PR #5, opened 2019-01-10
Update strings and dashboard html and js to display new version info.

EraYaN:
Tracks: https://github.com/jellyfin/jellyfin/pull/508


joshuaboniface:
@EraYaN With the final results of #508, I don't think this is required anymore - can you confirm?

nvllsvm:
https://github.com/jellyfin/jellyfin/pull/508 does indeed negate it

Issue #11, opened 2019-01-10
forgotpasswordpin submit retun page when entering pin 


grafixeyehero:
![login](https://user-images.githubusercontent.com/32230989/50998985-30a64400-153a-11e9-8907-311ec13a2464.jpg)

JustAMan:
Your fix is still not in public release (and not in `master`).
Please close only when the fix is merged to `master`, which should happen really soon.

grafixeyehero:
sorry

PR #12, opened 2019-01-10
fix return page to index.html #11

grafixeyehero:
bfayers commented on
https://github.com/jellyfin/jellyfin/pull/544#issuecomment-453202551


joshuaboniface:
@grafixeyehero Thanks for your first PR! I've updated the description, are you able to confirm if that is correct?

JustAMan:
@grafixeyehero I've fixed merge conflicts for you, I hope you don't object :)

PR #13, opened 2019-01-11
send android clients to select server before login

dkanada:
This will also need to be implemented for the standalone web client, since `login` assumes an ApiClient has been initialized but that isn't the case until `selectserver` has been called. The behavior is probably different when a user is logged in, but this would solve the inital login issue. We should use a new variable like `isBundledWithServer` to determine if a server must be selected first, and if `isNativeApp` is only used for this purpose we can replace it in favor of a unified variable.

### DIFF
--- a/src/scripts/site.js
+++ b/src/scripts/site.js
@@ -62,7 +62,8 @@ var Dashboard = {
         },
         logout: function(logoutWithServer) {
             function onLogoutDone() {
-                Dashboard.navigate("login.html")
+                var loginPage;
+                AppInfo.isNativeApp ? (loginPage = "selectserver.html", window.ApiClient = null) : loginPage = "login.html", Dashboard.navigate(loginPage)
             }!1 === logoutWithServer ? onLogoutDone() : ConnectionManager.logout().then(onLogoutDone)
         },
         getConfigurationPageUrl: function(name) {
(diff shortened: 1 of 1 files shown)


EraYaN:
We can absolutely can let any http server send a cookie or something like that to help it switch. Looks nice for now.

src/scripts/site.js
```diff
@@ -62,7 +62,8 @@ var Dashboard = {
         },
         logout: function(logoutWithServer) {
             function onLogoutDone() {
-                Dashboard.navigate("login.html")
+                var loginPage;
+                AppInfo.isNativeApp ? (loginPage = "selectserver.html", window.ApiClient = null) : loginPage = "login.html", Dashboard.navigate(loginPage)
```
JustAMan: can we do less poetic code please? :) like
```suggestion
                if (AppInfo.isNativeApp) {
                    loginPage = "selectserver.html";
                    window.ApiClient = null;
                } else {
                    loginPage = "login.html";
                }
                Dashboard.navigate(loginPage);
```

src/scripts/site.js
```diff
@@ -1033,7 +1034,7 @@ var Dashboard = {
             }, appRouter.showSelectServer = function() {
                 AppInfo.isNativeApp ? Dashboard.navigate("selectserver.html") : Dashboard.navigate("login.html")
             }, appRouter.showWelcome = function() {
-                Dashboard.navigate("login.html")
+                AppInfo.isNativeApp ? Dashboard.navigate("selectserver.html") : Dashboard.navigate("login.html")
```
JustAMan: If you want a ternary why so big?
```suggestion
                Dashboard.navigate(AppInfo.isNativeApp ? "selectserver.html" : "login.html")
```

src/scripts/site.js
dkanada: good change :) this change was actually copied over from an old working version so I didn't think to update the readability

src/scripts/site.js
dkanada: same here, pulled from older working version

PR #14, opened 2019-01-11
Reimplement page to add server

dkanada:
This functionality was in `connectlogin` so I stripped out all connect related code and renamed the files to `addserver` for clarity. It still needs a small amount of work to deminify since #1 won't catch this file.


src/scripts/addserver.js
```diff
@@ -0,0 +1,65 @@
+define(["appSettings", "loading", "browser", "emby-linkbutton"], function(appSettings, loading, browser) {
+    "use strict";
+
+    function handleConnectionResult(page, result) {
+        loading.hide();
+        switch (result.State) {
+            case "SignedIn":
+                var apiClient = result.ApiClient;
+                Dashboard.onServerChanged(apiClient.getCurrentUserId(), apiClient.accessToken(), apiClient), Dashboard.navigate("home.html");
```
JustAMan: Let's split in two lines
```suggestion
                Dashboard.onServerChanged(apiClient.getCurrentUserId(), apiClient.accessToken(), apiClient);
                Dashboard.navigate("home.html");
```

src/scripts/addserver.js
```diff
@@ -0,0 +1,65 @@
...
+                break;
+            case "ServerSignIn":
+                Dashboard.navigate("login.html?serverid=" + result.Servers[0].Id, false, "none");
+                break;
+            case "ServerSelection":
```
JustAMan: is this case really needed?

dkanada: I honestly don't know, I decided to leave it in until we figure out whether or not it's required. This return value is referenced a few times, not just here.

PR #16, opened 2019-01-11
Fix spinning circle at the end of config wizard.

drago-96:
This PR should fix jellyfin/jellyfin#405.
I added some functions to redirect to an absolute url.


drago-96:
Should be fixed now.

PR #20, opened 2019-01-11
Show album thumbnail and artist image in page itemdetail

### LINKED ISSUE #15: missing thumbnail on album view for mobile screen size
The android client has always been missing a nice thumbnail on `itemdetails` for albums similar to movies and shows. I will look into this once I have audio playback functional again.

drago-96:
Fixes #15.
Now the detail pages of albums and artists on mobile have the correct images.

### DIFF
--- a/src/scripts/itemdetailpage.js
+++ b/src/scripts/itemdetailpage.js
@@ -252,7 +252,7 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "cardBuild
         var imgUrl, screenWidth = screen.availWidth,
             hasbackdrop = !1,
             itemBackdropElement = page.querySelector("#itemBackdrop"),
-            usePrimaryImage = "Video" === item.MediaType && "Movie" !== item.Type && "Trailer" !== item.Type || item.MediaType && "Video" !== item.MediaType;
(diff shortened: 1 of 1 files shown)


src/scripts/itemdetailpage.js
```diff
@@ -252,7 +252,7 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "cardBuild
         var imgUrl, screenWidth = screen.availWidth,
             hasbackdrop = !1,
             itemBackdropElement = page.querySelector("#itemBackdrop"),
-            usePrimaryImage = "Video" === item.MediaType && "Movie" !== item.Type && "Trailer" !== item.Type || item.MediaType && "Video" !== item.MediaType;
+            usePrimaryImage = ("Video" === item.MediaType && "Movie" !== item.Type && "Trailer" !== item.Type) || (item.MediaType && "Video" !== item.MediaType) || ("MusicAlbum" === item.Type) || ("MusicArtist" === item.Type);
```
JustAMan: Care to explain why second part checks for `item.MediaType` without any comparison? Is it just to check if it is present at all (like not `undefined`)?

drago-96: I think yes: not every item object has the MediaType attribute.
Probably removing the check of `item.MediaType` would also work as a fix of this particular issue.

The right thing to do should be fixing the API endpoints to return consistent objects...

PR #28, opened 2019-01-12
Add transparent nav bar to BlueRadiance theme CSS

grafixeyehero:
left nav bar update to transparent


grafixeyehero:
@JustAMan this is the old
left panal backend server
left panal front

EraYaN:
Aah yes, much better

PR #30, opened 2019-01-13
Remove iap and simplify registrationservices

dkanada:


cvium:
LGTM

Is it necessary to keep registrationservices though?

dkanada:
At the moment those two methods in `registrationservices` are used all over the place and I was mainly focused on removing `iap` and the cordova dependencies, so I left them in for now.

cvium:
Might be a good idea to create an issue and/or add a TODO. I'm guessing we eventually want to remove the service entirely?

Issue #3, opened 2019-01-10
Determine versioning policies


joshuaboniface:
AFAICT there's no versioning in this repo, so it should closely track the main repo. Open to suggestions on how to accomplish this.

EraYaN:
This repo should mirror the main repo's versions and tags. So we should make a release checklist, and follow that every release.

hawken93:
I'd consider this and the jellyfin repo part of the same release, we even have to redirect to the deb files on the jellyfin repo on release

joshuaboniface:
The release checklist is a very good idea. I agree that they should be the same "version" across both, so that any release stuff is consistent. I'm not sure however if there's a good way to automate that.

hawken93:
Ansible it! :D

EraYaN:
Well if python wasn't an okay build dep...

POSIX shell! (or bash now) ;)

Ansible seems almost overkill

hawken93:
I'm not sure ansible would really be all that good at it.
anyway, it looks a bit like:

- make sure jellyfin points at the right commit in jellyfin-web
- tag both
- make release in jellyfin
I think that's mostly it

joshuaboniface:
Actually, we can make use of Org-level projects to help with this, but it doesn't directly help with versioning *per se*:

https://github.com/orgs/jellyfin/projects

We could make release projects there, include issues/PRs from the multiple relevant repos (jellyfin, jellyfin-web, and any submodules), and go from there.

I noticed that most of the PRs in this repo are against Master, so we should stop doing that ASAP and do them against `dev`, so we can follow the same procedure in both repositories. A master release PR would then happen in both at the same time, tracked in the same project.

cc @anthonylavado as our resident project master.

dkanada:
I actually think they shouldn't stay locked together, since the eventual goal is to be able to host the frontend separately if you want. The web frontend is already pretty close and we will need some backend changes to not serve the frontend at all with a toggle setting. The same argument that people were using for other clients (what if the frontend gets a major redesign or rewrite to Vue or React that modifies behavior and should get a major version bump but the server hasn't had enough changes) still stands for this repository.

joshuaboniface:
@dkanada I agree with you *one we reach full separation*, but for the near future they should closely track each other as much as possible, given how tightly integrated they are. That's something none of the other apps need to really deal with.

[...]

PR #33, opened 2019-01-13
WIP: Added basic webpack implementation.

EraYaN:
So this add a CopyPlugin based webpack.config.js

Since it won't just bundle site.js due to broken requires.

If anyone wants to have a look...

`yarn install` to install all deps.

`./setup-env.sh` to add the node_modules/.bin folder to path and `./webpack-prod.sh` to build production version.


package.json
```diff
@@ -0,0 +1,62 @@
+{
+    "name": "jellyfin-web",
+    "description": "",
+    "version": "10.0.1",
```
nvllsvm: Use `0.0.0` until a versioning scheme is defined in https://github.com/jellyfin/jellyfin-web/issues/3

package.json
```diff
@@ -0,0 +1,62 @@
...
+    "scripts": {
+        "test": "echo \"Error: no test specified\" && exit 1"
+    },
+    "keywords": [],
+    "author": "Jellyfin Project",
+    "license": "ISC",
```
JustAMan: What does `ISC` mean? Shouldn't it be `GPLv2`?

tsconfig.json
```diff
@@ -0,0 +1,63 @@
+{
+  "compilerOptions": {
+    /* Basic Options */
+    "target": "es5",                          /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */
+    "module": "es2015",                     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
+    //"lib": [ "es6", "dom"],                             /* Specify library files to be included in the compilation. */
+    "allowJs": true,                       /* Allow javascript files to be compiled. */
+    "checkJs": false,                       /* Report errors in .js files. */
```
JustAMan: why not `true`?

tsconfig.json
```diff
@@ -0,0 +1,63 @@
...
+    // "jsx": "preserve",                     /* Specify JSX code generation: 'preserve', 'react-native', or 'react'. */
+    // "declaration": true,                   /* Generates corresponding '.d.ts' file. */
+    "sourceMap": true,                     /* Generates corresponding '.map' file. */
+    // "outFile": "./",                       /* Concatenate and emit output to single file. */
+    "outDir": "./src/js",                        /* Redirect output structure to the directory. */
```
JustAMan: care to explain why output to `src`, not something like `dst`?

webpack.config.js
```diff
@@ -0,0 +1,137 @@
...
+            { from: 'dashboard/**/*', to: './' , context: 'src/'},
+            { from: 'devices/**/*', to: './' , context: 'src/'},
+            { from: 'home/**/*', to: './' , context: 'src/'},
+            { from: 'legacy/**/*', to: './' , context: 'src/'},
+            { from: 'list/**/*', to: './' , context: 'src/'},
+            { from: 'thirdparty/**/*', to: './' , context: 'src/'},
+          ])
+    ],
+    optimization: {
+        minimizer: [
```
JustAMan: will this also minify stuff, or only uglify?

EraYaN: uglify will also minify.

tsconfig.json
EraYaN: Because webpack will then take the generated *.js and bundle it. At least I think so, it's been a while since I set this template up.

tsconfig.json
EraYaN: The js is so shit, that it will take a rewrite to ever make that linter pass.

package.json
EraYaN: Yes forgot to change.

nvllsvm:
Temporarily closing - dev branch is being removed in favor of master branch. Also shrinking repo. https://github.com/jellyfin/jellyfin/issues/311

Please reopen in 3 hours.

tsconfig.json
JustAMan: Isn't it going to destroy contents in `src`? Or am I misunderstanding the purpose of this file?

package.json
JustAMan: @EraYaN can you fix the license?

tsconfig.json
EraYaN: Well it will only take *.ts and make *.js in the same location. So this should not overwrite anything. Especially since there is no TypeScript in the repo right now.

JustAMan: I just don't like build systems which pollute the source directory for building :)

EraYaN:
This should be closed in favour of #343

PR #34, opened 2019-01-13
Revert "Add Master codeowners"

joshuaboniface:
Reverts jellyfin/jellyfin-web#24

We reverted this in the main repo as unnecessary. Revert it here too.


joshuaboniface:
CC @nvllsvm as due to this change we need your approval ;-)

PR #36, opened 2019-01-15
Open videos in fullscreen on android devices

dkanada:


dkanada:
A native solution is the best way for good support across devices. This may be changed in the future depending on web support, I noticed some code in there to manage fullscreen state but it didn't work when I tried enabling it for android devices.

PR #37, opened 2019-01-15
Remove broken features from web interface

dkanada:
Several components in the javascript API don't have any source code available, including [wakeonlan](https://github.com/jellyfin/jellyfin-apiclient-javascript/blob/master/wakeonlan.js), [serverdiscovery](https://github.com/jellyfin/jellyfin-apiclient-javascript/blob/master/serverdiscovery.js), [cameraroll](https://github.com/jellyfin/jellyfin-apiclient-javascript/blob/master/cameraroll.js), and [fileupload](https://github.com/jellyfin/jellyfin-apiclient-javascript/blob/master/fileupload.js), which I assume was supposed to be a part of the `cameraroll` feature. We might want to implement `serverdiscovery` eventually, but `wakeonlan` and `cameraroll` should be removed. I would also like to remove all sync related functionality and simply extend the download feature for mobile as well. I will list some reasons for these actions below that I mentioned in chat, but they are all my personal opinions so everything is open to discussion. My primary reason for removing these features is to make the code easier to maintain without losing any functionality.

1. wakeonlan is bloat, it shouldn't be handled by a web client, and has no implementation in the API so it is useless at the moment.
2. cameraroll shouldn't be in the core functionality of this program, a media server should serve media, not upload media. I have no issues with a library type for photos, but there should be no option to upload any media.
3. sync has support in the API but no support in any client apps, and it shouldn't be handled by the server at all. The current interface for sync doesn't mesh at ALL with the native apps, since the interface (jellyfin-web) was designed for online use and any attempt at a bundled offline option (included in jellyfin-web) will not feel polished. This should be handled by clients themselves either with a portion of native code or an interface that was designed from the start to be used offline as well.
4. all of these can be supported with third party plugins. If someone wants to make them for their own client I have no problems with that.

This is also a good time to ask what we are doing to sync the jellyfin-apiclient-javascript between its own repository and this project. I don't want to make changes here and then have to manually edit the other repository in the future.


cvium:
> This is also a good time to ask what we are doing to sync the jellyfin-apiclient-javascript between its own repository and this project. I don't want to make changes here and then have to manually edit the other repository in the future.

We should probably do a hard fork from https://github.com/jellyfin/jellyfin-apiclient-javascript/commit/c390d773fe48539a6c989c962bbd4b6d2691b744 and publish it to npm or whatever popular package manager is used these days. Then import with <favorite package manager>.

hawken93:
Looking good :)

src/bower_components/emby-webcomponents/homesections/homesections.js
```diff
@@ -273,35 +273,12 @@
...
-
-        elem.querySelector('.appInfoSection').addEventListener('click', function (e) {
-
-            if (dom.parentWithClass(e.target, 'card')) {
-                registrationServices.showPremiereInfo();
-            }
-        });
+        elem.innerHTML = html;
+        bindHomeScreenSettingsIcon(elem, apiClient, user.Id, userSettings);
+        imageLoader.lazyChildren(elem);
```
cvium: I think this needs to return a promise? Quickly tracing the calls it looks like the result of this function is pushed to an array of promises.

src/bower_components/emby-webcomponents/homesections/homesections.js
```diff
@@ -699,21 +510,9 @@
...
-
-            if (infoHtml) {
-                bindAppInfoEvents(elem);
-            }
-
-            imageLoader.lazyChildren(elem);
-        });
+        elem.innerHTML = html;
+        bindHomeScreenSettingsIcon(elem, apiClient, user.Id, userSettings);
+        imageLoader.lazyChildren(elem);
```
cvium: Same as above

src/bower_components/emby-webcomponents/homesections/homesections.js
dkanada: done, thanks for catching this

src/bower_components/emby-webcomponents/homesections/homesections.js
dkanada: done

src/bower_components/emby-webcomponents/homesections/homesections.js
```diff
@@ -273,37 +273,16 @@
     }
 
     function loadlibraryButtons(elem, apiClient, user, userSettings, userViews) {
-
-        return Promise.all([getAppInfo(apiClient), getDownloadsSectionHtml(apiClient, user, userSettings)]).then(function (responses) {
-
-            var infoHtml = responses[0];
-            var downloadsHtml = responses[1];
-
+        return new Promise(function (resolve, reject) {
```
cvium: Hmm, there's nothing to resolve these new promises unless I'm missing something.

dkanada: Yeah, turns out I had it right the first time. The promises were only made to call some async methods and they were resolved right here. Without the method calls I removed the promises aren't necessary, I'll remove them in a minute.

cvium: The way I see it there's two solutions. Removing the promises and not adding them to the promise array (this one you missed), or resolving them immediately. I prefer the former. Sorry I didn't suggest that instead...

dkanada: Isn't that what I did the first time though? I removed the promises and didn't return any values from those two methods.

dkanada: Check the current source for my original solution, I added it again.

cvium: https://github.com/jellyfin/jellyfin-web/pull/37/files#diff-953f15eac453b39554bd82baac54837eR151

dkanada: ahhhh, okay. I'll fix it properly now haha

src/scripts/mypreferencescommon.js
```diff
@@ -1,16 +1,30 @@
...
+            var page = this;
+            var userId = params.userId || Dashboard.getCurrentUserId();
+
+            page.querySelector(".lnkDisplayPreferences").setAttribute("href", "mypreferencesdisplay.html?userId=" + userId);
+            page.querySelector(".lnkLanguagePreferences").setAttribute("href", "mypreferenceslanguages.html?userId=" + userId);
+            page.querySelector(".lnkSubtitleSettings").setAttribute("href", "mypreferencessubtitles.html?userId=" + userId);
+            page.querySelector(".lnkHomeScreenPreferences").setAttribute("href", "mypreferenceshome.html?userId=" + userId);
+            page.querySelector(".lnkMyProfile").setAttribute("href", "myprofile.html?userId=" + userId);
+
+            appHost.supports("multiserver") ? page.querySelector(".selectServer").classList.remove("hide") : page.querySelector(".selectServer").classList.add("hide");
```
JustAMan: de-uglify more?
```suggestion
            if (appHost.supports("multiserver")) {
                page.querySelector(".selectServer").classList.remove("hide")
            } else {
                page.querySelector(".selectServer").classList.add("hide");
            }
```

src/scripts/mypreferencescommon.js
```diff
@@ -1,16 +1,30 @@
...
+            connectionManager.user(ApiClient).then(function(user) {
+                user.localUser && !user.localUser.EnableAutoLogin ? view.querySelector(".btnLogout").classList.remove("hide") : view.querySelector(".btnLogout").classList.add("hide");
```
JustAMan:
```suggestion
                if (user.localUser && !user.localUser.EnableAutoLogin) {
                    view.querySelector(".btnLogout").classList.remove("hide");
                } else {
                    view.querySelector(".btnLogout").classList.add("hide");
                }
```

src/scripts/mypreferencescommon.js
```diff
@@ -1,16 +1,30 @@
...
+            });
+
+            Dashboard.getCurrentUser().then(function(user) {
+                page.querySelector(".headerUser").innerHTML = user.Name;
+                user.Policy.IsAdministrator ? page.querySelector(".adminSection").classList.remove("hide") : page.querySelector(".adminSection").classList.add("hide");
```
JustAMan:
```suggestion
                if (user.Policy.IsAdministrator) {
                    page.querySelector(".adminSection").classList.remove("hide");
                } else {
                    page.querySelector(".adminSection").classList.add("hide");
                }
```

src/scripts/selectserver.js
```diff
@@ -215,15 +116,13 @@ define(["loading", "appRouter", "layoutManager", "appSettings", "apphost", "focu
...
+        updatePageStyle(view, params);
         var backdropUrl = staticBackdrops.getRandomImageUrl();
         view.addEventListener("viewshow", function(e) {
             var isRestored = e.detail.isRestored;
-            appRouter.setTitle(null), backdrop.setBackdrop(backdropUrl), isRestored || (loadServers(), loadInvitations())
-        }), view.querySelector(".btnOffline").addEventListener("click", function(e) {
-            appRouter.show("/offline/offline.html")
+            appRouter.setTitle(null);
+            backdrop.setBackdrop(backdropUrl);
+            isRestored || loadServers();
```
JustAMan:
```suggestion
            if (!isRestored) loadServers();
```

src/scripts/site.js
```diff
@@ -285,12 +285,70 @@ var Dashboard = {
...
+        var bowerPath = getBowerPath();
+        var apiClientBowerPath = bowerPath + "/emby-apiclient";
+        var embyWebComponentsBowerPath = bowerPath + "/emby-webcomponents";
+
+        "android" === self.appMode
+            ? (define("filesystem", ["cordova/filesystem"], returnFirstDependency), define("cameraRoll", ["cordova/cameraroll"], returnFirstDependency))
```
JustAMan: you can drop `cameraroll` here

src/scripts/site.js
```diff
@@ -285,12 +285,70 @@ var Dashboard = {
...
+        var bowerPath = getBowerPath();
+        var apiClientBowerPath = bowerPath + "/emby-apiclient";
+        var embyWebComponentsBowerPath = bowerPath + "/emby-webcomponents";
+
+        "android" === self.appMode
```
JustAMan: can you de-uglify this to `if/else` instead of this ternary things?

src/scripts/site.js
```diff
@@ -285,12 +285,70 @@ var Dashboard = {
...
+        "cordova" === self.appMode || "android" === self.appMode
+            ? (define("apiclientcore", ["bower_components/emby-apiclient/apiclient"], returnFirstDependency), define("apiclient", ["bower_components/emby-apiclient/apiclientex"], returnFirstDependency))
+            : define("apiclient", ["bower_components/emby-apiclient/apiclient"], returnFirstDependency)
+        define("actionsheet", ["webActionSheet"], returnFirstDependency);
+        "registerElement" in document
+            ? define("registerElement", [])
+            : browser.msie
+                ? define("registerElement", [bowerPath + "/webcomponentsjs/webcomponents-lite.min.js"], returnFirstDependency)
+                : define("registerElement", [bowerPath + "/document-register-element/build/document-register-element"], returnFirstDependency);
+        "android" === self.appMode
```
JustAMan: can be combined like 
```js
"android" === self.appMode || "cordova" === self.appMode
```

src/scripts/site.js
JustAMan: Let's do it in a subsequent PR

src/bower_components/emby-webcomponents/homesections/homesections.js
dkanada: How about my latest commit? Removed the return statement so it just loads normally.

cvium:
Nice. LGTM now.

PR #39, opened 2019-01-16
Remove back button from dashboard and metadata manager

sparky8251:
Since the back button on the server settings dashboard stupidly moved you back through the list of recently selected server settings before going back to the home page, it was confusing.

The back button on the metadata manager page acted even more stupidly. It was functionally identical to the home button just to its right.

This pull deuglifies the library responsible for the back button and then fixes this inconsistent bit of UI behavior.

Best to test this, I've done what I can, but since I don't use all library types its possible my detection method will catch incorrect pages thus removing their back button.

Tested:
Server Dashboard
Metadata Manager
Home screen
Complete set of views for TV Shows and Movies including playback page

No regressions so far.

Now looks like this:
screenshot_20190116_111854
screenshot_20190116_111908

No more back button that does something stupid or acts just like the nearby home button.


src/scripts/librarymenu.js
```diff
@@ -155,297 +302,671 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
...
+        menuHtml += '<a is="emby-linkbutton" class="navMenuOption" href="' + item.href + '"' + pageIds + pageUrls + ">";
+
+        if (item.icon) {
+            menuHtml += '<i class="md-icon navMenuOptionIcon">' + item.icon + "</i>";
+        }
+
+        menuHtml += '<span class="navMenuOptionText">';
+        menuHtml += item.name;
+        menuHtml += "</span>";
+        return menuHtml += "</a>";
```
cvium: += is pointless
```suggestion
        return menuHtml + "</a>";
```

src/scripts/librarymenu.js
```diff
@@ -155,297 +302,671 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
...
-            for (menuHtml += '<div class="drawerContent">', i = 0, length = items.length; i < length; i++) item = items[i], item.href ? menuHtml += getToolsLinkHtml(item) : item.name && (menuHtml += '<h3 class="sidebarHeader">', menuHtml += item.name, menuHtml += "</h3>");
-            return menuHtml += "</div>"
-        })
+        return getToolsMenuLinks(apiClient).then(function (items) {
+            var i;
+            var length;
+            var item;
+            var menuHtml = "";
+
+            for (menuHtml += '<div class="drawerContent">', i = 0, length = items.length; i < length; i++) {
```
cvium: Deuglify fail
```suggestion
            menuHtml += '<div class="drawerContent">';
            for (i = 0, length = items.length; i < length; i++) {
```

src/scripts/librarymenu.js
```diff
@@ -155,297 +302,671 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
...
+                item = items[i];
+
+                if (item.href) {
+                    menuHtml += getToolsLinkHtml(item);
+                } else {
+                    if (item.name) {
```
cvium: The if shouldn't be nested here.

src/scripts/librarymenu.js
```diff
@@ -155,297 +302,671 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
...
+                        menuHtml += '<h3 class="sidebarHeader">';
+                        menuHtml += item.name;
+                        menuHtml += "</h3>";
+                    }
+                }
+            }
+
+            return menuHtml += "</div>";
```
cvium:
```suggestion
            return menuHtml + "</div>";
```

src/scripts/librarymenu.js
```diff
@@ -155,297 +302,671 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
...
+                var items = result;
+                var html = "";
+                html += '<h3 class="sidebarHeader">';
+                html += globalize.translate("HeaderMedia");
+                html += "</h3>";
+                html += items.map(function (i) {
+                    var icon = "folder";
+                    var itemId = i.Id;
+
+                    if ("channels" == i.CollectionType) {
```
cvium: All these double equals should be triple eg. `"channels" === i.CollectionType` since we know it's a string (it's just convention really)

src/scripts/librarymenu.js
```diff
@@ -155,297 +302,671 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
...
+
+                    return '<a is="emby-linkbutton" data-itemid="' + itemId + '" class="lnkMediaFolder navMenuOption" href="' + getItemHref(i, i.CollectionType) + '"><i class="md-icon navMenuOptionIcon">' + icon + '</i><span class="sectionName navMenuOptionText">' + i.Name + "</span></a>";
+                }).join("");
+                libraryMenuOptions.innerHTML = html;
+                var elem = libraryMenuOptions;
+                var sidebarLinks = elem.querySelectorAll(".navMenuOption");
+
+                for (var i = 0, length = sidebarLinks.length; i < length; i++) {
+                    sidebarLinks[i].removeEventListener("click", onSidebarLinkClick);
+                    sidebarLinks[i].addEventListener("click", onSidebarLinkClick);
```
cvium: What's the point of this?

src/scripts/librarymenu.js
```diff
@@ -155,297 +302,671 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
...
+        var i;
+        var length;
+        var isLiveTvPage = page.classList.contains("liveTvPage");
+        var isChannelsPage = page.classList.contains("channelsPage");
+        var isEditorPage = page.classList.contains("metadataEditorPage");
+        var isMySyncPage = page.classList.contains("mySyncPage");
+        var id = isLiveTvPage || isChannelsPage || isEditorPage || isMySyncPage || page.classList.contains("allLibraryPage") ? "" : getTopParentId() || "";
+        var elems = document.getElementsByClassName("lnkMediaFolder");
+
         for (i = 0, length = elems.length; i < length; i++) {
```
cvium: Should just move the variable declaration in here.

src/scripts/librarymenu.js
```diff
@@ -155,297 +302,671 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
...
-            var lnkMediaFolder = elems[i],
-                itemId = lnkMediaFolder.getAttribute("data-itemid");
-            isChannelsPage && "channels" == itemId ? lnkMediaFolder.classList.add("navMenuOption-selected") : isLiveTvPage && "livetv" == itemId ? lnkMediaFolder.classList.add("navMenuOption-selected") : isEditorPage && "editor" == itemId ? lnkMediaFolder.classList.add("navMenuOption-selected") : isMySyncPage && "manageoffline" == itemId && -1 != window.location.href.toString().indexOf("mode=download") ? lnkMediaFolder.classList.add("navMenuOption-selected") : isMySyncPage && "syncotherdevices" == itemId && -1 == window.location.href.toString().indexOf("mode=download") ? lnkMediaFolder.classList.add("navMenuOption-selected") : id && itemId == id ? lnkMediaFolder.classList.add("navMenuOption-selected") : lnkMediaFolder.classList.remove("navMenuOption-selected")
+            var lnkMediaFolder = elems[i];
+            var itemId = lnkMediaFolder.getAttribute("data-itemid");
+
+            if (isChannelsPage && "channels" == itemId) {
```
cvium: Again, should be triple equals imo.

src/scripts/librarymenu.js
```diff
@@ -155,297 +302,671 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
...
     }
 
     function updateTitle(page) {
         var title = page.getAttribute("data-title");
-        title ? LibraryMenu.setTitle(title) : page.classList.contains("standalonePage") && LibraryMenu.setDefaultTitle()
+
+        if (title) {
+            LibraryMenu.setTitle(title);
+        } else {
+            if (page.classList.contains("standalonePage")) {
```
cvium: Should not be nested

src/scripts/librarymenu.js
```diff
@@ -155,297 +302,671 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
...
     function updateBackButton(page) {
-        headerBackButton || (headerBackButton = document.querySelector(".headerBackButton")), headerBackButton && ("false" !== page.getAttribute("data-backbutton") && appRouter.canGoBack() ? headerBackButton.classList.remove("hide") : headerBackButton.classList.add("hide"))
+        var isDashboardPage = page.classList.contains("type-interior");
+        
+        if (!headerBackButton) {
+            headerBackButton = document.querySelector(".headerBackButton");
+        }
+
+        if (headerBackButton) {
+            if ("false" !== !isDashboardPage && page.getAttribute("data-backbutton") && appRouter.canGoBack()) {
```
cvium: Hmm. Isn't `isDashboardPage` a boolean? `"false"` is a string, so they will never be equal.
```suggestion
            if (!isDashboardPage && page.getAttribute("data-backbutton") && appRouter.canGoBack()) {
```

src/scripts/librarymenu.js
```diff
@@ -155,297 +302,671 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
...
+        });
     }
 
     function refreshLibraryDrawer(user) {
-        loadNavDrawer(), currentDrawerType = "library", (user ? Promise.resolve(user) : connectionManager.user(getCurrentApiClient())).then(function(user) {
-            refreshLibraryInfoInDrawer(user), updateLibraryMenu(user.localUser)
-        })
+        loadNavDrawer();
+        currentDrawerType = "library";
+        (user ? Promise.resolve(user) : connectionManager.user(getCurrentApiClient())).then(function (user) {
```
cvium: Deuglify fail. Should just be 
```
if (user) {
    Promise.resolve(user);
} else {
    doOtherStuff();
}
```

src/scripts/librarymenu.js
```diff
@@ -155,297 +302,671 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
...
+    var headerBackButton;
+    var headerUserButton;
+    var currentUser;
+    var headerSettingsButton;
+    var headerCastButton;
+    var headerSearchButton;
+    var enableLibraryNavDrawer = !layoutManager.tv;
+    var skinHeader = document.querySelector(".skinHeader");
+    var requiresUserRefresh = true;
+    var lastOpenTime = new Date().getTime();
```
cvium: Jesus, what's with all these global variables...

src/scripts/librarymenu.js
```diff
@@ -155,297 +302,671 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
...
+        headerCastButton = skinHeader.querySelector(".headerCastButton");
+        headerSearchButton = skinHeader.querySelector(".headerSearchButton");
+
+        if (!browser.chrome) {
+            skinHeader.classList.add("skinHeader-blurred");
+        }
+
+        lazyLoadViewMenuBarImages();
+        bindMenuEvents();
+    })();
```
cvium: Not a fan of a random self-invoking function. Out of the scope of this PR though imo.

src/scripts/librarymenu.js
sparky8251: And yet, this is the single bit of code I had to write to make the back button not appear on the dashboard.

Think it has to do with the getter. If its there, it returns true. Its used further up in the original code, hence why I used this to fix it up.

I can try the alternative you suggest, but changing that line has resulted in really stupid outcomes like total UI failure.

src/scripts/librarymenu.js
sparky8251: Fuck if I know. This is just the deuglify with my single fix. Didn't delve too deep into what unrelated parts of code did.

src/scripts/librarymenu.js
sparky8251: What do you mean?

src/scripts/librarymenu.js
sparky8251: It isnt? Did you check the latest commit or one of the earlier ones? I broke it up into pieces, looking at my code this isn't actually in the last commit.

src/scripts/librarymenu.js
sparky8251:
```
        if (user) {
            Promise.resolve(user);
        }
        else {
            connectionManager.user(getCurrentApiClient())).then(function (user);
            refreshLibraryInfoInDrawer(user);
            updateLibraryMenu(user.localUser);
        });
```

?

src/scripts/librarymenu.js
sparky8251: Not entirely sure. Touching most anything in this file completely breaks the UI though, so I'm not keen on the idea.

src/scripts/librarymenu.js
cvium: GitHub had issues when I reviewed it I think. I was looking at the total changes 🤷‍♂️

cvium: This is what I see
image

It should be `else if` and not `else { if {` (forgive the formatting).

src/scripts/librarymenu.js
cvium: `for (var i = 0; i < elems.length; i++) {` and then remove line 638 and 639.

src/scripts/librarymenu.js
cvium: Same as above, I still see this as nested under an else?

src/scripts/librarymenu.js
cvium: All I can say is, that `"false" !== !false` will always be false. The triple equality check matches the type as well as the value and a string and a boolean are not the same type.

sparky8251: well... `"false" !== page.getAttribute("data-backbutton") && appRouter.canGoBack()` is what it was before i changed it, and removing the `"false !=="` broke the function entirely.

Adding in `!isDashboardPage` at the beginning not only keeps it going for non-dashboard pages, it exempts the two I wanted to from having the back button shown.

JustAMan:
I'm reading this, pls don't merge before I finish.

cvium:
@JustAMan It's not ready yet though

sparky8251:
Yup. Still got to sit down and do cviums idea of html manipulation + a few more of his cleanups.

src/scripts/librarymenu.js
```diff
@@ -155,297 +302,669 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
...
-            }).join(""), libraryMenuOptions.innerHTML = html;
-            for (var elem = libraryMenuOptions, sidebarLinks = elem.querySelectorAll(".navMenuOption"), i = 0, length = sidebarLinks.length; i < length; i++) sidebarLinks[i].removeEventListener("click", onSidebarLinkClick), sidebarLinks[i].addEventListener("click", onSidebarLinkClick)
-        })
+        if (!user) {
+            showBySelector(".libraryMenuDownloads", false);
+            showBySelector(".lnkSyncToOtherDevices", false);
+            return void showBySelector(".userMenuOptions", false);
+        }
+
+        if (user.Policy.EnableContentDownloading) {
```
JustAMan: this could probably be replaced with
```js
showBySelector(".lnkSyncToOtherDevices", !!user.Policy.EnableContentDownloading);
```

src/scripts/librarymenu.js
```diff
@@ -155,297 +302,669 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
...
+            showBySelector(".lnkSyncToOtherDevices", true);
+        } else {
+            showBySelector(".lnkSyncToOtherDevices", false);
+        }
+
+        if (user.Policy.EnableContentDownloading && appHost.supports("sync")) {
```
JustAMan: same, more or less

src/scripts/librarymenu.js
```diff
@@ -155,297 +302,669 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
...
+                    else if ("tvshows" == i.CollectionType) {
+                        icon = "tv";
+                    }
+                    else if ("livetv" == i.CollectionType) {
+                        icon = "live_tv";
+                    }
+                    icon = i.icon || icon;
+
+                    if (i.onclick) {
+                        i.onclick;
```
JustAMan: what is this?..

src/scripts/librarymenu.js
```diff
@@ -155,297 +302,669 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
...
+                lnkMediaFolder.classList.add("navMenuOption-selected");
+            }
+            else if (isMySyncPage && "manageoffline" == itemId && -1 != window.location.href.toString().indexOf("mode=download")) {
+                lnkMediaFolder.classList.add("navMenuOption-selected");
+            }
+            else if (isMySyncPage && "syncotherdevices" == itemId && -1 == window.location.href.toString().indexOf("mode=download")) {
+                lnkMediaFolder.classList.add("navMenuOption-selected");
+            }
+            else if (id && itemId == id) {
+                lnkMediaFolder.classList.add("navMenuOption-selected");
```
JustAMan: all above `if`-s effect in same class added to `lnkMediaFolder`, maybe combine them all in one `if`?

src/scripts/librarymenu.js
```diff
@@ -155,297 +302,669 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
...
+        setTitle: function (title) {
+            if (null == title) {
+                return void LibraryMenu.setDefaultTitle();
+            }
+
+            if ("-" === title) {
+                title = "";
+            }
+
             var html = title;
```
JustAMan: I can see no difference from this point with `setDefaultTitle()`

src/scripts/librarymenu.js
JustAMan: This fix cannot be correct, I am with @cvium here. There should be some proper solution.

src/scripts/librarymenu.js
sparky8251: Possibly. Commented a FIXME and your suggestion. Personally unsure if its equivalent.

src/scripts/librarymenu.js
sparky8251: Same as previous. FIXME commented since its a bit beyond my current skills.

src/scripts/librarymenu.js
sparky8251: No idea. Used to be a huuuge amount of if/elses nested inside each other.

Just made a lazy equivalent to it. If you have a suggested change, I can merge it in.

src/scripts/librarymenu.js
sparky8251: They used to be a huge amount of nested if/elses. Unsure of how its actually executed. If you or someone else wants to suggest a change, I'll accept it.

Just did the lazy thing and undid 10+ layers of nesting :)

src/scripts/librarymenu.js
sparky8251: Not sure what you are hinting at here.

src/scripts/librarymenu.js
```diff
@@ -1,121 +1,272 @@
...
     }
 
     function onHeaderHomeButtonClick() {
-        Dashboard.navigate("home.html")
+        Dashboard.navigate("home.html");
     }
 
     function bindMenuEvents() {
-        (mainDrawerButton = document.querySelector(".mainDrawerButton")) && mainDrawerButton.addEventListener("click", toggleMainDrawer);
+         if (mainDrawerButton = document.querySelector(".mainDrawerButton")) {
```
JustAMan: this should be equivalent without in-`if` assignment:
```suggestion
        mainDrawerButton = document.querySelector(".mainDrawerButton");
        if (mainDrawerButton) {
```

sparky8251:
Well, as of now this method of fixing it is the cleanest with the least amount of unintended side effects. 

@cvium's idea (insert `data-backbutton="false"` into the html of pages that dont need the button and use that to evaluate) resulted in the back button appearing on the tv.html and movie.html page when it shouldn't have.

Several of @JustAMan's ideas resulted in entire failures of the UI.

Open to more ideas, but at this point this one seems "complete" to me.

src/scripts/librarymenu.js
JustAMan: I'm saying that this function seems to be a copy-paste of another one. But that should be cleaned up in subsequent PR anyway.

joshuaboniface:
Fixed up the conflicts using your changes only - most looked formatting related.

PR #43, opened 2019-01-19
Filter videos by audio stream language

taconaut:
Uses and exposes the audio language filtering being added with this PR: https://github.com/jellyfin/jellyfin/pull/624

It currently looks like this:
image

### DIFF
--- a/src/components/filterdialog/filterdialog.js
+++ b/src/components/filterdialog/filterdialog.js
@@ -21,6 +21,8 @@ define(["dialogHelper", "globalize", "connectionManager", "events", "browser", "
             return -1 != ("|" + (query.Tags || "") + "|").indexOf("|" + i + "|")
         }), renderOptions(context, ".yearFilters", "chkYearFilter", result.Years, function(i) {
             return -1 != ("," + (query.Years || "") + ",").indexOf("," + i + ",")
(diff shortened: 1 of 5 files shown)


taconaut:
@dkanada unfortunately not, I'm a mainlya backend guy. It would be handy to have a fallback onto a default language, probably en-US.

taconaut:
It would be nice to also localize the language code.
Does the backend have any localization so it could return a tuple containing the language code along with the localized language name? Or is everything done in the frontend? (I'm dicovering the project ;)

dkanada:
> Does the backend have any localization so it could return a tuple containing the language code along with the localized language name? Or is everything done in the frontend?

We have some very basic localization in the frontend, no clue about the backend. There must be some code somewhere to get the language name because I have seen it in a couple of locations in the server dashboard. One location is Dashboard -> Library -> Metadata -> Language if you want to do some digging. It would definitely be better to show the language name instead of the language code here.

Oliver-Hanikel:
I'm a bit confused. The PR was merged but I can't find the language option on my Jellyfin 10.6.4.
image

dkanada:
We had to revert this because the backend portion was never finished.

PR #46, opened 2019-01-20
minor changes to navbar elements

dkanada:
This fixes some broken elements that weren't merged correctly from my last pull request.

### DIFF
--- a/src/strings/en-US.json
+++ b/src/strings/en-US.json
@@ -138,7 +138,7 @@
   "ButtonShuffle": "Shuffle",
   "ButtonShutdown": "Shutdown",
   "ButtonSignIn": "Sign In",
-  "ButtonSignOut": "Sign out",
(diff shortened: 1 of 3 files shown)


src/strings/en-US.json
```diff
@@ -138,7 +138,7 @@
   "ButtonShuffle": "Shuffle",
   "ButtonShutdown": "Shutdown",
   "ButtonSignIn": "Sign In",
-  "ButtonSignOut": "Sign out",
+  "ButtonSignOut": "Sign Out",
   "ButtonSignUp": "Sign up",
```
Bond-009:
```suggestion
  "ButtonSignUp": "Sign Up",
```

dkanada: I think this needs to be removed anyway, it was probably only used for connect.

PR #51, opened 2019-01-21
remove more code for sync and camera roll

dkanada:
This doesn't need to go in until after 10.1.0 but there were some more files that needed to be pruned. I have tested this on Android and it still works fine.


src/bower_components/emby-apiclient/sync/serversync.js
```diff
@@ -12,11 +12,7 @@ define([], function() {
     }
 
     function uploadContent(connectionManager, server, options) {
-        return new Promise(function(resolve, reject) {
-            require(["contentuploader"], function(ContentUploader) {
-                (new ContentUploader).uploadImages(connectionManager, server).then(resolve, reject)
-            })
-        })
+        return new Promise().resolve();
```
JustAMan: shouldn't we `reject` here instead?

src/components/apphost.js
```diff
@@ -117,13 +117,38 @@ define(["appSettings", "browser", "events", "htmlMediaHelper"], function(appSett
...
-        }(); - 1 === supportedFeatures.indexOf("htmlvideoautoplay") && !1 !== supportsHtmlMediaAutoplay() && require(["autoPlayDetect"], function(autoPlayDetect) {
-        autoPlayDetect.supportsHtmlMediaAutoplay().then(function() {
-            appSettings.set(htmlMediaAutoplayAppStorageKey, "true"), supportedFeatures.push("htmlvideoautoplay"), supportedFeatures.push("htmlaudioautoplay")
-        }, function() {
-            appSettings.set(htmlMediaAutoplayAppStorageKey, "false")
-        })
+            return navigator.share && features.push("sharing");
```
JustAMan: Cannot be correct. This `return` completes after calling `features.push("sharing")`, so all the rest code is basically never going to execute.

src/scripts/site.js
```diff
@@ -202,61 +202,75 @@ var Dashboard = {
...
 
     function getSettingsBuilder(UserSettings, layoutManager, browser) {
-        return UserSettings.prototype.enableThemeVideos = function(val) {
-            return null != val ? this.set("enableThemeVideos", val.toString(), !1) : (val = this.get("enableThemeVideos", !1), val ? "false" !== val : !layoutManager.mobile && !browser.slow)
-        }, UserSettings
+        UserSettings.prototype.enableThemeVideos = function(val) {
+            return null != val
+                ? this.set("enableThemeVideos", val.toString(), !1)
+                : (val = this.get("enableThemeVideos", !1), val ? "false" !== val : !layoutManager.mobile && !browser.slow);
+        };
```
JustAMan: maybe de-uglify further?
```suggestion
        UserSettings.prototype.enableThemeVideos = function(val) {
            if (val != null) {
                return this.set('enableThemeVideos', val.toString(), false);
            } else {
                val = this.get('enableThemeVideos', false);
                if (val !== 'false') {
                    return !layoutManager.mobile;
                } else {
                    return !browser.slow;
                }
            }
        };
```

src/components/apphost.js
dkanada: Thanks for catching that one.

src/bower_components/emby-apiclient/sync/serversync.js
dkanada: Sure, I was just filling it with something since it will hopefully be removed soon.

src/scripts/site.js
dkanada: done, plus some extra in apphost

src/components/apphost.js
```diff
@@ -108,84 +108,118 @@ define(["appSettings", "browser", "events", "htmlMediaHelper"], function(appSett
...
+        (browser.tv || browser.xboxOne || browser.ps4 || browser.mobile) && features.push("physicalvolumecontrol");
+        browser.tv || browser.xboxOne || browser.ps4 || features.push("remotecontrol");
+        browser.operaTv || browser.tizen || browser.orsay || browser.web0s || browser.edgeUwp || features.push("remotevideo");
+        features.push("otherapppromotions");
+        features.push("targetblank");
+        browser.orsay || browser.tizen || browser.msie || !(browser.firefox || browser.ps4 || browser.edge || cueSupported()) || features.push("subtitleappearancesettings");
+        browser.orsay || browser.tizen || features.push("subtitleburnsettings");
+        browser.tv || browser.ps4 || browser.xboxOne || features.push("fileinput");
+        browser.chrome && features.push("chromecast");
+        return features;
```
hawken93: just a comment - this is not a blocker for merge:

This function could be deuglified further with if statements

dkanada: Yeah, I only deuglified it this much so GitHub will show the diff highlights when I remove things now.

src/components/apphost.js
```diff
@@ -108,84 +108,118 @@ define(["appSettings", "browser", "events", "htmlMediaHelper"], function(appSett
     }
 
     function onAppVisible() {
-        _isHidden && (_isHidden = !1, console.log("triggering app resume event"), events.trigger(appHost, "resume"))
+        isHidden && (isHidden = false, console.log("triggering app resume event"), events.trigger(appHost, "resume"));
```
JustAMan:
```suggestion
        if (isHidden) {
            isHidden = false;
            console.log("triggering app resume event");
            events.trigger(appHost, "resume");
        }
```

src/components/apphost.js
```diff
@@ -108,84 +108,118 @@ define(["appSettings", "browser", "events", "htmlMediaHelper"], function(appSett
...
-        }(); - 1 === supportedFeatures.indexOf("htmlvideoautoplay") && !1 !== supportsHtmlMediaAutoplay() && require(["autoPlayDetect"], function(autoPlayDetect) {
-        autoPlayDetect.supportsHtmlMediaAutoplay().then(function() {
-            appSettings.set(htmlMediaAutoplayAppStorageKey, "true"), supportedFeatures.push("htmlvideoautoplay"), supportedFeatures.push("htmlaudioautoplay")
-        }, function() {
-            appSettings.set(htmlMediaAutoplayAppStorageKey, "false")
-        })
+        isHidden || (isHidden = true, console.log("app is hidden"));
```
JustAMan:
```suggestion
        if (!isHidden) {
            isHidden = true;
            console.log("app is hidden");
        }
```

src/components/apphost.js
JustAMan: is it on purpose that you dropped both features (`multiserver` and `displaylanguage`) that depended on `AppInfo.isNativeApp`?

src/components/apphost.js
```diff
@@ -108,84 +108,118 @@ define(["appSettings", "browser", "events", "htmlMediaHelper"], function(appSett
...
+        browser.operaTv || browser.tizen || browser.orsay || browser.web0s || browser.edgeUwp || features.push("remotevideo");
+        features.push("otherapppromotions");
+        features.push("targetblank");
+        browser.orsay || browser.tizen || browser.msie || !(browser.firefox || browser.ps4 || browser.edge || cueSupported()) || features.push("subtitleappearancesettings");
+        browser.orsay || browser.tizen || features.push("subtitleburnsettings");
+        browser.tv || browser.ps4 || browser.xboxOne || features.push("fileinput");
+        browser.chrome && features.push("chromecast");
+        return features;
+    }();
+    if (supportedFeatures.indexOf("htmlvideoautoplay") === -1 && supportsHtmlMediaAutoplay() !=== false) {
```
JustAMan:
```suggestion
    if (supportedFeatures.indexOf("htmlvideoautoplay") === -1 && supportsHtmlMediaAutoplay() !== false) {
```
or am I missing a new Javascript comparison?..

src/components/apphost.js
```diff
@@ -108,84 +108,118 @@ define(["appSettings", "browser", "events", "htmlMediaHelper"], function(appSett
...
+        }
+    };
+
+    var doc = self.document;
     doc && (void 0 !== doc.visibilityState ? (visibilityChange = "visibilitychange", visibilityState = "hidden") : void 0 !== doc.mozHidden ? (visibilityChange = "mozvisibilitychange", visibilityState = "mozVisibilityState") : void 0 !== doc.msHidden ? (visibilityChange = "msvisibilitychange", visibilityState = "msVisibilityState") : void 0 !== doc.webkitHidden && (visibilityChange = "webkitvisibilitychange", visibilityState = "webkitVisibilityState"));
-    var _isHidden = !1;
-    return doc && doc.addEventListener(visibilityChange, function() {
+    var isHidden = false;
+    doc && doc.addEventListener(visibilityChange, function() {
```
JustAMan:
```suggestion
    if (doc) {
        doc.addEventListener(visibilityChange, function() {
```
etc.

src/components/apphost.js
```diff
@@ -108,84 +108,118 @@ define(["appSettings", "browser", "events", "htmlMediaHelper"], function(appSett
...
         document[visibilityState] ? onAppHidden() : onAppVisible()
-    }), self.addEventListener && (self.addEventListener("focus", onAppVisible), self.addEventListener("blur", onAppHidden)), appHost
+    }), self.addEventListener && (self.addEventListener("focus", onAppVisible), self.addEventListener("blur", onAppHidden));
```
JustAMan:
```suggestion
    });
    if (self.addEventListener) {
        self.addEventListener("focus", onAppVisible);
        self.addEventListener("blur", onAppHidden);
    }
```

src/components/apphost.js
dkanada: I don't think they will ever be true since this file is only loaded if the instance isn't a native app.

src/components/apphost.js
dkanada: Nope that definitely doesn't exist, I haven't tested this change in the web interface, only Android. I will probably start testing in web as well when I can run it separately from the server, but most of my changes will be platform agnostic. This being an exception :)

PR #56, opened 2019-01-22
use english for fallback translations and clean up language files

dkanada:
One step closer to getting a translation service up and running. We could probably find one that supports json right now if we wanted to start.


src/bower_components/emby-webcomponents/globalize.js
```diff
@@ -51,43 +47,37 @@ define(['connectionManager', 'userSettings', 'events'], function (connectionMana
...
             return Promise.resolve();
         }
 
         return loadTranslation(translationInfo.translations, culture).then(function (dictionary) {
-
             translationInfo.dictionaries[culture] = dictionary;
         });
     }
 
     function normalizeLocaleName(culture) {
```
cvium: Why is the normalization needed? And why only for 3 languages?

JustAMan: I would rather we stick to standard language codes, so far what I saw seems like a mumbo-jumbo of standard ones and invented by Emby...

dkanada: I was going to leave this for another day, it could break settings that rely on the current language codes. Up to you guys though, would be easy to just remove it and properly name the language files.

JustAMan: Can you just drop a `TODO` comment there, so we won't forget? Or open a tracking issue, up to you, just something so we won't lose it forever.

hawken93:
Awesome

PR #62, opened 2019-01-23
remove BOM

hawken93:
title says it all.. Simple sed on all files per https://unix.stackexchange.com/questions/381230/how-can-i-remove-the-bom-from-a-utf-8-file

.html 62
.js 81
.css 12


hawken93:
The BOM has been largely removed from the jellyfin repo, and I thought this was what we are going for

PR #75, opened 2019-01-26
User profile fixes

nvllsvm:
Removes the pointless dialog from appearing when clicking the add user button. 
image

### DIFF
--- a/src/scripts/userprofilespage.js
+++ b/src/scripts/userprofilespage.js
@@ -242,48 +242,11 @@ define(["loading", "dom", "globalize", "humanedate", "paper-icon-button-light", 
         });
     }
 
-    function showNewUserDialog(e__w) {
-        require(["dialog"], function (dialog) {
-            var items = [];
-            items.push({
-                name: globalize.translate("HeaderAddLocalUser"),
-                id: "manual",
-                type: "submit",
-                description: globalize.translate("AddUserByManually")
-            });
-            // TODO cvium
-            // items.push({
-            //     name: globalize.translate("HeaderInviteUser"),
-            //     id: "invite",
-            //     description: globalize.translate("HeaderInviteUserHelp")
-            // });
-            items.push({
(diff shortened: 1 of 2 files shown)


src/scripts/userprofilespage.js
```diff
@@ -242,48 +242,11 @@ define(["loading", "dom", "globalize", "humanedate", "paper-icon-button-light",
...
-            var items = [];
-            items.push({
-                name: globalize.translate("HeaderAddLocalUser"),
-                id: "manual",
-                type: "submit",
-                description: globalize.translate("AddUserByManually")
-            });
-            // TODO cvium
-            // items.push({
-            //     name: globalize.translate("HeaderInviteUser"),
```
joshuaboniface: @cvium Is this planned to be implemented? If so it could be a valid reason to keep this dialog, otherwise rip it out.

anthonylavado: Not @cvium, but this comes from https://github.com/jellyfin/jellyfin/pull/399 which was part of https://github.com/jellyfin/jellyfin/pull/398.

I’d assume it’s safe to take out for now if we want, as we’ll want to architect the e-mail sharing system together at some point.

cvium: My plan was to re-use the Connect UI components for the email invite thing

dkanada: What components would be reused specifically? I was thinking it would just be a token that allows a user to create their own account. Not sure how this dialog would be leveraged for an invite system.

nvllsvm: @cvium - do you plan on having a pull request ready to merge in the next 7 days? No sense keeping this around otherwise.

cvium: @dkanada It would be used for sending an invite email. Either add a user manually or generate an email.

@nvllsvm I'm okay with deleting it. I'm pretty sure the HTML was already removed in another pr anyway

PR #92, opened 2019-01-29
Added frontend support for a password field on setup

RazeLighter777:
This pull request is the frontend counterpart to the pull request here: https://github.com/jellyfin/jellyfin/pull/758

This will not work without that pull request, so wait until it is merged.

This adds a password prompt on to the setup screen so users can set a password when they setup Jellyfin.


src/strings/en-GB.json
```diff
@@ -82,6 +82,7 @@
   "LabelUrl": "URL:",
   "LabelZipCode": "Post Code:",
   "LabelffmpegPathHelp": "The path to the ffmpeg application file or folder containing ffmpeg.",
+  "LeaveBlankToNotSetAPassword": "This field can be left blank.",
```
joshuaboniface: I'd actually rather this be something like: `Optional - leave blank to set no password.`, both descriptive and functional.

src/strings/en-US.json
```diff
@@ -1028,6 +1028,7 @@
   "LanNetworksHelp": "Comma separated list of IP addresses or IP/netmask entries for networks that will be considered on local network when enforcing bandwidth restrictions. If set, all other IP addresses will be considered to be on the external network and will be subject to the external bandwidth restrictions. If left blank, only the server's subnet is considered to be on the local network.",
   "LatestFromLibrary": "Latest {0}",
   "LearnHowToCreateSynologyShares": "Learn how to share folders in Synology.",
+  "LeaveBlankToNotSetAPassword": "This field can be left blank.",
```
joshuaboniface: ibid.

RazeLighter777: Alright I fixed it.

src/scripts/wizarduserpage.js
```diff
@@ -36,7 +38,12 @@ define(["loading", "globalize", "dashboardcss", "emby-input", "emby-button", "em
     }
 
     function onSubmit(e) {
-        return submit(this), e.preventDefault(), !1
+        
+        var form = this;
+        return form.querySelector("#txtManualPassword").value != form.querySelector("#txtPasswordConfirm").value ? require(["toast"], function(toast) {
+            toast(Globalize.translate("PasswordMatchError"))
+        }) : submit(this), e.preventDefault(), !1
```
cvium: This is not that easy to read. I'm guessing you were trying to follow the coding style, but this file has been the victim of `uglify` and it's a long process of undoing it to all those poor JS files. We believe the original owner kept the real source in a private repo.

Anyway, I digress. Suggestion:
```
        var form = this;
        if (form.querySelector("#txtManualPassword").value != form.querySelector("#txtPasswordConfirm").value) {
            require(["toast"], function(toast) {
                toast(Globalize.translate("PasswordMatchError"));
            });
        } else {
            submit(form);
        }
        e.preventDefault();
        return false;
```

src/scripts/wizarduserpage.js
```diff
@@ -26,7 +26,9 @@ define(["loading", "globalize", "dashboardcss", "emby-input", "emby-button", "em
         apiClient.ajax({
             type: "POST",
             data: {
-                Name: form.querySelector("#txtUsername").value
+                Name: form.querySelector("#txtUsername").value,
+                Password:  form.querySelector("#txtManualPassword").value 
+
```
cvium: Extraneous whitespace.

src/scripts/wizarduserpage.js
```diff
@@ -26,7 +26,9 @@ define(["loading", "globalize", "dashboardcss", "emby-input", "emby-button", "em
         apiClient.ajax({
             type: "POST",
             data: {
-                Name: form.querySelector("#txtUsername").value
+                Name: form.querySelector("#txtUsername").value,
+                Password:  form.querySelector("#txtManualPassword").value
```
cvium: Extraneous whitespace after colon

src/scripts/wizarduserpage.js
```diff
@@ -36,7 +38,12 @@ define(["loading", "globalize", "dashboardcss", "emby-input", "emby-button", "em
     }
 
     function onSubmit(e) {
-        return submit(this), e.preventDefault(), !1
+
```
cvium: Unnecessary whitespace

src/scripts/wizarduserpage.js
```diff
@@ -45,6 +52,7 @@ define(["loading", "globalize", "dashboardcss", "emby-input", "emby-button", "em
             apiClient = getApiClient();
         apiClient.getJSON(apiClient.getUrl("Startup/User")).then(function(user) {
             page.querySelector("#txtUsername").value = user.Name || "", loading.hide()
+            page.querySelector("#txtManualPassword").value = user.Password || "", loading.hide()
```
cvium: This will call `loading.hide() twice` and it should be deuglified.
```suggestion
            page.querySelector("#txtUsername").value = user.Name || "";
            page.querySelector("#txtManualPassword").value = user.Password || "";
            loading.hide();
```

src/scripts/wizarduserpage.js
```diff
@@ -45,6 +52,7 @@ define(["loading", "globalize", "dashboardcss", "emby-input", "emby-button", "em
             apiClient = getApiClient();
         apiClient.getJSON(apiClient.getUrl("Startup/User")).then(function(user) {
             page.querySelector("#txtUsername").value = user.Name || "", loading.hide()
```
cvium: See below
```suggestion
```

src/strings/en-GB.json
```diff
@@ -82,6 +82,7 @@
   "LabelUrl": "URL:",
   "LabelZipCode": "Post Code:",
   "LabelffmpegPathHelp": "The path to the ffmpeg application file or folder containing ffmpeg.",
+  "LeaveBlankToNotSetAPassword": "Optional - leave blank to set no password",
```
dkanada: LGTM! One small change, the British translation should only be used for strings that require localization from English such as colour or programme, so a translation isn't required here.

There are probably lots of useless strings in this file still because my script only searched for direct matches to the English translation. Any translations in this file that don't use British terms need to be removed, I was just waiting for Weblate to get set up before I went and removed them all.

src/scripts/wizarduserpage.js
RazeLighter777: I addressed your changes.

src/strings/en-GB.json
RazeLighter777: Fixed.

src/scripts/wizarduserpage.js
RazeLighter777: Fixed

src/scripts/wizarduserpage.js
RazeLighter777: Fixed

src/scripts/wizarduserpage.js
RazeLighter777: Fixed

src/scripts/wizarduserpage.js
RazeLighter777: Fixed

src/scripts/wizarduserpage.js
RazeLighter777: Fixed

RazeLighter777:
Alright, all feedback is addressed.

PR #93, opened 2019-01-29
Update humanedate.js to pull client date and time

Maltahl:
Not entirely sure as this is my first PR.
Changes time from server time to client time as requested in issue jellyfin#50

### DIFF
--- a/src/components/humanedate.js
+++ b/src/components/humanedate.js
@@ -1,28 +1,8 @@
-define(["datetime"], function(datetime) {
-    "use strict";
-
-    function humane_date(date_str) {
-        var format, time_formats = [
-                [90, "a minute"],
-                [3600, "minutes", 60],
-                [5400, "an hour"],
-                [86400, "hours", 3600],
-                [129600, "a day"],
-                [604800, "days", 86400],
-                [907200, "a week"],
-                [2628e3, "weeks", 604800],
-                [3942e3, "a month"],
-                [31536e3, "months", 2628e3],
-                [47304e3, "a year"],
-                [31536e5, "years", 31536e3]
-            ],
-            dt = new Date,
-            date = datetime.parseISO8601Date(date_str, !0),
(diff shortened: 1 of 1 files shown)


src/components/humanedate.js
```diff
@@ -1,28 +1,8 @@
...
-});
\ No newline at end of file
+var currentdate = new Date(); 
+var datetime = "Last Sync: " + currentdate.getDate() + "/"
+                + (currentdate.getMonth()+1)  + "/" 
+                + currentdate.getFullYear() + " @ "  
+                + currentdate.getHours() + ":"  
+                + currentdate.getMinutes() + ":" 
+                + currentdate.getSeconds();
+    return window.Date = Date, Date
```
JustAMan: This cannot be right, that `humane_date` displayed something else, plus some other code used `window.humane_date` which you broke.
Oh, and you're not exposing `datetime` anywhere.

src/components/humanedate.js
```diff
@@ -1,28 +1,8 @@
...
-            i = 0;
-        for (seconds < 0 && (seconds = Math.abs(seconds)); format = time_formats[i++];)
-            if (seconds < format[0]) return 2 == format.length ? format[1] + " ago" : Math.round(seconds / format[2]) + " " + format[1] + " ago";
-        return seconds > 47304e5 ? Math.round(seconds / 47304e5) + " centuries ago" : date_str
-    }
-    return window.humane_date = humane_date, humane_date
-});
\ No newline at end of file
+var currentdate = new Date(); 
+var datetime = "Last Sync: " + currentdate.getDate() + "/"
```
JustAMan: This variable needs a more descriptive name, maybe `lastSyncDate` or something like this.

src/components/humanedate.js
Maltahl: Makes sense. Thanks for the insight.

PR #95, opened 2019-01-31
add display language option back

dkanada:
It seems I accidentally broke the display language setting in #51 so this PR reverts that change. There was originally a check for `isNativeApp` which you would expect to only be true for cordova instances but apparently there is some currently unknown behavior there.

### DIFF
--- a/src/components/apphost.js
+++ b/src/components/apphost.js
@@ -140,6 +140,8 @@ define(["appSettings", "browser", "events", "htmlMediaHelper"], function(appSett
         (browser.tv || browser.xboxOne || browser.ps4 || browser.mobile) && features.push("physicalvolumecontrol");
         browser.tv || browser.xboxOne || browser.ps4 || features.push("remotecontrol");
         browser.operaTv || browser.tizen || browser.orsay || browser.web0s || browser.edgeUwp || features.push("remotevideo");
+        // might require isNativeApp check if any other issues surface
+        features.push("displaylanguage");
         features.push("otherapppromotions");
         features.push("targetblank");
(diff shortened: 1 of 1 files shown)


dkanada:
I would rather not, it shouldn't even return true for the web client in the first place, unless I am misunderstanding the variable meaning. I can add it in but this PR would just always enable display language for the web client. I can't think of a situation where display language would cause issues.

src/components/apphost.js
```diff
@@ -140,6 +140,8 @@ define(["appSettings", "browser", "events", "htmlMediaHelper"], function(appSett
         (browser.tv || browser.xboxOne || browser.ps4 || browser.mobile) && features.push("physicalvolumecontrol");
         browser.tv || browser.xboxOne || browser.ps4 || features.push("remotecontrol");
         browser.operaTv || browser.tizen || browser.orsay || browser.web0s || browser.edgeUwp || features.push("remotevideo");
+        // might require isNativeApp check if any other issues surface
```
JustAMan: Maybe we should add a link to the comments?.. to have more context in place
```suggestion
        // might require isNativeApp check if any other issues surface
        // see https://github.com/jellyfin/jellyfin-web/pull/51 for when it was probably broken
```

dkanada: I actually traced isNativeApp throughout the project, this shouldn't be an issue at all anymore. I would say we could even remove the first comment but I can leave that one just in case.

Basically, that variable is set in a single location in the entire codebase when `appMode` is set to either android or cordova, which will never return true in the bundled web interface from now on.

PR #96, opened 2019-01-31
Add multiserver support for web client

dkanada:
This enables multiserver support for the bundled web client for two reasons.

1. Why hide an option for no reason?
2. Some users may want this enabled.

This is exactly the same in Plex where you can select another server without leaving the current web frontend. I personally think this is useful, and it also works really well from the bundled frontend from my testing. I was able to switch between my dev and prod servers and the API calls were routed to the active server while the frontend calls were left on the current connection. I tested media playback, modifying user settings, and using the dashboard and everything worked fine for me.

### DIFF
--- a/src/scripts/addpluginpage.js
+++ b/src/scripts/addpluginpage.js
@@ -28,28 +28,27 @@ define(["jQuery", "loading", "libraryMenu", "globalize", "connectionManager", "e
     }
 
     function renderPluginInfo(page, pkg, pluginSecurityInfo) {
-        if (!AppInfo.isNativeApp)
-            if (pkg.isPremium) {
-                $(".premiumPackage", page).show();
-                var regStatus = "";
-                if (pkg.isRegistered) regStatus += "<p style='color:green;'>", regStatus += globalize.translate("MessageFeatureIncludedWithSupporter");
-                else {
-                    var expDateTime = new Date(pkg.expDate).getTime(),
-                        nowTime = (new Date).getTime();
-                    expDateTime <= nowTime ? (regStatus += "<p style='color:red;'>", regStatus += globalize.translate("MessageTrialExpired")) : expDateTime > new Date(1970, 1, 1).getTime() && (regStatus += "<p style='color:blue;'>", regStatus += globalize.translate("MessageTrialWillExpireIn").replace("{0}", Math.round(expDateTime - nowTime) / 864e5))
-                }
-                if (regStatus += "</p>", $("#regStatus", page).html(regStatus), pluginSecurityInfo.IsMBSupporter)
-                    if ($(".premiumDescription", page).hide(), $(".supporterDescription", page).hide(), pkg.price > 0) {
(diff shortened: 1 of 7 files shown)


hawken93:
I'd like this to be something that could be decided by configuration at a later point :) for now I understand it's far easier to just enable it. LGTM

dkanada:
I want to make a few small changes before this gets merged to support running the web client on its own, should be done in a few days.

dkanada:
These changes will enable buttons to select a server in the user settings or navigation drawer for all web clients. All asset requests go to the hosted location regardless of the currently active server, and everything seems to be working fine. There shouldn't be any issues since this already works flawlessly on the Android client.



I have also enabled soft support for hosting the web client anywhere. I have been using nginx myself and it works great. The only change to get that working is adding the following two lines at the bottom of the index file. There is a single issue with a phantom server entry in the server selection page, but it has no effect on anything as far as I can tell.

'''
<script src='scripts/apploader.js'></script>
<script>window.appMode='standalone';</script>
'''

src/scripts/librarymenu.js
```diff
@@ -198,11 +198,7 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
         html += "</h3>";
         if (user.localUser) {
             html += '<a is="emby-linkbutton" class="navMenuOption lnkMediaFolder lnkMySettings" href="mypreferencesmenu.html"><i class="md-icon navMenuOptionIcon">settings</i><span class="navMenuOptionText">' + globalize.translate("ButtonSettings") + "</span></a>";
-        }
-        if (AppInfo.isNativeApp) {
             html += '<a is="emby-linkbutton" class="navMenuOption lnkMediaFolder" data-itemid="selectserver" href="selectserver.html?showuser=1"><i class="md-icon navMenuOptionIcon">wifi</i><span class="navMenuOptionText">' + globalize.translate("ButtonSelectServer") + "</span></a>";
-        }
-        if (user.localUser && !user.localUser.EnableAutoLogin) {
```
JustAMan: why remove this `if`?..

src/scripts/addpluginpage.js
```diff
@@ -28,28 +28,27 @@ define(["jQuery", "loading", "libraryMenu", "globalize", "connectionManager", "e
...
-                        var url = "https://mb3admin.local/admin/service/user/getPayPalEmail?id=" + pkg.owner;
-                        fetch(url).then(function(response) {
-                            return response.json()
-                        }).then(function(dev) {
-                            dev.payPalEmail && ($("#payPalEmail", page).val(dev.payPalEmail), $("#ppButton", page).show())
-                        })
-                    } else $(".premiumHasPrice", page).hide();
-                else pkg.price ? ($(".premiumDescription", page).show(), $(".supporterDescription", page).hide()) : ($(".premiumDescription", page).hide(), $(".supporterDescription", page).show()), $("#ppButton", page).hide()
-            } else $(".premiumPackage", page).hide()
+        if (pkg.isPremium) {
```
cvium: Eek... I think this function should just be removed. I know it's not really related to your changes, but it just seems odd to remove that top `if` instead of the whole function.

PR #112, opened 2019-02-02
Removed seasonal theme support

### LINKED ISSUE #71: Remove seasonal themes
These rely upon external assets of dubious licensing. https://github.com/jellyfin/jellyfin/blob/master/MediaBrowser.WebDashboard/dashboard-ui/bower_components/emby-webcomponents/skinmanager.js#L121

sparky8251:
As our sole seasonal theme was for Halloween and relied on dubiously licensed art assets to work, I removed the entire system.

Fixes #71


sparky8251:
Finished digging in the backend. Doesn't appear to be any logic there for managing seasonal themes.

All it does is hand the userdisplayprefs BLOB from the database without caring whats inside. All the magic was client side as far as I can see.

PR #115, opened 2019-02-02
WIP: Fix volume slider behavior

sparky8251:
This is based off my #112 PR, so its liable to appear like its changing more than it really is until that's merged or I fix it.

This is for staging my deuglification of `src/scripts/videoosd.js` so it can be properly reviewed.

I'll update this top level post when the actual fixes are in place and I've addressed the branch base problems.


sparky8251:
Closing to rebase off master instead of a personal branch.

PR #119, opened 2019-02-03
Make the toggle track visible on all themes

### LINKED ISSUE #67: Change Color of Switch Track
This may tie in to the theme/colour work that @sparky8251 is doing.

Essentially, right now with the new dark theme, the track for toggle switches is set to black (or nearly black). As a result, a switch just looks like a floating button.

I couldn't tell you where `emby-toggle.css` is being generated, but wherever it is, it ends up with this:
```css
.mdl-switch__track {
    background: rgba(0, 0, 0, .2);
    height: 1em;
    -webkit-border-radius: 1em;
    border-radius: 1em;
    cursor: pointer
}
```

That background should be changed to something lighter.

anthonylavado:
The default toggle track colour was essentially a transparent black. While this works on brighter themes, it didn't work on the now default dark theme. It is now set to a "middle grey" that is visible on all themes (light or dark).

Fixes #67.

### DIFF
--- a/src/bower_components/emby-webcomponents/emby-toggle/emby-toggle.css
+++ b/src/bower_components/emby-webcomponents/emby-toggle/emby-toggle.css
@@ -41,7 +41,7 @@
 }
 
 .mdl-switch__track {
-    background: rgba(0,0,0, 0.2);
+    background: rgba(255,255,255, 0.2);
     height: 1em;
     border-radius: 1em;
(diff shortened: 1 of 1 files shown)


anthonylavado:
It's a little out of place to see the lighter grey against some themes with lots of colour, like "Blue Radiance", but it's the best compromise I can think of, unless this can be customized by the themes.

Issue #120, opened 2019-02-05
Customize Home Screen never goes away


devenator:
The "Customize Home Screen"-Bar at bottom never goes away, even after clicking it and changing options for the user's homescreen.

**To Reproduce**

1. Open Jellyfin web-gui.
2. "Customize Home Screen" appears a few seconds after the whole page (content) has been loaded at the bottom of the homescreen.

**Expected behavior**
permanently remove it after clicking it once and/or changing any options in User Options.

(My Two Cents: Remove it completely. This was basically part of E-Who-Shall-Not-Be-Named's push for "Explore and Discover (Premier, buybuybuy)", and i find it insulting that every app feels a need to tell me that i can ... customize my options! </My To Cents>)

**Screenshots**

![chs](https://user-images.githubusercontent.com/45917913/52287985-86201600-296b-11e9-9064-fb2e0a21d7a1.jpg)

**System (please complete the following information):**
 - OS: Ubuntu 18.04 X64 (build with docker)
 - Browser: Firefox, Chrome
 - Jellyfin Version: 10.0.1

**Additional context**
This is on a clean, fresh installation, not a migrated one, user also new and not migrated. Also, E*** had the same problem multiple times in ~2017...

JustAMan:
Hmm, I would rather we have it there with a button that says something like "I know I can customize in profile settings", so that clicking that button removes that "customize" bar for this user completely.

So it's still there for new users to see, and is easily dismissable by veterans.

devenator:
Well i've just found out that this was actually added *after* my last E*** Version (old 3.2.x .net), i was pretty certain that i was seeing it there and it went away after clicking around, but acording to [the old forum](https://emby.media/community/index.php?/topic/61194-remove-customize-home-screen-button/), this is basically just trolling by luke, because TWO obvious ways (three-dot-menu AND the profile-icon) to get into the user-profile are not enough.

But the workaround still works:

```
Dashboard -> Settings -> custom css
 
#homeTab .customizeSection {display: none}

```
Again, my2cents: From gmail to youtube, the profile icon in the upper right is used everywhere, if someone really does not get this, it's probably better if he/she/it does not change settings in there anyways. But for me, the custom-css is good enough :)

anthonylavado:
This is fixed in the upcoming release. Thanks!

PR #131, opened 2019-02-11
remove "Customize Home Screen" Bar at Bottom of web ui

grafixeyehero:
#120

### DIFF
--- a/src/bower_components/emby-webcomponents/homesections/homesections.js
+++ b/src/bower_components/emby-webcomponents/homesections/homesections.js
@@ -68,16 +68,6 @@ define(['connectionManager', 'cardBuilder', 'registrationServices', 'appSettings
 
             return Promise.all(promises).then(function () {
 
-                html = '';
-
-                var style = 'margin-top:4em;';
-
-                if (layoutManager.tv) {
-                    style += 'padding: 0 7.5%;';
-                }
-
-                elem.insertAdjacentHTML('beforeend', html);
-
                 return resume(elem, {
                     refresh: true,
                     returnPromise: false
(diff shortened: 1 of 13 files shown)


dkanada:
I agree with @cvium, this button is too large and very useless in my opinion. I also don't think hiding the item when a user clicks it for the first time is a good solution either, which was mentioned in chat. Onboarding features like that are never very useful and create an ever-changing interface as they are hidden and shown to users.

sparky8251:
Yup. I also 100% back removing it entirely.

No need for 3 links to the same place and it's always been an odd UI element since its only there on the "true" home screen and not similar looking screens.

sparky8251:
That said, this removal appears incomplete.

The strings are still present in the strings files at the very least. There is also the possibility of some logic being stashed in a JS file.

Should first make sure removal is complete before accepting this PR.

cvium:
The change is _in_ in the js file. The removal is complete afaict. Strings removal should be done from weblate?

JustAMan:
> Strings removal should be done from weblate?

This is a tricky question, but I doubt weblate can remove items. So I would say remove them in this PR...

grafixeyehero:
@cvium suggested it's done

sparky8251:
The strings files are in `src/strings`.

You'll want to remove the lines containing `HeaderCustomizeHomeScreen` from any file in that folder to remove the string cruft.

grafixeyehero:
@sparky8251 it's done

sparky8251:
Just cause I've seen a lot of stupid with strings I checked for potential string variants that might have been from the olden days and never removed by Emby.

See no such thing! To me, it looks like this is ready for merging with no left over dead code.

grafixeyehero:
i agree

PR #135, opened 2019-02-13
Make sure fallback culture is always available

### LINKED ISSUE #134: Non-US languages fail to properly fallback to en_US translation when they should
In my testing, EN_GB fails to fallback to EN_US correctly, instead it shows the labels only.

This is less prominent in other languages as they have more translated strings, but the problem is still there all the same

JustAMan:
Fixes #134

### DIFF
--- a/src/bower_components/emby-webcomponents/globalize.js
+++ b/src/bower_components/emby-webcomponents/globalize.js
@@ -1,5 +1,6 @@
 define(['connectionManager', 'userSettings', 'events'], function (connectionManager, userSettings, events) {
     'use strict';
+    var fallbackCulture = 'en-us';
 
     var allTranslations = {};
     var currentCulture;
@@ -29,7 +30,7 @@ define(['connectionManager', 'userSettings', 'events'], function (connectionMana
             return navigator.languages[0];
         }
 
-        return 'en-us';
+        return fallbackCulture;
     }
 
     function updateCurrentCulture() {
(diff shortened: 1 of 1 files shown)


src/bower_components/emby-webcomponents/globalize.js
```diff
@@ -62,6 +63,11 @@ define(['connectionManager', 'userSettings', 'events'], function (connectionMana
         for (var i in allTranslations) {
             ensureTranslation(allTranslations[i], culture);
         }
+        if (culture !== fallbackCulture) {
```
cvium: Why not have the if inside the loop instead?

JustAMan: Because speed! :) This way it only checks once instead of how many entries there is in `allTranslations`.

thornbill: I doubt looping over all the entries twice is faster. Maybe it's faster if your locale is already `en-us`, but it seems like it would be worse for any other case. :wink:

LeoVerto:
I know this naming was part of the code previously but "culture" feels like the wrong term to me. Shouldn't this be something like "localization" instead?

src/bower_components/emby-webcomponents/globalize.js
JustAMan: Isn't it a moot point? :) This code executes once the application is loaded, and once a localization language is changed. Even if it adds like 50 extra milliseconds is it that bad? :)

For real, it's impossible to tell which one is faster without benchmarks, and I don't think we should really be concerned with performance here.

thornbill: I won't hold this up, but the double `for` loop still makes me cringe. :smile:

JustAMan:
@LeoVerto maybe `language` or `dialect`?

I was sticking to the style of the file to keep changes minimal as this targeted release branch.

PR #137, opened 2019-02-15
Replace HeaderEmbyServer with HeaderJellyfinServer on plugincatalog

grafixeyehero:


src/plugincatalog.html
```diff
@@ -7,7 +7,7 @@
 
                     <div class="selectContainer">
                         <select is="emby-select" id="selectSystem" label="${LabelDisplayPluginsFor}">
-                            <option value="Server">${HeaderEmbyServer}</option>
+                            <option value="Server">${HeaderJellyfinServer}</option>
```
JustAMan: Did you check that this localizes properly? I don't see corresponding localization files' change.

cvium: 'Emby' was replaced with 'Jellyfin' in all strings very early on -- even keys

dkanada: Just checked, it's included in the strings.

grafixeyehero: @JustAMan i checked it it's already there "HeaderJellyfinServer": "Jellyfin Server"

PR #141, opened 2019-02-17
Exit fullscreen automatically after video playback ends

cvium:
Ideally, the video element or its parent container should be fullscreened and none of this would be necessary, _BUT_ video osd is in a separate container, so video element cannot be fullscreened without the osd disappearing...

So this change is the next best solution I could think of. Maybe I'll look into refactoring videoosd to wrap the video element. 

Thoughts?

Fixes https://github.com/jellyfin/jellyfin/issues/923

### DIFF
--- a/src/bower_components/emby-webcomponents/fullscreen/fullscreenmanager.js
+++ b/src/bower_components/emby-webcomponents/fullscreen/fullscreenmanager.js
@@ -29,11 +29,14 @@ define(['events', 'dom'], function (events, dom) {
         }
         if (element.webkitEnterFullscreen) {
             element.webkitEnterFullscreen();
-        } 
+        }
     };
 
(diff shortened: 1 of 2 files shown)


joshuaboniface:
I think this works as a solution, given what we have already. It's not the best fix for #923 but it is a fix. Slated for inclusion in 10.2.1 as a bugfix.

CutterXYZ:
Is this necessary ?

Is there any way to re-enable fullscreen remotely from the android app without leaving the couch ? :)

Nickbert7:
Maybe we could keep it in fullscreen if the display mode is set to "TV" and leave it on the other display modes?

JustAMan:
Maybe we should automatically go full-screen when in TV mode?

PR #144, opened 2019-02-18
Add top offset and bottom offset margins in emby-scrollbuttons

grafixeyehero:
**Issues**
Arrow buttons cover the options menu in the dash #142

**Before**

before

**After**

after

**optional**  **removing gradient**

optional


grafixeyehero:
but which one did u prefer the 2nd or the 3rd

JustAMan:
I dunno, probably 2nd to better see where is that scroll thing... Though maybe remove gradient from it? Most our ui is flat.

PR #146, opened 2019-02-20
Load default lang for missing keys

vitorsemeano:
**Changes**
The globalize.js have a fallback implementation to load missing keys using the en-us dictionary. What happens is that this dictionary is not loaded until the user selects it in Display settings. To ensure this dictionary is always available, this PR forces that load when loading the user defined dictionary.


cvium:
This was already fixed in #135. Did it not get merged into master yet?
cc @joshuaboniface

vitorsemeano:
It's fixed in updstream, my bad. Closing this now.

Issue #25, opened 2019-01-12
close now playing page on media stop


dkanada:
When the stop button is clicked from `nowplaying.html` the queue resets and the user is left on a blank page. It would be better if the user was sent back to where they were before they opened the queue.

joshuaboniface:
Included in 10.2.2.

PR #152, opened 2019-02-23
Go back on Media stop

redSpoutnik:
Make ```nowplaying.html``` close on media stop.

**Changes**

When the stop button is clicked, the playing queue resets then users are sent to the page they were before ```nowplaying.html```, instead of an empty page.

**Issues**

#25


joshuaboniface:
@redSpoutnik Can you retarget/rebase this off of the `release-10.2.z` branch instead of `master`? It'll help us get this hotfix into 10.2.2.

joshuaboniface:
Thanks @redSpoutnik !

PR #154, opened 2019-02-23
Remove internet dependency for device icons

sparky8251:
**Changes**
Added icons to a local path and changed the requests to the now local icons.

**Oddities**
This PR will only fix part of the problem. The real solution is much larger. For some info on how this works and what else still needs doing read on.

When a client first authenticates it sends a huge chunk of data to `/emby/Sessions/Capabilities/Full` and the server writes this file to disk in a folder named by the device ID. In this file there is a key, `IconUrl` that has the URL the client gets its device icon from.

The fun comes from the fact that devices don't appear to update their capabilities very often, if at all and if they do, they certainly don't update the `IconUrl`. This means that when you go to the devices page and the web UI goes to get icons it uses the path the server tells it to, not the path it would tell the server to record.

Old devices then call out to github.com while new ones call to the local server. This problem will also cause breaks if ip/domain:port changes, the client will attempt to request the icon from a non-existent server.

Thankfully, the scope of the problem is minimal and doesn't actually have any negative impacts. The device page will not crash or error if it cannot find an icon, it just displays a blank box. 

Eventually we will have to go back and fix this for real. For now, this works to remove yet another dependency on internet access!


src/components/apphost.js
```diff
@@ -220,7 +220,7 @@ define(["appSettings", "browser", "events", "htmlMediaHelper"], function(appSett
...
+            return browser.edgeUwp, browser.edgeUwp ? "/web/css/images/devices/windowsrt.png" : browser.opera || browser.operaTv ? "/web/css/images/devices/opera.png" : browser.orsay || browser.tizen ? "/web/css/images/devices/samsungtv.png" : browser.web0s ? "/web/css/images/devices/lgtv.png" : browser.ps4 ? "/web/css/images/devices/ps4.png" : browser.chromecast ? "/web/css/images/devices/chromecast.png" : browser.chrome ? "/web/css/images/devices/chrome.png" : browser.edge ? "/web/css/images/devices/edge.png" : browser.firefox ? "/web/css/images/devices/firefox.png" : browser.msie ? "/web/css/images/devices/internetexplorer.png" : browser.safari ? "/web/css/images/devices/safari.png" : "/web/css/images/devices/html5.png"
```
cvium: Why `css/images`?

sparky8251: These images are only attached to the CSS element "cardBuilder".

I have no qualms with a better position for the images, just put them somewhere that kinda sorta made sense.

If you got one, let me know!

LeoVerto: Would this be more readable as a switch/case construct? More than one ternary operator in a return just is too much IMHO.

sparky8251: This is still uglified code. This entire file has not been deuglified and I didn't do it as part of this PR as its such a simple change compared to actually changing logic.

cvium: css seems like the wrong place tbh

sparky8251: Well, I could move it up a folder level and have it in a folder named `images` but then, instead of organizing, we have just added yet another folder with images in it.

JustAMan: I would say that putting them under `images` next to `css` suits me more than shoving them into `css`. Or maybe rename `css` to `assets` (but that's a less-preferable way for me).

JustAMan: Also certainly, even if all other file is uglified, make this a switch.

Plus with those absolute URLs you're effectively killing the possibility to host this under URL prefix :( There's a `baseUrl()` function that _might_ serve as a starter for URLs here.

sparky8251:
Well... I'll look into using the baseUrl() stuff, but I'm not sure the above is fully absolute paths. For me, it already prepends `ipaddress:port` to the beginning.

joshuaboniface:
Yea I agree with @JustAMan there, better to make sure it's set now rather than fight with breakage when we do eventually implement a `baseUrl` setup.

joshuaboniface:
Just a ping in this one - keep or close?

sparky8251:
Well, we certainly want to keep. This is removing a dependency on Emby keeping resources at a specific URL for parts of the UI to function normally.

Probably want to skip this for 10.3.0 though since I haven't found time since the new job to look at this again.

thornbill:
Perhaps something like https://simpleicons.org/ or https://ionicons.com/ would be a better source for these icons? Something that is definitely released under a compatible license.

JustAMan:
@sparky8251 I feel that since we just recently added `baseUrl` support, we can probably re-make this PR. Thoughts?

thornbill:
This was addressed by https://github.com/jellyfin/jellyfin-web/pull/379

PR #162, opened 2019-02-28
Convertion to webpack - part1 - npm and dynamic module resolution for site.js

vitorsemeano:
This is a first for a serie of PR i will be making for the convertion to webpack of the current master version.

This PR simple adds a package.json with basic configs, and also addresses some issues with dynamic modules resolution on site.js

Also the deuglify of site.js is present in these changes.

The expected behavior should be the same as the master version.


src/scripts/site.js
```diff
@@ -1,216 +1,333 @@
...
-                var promises = [apphost.getSyncProfile(), apphost.init()];
-                Promise.all(promises).then(function(responses) {
-                    var deviceProfile = responses[0],
-                        capabilities = Dashboard.capabilities(apphost);
-                    capabilities.DeviceProfile = deviceProfile;
-                    var connectionManager = new ConnectionManager(credentialProviderInstance, apphost.appName(), apphost.appVersion(), apphost.deviceName(), apphost.deviceId(), capabilities, window.devicePixelRatio);
-                    if (defineConnectionManager(connectionManager), bindConnectionManagerEvents(connectionManager, events, userSettings), !AppInfo.isNativeApp) return console.log("loading ApiClient singleton"), getRequirePromise(["apiclient"]).then(function(apiClientFactory) {
+        return require(["connectionManagerFactory", "apphost", "credentialprovider", "events", "userSettings"], function (ConnectionManager, apphost, credentialProvider, events, userSettings) {
+            var credentialProviderInstance = new credentialProvider(),
+                promises                   = [apphost.getSyncProfile(), apphost.init()];
```
cvium: I'm not a fan of this. I'd really prefer having variable declarations on separate lines.

src/scripts/site.js
```diff
@@ -1,216 +1,333 @@
...
+
+            Promise.all(promises).then(function (responses) {
+                var deviceProfile = responses[0],
+                    capabilities  = Dashboard.capabilities(apphost);
```
cvium: Same here

src/scripts/site.js
```diff
@@ -1,216 +1,333 @@
...
+                    console.log("loading ApiClient singleton");
+
+                    return require(["apiclient"], function (apiClientFactory) {
                         console.log("creating ApiClient singleton");
+
                         var apiClient = new apiClientFactory(Dashboard.serverAddress(), apphost.appName(), apphost.appVersion(), apphost.deviceName(), apphost.deviceId(), window.devicePixelRatio);
-                        apiClient.enableAutomaticNetworking = !1, apiClient.manualAddressOnly = !0, connectionManager.addApiClient(apiClient), window.ApiClient = apiClient, localApiClient = apiClient, console.log("loaded ApiClient singleton"), resolve()
+                        
+                        apiClient.enableAutomaticNetworking = false;
+                        apiClient.manualAddressOnly         = true;
```
cvium: I kind of get what you're going for here, but I'd prefer not having all this extra whitespace. If you add a variable with a longer name, you'll need to indent everything etc.
```suggestion
                        apiClient.manualAddressOnly = true;
```

src/scripts/site.js
```diff
@@ -287,103 +413,191 @@ var Dashboard = {
...
-            ? define("serverdiscovery", ["cordova/serverdiscovery"], returnFirstDependency)
-            : define("serverdiscovery", [apiClientBowerPath + "/serverdiscovery"], returnFirstDependency);
-        "cordova" === self.appMode && browser.iOSVersion && browser.iOSVersion < 11
-            ? define("imageFetcher", ["cordova/imagestore"], returnFirstDependency)
-            : define("imageFetcher", [embyWebComponentsBowerPath + "/images/basicimagefetcher"], returnFirstDependency);
+
+        if ("registerElement" in document) {
+            define("registerElement", []);
+        } else {
+            if (browser.msie) {
```
cvium: This should be merged with the else, so it becomes if-else if-else

src/scripts/site.js
```diff
@@ -287,103 +413,191 @@ var Dashboard = {
...
+        } else {
+            define("castSenderApiLoader", [], getCastSenderApiLoader);
+        }
+
+        if (self.Windows) {
+            define("bgtaskregister", ["environments/windows-uwp/bgtaskregister"], returnFirstDependency);
+            define("transfermanager", ["environments/windows-uwp/transfermanager"], returnFirstDependency);
+            define("filerepository", ["environments/windows-uwp/filerepository"], returnFirstDependency);
+        } else {
+            if ("cordova" === self.appMode) {
```
cvium: `deuglify`  doesn't handle nested if's very well, so this nesting should be fixed manually

src/scripts/site.js
```diff
@@ -287,103 +413,191 @@ var Dashboard = {
...
+                    });
+                });
+            });
+        });
     }
 
     function loadCoreDictionary(globalize) {
         var languages = ["ar", "be-by", "bg-bg", "ca", "cs", "da", "de", "el", "en-gb", "en-us", "es", "es-ar", "es-mx", "fa", "fi", "fr", "fr-ca", "gsw", "he", "hi-in", "hr", "hu", "id", "it", "kk", "ko", "lt-lt", "ms", "nb", "nl", "pl", "pt-br", "pt-pt", "ro", "ru", "sk", "sl-si", "sv", "tr", "uk", "vi", "zh-cn", "zh-hk", "zh-tw"];
-        var translations = languages.map(function(i) {
+        var translations = languages.map(function (i__t) {
```
cvium: Please replace variables like this with more descriptive names. This one should be `language`.

src/scripts/site.js
```diff
@@ -420,586 +644,747 @@ var Dashboard = {
...
-        var list = [
-            "bower_components/emby-webcomponents/playback/playbackvalidation",
-            "bower_components/emby-webcomponents/playback/playaccessvalidation",
-            "bower_components/emby-webcomponents/playback/experimentalwarnings",
-            "bower_components/emby-webcomponents/htmlaudioplayer/plugin",
-            "bower_components/emby-webcomponents/htmlvideoplayer/plugin",
-            "bower_components/emby-webcomponents/photoplayer/plugin",
-            "bower_components/emby-webcomponents/youtubeplayer/plugin"
-        ];
+        var list = ["bower_components/emby-webcomponents/playback/playbackvalidation", "bower_components/emby-webcomponents/playback/playaccessvalidation", "bower_components/emby-webcomponents/playback/experimentalwarnings", "bower_components/emby-webcomponents/htmlaudioplayer/plugin", "bower_components/emby-webcomponents/htmlvideoplayer/plugin", "bower_components/emby-webcomponents/photoplayer/plugin", "bower_components/emby-webcomponents/youtubeplayer/plugin"];
```
cvium: It was better before
```suggestion
        var list = [
            "bower_components/emby-webcomponents/playback/playbackvalidation",
            "bower_components/emby-webcomponents/playback/playaccessvalidation",
            "bower_components/emby-webcomponents/playback/experimentalwarnings",
            "bower_components/emby-webcomponents/htmlaudioplayer/plugin",
            "bower_components/emby-webcomponents/htmlvideoplayer/plugin",
            "bower_components/emby-webcomponents/photoplayer/plugin",
            "bower_components/emby-webcomponents/youtubeplayer/plugin"
        ];
```

src/scripts/site.js
```diff
@@ -420,586 +644,747 @@ var Dashboard = {
...
+        return new Promise(function (resolve, reject) {
+            require(["pluginManager"], function (pluginManager) {
+                pluginManager.loadPlugin(url).then(resolve, reject);
+            });
+        });
     }
 
     function enableNativeGamepadKeyMapping() {
-        return !(!window.navigator || "string" != typeof window.navigator.gamepadInputEmulation) && (window.navigator.gamepadInputEmulation = "keyboard", !0)
+        if (!(!window.navigator || "string" != typeof window.navigator.gamepadInputEmulation)) {
```
cvium: I hope I remember De Morgan's laws correctly...
```suggestion
        if (window.navigator && "string" == typeof window.navigator.gamepadInputEmulation) {
```

src/scripts/site.js
```diff
@@ -420,586 +644,747 @@ var Dashboard = {
...
+                    hashbang: true
+                });
+
+                if (!enableNativeGamepadKeyMapping() && isGamepadSupported()) {
+                    require(["bower_components/emby-webcomponents/input/gamepadtokey"]);
+                }
+
+                require(["bower_components/emby-webcomponents/thememediaplayer", "scripts/autobackdrops"]);
+
+                if (!("cordova" !== self.appMode && "android" !== self.appMode)) {
```
cvium:
```suggestion
                if ("cordova" === self.appMode || "android" === self.appMode) {
```

src/scripts/site.js
```diff
@@ -1213,103 +1619,314 @@ var Dashboard = {
...
+
+                if (!options) {
+                    options = {};
+                }
+
+                var url;
+                var itemType = item.Type || (options ? options.itemType : null);
+                var serverId = item.ServerId || options.serverId;
+
+                if ("settings" === item) {
```
cvium: We should probably convert all these to `switch` statements instead. It's out of the scope of this PR though.

src/scripts/site.js
vitorsemeano: fixed

src/scripts/site.js
vitorsemeano: sure, i will modify the cases i caught in this condition. I will arrange cases like this as long as i code. But cases already identified i fix them as you suggest.

src/scripts/site.js
vitorsemeano: done

src/scripts/site.js
```diff
@@ -420,586 +644,747 @@ var Dashboard = {
...
-                    //window.SyncRegistered = Dashboard.isConnectMode()
-                })
-            })
-        } catch (err) {
-            console.log("Error registering serviceWorker: " + err)
+        if (navigator.serviceWorker && "cordova" !== self.appMode && "android" !== self.appMode) {
+            try {
+                navigator.serviceWorker.register("serviceworker.js").then(function () {
+                    return navigator.serviceWorker.ready;
+                }).then(function (reg) {
```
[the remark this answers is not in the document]
vitorsemeano: should i remove the then you point (inclusing the registering of something related with the sync feature) or the then inside of it?

src/scripts/site.js
vitorsemeano: this could be converted to a switch but also moved to the module appRouter and not in site.js. I made a routes.js that removes all mappings from site.js but that is to be made in another PR. I dont remender if i already passed that code to appRouter module, but i could check later on.

src/scripts/site.js
```diff
@@ -420,586 +644,747 @@ var Dashboard = {
...
+                if (appHost.supports("fullscreenchange") && (browser.edgeUwp || -1 !== navigator.userAgent.toLowerCase().indexOf("electron"))) {
+                    require(["fullscreen-doubleclick"]);
+                }
+
+                if (appHost.supports("sync")) {
+                    initLocalSyncEvents();
+                }
+
+                if (!AppInfo.isNativeApp) {
+                    if (window.ApiClient) {
```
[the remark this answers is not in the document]
vitorsemeano: done

src/scripts/site.js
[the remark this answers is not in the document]
cvium: Yes, it's dead code.

src/scripts/site.js
```diff
@@ -32,23 +32,23 @@ function getParameterByName(name, url) {
 function pageClassOn(eventName, className, fn) {
     "use strict";
 
-    document.addEventListener(eventName, function (e__q) {
-        var target = e__q.target;
+    document.addEventListener(eventName, function (event) {
```
cvium: Just an fyi, `event` is also a global variable. Not that it matters here...

src/scripts/site.js
```diff
@@ -624,19 +608,16 @@ var AppInfo = {};
...
             console.log("Using system fonts with explicit sizes");
-            deps.push("systemFontsSizedCss");
+            require(["systemFontsSizedCss"]);
         } else {
             console.log("Using default fonts");
-            deps.push("systemFontsCss");
+            require(["systemFontsCss"]);
         }
 
-        deps.push("css!css/librarybrowser");
```
JustAMan: why did you drop this?

src/scripts/site.js
```diff
@@ -1219,106 +1208,87 @@ var AppInfo = {};
...
-        }
+        window.Emby = {};
 
         console.log("onAppReady - loading dependencies");
 
-        require(deps, function (appHost, pageObjects) {
-            if (console.log("Loaded dependencies in onAppReady"), window.Emby = {}, isBackgroundSync) {
-                return void syncNow();
+        if (isInBackground) {
+            syncNow();
```
JustAMan: so you're no longer loading `apphost` in background, why?

src/scripts/site.js
```diff
@@ -1354,19 +1324,18 @@ var AppInfo = {};
...
-        }
-
         initRequireWithBrowser(browser);
 
         if (self.appMode === 'cordova' || self.appMode === 'android' || self.appMode === 'standalone') {
             AppInfo.isNativeApp = true;
         }
 
-        require(initialDependencies, init);
+        if (!window.Promise || browser.web0s) {
```
JustAMan: why this was moved down?

src/scripts/site.js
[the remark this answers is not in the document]
vitorsemeano: removed

src/scripts/site.js
vitorsemeano: it's a global variable only for a few selected browsers. For example, firefox don't define event as a global variable, so it's safer to use the event passed as argument.

src/scripts/site.js
```diff
@@ -1,225 +1,337 @@
...
+
+        var urlLower = window.location.href.toLowerCase();
+        var index = urlLower.lastIndexOf("/web");
+
+        if (-1 != index) {
+            return urlLower.substring(0, index);
+        }
+
+        var loc = window.location;
+        var address = loc.protocol + "//" + loc.hostname;
```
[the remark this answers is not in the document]
vitorsemeano: what do you mean by url prefix?

src/scripts/site.js
cvium: I was more hinting at renaming the variable actually since I dislike shadowing

src/scripts/site.js
```diff
@@ -1,225 +1,337 @@
...
+        };
+        caps.IconUrl = appHost.deviceIconUrl();
+        caps.SupportsSync = appHost.supports("sync");
+        caps.SupportsContentUploading = appHost.supports("cameraupload");
+        appHost.getPushTokenInfo();
+        return caps = Object.assign(caps, appHost.getPushTokenInfo());
+    }
+};
+var AppInfo = {};
+!function () {
```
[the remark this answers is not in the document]
vitorsemeano: For what i searched in internet, is used to invoke a function without parenthesis around the function definition. Also it returns true. This can be changed to include the parenthesis and run the function.

src/scripts/site.js
```diff
@@ -1222,103 +1650,314 @@ var Dashboard = {
...
+                if ("wizard" === item) {
+                    return "wizardstart.html";
+                }
+
+                if ("manageserver" === item) {
+                    return "dashboard.html";
+                }
+
+                if ("recordedtv" === item) {
+                    return "livetv.html?tab=3&serverId=" + options.serverId;
```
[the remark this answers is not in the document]
vitorsemeano: i really don't know, i didn't look into the API of the server and routes logic. i think that should be left to another PR with defined objectives.

src/scripts/site.js
```diff
@@ -1222,103 +1650,314 @@ var Dashboard = {
...
+
+                if ("Person" == itemType) {
+                    return "itemdetails.html?id=" + id + "&serverId=" + serverId;
+                }
+
+                if ("Recording" == itemType) {
+                    return "itemdetails.html?id=" + id + "&serverId=" + serverId;
+                }
+
+                if ("MusicArtist" == itemType) {
```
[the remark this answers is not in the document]
vitorsemeano: put all options into an array and making a indexOf to check and then returning, i will make those ajustments.

src/scripts/site.js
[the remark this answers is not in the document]
vitorsemeano: or remove the function and just run the code inside of it, but if that functions defines a scope, don't know how the app would hold.

src/scripts/site.js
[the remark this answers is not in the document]
vitorsemeano: done

src/scripts/site.js
vitorsemeano: i didn't, i moved to the require function that comes right after that line. Webpack can't resolve dynamic modules passed as variables to require or define methods. So for certain cases, i merge and require them on the require calling.

src/scripts/site.js
vitorsemeano: That's something i didn't assume it would load things into other modules. But when i tested this, it all worked as expected. Looking in more detail into the code, it loads the appHost before reaching this block of code, when it creates the connectionManager. Once the module loads one time, even if you require it a second time, it will not load a second time, it will only return the first load. If we can assume the load is always made when defining a connectionManager, the changes i made for this block of code is working as intended, because you only need the instance of the appHost module when it's not in background.

But I could add a single require before this condition, forcing the load for appHost, just in case in the event of not loading it before this happens.

Tell me what you think.

src/scripts/site.js
vitorsemeano: This was moved down so i could load the dependencies statically when that condition could be met. If that condition fails, it doesn't need any dependencies to be loaded, so a call the callback init directly.

src/scripts/site.js
vitorsemeano: Even if you don't require apphost anywhere, it will load on the define definition of apphost. Ofc each modules needs to be loaded as required, as all define will be dropped in the end of this conversion to webpack

src/scripts/site.js
vitorsemeano: Oh i see, but is the same thing here. Window.event is equal to that variable. Still do you want me to change the name?

src/scripts/site.js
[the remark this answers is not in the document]
JustAMan: That was more a stray thought, not an issue with your PR :)

src/scripts/site.js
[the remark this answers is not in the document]
JustAMan: My desire is to serve Jellyfin via URL prefix. There are three ways of making it available:
* direct access: https://somesite.com/
* subdomain: https://jellyfin.somesite.com/
* URL prefix: https://somesite.com/jellyfin/

And I don't see how the part I highlighted works in the third case... that can easily be a future PR problem, though.

src/scripts/site.js
JustAMan: My assumptions here are that `apphost` should be already loaded here, but can we make an assert or something to check it?

src/scripts/site.js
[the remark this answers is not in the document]
vitorsemeano: The first two are here in this implementation. The third is partially if the lastindexof /web is present in the full url. But if this is not the case, the third option will fail. To you find any case where we might put the web project avail without web path in the url?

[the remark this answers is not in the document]
JustAMan: No idea. But seeing that web UI is essentially a (mostly) single-page application full URL could be just "everything until last `/` in the URL before `#` symbol" IMHO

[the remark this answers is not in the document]
vitorsemeano: well index.html is the main point of entry. Next pages goes after the hash (#). Perhaps we could split url into /, then the last part could be discarded if it contains a dot. What do you think?

src/scripts/site.js
vitorsemeano: i am not sure. we could add a flag to window object to signal that appHost is loaded, or we could force a require for appHost. Both options are valid, what do you prefer?

vitorsemeano: I made a require before all that to ensure appHost is loaded for all function scope. Check if it's ok for you

src/scripts/site.js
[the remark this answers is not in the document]
vitorsemeano: further analysis to this function, is mainly used to get server address for apiclient. so what we need here is server domain essentially. For jellyfin server instance, this always works, for a client like android or similar, the server address is handled elsewhere, not sure where..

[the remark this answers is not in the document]
JustAMan: Can you drop a `TODO` to investigate URL prefix support, so this no longer blocks current PR? It's out of scope of this PR anyway.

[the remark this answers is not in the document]
vitorsemeano: it's done.

vitorsemeano:
Will do in a couple of hours, thanks

ploughpuff:
Hey guys, probably something I'm doing wrong, but with this change my browser is showing just a blank screen.  No content is loaded.  It seems the task ProcessWebSocketRequest() is never triggered during startup phase...?

vitorsemeano:
I made #180 that should solve the issue @ploughpuff

ploughpuff:
My issue in above comment appears to be resolved by PR #196.

PR #164, opened 2019-03-01
 Better progressive playback controls of bottom pages

xlinbsd:
The volume slider is important to conserve at most width for audio playing.
The new value of hiding seems best balanced.

Only tested on last Firefox on Linux. Other tests for a better compromise are welcome.

### DIFF
--- a/src/bower_components/emby-webcomponents/nowplayingbar/nowplayingbar.css
+++ b/src/bower_components/emby-webcomponents/nowplayingbar/nowplayingbar.css
@@ -119,40 +119,47 @@
     height: 1.2em !important;
 }
 
+@media all and (max-width: 73em) {
+
+    .nowPlayingBarRight .nowPlayingBarUserDataButtons {
+        display: none;
+    }
+}
 
 @media all and (max-width: 70em) {
 
-    .nowPlayingBarUserDataButtons {
+    .nowPlayingBarRight .toggleRepeatButton {
         display: none;
     }
 }
 
 @media all and (max-width: 67em) {
 
-    .nowPlayingBar .muteButton, .nowPlayingBar .unmuteButton, .toggleRepeatButton {
+    .nowPlayingBarCenter .nowPlayingBarCurrentTime {
         display: none !important;
     }
-}
 
-
-@media all and (max-width: 64em) {
-
-    .nowPlayingBarCurrentTime {
-        display: none !important;
-    }
-    
 }
 
 @media all and (max-width: 61em) {
 
-    .nowPlayingBarCenter {
+    .nowPlayingBarCenter .stopButton {
         display: none !important;
     }
 
 }
 
 
-@media all and (min-width: 61em) {
+@media all and (max-width: 55em) {
+
+    .nowPlayingBarCenter .nextTrackButton, .nowPlayingBarCenter .previousTrackButton, .nowPlayingBarCenter .playPauseButton {
+        display: none !important;
+    }
+
+}
+
+
+@media all and (min-width: 55em) {
 
     .nowPlayingBarRight .playPauseButton {
         display: none;
@@ -161,7 +168,7 @@
 }
 
 @media all and (max-width: 41em) {
-    
+
     .nowPlayingImage {
         display: none;
     }
@@ -174,5 +181,13 @@
     .nowPlayingBarVolumeSliderContainer {
         display: none !important;
     }
-   
+
+}
+
+@media all and (max-width: 27em) {
+
+    .nowPlayingBar .muteButton, .nowPlayingBar .unmuteButton {
+        display: none !important;
+    }
+
 }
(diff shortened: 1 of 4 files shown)


xlinbsd:
Tested with latests Linux Chromium and Electron OK too.

sparky8251:
Do be aware, there are 2 other places in code that the volume slider exists. Its also in videoosd.js and remotecontrol.js.

Might not have to touch the videoosd.js one, but the remotecontrol.js is actually used in the second music playback screen. The one where you have your queued playlist along the bottom.

xlinbsd:
@sparky8251 OK Seen. Thanks

xlinbsd:
Hello, is the new ordering please you ?

xlinbsd:
I included my modifications of second pull since it is in the same objective to improve control bars

sparky8251:
Have any before and after pictures?

xlinbsd:
@sparky8251 I'll do it

xlinbsd:
after

xlinbsd:
Before.

before

xlinbsd:
Tested on latest iOS Safari, Lineage 16, macos 13's Safari, Linux Firefox && Linux Chromium

sparky8251:
It's looking good. Let us know when you feel its "complete".

xlinbsd:
I think it is complete for now. Need other users testing (in 10.2.3 ?) the new control style to be sure it is better for majority of us, no ?
I'll stay close if you think I can help a bit, I'm on the Riot room. I do know more than just some CSS coding.

src/bower_components/emby-webcomponents/nowplayingbar/nowplayingbar.css
```diff
@@ -119,38 +119,75 @@
...
         display: none !important;
     }
+
 }
 
-@media all and (max-width: 50em) {
+@media all and (max-width: 61em) {
 
-    .nowPlayingBarCenter {
+    .nowPlayingBarCenter .stopButton {
```
vitorsemeano: Why did you add the stopButton class here? this is inside the div with the class nowPlayingBarCenter

src/bower_components/emby-webcomponents/nowplayingbar/nowplayingbar.css
```diff
@@ -119,38 +119,75 @@
...
 
-    .toggleRepeatButton {
-        display: none;
+}
+
+
+@media all and (max-width: 55em) {
+
+    .nowPlayingBarCenter .nextTrackButton, .nowPlayingBarCenter .previousTrackButton, .nowPlayingBarCenter .playPauseButton {
+        display: none !important;
```
vitorsemeano: did you try without the !important? can't see to find more classes that use this sub classes with a display none important.

Also, nowPlayingBarCenter is parent of some of those classes. Try removing all sub classes of this one.

src/bower_components/emby-webcomponents/nowplayingbar/nowplayingbar.js
```diff
@@ -56,7 +56,7 @@ define(['require', 'datetime', 'itemHelper', 'events', 'browser', 'imageLoader',
 
         html += '<button is="paper-icon-button-light" class="muteButton mediaButton"><i class="md-icon">&#xE050;</i></button>';
 
-        html += '<div class="sliderContainer nowPlayingBarVolumeSliderContainer hide" style="width:100px;vertical-align:middle;display:inline-flex;">';
+        html += '<div class="sliderContainer nowPlayingBarVolumeSliderContainer hide" style="width:150px;vertical-align:middle;display:inline-flex;">';
```
vitorsemeano: I really don't like !important. could that be because of this inline style? display inline-flex? perhaps moving this rules to nowplayingbar.css will ease the !important? i could be asking too much already

src/bower_components/emby-webcomponents/nowplayingbar/nowplayingbar.css
vitorsemeano: forget that last part, you are applying the rules to the sub classes, not the parent one.

src/bower_components/emby-webcomponents/nowplayingbar/nowplayingbar.css
vitorsemeano: also forget this one, you are only applying the rule to stopButton, makes sense.

src/bower_components/emby-webcomponents/nowplayingbar/nowplayingbar.css
xlinbsd: I never myself use !important. That is a residue of the old code. I'll try without, it is a good idea.

src/bower_components/emby-webcomponents/nowplayingbar/nowplayingbar.js
xlinbsd: I will check if !important is really important.

xlinbsd: After some tests, `!important` is needed in this state of code. I will check why

xlinbsd:
Sorry for the last pushes, I was editing directly in github since I decided to do it right with git commands from my pc. I missed some editings while trying to do things right.

PR #172, opened 2019-03-12
Add web path as config option

dkanada:
related to jellyfin/jellyfin#1092


src/dashboard/dashboardgeneral.js
```diff
@@ -109,6 +109,23 @@ define(["jQuery", "loading", "fnchecked", "emby-checkbox", "emby-textarea", "emb
...
+            require(["directorybrowser"], function(directoryBrowser) {
+                var picker = new directoryBrowser;
+                picker.show({
+                    callback: function(path) {
+                        if (path) {
+                            view.querySelector("#txtWebPath").value = path;
+                        }
+                        picker.close();
+                    },
+                    validateWriteable: true,
```
cvium: Why?

joshuaboniface:
I agree with @JustAMan, the server-side change is something for packaging and really advanced users, not something that should be configurable within the UI itself.

dkanada:
I still think I should be able to edit the option from here as well, and I did add a warning to the description, but I guess I can remove it entirely from the general settings page if you think the warning isn't fair enough. I could also trigger a modal first with an extra large warning if that would be amenable, that sounds like plenty warning for inexperienced users.

src/strings/en-us.json
```diff
@@ -1123,7 +1123,7 @@
     "LabelMetadataPath": "Metadata path:",
     "LabelMetadataPathHelp": "Specify a custom location for downloaded artwork and metadata.",
     "LabelWebPath": "Web path:",
-    "LabelWebPathHelp": "The path where the web client source is located.",
+    "LabelWebPathHelp": "The path where the web client source is located. Do not change this unless you plan on moving the web files, or the web interface will break.",
```
JustAMan: I would rather we _not_ let users edit this via web UI (warning text is something no one would actually read).

cvium:
Changing the ui path feels more like a startup option. What's the use case exactly?

joshuaboniface:
So to clarify based on Riot chats around this - I like the idea of *showing* the web path, but not the idea of it being changeable from the UI - too much potential for a chicken-and-egg problem there if someone typos it. @dkanada if you can adjust it to just be for show I approve.

joshuaboniface:
Discussed more on Riot with @dkanada. I can see the usecase of setting this from the GUI for quick testing across all paltforms. Assuming it actually does save the configured value in `server.xml`and it works after a restart (@dkanada to test) , I'm OK with this option *with* a modal bearing a big "DANGER this is advanced and can break your install, don't change this unless you know what you're doing" on it (*before* the dialog selector) as the safety against users shooting themselves in the foot.

@JustAMan while I understand the argument, I think the potential for additional admin configurability outweighs the possible danger as long as the warning is clear enough. If the user clicks through a big warning and breaks it, OK, we can put a blurb in the docs.

dkanada:
I just removed it for now, the visible section on the dashboard is working fine though so this can be merged.

src/strings/en-us.json
```diff
@@ -1329,6 +1329,7 @@
     "LabelVideoCodec": "Video: {0}",
     "LabelVideoType": "Video Type:",
     "LabelView": "View:",
+    "LabelWeb": "Web",
```
JustAMan:
```suggestion
    "LabelWeb": "Web UI",
```
maybe?

PR #177, opened 2019-03-12
Conversion to webpack - part2 - emby-webcomponents into components and module naming resolution

vitorsemeano:
This PR is the second part of the webpack conversion that i am working on. In this PR, all emby-webcomponents were moved to components to unify all components into a single location. 

Also there were some module resolutions cases like alerts and apiInput.

The next part will focus more on module resolution naming.


src/components/skinmanager.js
```diff
@@ -219,7 +219,7 @@ define(['apphost', 'userSettings', 'browser', 'events', 'pluginManager', 'backdr
             selectedTheme = defaultTheme;
         }
 
-        var embyWebComponentsBowerPath = 'bower_components/emby-webcomponents';
```
JustAMan: place a TODO to fix variable name?

src/scripts/site.js
```diff
@@ -420,14 +420,14 @@ var AppInfo = {};
                 return self.ResizeObserver;
             });
         } else {
-            define("ResizeObserver", ["bower_components/emby-webcomponents/resize-observer-polyfill/ResizeObserver"], returnFirstDependency);
+            define("ResizeObserver", ["components/resize-observer-polyfill/ResizeObserver"], returnFirstDependency);
         }
     }
 
     function initRequireWithBrowser(browser) {
         var bowerPath = getBowerPath();
```
JustAMan: can we remove or rename these (both variable and function)?

src/scripts/site.js
```diff
@@ -454,8 +454,6 @@ var AppInfo = {};
             define("apiclient", ["bower_components/emby-apiclient/apiclient"], returnFirstDependency);
         }
 
-        define("actionsheet", ["webActionSheet"], returnFirstDependency);
```
JustAMan: why remove?

src/scripts/site.js
```diff
@@ -478,11 +476,7 @@ var AppInfo = {};
 
         var preferNativeAlerts = browser.tv;
```
JustAMan: this is probably not needed anymore

src/scripts/site.js
```diff
@@ -1240,29 +1217,29 @@ var AppInfo = {};
...
-                        require(["bower_components/emby-webcomponents/nowplayingbar/nowplayingbar"]);
+                        require(["components/nowplayingbar/nowplayingbar"]);
                     }
 
                     if (appHost.supports("remotecontrol")) {
-                        require(["playerSelectionMenu", "bower_components/emby-webcomponents/playback/remotecontrolautoplay"]);
+                        require(["playerSelectionMenu", "components/playback/remotecontrolautoplay"]);
                     }
 
                     if (!(appHost.supports("physicalvolumecontrol") && !browser.touch || browser.edge)) {
```
JustAMan: I wonder why we don't show volumecontrol for Edge... is it intentional, or a bug?

src/components/skinmanager.js
vitorsemeano: made a commit renaming all embyWebComponentsBowerPath to componentsPath. This fixes the TODO?

src/scripts/site.js
vitorsemeano: Well the objective is that when webpack conversion is done, there will be no components inside bower_components. And if still remain some, it will be properly renamed.

src/scripts/site.js
vitorsemeano: Because webActionSheet is defined as actionSheet and is only used to define actionsheet. So, i fixed the define for actionsheet, removing webActionSheet as a actual module, because WebActionSheet and actionsheet is the same module. 

This looks confusing, but is stupid to begin with defining something like this.

src/scripts/site.js
vitorsemeano: If the client never enter in a condition when browser.tv is true, yes this is useless. But there are lots of code regarding browser.tv. This should be targeted in another PR to remove all browser.tv related code. What do you think? 

Even if we add future support to put this inside a tv, the default alert mechanism should work out of the box, since it's based on dom manipulation to achieve a dialog.

src/scripts/site.js
vitorsemeano: Digging a little deep, i found that the control event for video html5 player is volumechange. This event is introduced in HTML5 specification, so it should be supported in edge without major problems. I think is fine to remove edge in this condition. Should i remove it?

src/scripts/site.js
```diff
@@ -1317,7 +1294,7 @@ var AppInfo = {};
 
         if (!window.Promise || browser.web0s) {
             initialDependencies.push();
```
cvium: What is this?

src/components/skinmanager.js
JustAMan: yeah, thx!

src/scripts/site.js
JustAMan: I was referring to the `preferNativeAlerts` variable only (`browser.tv` is needed, I'm not arguing with it).

src/scripts/site.js
JustAMan: I can only think of some funny edge (pun unintended) cases like Edge running as default browser on XBox maybe?..

src/scripts/site.js
vitorsemeano: Oh right, so looking into that, one could assume all clients support the non native method present here. We want that UI is equal to every client, so.. don't see the need to use nativeconfirm and nativeprompt here (window.confirm and window.prompt). What's your opinion? If you agree, i can remove nativeprompt and nativeconfirm, removing also preferNativeAlerts all together.

src/scripts/site.js
vitorsemeano: I can't test this for real and a real search on the web shows nothing regarding such issue. Edge should support html5 video without problems per spec: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video

src/scripts/site.js
vitorsemeano: I forgot to remove initialDependencies. The objective here was to remove initialDependencies from the require method. For webpack to resolve the require modules, it can't be passed as an array like initialDependencies. So i refactored the method so i could make the require explicitly.

src/scripts/site.js
JustAMan: I guess it was there for a reason, maybe native alerts/prompts on TVs could be more easy to navigate using remotes?.. maybe leave them in for now...

src/scripts/site.js
JustAMan: Then I suggest removing it, we can always fix it again later if things break.

src/scripts/site.js
vitorsemeano: sure np

src/scripts/site.js
vitorsemeano: condition removed.

PR #178, opened 2019-03-13
Remove unused localization strings (updated)

lostmypillow:
Improved version of original PR

**Changes**

**Issues**


src/strings/en-us.json
```diff
@@ -771,12 +760,9 @@
     "HeaderSupportTheTeam": "Support the Jellyfin Team",
     "HeaderSync": "Sync",
     "HeaderSyncJobInfo": "Sync Job",
-    "HeaderSyncRequiresSub": "Downloading requires an active Jellyfin Premiere subscription.",
-    "HeaderSystemDlnaProfiles": "System Profiles",
```
JustAMan: why is this gone?

lostmypillow: Since Sync doesn't require Premiere now, I thought it can be removed, so that when it syncs, it syncs, without prompting a Premiere message. As for DLNA profiles, I think it's a misclick. I'll add it back once you confirm Sync requires Sub deletion (or not, then I'll just add it back)

JustAMan: My comment is only about DLNA, the other removal is fine.

PR #187, opened 2019-03-16
Add CI for eslint

dkanada:
We dont have clean code yet so I couldn't enforce any actual rules but this will at least check for valid Javascript which is nice. I ran eslint locally but have no experience with CI so please check those files.


JustAMan:
This should be fixing #2

PR #188, opened 2019-03-16
Conversion to webpack - part3 - routes isolation, module resolution, refactoring appStorage

vitorsemeano:
Here goes the part 3 of the webpack conversion. This PR focus on module resolution for webpack, and it also focus in some relevant topics:

- Routes isolation: moved all controllers to a specific directory, and moved all routes definition from site.js to routes.js
- Refactor appStorage: merged all appStorage implementations into one single module
- Polifys removal: removed a bunch of unnecessary polifys, since the targets for this would be using a recent client

I also tested this PR with android, fixing some issues.


src/components/backdrop/package.json
```diff
@@ -0,0 +1,3 @@
+{
```
dkanada: Why did these folders get `package.json` files?

vitorsemeano: That's one of the ways webpack is able to resolve a specific module. When your have multiple js files inside a dir and only one is required by site.js. then the main js file requires the rest of the files in that dir. Ir could be other types of files inside a dir. The other way of resolving is defining a module in webpack.config with the exact path to it.

vitorsemeano: When you have a direct with multiple files inside that are required by site.js, I map that directory as a possible modules path that webpack can look for modules.

src/scripts/routes.js
```diff
@@ -0,0 +1,440 @@
...
+            roles: "admin",
+            controller: "device"
+        });
+        defineRoute({
+            path: "/dlnaprofile.html",
+            autoFocus: false,
+            roles: "admin"
+        });
+        defineRoute({
+            path: "/dlnaprofiles.html",
```
JustAMan: is it intentional that there's `dlnaprofile` and plural `dlnaprofiles`?

src/scripts/routes.js
```diff
@@ -0,0 +1,440 @@
...
+            autoFocus: false,
+            roles: "admin"
+        });
+        defineRoute({
+            path: "/dlnaserversettings.html",
```
JustAMan: what's the difference between `dlnaserversettings` and `dlnasettings`? can JF be a dlna client?..

src/scripts/routes.js
```diff
@@ -0,0 +1,440 @@
...
+    "emby-checkbox",
+    "emby-slider",
+    "listViewStyle",
+    "dashboardcss"], function () {
+
+    function defineRoute(newRoute, dictionary) {
+        var baseRoute = Emby.Page.baseUrl();
+        var path = newRoute.path;
+        path = path.replace(baseRoute, "");
+        console.log("Defining route: " + path);
```
JustAMan: hmm, why?..

src/scripts/site.js
```diff
@@ -620,514 +589,6 @@ var AppInfo = {};
...
-        });
-        defineRoute({
-            path: "/channelsettings.html",
-            dependencies: [],
-            autoFocus: false,
-            roles: "admin"
-        });
-        defineRoute({
-            path: "/addserver.html",
-            dependencies: ["emby-button", "emby-input"],
```
JustAMan: why dependencies are gone in new code?

src/scripts/site.js
```diff
@@ -620,514 +589,6 @@ var AppInfo = {};
...
-        });
-        defineRoute({
-            path: "/",
-            isDefaultRoute: true,
-            autoFocus: false,
-            dependencies: []
-        });
-    }
-
-    function getPluginPageContentPath() {
```
JustAMan: why remove?

src/scripts/site.js
```diff
@@ -1196,30 +657,27 @@ var AppInfo = {};
...
-                    defineCoreRoutes(appHost);
+                window.Emby.Page = appRouter;
 
+                require(['scripts/themeloader', 'libraryMenu', 'scripts/routes'], function () {
                     Emby.Page.start({
                         click: false,
                         hashbang: true
                     });
-
+
```
JustAMan: trailing spaces, here and below

src/bower_components/apiclient/appStorage.js
```diff
@@ -0,0 +1,52 @@
...
+        this.getItem = function(name) {
+            return localStorage.getItem(name);
+        };
+        
+        this.removeItem = function(name) {
+            localStorage.removeItem(name);
+
+            if (this.localData) {
+                this.localData[name] = null;
+                delete this.localData[name];
```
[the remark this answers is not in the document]
vitorsemeano: you don't. Take note that i just moved the code from one file to another mostly, making some fixes here and there. But i am very open to suggestion. I will make the cleanup so only delete persists.

src/bower_components/apiclient/appStorage.js
```diff
@@ -0,0 +1,52 @@
...
+                updateCache(this);
+            }
+        };
+
+        try {
+            self.caches && caches.open("embydata").then(onCacheOpened.bind(this));
```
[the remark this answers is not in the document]
vitorsemeano: I will make the proper adjustments as you suggest on a single commit.

src/components/appRouter.js
```diff
@@ -14,31 +14,6 @@ define(['loading', 'globalize', 'events', 'viewManager', 'layoutManager', 'skinM
         },
         showSettings: function () {
             show('/settings/settings.html');
-        },
```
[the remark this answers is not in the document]
vitorsemeano: Well almost all function removals you see are redefined in site.js. I just removed because they were doing nothing. Then we have the handleBackToDefault method that i also removed. That function would only be called when authenticate was called. In my point of view, that doesn't make any sense making a back operation right after a authenticate operation. Why would i want to exit the app (This can lead to a exit for a tizen context) when making a login? doesn't make much sense to me, so that's why i remove the condition and invocation. If that proofs any future problem, then we deal with it.

I also made a refactoring to the controllers that are loaded. Webpack doens't like dynamic module resolution, but it supports partial dynamic resolution. We can achieve that by using string literals. It gives a base location for webpack to fetch all files into possible locations given the literal. So that solves the problem for full dynamic module resolution by webpack.

I think it's all what i changed in this file.

src/components/skinManager.js
```diff
@@ -1,154 +1,9 @@
 define(['apphost', 'userSettings', 'browser', 'events', 'pluginManager', 'backdrop', 'globalize', 'require', 'appSettings'], function (appHost, userSettings, browser, events, pluginManager, backdrop, globalize, require, appSettings) {
     'use strict';
 
-    var currentSkin;
```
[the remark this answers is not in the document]
vitorsemeano: Because this was doing nothing. Theme support is visible in the clients, skin support is not supported. When i was working out this code, i asked in riot for this functionality, and they said this never worked. Can't remember who said it, but i confirmed it and removed the associated code accordingly.

src/components/skinManager.js
```diff
@@ -159,20 +14,61 @@ define(['apphost', 'userSettings', 'browser', 'events', 'pluginManager', 'backdr
...
+        if (options.start) {
+            Emby.Page.invokeShortcut(options.start);
+        } else {
+            Emby.Page.goHome();
         }
+    };
 
-        return [];
-    }
+    function getThemes() {
```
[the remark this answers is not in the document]
vitorsemeano: Well it comes from site.js. It was hardcoded in site.js, so i just moved the code from one place to another.

src/scripts/routes.js
```diff
@@ -0,0 +1,440 @@
...
+        newRoute.dictionary = newRoute.dictionary || dictionary || "core";
+        Emby.Page.addRoute(path, newRoute);
+    }
+
+    console.log("Defining core routes");
+        defineRoute({
```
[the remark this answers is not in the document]
vitorsemeano: done

src/scripts/routes.js
vitorsemeano: Don't have a clue, i just moved that code from site.js into routes.js for routes isolation.

src/scripts/routes.js
vitorsemeano: I don't know, i really didn't look into what each route does. Just kept the already made routes and general logic intact.

src/scripts/routes.js
vitorsemeano: Well i just moved from site.js to this file. But i made a commit optimizing the function.

src/scripts/site.js
vitorsemeano: They aren't. They are all required in the start of the module routes.js. The effect is the same, since if you manage to load all pages, the result would be requiring all modules present in dependencies. No specific usage for then in routes, just requiring is necessary.

src/scripts/site.js
vitorsemeano: probably a function that's not being used anywhere. I confirmed this with the other PR that don't have these changes.

src/scripts/site.js
vitorsemeano: done.

src/scripts/routes.js
JustAMan: Can you add a `// TODO: investigate` or something?

src/scripts/routes.js
JustAMan: Can you add a `// TODO: investigate` or something?

src/bower_components/apiclient/appStorage.js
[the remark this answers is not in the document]
JustAMan: So either clean it now or make a `TODO` comment IMHO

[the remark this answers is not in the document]
vitorsemeano: I cleaned yesterday, please check again if it's ok for you.

src/bower_components/apiclient/appStorage.js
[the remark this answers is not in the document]
vitorsemeano: I already made proper changes as you proposed. Please check again.

src/scripts/routes.js
vitorsemeano: Looking into the code, dlnaprofiles is the page to list all dlna profiles in server settings. dlnaprofile is the detail page for a given dlna profile that you select. So i should say it's intentional and makes sense this way.

src/scripts/routes.js
vitorsemeano: Atually, dlnaserversettings route doesn't have a html file nor any usage in the app. So i will remove it.

vitorsemeano: done

vitorsemeano:
I addresses all our comments, please check them again.

vitorsemeano:
For what i remember, i discussed that with @thornbill, and he said something like the code was there but it never worked or something similar. That's why i proceeded with removing it.

JustAMan:
> like the code was there but it never worked or something similar

I think that we can re-implement skins later if we want to. Less code to maintain for now is good - we need a clean foundation for building up anyway.

joshuaboniface:
@vitorsemeano Looks like we have some conflicts, are you able to resolve those?

vitorsemeano:
I have merged the changes from upstream, but i think we should wait for PR #197 to be merged into upstream before this one.

joshuaboniface:
Darn @vitorsemeano #197 caused two more conflicts, able to double-check them?

vitorsemeano:
Can you check if the merge is ok? I didn't tested. I can only test later tonight.

vitorsemeano:
After some mess in this branch trying to rebase it, i think i successfully managed to rebase it as intented.

PR #202, opened 2019-03-20
remove sync and registration services

dkanada:
This probably needs a lot of testing, but it seems to work on my device so far.


src/components/recordingcreator/recordingfields.template.html
```diff
@@ -1,20 +1,8 @@
 <div class="convertRecordingsContainer hide infoBanner" style="margin: 1em 0 2em;">
```
cvium: This seems Premiere related.

src/scripts/itemdetailpage.js
```diff
@@ -286,7 +281,11 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "cardBuild
...
         var canPlay = reloadPlayButtons(page, item);
-        (item.LocalTrailerCount || item.RemoteTrailers && item.RemoteTrailers.length) && -1 !== playbackManager.getSupportedCommands().indexOf("PlayTrailers") ? hideAll(page, "btnPlayTrailer", !0) : hideAll(page, "btnPlayTrailer"), setTrailerButtonVisibility(page, item), item.CanDelete && !item.IsFolder ? hideAll(page, "btnDeleteItem", !0) : hideAll(page, "btnDeleteItem"), renderSyncLocalContainer(page, params, user, item), "Program" !== item.Type || canPlay ? hideAll(page, "mainDetailButtons", !0) : hideAll(page, "mainDetailButtons"), showRecordingFields(instance, page, item, user);
+        (item.LocalTrailerCount || item.RemoteTrailers && item.RemoteTrailers.length) && -1 !== playbackManager.getSupportedCommands().indexOf("PlayTrailers") ? hideAll(page, "btnPlayTrailer", !0) : hideAll(page, "btnPlayTrailer");
```
cvium:
```suggestion
        ((item.LocalTrailerCount || item.RemoteTrailers && item.RemoteTrailers.length) && -1 !== playbackManager.getSupportedCommands().indexOf("PlayTrailers")) {
            hideAll(page, "btnPlayTrailer", true);
        } else {
            hideAll(page, "btnPlayTrailer");
        }
```

src/scripts/itemdetailpage.js
```diff
@@ -286,7 +281,11 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "cardBuild
...
+        setTrailerButtonVisibility(page, item);
+        item.CanDelete && !item.IsFolder ? hideAll(page, "btnDeleteItem", !0) : hideAll(page, "btnDeleteItem");
```
cvium:
```suggestion
        if (item.CanDelete && !item.IsFolder) {
            hideAll(page, "btnDeleteItem", true);
        } else {
            hideAll(page, "btnDeleteItem");
        }
```

src/scripts/itemdetailpage.js
```diff
@@ -286,7 +281,11 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "cardBuild
...
+        "Program" !== item.Type || canPlay ? hideAll(page, "mainDetailButtons", !0) : hideAll(page, "mainDetailButtons");
```
cvium:
```suggestion
        if ("Program" !== item.Type || canPlay) {
            hideAll(page, "mainDetailButtons", true);
        } else {
            hideAll(page, "mainDetailButtons");
        }
```

src/scripts/itemdetailpage.js
```diff
@@ -1166,30 +1165,55 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "cardBuild
                     }))
                 }
             }
-            var currentItem, self = this,
-                apiClient = params.serverId ? connectionManager.getApiClient(params.serverId) : ApiClient;
+
+            var currentItem;
+            var self = this;
+            var apiClient = params.serverId ? connectionManager.getApiClient(params.serverId) : ApiClient;
```
JustAMan: this changes the logic, before both `self` and `apiClient` were global variables, now they're local, this doesn't seem right.

src/scripts/itemdetailpage.js
```diff
@@ -1166,30 +1165,55 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "cardBuild
...
                 libraryMenu.setTransparentMenu(!0), e.detail.isRestored ? currentItem && (setTitle(currentItem, connectionManager.getApiClient(currentItem.ServerId)), renderTrackSelections(page, self, currentItem, !0)) : reload(self, page, params), events.on(apiClient, "message", onWebSocketMessage), events.on(playbackManager, "playerchange", onPlayerChange)
-            }), view.addEventListener("viewbeforehide", function() {
-                events.off(apiClient, "message", onWebSocketMessage), events.off(playbackManager, "playerchange", onPlayerChange), libraryMenu.setTransparentMenu(!1)
-            }), view.addEventListener("viewdestroy", function() {
-                currentItem = null, self._currentPlaybackMediaSources = null, self.currentRecordingFields = null
+            });
+            view.addEventListener("viewbeforehide", function() {
+                events.off(apiClient, "message", onWebSocketMessage);
+                events.off(playbackManager, "playerchange", onPlayerChange);
+                libraryMenu.setTransparentMenu(!1);
```
JustAMan:
```suggestion
                libraryMenu.setTransparentMenu(false);
```

src/scripts/tvrecommended.js
```diff
@@ -204,7 +204,7 @@ define(["events", "inputManager", "libraryMenu", "layoutManager", "loading", "do
             initialTabIndex = currentTabIndex;
         self.initTab = function() {
             var tabContent = self.tabContent;
-            setScrollClasses(tabContent.querySelector("#resumableItems"), enableScrollX()), categorysyncbuttons.init(tabContent)
+            setScrollClasses(tabContent.querySelector("#resumableItems"), enableScrollX())
```
JustAMan:
```suggestion
            setScrollClasses(tabContent.querySelector("#resumableItems"), enableScrollX());
```

src/scripts/itemdetailpage.js
cvium: It's the same syntax in C. Multiple local variables are defined with the comma operator

src/components/recordingcreator/recordingfields.template.html
dkanada: I didn't remove this section because I honestly have no clue how well this section was scrubbed of premiere elements, I don't have a tuner for live TV so I couldn't test.

src/components/itemcontextmenu.js
```diff
@@ -433,102 +413,48 @@ define(['apphost', 'globalize', 'connectionManager', 'itemHelper', 'appRouter',
...
-                        getResolveFunction(resolve, id)();
+                    getResolveFunction(resolve, id)();
                     break;
-                    }
                 case 'queueallfromhere':
                     {
                     getResolveFunction(resolve, id)();
                     break;
                     }
-                case 'convert':
```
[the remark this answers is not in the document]
dkanada: I was removing convert because users were reporting the feature as broken. As far as I can tell it uses the sync endpoints, the frontend relies on the sync dialog, and we are probably going to rewrite most of the backend code later.

src/components/multiselect/multiselect.js
```diff
@@ -218,13 +218,6 @@ define(['browser', 'appStorage', 'apphost', 'loading', 'connectionManager', 'glo
                 });
             }
 
-            if (user.Policy.EnableContentDownloading && appHost.supports('sync')) {
-                menuItems.push({
-                    name: globalize.translate('Download'),
```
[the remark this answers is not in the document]
dkanada: This isn't the normal download, it seems to be some variation of sync, which we don't support at the moment.

src/scripts/itemdetailpage.js
dkanada: This is fine then right? I don't think the variables have changed scope.

dkanada:
The frontend has a habit of using separate components for things that should be the same (add library versus edit library) and using the same component for things that should be atomic (sync versus convert versus download) in my opinion.

src/scripts/itemdetailpage.js
JustAMan: My bad :) TIL

src/components/recordingcreator/recordingfields.template.html
cvium: But in `recordingfields.js`, you're only ever adding the `hide` css class and it's already the default.

src/components/recordingcreator/recordingfields.js
```diff
@@ -215,7 +204,7 @@ define(['globalize', 'connectionManager', 'serverNotifications', 'require', 'loa
                 });
             }
         } else {
-            showSingleRecordingFields(options.parent, options.programId, apiClient);
+            context.querySelector('.recordSeriesContainer').classList.add('hide');
```
dkanada: @cvium I am on the fence about this line since it wasn't explicitly added in `showSingleRecordingFields`

cvium: I don't think it's necessary. We can just have some of the live tv folks test the RC and re-add it if it looks funny to them.

PR #203, opened 2019-03-21
Remove all unused strings from translation files

dkanada:
The first commit was finishing up #178 since I didn't see Weblate removing the leftover keys from the other languages. For the second commit I had a script looking for uses matching each key in the English file.


joshuaboniface:
Weblate hasn't actually been pushed yet or anything, but yea might as well kill 'em.

PR #205, opened 2019-03-22
Move assets and third party libs out of components and update some strings

dkanada:
Also included a fix for a broken setting and the usual cleanup for whatever files I was looking at during the changes. This can wait until after #188 so I can rebase to fix the issues.


src/controllers/dashboardpage.js
```diff
@@ -1,11 +1,6 @@
 define(["datetime", "events", "itemHelper", "serverNotifications", "dom", "globalize", "loading", "connectionManager", "playMethodHelper", "cardBuilder", "imageLoader", "components/activitylog", "humanedate", "listViewStyle", "emby-linkbutton", "flexStyles", "emby-button", "emby-itemscontainer"], function (datetime, events, itemHelper, serverNotifications, dom, globalize, loading, connectionManager, playMethodHelper, cardBuilder, imageLoader, ActivityLog) {
     "use strict";
 
-    function onConnectionHelpClick(evt) {
```
JustAMan: why remove?

src/components/apiInput/apiInput.js
```diff
@@ -27,28 +27,21 @@ define(['connectionManager', 'playbackManager', 'events', 'inputManager', 'focus
...
     }
 
     function processGeneralCommand(cmd, apiClient) {
-
-        // Full list
-        // https://github.com/MediaBrowser/MediaBrowser/blob/master/MediaBrowser.Model/Session/GeneralCommand.cs#L23
-        //console.log('Received command: ' + cmd.Name);
-
+        // https://github.com/jellyfin/jellyfin/blob/master/MediaBrowser.Model/Session/GeneralCommandType.cs
+        console.log('Received command: ' + cmd.Name);
```
JustAMan: why is this uncommented?

src/scripts/site.js
```diff
@@ -316,20 +316,6 @@ var AppInfo = {};
     }
 
     function getSettingsBuilder(UserSettings, layoutManager, browser) {
```
JustAMan: can we just get rid of this function?..

src/components/apiInput/apiInput.js
dkanada: It's useful info for the logs.

src/controllers/dashboardpage.js
dkanada: It was originally used for Emby Connect.

src/scripts/site.js
dkanada: It would take a lot of refactoring and I didn't want to break too much in `site.js` while @vitorsemeano is spending so much time in that file.

PR #210, opened 2019-03-25
Add Rescan Button to headerRight section

tbraeutigam:
This adds a rescan button to all libraries/collections and allows every user to rescan/update a library.

**My use case**
I have a few different users who can upload their photos to my NAS (mounted via nfs) into their library. This allows them to trigger a scan for new photos immediately without cron jobs (or me) getting involved.
(Or Script-Fu to work with the API from the NAS.)

**Changes**
1. Add a Rescan button to the headerRight Section of the home
2. Add getParentId function for libraries not identified by 'topParentId'
3. Add Function to perform rescan

**Issues**
Outstanding: Currently refreshDialog does not provide an option to limit the rescan types, meaning a user can have the full power of an admin rescan (refresh all metadata + replace images).
Outstanding: Would be good to implement a permission/policy item to only allow select users to use this.

### DIFF
--- a/src/scripts/librarymenu.js
+++ b/src/scripts/librarymenu.js
@@ -118,6 +118,9 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
         if (headerSearchButton) {
             headerSearchButton.addEventListener("click", showSearch);
         }
+        if (headerRefreshButton) {
+            headerRefreshButton.addEventListener("click", refreshCurrentLibrary);
+        }
 
         headerUserButton.addEventListener("click", onHeaderUserButtonClick);
         headerHomeButton.addEventListener("click", onHeaderHomeButtonClick);
@@ -126,6 +129,19 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
 
         if (headerSettingsButton) {
             headerSettingsButton.addEventListener("click", onSettingsClick);
+        }
+    }
+
+    function refreshCurrentLibrary(){
+        var currentLibraryId = getTopParentId() || getParentId();
+        if (currentLibraryId !== null) {
+            require(["refreshDialog"], function(refreshDialog) {
+                new refreshDialog({
+                    itemIds: [currentLibraryId],
+                    serverId: getCurrentApiClient().serverId(),
+                    mode: "scan"
+                }).show()
+            })
         }
     }
 
@@ -582,6 +598,9 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
         }
     }
 
+    function getParentId() {
+        return getParameterByName("parentId") || null;
+    }
     function getTopParentId() {
         return getParameterByName("topParentId") || null;
     }
@@ -723,6 +742,21 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
         }
     }
 
+    function updateRefreshButton(){
+        if (!headerRefreshButton) {
+            headerRefreshButton = document.querySelector(".headerRefreshButton");
+        }
+
+        var refreshId = getTopParentId() || getParentId();
+        
+        if ( refreshId !== null){
+            headerRefreshButton.classList.remove("hide");
+        }
+        else {
+            headerRefreshButton.classList.add("hide");
+        }
+    }
+
     function initHeadRoom(elem) {
         require(["headroom-window"], function (headroom) {
             headroom.add(elem);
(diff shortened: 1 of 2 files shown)


tbraeutigam:
Couple of Changes in the commits:

-> refreshDialog can now be customized by passing an extra option:
```javascript
show: {
    simple: true,
    scanNewAndUpdated: true,
    updateMissing: false,
    replaceAll: false
}
```
If it's not there, it gets ignored.

=================
Testing done:
Unminimized running against 10.2.2 server

First Test: Open scan Dialog with new restrictions in place
Result: Shows as expected

Second Test: Trigger a library refresh via the restricted dialog
Result: Scan happens as expected

Third Test: Is the refreshDialog unaffected when opening from Admin Dashboard?
Result: It's unaffected. Works as expected

tbraeutigam:
Two Screenshots

**Restricted refreshDialog Example**
jellyfin-web-restrictedDialog

**Refresh Button in headerRight Section**
jellyfin-web-refresh-button

src/scripts/librarymenu.js
```diff
@@ -129,6 +132,24 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
         }
     }
 
+    function refreshCurrentLibrary(){
+        var currentLibraryId = getTopParentId() || getParentId();
+        if (currentLibraryId !== null) {
+            require(["refreshDialog"], function(refreshDialog) {
```
JustAMan: ping @vitorsemeano - does this impact your webpack efforts?

src/scripts/librarymenu.js
```diff
@@ -723,6 +747,21 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
...
     }
 
+    function updateRefreshButton(){
+        if (!headerRefreshButton) {
+            headerRefreshButton = document.querySelector(".headerRefreshButton");
+        }
+
+        var refreshId = getTopParentId() || getParentId();
+        
+        if ( refreshId !== null){
```
JustAMan:
```suggestion
        if (refreshId !== null) {
```

vitorsemeano:
I don't think this will significantly affect wepack conversion. 

I only have one suggestion, to put the button more to the left, some users may already be accustomed to the user on the top right corner of the screen. Others should give their opinion also. 

But for me is LGTM.

thornbill:
I agree with @vitorsemeano I'm not a fan of the placement here. I think it might be better suited on either side of the cast icon?

tbraeutigam:
Well, moving it is at least simple :)

Instead of right of the user icon, move it left of it?

My thoughts on the various positions:
Left of Cast: Makes it feel more 'important' than it is.
Left of Search: I would already personally want search to be left of Cast, as the 'natural' spot for a search icon is the left-most icon of the top-right header section to me (from various different designs)
Left of User: Works for me. I put it to the right of the user as the other 'disappearing' icon (Admin Dash) was there.

dkanada:
The dialog changes are a good change for sure. Would it be fine to extend the current placement to work for non-admins as well? Right click on the library on the home screen, which could also be added to the small widget. It doesn't seem ideal to add buttons that will vanish from the header depending on the current page.

I'm not sure we should add this functionality until we have a per-user scan option regardless. I recently enabled multiserver functionality for all instances and got a lot of comments about exposing too much functionality to regular users. This seems like a similar case to me if we don't include any server permissions.

joshuaboniface:
I agree with @dkanada, let's do the whole thing at once. I've moved this to 10.4.0 to have some more time to go through the backend portions since I do like the idea of per-user library scanning.

dkanada:
This actually brings up a design flaw with the current user settings, specifically regarding the media deletion. It allows a per-library setting but the UI isn't extensible at all. We should really figure out the best way to introduce multiple per-library settings without having crazy amounts of lists on that page. Both the user and library setting pages haven't been deminified and cleaned up yet either, so we should get that done at the same time.

EraYaN:
The best option would be a table/matrix IMO.
Libraries on the left, permissions on top.
Checkboxes in every cell.

EraYaN:
Alright this should still be merged at some point.

joshuaboniface:
Agreed - can I get a second review please @jellyfin/web?

anthonylavado:
Alright - official position is we're holding off on this until we can enable better permissions that can limit it's usage for the right reasons.

sparky8251:
Basically, we want this but its also exposing how crappy the current permissions setup is.

Have to fix permissions first before we can mainline this.

dkanada:
I'm closing this since it's getting stale, but the feature itself is fine. As soon as someone adds the required backend changes we can add it to the web client.

PR #211, opened 2019-03-26
Adds web options to support better password resets

LogicalPhallacy:
This is related to https://github.com/jellyfin/jellyfin/pull/1149 which adds per user password resets
**Changes**
Added support for selecting which password reset provider to use per user (if more than one is installed)
Added Strings to support the same for Authentication providers that were missing
Changed the password reset pin creation complete dialog to link to the reset page.
Replaced @anthonylavado 's double credit in the contributors doc with myself.


src/controllers/forgotpassword.js
```diff
@@ -12,7 +12,7 @@ define([], function() {
         });
         if ("PinCode" == result.Action) {
             var msg = Globalize.translate("MessageForgotPasswordFileCreated");
-            return msg += "<br/>", msg += "<br/>", msg += result.PinFile, msg += "<br/>", void Dashboard.alert({
+            return msg += "<br/>", msg += "<br/>", msg += "<a href=\"/web/index.html#!/forgotpasswordpin.html\">Enter PIN here to finish Password Reset</a>" ,msg += "<br/>",msg += result.PinFile, msg += "<br/>", void Dashboard.alert({
```
JustAMan: please **nooo**, this leading slash in a URL kills all ability to run this prefixed... there should be some `getBaseUrl()` or something.

src/scripts/useredit.js
```diff
@@ -21,11 +21,23 @@ define(["jQuery", "loading", "libraryMenu", "fnchecked"], function($, loading, l
         })
     }
 
+    function loadPasswordResetProviders(page, user, providers) {
+        providers.length > 1 && !user.Policy.IsAdministrator ? page.querySelector(".fldSelectPasswordResetProvider").classList.remove("hide") : page.querySelector(".fldSelectPasswordResetProvider").classList.add("hide");
```
JustAMan: wtf? :) I mean, don't manually write uglified code, we're trying to de-uglify the codebase ;)

```suggestion
        if (providers.length > 1 && !user.Policy.IsAdministrator) {
            page.querySelector(".fldSelectPasswordResetProvider").classList.remove("hide");
        } else {
            page.querySelector(".fldSelectPasswordResetProvider").classList.add("hide");
        }
```
Yes, it's not a one-liner, but it's _way_ easier to read.

Also, why check if user is admin?

src/scripts/useredit.js
```diff
@@ -21,11 +21,23 @@ define(["jQuery", "loading", "libraryMenu", "fnchecked"], function($, loading, l
...
+        var currentProviderId = user.Policy.PasswordResetProviderId;
+        page.querySelector(".selectPasswordResetProvider").innerHTML = providers.map(function(provider) {
+            var selected = provider.Id === currentProviderId || providers.length < 2 ? " selected" : "";
```
JustAMan:
```suggestion
            var selected = (provider.Id === currentProviderId || providers.length < 2) ? " selected" : "";
```
IMHO better to add explicit braces for the sake of clarity.

src/strings/en-us.json
```diff
@@ -961,7 +964,7 @@
     "MessageNoServersAvailableToConnect": "No servers are available to connect to. If you've been invited to share a server, make sure to accept it below or by clicking the link in the email.",
     "MessageNoTrailersFound": "No trailers found. Install the Trailer channel to enhance your movie experience by adding a library of internet trailers.",
     "MessageNothingHere": "Nothing here.",
-    "MessagePasswordResetForUsers": "Passwords have been removed for the following users. To login, sign in with a blank password.",
+    "MessagePasswordResetForUsers": "The following users have had their passwords reset. They can now sign in with the PIN code that was used this reset.",
```
JustAMan:
```suggestion
    "MessagePasswordResetForUsers": "The following users have had their passwords reset. They can now sign in with the PIN codes that were used to perform the reset.",
```

src/strings/en-us.json
```diff
@@ -1170,6 +1173,7 @@
     "PasswordResetComplete": "The password has been reset.",
     "PasswordResetConfirmation": "Are you sure you wish to reset the password?",
     "PasswordResetHeader": "Reset Password",
+    "PasswordResetProviderHelp": "Choose a Password Reset Provider to be used if this user resets their password",
```
JustAMan: what does this "if" mean? like "choose which provider to use when in a distant future this user wishes to change password", or "choose which provider to use right now to reset the password"?

src/scripts/useredit.js
LogicalPhallacy: This is cloned from the select userauth provider which is also admin only it seemed appropriate to keep this option admin only for now as well

src/strings/en-us.json
LogicalPhallacy: The former

JustAMan: Then maybe put it like `Choose a Password Reset Provider to be used when this user requests a password reset`

dkanada: It would also be nice to manually trigger a password reset if you want to add that.

src/controllers/forgotpassword.js
dkanada: Just adding the page itself as `forgotpasswordpin.html` will work fine.

LogicalPhallacy: Switched it to just the html file, that sounds good.

src/strings/en-us.json
LogicalPhallacy: Using @JustAMan's phrasing, Adding a button to trigger a password reset is a project for later, but definitely might be worth doing

src/strings/en-us.json
```diff
@@ -40,6 +40,7 @@
     "AspectRatio": "Aspect ratio",
     "AttributeNew": "New",
     "Audio": "Audio",
+    "AuthProviderHelp": "Select an Authentication Provider to be used to Authenticate this user's password",
```
EraYaN: Capitalization of `Authenticate` should probably be removed.

joshuaboniface:
```suggestion
     "AuthProviderHelp": "Select an Authentication Provider to be used to authenticate this user's password", 
```

EraYaN:
Since the string now has dashes can we make the field auto add dashes? (Like some serial number entry systems do)

LogicalPhallacy:
I'm going to merge and do the dash check on the backend

PR #218, opened 2019-03-30
Solve random dark pane over video playing in fullscreen mode

vitorsemeano:
This PR addresses a bug that supposedly renders a dark pane over the video when playing.

I managed to replicate this problem when one file ends and the next one starts playing by itselt, for a specific file. This issue is caused because the property `playOptions` of the item being playing is missing sometimes. The dark pane shows when the property fullscreen represents a false result to a condition. This condition represents when a video is playing but not in the full viewport space available. To mitigate this problem, fallback points has been added when getting playOptions with an existing function for the effect.

I already tested these changes with jf-android.


joshuaboniface:

PR #219, opened 2019-03-31
Change Chromecast volume instead while casting

Froghut:
When changing volume connected to chromecast change the device volume instead of the jellyfin-player volume. This is more in line with other cast enabled apps.

### DIFF
--- a/src/components/chromecast/chromecastplayer.js
+++ b/src/components/chromecast/chromecastplayer.js
@@ -779,12 +779,15 @@ define(['appSettings', 'userSettings', 'playbackManager', 'connectionManager', '
         });
     };
 
-    ChromecastPlayer.prototype.volumeDown = function () {
+     ChromecastPlayer.prototype.volumeDown = function () {
+        vol = this._castPlayer.session.receiver.volume.level;
+        if (vol == null)
+            vol = 0.5;
+        vol -= 0.02;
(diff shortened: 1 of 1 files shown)


src/components/chromecast/chromecastplayer.js
```diff
@@ -779,12 +779,15 @@ define(['appSettings', 'userSettings', 'playbackManager', 'connectionManager', '
         });
     };
 
-    ChromecastPlayer.prototype.volumeDown = function () {
+     ChromecastPlayer.prototype.volumeDown = function () {
+        vol = this._castPlayer.session.receiver.volume.level;
+        if (vol == null)
+            vol = 0.5;
```
JustAMan: either wrap in `{}` or do one line, this way it's error-prone

src/components/chromecast/chromecastplayer.js
```diff
@@ -799,24 +802,22 @@ define(['appSettings', 'userSettings', 'playbackManager', 'connectionManager', '
     };
 
     ChromecastPlayer.prototype.volumeUp = function () {
+        vol = this._castPlayer.session.receiver.volume.level;
+        if (vol == null)
+            vol = 0.5;
```
JustAMan: same

joshuaboniface:

PR #220, opened 2019-03-31
Add subtitle synchronization on HTML video player

redSpoutnik:
This PR add the ability to manually set an offset to some subtitles in HTML video player:
* subtitle stream codec must be compatible with vtt (e.g. srt) or ass.
* subtitle stream display mode must be external (subtitle extraction must be allowed).

**Changes**

* If previous conditions are fullfilled, a slider shown between subtitle and parameters buttons.
* Current offset range is -30 sec to +30 sec.
* On selected subtitle changes, the offset is reset to 0.
* This feature works only in UI and doesn't require backend call.

**Tested on Linux Mint 18.2 with :**
* Brave Browser 0.61.51
* Google Chrome 73.0.3683.75
* Firefox Quantum 65.0.1

**Issues**

May partially solve [jellyfin/jellyfin#130](https://github.com/jellyfin/jellyfin/issues/130).


redSpoutnik:
null-offset
negative-offset
positive-offset

cvium:
I'd prefer if it wasn't a slider and it should be inside the options menu tbh. The current placement is a bit crowded

JustAMan:
I would say it should be inside `CC` menu which appears when one clicks respective button.
Also having a field which could be filled manually (in addition to a slider) would be good.

redSpoutnik:
Thanks, I will take the time to test those changes and I will give you some feedback.

dkanada:
One quick note, I fixed the CI issue here in another PR so you won't have any issues in the future :) The backend changes look good, I agree about moving it out of the main bar though, I would say inside the settings button since that one is for general options. I think the subtitle and audio buttons should stay as track selection.

redSpoutnik:
Hi all,
I want to share with you my recent tests on this PR.

To avoid grouping slider and editable field in a tight menu item, I took inspiration from kodi subtitle offset.

If a subtitle track is selected, a localized *"SubtitleOffset"* option appears in the settings:
subtitleoffset-menu

This option display a panel with the slider and a text field:
subtitle-sync-overlay
subtitle-sync-overlay-background

The slider has the exact same behavior as before:
subtitle-sync-slider

The textfield is editable:
subtitle-sync-textfield

* Slider and textfield values are synchronized.
* On mouse inactivity the panel fades with bottom controls, except if the textfield is being edited.
* If the panel is closed using the cross, it will not show up until the *"SubtitleOffset"* option is clicked again, but the subtitle offset remain.

What do you think?

dkanada:
@redSpoutnik there seem to be a few conflicts but I think this is ready to merge once those are fixed up.

cc @JustAMan @cvium for comments

anthonylavado:
Ready to put this in as soon as the conflicts are sorted.

redSpoutnik:
@dkanada  @anthonylavado Conflicts merged :) but be aware that I let the "subtitle offset" button in settings, whereas the "subtitle settings" has been removed.
This PR doesn't include translations of the button's name btw.

anthonylavado:
> This PR doesn't include translations of the button's name btw.
We'll want to make an issue to track this!
Thanks for the great PR @redSpoutnik 👏

PR #221, opened 2019-04-01
Fix local chromecast play behind reverse proxy

agrenott:
**Changes**
When using a HTTPS reverse proxy (for eg. traefik in fron of docker
jellyfin), there's no valid advertised local address.
Let's default to external one in such case, relying on the home router
to properly route it internally.

**Issues**
Fixes https://github.com/jellyfin/jellyfin/issues/791 for me.

### DIFF
--- a/src/components/chromecast/chromecasthelpers.js
+++ b/src/components/chromecast/chromecasthelpers.js
@@ -187,8 +187,13 @@ define(['events'], function (events) {
         return apiClient.getEndpointInfo().then(function (endpoint) {
             if (endpoint.IsInNetwork) {
                 return apiClient.getPublicSystemInfo().then(function (info) {
-                    addToCache(serverAddress, info.LocalAddress);
-                    return info.LocalAddress;
+                    var localAddress = info.LocalAddress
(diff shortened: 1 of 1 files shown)


src/components/chromecast/chromecasthelpers.js
```diff
@@ -187,8 +187,13 @@ define(['events'], function (events) {
         return apiClient.getEndpointInfo().then(function (endpoint) {
             if (endpoint.IsInNetwork) {
                 return apiClient.getPublicSystemInfo().then(function (info) {
-                    addToCache(serverAddress, info.LocalAddress);
-                    return info.LocalAddress;
+                    var localAddress = info.LocalAddress
+                    if (! localAddress) {
```
JustAMan:
```suggestion
                    if (!localAddress) {
```

joshuaboniface:

src/components/chromecast/chromecasthelpers.js
agrenott: Right, fixed.

PR #222, opened 2019-04-01
Remember device volume/mute on Chromecast playback stop

Froghut:
when playback stops while connected to a chromecast device the playstate was completely reset, including volume/mute. Those should not be reset since the volume of the device does not change on playback stop and i.e. the remote control panel should still show the correct (device) volume.

### DIFF
--- a/src/components/chromecast/chromecastplayer.js
+++ b/src/components/chromecast/chromecastplayer.js
@@ -574,8 +574,15 @@ define(['appSettings', 'userSettings', 'playbackManager', 'connectionManager', '
 
             events.trigger(instance, "playbackstop", [state]);
 
+            var state = instance.lastPlayerData.PlayState || {};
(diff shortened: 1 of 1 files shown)


joshuaboniface:

src/components/chromecast/chromecastplayer.js
```diff
@@ -574,8 +574,15 @@ define(['appSettings', 'userSettings', 'playbackManager', 'connectionManager', '
 
             events.trigger(instance, "playbackstop", [state]);
 
+            var state = instance.lastPlayerData.PlayState || {};
+            var volume = state.VolumeLevel;
```
JustAMan: should we add a default like `|| 0.5`?.. and `false` to `mute`?

Froghut: yeah, makes sense.

PR #223, opened 2019-04-02
When playing on Chromecast make volume buttons change chromecast volume

Froghut:
The ideal solution would be to update the cordova chromecast plugin to the new CAF version, that would handle this automatically.
Until then this workaround will intercept the volume keys and change the chromecast device volume instead of the mobile phone's. I also changed the volume steps from 2% to 5% per press - otherwise changing volume would feel like it's taking forever.


anthonylavado:
Thank you for reminding me to test the CAF version

PR #224, opened 2019-04-03
Conversion to webpack - part5 - module resolution, data-require cleaned

vitorsemeano:
There's one more part to add to webpack conversion. This chunk of changes addresses the main points:

- Name module resolution
- Removes all data-require present in views, and also converts some of these requires intro controllers
- refactor some init code into connectionManager, with some cleanup associated (also deuglification)
- vibrant completely removed since this was a dead module.


src/bower_components/apiclient/connectionManager.js
```diff
@@ -0,0 +1,1292 @@
+define(["events", "apiclient", "appStorage", "credentialprovider", "apphost"], function (events, apiClientFactory, appStorage, credentialProvider, apphost) {
+    "use strict";
+
+    function getServerAddress(server, mode) {
+        switch (mode) {
```
dkanada: If you made any other changes besides deminification here then you might want to mirror them in the ES6 repo so we can migrate with less issues.

src/components/cardbuilder/roundcard.css
```diff
@@ -1,7 +0,0 @@
-.card-round:focus > .cardBox-focustransform {
-    transform: scale(1.26, 1.26);
```
dkanada: Isn't this used be people in the item details page?

src/library.html
```diff
@@ -1,4 +1,4 @@
-<div id="mediaLibraryPage" data-role="page" class="page type-interior mediaLibraryPage librarySectionPage withTabs fullWidthContent" data-require="scripts/medialibrarypage">
+<div id="mediaLibraryPage" data-role="page" class="page type-interior mediaLibraryPage librarySectionPage withTabs fullWidthContent">
```
dkanada: Lots of plugins use `data-require` so you can't remove that from the code itself until they get updated. You can remove it from the pages here though.

vitorsemeano: I believe most of those requires can be removed in the plugins repo. Most of them are the ones that begin with emby-*. Most of these are already required when loading the routes.js module. We need to look case by case to remove those requires, but should work as expected.

src/bower_components/apiclient/connectionManager.js
vitorsemeano: I modified this file to reflect some module resolution problems with this module. I can look into it and make a PR with the changes i made here into the connectionManager.js in the apiclient repo.

src/components/cardbuilder/roundcard.css
vitorsemeano: I didn't find any reference to this file, making a search in the repository i can't find any reference. Can you confirm it?

dkanada: right [here](https://github.com/jellyfin/jellyfin-web/blob/master/src/scripts/site.js#L787)

vitorsemeano: Right, but that module isn't used anywhere, so that's why I removed it. But I didn't checked on the plugins so it could be used there

vitorsemeano:
Removed all connectionmanager changes, i also reverted the data-require and data-controller removal.

I also proceeded into squashing all those module resolution commits into 2 commits.

I already tested in jf-android and it seems fine.

src/scripts/site.js
```diff
@@ -831,7 +829,6 @@ var AppInfo = {};
         define("viewSettings", [componentsPath + "/viewsettings/viewsettings"], returnFirstDependency);
         define("filterMenu", [componentsPath + "/filtermenu/filtermenu"], returnFirstDependency);
         define("sortMenu", [componentsPath + "/sortmenu/sortmenu"], returnFirstDependency);
-        define("connectionmanager", [apiClientBowerPath + "/connectionmanager"]);
```
dkanada: Should this be kept now?

src/scripts/dlnaprofile.js
```diff
@@ -1,371 +0,0 @@
...
-        var usersHtml = "<option></option>" + users.map(function(u) {
-            return '<option value="' + u.Id + '">' + u.Name + "</option>"
-        }).join("");
-        $("#selectUser", page).html(usersHtml).val(profile.UserId || ""), renderSubProfiles(page, profile)
```
dkanada: Since we are leaving webpack alone for now could you avoid the rename of this file so the diff will work properly? It shows up as two different files at the moment.

src/scripts/site.js
vitorsemeano: connectionmanager does nothing, connectionManager is the module that is used everywhere. connectionManager is defined after instantiate the connectionManagerFactory module.

src/scripts/dlnaprofile.js
vitorsemeano: seeing what i have done, dlnaprofile seems to be a controller, so to load that properly, the file needs to be inside the dir controllers. The commit where i moved it shows as a rename only, perhaps it shows that way because of deuglification i made on dlnaprofile on a different commit. Not sure what to do here

PR #231, opened 2019-04-06
Improve the plugin catalog and related controllers

dkanada:


src/controllers/addpluginpage.js
```diff
@@ -79,62 +87,69 @@ define(["jQuery", "loading", "libraryMenu", "globalize", "connectionManager", "e
...
     function performInstallation(page, packageName, guid, updateClass, version) {
-        var developer = $("#developer", page).html().toLowerCase(),
-            alertCallback = function(confirmed) {
-                confirmed && (loading.show(), page.querySelector("#btnInstall").disabled = !0, ApiClient.installPlugin(packageName, guid, updateClass, version).then(function() {
-                    loading.hide(), alertText(globalize.translate("PluginInstalledMessage"))
-                }))
-            };
-        if ("luke" != developer && "ebr" != developer) {
+        var developer = $("#developer", page).html().toLowerCase();
+        var alertCallback = function() {
```
thornbill: It doesn't look like this handles the case where a user doesn't confirm installation?

dkanada: I just don't call the function if they don't confirm the installation.

Issue #201, opened 2019-02-17
Remove "Add Server" Option from Login Screen on Public Facing Address


triDcontrols:
**Describe the bug**
Either it was me and did not notice, or latest update changed this. I now see a "Change Server" Option on the login screen when the server is public facing or private internal access. I do use Nginx as a reverse proxy, and not sure if jellyfin is seeing the reverse proxy server and things it's on the local network thus giving this option. What's odd, is it also gives the option to add additional servers from the same page, does not have to be my own jelly fin, I can add any other jellyfin public facing server to the list. 

**To Reproduce**

setup nginx as reverse proxy to jellyfin, and use public facing web address to access jellyfin

**Expected behavior**
Only provide Login Screen, and no "Change Server" Option. 

**Logs**

**Screenshots**

![screen shot 2019-02-16 at 7 41 05 pm](https://user-images.githubusercontent.com/22411662/52908076-0f183680-3223-11e9-93fa-37a668294b77.png)

![screen shot 2019-02-16 at 7 46 37 pm](https://user-images.githubusercontent.com/22411662/52908117-bac18680-3223-11e9-88d2-73f8258e53a5.png)

![screen shot 2019-02-16 at 7 47 13 pm](https://user-images.githubusercontent.com/22411662/52908118-bbf2b380-3223-11e9-8c3b-ec1d072665c5.png)

**System (please complete the following information):**
 - OS: Docker
 - Browser: Firefox, Chrome, Safari
 - Jellyfin Version: Version 10.2.0

**Additional context**

PR #235, opened 2019-04-07
Add check for multiserver on login page

dkanada:


anthonylavado:
Just to help me follow this: is it now optional on whether or not to show the “select server” button?

dkanada:
This is actually related to #201 since this was still showing up on the web frontend. It isn't configurable yet but we should add that at some point.

Issue #233, opened 2019-04-07
Servers with one user will not allow the account to sign out


dkanada:
Even if the sign out button is enabled the frontend won't actually destroy the session, which isn't ideal.

cc @joshuaboniface

dkanada:
this will be fixed by #292

PR #241, opened 2019-04-11
Fix logout session storage

DrPandemic:
When logging out, the user session wouldn't be destroyed on the client. This PR cleans the user session from local storage when logging out.

Fixes https://github.com/jellyfin/jellyfin-web/issues/233

### DIFF
--- a/src/bower_components/apiclient/apiclientcore.js
+++ b/src/bower_components/apiclient/apiclientcore.js
@@ -303,6 +303,7 @@ define(["events", "appStorage"], function(events, appStorage) {
     }, ApiClient.prototype.logout = function() {
         stopBitrateDetection(this), this.closeWebSocket();
         var done = function() {
+            appStorage.removeItem("user-" + this._currentUser.Id + "-" + this._currentUser.ServerId)
             this.setAuthenticationInfo(null, null)
         }.bind(this);
         if (this.accessToken()) {


DrPandemic:
Thanks @dkanada for the review, but I'm not 100% to understand how you see the fix (sorry I'm really unfamiliar with how this codebase is structured).

> It might be better to add a method to appStorage

Why is `appStorage` more appropriated? That file seems to be a generic cache management file. Wouldn't it feel awkward to add a function specific to session/credential management in that file?

Just to be sure I understand your comment. You think the best solution is to do something like that in the `appStorage.js`
```javascript
this.logout = function(userKey) {
  this.removeItem(userKey)
}
```
and call it in the logout function in `apiclientcore.js`?

dkanada:
I wasn't saying this is the incorrect way to fix the issue, I was just thinking that there must be a location in the apiclient where the exact value `"user-" + this._currentUser.Id + "-" + this._currentUser.ServerId` gets added to the cache. If that method exists we might want to add a new method to remove in that specific location.

Also, does this actually add back the logout button in the navbar and user preferences? Those buttons check for [EnableAutoLogin](https://github.com/jellyfin/jellyfin-web/blob/0a5f08ecb06ae86a5ab0b1ff2d2099c1b012f162/src/scripts/librarymenu.js#L204) and I haven't exactly checked what this does, or how that value gets changed.

DrPandemic:
> I was just thinking that there must be a location in the apiclient where the exact value "user-" + this._currentUser.Id + "-" + this._currentUser.ServerId gets added to the cache

Oh I see what you mean. In fact setting the key is not much cleaner. It's done 12 lines higher in the file https://github.com/jellyfin/jellyfin-web/blob/8b5004dda4bd0a62a0f06fa650a76d49e32d9b9a/src/bower_components/apiclient/apiclientcore.js#L294 and it's also concatenating random strings. I could refactor out the string building, but I don't see too much added value there  🤷‍♂️.

> Also, does this actually add back the logout button in the navbar and user preferences? 

I don't think it's related. `EnableAutoLogin` is set in the backend when there's a single user on the server and that user has no password. So if those conditions are met the logout button won't be present.

https://github.com/jellyfin/jellyfin/blob/f97f6b8061a4127d5a18c6a132e546ed424f88c7/Emby.Server.Implementations/Library/UserManager.cs#L619

DrPandemic:
I didn't know it existed :mindblown:, I'll push a PR there.

PR #243, opened 2019-04-13
Changing some layout 

grafixeyehero:
Screenshot_2019-04-13 Jellyfin
Screenshot_2019-04-13 Jellyfin


grafixeyehero:
this screen shot for mediainfo official rathings dark baground & light 

Screenshot_2019-04-13 Jellyfin(2)
Screenshot_2019-04-13 Jellyfin(3)

src/components/cardbuilder/card.css
```diff
@@ -280,7 +281,7 @@ button {
 
 .visualCardBox {
     box-shadow: 0 0.0725em 0.29em 0 rgba(0, 0, 0, 0.37);
-    border-radius: .145em;
+    border-radius: 15px
```
oddstr13: `em` is relative to the size of `m`, `px` is number of pixels.
This will likely look wildly different on a high DPI screen compared to a normal computer screen.

src/components/cardbuilder/card.css
```diff
@@ -173,6 +173,7 @@ button {
     /* This is only needed for scalable cards */
     height: 100%;
     contain: strict;
+    border-radius: 15px
```
oddstr13: `px` vs `em` in context of high DPI.

oddstr13:
That being said, I think I personally prefer square corners on the images.

lostmypillow:
I disagree, I love the rounded corners

Nickbert7:
I like the rounded corners, but would it be possible to add them as an option or as an additional theme?
You might also add the rounded corners to the overlay:

.cardOverlayContainer {
      border-radius: 0.8em;
}
.visualCardBox .cardOverlayContainer {
    border-bottom-right-radius: 0em;
    border-bottom-left-radius: 0em;
}

image vs. image

image vs. image

grafixeyehero:
I agree with the comment you provided to be optional. So i will move it to its on independent themes so the users can have an option to select from

grafixeyehero:
move it to its on independent themes so the users can have an option 
https://github.com/jellyfin/jellyfin-web/pull/251

PR #244, opened 2019-04-13
Fix a deminification issue with the plugin page

dkanada:


joshuaboniface:
Without this, plugin installs just hang since the first conditional doesn't match.

PR #251, opened 2019-04-17
Adding new Purple haze themes with rounded layout

grafixeyehero:
Screenshot_2019-04-18 Jellyfin(3)


src/components/themes/purple-haze/theme.css
```diff
@@ -0,0 +1,476 @@
...
+    color: inherit
+}
+
+.checkboxListLabel,
+.inputLabel,
+.inputLabelUnfocused,
+.paperListLabel,
+.textareaLabelUnfocused {
+    color: #bbb;
+    color: rgba(255, 255, 255, .7)
```
DrPandemic: I guess having 2 `color` is a typo?

grafixeyehero: Not all browsers support RGBa, so if the design permits, you should declare a "fallback" color. This color will be most likely be solid (fully opaque). Not declaring a fallback means no color will be applied in browsers that don't support it. This fallback does fail in some really old browsers.

anthonylavado:
I still really like this, so unless there's objection from @jellyfin/web...

grafixeyehero:
@thornbill 
Screenshot_2019-04-25 Jellyfin

Issue #228, opened 2019-04-04
Click on Layout format goes to movie details and does not change the layout


daves00:
**Describe the bug**

Go into movies, change the layout, this works for some correctly, but when viewing layout (see screenshot j1), and clicking on the layout (circled in red on the screenshot) I am taken to movie details and can no longer change the layout.

**To Reproduce**
1. Click on Home
2. Click on Movies
3. Change Layout, selecting from the bottom of the list
4. Change layout to next one up from last - then repeat going up the list of layouts until issue occurs
5. Change layout and taken to movie details screen - see screen shot j2 

**Expected behavior**
Be able to change between the different layout options

**Logs**  Log pasted but only INF messages, no errors
<details>

```
[2019-04-04 23:58:20.958 +01:00] [INF] HTTP Response 200 to "Client-IP". Time: "3""ms". "http://Server-IP:8096/emby/Users/c004d95e46cf44acb8ffaf31926ae1b9/Views" ""
[2019-04-04 23:58:21.106 +01:00] [INF] "HTTP GET" "http://Server-IP:8096/emby/Users/c004d95e46cf44acb8ffaf31926ae1b9/Views". UserAgent: "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:66.0) Gecko/20100101 Firefox/66.0"
[2019-04-04 23:58:21.109 +01:00] [INF] HTTP Response 200 to "Client-IP". Time: "3""ms". "http://Server-IP:8096/emby/Users/c004d95e46cf44acb8ffaf31926ae1b9/Views" ""
[2019-04-04 23:58:21.132 +01:00] [INF] "HTTP GET" "http://Server-IP:8096/emby/LiveTv/Programs/Recommended?userId=c004d95e46cf44acb8ffaf31926ae1b9&IsAiring=true&limit=1&ImageTypeLimit=1&EnableImageTypes=Primary%2CThumb%2CBackdrop&EnableTotalRecordCount=false&Fields=ChannelInfo%2CPrimaryImageAspectRatio". UserAgent: "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:66.0) Gecko/20100101 Firefox/66.0"
[...]

PR #264, opened 2019-04-22
Fix Thumb rendering without thumbnails

DrPandemic:
**Changes**
The initial problem was occurring when rendering a page in `Thumb` (`options.cardLayout = false`) mode and the media didn't have a thumbnail (`forceName = true`). In this specific situation, it would try to render the title in the `innerCardFooter` rather than in the `outerCardFooter`. Those footers would render the title in a `<button>`, but the `innerCardFooter` was already rendered in a `<button>`. This double `<button>` was breaking the browser's HTML renderers. 

This fix prevents the usage of `innerCardFooter` and fallbacks to `outerCardFooter`.

Before:
image

After (you can see it also fixed the title positioning):
image

**Issues**
Closes https://github.com/jellyfin/jellyfin-web/issues/228


DrPandemic:
I didn't every possible combinations, but for the one I tried, they seemed to be working well.

PR #265, opened 2019-04-22
Update links for most of the old documentation

dkanada:


anthonylavado:
This is missing the corresponding updates in Docs, right?

PR #271, opened 2019-04-25
Remove similar themes with accent changes

dkanada:
I would like to shrink down the theme files but it's really tedious to maintain ten different themes that only change the accent color. This doesn't have to get merged but it would make style changes a lot easier.


anthonylavado:
Sorry for breaking this!

PR #272, opened 2019-04-25
Remove option to enable debug logging

dkanada:
Just wanted to kill some time and I remember someone mentioning this option doesn't actually work. We could also add an option to set the specific logging level or just leave this option in the runtime flags, opened this PR for discussion.


anthonylavado:
@DrPandemic We do actually have a debug log level, but you have to enable it in a file on the server.

anthonylavado:
@dkanada Should we keep a link to the documentation, and add a corresponding page with details on turning on Debug logging?

dkanada:
@anthonylavado I'd like to get some of the current PRs merged before I work on more. It's getting hard to avoid possible merge conflicts with all the changes and I have like nine or ten branches on my local copy of the source.

Issue #198, opened 2019-03-17
Fullscreen on doubleclick


platinum55:
just trigger fullscreen on doubleclick, maybe with an option to disable it, for users who have problems with it.

As a workaround I added this piece of code to videoosd.js around line #1255:

        dom.addEventListener(view, "dblclick", function (e) {
            playbackManager.toggleFullscreen(currentPlayer);
        }, {
            passive: true
        });

This works so far, but sometimes the poster and osd does not hide.. would be cool to have it nicely implemented :)

DrPandemic:
I've been playing around with your snippet and it seems to be working. I'm unable to reproduce your bug where the osd doesn't hide. Which browser are you using? Are you getting this issue on some specific media?

The only thing I would do differently is that I wouldn't bind the event on mobile
```javascript
        if (browser.touch) {
            dom.addEventListener(view, "dblclick", onDoubleClick, {});
        } else {
            dom.addEventListener(view, "dblclick", function () {
                playbackManager.toggleFullscreen(currentPlayer);
            }, {
                passive: true
            });
        }
```

I found only one outside thing that can prevent the osd from hiding and it's the `viewbeforehide` event. But this seems to only be trigger when switching page. A workaround could be to call `startOsdHideTimer()` in the event listener to be sure the timer was not cleared before or during the fullscreen action.

platinum55:
hello mate, I hadn't had the issue again, maybe it disappeared while upgrading from 10.2.2 to 10.3.0 :-) thanks for your enhancement! have a nice day

PR #275, opened 2019-04-25
Add double click to fullscreen

DrPandemic:
**Changes**
Double click to fullscreen.

**Issues**
Closes https://github.com/jellyfin/jellyfin-web/issues/198

### DIFF
--- a/src/controllers/videoosd.js
+++ b/src/controllers/videoosd.js
@@ -368,7 +368,7 @@ define(["playbackManager", "dom", "inputmanager", "datetime", "itemHelper", "med
             if ("osd" === currentVisibleMenu) {
                 hideOsd();
             } else if (!currentVisibleMenu) {
-                    showOsd();
+                showOsd();
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/controllers/videoosd.js
```diff
@@ -1254,6 +1254,12 @@ define(["playbackManager", "dom", "inputmanager", "datetime", "itemHelper", "med
 
         if (browser.touch) {
             dom.addEventListener(view, "dblclick", onDoubleClick, {});
+        } else {
+            dom.addEventListener(view, "dblclick", function () {
```
cvium: This is just a style suggestion that I find easier to read.
```suggestion
            var options = { passive: true };
            dom.addEventListener(view, "dblclick", function () { playbackManager.toggleFullscreen(currentPlayer); }, options);
```

DrPandemic: I like that suggestion, thanks. Tho, I'm pretty sure this function is pure JS without any jQuery

DrPandemic:
I didn't look at the result after accepting the modification on Github... Should be fixed

Issue #246, opened 2019-04-15
Option to disable controller input


PrimordialHelios:
**Is your feature request related to a problem? Please describe.**
I frequently watch videos while gaming, and I occasionally use an xbox360 controller. Unfortunately, the controller input is *always* sent to jellyfin's web client, even when it isn't the focused window. This makes watching videos impossible while playing games with a controller, as each button stops, pauses, fast-forwards, etc.

**Describe the solution you'd like**
A toggle option somewhere in the settings to disable controller input. Alternatively, it would be useful if jellyfin-web just didn't grab any input when it isn't focused.

Reynbow:
Please! This is doing my head in.

anthonylavado:
This is available in the next release :-)

PrimordialHelios:
Fantastic! Thanks for all the great work :)

TheFeelTrain:
I would like this to be revisited. 

This is probably an extremely niche problem but I use an Xbox controller to control my mouse cursor among other things with some custom bindings. During video playback (not during music as far as I can tell) the left thumbstick controls seeking so the player will jump around when I try to move my mouse.

The previous fixes do not account for this use case, so a way to outright disable controller input would be appreciated.

PR #278, opened 2019-04-25
New keyboard shortcuts

DrPandemic:
**Changes**
Tries to follow youtube keyboard shortcuts as closely as possible.
- `k/space` pause/play
- `m` mute/unmute
- `j/l/left/right` forward/backward
- `f` fullscreen

Gamepad events are ignored if the tab is not focused.

This adds the shortcuts on the tooltips, so it's "self-documented".

**Questions**
Do we manage any language that is right to left? For those languages, should we change the translations for something like `(shortcut) foo bar`?

**Issues**
Closes https://github.com/jellyfin/jellyfin/issues/1283
Closes https://github.com/jellyfin/jellyfin-web/issues/246

### DIFF
--- a/src/controllers/videoosd.js
+++ b/src/controllers/videoosd.js
@@ -530,7 +530,7 @@ define(["playbackManager", "dom", "inputmanager", "datetime", "itemHelper", "med
                 view.querySelector(".btnFullscreen").setAttribute("title", globalize.translate("ExitFullscreen"));
                 view.querySelector(".btnFullscreen i").innerHTML = "&#xE5D1;";
             } else {
-                view.querySelector(".btnFullscreen").setAttribute("title", globalize.translate("Fullscreen"));
+                view.querySelector(".btnFullscreen").setAttribute("title", globalize.translate("Fullscreen") + " (f)");
                 view.querySelector(".btnFullscreen i").innerHTML = "&#xE5D0;";
             }
         }
@@ -713,7 +713,14 @@ define(["playbackManager", "dom", "inputmanager", "datetime", "itemHelper", "med
         }
 
         function updatePlayPauseState(isPaused) {
-            view.querySelector(".btnPause i").innerHTML = isPaused ? "&#xE037;" : "&#xE034;";
+            var button = view.querySelector(".btnPause i");
+            if (isPaused) {
+                button.innerHTML = "&#xE037;";
+                button.setAttribute("title", globalize.translate("ButtonPlay") + " (k)");
+            } else {
+                button.innerHTML = "&#xE034;";
+                button.setAttribute("title", globalize.translate("ButtonPause") + " (k)");
+            }
         }
 
         function updatePlayerStateInternal(event, player, state) {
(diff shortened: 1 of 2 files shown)


src/controllers/videoosd.js
```diff
@@ -1013,43 +1020,59 @@ define(["playbackManager", "dom", "inputmanager", "datetime", "itemHelper", "med
...
+                case "j":
+                case "ArrowLeft":
+                case "Left":
+                    playbackManager.rewind(currentPlayer);
+                    showOsd();
+                break;
+
                 case "f":
-                if (!e.ctrlKey) {
+                if (!e.ctrlKey && !e.metaKey) {
```
DrPandemic: Nicer on Mac

DrPandemic:
You are right, adding a menu would be nice. I'll add it.

Tho, I personally also like the updated tooltips. It's how Youtube does it and I think it's an easy way to discover them.

dkanada:
The tooltips can be nice but it means we would have to update strings in every language if we ever change or a key, and they don't even show every key option for a certain action. It also means adding even more strings for translation because we would need values with and without the tooltips. I would at least say make the translation and then add the keybinding at the end in the script, but that wouldn't have very good RTL support.

I am going to point to Netflix as a service that doesn't even have a dialog to show their keybindings :p they only have them listed on a page in their support site. I think the dialog is a fine solution but I will defer to the group opinion.

DrPandemic:
Thanks @dkanada, I'll wait to see if there's more opinions and I'll update the PR accordingly

anthonylavado:
I don't know that we'll add keys very often, or change them for that matter. If Youtube does something, then it's pretty nice to "copy" it, because it's like a de facto standard. I think the tooltips are fine as well, but a dialog would be nice too! Like pressing "?" to bring up a little transparent overlay.

Either way, I like this as it is now.

dkanada:
> add the keybinding in the javascript

How about this so there's less work for translators?

anthonylavado:
@dkanada Good thinking. Was that in chat? I missed it

DrPandemic:
> add the keybinding in the javascript

I like that, I'll try to update this PR tonight

DrPandemic:
I did a quick refactor and removed all the new translations

DrPandemic:
I think we should merge this version and I'll do another PR to add the menu (if it's ok with everyone else). I personally prefer to go for smaller PRs

vitorsemeano:
For me its«s looks pretty neat

PrimordialHelios:
Thanks to all for the great work!

beyondunreal79:
Is there anyway to add a subtitle shortcut key? Especially now since we have a the jellyfin media player?

pyorot:
i would also appreciate a comment on the lack of a shortcut key to toggle subtitles on/off (without otherwise changing the subtitle track)

it’s a fairly important quality-of-life feature for people with all sorts of hearing difficulties i think. i have a noise sensitivity that means i prefer to only have them on when i miss a bit of noisy dialogue

jackrosenberg:
I sometimes watch porn and would like to quickly close if my mom walks in, can we add a close and pause button please sir



?
_

PR #291, opened 2019-04-26
Correct 404 on autoPlayDetect.js

### LINKED ISSUE #267: Unable to load autoPlayDetect.js
**Describe the bug**
Unable to load `/web/components/playback/autoPlayDetect.js`

> from @bugfixin:matrix.org:
OK I think I see what's going on - it appears that the web interface loads autoPlayDetect.js when it can't verify html media autoplay functionality on a site - with chrome, this occurs when visiting jellyfin over non-https - and if you have already visited via https, the browser sets an appsettings key which causes loading of autoPlayDetect.js not to be attempted (this is why i saw it initially on my dev and not my prod)
clearing storage and visiting over non-https i can reproduce this on 10.3.0,10.3.1,and current master
There are two more layers - this loading fails with a 404 because jellyfin's web server is case sensitive, and the URL is autoPlayDetect.js while the filename is autoplaydetect.js

**To Reproduce**

1. Load jellyfin with cache disabled.
2. View log

**Expected behavior**

Error to not occur

**Logs**
```
app is hidden apphost.js:203:13
The resource from “https://jellyfin.domain.com/web/components/playback/autoPlayDetect.js?v=10.3.1” was blocked due to MIME type (“text/html”) mismatch (X-Content-Type-Options: nosniff).[Learn More] index.html
Loading failed for the <script> with source “https://jellyfin.domain.com/web/components/playback/autoPlayDetect.js?v=10.3.1”. index.html:1:1
RequireJS error: Error: Load failed: autoPlayDetect: https://jellyfin.domain.com/web/components/playback/autoPlayDetect.js?v=10.3.1. Failed modules: site.js:389:9
Error: Load failed: autoPlayDetect: https://jellyfin.domain.com/web/components/playback/autoPlayDetect.js?v=10.3.1 alameda.js:284:30
Begin ConnectionManager constructor
```

 - OS: Docker
 - Browser: Firefox
 - Jellyfin Version: 10.3.1

bugfixin:
**Changes**
This modifies a hardcoded reference to autoPlayDetect.js which has been renamed to autoplaydetect.js - because this is case sensitive it results in a 404.

**Issues**
Fixes #267

### DIFF
--- a/src/scripts/site.js
+++ b/src/scripts/site.js
@@ -703,7 +703,7 @@ var AppInfo = {};
             itemHelper: componentsPath + "/itemhelper",
             itemShortcuts: componentsPath + "/shortcuts",
             playQueueManager: componentsPath + "/playback/playqueuemanager",
-            autoPlayDetect: componentsPath + "/playback/autoPlayDetect",
+            autoPlayDetect: componentsPath + "/playback/autoplaydetect",
             nowPlayingHelper: componentsPath + "/playback/nowplayinghelper",
             pluginManager: componentsPath + "/pluginmanager",
             packageManager: componentsPath + "/packagemanager"


bugfixin:
Ah, I was just going for consistency as the majority of the other paths for js files are also lowercased, but I'd be just as happy with renaming the file to camelcase. It may also be worth merging this pull even if it is changed (along with all of the other js paths) to camelcase in the future.

vitorsemeano:
The reason is simple why I change some files to camel case. When the module is defined in site.js, the name is in camel case, pointing to the file on lower case. For webpack to solve automatically this files, you almost just need to change the filename to camel case, matching the definition found in site.js

vitorsemeano:
I don't change every module to camel case because I only change what webpack tells me he can't resolve. So is a case to case change. I will commit the webpack.config.js so you can see the state of webpack conversion.

PR #299, opened 2019-05-03
remove networking information from the dashboard

dkanada:
This is some work towards fixing the ten second delay with the system info endpoint. We can just remove the networking information until someone implements it asynchronously.

I also removed the following strings because I believe they are unused. I check for the dynamic string translations now.

```
ButtonMoreInformation
EnableDebugLoggingHelp
HeaderJellyfinServer
LabelDisplayPluginsFor
LabelEnableDebugLogging
LabelLocalAccessUrl
LabelRemoteAccessUrl
LabelRunningOnPort
LabelRunningOnPorts
LearnHowToCreateSynologyShares
PluginTabAppClassic
```


vitorsemeano:
But that info is not generated when requesting systeminfo through the apiclient? just trying to understand the problem and give you approval. everything else looks fine for me

PR #300, opened 2019-05-03
Fix dashboard column & update purple haze 

grafixeyehero:
Dashboard column **Before**

Screenshot_2019-05-03 Jellyfin(2)

Dashboard column **After** and some update on  **Purple haze color**

Screenshot_2019-05-03 Jellyfin

### DIFF
--- a/src/components/themes/purple-haze/theme.css
+++ b/src/components/themes/purple-haze/theme.css
@@ -1,7 +1,7 @@
 .skinHeader,
 html {
-    color: #ddd;
-    color: rgba(255, 255, 255, .8)
+    color: #f8f8fe;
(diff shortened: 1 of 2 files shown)


dkanada:
What exactly is fixed on the dashboard page?

grafixeyehero:
Look closely on the dashbord before on activity and  alerts cloumn

joshuaboniface:
So it looks like we're just removing the Alerts column, correct? I'm Ok with this, though ideally it would be useful instead.

grafixeyehero:
Not removing scrolldown it's there

grafixeyehero:
@joshuaboniface 
Didn't remove alerts scrolldown u can find on the bottom

PR #304, opened 2019-05-05
Fixed Progressive Web App for Android and iOS

skaro13:
**Changes**
- Fixed manifest.json to allow standalone app start on iOS and Android
- Replaced PWA short name (Emby --> Jellyfin)
- Added files to support launch screens for iOS web app.
- Added a 512px icon png for Android launch screen.

This is an interesting replacement for an iOS app for Jellyfin. Launchcreens are a bit of a pain to implement since they must support every resolution available. If having all these links is problematic, I guess we could do without them but the launch screen will be white.

Before, the home screen link opened the browser and navigated to the jellyfin server:

out

Now, the home screen link appear as a native app:

ezgif-1-9bc331c46c36


src/index.html
```diff
@@ -20,7 +20,46 @@
     <meta property="fb:app_id" content="1618309211750238">
     <link rel="apple-touch-icon" sizes="180x180" href="touchicon.png">
 
-    <!--<link rel="apple-touch-startup-image" href="css/images/iossplash.png">-->
+    <!-------------------- iPhone 5 ------------------->
+    <link href="splashscreens/iphone5_splash.png" media="screen and (device-width: 320px) and (device-height: 568px) and (-webkit-device-pixel-ratio: 2) and (orientation: portrait)" rel="apple-touch-startup-image" />
+    <link href="splashscreens/iphone5_splash_l.png" media="screen and (device-width: 320px) and (device-height: 568px) and (-webkit-device-pixel-ratio: 2) and (orientation: landscape)" rel="apple-touch-startup-image" />
```
dkanada: It looks like the documentation [here](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html) doesn't require several images for different resolutions. We might also be able to just specify one image with a small icon to avoid all these splash screens, or maybe just three or four to use between different resolution screens.

skaro13: It used to be working with a single <link> entry but this is [no longer the case](https://medium.com/appscope/adding-custom-ios-splash-screens-to-your-progressive-web-app-41a9b18bdca3):

> What is not mentioned is that the href attribute has to refer to an image that is of the exact same resolution as the iOS device that uses the app. The problem that arises here is that there are multiple iOS devices with different resolutions, and unfortunately we cannot just simply repeat this code multiple times for images of different sizes. 

It's kind of ugly but this is the only way to do this currently. I'll try to merge them.

dkanada: If that's the case you can probably just leave them. You might be able to shorten the media attribute a bit by declaring the resolutions like `1080x1920` but that might not work with the orientation check.

skaro13: Resolution is not working with orientation indeed. This seems to be [the only working method](https://gist.github.com/EvanBacon/7fd4dc3be3d00096579bb0b134c56ec7) currently... Ii might get better with future releases of iOS. 

Android does not have this problem since the splash screen is made up from the manifest background color and the 512px icon for every device.

joshuaboniface:
It would need to be rebased off the release branch to make the release, but I'd be OK with that.

anthonylavado:
There appear to be no merge conflicts... It's a pretty straightforward change tbh, with files we don't touch often.

skaro13:
Let me know what you prefer. I could always fix things on my side and submit a new pull request with the release branch as source.

DrPandemic:
You don't need a new PR. If you would like to keep this PR you can 
1. Rebase your branch on your repo on `release-10.3.z`
2. Force push
3. Change the target branch at the top of this page (you first need to click on edit)

skaro13:
Now on release-10.3.z

skug67:
Doesn't seem to pick up when jellyfin is being served at suburl [https://host.domain.com/emby].

skug67:
Dug around a bit and you can get the suburl working by modifying the "start_url" entry in manifest.json from "/web/index.html" to "/emby/web/index.html"  Given that the system is currently listening at /emby as well as the root url, is there any reason not to change this entry so that it works for both default and non-default users?  (In the long run we'll hopefully get the ability to configure the listening url on the dashboard, at which point we'll need to link up that setting with the start_url entry in manifest.json -- not that I know how that could be accomplished.)

PR #307, opened 2019-05-05
Redirect when loading video player fails

DrPandemic:
**Changes**
When reloading the video player, it gets stuck in a weird broken state.
image

I'm not 100% sure it's the right solution to the problem, but this PR force a redirect to the home page when loading the player fails. We can't really rescue by trying to reload the right video, because AFAIK we don't know what's the video after a page reload.

I personally prefer to go back to a good state (home page) than staying on a broken page.

### DIFF
--- a/src/controllers/videoosd.js
+++ b/src/controllers/videoosd.js
@@ -1227,16 +1227,22 @@ define(["playbackManager", "dom", "inputmanager", "datetime", "itemHelper", "med
             Emby.Page.setTransparency("full");
         });
         view.addEventListener("viewshow", function (e) {
-            events.on(playbackManager, "playerchange", onPlayerChange);
-            bindToPlayer(playbackManager.getCurrentPlayer());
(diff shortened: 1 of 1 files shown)


EraYaN:
Can we not simulate the "back button" action? history.go(-1) might do it on most platforms.

joshuaboniface:
@EraYaN I've found when dealing with this particular issue that a simple "back" leaves it broken, because of the weird frame-like handling of pages. So a single `history.go(-1)` might not actually be sufficient. I'm more a fan of the explicit "go to home" idea.

DrPandemic:
@EraYaN I'll be honest, I didn't even think about it. If `go(-1)` were to work, I would prefer that solution

dkanada:
#152 courtesy of @redSpoutnik

DrPandemic:
I just tried with `Emby.Page.back();` and this breaks the app

PR #309, opened 2019-05-05
Various fixes for webOS 3.3

### LINKED ISSUE #306: RequireJS regression in old Chromium versions (webOS) [10.2.2 -> 10.3.0]
**Describe the bug**
`RequireJS error: TypeError: undefined is not a function. Failed modules: ` at `site.js?v=10.3.2:389`

This did not occur in the 10.2.2 version

**To Reproduce**
Try to load the web UI on an LG webOS tv v3 (and probably earlier)

So Chromium 38 and WebKit 538.2: http://webostv.developer.lge.com/discover/specifications/web-engine/

**Expected behavior**

**Logs**
`RequireJS error: TypeError: undefined is not a function. Failed modules: ` at `site.js?v=10.3.2:389`

Is the first (and only) line.

 - OS: webOS v3.x
 - Browser: WebAppManager (the "native" apps)
 - Jellyfin Version: 10.3.x

**Additional context**
Not the browser app, although that also does not work.

oddstr13:
**Changes**
  * Add back `Object.assign` and `fetch` polyfills.
  * remove occurances of `Promise.finally()`.
  * move functions out of if statements (strict mode)
  * make `delayedError` spit out a stack trace.

**Issues**
Fixes #306 (for webOS 3.3, not tested with webOS 2.0)

### DIFF
--- a/src/bower_components/alameda/alameda.js
+++ b/src/bower_components/alameda/alameda.js
@@ -239,6 +239,7 @@ var requirejs, require, define;
         }
 
         function delayedError(e) {
+            console.log(e.stack);
             return setTimeout(function() {
                 e.dynaId && trackedErrors[e.dynaId] || (trackedErrors[e.dynaId] = !0, req.onError(e))
             }), e
--- a/src/bower_components/alameda/alameda.js
+++ b/src/bower_components/alameda/alameda.js
@@ -266,7 +266,7 @@ var requirejs, require, define;
             trackedErrors = obj(),
             urlFetched = obj(),
             bundlesMap = obj(),
-            asyncResolve = Promise.resolve();
+            asyncResolve = Promise.resolve(undefined);
         return load = "function" == typeof importScripts ? function(map) {
             var url = map.url;
             urlFetched[url] || (urlFetched[url] = !0, getDefer(map.id), importScripts(url), takeQueue(map.id))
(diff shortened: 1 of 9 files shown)


src/bower_components/alameda/alameda.js
```diff
@@ -239,6 +239,7 @@ var requirejs, require, define;
         }
 
         function delayedError(e) {
+            console.log(e.stack);
```
thornbill: I'm guessing this was just added for debugging? Should probably remove it so we aren't modifying 3rd party libs.

src/components/appRouter.js
```diff
@@ -343,7 +342,6 @@ define(['loading', 'globalize', 'events', 'viewManager', 'layoutManager', 'skinM
                 hashbang: options.hashbang !== false,
                 enableHistory: enableHistory()
             });
-        }).finally(function () {
```
thornbill: This may end up in the loading indicator not being hidden if an error occurs. We may want to do something like `.catch().then(function() { ... })` to emulate the behavior of `.finally()`.

src/controllers/loginpage.js
```diff
@@ -156,8 +156,6 @@ define(["apphost", "appSettings", "dom", "connectionManager", "loading", "cardSt
                     view.querySelector("#txtManualName").value = "";
                     showManualForm(view, false, false);
                 }
-
-            }).finally(function () {
```
thornbill: Same as above.

src/bower_components/alameda/alameda.js
oddstr13: I left this one in intentionally, as without it you don't get the source of errors, just a cryptic line of text (see #306 for example)

src/components/appRouter.js
oddstr13: Wasn't entirely sure of how to handle the lack of `.finally()`

PR #312, opened 2019-05-07
Always show the start task button except on errors

dkanada:
The start button was never getting shown.


src/controllers/scheduledtaskspage.js
```diff
@@ -106,15 +106,15 @@ define(["jQuery", "loading", "events", "globalize", "serverNotifications", "huma
         if (state === "Running") {
             elem.classList.remove("btnStartTask");
             elem.classList.add("btnStopTask");
-            elem.classList.remove("hide");
```
thornbill: Should we still remove the `hide` class in the "Running" and "Idle" states? It looks like we could end up with cases where the element is hidden and then not unhidden.

dkanada: I just removed the hide stuff entirely, now it will always be shown.

PR #313, opened 2019-05-07
Improve the library page and dialog UX

dkanada:
The main change was standardizing the submit button on the create and edit dialogs, which was missing on the edit dialog previously. I also removed a useless dialog that just told users to remove and recreate their library to change the content type. I didn't get all the deminification done for these pages but I am done for now, it was taking ages.


dkanada:
I use a mix of both, but I probably do it by hand more often because it's an easy way to get familiar with the code.

joshuaboniface:
Approved but we have CLI-only merge conflicts from previous PRs @dkanada

dkanada:
@PrplHaz4 are you seeing an issue? I didn't change the behavior for that input, only when saving changes to an existing library.

PrplHaz4:
@dkanada I haven't tested this PR - was just curious what happened now when you would have normally seen this dialog (only because I recently tried to change the media type of a library and was annoyed by this dialog)

dkanada:
OH! That dialog. Yeah it was annoying when I came across it as well. There isn't a simple fix to that since it would require backend changes so I just removed it for now. I would like to move the rename dialog into the edit page at some point but that's for another time.

Issue #303, opened 2019-05-04
Add keyboard/remote navigation


EraYaN:
**Describe the feature you'd like**
To use the webUI as a hosted app on LG webOS 4 TV's (browser won't do streaming) it would be nice to be able to use a keyboard or preferably the remote to control the UI. Right now the magic remote works (since it acts as a mouse). And the main menu does show a highlight around the first menu, it just does not move with arrow keys.

**Additional context**
Both tested using the magic remote and one of the "normal" remotes, and a keyboard.

**Key codes**
The keypress event acts like [Chrome and IE](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) and below are the keyup and keydown codes. So for controls keyup and keydown are the correct events.

*Common keys*

| Key      | Key code | Suggested Action/Notes                                     |
|----------|----------|------------------------------------------------------------|
| 0-9      | 48-57    | Channel switching and number entry (same as keyboard)      |
| OK       | 13       | Enter/OK/Show OSD                                          |
| TEXT     | 459      | Cycle through subtitles (Mimic Kodi behaviour for key `L`) |
| T.OPT    | 1004     | Show subtitle options                                      |
| Ch. Up   | 33       | Next Channel/Next Item in playlist                         |
| Ch. Down | 34       | Previous Channel/PreviousItem in playlist                  |
| Red      | 403      | Remappable action 1                                        |
| Green    | 404      | Remappable action 2                                        |
| Yellow   | 405      | Remappable action 3                                        |
| Blue     | 406      | Remappable action 4                                        |
| Left     | 37       | (same as keyboard)                                         |
| Up       | 38       | (same as keyboard)                                         |
| Right    | 39       | (same as keyboard)                                         |
| Down     | 40       | (same as keyboard)                                         |
	
*Magic Remote keys*

| Key              | Key code | Suggested Action           |
|------------------|----------|----------------------------|
| RemoteTypeButton | 1062     | Toggle Mouse/Keyboard mode |
| cursor on        | 1536     | Switch to Mouse mode       |
| cursor off       | 1537     | Switch to Keyboard mode    |

*Normal Remote keys*

| Key      | Key code | Suggested Action                                                    |
|----------|----------|---------------------------------------------------------------------|
| Subtitle | 460      | Toggle Subtitle on/off (Mimic Kodi key `T`)                         |
| Info     | 457      | Show info                                                           |
[...]

PR #314, opened 2019-05-08
Initial support for keyboard navigation (for webOS remote etc.)

oddstr13:
Adds handling of keyboard events in order to support TV remotes such as in webOS.
Currently only handles arrow keys (up, down, left, right) and this is enough for basic navigation and watching movies.
It currently *partially* implements #303.

### There are still a few issues that should probably be fixed before merging: ###
- [ ] Users on login screen do not highlight when selected
- [ ] Radio buttons do not highlight when they are not checked
- [ ] Item listings need focus move on load, similar to what I've done for the `Resume` and `Play` buttons on item details.
- [ ] Navigating (focusing) over tabs load their content, potentially making it awkward to navigate to the user menu in the upper right corner.
- [ ] Moving focus to items (partially) off screen does not scroll/change the view in webOS 3
 *Potentially related to the fact that scrolling on any page moves only the top bar - but selecting items off the screen does not move the bar.*
  Seems to be working in webOS 4

Some of those issues are less severe, and might not belong in this PR rather than a general UI cleanup?
#### Scrolling not working in webOS 3 ####
Scrolling moves only the top bar

PS: This pr also includes the commits from #309, as they are a requirement for even loading the web app to test this - I'm assuming the related diffs will disappear from this view once that one is merged)

### DIFF
--- a/src/components/keyboardnavigation.js
+++ b/src/components/keyboardnavigation.js
@@ -1,0 +1,36 @@
+define(['inputmanager', 'focusManager'], function(inputmanager, focusManager) {
+    'use strict';
+
+    console.log("keyboardnavigation");
+
+    function enable() {
+        document.addEventListener('keydown', function(e) {
+            var capture = true;
+
+            switch (e.keyCode) {
+            case 37: // ArrowLeft
+                inputmanager.handle('left');
+                break;
+            case 38: // ArrowUp
+                inputmanager.handle('up');
+                break;
+            case 39: // ArrowRight
+                inputmanager.handle('right');
+                break;
+            case 40: // ArrowDown
+                inputmanager.handle('down');
+                break;
+            default:
+                capture = false;
+            }
+            if (capture) {
+                console.log("Disabling default event handling");
+                e.preventDefault();
+            }
+        });
+
+    }
+    return {
+        enable: enable,
+    };
+});
--- a/src/controllers/itemdetailpage.js
+++ b/src/controllers/itemdetailpage.js
@@ -322,6 +322,25 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "cardBuild
             itemBirthLocation.classList.remove("hide"), itemBirthLocation.innerHTML = globalize.translate("BirthPlaceValue").replace("{0}", gmap)
         } else itemBirthLocation.classList.add("hide");
         setPeopleHeader(page, item), loading.hide()
+
+        try {
+            require(["focusManager"], function(focusManager) {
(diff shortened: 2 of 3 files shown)


joshuaboniface:
>PS: This pr also includes the commits from #309, as they are a requirement for even loading the web app to test this - I'm assuming the related diffs will disappear from this view once that one is merged)

Looks like they do duplicate, but that should be fine unless you want to go in and remove them.

oddstr13:
> Looks like they do duplicate, but that should be fine unless you want to go in and remove them.
Yea, that's because it was merged in, not pulled in I think -- a rebase should fix that.

oddstr13:
Rebased so that diffs are proper

DrPandemic:
You removed the `[WIP]`, do you want reviews now or do you prefer to solve all the small issues in this PR before?

oddstr13:
I removed WIP in the title in favor of the issue tag, as I have access to those now.

I'm always open to comments however, so review if you want to.

I do feel like some of the issues should probably be fixed before merging, but it would also be preferable to merge before next release as this makes the webOS app minimally functioning for webOS 3 and up.

On May 11, 2019 5:57:49 PM UTC, Jean-Samuel Aubry-Guzzi <notifications@github.com> wrote:
>You removed the `[WIP]`, do you want reviews now or do you prefer to
>solve all the small issues in this PR before?
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/jellyfin/jellyfin-web/pull/314#issuecomment-491531820

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

DrPandemic:
> I removed WIP in the title in favor of the issue tag, as I have access to those now.

That make sense :sweat_smile:. I'll wait for you to be done before testing it.

oddstr13:
The lack of scroll seems to be related to `window.scroll(0, 100)` doing nothing (`window.scrollY` is `0` after).
This is true for webOS emulator versions 1.2, 2.0 and 3.0. It is also true for my webOS 3.3 TV.

Scroll wheel that moves the top bar only *does* change the `window.scrollY` for some reason, and I have no idea how or why.

It is not exclusive to the Jellyfin app.
Test case: https://gist.github.com/oddstr13/419aec64e4087d125cbf49f838de910e

I have not been able to figure out why yet - any input on this would be apreciated.

EraYaN:
Can we leverage any native scrolling instead of javascript scrolling on the TVs? Like maybe using the actual `focus` functions? Some browsers (on console) use this to scroll pages, as in scroll things into view.

Issue #302, opened 2019-05-04
Minor UI Issues


T0W1:
the scroll button overlaps with the 3dots. #142 #144 
![Screenshot from 2019-04-29 21-59-22](https://user-images.githubusercontent.com/50248556/57165021-3b8d2900-6df6-11e9-8dd1-871347c30bfa.png)

also this, there is nothing there.
![Screenshot at 2019-05-03 22:51:06](https://user-images.githubusercontent.com/50248556/57165073-58c1f780-6df6-11e9-8939-7c126367efb1.png)

Alphabet list overlaps Dashboard icon when **not** fullscreen.
![Fullscreen-Screenshot from 2019-04-29 23-16-59](https://user-images.githubusercontent.com/50248556/57164054-38dd0480-6df3-11e9-9f14-8e0b167c8eb7.png)
![Screenshot at 2019-04-29 23:15:54](https://user-images.githubusercontent.com/50248556/57164057-39759b00-6df3-11e9-9363-a2eace088984.png)

The Movie and TV Show posters are inconsistent when it comes to showing YEAR.
![Screenshot at 2019-05-04 13:11:08](https://user-images.githubusercontent.com/50248556/57178307-69ae4f80-6e6e-11e9-9ded-ed4dfb8f286e.png)
![Screenshot from 2019-05-04 13-20-35](https://user-images.githubusercontent.com/50248556/57178403-b181a680-6e6f-11e9-8ff0-c68408278fde.png)

anthonylavado:
Just to be clear, is it the display of the years that is the inconsistent part you mention, or the fact that only some shows (like The Grand Tour) show a year in the Poster view?

IIRC there are multiple shows called “The Grand Tour”, so the year must be used to help mark which one you’re looking at.

T0W1:
The first one, I know why the grand tour shows it. However some may even consider that to be inconsistent for the fact that not all shows do this and can even be considered to be redundant because the years can be shown below the name.

T0W1:
@DrPandemic  @dkanada #315
For the left and right scroll buttons maybe change it to the one used in HOME VIDEOS AND PICTURES
as has a smaller area and is unlikely to interfere. (it also look cleaner to me)
![Screenshot from 2019-05-15 20-22-10](https://user-images.githubusercontent.com/50248556/57800254-3bbfe980-7751-11e9-908d-253e947ccf38.png)

T0W1:
[...]

PR #315, opened 2019-05-10
Move buttons to top right for custom scroll element

dkanada:
The only huge issue is that I haven't figured out how to hide the scroll buttons if the elements don't actually go past the page width. There might also be some style kinks to work out.

This fixes the overlap issue several people mentioned since the buttons no longer show up in the scroller at all. The scroll buttons also no longer break and let you navigate past the last element either.

@dinki can you test these changes with the TV guide element I mentioned a bit ago and let me know if this breaks anything?

### DIFF
--- a/src/components/emby-scrollbuttons/emby-scrollbuttons.css
+++ b/src/components/emby-scrollbuttons/emby-scrollbuttons.css
@@ -1,60 +1,20 @@
-.emby-scrollbuttons-scroller {
+.emby-scroller-container {
     position: relative;
 }
 
-.scrollbuttoncontainer {
+.emby-scroller {
+    margin-left: 3.3%;
+    margin-right: 3.3%;
+}
+
+.emby-scrollbuttons {
     position: absolute;
-    top: 10%;
-    bottom: 35%;
+    top: 0;
+    right: 0;
     align-items: center;
     justify-content: center;
     z-index: 1;
-    font-size: 3em;
-    color: #fff;
-    display: none;
-    overflow: hidden;
-}
-
-.scrollbuttoncontainer-left {
-    background: rgba(20, 20, 20, .5);
-    background: -moz-linear-gradient(left,#000 0,rgba(0,0,0,0) 100%);
-    background: -webkit-linear-gradient(left,#000 0,rgba(0,0,0,0) 100%);
-    background: linear-gradient(to right,#000,rgba(0,0,0,0));
-}
-
-.scrollbuttoncontainer-right {
-    background: rgba(20, 20, 20, .5);
-    background: -moz-linear-gradient(right,#000 0,rgba(0,0,0,0) 100%);
-    background: -webkit-linear-gradient(right,#000 0,rgba(0,0,0,0) 100%);
-    background: linear-gradient(to left,#000,rgba(0,0,0,0));
-}
-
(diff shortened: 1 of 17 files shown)


dkanada:
Oh yeah, this fixes the overlap issue several people mentioned since the buttons no longer show up in the scroller at all. The scroll buttons also no longer break and let you navigate past the last element either.

dinki:
I'm not sure when I'll have time to get to this, but you should be able to use the 'fake live tv' files I made to test it. You can also replicate the single channel in the M3U file multiple times, give the copies different names, and assign all the same guide data. That should populate your guide enough to test. I can certainly test but it might not be until mid next week.

dinki:
[livetvtestingfiles.zip](https://github.com/jellyfin/jellyfin-web/files/3166552/livetvtestingfiles.zip)

src/components/emby-scrollbuttons/emby-scrollbuttons.css
```diff
@@ -1,60 +1,20 @@
...
-.emby-scrollbuttons-scrollbutton:after {
-    content: '';
-    display: none !important;
-}
-
-.emby-scrollbuttons-scrollbutton:focus {
-    color: inherit !important;
-}
+    overflow: hidden;
+}
```
DrPandemic: No biggie, but you could set your editor to fix that.

DrPandemic:
This should fix the first element of https://github.com/jellyfin/jellyfin-web/issues/302

dkanada:
The thing that bothers you is what I meant above with hiding the scroll buttons, I have no clue how to easily fix that yet. This is the one of several UI changes I wanted to pull over from Plex with the goal of making Jellyfin more polished.

src/components/emby-scrollbuttons/emby-scrollbuttons.css
dkanada: I added an editor config so all files should get standardized eventually.

dkanada:
That file is a dependency though, so we might want to remove it and use the upstream version if we can. What exactly was the issue your change fixed?

Also, what browser are you using? At the moment, only chromium browsers properly hide the buttons when the scroller is too short. Each scroller also needs the afterRefresh method linked as seen [here](https://github.com/jellyfin/jellyfin-web/pull/315/files#diff-182d1e7d3570466ecc775a05b555d587R469) to render properly on load.

DrPandemic:
> so we might want to remove it and use the upstream version if we can

Not sure to understand what this means.

Ohh I was testing on firefox. I just tried on chrome and it totally break it. Oh that's interesting, with my "fix" FF hides arrows when they are not needed, but not on chrome. But without it, it's the inverse.

So, what I'm saying is that the scroller is not hidden on FF with your current PR with the list is too short.

oddstr13:
Reddit post mentioning the arrow problem https://www.reddit.com/r/jellyfin/comments/c2tp5m/very_tiny_bug/

dkanada:
There are still a few underlying problems but this is the easiest solution I found that didn't require sweeping changes to the code.

dkanada:
@DrPandemic without the latest commit the negative margin was broken in the android app and I have no clue why. It added a horizontal scrollbar for no reason.

Issue #263, opened 2019-04-16
Cannot Skip or Continue a video


fromage9747:
I tried compiling this from source today but all it ever gave me was a black screen.

I then found [this forum entry](https://forum.jellyfin.org/t/temporary-testing/55) which directed me to some beta apk's.

I am able to load the app and run it and start any video which works not as great as the web client though.  However whenever trying to skip through or continue where I left off the progress bar at the bottom skips to the beginning, starting the counter from the start and the screen is black.

I've tried with various videos and they are provide the same result. 

I am using a Samsung Note8 running Android pie.

fromage9747:
@retro-78 It happens on MKV's as well.  pretty much most of my library...  But yes, the web client works well except for going full screen in chrome on the phone.  Due to having a Samsung note8 with a software navigation bar, I will have a bright white navigation bar starring at me whilst watching something in full screen. 

If not for the inability to skip through videos, I would ditch Plex.

fromage9747:
using build 0.9.0 and still experiencing the same issue.

PR #323, opened 2019-05-16
Resolve some android seeking issues

bugfixin:
I believe this allows android to seek, but I think there's more to it than this quick fix - the cordova app seems to be getting transcodes / remuxes when it is not fully necessary which makes this bug worse.

**Changes**
Prevents a removal of the startPositionTicks variable so that the progress bar is not reset to 0:00 when seeking

**Issues**
#263
possibly https://github.com/jellyfin/jellyfin-android/issues/133

### DIFF
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -2536,7 +2536,6 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
 
                     } else {
 
-                        playerStartPositionTicks = null;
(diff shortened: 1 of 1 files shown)


fromage9747:
Where can I get the apk for this tracking issue fix? I am using the 0.9.0 release and still experience the inability to skip or resume a video.

anthonylavado:
@fromage9747 The last upload of 0.9.0 used an old build of certain modules for some reason. We will be making a new one soon.

ckcr4lyf:
So essentially, the new apk build will have the seeking issue fixed? Because for me as well, this is the biggest drawback right now.

dag420:
plz seeking is important to me plz fix

PR #324, opened 2019-05-16
Change warning icons on dashboard to a bell instead of "pvr"

Nickbert7:
new:
image

old:
image

### DIFF
--- a/src/components/activitylog.js
+++ b/src/components/activitylog.js
@@ -4,13 +4,11 @@ define(["events", "globalize", "dom", "datetime", "userSettings", "serverNotific
     function getEntryHtml(entry, apiClient) {
         var html = "";
         html += '<div class="listItem listItem-border">';
+        var color = "#00a4dc";
+        var icon =  "notifications";
         if ("Error" == entry.Severity || "Fatal" == entry.Severity || "Warn" == entry.Severity) {
-            var color = "#cc0000";
-            var icon = "notification_important";
-        }
-        else {
-            var color = "#00a4dc";
-            var icon =  "notifications";
+            color = "#cc0000";
+            icon = "notification_important";
         }
         if (entry.UserId && entry.UserPrimaryImageTag) {
             html += '<i class="listItemIcon md-icon" style="width:2em!important;height:2em!important;padding:0;color:transparent;background-color:' + color + ";background-image:url('" + apiClient.getUserImageUrl(entry.UserId, {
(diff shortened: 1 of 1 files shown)


src/components/activitylog.js
```diff
@@ -5,13 +5,14 @@ define(["events", "globalize", "dom", "datetime", "userSettings", "serverNotific
         var html = "";
         html += '<div class="listItem listItem-border">';
         var color = "Error" == entry.Severity || "Fatal" == entry.Severity || "Warn" == entry.Severity ? "#cc0000" : "#00a4dc";
+        var icon = "Error" == entry.Severity || "Fatal" == entry.Severity || "Warn" == entry.Severity ? "notification_important" : "notifications";
```
thornbill: Since these two lines use the same logic, it may be cleaner to use an `if` statement rather than duplicating it.

Nickbert7: That's right, hopefully better now

src/components/activitylog.js
```diff
@@ -4,14 +4,21 @@ define(["events", "globalize", "dom", "datetime", "userSettings", "serverNotific
...
+            var icon =  "notifications";
+        }
         if (entry.UserId && entry.UserPrimaryImageTag) {
             html += '<i class="listItemIcon md-icon" style="width:2em!important;height:2em!important;padding:0;color:transparent;background-color:' + color + ";background-image:url('" + apiClient.getUserImageUrl(entry.UserId, {
                 type: "Primary",
                 tag: entry.UserPrimaryImageTag,
                 height: 40
             }) + "');background-repeat:no-repeat;background-position:center center;background-size: cover;\">dvr</i>"
-        } else html += '<i class="listItemIcon md-icon" style="background-color:' + color + '">dvr</i>';
+        } else html += '<i class="listItemIcon md-icon" style="background-color:' + color + '">' + icon + '</i>';
```
thornbill: To improve readability here I would recommend adding braces here. 
```
} else {
    html += ...
}
```

src/components/activitylog.js
```diff
@@ -4,14 +4,21 @@ define(["events", "globalize", "dom", "datetime", "userSettings", "serverNotific
     function getEntryHtml(entry, apiClient) {
         var html = "";
         html += '<div class="listItem listItem-border">';
-        var color = "Error" == entry.Severity || "Fatal" == entry.Severity || "Warn" == entry.Severity ? "#cc0000" : "#00a4dc";
+        if ("Error" == entry.Severity || "Fatal" == entry.Severity || "Warn" == entry.Severity) {
+            var color = "#cc0000";
+            var icon = "notification_important";
```
thornbill: While I believe this will function as expected, it’s generally better to define variables outside of blocks like this to improve readability.

src/components/activitylog.js
```diff
@@ -4,14 +4,21 @@ define(["events", "globalize", "dom", "datetime", "userSettings", "serverNotific
...
+        }
+        else {
```
thornbill: I would combine this with the line above.

Nickbert7: Thank you for the feedback
Those comments are the best way to improve my JS :)

Issue #308, opened 2019-05-05
Library icons are different in different places


stephanrenggli:
**Is your feature request related to a problem? Please describe.**
The library icons (movies, tv shows, music, etc.) are not the same as in library settings.

Sidebar:
![image](https://user-images.githubusercontent.com/5135902/57200134-f4cf3880-6f87-11e9-918f-d4d051ed6d4e.png)

Library Settings:
![image](https://user-images.githubusercontent.com/5135902/57200128-dec17800-6f87-11e9-8c2f-f42967a82325.png)

**Describe the solution you'd like**
The same icons should be used everywhere.

**Describe alternatives you've considered**

**Additional context**
Low priority as it doesn't have any impact on jellyfin features.

dkanada:
fixed by #326

PR #326, opened 2019-05-16
Fix icon 

grafixeyehero:
Fix Icon
Library icons are different in different places #308
Before

home page
Screenshot_2019-05-17 Jellyfin(8)
Screenshot_2019-05-17 Jellyfin(6)

nav menu
Screenshot_2019-05-17 Jellyfin(7)

dashbord
Screenshot_2019-05-17 Jellyfin(5)

After
home page
Screenshot_2019-05-17 Jellyfin(4)
Screenshot_2019-05-17 Jellyfin

nav menu
Screenshot_2019-05-17 Jellyfin(1)

dashborad
Screenshot_2019-05-17 Jellyfin(3)

### DIFF
--- a/src/components/cardbuilder/cardBuilder.js
+++ b/src/components/cardbuilder/cardBuilder.js
@@ -1572,7 +1572,7 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
                 return '<i class="cardImageIcon md-icon">&#xe333;</i>';
             }
             if (collectionType === 'homevideos' || collectionType === 'photos') {
-                return '<i class="cardImageIcon md-icon">&#xe071;</i>';
+                return '<i class="cardImageIcon md-icon">&#xe413;</i>';
             }
             if (collectionType === 'music') {
                 return '<i class="cardImageIcon md-icon">&#xE310;</i>';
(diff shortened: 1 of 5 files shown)


src/components/cardbuilder/cardBuilder.js
```diff
@@ -1565,12 +1565,24 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
...
             }
+            if (collectionType === 'movies') {
+                return '<i class="cardImageIcon md-icon">&#xe04a;</i>';
+            }
+            if (collectionType === 'tvshows') {
+                return '<i class="cardImageIcon md-icon">&#xe333;</i>';
+            }
             if (collectionType === 'homevideos' || collectionType === 'photos') {
-                return '<i class="cardImageIcon md-icon">&#xE412;</i>';
+                return '<i class="cardImageIcon md-icon">&#xe071;</i>';
```
DrPandemic: I'm not the biggest fan of choosing this icon. I would go for `e413` `photo library` 
image. It would match with `src/scripts/librarymenu.js`. 

Tho, it might clash with `homevideos`'s icon since both seems to always be linked... If both are always linked, I would choose `photo library` since IMHO this library type is more representative of pictures.

But honestly, I don't have strong opinions here. So both are good with me.

grafixeyehero: Thanks as u suggested i updated homevideos and photo

grafixeyehero:
@dkanada i agree with you, i think for the time being it is enough to use the pull request. Eventhoug i can't do it at the moment but it's possible to adjust it as per your suggestion in the future.

Issue #114, opened 2019-02-02
The repeat button on nowplaying.html is highlighted improperly


sparky8251:
If you have a music library and are on the now playing page, there is a repeat button.

If you click it once, it turns blue and gets a blue "bubble" around it and turns on repeat, again and the number 1 appears in the middle of the icon. 

3rd click you would expect the repeat button to return to white to indicate the option is off. This does not happen until you click elsewhere in the UI.

DrPandemic:
I actually wrote a commit to "fix" that some weeks ago, but never submitted it. After checking other sites (like gmail and anchor.fm) I realized that they were doing the same thing.

![image](https://user-images.githubusercontent.com/3250155/57576158-b1f9de80-7427-11e9-976d-c54b096e0c3d.png)
![image](https://user-images.githubusercontent.com/3250155/57576222-6562d300-7428-11e9-8d65-35f1a89c6b49.png)


Pretty much all the icons kept the bubble.

I think it's done like that for people that navigate with the keyboard (eg. blind people). I agree that it's annoying, but I'm not sure what's the right approach here.

dkanada:
This issue is probably resolved since we removed the annoying hover background. The new behavior for paper icons (the offending css class) is a blue icon for focus and a circular background on hover. @DrPandemic also made some excellent changes to improve the checkbox focus behavior.

PR #327, opened 2019-05-17
Changes checkboxes focus appearance

DrPandemic:
Follow up on https://github.com/jellyfin/jellyfin-web/pull/318. The goal is to improve the focus styling.

The current focus styling for checkboxes is really aggressive IMHO.
image

This PR changes checkboxes focus styling.

image
image

In order:
- Checked, no focus
- Unchecked, no focus
- Unchecked, focus
- Checked, focus

Kinda related to https://github.com/jellyfin/jellyfin-web/issues/114

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -22,6 +22,7 @@
  - [LogicalPhallacy](https://github.com/LogicalPhallacy)
  - [thornbill](https://github.com/thornbill)
  - [redSpoutnik](https://github.com/redSpoutnik)
+ - [DrPandemic](https://github.com/drpandemic)
 
 # Emby Contributors
 
--- a/src/components/emby-checkbox/emby-checkbox.css
+++ b/src/components/emby-checkbox/emby-checkbox.css
@@ -63,29 +63,6 @@
     justify-content: center;
 }
 
-/* Commenting this out - set by theme */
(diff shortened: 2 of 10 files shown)


dkanada:
@DrPandemic any chance you could make this a bit more generic and also add a similar style to the text input? I would say select as well but that one might need an overhaul.

DrPandemic:
I'll take a look to see how it's done right now. I'll keep you informed

DrPandemic:
@dkanada I took a look, and I'm not sure what you have in mind.

Currently, text inputs look like
image

and select 
image

which I think look similar to the new checkboxes style. How would you see them?

dkanada:
The checkboxes have a nice thick border that looks great, but the input and select elements have a paper thin border that looks like an afterthought. The text above the elements is also highlighted, which is not the case for checkboxes. Another minor issue is that the drop down for select isn't flush with the main select element, which keeps its curved border. I think that would require code to fix though.

DrPandemic:
I'm not sure if I prefer the label highlighted or not. It feels uglier, but it's easier to see. Which one do you prefer?

dkanada:
I like it better not highlighted because it's not really the element that has focus. The text for settings is always above the element so I don't think we really need to highlight it as well.

PR #334, opened 2019-05-21
Add getPlugins from NativeShell API

vitorsemeano:
This PR adds the method getPlugins through the NativeShell API, to the loading of plugins on start. This should allow external plugins like video, audio, between others, to be loaded dynamically if this is to be used in other environments, like android.

**Changes**
Adds method getPlugins to the loadPlugins method present in site.js


src/scripts/site.js
```diff
@@ -546,8 +546,8 @@ var AppInfo = {};
             }
         }
 
-        for (var index = 0, length = externalPlugins.length; index < length; index++) {
-            list.push(externalPlugins[index]);
+        if (window.NativeShell) {
+            list = list.concat(window.NativeShell.getPlugins());
```
dkanada: Don't we want to keep the web players just in case?

vitorsemeano: And they are kept in place. I just removed an array that is never filled with plugins. The list array contains all current plugins, including the web ones.

dkanada:
This looks good but can you run the commands below to avoid the merge commits on the master branch? Merge commits cause a lot of issues when trying to revert previous changes to the codebase.

```
git checkout exoplayer_changes
git reset HEAD~1
git stash
git reset --hard HEAD~4
git stash pop
git add .
git commit -m "add getPlugins method"
```

dkanada:
merged in #360

PR #337, opened 2019-05-24
Correction and translation of texts or words in spanish

97carmine:
**Changes**
I've corrected the spelling of words where things like accents and tenses have been added
Consistency has been given to sentences that were not understood.
Translations of words that were in English within a sentence have also been made.


joshuaboniface:
Just an FYI for future @97carmine: we have a Weblate up at https://translate.jellyfin.org to do translations - you might find it more productive than manual editing!

PR #342, opened 2019-05-25
Add span class to fix a margin issue with some icons

grafixeyehero:


DrPandemic:
I'm trying to test the change, but I'm unsure how to reach that section.

grafixeyehero:
@DrPandemic on version 10.3 home library section look like 
Screenshot_2019-05-16 Jellyfin
on master version 
Screenshot_2019-05-26 Jellyfin(4)

this fix the margin 
Screenshot_2019-05-26 Jellyfin(2)
Screenshot_2019-05-26 Jellyfin(1)
Screenshot_2019-05-26 Jellyfin

dkanada:
I would say the icons could use even more spacing before the text, but this is an excellent change. I assumed the issue was my unpopular browser.

PR #343, opened 2019-05-26
Bundle npm dependencies using webpack

thornbill:
**Changes**
Proof of concept implementation to use webpack to bundle npm dependencies and expose them to require.js in browser. Currently only moves a single dependency (jstree) to being bundled in this fashion.

One issue we still need to address is how we handle distributing the output of the webpack build. While it's not ideal, we may need to commit the output in `dist` for it to be available in all the places it is currently used. I'm going to mark this as a draft until we have agreed on a solution for this.

**Issues**
Continues work on #21


dkanada:
I think we only use it in the server and cordova client at the moment. Since the latter already has a build step with npm tools I think the easiest solution is to stick with building it for the server as well. We can either use npm to build it before the server or use a prebuilt web client at a specific location. The prebuilt version could be downloaded from here if someone doesn't want to bother but it also means devs can use a separate git repo as the path and avoid the submodule issues. If we also make this change on cordova we can use the same local copy of web for both repos during development which I think is convenient.

EraYaN:
We can also host versioned GitHub releases with prebuild webclients for download. Or just make the server build system build the webclient as well, as needed, that probably gets us the most flexibility.

Committing `dist` is how you get bad bloated repos like the original Emby repo.

src/scripts/apploader.js
```diff
@@ -1,12 +1,24 @@
-! function() {
+(function() {
```
dkanada: Can we just unwrap the function entirely? I've always disliked this shorthand.

src/scripts/editorsidebar.js
```diff
@@ -149,7 +149,9 @@ define(["datetime", "jQuery", "material-icons"], function(datetime, $) {
         nodesToLoad = [], selectedNodeId = null, $.jstree.destroy(), $(".libraryTree", page).jstree({
             plugins: ["wholerow"],
             core: {
-                check_callback: !0,
+                // Disable animations because jQuery slim does not support them
```
dkanada: Can we migrate to normal jQuery with no issues to support animations?

webpack.config.js
```diff
@@ -20,16 +28,27 @@ module.exports = {
             path.resolve(__dirname, 'src/bower_components/apiclient'),
             path.resolve(__dirname, 'src/bower_components/apiclient/sync'),
             path.resolve(__dirname, 'src/components/cardbuilder'),
```
dkanada: Do we still need the above modules or can we remove them?

DrPandemic:
I really like this PR and I can't wait to use bundler for every dependency. My 2 cents on the `how we handle distributing the output of the webpack build.`, would be to add a git submodule. So the cordova client would add this repo as a submodule and during the build process we could "simply" run webpack in the web folder/submodule.

I'm not fan of adding `dist/` to git since it will bloat and slow down every git action on the repo.

thornbill:
We had some good discussions around this on Matrix today. I think the best solution is to put it on npm. My biggest concern was with how to handle it in the server and it would appear that can be handled in a csproj file.

One thing that was pointed out is we should maintain a way to use a different version for testing pull requests or local changes. In Android and probably others we could use an environment variable to specify a different location to use for the web directory.

EraYaN:
We can always just use GitHub Releases (at least for the server) so one does not need to have npm in their path to build the server.

And the server could easily make the path to the web-ui configurable.

src/scripts/apploader.js
thornbill: I refactored this file to further simplify the logic and removed the other IIFE, but imo this one should stay to avoid polluting the browser's global scope with values defined here.

src/scripts/editorsidebar.js
thornbill: I'm not opposed to that, but I don't think it should be included in the scope of the webpack changes. This is only dealing with animating the metadata tree view which seems to have been disabled by Emby by editing the jstree source.

webpack.config.js
thornbill: They have been removed now.

PR #348, opened 2019-05-29
Fix firefox dropdown

grafixeyehero:
Before 
Screenshot_2019-05-30 Jellyfin(1)

After
Screenshot_2019-05-30 Jellyfin

Tested on Version 67.0 (64-bit)

### DIFF
--- a/src/components/emby-select/emby-select.js
+++ b/src/components/emby-select/emby-select.js
@@ -116,7 +116,7 @@ define(['layoutManager', 'browser', 'actionsheet', 'css!./emby-select', 'registe
             inputId++;
         }
 
-        if (!browser.firefox) {
+        if (browser) {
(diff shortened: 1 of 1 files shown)


src/components/emby-select/emby-select.js
```diff
@@ -116,7 +116,7 @@ define(['layoutManager', 'browser', 'actionsheet', 'css!./emby-select', 'registe
             inputId++;
         }
 
-        if (!browser.firefox) {
+        if (browser) {
```
dkanada: this conditional is useless now so it should be removed

grafixeyehero: @dkanada If i remove the codition some part doesn't work that look like this
Screenshot_2019-06-01 Jellyfin(2)

PR #350, opened 2019-05-31
Move fanart provider to plugin

dkanada:


PrplHaz4:
i think this link belongs here too: https://github.com/jellyfin/jellyfin-plugin-fanart

anthonylavado:
Noted: https://github.com/jellyfin/jellyfin-plugin-fanart/issues/1

PR #351, opened 2019-05-31
General improvements to user's profile photos

ferferga:
This PR attempts to improve some quirks of how profile photos were displayed in Jellyfin Web in User's dashboard page and login page (when users aren't hidden from login screens)

Before: 

After:

EDIT: Latest commit fixes a seam I forgot in activity page:
Captura

Also, JS scripts now doesn't call for specific resolutions of the picture (but for the main one in the header). This might be seen as a concern for mobile devices where data traffic is a problem, but asking two resolutions instead of many across the pages means that the browser caches only two versions and need to make less requests. This is also great for High DPI devices (Surface tablets screens' for example) and mobile devices:

Screenshot_2019-06-01-21-05-49


src/scripts/librarymenu.js
```diff
@@ -79,8 +79,8 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
 
     function updateHeaderUserButton(src) {
         if (src) {
-            headerUserButton.classList.add("headerUserButtonRound");
-            headerUserButton.innerHTML = '<img src="' + src + '" />';
+            headerUserButton.classList.add("headerUserButtonRound");
```
thornbill: Looks like there was some trailing whitespace added here.

src/controllers/dashboardpage.js
```diff
@@ -532,7 +532,7 @@ define(["datetime", "events", "itemHelper", "serverNotifications", "dom", "globa
             if (session.UserId && session.UserPrimaryImageTag) {
                 return ApiClient.getUserImageUrl(session.UserId, {
                     tag: session.UserPrimaryImageTag,
-                    height: 24,
+                    // height: 52,
```
thornbill: If this is no longer needed, I would just remove it.

src/components/emby-button/emby-button.css
```diff
@@ -148,15 +148,18 @@
         vertical-align: middle;
     }
 
-    .paper-icon-button-light > img {
-        width: 1.72em;
+    .paper-icon-button-light > div {
+        /* width: 2.72em; */
```
thornbill: If this is no longer needed, I would just remove it.

ferferga:
Performed all your suggested changes in the new commit!

Thanks!

src/scripts/librarymenu.js
thornbill: Still looks like there are some spaces at the end of this line. All other changes look good!

ferferga:
Pushed a new commit with the fix for the trailing whitespace. Seems that some changes made by Notepad++ after editing the original with VSCode left the trailing whitespace there. Sorry!

ferferga:
I did the same with the profile photo at the top of the page that I've done with the profile photos in the dashboard: point to the full resolution image and let CSS and the browser display the images with the correct size.

I found that at some devices with really small sizes like my old LG-E460 or old iPods Touchs display some seams. After discovering this, it also seemed pointless to me to even download two copies of the same photo in different sizes (as explained in the first post).

src/scripts/librarymenu.js
```diff
@@ -816,7 +811,7 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
                 pageTitleElement.innerHTML = "";
             }
 
-            document.title = "Jellyfin";
+            document.title = "Cockflix";
```
joshuaboniface: ??????

src/scripts/librarymenu.js
```diff
@@ -840,7 +835,7 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
                 pageTitleElement.innerHTML = html || "";
             }
 
-            document.title = title || "Jellyfin";
+            document.title = title || "Cockflix";
```
joshuaboniface: ibid.

ferferga:
Broke the repository while reverting some changes pushed to the public repository from my own local testing branch. Closing PR until I figure this out.

PR #352, opened 2019-05-31
Remove media info detial form none admin user page

grafixeyehero:
Before 
Screenshot_2019-05-31 Jellyfin(2)

After 
Screenshot_2019-05-31 Jellyfin(1)


cvium:
This seems counterproductive. Makes it very hard for non admin users to see if they can direct play it

grafixeyehero:
@cvium Isn't it enough this for user
Screenshot_2019-05-31 Jellyfin(3)

Hello-User:
The more vebose level of media information has been useful to me in the past and it would be very counterproductive to remove it imo. Maybe just compress it to what you have and have a button to extend the rest of the info?

PR #353, opened 2019-05-31
Add proper styling to DlnaProfile and ScheduledTask pages

grafixeyehero:
Before
Screenshot_2019-06-01 Jellyfin(1)

After
Screenshot_2019-06-01 Jellyfin

### DIFF
--- a/src/components/dialogHelper/dialoghelper.css
+++ b/src/components/dialogHelper/dialoghelper.css
@@ -139,8 +139,8 @@
     }
 
     .dialog-small {
-        width: 60%;
-        height: 80%;
+        width: 100%;
+        height: 100%;
     }
 
     .dialog-fullscreen-border {
--- a/src/components/dialogHelper/dialoghelper.css
+++ b/src/components/dialogHelper/dialoghelper.css
@@ -139,8 +139,8 @@
     }
 
     .dialog-small {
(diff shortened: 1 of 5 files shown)


src/components/dialogHelper/dialoghelper.css
```diff
@@ -139,8 +139,8 @@
     }
 
     .dialog-small {
-        width: 60%;
-        height: 80%;
+        width: 100%;
```
DrPandemic: Something seems a little bit weird here, `.dialog-small` becomes bigger than `.dialog-medium`. Maybe you should use another class or rename this one?

grafixeyehero:
> 1
> image
> 
> 2
> image
> 
which one did u prefer?

DrPandemic:
This page is already weird with all the bullet points, but disregarding that, I think 1 is more readable.

grafixeyehero:
@DrPandemic way u don't get the same rendering on firefox and chrome. b/c on firefox the emby-button inject button-inline class for this i will open a new PR that fix for both. And for the bullet points i will remove it only it appere on the title

grafixeyehero:
> 
> 
> Except for the `button-link-inline` this looks good!

See PR  #356 Fix some style

PR #356, opened 2019-06-05
Fix certain elements getting incorrectly centered on firefox

grafixeyehero:
note it's better to remove the class button-link-inline. For know i just disabled form css 

before 
Firefox
58764365-1514f780-8534-11e9-8962-2ce7ebe96016
Screenshot_2019-06-05 Jellyfin
Screenshot_2019-06-05 Jellyfin(1)

chrome
58764371-2e1da880-8534-11e9-9d55-eecd4151dfe6
Screenshot (4)
Screenshot (5)

### DIFF
--- a/src/components/emby-button/emby-button.css
+++ b/src/components/emby-button/emby-button.css
@@ -50,9 +50,9 @@
     vertical-align: initial;
 }
 
-.button-link-inline {
+/*.button-link-inline {
     display: inline;
-}
+}*/
 
 .button-link:hover {
     text-decoration: underline;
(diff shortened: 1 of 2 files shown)


dkanada:
It could have been a workaround for older versions. I tested the class as well and it definitely fixes the spacing when removed. @grafixeyehero rather than comment out the changes I'd say you can just remove it entirely.

I think @DrPandemic was talking about the lack of styling for the supported formats, which are shown as plain text. It's not in the scope of this pull request, but they definitely look out of place. I am wondering if it's a bug with the routing that breaks an element or if they just don't have any style in the first place, which might be worse.

dkanada:
I updated the title to explain exactly what issue is getting fixed here, let me know if I missed the issue you are targeting.

grafixeyehero:
@dkanada i am agree with u removing the class entirely.

PR #357, opened 2019-06-06
Improvements to the quality of artworks and user images

ferferga:
This Pull Request attempts to improve some quirks of how profile photos were displayed in Jellyfin Web in User's dashboard page (activity log and user panel) and login page (when users aren't hidden from login screens), as well as how artwork and photos from actors and movies are displayed.

**Before (Profile Photos):**

58699115-0a742a00-839d-11e9-837c-d90af9aa4347
58699119-0b0cc080-839d-11e9-8d97-2331cda29882
Captura1-3
Captura1-1

**After (Profile Photos):**

58699117-0b0cc080-839d-11e9-8294-7b9f3277a3fc
58699118-0b0cc080-839d-11e9-9e55-5217f2542110
58752554-19e49780-84b1-11e9-96b0-6980c53c08ac
Captura2-1

All the profile photo scaling performed by the server is now removed. Now images are scaled by the browser using CSS. Although using the full resolution image might be seen as an overkill for data usage (specially in mobile devices), modern browsers use the cache for that matter in an intelligent way, so, in my opinion, we have two benefits here: 

1. Images always have great quality, no matter the resolution of the device
2. The user image is only downloaded one time at full resolution, instead of downloading it in smaller sizes multiple times in different places of the web app.

(The mobile version is one place where you can see the benefit of having the high resolution photo. As you can see, there are no seams, no matter how you scale the page):

58752578-7fd11f00-84b1-11e9-93dc-52f58ddfe0c3

The same has been applied to the artwork, backgrounds and actor's portrait's of the media: Now the images are always downloaded at full resolution and are scaled by the browser itself. The gain in quality is really noticeable:

**Before:**

Captura1-2
Captura1

**After:**

Captura2
Captura2-2

### DIFF
--- a/src/components/activitylog.js
+++ b/src/components/activitylog.js
@@ -13,8 +13,7 @@ define(["events", "globalize", "dom", "datetime", "userSettings", "serverNotific
         if (entry.UserId && entry.UserPrimaryImageTag) {
             html += '<i class="listItemIcon md-icon" style="width:2em!important;height:2em!important;padding:0;color:transparent;background-color:' + color + ";background-image:url('" + apiClient.getUserImageUrl(entry.UserId, {
                 type: "Primary",
-                tag: entry.UserPrimaryImageTag,
-                height: 40
+                tag: entry.UserPrimaryImageTag
             }) + "');background-repeat:no-repeat;background-position:center center;background-size: cover;\">dvr</i>"
         } else {
             html += '<i class="listItemIcon md-icon" style="background-color:' + color + '">' + icon + '</i>';
--- a/src/components/backdrop/backdrop.js
+++ b/src/components/backdrop/backdrop.js
@@ -181,21 +181,6 @@ define(['browser', 'connectionManager', 'playbackManager', 'dom', 'css!./style']
         currentLoadingBackdrop = instance;
     }
 
-    var standardWidths = [480, 720, 1280, 1440, 1920];
-    function getBackdropMaxWidth() {
-
-        var width = dom.getWindowSize().innerWidth;
-
-        if (standardWidths.indexOf(width) !== -1) {
-            return width;
-        }
-
-        var roundScreenTo = 100;
-        width = Math.floor(width / roundScreenTo) * roundScreenTo;
-
-        return Math.min(width, 1920);
-    }
-
     function getItemImageUrls(item, imageOptions) {
 
         imageOptions = imageOptions || {};
@@ -209,7 +194,6 @@ define(['browser', 'connectionManager', 'playbackManager', 'dom', 'css!./style']
                 return apiClient.getScaledImageUrl(item.BackdropItemId || item.Id, Object.assign(imageOptions, {
                     type: "Backdrop",
                     tag: imgTag,
-                    maxWidth: getBackdropMaxWidth(),
                     index: index
                 }));
             });
@@ -222,7 +206,6 @@ define(['browser', 'connectionManager', 'playbackManager', 'dom', 'css!./style']
                 return apiClient.getScaledImageUrl(item.ParentBackdropItemId, Object.assign(imageOptions, {
                     type: "Backdrop",
                     tag: imgTag,
-                    maxWidth: getBackdropMaxWidth(),
                     index: index
                 }));
             });
--- a/src/components/cardbuilder/card.css
+++ b/src/components/cardbuilder/card.css
@@ -232,9 +232,10 @@ button {
     left: 0;
     right: 0;
     bottom: 0;
-    background-size: contain;
+    background-size: cover;
     background-repeat: no-repeat;
-    background-position: center bottom;
+    background-position: center;
+    border: none;
 }
 
 .cardImage-img {
@@ -259,6 +260,10 @@ button {
 .coveredImage {
     background-size: 100% 100%;
     background-position: center center;
+}
+
+.cardProfilePhoto {
+    border-radius: 100%;
 }
 
 .coveredImage-noScale {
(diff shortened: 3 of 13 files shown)


cvium:
Wrong approach imo. Images can be quite big. We should do something like this instead: https://css-tricks.com/responsive-images-css/

ferferga:
**Important**: I would consider the changes to the quality of the artwork as a WIP. I did the changes yesterday and I didn't have the time to test them as much as I did with the profile photos. They work perfectly in Chrome, both in mobile and desktop. I encourage testing it in other browsers/environments, as maybe the reduction in sizes was there for something else apart from (I suppose) data traffic reduction.
Also, the difference in quality can't be seen in screenshots.

ferferga:
> Wrong approach imo. Images can be quite big. We should do something like this instead: https://css-tricks.com/responsive-images-css/

Wow, you were faster than me :). That was what I wanted to clarify, that the approach might not be the correct one as it was a quick fix, and that most of the code is untidy (JS methods for resizing the images are still there, for example).
I did the changes yesterday morning and I didn't see any quirk or issue with them however, that's why I publish it in this PR. 

However, if you are seeing any "display" issues after cloning my repository and testing it, kindly tell me so I can take a look.
I double quote display because I want to make sure that the rendering of the DOM is correct first, and, once that it's done, dive more into cleaning the code.

dkanada:
I like the circlular shape everywhere except the login page. The current interface uses cards for certain lists (servers, libraries, and users are the three I can think of right now) and so the square image makes sense with those cards, which are also used on the login and server selection pages.

ferferga:
> I like the circlular shape everywhere except the login page. The current interface uses cards for certain lists (servers, libraries, and users are the three I can think of right now) and so the square image makes sense with those cards, which are also used on the login and server selection pages.

So, you think it's right as well for the "Profile" page?. I see your point, and somehow I think the same. However, I think that the card design on login without any effects while hovering and the lack of shadows (like the items in the library) gives the impression that the image isn't "clickable", at least for me. I thought about the actual design because the profile photo was actually rounded in some places and because I think that people already expect profile pictures + circular shapes to be clickable.

I think the main point here is whether we stick to profile photos squared everywhere (and distinguish it between other UI elements) or keep all using cards.

Bond-009:
I would rather have the full size images than the current blurry mess. The resolution of the artwork is currently just to low for big and high res screens.

dkanada:
I think removing the max size on user avatars is fine for now but we should hold off on all other instances. I also think the changes to the user cards should be reverted, but all the other circular changes are fine.

We should actually just remove the user cards from the login page. I can't think of another service that has cards (netflix has it but you have to login normally first) to avoid typing the username, and it's a huge security flaw that we should probably just close. @sparky8251 I think you're right on this one.

ferferga:
Good afternoon. I've been out for a while and I thought that I replied to this at the same time as this [PR](https://github.com/jellyfin/jellyfin/pull/1455)



> I think removing the max size on user avatars is fine for now but we should hold off on all other instances. I also think the changes to the user cards should be reverted, but all the other circular changes are fine.
> 
> We should actually just remove the user cards from the login page. I can't think of another service that has cards (netflix has it but you have to login normally first) to avoid typing the username, and it's a huge security flaw that we should probably just close. @sparky8251 I think you're right on this one.

What's the reasoning for leaving other instances untouched? They scale fine and the blurry mess as @Bond-009 mention is gone.

I completely agree with you about removing the users from login page. It's a huge security flaw, though maybe some people who use it might be annoyed by the removal of the feature. One thing that I noticed the other day that needs fixing for sure is the "Manual login" button that still appears when all users are hidden: it's plain redundant. But that's for another PR.

dkanada:
> What's the reasoning for leaving other instances untouched?

I like the circle icon changes everywhere but they just don't make sense with the square card layout. The cards themselves do need some work but that's another issue entirely that we can address in another pull request.

ferferga:
@dkanada Ah, nvm, I thought you were talking about the resolution changes to all posters and images.

oddstr13:
As this is a feature/enhancement pr, and not a regression/security bug fix, it should be retargeted and rebased to the master branch.

joshuaboniface:
@ferferga I've retargeted against master, but there's a merge conflict. Are you able to fix that up?

ferferga:
@joshuaboniface Yeah of course. I'm out of home right now, as soon as I get back and find some time I will take a look.

joshuaboniface:
@dkanada Any other thoughts here? Is it good to you?

@ferferga Sorry another merge issue, been a couple of PRs going in in prep for 10.4.0.

If we're going to do this, I want to get it in this release

ferferga:
@joshuaboniface Let me look this weekend :)

ferferga:
I could finally make a revision of the changes with the newer commits in a short spare time I had during studies. Hope that everything is good now.

joshuaboniface:
Hey @ferferga I hate to say this, but I think this PR is irrevocably messed up. I can't even check it out using `hub pr checkout`. Plus a bunch more merge conflicts from the first commit, which is really weird.

I'm sorry to ask more work, but are you able to make up a fresh PR with just the expected, latest changes, based on the `master` branch? If you need a hand, please pop onto our Matrix chats (room #jellyfin-dev:matrix.org) and give me a ping. We do want this in!

ferferga:
> Hey @ferferga I hate to say this, but I think this PR is irrevocably messed up. I can't even check it out using `hub pr checkout`. Plus a bunch more merge conflicts from the first commit, which is really weird.
> 
> I'm sorry to ask more work, but are you able to make up a fresh PR with just the expected, latest changes, based on the `master` branch? If you need a hand, please pop onto our Matrix chats (room #jellyfin-dev:matrix.org) and give me a ping. We do want this in!

Will try to check this weekend

ferferga:
Created the new PR as suggested in #512. Closing this PR.

PR #361, opened 2019-06-13
Allow Jellyfin and Emby to coexist on the same domain

oddstr13:
Jellyfin and Emby both overwrite the `servercredentials3` local storage object when saved auth doesn't match. Renaming this key allows both to stay logged in on the same domain.

See jellyfin/jellyfin-apiclient-javascript#14 for corresponding apiclient pr

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -21,6 +21,7 @@
  - [RazeLighter777](https://github.com/RazeLighter777)
  - [LogicalPhallacy](https://github.com/LogicalPhallacy)
  - [thornbill](https://github.com/thornbill)
+ - [Oddstr13](https://github.com/oddstr13)
 
 # Emby Contributors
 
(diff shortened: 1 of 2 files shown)


thornbill:
I added a comment in the corresponding apiclient pull request https://github.com/jellyfin/jellyfin-apiclient-javascript/pull/14#pullrequestreview-249686649

oddstr13:
Targeting of release branch was intentional, as I consider this a bug fix and not a feature.
(I would also prefer getting it in sooner rather than later - next build including master is likely db rewrite, which is likely a while before it is ready)

There is one side effect from this patch - users will get logged out on update.
This could be mitigated by including code to  copy the old key to the new - but I'm not currently familiar enough with the code base to do that, I just did a search and replace.

As for the name of the variable, this doesn't really matter to me, but I would prefer `jellyfin_credentials` over `JellyfinCredentials` (CamelCase screams "this is a class" at me, and that just feels wrong)

thornbill:
`jellyfin_credentials` sounds good to me. 👍

oddstr13:
`jellyfin_credentials` rather than `jf_servercredentials`?

@joshuaboniface  @DrPandemic

DrPandemic:
I think `jellyfin_credentials` reads better since we don't really ever use the jf acronym, but both are good.

PR #365, opened 2019-06-19
Fix repeated automatic attempts to cast to a device

dkanada:


src/components/chromecast/chromecastplayer.js
```diff
@@ -211,19 +189,16 @@ define(['appSettings', 'userSettings', 'playbackManager', 'connectionManager', '
      * session update listener
      */
     CastPlayer.prototype.sessionUpdateListener = function (isAlive) {
-
-        //console.log('sessionUpdateListener alive: ' + isAlive);
-
         if (isAlive) {
-        }
-        else {
+            console.log('sessionUpdateListener: not alive');
```
oddstr13: either variable is inverted logic, or log message is wrong

dkanada: forgot an exclamation mark, you're right

PR #366, opened 2019-06-21
Minor fixes for section headers

dkanada:


dkanada:
The left margins for section headers were all messed up when they had cards below them in some places. I just made a few adjustments to fix that issue everywhere.

PR #379, opened 2019-07-01
Add an ImageHelper class to consolidate frequently used icons

dkanada:
I implemented my suggestion in #326 and also found a clean solution for #154 so we can stop using icons from Emby's repository.

I don't think clients should report an icon at all, which means the icon feature is unique to the web client and doesn't need icons hosted anywhere. I use the device name to generate an icon at the moment but if a client ever wants to add unique device names we might need to implement something like device type.


src/components/apphost.js
```diff
@@ -310,9 +311,11 @@ define(["appSettings", "browser", "events", "htmlMediaHelper"], function (appSet
 
     var deviceId;
     var deviceName;
+    var appName = "Jellyfin Web";
+    var appVersion = "10.3.6";
```
EraYaN: If this the only place where the version is kept in the Web UI?

src/scripts/imagehelper.js
```diff
@@ -0,0 +1,64 @@
+define(["browser"], function (browser) {
+    "use strict";
+
+    function getDeviceIcon(device) {
+        var baseUrl = "img/devices/";
+        switch (device) {
+            case "Opera":
+            case "Opera TV":
+                return baseUrl + "opera.svg";
+            case "Samsung Smart TV":
```
EraYaN: Might want to add LG webOS TV too.

src/components/apphost.js
dkanada: yes

src/scripts/imagehelper.js
dkanada: The icon pack I used didn't have any for those devices, if someone finds a decent icon later I figure we can add them then.

anthonylavado:
@dkanada Just need to check up the conflicts and I'm ready to approve/merge

PR #381, opened 2019-07-02
Fix homesection programs button to go to correct page

dinki:
The button on the homesection/homepage for Live Tv-> Programs 

image

This did not work due to not having a section assigned.  Without the assignment the link just went to livetv.html which brings up the guide.  This PR sends the user to the correct page containing the 'Programs'

### DIFF
--- a/src/components/homesections/homesections.js
+++ b/src/components/homesections/homesections.js
@@ -655,7 +655,7 @@ define(['connectionManager', 'cardBuilder', 'appSettings', 'dom', 'apphost', 'la
         }
 
         var userId = user.Id;
-        return apiClient.getLiveTvRecommendedPrograms({
+        apiClient.getLiveTvRecommendedPrograms({
             userId: apiClient.getCurrentUserId(),
             IsAiring: true,
             limit: 1,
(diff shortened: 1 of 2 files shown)


dinki:
> So many Live TV improvements, nice!

Glad I can finally contribute.  It's not much but fixes some annoyances for me at least!

PrplHaz4:
@dinki this does not appear to be broken for me in FF on 10.3.5 - was this broken in something that hasn't been released yet?  unfortunately can't test the PR for a few days....

dinki:
@PrplHaz4 Really?  Using Chrome/Opera the button takes me to the 'Guide' page rather than to the 'Programs' page.  I do have my default screen for Live TV set to 'guide, but I can't see any case where clicking on the 'Programs' button shouldn't bring you anywhere other than the 'Programs' page.  I modeled the PR after the other buttons in that section and the way it is done in the livetv.html header.

PrplHaz4:
@dinki looking at your changes I was surprised it worked at all before...I was able to repro by changing the default LiveTV page from `Suggestions` to `Guide` 👍

dinki:
@PrplHaz4 Yeah, I just switched mine from 'Guide' to 'Suggestions':

image

and clicking on the 'Live TV' link from here:
image

brings me to the programs part of 'livetv.html'.  That is also the default landing page if you go to 'livetv.html?serverId=[yourid]' .  I think that's why yours is working fine and mine is not.  Can you try changing your Home Settings to 'Guide' and test it?

dinki:
@PrplHaz4 ah okay.. so you have confirmed what I thought.  Excellent.  Thanks

PR #382, opened 2019-07-02
Fix homesection 'On now' display elements

### LINKED ISSUE #373: 'On Now' section on main page not displaying elements (REGRESSION)
image

The 'On Now' section from the above image used to show currently air programs on Live TV.  @thornbill confirmed that this was working as of 10.2.2 but is broken currently.  I am running 10.3.3 and it was broken in that release as well.

dinki:
The 'On now' section on the homesection/homepage  was not returning any media elements.  This fixes that.  The 'return' was left out so data never arrived.

Thanks for spotting it @thornbill 

Fixes #373


dinki:
Fixes this issue https://github.com/jellyfin/jellyfin-web/issues/373

PR #388, opened 2019-07-10
Enable display mode setting

Nickbert7:
image

Hosted preview with some other changes: https://nickbert7.github.io/jellyfin-web/src/#!/selectserver.html

### DIFF
--- a/src/components/focusManager.js
+++ b/src/components/focusManager.js
@@ -40,7 +40,7 @@ define(['dom'], function (dom) {
 
         try {
             element.focus({
-                preventScroll: true
+                preventScroll: false
             });
         } catch (err) {
             console.log('Error in focusManager.autoFocus: ' + err);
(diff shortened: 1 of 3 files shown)


Nickbert7:
In TV it shows an additional home screen setting:
image

But so far it seems not to have any effect...

EraYaN:
The horizontal home screen seems to be completely missing for some reason.

src/components/focusManager.js
```diff
@@ -40,7 +40,7 @@ define(['dom'], function (dom) {
 
         try {
             element.focus({
-                preventScroll: true
+                preventScroll: false
```
EraYaN: Maybe this fixes some the the webOS problems? @oddstr13

dkanada: According to chat logs this didn't fix anything for WebOS.

PR #389, opened 2019-07-15
Fix menu jumping up and down

### LINKED ISSUE #158: Whole menu is jumping up and down when selecting different items
**Describe the bug**
Whole menu is jumping up and down when selecting different items

**To Reproduce**
1. Go to the dash
2. Click on some menu items

**Expected behavior**
The menu stays where it is.

**Screenshots**

 - OS: Raspbian
 - Browser: Firefox
 - Jellyfin Version: N/A (build from master)

redSpoutnik:
This is an attempt to fix #158 without changing header tabs behavior.

**Changes**

Just move up headerTabs and remove 'withSectionTabs' specific margin-tops in css.

## Before: 
before-without-tabs
before-with-tabs

## After:
after-without-tabs
after-with-tabs

**Issues**

#158

### DIFF
--- a/src/css/librarybrowser.css
+++ b/src/css/librarybrowser.css
@@ -247,23 +247,17 @@
         width: 20.205em !important;
         font-size: 94%
     }
+    
+    .headerTabs {
+    	margin-top: -2em;
+    }
 
     .dashboardDocument .mainDrawer-scrollContainer {
         margin-top: 5em !important
     }
 
-    .dashboardDocument withSectionTabs .mainDrawer-scrollContainer {
-        margin-top: 8.7em !important
-    }
-
     .dashboardDocument .skinBody {
         left: 20em
-    }
-}
-
-@media all and (min-width:40em) and (max-width:84em) {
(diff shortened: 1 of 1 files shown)


dkanada:
Can you explain the fix a bit more? The only reason the behavior changes is to add a new row for smaller screens. Otherwise, the tabs should stay in the same row. Why was this causing the menu to jump around?

redSpoutnik:
Sure, section-tabs row is on the same level as scrollcontainer's top.

To avoid scrollcontainer hiding section-tabs, scrollcontainer has a custom margin-top when section-tabs are displayed.

But, as some sections don't have section-tabs and jellyfin's theme is mostly dark, switching scrollcontainer's margin-top between section-tabs and not section-tabs produces a visual effect: like if scrollcontainer's size was jumping.

Sorry for my bad english, tell me if it isn't clear?

redSpoutnik:
@dkanada Fair enough, should we then close this PR?

anthonylavado:
Closing for now.

PR #396, opened 2019-07-28
Move download action to major button for books (itemdetailspage)

Abbe98:
**Changes**
Because EBooks can't be viewed within jellyfin-web this change promotes the download action from the context menu to a always visible button.


anthonylavado:
@Abbe98 Awesome. Thank you for contributing to Jellyfin!

PR #398, opened 2019-07-29
Add stale app config.

EraYaN:
Adds a config file for the Stale App, to help keep the issues list fresh, I think 60 days is fair as a starting point. I added most tags I think should be exempt.

Link: https://github.com/apps/stale


anthonylavado:
Just as a note, Stale-bot is added, so when this is merged, it should go to work (with a short delay).

PR #400, opened 2019-07-31
Add base url option to web client

dkanada:
merge this one with jellyfin/jellyfin#1588

### DIFF
--- a/src/addserver.html
+++ b/src/addserver.html
@@ -1,20 +1,16 @@
 <div data-role="page" class="page standalonePage">
     <div class="padded-left padded-right padded-bottom-page">
-        <form class="manualServerForm" style="margin: 0 auto;">
-            <h1 style="text-align: left;">${HeaderConnectToServer}</h1>
+        <form class="addServerForm" style="margin: 0 auto;">
+            <h1>${HeaderConnectToServer}</h1>
             <div class="inputContainer">
                 <input is="emby-input" type="text" id="txtServerHost" required="required" label="${LabelServerHost}" autocomplete="off" spellcheck="false" autocapitalize="none" autocorrect="off" />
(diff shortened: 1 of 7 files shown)


src/addserver.html
```diff
@@ -1,20 +1,16 @@
...
+        <form class="addServerForm" style="margin: 0 auto;">
+            <h1>${HeaderConnectToServer}</h1>
             <div class="inputContainer">
                 <input is="emby-input" type="text" id="txtServerHost" required="required" label="${LabelServerHost}" autocomplete="off" spellcheck="false" autocapitalize="none" autocorrect="off" />
-                <div class="fieldDescription" style="text-align: left;">${LabelServerHostHelp}</div>
+                <div class="fieldDescription">${LabelServerHostHelp}</div>
             </div>
-            <br />
-            <div class="inputContainer">
-                <input is="emby-input" type="number" id="txtServerPort" step="1" min="0" value="8096" label="${LabelServerPort}" />
```
cvium: Not sure I agree with this change

dkanada: What would you suggest as an alternative? Makes no sense to have a port if we expect the user to use a Base URL in the hostname field. I'm also not a fan of this change just because it makes the page look a bit empty but don't think it will be confusing for users.

cvium: I guess I'd rather have them all separate.

dkanada: I don't have a strong opinion but I'd like to get more input since this is the first page you see on android and electron clients. It will also exist in the future on web instances that support a multiserver toggle. @joshuaboniface @anthonylavado @grafixeyehero @vitorsemeano and whoever else, if you have an opinion post it here.

anthonylavado: My opinion is to have a consistent dialog amongst all possible clients. Having a single field looks cleaner, and as will probably serve us better as we get into wacky URL land.

Issue #384, opened 2019-07-04
UI Bug - Artist name overlaps on the control buttons


ar7x9:
The bug is that the names of the artist of the songs are overlapping with control buttons (play, pause, etc.).

Bug occurs when artist name is extremely long.

Overlap should not have occurred

![Image](https://user-images.githubusercontent.com/51527759/60668148-c9b77700-9e5a-11e9-965f-f0b9b137c6d5.png)

**System (please complete the following information):**
 - OS: Windows 1080p 125% scale
 - Browser: Firefox
 - Jellyfin Version: 10.3.5.0

anthonylavado:
Fixed by #403.

PR #403, opened 2019-08-03
Fix text overlap on now playing bar

grafixeyehero:
**Issues**
UI Bug - Artist name overlaps on the control buttons #384

### DIFF
--- a/src/components/nowplayingbar/nowplayingbar.css
+++ b/src/components/nowplayingbar/nowplayingbar.css
@@ -6,6 +6,11 @@
     overflow: hidden;
 }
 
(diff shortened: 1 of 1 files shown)


dkanada:
I don't think we need a special rule for this, the `nowPlayingBarInfoContainer` class includes the entire left side and already hides any overflow. We should just constrain the right end of this container to the left end of the center controls so they don't overlap.

grafixeyehero:
@dkanada i set for nowPlayingBarInfoContainer to use 40% width & nowPlayingBarRight use the rest on full-screen on 576px  nowPlayingBarInfoContainer use 70% of the screen

PR #406, opened 2019-08-07
Enable blurred footer in chrome

Nickbert7:
Enables blurred footer also in chrome

**Changes**
Displays the now playing bar blurred in chrome as in other browsers

Before
image

After
image


Nickbert7:
My intention was just to remove the special handling for chrome... but somehow also Firefox has no blur...

Without the change:
Chrome:
image

Firefox:
image

Edge:
image

It would also be fine for me to remove the blur completely, it should just look the same within every browser...

PR #407, opened 2019-08-10
[WIP] Deuglify part-2

grafixeyehero:


dkanada:
Before you work on this any more could you link the source you're using to deminify the code? I'm worried the code may have changed since it was last updated, which makes these two pull requests harder to review since they're so large. It honestly might be better to just run `eslint --fix` one time on these files, which will use the existing source and avoid any possibility of breakage from code changes.

grafixeyehero:
> 
> 
> Before you work on this any more could you link the source you're using to deminify the code? I'm worried the code may have changed since it was last updated, which makes these two pull requests harder to review since they're so large. It honestly might be better to just run `eslint --fix` one time on these files, which will use the existing source and avoid any possibility of breakage from code changes.

i link it on matrix

grafixeyehero:
closing the pull requests for know.
@dkanada commited harder to review since they're so large.
I will gone to reduce it

PR #412, opened 2019-08-16
Move transcoding page to playback settings and improve strings

dkanada:


cvium:
Needs some screens

dkanada:
@cvium the library pages were just internal, the only UI change (excluding the strings) was moving the transcode settings to the playback section.

before
after

PR #417, opened 2019-08-22
add favorite books section to favorite tab

Abbe98:
Adds support for Books in the favorite tab in the home view.

### DIFF
--- a/src/controllers/favorites.js
+++ b/src/controllers/favorites.js
@@ -108,6 +108,15 @@ define(["appRouter", "cardBuilder", "dom", "globalize", "connectionManager", "ap
             overlayMoreButton: !0,
             action: "instantmix",
             coverImage: !0
+        }, {
+            name: "HeaderFavoriteBooks",
+            types: "Book",
+            shape: getPosterShape(),
+            showTitle: true,
+            showYear: true,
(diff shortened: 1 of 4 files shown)


src/controllers/favorites.js
```diff
@@ -108,6 +108,15 @@ define(["appRouter", "cardBuilder", "dom", "globalize", "connectionManager", "ap
             overlayMoreButton: !0,
             action: "instantmix",
             coverImage: !0
+        }, {
+            name: "HeaderFavoriteBooks",
+            types: "Book",
+            shape: getPosterShape(),
+            showTitle: !0,
```
thornbill: We should use true/false for these. The others are only using this format because the source was originally uglified.

Abbe98: @thornbill does it make sense to leave this for a separate clean up effort? I could do the entire file but it would make sense to do it as a separate effort to have the change under a different merge.

dkanada: I think @thornbill just means the small section you added should use boolean values, we definitely don't expect you to change all of them in this pull request.

thornbill: Ah right. I meant just in the section you added. Thanks for clarifying @dkanada!

PR #424, opened 2019-08-27
Add a "Copy Stream URL" button in the dropdown menu, for easy Direct Play in VLC

### LINKED ISSUE #423: Add option to copy download URL to easily direct play using VLC
Right now Jellyfin has no easy way to force direct play on Windows, especially with x265 or 10bit files.

However, there is a hack I've discovered, if the server allows downloading, a user can:

1. Start download (Movie or TV episode)
2. Cancel it in browser
3. Copy the URL from downloads page
4. Open it as network stream in VLC

This begins an automatic sequential download, as well as direct stream (thanks to the power of VLC) of any media, as well as subtitle selection.

The problem is there is no easy way to directly copy this URL, forcing a download, cancel, and navigating to browser downloads to copy the URL.

Suggestion: Have an additional button to copy the URL to clipboard.

ckcr4lyf:
**Changes**
Modified the JS file that generates the dropdown options
If download is enabled on the server, allow easy copy of the Download URL

Copy Stream URL

Users can put this in VLC or a 3rd party player, allowing playback of x265 content without transcoding on PC.
Minor change, only one script file, and two language packs.

**Issues**
Fixes #423


ckcr4lyf:
P.S. Browsers can't launch applications, but this could be incorporated in the Electron Theater, to launch a local instance of VLC or MPV or something with that URL.

Would help fix https://github.com/jellyfin/jellyfin-theater-electron/issues/23

src/components/itemcontextmenu.js
```diff
@@ -305,6 +310,22 @@ define(['apphost', 'globalize', 'connectionManager', 'itemHelper', 'appRouter',
...
+                    var downloadHref = apiClient.getItemDownloadUrl(itemId);
+                    var textArea = document.createElement("textarea");
+                    textArea.value = downloadHref;
+                    document.body.appendChild(textArea);
+                    textArea.focus();
+                    textArea.select();
+                    try {
+                        document.execCommand('copy');
+                    } catch (err) {
+
```
cvium: Shouldn't we at least write the error to the console?

ckcr4lyf: Fixed. Thanks for the review.

puschie286:
is it possible to hide this ? 
guess 99.9% of the users never use this or just miss clicked

PR #430, opened 2019-08-31
Fix layout issues for device cards on dashboard

dkanada:
The information was getting pushed to a new line when the content was too long. I also stacked the progress bars on top of each other and removed the border radius.


joshuaboniface:
@dkanada Before and after pics?

Issue #431, opened 2019-08-31
Now playing bar controls not working


dkanada:
**Describe the bug**
Any click on the now playing bar will redirect to the now playing page rather than pause or control audio playback.

**To Reproduce**
Start playing music and try to pause or rewind the music.

PR #433, opened 2019-08-31
Quick fix Now playing bar controls

grafixeyehero:
**Issues**
Now playing bar controls not working #431

### DIFF
--- a/src/components/nowplayingbar/nowplayingbar.css
+++ b/src/components/nowplayingbar/nowplayingbar.css
@@ -94,7 +94,6 @@
     z-index: 2;
     display: flex;
     justify-content: flex-end;
-    flex-grow: 1;
     align-items: center;
     flex-shrink: 0;
 }


cvium:
Could you elaborate on why you removed the flex-grow?

grafixeyehero:
> Could you elaborate on why you removed the flex-grow?
`.nowPlayingBarRight` take 60% of the screen and `.nowPlayingBarInfoContainer` take 40% of the screen. when we add a flex-grow in `.nowPlayingBarRight` It dictates what amount of the available space inside the flex container the item should take the space. so the `.nowPlayingBarRight` cover the `.nowPlayingBarCenter `

grafixeyehero:
Before


After

cvium:
Shouldn't we remove flex-grow from the other two as well?

grafixeyehero:
> 
> 
> Shouldn't we remove flex-grow from the other two as well?

which one?

Issue #419, opened 2019-08-25
Delete Media confirmation is highlighted the wrong way around


Cyan101:
![ss (2019-08-25 at 23 29 22)](https://user-images.githubusercontent.com/2821230/63652276-f9b33480-c790-11e9-93e6-435ef84ed632.png)

Might be a bit opinionative but deleting should be the highlighted button as its the button taking action, while cancel isn't.

Many programs like discord, reddit, google drive all have their deletes highlighted and their cancels as the background color/non-highlighted.

anthonylavado:
I suppose we could do that and make the Delete button red as well

Cyan101:
Sure.

I've also been changing a bunch of settings on my server and it's not just
the delete popup that has cancel highlighted, so I'd suggest changing the
cancel and "action" button both.

I can have a look at making a PR for this later if you'd like.

Cyan101:
seems like it's actually mislabeled in the HTML

You can see here that the delete button has the button-cancel class, while the cancel button has the button-submit class.

![image](https://user-images.githubusercontent.com/2821230/63678786-80fcb880-c822-11e9-95e0-5450e95f8af9.png)

dkanada:
I think it was a design choice at the time. It's not a bad idea to default to the safe operation rather than the one that will destroy media.

Cyan101:
problem is the design of it clashes against "standards" that most other programs follow, I personally almost deleted some files because I avoided the highlighted button cause I was trying to cancel.

I was going to just make a PR with some CSS changes but since its in the HTML i'm not too sure which files I need to be looking for and changing, especially since its not just for the delete prompt.

dkanada:
Ah I see what you mean. The classes `button-cancel` and `button-submit` apply the color schemes. If you wanted to add another class next to those called `button-red` or something and then flip the button styles that would work.

It would be best to find all instances of inconsistent button coloring and change them all at once with the new class though.

Cyan101:
Yeah, the styles which are correctly coloured are linked to the wrong buttons, you could make another css class with `button-red` or `button-delete` but I think that's a second thing to worry about, first thing should be to find where the buttons are created and fix it up so that they have the correct class on both buttons.

Again, i'm not too sure where these buttons are being built so I don't think i'll be able to PR this

PR #435, opened 2019-08-31
Change primary highlight for delete confirmation dialog

grafixeyehero:
After
Screenshot (16)

**Issues**
Delete Media confirmation is highlighted the wrong way around #419

### DIFF
--- a/src/components/confirm/confirm.js
+++ b/src/components/confirm/confirm.js
@@ -18,13 +18,13 @@ define(['dialog', 'globalize'], function (dialog, globalize) {
         items.push({
             name: options.cancelText || globalize.translate('ButtonCancel'),
             id: 'cancel',
-            type: options.primary === 'cancel' ? 'submit' : 'cancel'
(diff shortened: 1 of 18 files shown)


dkanada:
There are exactly five instances of reversed behavior based on a quick search. @Cyan101 does that sound accurate? @grafixeyehero a better solution would be to remove the `primary` option and add a boolean to the options that will turn the primary option red, which is what we were talking about in the issue linked above.

Cyan101:
sounds about right, I think the css classes for them should be swapped
around and can be left at that, but if you do want to add a red one can
perhaps just add another class but might have to change how the buttons are
generated

dkanada:
@grafixeyehero your latest commit removes the submit functionality which we need to keep as an option. I can add something to keep both if you want.

dkanada:
@anthonylavado I'd like to get three reviews here since we're modifying a destructive action. I haven't tested this but the logic should be fine.

PR #386, opened 2019-07-05
Change media info display 

grafixeyehero:
Screenshot_2019-07-24 Jellyfin
Screenshot_2019-07-25 Jellyfin

PR #439, opened 2019-09-03
Add media file size to Playback Data

leftl:
**Changes**

Feature request, added original media file size on the "Stats for Nerds" (now "Playback Data") pop-up. Used base-10 converting bytes to mega/gigabytes, this may not be the same size as reported by all file managers.
Screenshot from 2019-09-02 16-28-18

**Issues**

[#1492 at jellyfin/jellyfin](https://github.com/jellyfin/jellyfin/issues/1492)


anthonylavado:
We do have the Media Info dialog in the upcoming release that shows size: https://github.com/jellyfin/jellyfin-web/pull/386

joshuaboniface:
@anthonylavado I just checked mine, and it doesn't show the original file size, just bitrate.

leftl:
@anthonylavado the feature request was specifically on the Stats for Nerds dialog for currently playing media. I did however notice within the Media Info dialog, file size is handled as base-2 and will report a different file size than proposed change. Is there an established standard?
https://github.com/jellyfin/jellyfin-web/blob/a4dabc1644ac3e4f0b242a5760698e658c91fcda/src/components/itemMediaInfo/itemMediaInfo.js#L30
@joshuaboniface thank you!

EraYaN:
That one is just incorrect, and should either be `MiB` or the factor should be `1e6`.

joshuaboniface:
Yea I'd just change it to MiB as a label and call it good. i hate 10^X representations of storage space anyways ;-)

EraYaN:
MiB for storage and bytes, MB for bandwidth and throughputs. (Resp. 1024 and 1000)

leftl:
Made this change, now showing MiB, GiB, etc. Also dropped trailing zeros from the decimal.

leftl:
@anthonylavado Thanks!
Just wanted to make a quick note, using the Stats for Nerds/Playback Stats dialog, any user can view currently playing media's file size. Within the Media Info dialog via #386, only administrators can view the file size.

anthonylavado:
@dkanada @cvium thoughts on the restriction? I seem to think you guys had an opinion on it in #386

dkanada:
I'm fine with it, the size isn't really sensitive information for somone who already has a valid access token for the server. @grafixeyehero actually removed that behavior [here](https://github.com/jellyfin/jellyfin-web/pull/444) so it should match once both are merged.

src/components/playerstats/playerstats.js
```diff
@@ -210,6 +221,13 @@ define(['events', 'globalize', 'playbackManager', 'connectionManager', 'playMeth
             });
         }
 
+        if (mediaFileSize) {
+            sessionStats.push({
+                label: 'File size:',
```
dkanada: Can you also add this to the i18n files and call `globalize.translate()` here so it can be translated to other languages?

anthonylavado:
Ah @leftl, sorry about that - small merge conflict.

leftl:
Added globalization to labels within the dialog, and fixed conflict. Thanks for the help with my first PR everyone.

src/components/playerstats/playerstats.js
```diff
@@ -196,24 +196,42 @@ define(['events', 'globalize', 'playbackManager', 'connectionManager', 'playMeth
         }
     }
 
+    function getReadableSize(size) {
+        if (size >= 1073741824) {
```
cvium: `Math.pow(1024, 3)` is a tad more readable imo

PR #446, opened 2019-09-05
Dashboard layout improvements

dkanada:


src/controllers/dashboardpage.js
```diff
@@ -12,27 +12,23 @@ define(["datetime", "events", "itemHelper", "serverNotifications", "dom", "globa
...
-            } else if (isTranscode) {
+            } else if (displayPlayMethod === "Transcode") {
                 title = globalize.translate("Transcoding");
                 text.push(globalize.translate("MediaIsBeingConverted"));
-
                 if (session.TranscodingInfo && session.TranscodingInfo.TranscodeReasons && session.TranscodingInfo.TranscodeReasons.length) {
                     text.push("<br/>");
                     text.push(globalize.translate("LabelReasonForTranscoding"));
                     session.TranscodingInfo.TranscodeReasons.forEach(function (transcodeReason) {
-                        text.push(globalize.translate("" + transcodeReason));
```
JustAMan: why change? this looks like a stringification safety net to me

src/controllers/dashboardpage.js
```diff
@@ -398,13 +395,12 @@ define(["datetime", "events", "itemHelper", "serverNotifications", "dom", "globa
...
             if (showTranscodingInfo) {
                 var line = [];
-
                 if (session.TranscodingInfo) {
                     if (session.TranscodingInfo.Bitrate) {
                         if (session.TranscodingInfo.Bitrate > 1e6) {
                             line.push((session.TranscodingInfo.Bitrate / 1e6).toFixed(1) + " Mbps");
                         } else {
-                            line.push(Math.floor(session.TranscodingInfo.Bitrate / 1e3) + " kbps");
+                            line.push(Math.floor(session.TranscodingInfo.Bitrate / 1e3) + " Kbps");
```
JustAMan: I understand that per ISO 1 Kb is 1000 bytes, but maybe move to 1 KiB = 1024 bytes here somehow?..

dkanada: Generally Kbps is used for any kind of streaming measurement which is why I kept the label the same as before.

src/controllers/dashboardpage.js
dkanada: Javascript is usually very good about casting types and I didn't notice any issues when I was testing this pull request.

dkanada:
@grafixeyehero that's definitely an issue but I didn't add that functionality. It already existed, someone just added a toggle in the user settings.

grafixeyehero:
OK Merge it

grafixeyehero:
Fix toggle https://github.com/jellyfin/jellyfin-web/pull/466

PR #449, opened 2019-09-07
Don't hide play all and shuffle buttons on mobile

mitchfizz05:
**Changes**
Play all and shuffle buttons are no longer are hidden on mobile.
It looks like this was done so it would fit better on the smaller screen, so my proposed solution is to move the pagination controls above the buttons on mobile.

Example of what it looks like

### DIFF
--- a/src/music.html
+++ b/src/music.html
@@ -3,8 +3,8 @@
     <style>
         @media all and (max-width: 32em) {
 
-            .musicglobalButton {
-                display: none;
+            .paging {
+                width: 100%;
             }
         }
     </style>


dkanada:
I agree that this is an issue, but do you have any thoughts on a more permanent solution that would work on more than one page? This might work for now but I'd like to add shuffle everywhere at some point. We also have [this](https://github.com/jellyfin/jellyfin-web/pull/210) pull request open that introduces another library button I have no idea where to place in the interface. I'd also like to move all CSS out of the HTML so that's another reason this is more of a temporary fix.

mitchfizz05:
@dkanada This is definitely more of a quick 2 line temporary fix so it's at least *possible* to shuffle on mobile.

As for a more long-term solution; having the current button bar be consistent across all tabs (kinda weird that shuffle appears on the album tab but not the songs tab), and put more obscure actions into a 3-dot menu.

Re-scan could go in there, as well as select view, filter and maybe sort by (possibly depending on screen size).  These more obscure actions could then be labelled if they're in a sub-menu to make it clearer what they do, rather than just being mystery icons.

While putting actions in menus isn't ideal, I can't think how else it could be done without cluttering the UI and getting confusing.

PR #451, opened 2019-09-07
Improve button style when focused

dkanada:


src/components/emby-button/emby-button.css
```diff
@@ -29,6 +29,13 @@
     text-decoration: none;
     /* Not crazy about this but it normalizes heights between anchors and buttons */
     line-height: 1.35;
+    transform-origin: center;
+    transition: 0.2s;
+}
+
+.emby-button.show-focus:focus {
+    transform: scale(1.4);
```
JustAMan: isn't it getting a bit too big?

src/components/emby-button/emby-button.css
```diff
@@ -145,8 +149,6 @@
 
     .paper-icon-button-light > img {
         width: 1.72em;
-        /* Can't use 100% height or it will stretch past the boundaries in safari */
```
JustAMan: why remove?

src/components/emby-button/emby-button.js
```diff
@@ -21,30 +21,23 @@ define(['browser', 'dom', 'layoutManager', 'shell', 'appRouter', 'apphost', 'css
...
         this.classList.add('emby-button');
-
         if (layoutManager.tv) {
-            if (this.getAttribute('data-focusscale') !== 'false') {
-                this.classList.add('emby-button-focusscale');
-            }
-            this.classList.add('emby-button-tv');
+            // handles all special css for tv layout
+            // the old method was element-showfocus
+            // this method utilizes class chaining instead
```
JustAMan: no use leaving comment like "old method was" - we no longer see it in the code, so there's no "old method". probably something like
```suggestion
            // handles all special css for tv layout;
            // this method utilizes css class chaining
```

JustAMan:
also @dkanada would be very helpful to see a pair of screenshots "before"-"after" to better understand what's going on

src/components/emby-button/emby-button.css
dkanada: That scale was the default value before, I just moved it here using the new method so we can more easily search and modify TV layout changes in the future.

src/components/emby-button/emby-button.css
dkanada: The item it refers to was commented out so I figured it was useless.

dkanada:
@JustAMan I really didn't change any styling, only the behavior. On the current nightly when you click a button and move your mouse the focus style continues to show until you click another button or element on the page. With these changes the focus style will only apply if the TV layout is active, thus fixing a lot of focus issues. I also disabled the focus style for disabled buttons.

dkanada:
Tab Button


Icon Button


User Settings



I didn't actually change the styles much. This will just help people search and modify the TV styling much faster. The old method was a unique class for every element and this pull request moves several elements over to a single class that can be chained with the base class to selectively apply styles.

dkanada:
Tab Button


Icon Button


User Settings



I didn't actually change the styles much. This will just help people search and modify the TV styling much faster. The old method was a unique class for every element and this pull request moves several elements over to a single class that can be chained with the base class to selectively apply styles.

PR #458, opened 2019-09-11
Refactor several files and methods

dkanada:


src/bundle.js
```diff
@@ -7,8 +7,30 @@ var _define = window.define;
 // jstree
 var jstree = require("jstree");
 require("jstree/dist/themes/default/style.css");
-_define("jstree", ["jQuery"], function() { return jstree; });
+_define("jstree", function() { return jstree; });
```
thornbill: jstree has a dependency on jQuery so I wonder if removing this could introduce some weird issues.

src/scripts/site.js
```diff
@@ -227,12 +227,6 @@ var AppInfo = {};
 !function () {
     "use strict";
 
-    function onApiClientCreated(e, newApiClient) {
-        if (window.$) {
```
thornbill: Any ideas why they were replacing the ajax method here?

src/components/inputManager.js
```diff
@@ -18,27 +17,18 @@ define(['playbackManager', 'focusManager', 'appRouter', 'dom'], function (playba
...
     function off(scope, fn) {
-
-        if (eventListenerCount) {
-            eventListenerCount--;
-        }
-
-        dom.removeEventListener(scope, 'command', fn, {
-
-        });
+        eventListenerCount--;
```
thornbill: This could cause an error if eventListenerCount is undefined.

src/components/inputManager.js
```diff
@@ -214,53 +204,9 @@ define(['playbackManager', 'focusManager', 'appRouter', 'dom'], function (playba
             case 'toggledisplaymirror':
                 playbackManager.toggleDisplayMirroring();
                 break;
-            case 'togglestats':
-                //playbackManager.toggleStats();
-                break;
-            case 'movies':
-                // TODO
-                appRouter.goHome();
```
thornbill: Why remove these that call `goHome`?

src/bundle.js
dkanada: I tested the metadata manager after removing this and didn't notice anything wrong. I think it's fine because I bundled jquery as well.

src/scripts/site.js
dkanada: Absolutely no clue, but it ran totally fine after removing these sections so I think we're in the clear.

src/components/inputManager.js
dkanada: I don't think it should since the variable is initialized right when it gets created, but I can add this back if you want. The increment section doesn't have a check.

src/components/inputManager.js
dkanada: I removed those three because it didn't seem like they served their intended purpose, which I interpreted as to go to a section with music. We haven't documented the available commands yet so my plan was to just not include these three.

joshuaboniface:
@thornbill Any follow-up thoughts?

src/components/inputManager.js
thornbill: I would add it back just to err on the side of safety.

PR #459, opened 2019-09-11
Added more details to itemdetails page

dhartung:
Useful changes for me, may be useful for others as well.

**Changes**

* When using a translated title, the original title will be shown too.
* Information about the last date media was played is added.
* Un-minified the `renderDetails`-Function a bit.

grafik


dkanada:
I really like the original name shown below the localized title, but this page will hopefully get an overhaul at some point. It might be the worst page on the website in terms of general layout and polish, for both mobile and desktop layouts.

dkanada:
Yeah, the minified code is still all over the source, but we have slowly been removing it. At some point the plan was just to run eslint but it will break `git blame` so we haven't gotten around to it yet.

PR #463, opened 2019-09-12
Changed to non-blocking subtitles

dhartung:
_Behavior before_:   
When the user starts the video playback, for every `SUBRIP` a `<track />` element was created inside the main `<video />` tag. The propper playback started after all tracks were loaded.

In my case most of the subtitles are embedded, so jellyfin extracted the subtitles on-the-fly which took (depending on the file size) quite a while - even if I don't want to display a subtitle.

--> For a few minutes the `playing` event wasn't triggered, the `videoosd` interface was not shown and depending on the webbrowser the `<video/>` element behaved very strange.

**Changes** (_Behavior after_)
No tracks are added anymore. Jellyfin uses either the custom element (Firefox) to render the subtitle, or uses the TextTrackCue method (Chrome). These are loaded asynchronos while the playback starts. A little text indicates, that additional data is loading. Only the selected subtitle is loaded.

**Notes**
I am surprised that nobody has reported this behavior as a bug yet. While waiting for subtitles the first few seconds of the video were played in permanent loop (Chrome). Was hard for me to figure out that the mistake is in the subtitles.

Actually there were 3 ways to display subtitles (excluding the server-side burning):
(1) Using the trackCueObject
(2) Using the custom `videoSubtitles` event
(3) Using native `<tracks/>`

I think due to a bug, method (1) was never used. I eliminated method (3) and stick with (1) and (2).
We could use blobs for method (3) to enable asnychronous loading, but I do not see any advantages in using this method.

But may I'm wrong and this waiting-for-subtitles behavior is intended. For my use case this behavior was ... anoying :D

### DIFF
--- a/src/components/htmlvideoplayer/plugin.js
+++ b/src/components/htmlvideoplayer/plugin.js
@@ -200,6 +200,27 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
             return self._currentSrc;
         };
 
+        self._fetchQueue = 0;
+        self.isFetching = false;
+
+        function incrementFetchQueue() {
+            if (self._fetchQueue <= 0) {
+                self.isFetching = true;
+                events.trigger(self, "beginFetch");
+            }
+
+            self._fetchQueue++;
+        }
+
+        function decrementFetchQueue() {
+            self._fetchQueue--;
+
+            if (self._fetchQueue <= 0) {
+                self.isFetching = false;
+                events.trigger(self, "endFetch");
+            }
+        }
+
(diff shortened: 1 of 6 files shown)


dhartung:
What event do you mean? In the osd interface all subtitles are listed.

A subtitle is only requested if the user selects one (either in the videoosd page or in the itemdetailpage), i.e. he can select a subtitle that haven't been fully loaded.

ThatNerdyPikachu:
Seems to work fine, good job!

ThatNerdyPikachu:
(others should probably test aswell, I only tested on one browser^^)

PR #467, opened 2019-09-14
Fix overlay header video playback

### LINKED ISSUE #454: Themes in 10.4 RC overlay video playback
**Describe the bug**

**To Reproduce**

1. Choose a theme on 10.4 RC (Haze is most noticeable however other themes like Dark also has transparency issues).
2. Play any video.

**Expected behavior**
See before picture.

**Logs**
N/A

**Screenshots**
Before:
image

After:
image 2

 - OS: Windows
 - Browser: Firefox/Vivaldi
 - Jellyfin Version: 10.4 RC

**Additional context**

grafixeyehero:
**Issues**
Fixes #454

### DIFF
--- a/src/controllers/itemdetailpage.js
+++ b/src/controllers/itemdetailpage.js
@@ -330,9 +330,9 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "cardBuild
         try {
             require(["focusManager"], function(focusManager) {
                 [".btnResume", ".btnPlay"].every(function (cls) {
-                    
+
                     var elems = page.querySelectorAll(cls);
-                    
+
                     for (var i = 0; i < elems.length; i++) {
                         if (focusManager.isCurrentlyFocusable(elems[i])) {
                             focusManager.focus(elems[i]);
@@ -919,7 +919,7 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "cardBuild
         if (!canPlaySomeItemInCollection(items)) {
             hideAll(page, "btnPlay", false);
             hideAll(page, "btnShuffle", false);
-        }   
+        }
     }
 
     function renderCollectionItemType(page, parentItem, type, items) {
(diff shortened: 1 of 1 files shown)


joshuaboniface:
@grafixeyehero The deminification caused conflicts with a previous PR. Can you remove the deminification from this PR and we'll do a full deminification soon?

grafixeyehero:
@joshuaboniface as you suggested i update the PR

src/controllers/itemdetailpage.js
```diff
@@ -1237,7 +1237,7 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "cardBuild
             view.addEventListener("viewbeforehide", function() {
                 events.off(apiClient, "message", onWebSocketMessage);
                 events.off(playbackManager, "playerchange", onPlayerChange);
-                libraryMenu.setTransparentMenu(false);
+                libraryMenu.setTransparentMenu(true);
```
JustAMan: How can this be a fix? According to `git blame` this `setTransparentMenu(false)` was so from the get-go, how can it be broken in 10.4-RC?

grafixeyehero: i think this was broken form the start.

JustAMan: If I read the bug correctly, it says that menu started being opaque after 10.3.7... How can it be broken from the beginning?..

grafixeyehero: i will check it on 10.3.7

grafixeyehero: yeah you are correct. This was removed https://github.com/jellyfin/jellyfin-web/pull/325/files

thornbill:
Proposed fix using css: https://github.com/jellyfin/jellyfin-web/pull/478

anthonylavado:
Since @thornbill has added #478, do we still want this PR? Or can both exist together?

Issue #199, opened 2019-03-17
Duplicate loading animation


platinum55:
when clicking on a movie, there are two loading animations:

http://i.imgur.com/KJkxAGM.png
http://i.imgur.com/dbjqwkz.png

- one from jellyfin
- one from the videoosd I guess

I couldn't find an html element in the player, I think it comes from the videoplayer itself.
Just a cosmetic issue.

JustAMan:
@sparky8251 does this ring a bell from your adventures in fixing the volume control?

noah2121:
+1, same issue here.

JustAMan:
I feel it should be fixed in 10.4.0 release.
@platinum55 or @noah2121 if you see it again please ping us here.

PR #468, opened 2019-09-15
Fix Duplicate loading animation

grafixeyehero:
**Issues**
Duplicate loading animation #199

### DIFF
--- a/src/components/htmlvideoplayer/plugin.js
+++ b/src/components/htmlvideoplayer/plugin.js
@@ -1398,13 +1398,10 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
                             cssClass += ' htmlvideoplayer-moveupsubtitles';
                         }
 
-                        // Can't autoplay in these browsers so we need to use the full controls, at least until playback starts
-                        if (appHost.supports('htmlvideoautoplay')) {
-                            html += '<video class="' + cssClass + '" preload="metadata" autoplay="autoplay" controls="controls" webkit-playsinline playsinline>';
-                        } else {
-
-                            // Chrome 35 won't play with preload none
+                        if (!appHost.supports('htmlvideoautoplay')) {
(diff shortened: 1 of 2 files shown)


src/components/htmlvideoplayer/plugin.js
```diff
@@ -1398,12 +1398,7 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
                             cssClass += ' htmlvideoplayer-moveupsubtitles';
                         }
 
-                        // Can't autoplay in these browsers so we need to use the full controls, at least until playback starts
                         if (!appHost.supports('htmlvideoautoplay')) {
-                            html += '<video class="' + cssClass + '" preload="metadata" autoplay="autoplay" controls="controls" webkit-playsinline playsinline>';
-                        } else {
-
-                            // Chrome 35 won't play with preload none
                             html += '<video class="' + cssClass + '" preload="metadata" autoplay="autoplay" webkit-playsinline playsinline>';
```
JustAMan: What about browsers which do have autoplay? Won't they now have no video player at all?

grafixeyehero:
What did you mean no video player at all?

grafixeyehero:
Wrong btn sorry

grafixeyehero:
Tested
Chrome 76.0.3809.132
Firefox 70.0b6

dkanada:
This adds a duplicate loading animation for firefox on my machine. I definitely want to merge this once we figure out the behavior but this doesn't seem like the right solution.

grafixeyehero:
> 
> 
> This adds a duplicate loading animation for firefox on my machine. I definitely want to merge this once we figure out the behavior but this doesn't seem like the right solution.

the first comment did'nt hide on firefox. i hope this fix https://github.com/jellyfin/jellyfin-web/pull/468/commits/20c23dc7285b831bdf130275c1e6aa760a7282dc

grafixeyehero:
@dkanada could you explain me a in which version you test and the problem that you faced. is it the controller or the loading animation

PR #470, opened 2019-09-17
Azure pipelines

thornbill:
**Changes**
Add Azure Pipelines config to run eslint and yarn package.

**Issues**
N/A


EraYaN:
Alright how do you get it to fill in the WebVersion variable?

thornbill:
@EraYaN Right now that is just set statically https://github.com/thornbill/jellyfin-web/blob/azure-pipelines/.ci/azure-pipelines.yml#L4. The tarball generated by `yarn pack` includes the version number from `package.json`. We aren't currently tracking the web version there so it just stays `0.0.0`. If you have any better ideas on how to pull that value, I am definitely open to that.

EraYaN:
I mean you can just upload it with a wild card `*` ;) I do that with the setup as well.

thornbill:
Right that makes perfect sense... let me fix that.

.npmignore
```diff
@@ -1,7 +1,9 @@
+.editorconfig
 .gitattributes
 .github
 .drone.yml
 .eslintrc.yml
+azure-pipelines.yml
```
EraYaN: This should be `.ci/` maybe?

thornbill: Yeah I missed that when I moved it.

thornbill:
Well it doesn't look like wildcards are supported in the `targetPath` for publishing artifacts. May need to add an intermediate copy task.

EraYaN:
The setup is moved to the Artifacts Directory and then the whole thing is uploaded.

anthonylavado:
This good?

thornbill:
Yeah it should be all good.

PR #472, opened 2019-09-19
Remove WAN domain name options

sparky8251:
This was used in generation and modification of certificates, but since JF doesn't generate them and shouldn't modify them, I have removed these.

There is a lot to untangle with regard to JF at least attempting to modify certs given to it, but this is a start.

Merge conflicts are likely easily resolvable. Caused by me removing unused translations.


sparky8251:
Should be merged with https://github.com/jellyfin/jellyfin/pull/1773

PR #473, opened 2019-09-19
Add better "Date *" base strings

ThatNerdyPikachu:


dkanada:
Where are these used in the web client?

ThatNerdyPikachu:
See #459.

thornbill:
It looks like these are also used for sorting options (See https://github.com/jellyfin/jellyfin-web/blob/82f171e99a705b6deb46805ff2332363cfe23f7b/src/components/tabbedview/itemstab.js#L442). We should either add new translation strings or append the colon in the item details view to avoid impacting the other places these translations are used.

joshuaboniface:
I agree with @thornbill here, the colons belong in the actual string read, rather than in the message text. This just complicates translations.

dkanada:
@ThatNerdyPikachu did you want to revert these strings and add new ones?

Issue #453, opened 2019-09-03
Volume slider generates hundreds of Playback requests


sachk:
**Describe the bug**
A progress request is sent by the browser for each fractional movement of the volume bar. This blocks requests for new HLS chunks, slowing down playback.

**To Reproduce**

1. Play anything
2. Open developer tools
3. Move the volume bar
4. See all the progress requests being made

**Expected behavior**
Progress requests are only sent when the volume control stops being moved or aren't sent when volume is changed (I can't think of why the server needs to know the clients volume, but perhaps it's necessary)

**Screenshots**
[Gif of the problem (had to censor personal information)
](https://gfycat.com/inexperiencedtamebelugawhale)

**System (please complete the following information):**
 - OS: Debian 10
 - Browser: Chrome
 - Jellyfin Version: 10.3.7
 - Reverse proxy: Built in SSL Support

redSpoutnik:
This is caused by "mousemove" and "touchmove" volumeSlider's event listeners in videoosd.js, they are triggered for every increment when moving the slider.

I don't know why those two listeners have been set.

I tested to remove them and the "change" listener was enough (triggered only on mouse or touch release) on Chrome, Firefox and Brave.

anthonylavado:
Cc @dkanada, I remember he had some thoughts on why this was the case

dkanada:
This was last modified by @sparky8251 because he felt the change listener wasn't responsive enough, which is a valid point. The two solutions are to either revert back to change or add a time delta check to limit the requests to one per second or something.

PR #474, opened 2019-09-21
add 700ms timer for 'mousemove' and 'touchmove' volume listeners

redSpoutnik:
This PR is a proposal for #453 (Volume slider generates hundreds of Playback requests) : 

The 'mousemove' and 'touchmove' volume slider listeners make a server request for each increment when moving the slider. 
This PR set a timer on those two listeners to limit the number of requests.

**Changes**

* When 'mousemove' or 'touchmove' listeners are triggered, a new request is sent to the server only if 700ms elapsed since the last one.
* The timer's limit of 700ms is arbitrary (same as onTimeUpdate(e) limit in videoosd.js). 
* 700ms seems to me a good compromise between responsiveness and number requests limitation, but it could be changed.

**Issues**

Possibly #453

### DIFF
--- a/src/controllers/videoosd.js
+++ b/src/controllers/videoosd.js
@@ -1209,7 +1209,7 @@ define(["playbackManager", "dom", "inputmanager", "datetime", "itemHelper", "med
         var programEndDateMs = 0;
         var playbackStartTimeTicks = 0;
         var subtitleSyncOverlay;
-        var volumeSliderTimer = 0;
+        var volumeSliderTimer;
         var nowPlayingVolumeSlider = view.querySelector(".osdVolumeSlider");
         var nowPlayingVolumeSliderContainer = view.querySelector(".osdVolumeSliderContainer");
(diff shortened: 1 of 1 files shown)


thornbill:
It might be better to use setTimeout/clearTimeout for this so we only send the last event triggered after a certain delay. I think that is a more standard way to handle debouncing an event like this.

redSpoutnik:
Thanks for advice @thornbill, the behavior is now slightly different: 
* On 'mousemove' or 'touchmove' event a new timer is created (except if there is already one running) and after 700ms a new request is send with current volume slider value.
* On 'change' event (slider click, slider start moving or stop moving), if a timer is currently running it will be destroyed.

dkanada:
@thornbill if I recall correctly, since the volume slider isn't a component this won't fix the issue for the other two instances in the music controls or remote control page. We will have to move all three to a component eventually to fix the fragmentation issue.

PR #477, opened 2019-09-24
Fix mediainfo display on tv mode

grafixeyehero:
**To Reproduce**
1. Set a Display Mode 'Tv' on user setting page
2. Go to 'detail page.'
3. Click on more option button and click 'More Info'
4. See

### DIFF
--- a/src/components/themes/appletv/theme.css
+++ b/src/components/themes/appletv/theme.css
@@ -462,9 +462,3 @@ html {
 .metadataSidebarIcon {
     color: #00a4dc
 }
-
-.mediaInfoContent {
-    margin-left: auto;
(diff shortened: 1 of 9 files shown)


grafixeyehero:
it work for both of them

src/components/themes/appletv/theme.css
```diff
@@ -462,3 +462,9 @@ html {
 .metadataSidebarIcon {
     color: #00a4dc
 }
+
+.mediaInfoContent {
```
dkanada: I would remove this from the theme styles and add it to `site.css` instead.

grafixeyehero: done

PR #479, opened 2019-09-25
Improve the user settings interface

dkanada:
The first commit removes the section below and replaces it with some buttons, the old process was awkward to get through. I'll add a few more commits that change the headers and remove old strings when I have time.

remove


JustAMan:
@dkanada can you show "before" and "after"? I cannot grasp the pic you show, is it the old state or the proposed?

dkanada:
The photo in my last comment is before the change, I just replaced it with a single button to add the photo. It looks like the `Scan All Libraries` button in the dashboard.

before
after

src/strings/de.json
```diff
@@ -1296,7 +1290,7 @@
     "ButtonTrailer": "Trailer",
     "ButtonWebsite": "Webseite",
     "ChangingMetadataImageSettingsNewContent": "Änderungen an Metadaten- und Artwork-Einstellungen betreffen nur der Bibliothek neu hinzugefügte Inhalte. Um diese Änderungen auf existierende Medien anzuwenden müssen die Metadaten manuell aktualisiert werden.",
-    "CopyStreamURL": "Kopiere Streaminglink",
+    "CopyStreamURL": "Stream URL kopieren",
     "CopyStreamURLSuccess": "Link erfolgreich kopiert.",
```
thornbill: Is this an intentional change to the translation?

dkanada: Apparently someone added the key twice in this file, so my script replaced this instance with a value from the same key at the end of the file.

PR #482, opened 2019-09-26
Fix Footer styles

grafixeyehero:
Before
Screenshot_2019-09-27 Jellyfin(2)
Screenshot_2019-09-27 Jellyfin
Screenshot_2019-09-27 Jellyfin(3)

After
Screenshot_2019-09-27 Jellyfin(8)
Screenshot_2019-09-27 Jellyfin(9)
Screenshot_2019-09-27 Jellyfin(7)

### DIFF
--- a/src/components/cardbuilder/card.css
+++ b/src/components/cardbuilder/card.css
@@ -271,6 +271,10 @@ button {
     position: relative;
 }
 
+.visualCardBox-cardFooter {
+    height: 5.5em;
+}
+
 .visualCardBox {
     box-shadow: 0 0.0725em 0.29em 0 rgba(0, 0, 0, 0.37);
     border-radius: .145em;
--- a/src/components/cardbuilder/card.css
+++ b/src/components/cardbuilder/card.css
@@ -267,13 +267,11 @@ button {
 
 .cardFooter {
     padding: .3em .3em .5em .3em;
-    height: 3.5em;
+    min-height: 3.5em;
     position: relative;
 }
 
-.visualCardBox-cardFooter {
-    height: 5.5em;
-}
+
 
 .visualCardBox {
     box-shadow: 0 0.0725em 0.29em 0 rgba(0, 0, 0, 0.37);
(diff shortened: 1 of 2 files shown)


dinki:
Looked better in the before images, IMO

joshuaboniface:
I agree with @dinki, for entries that don't overflow that extra space looks really ugly. Can we make each card dynamically sized to its content? i.e. only add the additional line if the line is actually present? In any case though, I think the original is too tight on the bottom, maybe another 2em of spacing after the text would look much better.

src/components/cardbuilder/card.css
```diff
@@ -271,6 +271,10 @@ button {
     position: relative;
 }
 
+.visualCardBox-cardFooter {
+    height: 5.5em;
```
thornbill: Instead of just setting a larger value for the height of the footer and hoping the content will fit within the larger height, perhaps we should just change the `height: 3.5em;` of `.cardFooter` to a `min-height`. That will allow the footer to expand dynamically based on the size of the content.

thornbill: Screenshot_2019-09-27 Jellyfin(2)

dkanada: Not a fan of the different sizes. I think the original method works fine until we can figure out a way to make these look half decent. They're absolutely terrible right now.

thornbill: I agree it would be nice if they were a consistent size, but this change doesn't _fix_ the problem here. It only makes it less likely to happen, if the footer content is too large (taller than `5.5em`) it will still overflow the container. Even if we increase the size, we should still use `max-height` to avoid this or hide the overflow. To properly fix this so all cards are equal in size even if the content of one is larger would probably need a more extensive overhaul.

grafixeyehero: @thornbill done

dkanada:
merged #484 instead for now since we couldn't find a complete solution

PR #491, opened 2019-09-30
Bundle flvjs and shaka and move all libs to one folder

dkanada:


src/bundle.js
```diff
@@ -21,6 +25,10 @@ _define("hlsjs", function() { return hlsjs; });
 var howler = require("howler");
 _define("howler", function() { return howler; });
 
+// shaka
+var shaka = require("shaka-player");
+_define("shaka", function() { return shaka });
```
JustAMan:
```suggestion
_define("shaka", function() { return shaka; });
```

so that style is consistent 😄

src/components/viewContainer.js
```diff
@@ -44,11 +44,11 @@ define(["browser", "dom", "layoutManager", "css!components/viewManager/viewConta
 
 
             if (isPluginpage) {
-                require(["legacy/dashboard"]);
+                require(["legacyDashboard"]);
```
JustAMan: could you explain this?

src/controllers/user/menu.js
```diff
@@ -14,7 +14,7 @@ define(["apphost", "connectionManager", "listViewStyle", "emby-button"], functio
             page.querySelector(".lnkMyProfile").setAttribute("href", "myprofile.html?userId=" + userId);
             page.querySelector(".lnkDisplayPreferences").setAttribute("href", "mypreferencesdisplay.html?userId=" + userId);
             page.querySelector(".lnkHomePreferences").setAttribute("href", "mypreferenceshome.html?userId=" + userId);
-            page.querySelector(".lnkLanguagePreferences").setAttribute("href", "mypreferenceslanguages.html?userId=" + userId);
+            page.querySelector(".lnkPlaybackPreferences").setAttribute("href", "mypreferencesplayback.html?userId=" + userId);
```
JustAMan: eh? why?

src/scripts/site.js
```diff
@@ -309,6 +311,10 @@ var AppInfo = {};
         return "bower_components";
     }
 
+    function getComponentsPath() {
```
JustAMan: why is this needed?

src/scripts/site.js
```diff
@@ -381,14 +387,13 @@ var AppInfo = {};
...
             });
         } else {
-            define("ResizeObserver", ["thirdparty/resize-observer-polyfill/ResizeObserver"], returnFirstDependency);
+            define("ResizeObserver", [getBowerPath() + "/resize-observer-polyfill/ResizeObserver"], returnFirstDependency);
         }
     }
 
     function initRequireWithBrowser(browser) {
         var bowerPath = getBowerPath();
-        var apiClientBowerPath = bowerPath + "/apiclient";
```
JustAMan: why remove that and copy-paste `/apiclient` all over the place?

src/scripts/site.js
```diff
@@ -706,35 +680,87 @@ var AppInfo = {};
...
+        define("appFooter-shared", ["appFooter"], createSharedAppFooter);
+
+        // TODO pull apiclient out of this repository
+        define('events', [bowerPath + "/apiclient/events"], returnFirstDependency);
+        define('credentialprovider', [bowerPath + "/apiclient/credentialprovider"], returnFirstDependency);
+        define('connectionManagerFactory', [bowerPath + "/apiclient/connectionmanager"], returnFirstDependency);
+        define('appStorage', [bowerPath + "/apiclient/appStorage"], returnFirstDependency);
+        define("serversync", [bowerPath + "/apiclient/sync/serversync"], returnFirstDependency);
+        define("multiserversync", [bowerPath + "/apiclient/sync/multiserversync"], returnFirstDependency);
+        define("mediasync", [bowerPath + "/apiclient/sync/mediasync"], returnFirstDependency);
```
JustAMan: are these three `-sync` things needed?

src/components/viewContainer.js
dkanada: I'm just consolidating the legacy script locations in `site.js` so we have a central location for items we need to remove eventually.

src/controllers/user/menu.js
dkanada: This is the page that also controls other aspects of media playback like bitrate and skip length, among other things.

src/scripts/site.js
dkanada: I was just matching the `getBowerPath()` method above, we might want to eventually remove this but right now the path is set in multiple locations.

src/scripts/site.js
dkanada: I did this because the goal is to eventually remove `bowerPath` entirely and pull the dependencies in during the build, and consolidating all of the libraries to remove under `bowerPath` helps with searching for now.

src/scripts/site.js
dkanada: I tried removing some of these but was getting a ton of errors, I'll look into it more once we bundle the API client during build.

PR #499, opened 2019-10-03
Change Date Added string for Item Details

ThatNerdyPikachu:


src/strings/en-us.json
```diff
@@ -165,6 +165,8 @@
     "CustomDlnaProfilesHelp": "Create a custom profile to target a new device or override a system profile.",
     "DateAdded": "Date added",
     "DatePlayed": "Date played",
+    "DateAddedColon": "Date added:",
```
dkanada: It looks like you didn't use this string anywhere.

ThatNerdyPikachu: Added for redundancy, in case Date Added was used anywhere we want a colon as well.

joshuaboniface: Why not just put the colon in the in-code string where required? It seems weird to have a single string that exists just to "have a colon at the end".

ThatNerdyPikachu: That works.

PR #501, opened 2019-10-04
Resolves missing policy error on user creation

anthonylavado:
Restores the previous method of posting the Users/New request, but adds the password field as well.

**Note**
I'm sure this isn't right, but this does fix the problem here.

### DIFF
--- a/src/bower_components/apiclient/apiclientcore.js
+++ b/src/bower_components/apiclient/apiclientcore.js
@@ -1195,7 +1195,7 @@ define(["events", "appStorage"], function(events, appStorage) {
         return this.ajax({
             type: "POST",
             url: url,
-            data: {
+            data: user,
                     Name: user.Name,
                     Password: user.Password
                 },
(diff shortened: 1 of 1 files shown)


src/bower_components/apiclient/apiclientcore.js
```diff
@@ -1195,8 +1195,11 @@ define(["events", "appStorage"], function(events, appStorage) {
         return this.ajax({
             type: "POST",
             url: url,
-            data: JSON.stringify(user),
-            contentType: "application/json"
+            data: {
```
thornbill:
```suggestion
            data: user,
```

src/bower_components/apiclient/apiclientcore.js
```diff
@@ -1195,8 +1195,11 @@ define(["events", "appStorage"], function(events, appStorage) {
         return this.ajax({
             type: "POST",
             url: url,
-            data: JSON.stringify(user),
-            contentType: "application/json"
+            data: user,
+                    Name: user.Name,
```
thornbill: Oh the Github selection didn't let me select multiple lines. Lines #1199-1201 should be removed.

anthonylavado: Whoops, missed that

PR #502, opened 2019-10-04
Update Docs to docs.jellyfin.org URL

joshuaboniface:
**Changes**
As per discussions on Riot, pull docs out of main site and create a separate site for them. This just updates the URL for now, work required on the infra side to actually support this.

**Issues**
N/A


joshuaboniface:
HAProxy redirects have been added for the subdomain so no functional change is required for now.

PR #484, opened 2019-09-29
Revert "Allow card text to continue through line breaks"

thornbill:
**Changes**
Reverts changes to allow multiline text in card footers.

**Issues**
Fixes #485

PR #503, opened 2019-10-06
Fix some minor issues with the dashboard page

dkanada:
The icon was showing up too small in the dashboard, this just makes it the right size.


thornbill:
Your second commit seems unrelated to the icon. What is that changing?

dkanada:
The playback progress element itself needed to be updated, the old behavior was inserting a copy inside the progress bar.

joshuaboniface (approved):

JustAMan (approved):

miguelaguiar01:
You are right, and the probe you built is the case I should have thought of: Outline's own first push is `added: 1919, seeded: 0, seedsIdentical: 12466` for a language whose file is full of the English text, and my gate silenced exactly the line that explains why that language reads nought.

Taken as you proposed. The clause now rides on the push having done something (`added || changed || archived || seeded`), and the seeded-nothing branch reads as the catalogue's state, never as work: `1920 added, 0 changed, 0 stale, 0 archived, 13 repository translation(s) kept as Corpus has them; 3840 repository translation(s) identical to the source, kept untranslated`. `0 translation(s) seeded` cannot be printed any more. A 0/0/0/0 push stays quiet, which is all #484 asked for. Your case is now a test, so the hole is no longer invisible to the suite.

Spec §8's seed paragraph says the same three things. The changelog says what the line does on a first push, and its stray blank line is gone. `--dry-run` shares the path deliberately: a dry run of a first push should explain the same rows, and the label already says it is a dry run.

PR #508, opened 2019-10-08
Rearrange and add metadata button in top-right

joshuaboniface:
**Changes**
1. Adds a Metadata editor button on the top right menu, to match the dashboard button. Avoids this visually being hidden away in the user menu. IMO if we have a dashboard button here, this probably should be too.
2. Move the User button to the end. IMO this is visually nicer as it's the "main" button.

**Issues**
N/A

**Visual**

Screenshot from 2019-10-08 19-07-58


dkanada:
I'm not a fan of this to be honest. Four icons was already crowded from a UX standpoint and I was planning on removing the dashboard icon and leaving it in the user settings menu and navigation bar.

joshuaboniface:
I just want the two to be consistent and obvious. Removing the other button would accomplish that.

dkanada:
If I open another PR that removes the dashboard button instead would that be alright for the 10.4.1 release? I have a bit of spare time this weekend.

dkanada:
closing in favor of #551

PR #521, opened 2019-10-12
Fix save button in now playing controller

dkanada:
This was caused by a mistake in an old deminification attempt. I tested the fix and saving the queue to a playlist is working again.


joshuaboniface:
Cherry-picked into 10.4.1.

PR #522, opened 2019-10-12
Tizen

dmitrylyzo:
These are the changes I have to make to get Jellyfin to work (tolerably) on Samsung TV with Tizen. Tizen App itself is on [.](https://github.com/dmitrylyzo/jellyfin-tizen) Maybe it is better to split my work into small pull requests. There are still problems that cannot be solved without you.
Sorry for asking you such way.


src/components/scrollManager.js
```diff
@@ -0,0 +1,246 @@
+define([], function () {
```
JustAMan: is this a new file, or is it some 3rd-party library?

src/components/scrollManager.js
```diff
@@ -0,0 +1,246 @@
...
+    var scrollElementRect;
+    var scrollTarget = {};
+    var scrollTimer;
+    var oldActiveElement;
+
+    function clamp(min, value, max) {
+        return value < min ? min : value > max ? max : value;
+    }
+
+    function nearestScroll(v1, v2, min, max) {
```
JustAMan: add some comment explaining what this function does please

src/components/scrollManager.js
```diff
@@ -0,0 +1,246 @@
...
+        if (delta1 < 0 && delta1 < delta2) {
+            return delta1;
+        } else if (delta2 < 0) {
+            return -delta2;
+        }
+        return 0;
+    };
+
+    function trend(v, v1, v2) {
+        return (v1 < v && v2 < v1) || (v1 > v && v2 > v1) ? v2 : v1;
```
JustAMan: I cannot grasp what this function does, please explain (or better expand it so it's clearer).
Or I might understand, but if I do then this function has a bug :) so an explanation is needed to understand whether I misunderstand or this has a bug.

src/components/scrollManager.js
```diff
@@ -0,0 +1,246 @@
...
+        return scrollBy(options.left, options.top, options.behavior === 'smooth');
+    };
+
+    function resetScroll() {
+        cancelAnimationFrame(scrollTimer);
+        scrollTimer = undefined;
+    }
+
+    function scrollTo(x, y, smooth) {
+        //console.log('scrollTo (' + x + ', ' + y + ', ' + smooth + ')');
```
JustAMan: don't commit "development debug" comments please

src/components/scrollManager.js
```diff
@@ -0,0 +1,246 @@
...
+        if (scrollTimer) {
+            scrollTarget.x = trend(window.scrollX, scrollTarget.x, x);
+            scrollTarget.y = trend(window.scrollY, scrollTarget.y, y);
+        } else {
+            scrollTarget.x = x;
+            scrollTarget.y = y;
+        }
+
+        var scrollMaxX = Math.max(0, Math.max(document.body.scrollWidth, document.body.offsetWidth, document.documentElement.clientWidth, document.documentElement.scrollWidth, document.documentElement.offsetWidth) - window.innerWidth);
+        var scrollMaxY = Math.max(0, Math.max(document.body.scrollHeight, document.body.offsetHeight, document.documentElement.clientHeight, document.documentElement.scrollHeight, document.documentElement.offsetHeight) - window.innerHeight);
```
JustAMan: please change those lines into multiline statements, otherwise reading them is painfully hard

src/components/scrollManager.js
```diff
@@ -0,0 +1,246 @@
...
+
+        scrollTarget.x = clamp(0, scrollTarget.x, scrollMaxX);
+        scrollTarget.y = clamp(0, scrollTarget.y, scrollMaxY);
+
+        if (scrollTarget.x === window.scrollX && scrollTarget.y === window.scrollY) {
+            //console.log('cancel scroll animation');
```
JustAMan: again dev-debug comment

src/components/scrollManager.js
```diff
@@ -0,0 +1,246 @@
...
+            resetScroll();
+            return true;
+        }
+
+        //console.log('scroll from (' + window.scrollX + ', ' + window.scrollY + ') to (' + scrollTarget.x + ', ' + scrollTarget.y + ')');
```
JustAMan: again dev-debug

src/components/scrollManager.js
```diff
@@ -0,0 +1,246 @@
...
+
+        if (!scrollTimer) {
+            var start;
+
+            function scrollAnim(currentTimestamp) {
+                start = start || currentTimestamp;
+
+                var dx = scrollTarget.x - window.scrollX;
+                var dy = scrollTarget.y - window.scrollY;
+                //console.log('> ' + dx + ', ' + dy + ' = ' + window.scrollX + ', ' + window.scrollY);
```
JustAMan: again dev-debug... I'll stop posting these, please clean up all dev-debug comments

src/components/scrollManager.js
```diff
@@ -0,0 +1,246 @@
...
+
+        scrollX = scrollX || 0;
+        scrollY = scrollY || 0;
+
+        // scroll to element container instead
+        elem = elem.closest('.focuscontainer-scroll, .inputContainer, .selectContainer, .checkboxContainer, .paperList') || elem;
+
+        //console.log('scrollIntoView (' + elem.tagName + ', ' + scrollX + ', ' + scrollY + ')');
+
+        if (elem !== scrollElement) {
```
JustAMan: what is the reason to make this a global variable?

src/components/scrollManager.js
```diff
@@ -0,0 +1,246 @@
...
+
+        return false;
+    }
+
+    function scrollBy(dx, dy, smooth) {
+        //console.log('scrollBy (' + dx + ', ' + dy + ', ' + smooth + ')');
+        return scrollTo(window.scrollX + dx, window.scrollY + dy, smooth);
+    }
+
+    function scrollIntoView(elem, scrollX, scrollY) {
```
JustAMan: a comment explaining what this function should do and its parameters would be nice

src/components/scrollManager.js
```diff
@@ -0,0 +1,246 @@
...
+        }
+
+        var windowRect = getWindowRect();
+
+        return ((scrollX < 0 && rect.left >= windowRect.left) ||
+                (scrollX > 0 && rect.right <= windowRect.right) ||
+                (scrollX === 0 && rect.left >= windowRect.left && rect.right <= windowRect.right)) &&
+            ((scrollY < 0 && rect.top >= windowRect.top) ||
+                (scrollY > 0 && rect.bottom <= windowRect.bottom) ||
+                (scrollY === 0 && rect.top >= windowRect.top && rect.bottom <= windowRect.bottom));
```
JustAMan: pls add a comment explaining this giant condition

src/components/scrollManager.js
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    function prepare() {
```
JustAMan: when should this be called?

src/controllers/addserver.js
```diff
@@ -46,6 +46,16 @@ define(["appSettings", "loading", "browser", "emby-button"], function(appSetting
         view.querySelector(".addServerForm").addEventListener("submit", onServerSubmit);
         view.querySelector(".btnCancel").addEventListener("click", goBack);
 
+        // HACK: To catch the end of loading
```
JustAMan: eh, what this changes?

src/controllers/videoosd.js
```diff
@@ -1071,6 +1071,12 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
         }
 
         function onWindowKeyDown(e) {
+            // FIXME: Conflicts with keyboard navigation
+            // FIXME: Then the keyboard is completely ignored. Need another solution.
```
JustAMan: this doesn't sound pretty... I would rather we have a proper navigation for all modes - AFAIK TV remote navigation is very similar to normal keyboard, but maybe it just uses another key codes - those should be extracted in a single file and defined there depending on context then

src/components/emby-tabs/emby-tabs.js
```diff
@@ -18,13 +18,13 @@ define(['dom', 'scroller', 'browser', 'layoutManager', 'focusManager', 'register
 
     function onFocus(e) {
 
-        if (layoutManager.tv) {
+        /*if (layoutManager.tv) {
```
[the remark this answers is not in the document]
dmitrylyzo: I turned it off because when I focus on the profile settings in the top menu using the keyboard, this code activates the last tab. So, I leave activation by click only.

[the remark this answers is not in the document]
JustAMan: It is fine to remove old code, sure, but you should do so by _deleting_ that code, not commenting it out.

src/components/focusManager.js
```diff
@@ -38,13 +44,22 @@ define(['dom'], function (dom) {
...
         } catch (err) {
             console.log('Error in focusManager.autoFocus: ' + err);
         }
+
+        // Restore scrolling - Tizen does not support 'preventScroll' option for 'focus'
+        if (browser.tizen) {
+            window.scroll(currentScroll.x, currentScroll.y);
+        }
+
+        scrollManager.scrollIntoView(element);
```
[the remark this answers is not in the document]
dmitrylyzo: It seems that Tizen 4.0 does not support smooth scrolling and ignores 'preventScroll' option. Instant scroll on focus looks ugly. So, I wrote scroll manager to handle all scrolling tasks, set 'preventScroll' to true again and block scrolling by reverting window scroll for Tizen.

src/components/focusManager.js
```diff
@@ -131,6 +146,18 @@ define(['dom'], function (dom) {
...
 
+        var elementRect = getOffset(elem);
+
+        // not currently visible
+        if (!elementRect.width && !elementRect.height) {
+            return false;
+        }
+
+        // out of view
+        if (viewport !== false && (elementRect.right < 0 || elementRect.bottom < 0 || elementRect.right >= window.innerWidth || elementRect.top >= window.innerHeight)) {
```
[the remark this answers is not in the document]
dmitrylyzo: By default it should test viewport boundaries (means 'not explicitly false'), so we cannot focus on element out of view (not-seen), otherwise the screen will jump around. But if we need to focus on element out of view (for example scroll up to default button) then caller pass 'false' explicitly. I used this once at viewManager.onViewChange.

src/controllers/videoosd.js
dmitrylyzo: Codes of navigation buttons the same as keyboard arrow keys ones. When I press a key on the remote control, it controls a playback, and I cannot navigate in video player toolbar. Actually, it navigate and rewind/forward in parallel. For playback control I use media buttons on the remote control. For now, I have no idea how it should be. So this needs to be discussed. Maybe playback control must be active when focused on video player itself with hidden toolbar. And we have to press some key to show up the osd.

src/components/scrollManager.js
dmitrylyzo: It is new (self-written).

src/components/scrollManager.js
dmitrylyzo: To ensure the element rect is retrieved only once (on change scroll target). There is a call of scrollIntoView in scrollInDirection to ensure we are done and may scroll by increment. Rect can be retrieved each time, but then (as I remember) the scroll manager have extra scroll on scaled elements after 'we are done'.

src/controllers/addserver.js
dmitrylyzo: At dedicated Tizen app project (dmitrylyzo/jellyfin-tizen) I catch 'end of loading' by watching for loading indicator 'hide' style class. Actually it is better to have some event from viewManager to call my onPageLoad.

dmitrylyzo:
Will adding cleanup commits be enough? Or should I redo the branch with your suggestions?

src/components/focusManager.js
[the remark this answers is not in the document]
JustAMan: This doesn't sound like a good variable name and default value to me... a default value for omitted variable is `undefined` which is falsy, so I propose to name it `forceScroll` and set it to `true` when needed

src/components/focusManager.js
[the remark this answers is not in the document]
JustAMan: Did you check that it doesn't prevent scrolling on non-Tizen then?

src/components/scrollManager.js
JustAMan: can we encapsulate this element and "prepare, now scroll" logic into a single class?.. I don't like global variables much...

src/controllers/addserver.js
JustAMan: mhm, I see... can you introduce a new event in this case instead of hijacking the loading indicator?..

JustAMan:
> Will adding cleanup commits be enough? Or should I redo the branch with your suggestions?

As I stated before I would prefer to see this in a series of PRs, each one adding a single feature, like "adding scroll manager", "fixing keyboard controlling in X part", etc., so we can potentially rollback or re-test a single piece of this big set of changes.

P.S. Thanks for working on enabling Tizen, I personally is much interested as I have a Tizen of 2015 year as one of my devices :)

src/components/focusManager.js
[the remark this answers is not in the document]
dmitrylyzo: ScrollManger was not tested on non-Tizen devices.
Originally it was in the Tizen app itself (adapter) while I played with it on TV. Keyboard navigation and scroll on focus was in the Tizen app too. Then jellyfin-web got keyboardnavigation and scroll on focus (preventScroll=false), that broke my app. Since I have to make a lot of changes to jellyfin-web, I decided to integrate scrollManager and some code of Tizen app to jellyfin-web and share it with other TVs (as I thought) to have the same scroll behavior.
Is keyboard navigation expected in the browser (not app) on the TV?

dkanada:
@dmitrylyzo we will accept any changes as long as they don't affect the web client in a negative way for more common clients such as web browsers. Hacks are also generally undesireable unless they are relatively atomic, but there are exceptions.

src/components/focusManager.js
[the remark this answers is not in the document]
JustAMan: Ideally web UI should work mostly the same on a browser opened in a TV, plus keyboard navigation on non-TVs should also work (eventually, not sure if it works okay now - I think we had reports that it behaves strangely).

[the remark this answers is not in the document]
dmitrylyzo: Have tested ScrollManager in browser on TV (Samsung UE65NU7400) - works same way (not taking into account loosing of the focus due to missing my app wrapper).

Also have installed webOS TV emulator. Have tested in jellyfin-webos app (it is almost "clear" browser, i.e. with no extra code) and have found out that webOS also ignores 'preventScroll=true'. Have blocked scroll with window.scroll as well, and ScrollManager started to work same way.

ScrollManager has an issue with scrolling of itemcontainer (for example, at "Movies/Recommended" page), possibly because I scroll the window but not the container itself. One way is to wrap all single row containers with emby-scroller (it is "self-scrolled"). Other is to try to scroll parent container with ScrollManager.

I began to split the changes into smaller parts, testing them separately.

dmitrylyzo:
Do you require string to be double-quoted? Need to know to commit with valid style.

dkanada:
We use double quotes for strings unless single quotes are more convenient or required for a certain instance.

dkanada:
Since we are splitting this PR into several smaller ones I think this can be closed. As soon as we finish merging in the deminification PRs I'll focus more on your open changes @dmitrylyzo :) we have a backlog right now.

PR #523, opened 2019-10-12
Add icon on item context menu

grafixeyehero:
After
Screenshot_2019-10-12 Jellyfin


src/components/itemcontextmenu.js
```diff
@@ -13,14 +13,16 @@ define(['apphost', 'globalize', 'connectionManager', 'itemHelper', 'appRouter',
             if (options.play !== false) {
                 commands.push({
                     name: globalize.translate('Play'),
-                    id: 'resume'
+                    id: 'resume',
+                    icon: "play_arrow"
```
JustAMan: can we use consistent quotes please (here and in other places)?

grafixeyehero: done

joshuaboniface:
Cherry-picked into 10.4.1.

Issue #66, opened 2018-12-20
Live TV display in web GUI lacks pagination


dinki:
This is minor but thought I'd report it.

The Live TV display of upcoming Movies and Shows in the web GUI lacks pagination and only the first page is shown.  This can be reviewed by clicking on:

Live TV->Programs->Movies
or
Live TV->Programs->Shows

Expected behavior would be to show all items separated by page with page navigation at top, bottom or both.

PrplHaz4:
I have confirmed this bug is still present for me in 10.3.2

dinki:
bump.  Still needs to be fixed

MrTimscampi:
Looking into this, it's because this page is treated as a suggestion page.

As such, it only gets a few items (9 or 12 -or 14? For some reason it adds 2 more when it's 12 items-). 

It would be trivial to make it respect the page limit settings, but adding proper pagination would be a bit more involved,

MrTimscampi:
As this is a feature request and not a bug, please redirect this to https://features.jellyfin.org/

dinki:
I disagree.  The user is only able to see the first page.  In my case, only programs that start with a numeric or the letter 'a' are shown.  Anything 'b-z' is not accessible.

While I understand that live tv is not a big priority for most it does not change the fact that this is indeed a bug.

MrTimscampi:
This has nothing to do with Live TV being priority or not.

The page is a suggestions page and not a library one. We currently don't support pagination on suggestion pages, so this is a feature and not a bug. As such, it goes on https://features.jellyfin.org/

We are cleaning up issues because we have/had a lot of feature requests and suggestions among legitimate issues, making it hard to track what is REALLY an issue for people.

DMouse10462:
I think there's a distinction to be made between "operating as expected" from a written code standpoint vs. a UI/UX standpoint. This is a frontend application, and I would expect that a frontend application treats design and implementation inconsistencies as bugs from a UI/UX standpoint because the view (a list of all programs listed as Movies/TV Shows) is expected (based on the other cardviews) to do something (list all of its contents with pagination/lazy loading/etc.) that isn't happening. The code may be operating as expected from what you see in the .js file, but a user looking at the view would argue otherwise.

In a user interface, not providing the necessary UI in a consistent manner is, in my argument, a bug.

PR #524, opened 2019-10-13
Fix minor style on home section Live TV

grafixeyehero:
Issue
https://github.com/jellyfin/jellyfin-web/issues/66

Before
Screenshot (2)

After
Screenshot (1)


JustAMan:
I think @dkanada said that padding was needed for mobile

joshuaboniface:
Cherry-picked into 10.4.1.

PR #533, opened 2019-10-16
Tizen: Remove tab activation by focus

dmitrylyzo:
When navigating the top menu (on TV), the tab is activated by focus, sometimes losing focus. This behavior has been removed to make the top menu navigation "smoother".


dmitrylyzo:
Found out that in "src/components/serverNotifications/mouseManager.js" focus on mouseover is enabled for Tizen (but disabled for webOS). Perhaps this is somewhat related (the same concept) to the activation of tabs by focus.

PR #534, opened 2019-10-16
Set "Resume" action on season item on TV

dmitrylyzo:
Made to be able to play/resume an episode of a TV season on the TV ^_^. Prior to this, there was no action at all. I have doubts about where the action is specified, but it works.


dkanada:
Does this resume the first unwatched episode or only the first episode in a season?

dmitrylyzo:
This resumes the selected episode (immediately, like "Play" hover button in Desktop mode). I started 2 episodes of the same season, and Jellyfin saves positions of both.

PR #535, opened 2019-10-16
Make login card scalable on focus

dmitrylyzo:
Made a login card similar to a server card to somewhat indicate the focus.
jellyfin-web_issue03_01

### DIFF
--- a/src/components/cardbuilder/cardBuilder.js
+++ b/src/components/cardbuilder/cardBuilder.js
@@ -3,7 +3,7 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
         'use strict';
 
         var devicePixelRatio = window.devicePixelRatio || 1;
-        var enableFocusTransfrom = !browser.slow && !browser.edge;
+        var enableFocusTransform = !browser.slow && !browser.edge;
 
         function getCardsHtml(items, options) {
             if (arguments.length === 1) {
@@ -1221,8 +1221,13 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
                 className += ' card-hoverable';
             }
 
-            if (!enableFocusTransfrom || !layoutManager.tv) {
-                className += ' card-nofocustransform';
+            if (layoutManager.tv) {
+
+                className += ' show-focus'
+
+                if (enableFocusTransform) {
+                    className += ' show-animation'
+                }
             }
 
             var imgInfo = getCardImageUrl(item, apiClient, options, shape);
@@ -1249,23 +1254,6 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
             }
 
             var cardBoxClass = options.cardLayout ? 'cardBox visualCardBox' : 'cardBox';
-
-            if (layoutManager.tv) {
-
-                if (enableFocusTransfrom) {
-                    cardBoxClass += ' cardBox-focustransform cardBox-withfocuscontent';
-                } else {
-                    cardBoxClass += ' cardBox-withfocuscontent-large';
-                }
-
-                if (options.cardLayout) {
-                    cardBoxClass += ' card-focuscontent';
-
-                    if (!enableFocusTransfrom) {
-                        cardBoxClass += ' card-focuscontent-large';
-                    }
-                }
-            }
 
             var footerCssClass;
             var progressHtml = indicators.getProgressBarHtml(item);
@@ -1384,15 +1372,6 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
             }
 
             var cardScalableClass = 'cardScalable';
-
-            if (layoutManager.tv && !options.cardLayout) {
-
-                cardScalableClass += ' card-focuscontent';
-
-                if (!enableFocusTransfrom) {
-                    cardScalableClass += ' card-focuscontent-large';
-                }
-            }
 
             cardImageContainerOpen = '<div class="' + cardBoxClass + '"><div class="' + cardScalableClass + '"><div class="cardPadder-' + shape + '"></div>' + cardImageContainerOpen;
             cardBoxClose = '</div>';
--- a/src/components/cardbuilder/cardBuilder.js
+++ b/src/components/cardbuilder/cardBuilder.js
@@ -1222,10 +1222,10 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
             }
 
             if (layoutManager.tv) {
-                className += ' show-focus'
+                className += ' show-focus';
 
                 if (enableFocusTransform) {
-                    className += ' show-animation'
+                    className += ' show-animation';
                 }
             }
 
--- a/src/components/cardbuilder/cardBuilder.js
+++ b/src/components/cardbuilder/cardBuilder.js
@@ -1203,6 +1203,8 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
                 shape = shape || 'mixedSquare';
             }
 
+            // TODO move card creation code to Card component
+
             var className = 'card';
 
             if (shape) {
--- a/src/components/cardbuilder/cardBuilder.js
+++ b/src/components/cardbuilder/cardBuilder.js
@@ -1222,7 +1222,6 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
             }
 
             if (layoutManager.tv) {
-
                 className += ' show-focus'
 
                 if (enableFocusTransform) {
--- a/src/components/cardbuilder/chaptercardbuilder.js
+++ b/src/components/cardbuilder/chaptercardbuilder.js
@@ -5,6 +5,8 @@ define(['datetime', 'imageLoader', 'connectionManager', 'layoutManager', 'browse
 
     function buildChapterCardsHtml(item, chapters, options) {
 
+        // TODO move card creation code to Card component
+
         var className = 'card itemAction chapterCard';
 
         if (layoutManager.tv) {
--- a/src/components/cardbuilder/chaptercardbuilder.js
+++ b/src/components/cardbuilder/chaptercardbuilder.js
@@ -8,10 +8,10 @@ define(['datetime', 'imageLoader', 'connectionManager', 'layoutManager', 'browse
         var className = 'card itemAction chapterCard';
 
         if (layoutManager.tv) {
-            className += ' show-focus'
+            className += ' show-focus';
 
             if (enableFocusTransform) {
-                className += ' show-animation'
+                className += ' show-animation';
             }
         }
 
--- a/src/components/cardbuilder/chaptercardbuilder.js
+++ b/src/components/cardbuilder/chaptercardbuilder.js
@@ -1,12 +1,19 @@
 define(['datetime', 'imageLoader', 'connectionManager', 'layoutManager', 'browser'], function (datetime, imageLoader, connectionManager, layoutManager, browser) {
     'use strict';
 
+    var enableFocusTransform = !browser.slow && !browser.edge;
+
     function buildChapterCardsHtml(item, chapters, options) {
 
         var className = 'card itemAction chapterCard';
 
-        if (layoutManager.tv && (browser.animate || browser.edge)) {
-            className += ' card-focusscale';
+        if (layoutManager.tv) {
+
+            className += ' show-focus'
+
+            if (enableFocusTransform) {
+                className += ' show-animation'
+            }
         }
 
         var mediaStreams = ((item.MediaSources || [])[0] || {}).MediaStreams || [];
@@ -91,19 +98,6 @@ define(['datetime', 'imageLoader', 'connectionManager', 'layoutManager', 'browse
 
         var cardBoxCssClass = 'cardBox';
         var cardScalableClass = 'cardScalable';
-
-        if (layoutManager.tv) {
-            var enableFocusTransfrom = !browser.slow && !browser.edge;
-
-            cardScalableClass += ' card-focuscontent';
-
-            if (enableFocusTransfrom) {
-                cardBoxCssClass += ' cardBox-focustransform cardBox-withfocuscontent';
-            } else {
-                cardBoxCssClass += ' cardBox-withfocuscontent-large';
-                cardScalableClass += ' card-focuscontent-large';
-            }
-        }
 
         var html = '<button type="button" class="' + className + '"' + dataAttributes + '><div class="' + cardBoxCssClass + '"><div class="' + cardScalableClass + '"><div class="cardPadder-' + shape + '"></div>' + cardImageContainer + '</div><div class="innerCardFooter">' + nameHtml + '</div></div></div></button>';
 
--- a/src/components/cardbuilder/chaptercardbuilder.js
+++ b/src/components/cardbuilder/chaptercardbuilder.js
@@ -8,7 +8,6 @@ define(['datetime', 'imageLoader', 'connectionManager', 'layoutManager', 'browse
         var className = 'card itemAction chapterCard';
 
         if (layoutManager.tv) {
-
             className += ' show-focus'
 
             if (enableFocusTransform) {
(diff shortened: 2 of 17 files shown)


src/controllers/loginpage.js
```diff
@@ -60,8 +60,15 @@ define(["apphost", "appSettings", "dom", "connectionManager", "loading", "cardSt
         var html = "";
         for (var i = 0; i < users.length; i++) {
             var user = users[i];
-            html += '<button type="button" class="card squareCard scalableCard squareCard-scalable"><div class="cardBox cardBox-bottompadded">';
-            html += '<div class="cardScalable">';
+
+            var cardBoxCssClass = "cardBox cardBox-bottompadded";
+            if (layoutManager.tv) {
+                cardBoxCssClass += " cardBox-focustransform";
```
dkanada: In the future I'd like to move all focus related styling to a single class called `show-focus` to speed up development. That way we can just search for a single class and find every instance, rather than all the custom classes that currently exist.

dmitrylyzo: Do you want me to replace _cardBox-focustransform_ to _show-focus_ (or even mark the top element with it)?
Then `components/cardbuilder/card.css` with backward compatibility (not tested):
```
.cardBox-focustransform,
.cardBox.show-focus,           /* or */
.card.show-focus .cardBox { /* in case of marking the top element */
    will-change: transform;
    transition: transform 200ms ease-out;
}

.card:focus > .cardBox-focustransform,
.card:focus > .cardBox.show-focus,  /* or */
.card.show-focus:focus > .cardBox { /* in case of marking the top element */
    transform: scale(1.18);
}
```
Class _card-focuscontent_ adds some style by theme.

dkanada: I like the `.cardBox.show-focus` method best. Also, there's no need to maintain backward compatibility since we are changing all instances of the class at the same time. The `.card-focuscontent` instances should also just be changed to use `.cardBox.show-focus` because there's really no need for another class that has the same function.

dmitrylyzo: It turned out to be not so simple.

The code controls the focus indication style. For example, in `components/itemidentifier/itemidentifier.js` class `cardBox-focustransform` added in TV mode for "not slow" browser (and at some places not for Edge). On "slow" browsers the color frame is used. The same in
```
components/imageeditor/imageeditor.js
components/cardbuilder/chaptercardbuilder.js
components/cardbuilder/cardBuilder.js
components/imagedownloader/imagedownloader.js
```
Some "simple focus" marker (like `layout-tv`) is needed to save current logic. Maybe it should be configurable like a layout.

**Focus related classes (and possible replacement)**
- `cardBox-focustransform` ~ `.focus-cool .cardBox.show-focus`
Applied to `cardBox` (in some places for "fast" browser only) when transform should be used.

- `card-focusscale`
Applied to `cardBox`. Only appended, no reference in CSS and JS. Safe to remove.

- `card-focuscontent` ~  `.outerBorder .cardBox.show-focus` / `.innerBorder .cardBox.show-focus > .cardScalable`
Set border color of focused card.
In major it is applied to `cardBox`.
In `selectserver.js` and `chaptercardbuilder.js` it is applied to `cardScalable`.
In `cardBuilder.js` it is applied to both (`cardBox` and `cardScalable`) in depend of view style ("Poster" or "Poster card").
So it controls who will have a border: inner (`cardScalable`) or outer (`cardBox`).
This one requires "card border style" marker. Or just leave it in place.
Works in combination with `card-focuscontent-large`.

- `card-focuscontent-large` ~ `.focus-simple .outerBorder  .cardBox.show-focus` / `.focus-simple .innerBorder  .cardBox.show-focus > .cardScalable`
Same as `card-focuscontent`. Set border size when transform is not in use.

- `cardBox-withfocuscontent`
Applied to `cardBox`. Only appended, no reference in CSS and JS. Safe to remove.

- `cardBox-withfocuscontent-large` ~ `.focus-simple .cardBox.show-focus`
Applied to `cardBox` for "slow" browser (and Edge). Set margins when transform is not in use.

dkanada: > It turned out to be not so simple.

That's definitely an understatement :laughing: I remember finding all these classes and slowly backing away, thank you for finally documenting all of them.

What about two classes (show-focus and show-animation) to replace these? The show-focus class will work as mentioned above and the show-animation class will be selectively enabled the same as now. We can then move all transform styles and animations under that class and add it below the layout preference in the user settings.

dmitrylyzo: > What about two classes (show-focus and show-animation) to replace these?

It seems it will be enough. I tried them on cardBuilder only (it is most "styled").
`show-focus` - focus should be indicated
`show-animation` - (as addition) focus should be "beautiful"
They can be moved to card (button) itself like in emby-button you mentioned, but this complicates the selector.
Class `.visualCardBox` can be used as marker inner/outer border.

- Somewhere in `components/cardbuilder/cardBuilder.js`
```
if (layoutManager.tv) {

    cardBoxClass += ' show-focus'

    if (enableFocusTransfrom) {
        cardBoxClass += ' show-animation'
    }
}
```
- `components/cardbuilder/card.css` (_replaced classes for reference only_)
```
/*.cardBox-withfocuscontent-large {*/
.cardBox.show-focus:not(.show-animation) {
    margin: .4em;
}
 
/*.card-focuscontent-large {*/
.cardBox.show-focus:not(.show-animation).visualCardBox,
.cardBox.show-focus:not(.show-animation):not(.visualCardBox) .cardScalable {
    border: .5em solid transparent;
}
 
/*.cardBox-focustransform {*/
.cardBox.show-animation {
    will-change: transform;
    transition: transform 200ms ease-out;
}
 
/*.card:focus > .cardBox-focustransform {*/
.card:focus > .cardBox.show-animation {
    transform: scale(1.18, 1.18);
}
```
- `components/themes/dark/theme.css` (_replaced classes for reference only_)
```
/*.card:focus .card-focuscontent,*/
/* if 'show-animation' the border has zero size */
.card:focus .cardBox.visualCardBox,
.card:focus .cardBox:not(.visualCardBox) .cardScalable {
    border-color: #00a4dc
}
```

Also I missed `focus-nofocustransform`. It applied to `card` when transform is not in use.
If `show-focus` and `show-animation` will be applied to a card itself, then it could be removed.
```
/*.card-nofocustransform {*/
.card.show-focus:not(.show-animation) {
    contain: layout style paint;
}
```

dkanada: That sounds excellent. I'm not sure I have a preference on where to apply the classes right now, either way is an improvement.

dmitrylyzo: For now, this can be as in `cardBuilder.js`
```
var enableFocusTransform = !browser.slow && !browser.edge;
```
Also I think is better to add these classes to the `card` instead of the `cardBox`. In this case you could tune the `card` style too.

JustAMan: go for it! :) should it be done in this PR?

dkanada:
Let's exit that extremely long comment chain for the main thread :p I would prefer to merge this with the changes we discussed, but if you don't want to update the classes I can review this without them.

dmitrylyzo:
I am ready to push with replacements (with rebase on the current master). I have to deminify one method in `controllers/livetvstatus.js`.

In some places focustransform was forced, I changed all to discussed logic (animation on "fast" browsers). Now all cards show their focus equally.

**Checks**
```
                            A  !A
tunerpicker.js              *   *
itemidentifier.js           +   +
chaptercardbuilder.js       +   +
cardBuilder.js              +   +
imageeditor.js              +   +
imagedownloader.js          +   +
loginpage.js                +   +
livetvstatus.js             *   *
selectserver.js             +   +
```
* In `livetvstatus.js` not a `button` but `div` - no focus. I checked it with `div` changed to `button`, but left it as is.
* `tunerpicker.js` is broken on TV (no `centerFocus` and `focusManager`), I checked it too, but did not fix it.

**Possibly focus related combinations:**
```
card scalableCard
backdropCard backdropCard-scalable
visualCardBox cardScalable visualCardBox-cardScalable
squareCard scalableCard squareCard-scalable
overflowSquareCard scalableCard overflowSquareCard-scalable
```

There is some style issue with "Purple Haze" theme due to round border:
jellyfin-web_#525_01

dkanada:
> I have to deminify one method

We just merged the final large deminify pull request, maybe this isn't required anymore?

> Possibly focus related combinations

Wow that's a lot of combinations, we might want to fix that eventually but as long as your pull request doesn't break those I don't think it's required to investigate that now, what do you think?

dmitrylyzo:
I force push with class replacements.

As I said, there is some places where style is changed (only "slow" browsers are affected - real TVs):
* `src/components/itemidentifier/itemidentifier.js`:
`focuscontent` was on `cardBox`. But it is not `visualCardBox`, so frame will be on `cardScalable`.
* `src/controllers/loginpage.js`
Now it shows focus with a frame.
* `src/controllers/selectserver.js`
Now it shows focus with a frame.

These are set but not used:
```
scalableCard
backdropCard-scalable
visualCardBox-cardScalable - looks like a shortcut
squareCard-scalable
overflowSquareCard-scalable
```

dkanada:
> These are set but not used

If they aren't used then feel free to remove them everywhere :) otherwise I'll get around to it eventually.

src/components/cardbuilder/cardBuilder.js
```diff
@@ -3,7 +3,7 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
         'use strict';
 
         var devicePixelRatio = window.devicePixelRatio || 1;
-        var enableFocusTransfrom = !browser.slow && !browser.edge;
+        var enableFocusTransform = !browser.slow && !browser.edge;
```
JustAMan: what changed here?..

src/components/cardbuilder/cardBuilder.js
```diff
@@ -1221,8 +1221,13 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
                 className += ' card-hoverable';
             }
 
-            if (!enableFocusTransfrom || !layoutManager.tv) {
-                className += ' card-nofocustransform';
+            if (layoutManager.tv) {
+
```
JustAMan:
```suggestion
```

src/components/cardbuilder/chaptercardbuilder.js
```diff
@@ -1,12 +1,19 @@
...
+    var enableFocusTransform = !browser.slow && !browser.edge;
+
     function buildChapterCardsHtml(item, chapters, options) {
 
         var className = 'card itemAction chapterCard';
 
-        if (layoutManager.tv && (browser.animate || browser.edge)) {
-            className += ' card-focusscale';
+        if (layoutManager.tv) {
+
```
JustAMan:
```suggestion
```

src/components/imagedownloader/imagedownloader.js
```diff
@@ -196,12 +198,13 @@ define(['loading', 'apphost', 'dialogHelper', 'connectionManager', 'imageLoader'
         if (tagName === 'button') {
             cssClass += ' btnImageCard';
 
-            if (layoutManager.tv && !browser.slow) {
-                cardBoxCssClass += ' cardBox-focustransform';
-            }
-
             if (layoutManager.tv) {
-                cardBoxCssClass += ' card-focuscontent cardBox-withfocuscontent';
+
```
JustAMan:
```suggestion
```

src/components/imageeditor/imageeditor.js
```diff
@@ -105,12 +107,13 @@ define(['dialogHelper', 'connectionManager', 'loading', 'dom', 'layoutManager',
         if (tagName === 'button') {
             cssClass += ' btnImageCard';
 
-            if (layoutManager.tv && !browser.slow) {
-                cardBoxCssClass += ' cardBox-focustransform';
-            }
-
             if (layoutManager.tv) {
-                cardBoxCssClass += ' card-focuscontent cardBox-withfocuscontent';
+
```
JustAMan:
```suggestion
```

src/components/itemidentifier/itemidentifier.js
```diff
@@ -190,16 +192,17 @@ define(["dialogHelper", "loading", "connectionManager", "require", "globalize",
             padderClass = "cardPadder-portrait";
         }
 
-        if (layoutManager.tv && !browser.slow) {
-            cardBoxCssClass += " cardBox-focustransform";
-        }
+        if (layoutManager.tv) {
```
JustAMan:
```suggestion
```

src/controllers/livetvstatus.js
```diff
@@ -10,11 +12,14 @@ define(["jQuery", "globalize", "scripts/taskbutton", "dom", "libraryMenu", "layo
         padderClass = "cardPadder-backdrop";
 
         if (layoutManager.tv) {
-            cssClass += " card-focusscale";
-            cardBoxCssClass += " cardBox-focustransform";
+
```
JustAMan:
```suggestion
```

src/controllers/loginpage.js
```diff
@@ -60,7 +62,22 @@ define(["apphost", "appSettings", "dom", "connectionManager", "loading", "cardSt
         var html = "";
         for (var i = 0; i < users.length; i++) {
             var user = users[i];
-            html += '<button type="button" class="card squareCard scalableCard squareCard-scalable"><div class="cardBox cardBox-bottompadded">';
+
+            var cssClass = "card squareCard scalableCard squareCard-scalable";
+
+            if (layoutManager.tv) {
+
```
JustAMan:
```suggestion
```

src/controllers/selectserver.js
```diff
@@ -21,17 +23,24 @@ define(["loading", "appRouter", "layoutManager", "appSettings", "apphost", "focu
...
 
             if (layoutManager.tv) {
-                cardBoxCssClass += " cardBox-focustransform";
+
+                cssClass += ' show-focus'
+
+                if (enableFocusTransform) {
+                    cssClass += ' show-animation'
+                }
             }
```
JustAMan: can this be extracted to a common function instead of being copy-pasted in like hundred places?

src/components/cardbuilder/cardBuilder.js
dmitrylyzo: TransfROm -> TransfORm

src/controllers/selectserver.js
dmitrylyzo: I also thought about it, but the question is, where should it be?
Currently, this code placed next to the card generating code. I think that all card generating code should be in `cardBuilder` (need to export `buildCard` function), but I did not dare to do such refactoring.

dmitrylyzo:
> Mostly fine, but cannot check css - I have to hope you tested that :)

To check CSS on PC, you may force "slow" browser here:
https://github.com/jellyfin/jellyfin-web/blob/c10f3feb3149d761c993a0c3b007c09c8a848f14/src/scripts/browser.js#L286-L288

src/controllers/selectserver.js
dkanada: Yes, the main issue is that CardBuilder only really works with library items, but we use the same style in several other elements (login cards, server cards, library cards, device cards) without the component. We need a more generic Card element to use in those cases, but that's a huge change.

dkanada: Maybe adding a TODO comment in every location it gets copied so other people might notice in the future and make the refactor?

dmitrylyzo: For now, it is possible to extract focus related code into some function, and use it even with emby-button:
```
function focusClass() {

    var cssClass = ' show-focus';    // space because it must be added to the base class

    if (enableFocusTransform) {
        cssClass += ' show-animation';
    }

    return cssClass;
}
```
Or even
```
function focusClass() {

    var cssClass = '';

    if (layoutManager.tv) {
        cssClass += ' show-focus';    // space because it must be added to the base class

        if (enableFocusTransform) {
            cssClass += ' show-animation';
        }
    }

    // Here may be variations for other layouts

    return cssClass;
}
```

dkanada: I would just add the method as an export to CardBuilder for now, and place a TODO there that states this is a workaround until the component is more generic.

JustAMan: I would be fine with you just adding `TODO`s for now

PR #538, opened 2019-10-16
Fix emby-scroller padded for mobile

grafixeyehero:
Before 
Screenshot

After
Screenshot (3)

### DIFF
--- a/src/components/emby-scrollbuttons/emby-scrollbuttons.css
+++ b/src/components/emby-scrollbuttons/emby-scrollbuttons.css
@@ -2,6 +2,7 @@
     position: absolute;
     top: 0;
     right: 0;
+    margin-right:3.3%;
     align-items: center;
     justify-content: center;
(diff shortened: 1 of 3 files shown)


dkanada:
1. The scroll buttons shouldn't be visible on the mobile layout, so that's an issue.
2. I was concerned about the left margin, which should match the section header margin.

grafixeyehero:
> The scroll buttons shouldn't be visible on the mobile layout, so that's an issue.

a simple fix like this 
@media all and (max-width:50em) {
    .emby-scrollbuttons {
        display: none !important
    }
}

dkanada:
I thought I had only enabled them for the mobile layout though, which would be preferable to a css change.

grafixeyehero:
> 
> 
> I thought I had only enabled them for the mobile layout though, which would be preferable to a css change.

closing the PR

Issue #528, opened 2019-10-15
Can't add a Live TV Guide Provider


hoshsadiq:
**Describe the bug**
Unable to add xmltv epg data

**To Reproduce**

1. Add tuner device
2. Add xmltv data
3. Notice error:
   ```
   Could not find file 'jellyfin-web/components/tvproviders/.template.html'.
   ```

**Logs**
No server logs, but here are relevant JS console logs:
```
Load failed: components/tvproviders/.js: http://localhost:8097/web/components/tvproviders/.js?v=10.4.0
```

**Screenshots**

**System (please complete the following information):**
 - OS: Docker
 - Browser: Firefox
 - Jellyfin Version: 10.4.0

**Additional context**
The URL goes to, so it seems to connect to the correct type: http://localhost:8097/web/index.html?start=wizard#!/livetvguideprovider.html?type=xmltv
I've also looked in the relevant directory and `xmltv.template.html` does exit, but for some reason the it's unable to read type when trying to find the file.

Possibly related to jellyfin/jellyfin#108 jellyfin/jellyfin#111 jellyfin/jellyfin#57 or perhaps even dupes of one of those, however I didn't reopen an old thread.

JustAMan:
@hoshsadiq moved

hoshsadiq:
Thank you @JustAMan 

Also possibly related/duplicate of #493 but this is in stable, not in nightly.

JustAMan:
@thornbill can you please have a look?

grafixeyehero:
@hoshsadiq did you test like this http://localhost:8097/web/index.html#!/livetvguideprovider.html?type=xmltv
maybe the problem was `start=wizard#!`

hoshsadiq:
I haven't.  Didn't even realise. I'll check it out later

hoshsadiq:
So weirdly it works now, even when the wizard bit is there (specifically it automatically redirects to a url without the `start=wizard`).

PR #542, opened 2019-10-19
Improve wizard finish page redirect

grafixeyehero:
**Issues**
https://github.com/jellyfin/jellyfin-web/issues/528

### DIFF
--- a/src/controllers/wizardfinishpage.js
+++ b/src/controllers/wizardfinishpage.js
@@ -8,7 +8,6 @@ define(["loading"], function(loading) {
         }).then(function() {
             loading.hide();
             window.location.href = "index.html";
-
         });
     }
     return function(view, params) {
--- a/src/controllers/wizardfinishpage.js
+++ b/src/controllers/wizardfinishpage.js
@@ -6,17 +6,12 @@ define(["loading"], function(loading) {
             url: ApiClient.getUrl("Startup/Complete"),
             type: "POST"
         }).then(function() {
-            Dashboard.navigate("dashboard.html");
             loading.hide();
+            window.location.href = "index.html";
+
         });
     }
     return function(view, params) {
         view.querySelector(".btnWizardNext").addEventListener("click", onFinish);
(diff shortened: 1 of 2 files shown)


src/controllers/wizardfinishpage.js
```diff
@@ -6,17 +6,12 @@ define(["loading"], function(loading) {
             url: ApiClient.getUrl("Startup/Complete"),
             type: "POST"
         }).then(function() {
-            Dashboard.navigate("dashboard.html");
             loading.hide();
+            window.location.href = "index.html";
+
```
dkanada: maybe remove this newline before merge

grafixeyehero: done

src/controllers/wizardfinishpage.js
```diff
@@ -6,17 +6,11 @@ define(["loading"], function(loading) {
             url: ApiClient.getUrl("Startup/Complete"),
             type: "POST"
         }).then(function() {
-            Dashboard.navigate("dashboard.html");
             loading.hide();
+            window.location.href = "index.html";
```
dkanada: Did you test this redirect? @thornbill mentioned it might need `web` at the front, but if this works fine I'd rather leave it this way.

grafixeyehero: i tested but @thornbill why `web` needed?

thornbill: I haven’t had a chance to test this pr, but I know going directly to `/index.html` on my server returns an error. It must either be `/` or `/web/index.html` for the page to load correctly.

dkanada: @thornbill wouldn't it be fine since the location property can be a relative link?

thornbill: Tested and it works as intended!

joshuaboniface:
Cherry-picked into 10.4.1.

PR #544, opened 2019-10-20
Add ScrollManager to control scrolling on focus

dmitrylyzo:
To make another step for #303 

**Changes**
I rewrote my scroll manager.
- Now it just keeps track of the focus and adjusts the scrolling.
- By default, it is enabled only on Tizen and webOS in TV mode.
- Scrolls to center of view.
- Smooth scrolling on Tizen 4 is not supported, so it is done through requestAnimationFrame. In the webOS 4 emulator, the animation also works, but I left it off, because other scrollers work without animation.

**Limitations**
- No cascade scrolling. Only two scrollers at a time are supported: horizontal and vertical. Vertical scroller is always a window.

**Todo**
- Need some marker (class) to force to scroll to the top to show the top menu. When focus goes on hidden menu, window scrolls a little (because menu is fixed and placed out of view), but the top menu remains hidden. For now I "hacked" it by test "out of view" state.
- Need some marker (class) to mark UI block (`detailPagePrimaryContainer` as example of such block) which also force to scroll to the top (sticked at the top). For now I "hacked" it by minimum vertical scroll amount - the size of top menu.
- Need some marker (if no such) to mark a group of focusable elements to scroll on the group instead of element itself. For example, "Library Order" in settings.

_(So far, we have decided that the tasks do not require immediate implementation)_

**Bugs**
- [x] Due to "scroll to center of view" I realized that document has size bigger than TV screen, and it scrolls when navigating library. ~~Another consequence of bigger size is that navigating in alphaPicker is broken - navigates out of alphaPicker on an overlaped card.~~

- [x] Changing "Library Order" makes the focus lost. _(Not actually related)_


dmitrylyzo:
Now vertical scroller may not be a window, but still no cascade scrolling (is this necessary?).

ScrollManager supports (in TV mode) webOS 2/3 (browser only), Tizen 4/5 (native/browser), Firefox 69/70, Chrome 76.0.3809.100 - checked with animated scroll enabled everywhere.

**Bugs**
At the moment, it looks jerky when mouse autofocus is activated (`components/serverNotifications/mouseManager.js`). When you hover over an element, focus is triggered, the window scrolls and drags another element under the cursor, and everything repeats.
Native focus looks a bit better, because it does not scroll in the center every time (centering when out of view), but emby-scroller still centering.
If there is a way to detect that the focus triggered by the mouse (or opposite - by the keyboard), I could scroll to the nearest position.

**Notes**
webOS 2/3 browser is not detected as webOS.

**Experiment**
Native app webOS 3 does not scroll window (as mentioned by @oddstr13 in #314). But it scrolls div. I tried adding the line below and it started to scroll with the mouse wheel, focus (native) and my ScrollManager. But it breaks the top menu - it stops hiding.
```
--- a/src/components/viewContainer.js
+++ b/src/components/viewContainer.js
@@ -70,6 +70,7 @@ define(["browser", "dom", "layoutManager", "css!components/viewManager/viewConta
                 }
 
                 view.classList.add("mainAnimatedPage");
+                view.classList.add("scrollY");
 
                 if (currentPage) {
                     if (newViewInfo.hasScript && window.$) {
```

**Reference**
_Taken from emulators except Tizen 4_
```
webOS 1 (app): Mozilla/5.0 (WebOS; Linux i686) AppleWebKit/537.41 (KHTML, like Gecko) Large Screen WebAppManager Safari/537.41
webOS 1 (browser): Mozilla/5.0 (Linux; NetCast; U) AppleWebKit/537.31 (KHTML, like Gecko) Chrome/26.0.1410.33 Safari/537.31 SmartTV/5.0
webOS 2 (app): Mozilla/5.0 (WebOS; Linux i686) AppleWebKit/538.2 (KHTML, like Gecko) Large Screen WebAppManager Safari/538.2
webOS 2 (browser): Mozilla/5.0 (Linux; NetCast; U) AppleWebKit/537.31 (KHTML, like Gecko) Chrome/34.0.1847.137 Safari/537.31 SmartTV/6.0
webOS 3 (app): Mozilla/5.0 (WebOS; Linux/SmartTV) AppleWebKit/537.36 (KHTML, like Gecko) QtWebEngine/5.2.1 Chrome/38.0.2125.122 Safari/537.36 WebAppManager
webOS 3 (browser): Mozilla/5.0 (Linux; NetCast; U) AppleWebKit/537.31 (KHTML, like Gecko) Chrome/38.0.2125.122 Safari/537.31 SmartTV/6.0
webOS 4 (app): Mozilla/5.0 (WebOS; Linux/SmartTV) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/53.0.2785.34 Safari/537.36 WebAppManager
Tizen 4 (app): Mozilla/5.0 (SMART-TV; LINUX; Tizen 4.0) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 TV Safari/537.36
Tizen 4 (browser): Mozilla/5.0 (SMART-TV; Linux; Tizen 4.0) AppleWebKit/537.36 (KHTML, like Gecko) SamsungBrowser/2.1 Chrome/56.0.2924.0 TV Safari/537.36
Tizen 5 (app): Mozilla/5.0 (SMART-TV; LINUX; Tizen 5.0) AppleWebKit/537.36 (KHTML, like Gecko) Version/5.0 TV Safari/537.36
Tizen 5 (browser): Mozilla/5.0 (SMART-TV; Linux; Tizen 5.0) AppleWebKit/537.36 (KHTML, like Gecko) SamsungBrowser/2.2 Chrome/63.0.3239.84 TV Safari/537.36
```

dmitrylyzo:
I found out why the page content is wider than the screen. In 38bed87e92595088ea9644d719df3d2fb914d44e the lines below was added to `components/cardbuilder/card.css` (I suppose to line up the section title and the card in the scroller).
```
 .itemsContainer {
     display: flex;
+    margin-left: -0.6em;
+    margin-right: -0.6em;
 }
```
I could add `overflow-x/y: hidden` condition to ScrollManager (tested already), but it could be better to fix the page width instead. I tried to fix page by adding `overflow-x: hidden` to `.page` but it move scrollbar from the documentElement/body to the page itself breaking the top menu hiding. What do you think?

dkanada:
What about using the `first` selector to remove the left padding on the first card in a scroller and the same for last? Also, this scroll manager will only be enabled for the TV layout right?

dmitrylyzo:
> Also, this scroll manager will only be enabled for the TV layout right?

Yes.

> What about using the first selector to remove the left padding on the first card in a scroller and the same for last?

It will work for single row container.

I thought it over again. Since browsers do not scroll horizontally due to `overflow-x: hidden` on `body`, ScrollManager should not do this either.

dkanada:
Can you link the location of the negative margin? Also, we might want to remove that anyways, it was supposed to fix the scroller margin on mobile devices and is completely failing so there's really no reason to keep it.

dmitrylyzo:
https://github.com/jellyfin/jellyfin-web/blob/f64c1c08a3ada693b60decdd37d2863b0235d7ce/src/components/cardbuilder/card.css#L33-L47
This CSS trick is necessary due to the lack of `gap` property for `flex`. So it actually removes the gap between the edge card and the scroller edge.
I think that in the future there will be many more such tricks (at least for compatibility).

dkanada:
@grafixeyehero this is the fix I was talking about that breaks mobile. Maybe we can remove the hack and just decrease the padding for the scroller element? I'd like to fix this for all platforms without relying on negative margins, which seems to break mobile.

dkanada:
> it looks jerky when mouse autofocus is activated

What about disabling the mouse manager for the TV layout?

dmitrylyzo:
> What about disabling the mouse manager for the TV layout?

Disabling it for TV layout means completely removing this feature.
https://github.com/jellyfin/jellyfin-web/blob/a944bc62eff8fd7f95b6c0efe18b750aaed700eb/src/components/serverNotifications/mouseManager.js#L89-L104

dkanada:
Ah. I had no clue it was only used for TV browsers haha. What do you think? I feel like most of the time if someone has a mouse hooked up to their TV they would prefer to focus with a click rather than hover.

dmitrylyzo:
I prefer keyboard navigation. In the native app, the remote control works like a keyboard (Tizen 4/5, webOS 4) - no problem. But in the browser on TV, it works like a mouse :disappointed: (true for Tizen 4/5) - I just use a wireless keyboard in this case.

In case of using a mouse with autofocus feature disabled, user can (in theory) navigate with a keyboard on one button but click (and then focus) another. Maybe this is not so bad - similar to a desktop browser. :thinking:

Also, maybe this was done to highlight the buttons on TV. If autofocus feature will be disabled and highlighting will be done with `:hover`, user will see two "focused" buttons at the same time. :dizzy_face:

dkanada:
We could also use the user agent strings you extracted above and automatically set which focus managers are enabled. Or use a sane default and then expose more options as a setting in the user preferences. The latter sounds like a good idea in my opinion.

dmitrylyzo:
Tried some "focus device detecting" idea, but not sure about reliability. Also this does not help with `emby-scroller` and `scrollhelper` (for example, itemidentifier dialog) - they scroll by themselves to the center of the screen (horizontally/vertically).

The idea is to store the device by the last event.
```
    if (isEnabled()) {
        var lastDevice;

        dom.addEventListener(window, "mouseenter", function(e) {
            lastDevice = "mouse";
        }, {
            capture: true,
            passive: true
        });

        dom.addEventListener(window, "pointerenter", function(e) {
            lastDevice = "mouse";
        }, {
            capture: true,
            passive: true
        });

        dom.addEventListener(window, "click", function(e) {
            lastDevice = "mouse";
        }, {
            capture: true,
            passive: true
        });

        dom.addEventListener(window, "keydown", function(e) {
            lastDevice = "keyboard";
        }, {
            capture: true,
            passive: true
        });

        dom.addEventListener(window, "focusin", function(e) {
            if (lastDevice === "mouse") {
                return;
            }

            setTimeout(function() {
                scrollToElement(e.target, useSmoothScroll());
            }, 0);
        }, {capture: true});
    }
```

dmitrylyzo:
What you think about the markers (from TODO)?
* `scrollfocus-top` for scroll to the top (top menu and `detailPagePrimaryContainer`).
* `scrollfocus` for containers that should be completely visible (doesn't mean centered).

dkanada:
Can you link the TODO entries you saw for those two classes? I haven't seen them.

dmitrylyzo:
From this PR description.

dmitrylyzo:
I wanted to replace "hacks" with some adequate solution. I just need marker names.
_Spoiler: it seems that hacks are becoming a feature._

**HACK 1**
https://github.com/jellyfin/jellyfin-web/blob/2a96b67802003846bfe2aca37707bcd1df3a5500/src/components/scrollManager.js#L445-L449
`scrollfocus-top` will solve this.
In webOS 2/3 top menu doesn't hide. Hack doesn't force to scroll to the top. But marker will - maybe this is not good for user experience.

**HACK 2**
https://github.com/jellyfin/jellyfin-web/blob/2a96b67802003846bfe2aca37707bcd1df3a5500/src/components/scrollManager.js#L451-L455
`scrollfocus-top` works, but it is too aggressive and cannot be used everywhere. For example, library item lister is too big, and it would be impossible to see the bottom items. So, this hack has become essential.

`scrollfocus` is not needed for now due to the "centering", but if there is a nearest scroll (only to display the element), it will be useful.

I'm starting to think, to leave everything (hacks) as it is :smiley:

### Document size issue (negative margins)
I think it's okay if I add the check "overflow-x/y: hidden" so as not to scroll. As I said, browsers do not scroll body with "overflow-x/y: hidden". So why should I?
```
        get scrollWidth() {
            var overflowX = window.getComputedStyle(document.body).overflowX;
            return overflowX !== "hidden" ? Math.max(document.documentElement.scrollWidth, document.body.scrollWidth) : this.clientWidth;
        },

        get scrollHeight() {
            var overflowY = window.getComputedStyle(document.body).overflowY;
            return overflowY !== "hidden" ? Math.max(document.documentElement.scrollHeight, document.body.scrollHeight) : this.clientHeight;
        },
```

### Mouse autofocus
I think to leave it as it is (without device detection). You will decide after you try ScrollManager.

dkanada:
> Document size

I think the fix here should be to just find a different solution and avoid negative margins entirely. The one you linked above should be the only instance and there must be a different solution that avoids the negative margins.

> Mouse autofocus

I'll try it out soon, but I still think we should try to disable the mouse focus when possible. This layout will also be used by people who want keyboard control with thier PC connected to a TV, which would cause issues with a mouse connected.

> Hacks

I'm fine with the hacks for now since they're well documented and would cause headaches to work around, but I'll keep thinking of potential solutions.

dmitrylyzo:
Now animated scroll is similar to "270ms ease-out" of `emby-scroller`. And `ScrollTime` is used as expected.

Nickbert7:
As you are reworking the focus and the keyboard navigation, I'm not sure if it helps but you might want to have a look on my PR from the past:
https://github.com/jellyfin/jellyfin-web/pull/388/commits/9cc0491e3a6ae2a63bc735d8e438066ac6ae40ff

It was a try/hack to hide the nav bar so the menu entries "outside" the screen wouldn’t be focused.
Unfortunately I only worked after a page refresh...
So you might find a better solution or even get the menu working in the tv layout.

dmitrylyzo:
@Nickbert7,
I also skipped this in my first draft PR (#522) - I did `focusManager` to focus only on currently visible elements, and implemented incremental scrolling (refused native one) to read large descriptions, video metadata (this was at the very bottom), navigate to the very top to make the top menu visible.

Now this is much simpler - scrollManager just fixes scroll.

If the left menu is required, there is some options:

1. Add/return top menu button to show left menu and focus on its first entry; hide it on blur. _(like in desktop layout)_

2. Show/hide the left menu on focus/blur of its elements. _(disliked this idea while writing - user can use TV layout with a mouse, but will not be able to open the left menu)_
Currently, the left menu is `focuscontainer-y` - only vertical navigation. "Right" button will pop out from it. But for now, it is a bit tricky to get into the left menu. For example, on home screen, you have to move up, because section scrollers is `focuscontainer-x` - release when moving vertically. So, `focusManager` and/or "navigation layout" (assigned focus classes) need to be reworked to be able to focus in left menu by "Left" button.

At the beginning  of making Jellyfin for Tizen, I tried [luke-chang/js-spatial-navigation.](https://github.com/luke-chang/js-spatial-navigation) But then I thought it would be difficult to sync all wrapped Jellyfin pages. Mentioned project uses some focus rules, remembers the last focused element of focuscontainer.

**(!)** There is another forcibly hidden menu - `nowplayingbar`, and I have no idea what to do with it. Because of this, you cannot control the playback of music.

P.S.: I found a bug with "minimumScrollY" hack when testing navigating the left menu after your message.

Nickbert7:
I just wanted that you are aware of what I have "done" with the menue.
If it is OK for you how it is now, then it is also fine for me.
Btw good work so far👍

JustAMan:
@dmitrylyzo what makes this PR a WIP, i.e. what is not ready for it to be merged (except a review)?

dmitrylyzo:
@JustAMan,
I was waiting for "document oversize" issue. Looks like this has already been fixed, and ScrollManager now scrolls properly.

src/components/scrollManager.js
```diff
@@ -0,0 +1,501 @@
...
+     * Returns minimum vertical scroll.
+     * Scroll less than that value will be zeroed.
+     *
+     * @return {number} minimum vertical scroll
+     */
+    function minimumScrollY() {
+        if (_minimumScrollY === 0) {
+            var topMenu = document.querySelector(".headerTop");
+            if (topMenu) {
+                _minimumScrollY = topMenu.clientHeight;
```
JustAMan: what happens if `.headerTop` ever changes size? cannot we just never cache it and always return `topMenu.clientHeight` if it's available

src/components/scrollManager.js
```diff
@@ -0,0 +1,501 @@
...
+     * @param {HTMLElement} element element for which parent is being searched
+     * @param {boolean} vertical search for vertical scrollable parent
+     */
+    function getScrollableParent(element, vertical) {
+        if (element) {
+            var parent = element.parentElement;
+
+            while (parent) {
+                if ((!vertical && parent.scrollWidth > parent.clientWidth && parent.classList.contains("scrollX")) ||
+                    (vertical && parent.scrollHeight > parent.clientHeight) && parent.classList.contains("scrollY")) {
```
JustAMan:
```suggestion
                    (vertical && parent.scrollHeight > parent.clientHeight && parent.classList.contains("scrollY"))) {
```
otherwise condition looks broken

src/components/scrollManager.js
```diff
@@ -0,0 +1,501 @@
...
+    function calcScroll(scrollerData, elementPos, elementSize, centered) {
+        var maxScroll = scrollerData.scrollSize - scrollerData.clientSize;
+
+        var scroll;
+
+        if (centered) {
+            scroll = elementPos + (elementSize - scrollerData.clientSize) / 2;
+        } else {
+            var delta = fitRange(elementPos, elementPos + elementSize - 1, scrollerData.scrollPos, scrollerData.scrollPos + scrollerData.clientSize - 1);
+            scroll = scrollerData.scrollPos - delta;
```
JustAMan: why is it so instead of `scroll = elementPos + delta`?

src/components/scrollManager.js
```diff
@@ -0,0 +1,501 @@
...
+        function scrollAnim(currentTimestamp) {
+
+            start = start || currentTimestamp;
+
+            var k = Math.min(1, (currentTimestamp - start) / ScrollTime);
+
+            if (k === 1) {
+                resetScrollTimer();
+                xScroller.scrollLeft = scrollX;
+                yScroller.scrollTop = scrollY;
```
JustAMan: shouldn't this be done by calling `scrollToHelper()`?

src/components/scrollManager.js
dmitrylyzo: I tried to minimize the code that needs to be executed. I worry about `querySelector`. But maybe I over-optimized it :smile: , because this function is called only once on each user navigation.

src/components/scrollManager.js
dmitrylyzo: This completes the scroll if some fractional parts remain.
At that time, I still hadn't decided if I wanted to call the function.
If only for uniformity.
```
builtinScroll(xScroller, scrollX, yScroller, scrollY, false);
```

JustAMan: I'm worried about elements not having a setter for e.g. `.scrollLeft`, which is handled by your helper but not by this place in the code

src/components/scrollManager.js
JustAMan: Yeah, this looks like premature optimization :) I think you shouldn't cache stuff here. Also that selector should be pretty easy to query for...

src/components/scrollManager.js
dmitrylyzo: `fitRange` returns delta to move `element` into `scroller` viewport. But, in fact, we should move `scroller` - move its viewport over `element`.

PR #547, opened 2019-10-22
Use JavascriptSubtitlesOctopus to render ASS subtitles with embedded fonts.

Unhelpful:
Use JavascriptSubtitlesOctopus, which uses a WebAssembly or asm.js version of libass, to render subtitles. Requires https://github.com/jellyfin/jellyfin/pull/2195 which exposes attached fonts in the API.


src/components/htmlvideoplayer/plugin.js
```diff
@@ -1071,6 +1101,14 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
             });
         }
 
+        function renderSsaAss(videoElement, track, item) {
+            if (supportsCanvas() && supportsWebWorkers()) {
+                renderWithSubtitlesOctopus(videoElement, track, item);
+            } else {
+                renderWithLibjass(videoElement, track, item);
```
dkanada: I was told that libjass wasn't working right now, is that not the case?

src/components/htmlvideoplayer/plugin.js
```diff
@@ -1024,6 +1039,21 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
             lastCustomTrackMs = 0;
         }
 
+        function renderWithSubtitlesOctopus(videoElement, track, item) {
+            var options = {
+                video: videoElement,
+                subUrl: getTextTrackUrl(track, item),
+                fonts: self._currentPlayOptions.mediaSource.MediaAttachments.map(i => i.DeliveryUrl),
+                workerUrl: appRouter.baseUrl() + "/JavascriptSubtitlesOctopus/subtitles-octopus-worker.js",
```
dkanada: Is there any way to avoid a hardcoded URL here or are we stuck with it?

webpack.common.js
```diff
@@ -10,9 +10,18 @@ module.exports = {
...
+            {
+                from: "**/*",
+                to: "."
+            },
+            {
+                from: "../node_modules/libass-wasm/dist/subtitles-octopus-worker.*",
+                to: "JavascriptSubtitlesOctopus",
+                transformPath(targetPath, absolutePath) {
+                    return Promise.resolve(path.join("JavascriptSubtitlesOctopus", path.basename(targetPath)));
+                }
```
dkanada: I'm guessing this section was to accomodate the URL I mentioned above?

Unhelpful: I've removed the use of transformPath, and modified the options to reflect the new installation path of the worker script. I don't think we can completely avoid hard-coding /some/ path to the worker, unless it's installed at top level.

src/components/htmlvideoplayer/plugin.js
Unhelpful: libjass works for me, but the subtitles are rendered as though the window height is the video height, so they can end up huge and behind the OSD if the window is narrow.

dkanada: So from now on libjass will be used as a fallback for ssa and ass subtitles?

webpack.common.js
dkanada: That's fine, thanks for the improvements.

dali99:
> The changes look good to me, and I'm quite interested to see what the improved results look like! Just out of curiosity, can you compare the new behavior with support from other media servers? I have absolutely no idea how well any of them handle custom fonts for subtitles.

See https://github.com/jellyfin/jellyfin/issues/1589

Plex handles it perfectly

webpack.common.js
```diff
@@ -10,9 +10,15 @@ module.exports = {
...
-            to: "."
-        }])
+        new CopyPlugin([
+            {
+                from: "**/*",
+                to: "."
+            },
+            {
+                context: path.resolve(__dirname),
+                from: "node_modules/libass-wasm/dist/subtitles-octopus-worker.*",
```
dkanada: We recently added a new directory to replace the node_modules directory in the built client, can you change this path to match that?

Unhelpful: fixed

src/components/htmlvideoplayer/plugin.js
```diff
@@ -1069,6 +1100,14 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
             });
         }
 
+        function renderSsaAss(videoElement, track, item) {
+            if (supportsCanvas() && supportsWebWorkers()) {
+                renderWithSubtitlesOctopus(videoElement, track, item);
+            } else {
+                renderWithLibjass(videoElement, track, item);
```
JustAMan: can we have a `console.log()` about falling back to libjass?

dkanada: there is also a new conflict if you have time to fix it

PR #548, opened 2019-10-22
Add favorite people section to favorite tab

grafixeyehero:
https://github.com/jellyfin/jellyfin/pull/1931  PR merge first

Screenshot (13)


src/components/itemhelper.js
```diff
@@ -273,8 +273,13 @@ define(['apphost', 'globalize'], function (appHost, globalize) {
         },
 
         canRate: function (item) {
+            var itemType = item.Type;
+
+            if (item.Type === 'Program' || item.Type === 'Timer' || item.Type === 'SeriesTimer' || item.Type === 'CollectionFolder' || item.Type === 'UserView' || item.Type === 'Channel' || itemType === 'Season' || itemType === 'Studio') {
+                return false;
+            }
```
dkanada: There's no need to create a new variable here, the other checks just use the item type directly.

src/controllers/favorites.js
```diff
@@ -117,59 +128,82 @@ define(["appRouter", "cardBuilder", "dom", "globalize", "connectionManager", "ap
...
-            return section.showTitle && lines++, section.showYear && lines++, section.showParentTitle && lines++, cardBuilder.getCardsHtml({
+        return function (items) {
+            var supportsImageAnalysis = appHost.supports("imageanalysis");
+            var cardLayout = (appHost.preferVisualCards || supportsImageAnalysis) && section.autoCardLayout && section.showTitle;
+            cardLayout = false;
+            var serverId = this.apiClient.serverId();
+            var leadingButtons = layoutManager.tv ? [{
+                name: globalize.translate("All"),
+                id: "more",
+                icon: "&#xE87D;",
```
dkanada: We should replace these with ASCII text, the unicode values are a pain to deal with. I haven't found an easy way to update all of them automatically though.

src/components/itemhelper.js
grafixeyehero: need by season and studio item.Type not define.  if i remove those to yay it not need.  otherwise we can add season on the favorite section

src/controllers/favorites.js
grafixeyehero: PR: Replace all Unicode icon with there value https://github.com/jellyfin/jellyfin-web/pull/582

src/controllers/favorites.js
```diff
@@ -178,39 +212,100 @@ define(["appRouter", "cardBuilder", "dom", "globalize", "connectionManager", "ap
...
+            html += '<div class="sectionTitleContainer sectionTitleContainer-cards padded-left">';
+
+            if (layoutManager.tv) {
+                html += '<h2 class="sectionTitle sectionTitle-cards">' + globalize.translate(section.name) + "</h2>";
+            } else {
+                html += '<a is="emby-linkbutton" href="' + getRouteUrl(section, apiClient.serverId()) + '" class="more button-flat button-flat-mini sectionTitleTextButton">';
+                html += '<h2 class="sectionTitle sectionTitle-cards">';
+                html += globalize.translate(section.name);
+                html += "</h2>";
+                html += '<i class="md-icon">&#xE5CC;</i>';
```
JustAMan: pls use name not number here

grafixeyehero: PR: Replacing all Unicode icon https://github.com/jellyfin/jellyfin-web/pull/582

src/components/itemhelper.js
```diff
@@ -272,7 +272,13 @@ define(['apphost', 'globalize'], function (appHost, globalize) {
...
         canRate: function (item) {
 
-            if (item.Type === 'Program' || item.Type === 'Timer' || item.Type === 'SeriesTimer' || item.Type === 'CollectionFolder' || item.Type === 'UserView' || item.Type === 'Channel') {
+            if (item.Type === 'Program' ||
+                item.Type === 'Timer' ||
+                item.Type === 'SeriesTimer' ||
+                item.Type === 'CollectionFolder' ||
+                item.Type === 'UserView' ||
+                item.Type === 'Channel' ||
+                !item.UserData) {
```
dkanada: Can you put the operators at the beginning of the line rather than the end?

grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/548/commits/b77143797c6b846f0c6074485b2acc899d4eb746

PR #549, opened 2019-10-23
Move bower_components libraries from source control to npm

grafixeyehero:
- [x] alameda 1.3.0

-  [x] requirejs 2.3.5


dkanada:
I think we can avoid pagejs and query string for now since those two require a lot more work than the rest. We can focus on this one when some of your earlier pull requests get merged in, there's a bit of a backlog.

assets.js
```diff
@@ -0,0 +1,14 @@
+//const CSS = [
+
+//];
```
dkanada: I'd remove this commented out section.

assets.js
```diff
@@ -0,0 +1,14 @@
...
+const JS = [
+    'requirejs/require.js',
+    'alameda/alameda.js',
+    'webcomponents.js-2/webcomponents-lite.min.js',
+    'document-register-element/build/document-register-element.js',
+    'native-promise-only/lib/npo.src.js',
+    'resize-observer-polyfill/dist/ResizeObserver.js',
+    'whatwg-fetch/fetch.js'
```
dkanada: I think alameda and requirejs are the only two that need this special method. The rest can just use the bundle method that @thornbill started using.

webpack.common.js
```diff
@@ -10,9 +14,18 @@ module.exports = {
...
-        }])
+        }]),
+        new CopyPlugin(
+            Assets.map(asset => {
+              return {
+                from: path.resolve(__dirname, `./node_modules/${asset}`),
+                to: path.resolve(__dirname, './dist/bower_components')
+              };
+            })
+          )
```
dkanada: Something is weird with the indentation here...

assets.js
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/549/commits/651a53d64a2d44abaa7bfd9bf1bf58e5a7fe4122

webpack.common.js
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/549/commits/651a53d64a2d44abaa7bfd9bf1bf58e5a7fe4122

assets.js
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/549/commits/651a53d64a2d44abaa7bfd9bf1bf58e5a7fe4122

webpack.common.js
```diff
@@ -10,9 +14,18 @@ module.exports = {
...
         new CopyPlugin([{
             from: "**/*",
             to: "."
-        }])
+        }]),
+        new CopyPlugin(
+            Assets.map(asset => {
+                return {
+                    from: path.resolve(__dirname, `./node_modules/${asset}`),
+                    to: path.resolve(__dirname, './dist/bower_components')
```
thornbill: I would prefer this directory be named something else to differentiate between resources that are still in `bower_components`.

dkanada: How about `dist/libraries` for things we need copied directly?

grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/549/commits/af2b49ac303d36de2c107a0eeb98968679609cce

webpack.common.js
```diff
@@ -1,6 +1,10 @@
 const path = require("path");
+const { CleanWebpackPlugin} = require("clean-webpack-plugin");
 const CopyPlugin = require("copy-webpack-plugin");
 
+// assets.js
+const Assets = require('./assets');
```
thornbill: Should we just inline the assets here? I can't think of an advantage to having them in a separate file.

grafixeyehero: @thornbill you say like this 
`const Assets = [
    'alameda/alameda.js',
    'requirejs/require.js'
];`

thornbill: Yeah I prefer that personally.

grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/549/commits/863eb067c22b45a1416800d3dc2667d896dbf12a

PR #550, opened 2019-10-24
Fix Metadata Manager Live TV Loading..

grafixeyehero:
Before
Screenshot (15)
After
Screenshot (14)


joshuaboniface:
Cherry-picked into 10.4.1.

PR #551, opened 2019-10-25
Remove dashboard icon from header

dkanada:
> I just want the two to be consistent and obvious. Removing the other button would accomplish that.

I think the best option for now is two leave both icons out of the header. This pull request has been tested locally so it works fine, and it's a small change so we could technically backport if we wanted to.

Long term, this would be an excellent candidate for the frontend configuration JSON file we were talking about on Matrix. We could even use a key-value pair or something to store the URLs and let users add custom links. I believe there was a feature request for this somewhere.

replaces #508

### DIFF
--- a/src/scripts/librarymenu.js
+++ b/src/scripts/librarymenu.js
@@ -52,14 +52,6 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
                 headerSearchButton.classList.remove("hide");
             }
 
-            if (headerSettingsButton) {
-                if (user.localUser.Policy.IsAdministrator) {
-                    headerSettingsButton.classList.remove("hide");
-                } else {
-                    headerSettingsButton.classList.add("hide");
-                }
-            }
-
             headerCastButton.classList.remove("hide");
         } else {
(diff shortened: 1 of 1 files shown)


sparky8251:
Not sure I agree with the backport label. This isn't a bug fix, its a change. It changes the way the UI behaves by removing a link entirely.

Not exactly a stable release if we start including changes and not just fixes.

dkanada:
You could consider this a usability fix in my opinion. I'll leave it up to others on whether or not we should backport this though.

KeepBotting:
I agree with this change from a usability perspective. Don't really care how/when it's added, or if it's backported or not. That seems not as relevant as how it affects UX.

Anecdotally, I *never* use the left hamburger menu, because everything is accessible from the dashboard, which I navigate to by using the button in the upper-right.

Thinking about it, it makes more sense to place navigation buttons in the left menu. The upper-right seems better suited to *operation* buttons, like Search or Cast.

joshuaboniface:
@sparky8251 It's trivial enough to be backported without impacy, as a "usability fix" as @dkanada mentions, but I don't care much either way.

Issue #504, opened 2019-10-07
Unable to select saved login+password in login form using arrow


agilob:
**Describe the bug**
I use keyboard to navigate webui and can't login to jellyfin by selecting username+password using arrows

**To Reproduce**
https://gfycat.com/greenunrulyhalicore

Start typing your username, as autocomplete arrives, press arrow down to select it from the list

**Expected behavior**
Pressing arrow down navigates to a field below.

**System (please complete the following information):**
 - OS: docker
 - Browser:  Firefox
 - Jellyfin Version: 10.4.0

agilob:
Standard way to navigate between fields is to use tab [down] and tab+shift [up]

anthonylavado:
@JustAMan I think it's definitely the keyboard nav. It's probably active on all screens, including this one, so that it could be navigated with a remote.

Relevant file: https://github.com/jellyfin/jellyfin-web/blob/release-10.4.z/src/components/keyboardnavigation.js

agilob:
Still a bug and regression.

JustAMan:
@dmitrylyzo does this ring any bells for your experiments with TV + keyboard support?

dkanada:
I don't personally consider this a regression, it was a concious decision to disable keyboard navigation for the desktop layout. My personal solution for this would be an additional preference to manually enable keyboard navigation rather than enable it globally for the desktop layout, but that's open to discussion.

JustAMan:
The problem here is that keyboard navigation looks to be _not_ disabled here, but rather uses TV remote-like behaviour (where arrow keys move focus around instead of missing-on-the-remote Tab).

agilob:
This is broken user experience comparing to previous versions and unmet  expectations from generally known web from behaviour. Navigation with tab / tab+shift is a common thing that works across all devices, whether you're in a browser, desktop app or mobile, it always works the same when using keyboard.

Here I would like to add that this affects not only login form but all forms in jellyfin, library management too.

dmitrylyzo:
@JustAMan 
I mainly use the keyboard for navigation only. On a real TV I use a remote control (keyboard for some tests and when I am stuck with a broken focus). So, I almost do not edit the inputs.

Tizen 4 (real)/5: no autocompletion, no dropdown list.
WebOS 3/4: autocompletion is part of screen keyboard, no dropdown list.

No dropdown lists - no problem (almost).

Why "almost"?
[...]

PR #563, opened 2019-11-02
Add autofocus on page load

dmitrylyzo:
As continue of #314 (#303 as main goal).

When view/page is loaded, focus usually is on `body`. When navigating with the keyboard, focus moves to the element at the bottom of the current view (at least it looks like this). To fix focus after page is loaded, I added `AutoFocuser`. :rofl:

**Detailed description**
In Tizen App, I fix focus in the function `onPageLoad` watching for style of loading element (https://github.com/dmitrylyzo/jellyfin-tizen/blob/ea271573c1342276f63f13b9ee04aa5cadd5d54f/tizen.js#L82-L128). In my first PR, we slightly discuss this approach (https://github.com/jellyfin/jellyfin-web/pull/522#discussion_r334903850), ended with the need for some event. At some point, event will be useful, but for now I decided to use some shared module directly.

Actually, `FocusManager` is the best candidate because focus is its duty (`ViewManager` or `loading` could be used for event emitting). But I decided not to clog `FocusManger` with some site specific classes (`btnPlay`, `btnNextPage`, etc.) and logic, and extracted feature in a separate module, `AutoFocuser`. If you think that it should be integrated into `FocusManager`, it will be integrated.

* For now, focus is fixed on pages which I can check somehow.
* `AutoFocuser` restores navigating buttons (`btnPreviousPage`, `btnNextPage`) for "unstoppable" library navigating.
* Play/Resume focus moved to it too.

**Issues**
1. Not all pages are fixed. Maybe some of them do not require it.
* `moviegenres` and `tvgenres` - could not find a place to call `AutoFocuser.autoFocus`.
* `livetv` group, `movietrailers` and `tvupcoming` - skipped due to no examples for check.

2. In pages with item listing, `FocusManager.autoFocus` moves focus on the "toolbuttons" (`btnNextPage`, `btnSelectView`) since they are not `noautofocus`. In my first PR, I have to mark all "toolbuttons" with `noautofocus` and swap alphaPicker (maybe `noautofocus` on each letter helps too).

3. Some pages do not use `AutoFocuser`:
* Settings loaded by the helper object - `autoFocus` option on it.

4. On "Profile" page, `AutoFocuser.autoFocus` called by UserPasswordPage.

### DIFF
--- a/src/components/autoFocuser.js
+++ b/src/components/autoFocuser.js
@@ -29,6 +29,13 @@ define(["focusManager", "layoutManager"], function (focusManager, layoutManager)
     }
 
     /**
+     * Create an array from some source.
+     */
+    var arrayFrom = Array.prototype.from || function (src) {
+        return Array.prototype.slice.call(src);
+    }
+
+    /**
      * Set focus on a suitable element, taking into account the previously selected.
      */
     function autoFocus(container) {
@@ -55,8 +62,8 @@ define(["focusManager", "layoutManager"], function (focusManager, layoutManager)
             candidates.push(activeElement);
         }
 
-        candidates = candidates.concat(Array.from(container.querySelectorAll(".btnResume")));
-        candidates = candidates.concat(Array.from(container.querySelectorAll(".btnPlay")));
+        candidates = candidates.concat(arrayFrom(container.querySelectorAll(".btnResume")));
+        candidates = candidates.concat(arrayFrom(container.querySelectorAll(".btnPlay")));
 
         var notFound = candidates.every(function (element) {
             if (focusManager.isCurrentlyFocusable(element)) {
--- a/src/components/autoFocuser.js
+++ b/src/components/autoFocuser.js
@@ -53,6 +53,15 @@ define(["focusManager"], function (focusManager) {
         });
 
         if (notFound) {
+            // FIXME: Multiple itemsContainers
+            var itemsContainer = container.querySelector(".itemsContainer");
+
+            if (itemsContainer) {
+                notFound = !focusManager.autoFocus(itemsContainer);
+            }
+        }
+
+        if (notFound) {
             focusManager.autoFocus(container);
         }
     }
--- a/src/components/autoFocuser.js
+++ b/src/components/autoFocuser.js
@@ -65,27 +65,32 @@ define(["focusManager", "layoutManager"], function (focusManager, layoutManager)
         candidates = candidates.concat(arrayFrom(container.querySelectorAll(".btnResume")));
         candidates = candidates.concat(arrayFrom(container.querySelectorAll(".btnPlay")));
 
-        var notFound = candidates.every(function (element) {
+        var focusedElement;
+
+        candidates.every(function (element) {
             if (focusManager.isCurrentlyFocusable(element)) {
                 focusManager.focus(element);
+                focusedElement = element;
                 return false;
             }
 
             return true;
         });
 
-        if (notFound) {
+        if (!focusedElement) {
             // FIXME: Multiple itemsContainers
             var itemsContainer = container.querySelector(".itemsContainer");
 
             if (itemsContainer) {
-                notFound = !focusManager.autoFocus(itemsContainer);
+                focusedElement = focusManager.autoFocus(itemsContainer);
             }
         }
 
-        if (notFound) {
-            focusManager.autoFocus(container);
+        if (!focusedElement) {
+            focusedElement = focusManager.autoFocus(container);
         }
+
+        return focusedElement;
     }
 
     return {
--- a/src/components/autoFocuser.js
+++ b/src/components/autoFocuser.js
@@ -1,4 +1,4 @@
-define(["focusManager"], function (focusManager) {
+define(["focusManager", "layoutManager"], function (focusManager, layoutManager) {
     "use strict";
 
     /**
@@ -7,9 +7,20 @@ define(["focusManager"], function (focusManager) {
     var activeElement;
 
     /**
+     * Returns true if AutoFocuser is enabled.
+     */
+    function isEnabled() {
+        return layoutManager.tv;
+    };
+
+    /**
      * Start AutoFocuser
      */
     function enable() {
+        if (!isEnabled()) {
+            return;
+        }
+
         window.addEventListener("focusin", function (e) {
             activeElement = e.target;
         });
@@ -21,6 +32,10 @@ define(["focusManager"], function (focusManager) {
      * Set focus on a suitable element, taking into account the previously selected.
      */
     function autoFocus(container) {
+        if (!isEnabled()) {
+            return;
+        }
+
         container = container || document.body;
 
         var candidates = [];
@@ -67,6 +82,7 @@ define(["focusManager"], function (focusManager) {
     }
 
     return {
+        isEnabled: isEnabled,
         enable: enable,
         autoFocus: autoFocus
     };
--- a/src/components/autoFocuser.js
+++ b/src/components/autoFocuser.js
@@ -1,0 +1,64 @@
+define(["focusManager"], function (focusManager) {
+    "use strict";
+
+    /**
+     * Previously selected element.
+     */
+    var activeElement;
+
+    /**
+     * Start AutoFocuser
+     */
+    function enable() {
+        window.addEventListener("focusin", function (e) {
+            activeElement = e.target;
+        });
+
+        console.log("AutoFocuser enabled");
+    }
+
+    /**
+     * Set focus on a suitable element, taking into account the previously selected.
+     */
+    function autoFocus(container) {
+        container = container || document.body;
+
+        var candidates = [];
+
+        if (activeElement) {
+            // These elements are recreated
+            if (activeElement.classList.contains("btnPreviousPage")) {
+                candidates.push(container.querySelector(".btnPreviousPage"));
+                candidates.push(container.querySelector(".btnNextPage"));
+            } else if (activeElement.classList.contains("btnNextPage")) {
+                candidates.push(container.querySelector(".btnNextPage"));
+                candidates.push(container.querySelector(".btnPreviousPage"));
+            } else if (activeElement.classList.contains("btnSelectView")) {
+                candidates.push(container.querySelector(".btnSelectView"));
+            }
+
+            candidates.push(activeElement);
+        }
+
+        candidates = candidates.concat(Array.from(container.querySelectorAll(".btnResume")));
+        candidates = candidates.concat(Array.from(container.querySelectorAll(".btnPlay")));
+
+        var notFound = candidates.every(function (element) {
+            if (focusManager.isCurrentlyFocusable(element)) {
+                focusManager.focus(element);
+                return false;
+            }
+
+            return true;
+        });
+
+        if (notFound) {
+            focusManager.autoFocus(container);
+        }
+    }
+
+    return {
+        enable: enable,
+        autoFocus: autoFocus
+    };
+});
--- a/src/components/autoFocuser.js
+++ b/src/components/autoFocuser.js
@@ -11,7 +11,7 @@ define(["focusManager", "layoutManager"], function (focusManager, layoutManager)
      */
     function isEnabled() {
         return layoutManager.tv;
-    };
+    }
 
     /**
      * Start AutoFocuser
--- a/src/controllers/itemdetailpage.js
+++ b/src/controllers/itemdetailpage.js
@@ -1715,6 +1715,12 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "cardBuild
             hideAll(page, "btnPlay", false);
             hideAll(page, "btnShuffle", false);
         }
+
+        // HACK: Call autoFocuser again because btnPlay may be hidden, but focused by reloadFromItem
+        // FIXME: Sometimes focus does not move until all (?) sections are loaded
+        require(["autoFocuser"], function (autoFocuser) {
+            autoFocuser.autoFocus(page);
+        });
     }
 
     function renderCollectionItemType(page, parentItem, type, items) {
--- a/src/controllers/itemdetailpage.js
+++ b/src/controllers/itemdetailpage.js
@@ -617,24 +617,9 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "cardBuild
             hideAll(page, "btnDownload", true);
         }
 
-        try {
-            require(["focusManager"], function (focusManager) {
-                [".btnResume", ".btnPlay"].every(function (cls) {
-                    var elems = page.querySelectorAll(cls);
-
-                    for (var i = 0; i < elems.length; i++) {
-                        if (focusManager.isCurrentlyFocusable(elems[i])) {
-                            focusManager.focus(elems[i]);
-                            return false;
-                        }
-                    }
-
-                    return true;
-                });
-            });
-        } catch (e) {
-            console.log(e);
-        }
+        require(["autoFocuser"], function (autoFocuser) {
+            autoFocuser.autoFocus(page);
+        });
     }
 
     function logoImageUrl(item, apiClient, options) {
--- a/src/controllers/movies/moviesrecommended.js
+++ b/src/controllers/movies/moviesrecommended.js
@@ -36,6 +36,9 @@ define(["events", "layoutManager", "inputManager", "userSettings", "libraryMenu"
                 showYear: true,
                 centerText: true
             });
+
+            // FIXME: Wait for all sections to load
+            autoFocus(page);
         });
     }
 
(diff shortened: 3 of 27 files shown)


src/controllers/itemdetailpage.js
```diff
@@ -341,19 +341,8 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "cardBuild
...
-                    for (var i = 0; i < elems.length; i++) {
-                        if (focusManager.isCurrentlyFocusable(elems[i])) {
-                            focusManager.focus(elems[i]);
-                            return false;
-                        }
-                    }
-                    return true;
-                });
+            require(["autoFocuser"], function(autoFocuser) {
+                autoFocuser.autoFocus(page);
```
dkanada: Can the try-catch block can be removed here since you added a separate handler?

src/components/autoFocuser.js
```diff
@@ -0,0 +1,67 @@
...
+                candidates.push(container.querySelector(".btnSelectView"));
+            }
+
+            candidates.push(activeElement);
+        }
+
+        candidates = candidates.concat(Array.from(container.querySelectorAll(".btnResume")));
+        candidates = candidates.concat(Array.from(container.querySelectorAll(".btnPlay")));
+
+        var notfound = candidates.every(function(element) {
```
dkanada: use camelCase

src/components/autoFocuser.js
```diff
@@ -0,0 +1,67 @@
...
+                candidates.push(container.querySelector(".btnPreviousPage"));
+            } else if (activeElement.classList.contains("btnSelectView")) {
+                candidates.push(container.querySelector(".btnSelectView"));
+            }
+
+            candidates.push(activeElement);
+        }
+
+        candidates = candidates.concat(Array.from(container.querySelectorAll(".btnResume")));
+        candidates = candidates.concat(Array.from(container.querySelectorAll(".btnPlay")));
```
dkanada: Can we avoid the array creation here somehow?

src/controllers/user/display.js
```diff
@@ -13,7 +13,8 @@ define(["displaySettings", "userSettingsBuilder", "userSettings"], function(Disp
                 element: view.querySelector(".settingsContainer"),
                 userSettings: userSettings,
                 enableSaveButton: !1,
-                enableSaveConfirmation: !1
+                enableSaveConfirmation: !1,
+                autoFocus: true
```
dkanada: Where is this value checked?

dmitrylyzo: https://github.com/jellyfin/jellyfin-web/blob/c10f3feb3149d761c993a0c3b007c09c8a848f14/src/components/displaysettings/displaysettings.js#L295-L297

dkanada:
The focus seems to work fine on my computer in TV mode, is this a browser issue? Maybe it should only be enabled on TV devices.

dmitrylyzo:
> The focus seems to work fine on my computer in TV mode, is this a browser issue?

If you mean focus on `body`. _(in all cases, keyboard navigation only)_
In Chrome 78.0.3904.87 in TV mode, when I open "Music/TV Shows/Movies", focus is nowhere (actually on `body`). Then I navigate down, focus moves to first element of second row.
In settings (for example, Home), focus is nowhere again. Then I navigate down, focus moves on 4-th section.
In Firefox 70.0.1, after page loads, focus is on last clicked element (not seen). Then I navigate down, focus moves to "Back" button.

I believe that user go to library to watch/listen some item, so focus must be on first element of library as the most likely option. Same for settings.

In addition, when using paging buttons, they are recreated, so focus moves to body in Chrome and Firefox.

> Maybe it should only be enabled on TV devices.

If keyboard navigation should be used on TV only, then sure yes.
At current moment, `keyboardnavigation` is enabled everywhere. In non-TV modes, there is no focus indication, so probably it should be turned off as well. Moreover, this creates some inconvinience #504.

src/components/autoFocuser.js
dmitrylyzo: Do you mean `Array.from`?
Or `candidates` itself?

dkanada: Array.from

dmitrylyzo: Currently, there is one `btnResume` on `itemdetails.html`, so I can change it to
```
candidates.push(container.querySelector(".btnResume"));
```
But there is two `btnPlay`. One of them (the first one) is hidden in TV layout in `itemBackdrop`.

Are you worried about performance?

dkanada:
I did a bit of testing on master and agree this is needed for the TV layout, what else is left to do for this pull request?

dmitrylyzo:
> what else is left to do for this pull request?

- I skiped some pages (see PR description). Maybe missed some of them. Actually, some pages do not require autoFocus by now, since they are called by tabs, and focus will be on tab button.

- It do not focus on first card as I want. There is 3 options:
1. Make all "toolbuttons" (prev/next page, sort, select view, etc.) as `noautofocus`. _Old approach_
2. In `AutoFocuser`: if `notFound`, find first occurence of `itemsContainer`, and call autoFocus on it instead of page. _There may be several `itemsContainer` on page_
```
        if (notFound) {
            var itemsContainer = container.querySelector(".itemsContainer");

            if (itemsContainer) {
                notFound = !focusManager.autoFocus(itemsContainer);
            }
        }

        if (notFound) {
            focusManager.autoFocus(container);
        }
 ```
3. Pass prefered container as second param.

**Comment by dkanada** [comment]
> some pages do not require autoFocus by now

Could you check for this case? No sense in focusing when it isn't required.

> focus on first card

I actually think we should keep focus on these buttons rather than move focus to the first card. Navigation buttons often need to be pressed more than once so moving focus away wouldn't be ideal.

Yeah, I think for now the AutoFocuser and keyboard navigation should only be enabled for TV layouts. I'm not a fan of partially supporting keyboard navigation from the desktop layout, so we would need to support it everywhere. That would be done in a separate pull request though.

**Comment by dmitrylyzo** [comment]
> Navigation buttons often need to be pressed more than once so moving focus away wouldn't be ideal.

Navigation buttons have a priority if they have been focused before. In other words, the first card should be focused if no other element was focused before.

**Comment by dkanada** [comment]
I think option two might be best then.

**Comment by dmitrylyzo** [comment]
Rebased.

<details>
<summary>Checked pages</summary>

```
addserver - opens with refresh
itemdetailpage - opens with refresh
loginpage - opens with refresh

livetv/
    livetvsuggested - opens with refresh
    livetvchannels - have arrows

movies/
    moviecollection - may have arrows (judging by the source)
    movierecommended - opens with refresh (user settings)
    movies - opens with refresh

music/
    musicalbums - have arrows / opens with refresh (user settings)
    musicartists - have arrows / opens with refresh (user settings)
    musicgenres - opens with refresh (user settings)
    musicplaylists - opens with refresh (user settings)
    musicrecommended - opens with refresh
    songs - have arrows

shows/
    episodes - have arrows
    tvlatest - opens with refresh (user settings)
    tvrecommended - opens with refresh (user settings)
    tvshows - opens with refresh
*   tvstudios - not sure. can it have arrow buttons?

user/
    display - opens with refresh
    home - opens with refresh
    menu - opens with refresh
    playback - opens with refresh
    subtitles - opens with refresh
    userpasswordpage (profile) - opens with refresh
```
</details>

As I said, some pages are skipped due to the inability to check, or I do not know where `AutoFocuser.autoFocus` call should be, or it seems that they do not require autofocus yet. Also, I could have missed some.

- `moviegenres` - accessed through tab (autofocus is not required).
- `tvgenres` -  accessed through tab (autofocus is not required).
- `movietrailers` - skipped due to no examples for check.
- `tvupcoming` - skipped due to no examples for check.
- `livetvguide` uses `guide` component. Should I call `AutoFocuser` from components?
- Component `CollectionEditor` (creation of movie collection). Component again.

**Some issues**

- Commented hack in `src/controllers/itemdetailpage.js`
_Sometimes it is not reproducible_
1. Go to "Movies/Collections"
2. Click "+" (add)
3. Enter title, then OK
4. Collection opened, focus placed (usually) on some button (on Desktop browsers, focus on `itemDetailImage`).
5. Back and repeat 2,3
6. Collection opens with no focus. Focus is placed after some delay.

- _Not related to AutoFocuser - reprodusable on clean jellyfin-web_
1. Go to "Home" user settings.
2. Place focus on some top select (for example, "TV mode home screen")
3. Scroll down and click on other select (for example, Movies/Default screen)
4. Leave select screen (back or select some item)
5. Window scrolls to the focused top element

This is due to the fact that select does not focus on the click.

**Comment by JustAMan** [comment]
@dmitrylyzo I take it doesn't impact non-TV state currently, and improves quality of life for TV, even if not to ideal state, right? If so, I feel we should review and merge it.

**Comment by dmitrylyzo** [comment]
@JustAMan,
Yes, it is off in a non-TV layout. Most of the pages that I can reach is ready (see "Checked page" spoiler above).

File: src/components/autoFocuser.js [hunk 5]
```diff
@@ -0,0 +1,96 @@
...
+            return;
+        }
+
+        container = container || document.body;
+
+        var candidates = [];
+
+        if (activeElement) {
+            // These elements are recreated
+            if (activeElement.classList.contains("btnPreviousPage")) {
```
JustAMan: can we do this as a loop over an array of all potential candidates handled similar here?

src/components/autoFocuser.js
```diff
@@ -0,0 +1,96 @@
...
+                candidates.push(container.querySelector(".btnSelectView"));
+            }
+
+            candidates.push(activeElement);
+        }
+
+        candidates = candidates.concat(arrayFrom(container.querySelectorAll(".btnResume")));
+        candidates = candidates.concat(arrayFrom(container.querySelectorAll(".btnPlay")));
+
+        var notFound = candidates.every(function (element) {
```
JustAMan: `notFound` -> `noFocusableElements`?

src/controllers/movies/moviesrecommended.js
```diff
@@ -36,6 +36,9 @@ define(["events", "layoutManager", "inputManager", "userSettings", "libraryMenu"
                 showYear: true,
                 centerText: true
             });
+
+            // FIXME: Wait for all sections to load
```
JustAMan: do you have any idea on how to fix it?

src/controllers/user/display.js
```diff
@@ -24,7 +24,8 @@ define(["displaySettings", "userSettingsBuilder", "userSettings"], function (Dis
                     element: view.querySelector(".settingsContainer"),
                     userSettings: userSettings,
                     enableSaveButton: false,
-                    enableSaveConfirmation: false
+                    enableSaveConfirmation: false,
+                    autoFocus: autoFocuser.isEnabled()
```
JustAMan: why do we need this?

src/controllers/user/home.js
```diff
@@ -24,7 +24,8 @@ define(["homescreenSettings", "userSettingsBuilder", "dom", "globalize", "loadin
                     element: view.querySelector(".homeScreenSettingsContainer"),
                     userSettings: userSettings,
                     enableSaveButton: false,
-                    enableSaveConfirmation: false
+                    enableSaveConfirmation: false,
+                    autoFocus: autoFocuser.isEnabled()
```
JustAMan: why do we need it?

src/controllers/user/playback.js
```diff
@@ -24,7 +24,8 @@ define(["playbackSettings", "userSettingsBuilder", "dom", "globalize", "loading"
                     element: view.querySelector(".settingsContainer"),
                     userSettings: userSettings,
                     enableSaveButton: false,
-                    enableSaveConfirmation: false
+                    enableSaveConfirmation: false,
+                    autoFocus: autoFocuser.isEnabled()
```
JustAMan: again, why?

src/controllers/user/subtitles.js
```diff
@@ -24,7 +24,8 @@ define(["subtitleSettings", "userSettingsBuilder", "userSettings"], function (Su
                     element: view.querySelector(".settingsContainer"),
                     userSettings: userSettings,
                     enableSaveButton: false,
-                    enableSaveConfirmation: false
+                    enableSaveConfirmation: false,
+                    autoFocus: autoFocuser.isEnabled()
```
JustAMan: why?

src/controllers/user/display.js
dmitrylyzo: Otherwise, it will focus on non-TV. Initially it was without `autoFocus`, and it is currently in sync with `AutoFocuser` which is enabled on TV only.

src/controllers/movies/moviesrecommended.js
dmitrylyzo: 1. Dirty way: add count of sections to load (like semaphore).
2. Make all `load*` functions to return `Promise` and wait for all in `loadSuggestionsTab`.
All above is not tested - only thoughts.

src/controllers/user/display.js
JustAMan: oh, so `autoFocus` is an already existing value?

src/components/autoFocuser.js
dmitrylyzo: I planned:
if `btnPreviousPage` was focused then it should be focused next, but if it is disabled, then `btnNextPage` should. And vice versa. Then user stays on paging buttons for as long as necessary.

To make this logic as loop :thinking: :
```
var focusRules = [
    {
        prevCss: ".btnPreviousPage",
        nextCss: [".btnNextPage", ".btnPreviousPage"]
    },
    {
        prevCss: ".btnNextPage",
        nextCss: [".btnPreviousPage", ".btnNextPage"]
    },
    {
        prevCss: ".btnSelectView"
        // Suppose nextCss: [prevCss]
    },

    // Here we must check activeElement as it is
    {}, // unobvious HACK :)

    // If no prevCss, then this could be perceived as "no activeElement"
    {
        nextCss: [".btnResume", ".btnPlay"]
    }
];
```

src/controllers/user/display.js
dmitrylyzo: Yes, it has not been used here before.
https://github.com/jellyfin/jellyfin-web/blob/c10f3feb3149d761c993a0c3b007c09c8a848f14/src/components/displaysettings/displaysettings.js#L295-L297

src/controllers/user/home.js
dmitrylyzo: https://github.com/jellyfin/jellyfin-web/blob/1d0e56c0ae5f74d38ea336cd10ffb9a276069f48/src/components/homescreensettings/homescreensettings.js#L525-L527

src/controllers/user/playback.js
dmitrylyzo: https://github.com/jellyfin/jellyfin-web/blob/1d0e56c0ae5f74d38ea336cd10ffb9a276069f48/src/components/playbacksettings/playbacksettings.js#L300-L302

src/controllers/user/subtitles.js
dmitrylyzo: https://github.com/jellyfin/jellyfin-web/blob/1d0e56c0ae5f74d38ea336cd10ffb9a276069f48/src/components/subtitlesettings/subtitlesettings.js#L175-L177

src/components/autoFocuser.js
JustAMan: okay, I see your point... I was thinking too Pythonic here, js arrays and for-s might be too strict to implement

<details><summary>Python illustration</summary>

```python
for prevCss, nextCss in ((".btnPreviousPage", (".btnNextPage", ".btnPreviousPage")),
                         (".btnNextPage", (".btnPreviousPage", ".btnNextPage")),
                         (".btnSelectView", (".btnSelectView",))):
    if activeElement.classList.contains(prevCss):
        for nextItem in nextCss:
            candidates.append(container.querySelector(nextItem))
```

</details>

src/components/autoFocuser.js
dmitrylyzo: How about this?
```
        var focusedElement;

        candidates.every(function (element) {
            if (focusManager.isCurrentlyFocusable(element)) {
                focusManager.focus(element);
                focusedElement = element;
                return false;
            }

            return true;
        });

        if (!focusedElement) {
            // FIXME: Multiple itemsContainers
            var itemsContainer = container.querySelector(".itemsContainer");

            if (itemsContainer) {
                focusedElement = focusManager.autoFocus(itemsContainer);
            }
        }

        if (!focusedElement) {
            focusedElement = focusManager.autoFocus(container);
        }

        return focusedElement;
```

JustAMan: Cool, I like this!

src/components/autoFocuser.js
```diff
@@ -65,27 +65,32 @@ define(["focusManager", "layoutManager"], function (focusManager, layoutManager)
         candidates = candidates.concat(arrayFrom(container.querySelectorAll(".btnResume")));
         candidates = candidates.concat(arrayFrom(container.querySelectorAll(".btnPlay")));
 
-        var notFound = candidates.every(function (element) {
+        var focusedElement;
+
+        candidates.every(function (element) {
             if (focusManager.isCurrentlyFocusable(element)) {
```
JustAMan: shouldn't you check here if `focusedElement` is still false-y?

dmitrylyzo: `return false` in callback stops `every`.

PR #566, opened 2019-11-03
Make navigation and playback control TV-friendly

dmitrylyzo:
On the TV, keyboard navigation during playback is broken, because at the same time it controls playback.

* I left all the keys except the navigation ones as they are.
* Made possible to seek with the keyboard like with a mouse ("keyboard dragging"), when slider is focused (I made this the only focusable in TV mode).
* "Keyboard dragging" is allways enabled, even in Desktop mode (but this also works), because focus is set on click.
* There is a volume slider. I tried to make it focusable too, but then you are not be able to navigate from it horizontally.
* Cannot check video playback in webOS 4 emulator (not playing :disappointed:), but visually keyboard dragging works. In Tizen 4/5, good.
* **LiveTV is not tested.**

### DIFF
--- a/src/components/emby-slider/emby-slider.css
+++ b/src/components/emby-slider/emby-slider.css
@@ -87,6 +87,10 @@ _:-ms-input-placeholder {
     transform: scale(1.6);
 }
 
+.mdl-slider.show-focus:focus::-webkit-slider-thumb {
+    transform: scale(1.6);
+}
+
 .slider-no-webkit-thumb::-webkit-slider-thumb {
     opacity: 0 !important;
 }
--- a/src/components/emby-slider/emby-slider.js
+++ b/src/components/emby-slider/emby-slider.js
@@ -250,7 +250,6 @@ define(['browser', 'dom', 'layoutManager', 'css!./emby-slider', 'registerElement
      * Handle KeyDown event
      */
     function onKeyDown(e) {
-
         switch (e.key) {
             case 'ArrowLeft':
             case 'Left':
@@ -258,7 +257,6 @@ define(['browser', 'dom', 'layoutManager', 'css!./emby-slider', 'registerElement
                 e.preventDefault();
                 e.stopPropagation();
                 break;
-
             case 'ArrowRight':
             case 'Right':
                 stepKeyboard(this, this.keyboardStepUp || 1);
--- a/src/components/emby-slider/emby-slider.js
+++ b/src/components/emby-slider/emby-slider.js
@@ -19,6 +19,11 @@ define(['browser', 'dom', 'layoutManager', 'css!./emby-slider', 'registerElement
 
     function updateValues() {
 
+        // Do not update values when dragging with keyboard to keep current progress for reference
+        if (!!this.keyboardDragging) {
+            return;
+        }
+
         var range = this;
         var value = range.value;
 
@@ -81,6 +86,9 @@ define(['browser', 'dom', 'layoutManager', 'css!./emby-slider', 'registerElement
         }
         if (!layoutManager.mobile) {
             this.classList.add('mdl-slider-hoverthumb');
+        }
+        if (layoutManager.tv) {
+            this.classList.add('show-focus');
         }
 
         var containerElement = this.parentNode;
@@ -176,6 +184,116 @@ define(['browser', 'dom', 'layoutManager', 'css!./emby-slider', 'registerElement
             }
         }
     };
+
+    /**
+     * Keyboard dragging timeout.
+     * After this delay "change" event will be fired.
+     */
+    var KeyboardDraggingTimeout = 1000;
+
+    /**
+     * Keyboard dragging timer.
+     */
+    var keyboardDraggingTimer;
+
+    /**
+     * Start keyboard dragging.
+     *
+     * @param {Object} elem slider itself
+     */
+    function startKeyboardDragging(elem) {
+
+        elem.keyboardDragging = true;
+
+        clearTimeout(keyboardDraggingTimer);
+
+        keyboardDraggingTimer = setTimeout(function () {
+            finishKeyboardDragging(elem);
+        }, KeyboardDraggingTimeout);
+    }
+
+    /**
+     * Finish keyboard dragging.
+     *
+     * @param {Object} elem slider itself
+     */
+    function finishKeyboardDragging(elem) {
+
+        clearTimeout(keyboardDraggingTimer);
+        keyboardDraggingTimer = undefined;
+
+        elem.keyboardDragging = false;
+
+        var event = new Event('change', {
+            bubbles: true,
+            cancelable: false
+        });
+        elem.dispatchEvent(event);
+    }
+
+    /**
+     * Do step by delta.
+     *
+     * @param {Object} elem slider itself
+     * @param {number} delta step amount
+     */
+    function stepKeyboard(elem, delta) {
+
+        startKeyboardDragging(elem);
+
+        elem.value = Math.max(elem.min, Math.min(elem.max, parseFloat(elem.value) + delta));
+
+        var event = new Event('input', {
+            bubbles: true,
+            cancelable: false
+        });
+        elem.dispatchEvent(event);
+    }
+
+    /**
+     * Handle KeyDown event
+     */
+    function onKeyDown(e) {
+
+        switch (e.key) {
+            case 'ArrowLeft':
+            case 'Left':
+                stepKeyboard(this, -this.keyboardStepDown || -1);
+                e.preventDefault();
+                e.stopPropagation();
+            break;
+
+            case 'ArrowRight':
+            case 'Right':
+                stepKeyboard(this, this.keyboardStepUp || 1);
+                e.preventDefault();
+                e.stopPropagation();
+            break;
+        }
+    }
+
+    /**
+     * Enable keyboard dragging.
+     */
+    EmbySliderPrototype.enableKeyboardDragging = function () {
+
+        if (!this.keyboardDraggingEnabled) {
+            this.addEventListener('keydown', onKeyDown);
+            this.keyboardDraggingEnabled = true;
+        }
+    }
+
+    /**
+     * Set steps for keyboard input.
+     *
+     * @param {number} stepDown step to reduce
+     * @param {number} stepUp step to increase
+     */
+    EmbySliderPrototype.setKeyboardSteps = function (stepDown, stepUp) {
+
+        this.keyboardStepDown = stepDown || stepUp || 1;
+        this.keyboardStepUp = stepUp || stepDown || 1;
+    }
 
     function setRange(elem, startPercent, endPercent) {
 
--- a/src/components/emby-slider/emby-slider.js
+++ b/src/components/emby-slider/emby-slider.js
@@ -202,11 +202,9 @@ define(['browser', 'dom', 'layoutManager', 'css!./emby-slider', 'registerElement
      * @param {Object} elem slider itself
      */
     function startKeyboardDragging(elem) {
-
         elem.keyboardDragging = true;
 
         clearTimeout(keyboardDraggingTimer);
-
         keyboardDraggingTimer = setTimeout(function () {
             finishKeyboardDragging(elem);
         }, KeyboardDraggingTimeout);
@@ -218,7 +216,6 @@ define(['browser', 'dom', 'layoutManager', 'css!./emby-slider', 'registerElement
      * @param {Object} elem slider itself
      */
     function finishKeyboardDragging(elem) {
-
         clearTimeout(keyboardDraggingTimer);
         keyboardDraggingTimer = undefined;
 
@@ -238,7 +235,6 @@ define(['browser', 'dom', 'layoutManager', 'css!./emby-slider', 'registerElement
      * @param {number} delta step amount
      */
     function stepKeyboard(elem, delta) {
-
         startKeyboardDragging(elem);
 
         elem.value = Math.max(elem.min, Math.min(elem.max, parseFloat(elem.value) + delta));
@@ -261,14 +257,14 @@ define(['browser', 'dom', 'layoutManager', 'css!./emby-slider', 'registerElement
                 stepKeyboard(this, -this.keyboardStepDown || -1);
                 e.preventDefault();
                 e.stopPropagation();
-            break;
+                break;
 
             case 'ArrowRight':
             case 'Right':
                 stepKeyboard(this, this.keyboardStepUp || 1);
                 e.preventDefault();
                 e.stopPropagation();
-            break;
+                break;
         }
     }
 
@@ -276,7 +272,6 @@ define(['browser', 'dom', 'layoutManager', 'css!./emby-slider', 'registerElement
      * Enable keyboard dragging.
      */
     EmbySliderPrototype.enableKeyboardDragging = function () {
-
         if (!this.keyboardDraggingEnabled) {
             this.addEventListener('keydown', onKeyDown);
             this.keyboardDraggingEnabled = true;
@@ -290,7 +285,6 @@ define(['browser', 'dom', 'layoutManager', 'css!./emby-slider', 'registerElement
      * @param {number} stepUp step to increase
      */
     EmbySliderPrototype.setKeyboardSteps = function (stepDown, stepUp) {
-
         this.keyboardStepDown = stepDown || stepUp || 1;
         this.keyboardStepUp = stepUp || stepDown || 1;
     }
--- a/src/controllers/videoosd.js
+++ b/src/controllers/videoosd.js
@@ -1083,11 +1083,13 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
         function onWindowKeyDown(e) {
             if (!currentVisibleMenu && 32 === e.keyCode) {
                 playbackManager.playPause(currentPlayer);
-                return void showOsd();
+                showOsd();
+                return;
             }
 
             if (layoutManager.tv && NavigationKeys.indexOf(e.key) != -1) {
-                return void showOsd();
+                showOsd();
+                return;
             }
 
             switch (e.key) {
--- a/src/controllers/videoosd.js
+++ b/src/controllers/videoosd.js
@@ -770,6 +770,10 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
             var isProgressClear = state.MediaSource && null == state.MediaSource.RunTimeTicks;
             nowPlayingPositionSlider.setIsClear(isProgressClear);
 
+            if (nowPlayingItem.RunTimeTicks) {
+                nowPlayingPositionSlider.setKeyboardSteps(userSettings.skipBackLength() * 1000000 / nowPlayingItem.RunTimeTicks, userSettings.skipForwardLength() * 1000000 / nowPlayingItem.RunTimeTicks);
+            }
+
             if (-1 === supportedCommands.indexOf("ToggleFullscreen") || player.isLocalPlayer && layoutManager.tv && playbackManager.isFullscreen(player)) {
                 view.querySelector(".btnFullscreen").classList.add("hide");
             } else {
@@ -1070,9 +1074,18 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
             }
         }
 
+        /**
+         * Keys used for keyboard navigation.
+         */
+        var NavigationKeys = ["ArrowLeft", "ArrowRight", "ArrowUp", "ArrowDown"];
+
         function onWindowKeyDown(e) {
             if (!currentVisibleMenu && 32 === e.keyCode) {
                 playbackManager.playPause(currentPlayer);
+                return void showOsd();
+            }
+
+            if (layoutManager.tv && NavigationKeys.indexOf(e.key) != -1) {
                 return void showOsd();
             }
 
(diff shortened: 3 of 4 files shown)


dmitrylyzo:
In webOS 2/3, playback control does not work, because their old browsers do not have `key` property in `keydown` event (`keyCode` only).
If it is necessary to support them, I suggest to add some "key detection" function (for example, in `inputManager`), and replace all instances of `key` (maybe `keyCode` also) with it. This is for future.

dkanada:
How old are the devices that use WebOS 2 and 3 by now? if they're too old we could just ignore them for now.

src/components/emby-slider/emby-slider.css
```diff
@@ -87,6 +87,10 @@ _:-ms-input-placeholder {
     transform: scale(1.6);
 }
 
+.mdl-slider:focus::-webkit-slider-thumb {
+    transform: scale(1.6);
+}
+
```
dkanada: Won't this have the same focus issue we are trying to solve in your scaling pull request? We only want focus to show for TV layouts normally.

dmitrylyzo: It is also scaled on hover.
https://github.com/jellyfin/jellyfin-web/blob/8aae12d0147e7a421d41b40a7cb75f9ca3d2d249/src/components/emby-slider/emby-slider.css#L86-L88

But it is not a problem to add condition in `emby-slider`.

Just got a thought. In fact, the focus should be displayed if keyboard navigation is enabled. And keyboard navigation should (?) be enabled only on TV. But there is also TAB navigation, so maybe I'm wrong.

dmitrylyzo:
> How old are the devices that use WebOS 2 and 3 by now? if they're too old we could just ignore them for now.

webOS 2.0 is 2014-1015, but not sure. @JustAMan has one (https://github.com/jellyfin/jellyfin/issues/588#issuecomment-454401716).

dmitrylyzo:
There are some issues caused by "keyboard dragging".

I left slider step as `stepDown`/`stepUp`. As a result:
1. User can move slider with a mouse only by last used step.
2. Firefox marks slider as invalid because value is not aligned.
3. Chrome rounds slider value to step - progress looks discrete.
4. User does not seek in steps relative to current position, but on some grid.

I did not pay attention. :disappointed:

JustAMan:
> > How old are the devices that use WebOS 2 and 3 by now? if they're too old we could just ignore them for now.
> 
> webOS 2.0 is 2014-1015, but not sure. @JustAMan has one ([jellyfin/jellyfin#588 (comment)](https://github.com/jellyfin/jellyfin/issues/588#issuecomment-454401716)).

Indeed, mine was bought in late Fall 2015. Also it's one of the last remnants that sport passive 3D which I still use from time to time, so I don't want it to be deprecated :)

That said, I personally have a magic remote, so if an app is operatable by it but not by other buttons that is bearable for me.

dmitrylyzo:
I re-implemented keyboard dragging. Actually, I started with this variant, but chose more native one (as it turned out, it is not suitable).

- Now keyboard dragging is only enabled in TV layout.
- Dragging is relative to the current position.
- Dragging is snapped to step of slider and does not break it.

Do you mind if I squash commits?

@JustAMan, do you have same issue in real TV with webOS 2 (with clean jellyfin-web tree).
jellyfin-web_issue06_01
1. Slider thumb is not scaled.
2. Bubble is below the slider.
3. Progress is outside.
4. Is the video playing normally? The emulator either shows a black screen, or crushes, or back to previous page. _But maybe this is emulator issue._

dkanada:
Squashing your commits is fine. What error does the emulator console show during video playback? We can narrow it down from there.

dmitrylyzo:
> What error does the emulator console show during video playback?

The problem is that I test jellyfin in built-in browser and don't figure out how to connect inspector to it. For debugging purpose, I added fixed textarea and redirect console to it. :sunglasses:

At current moment, I got error on loading of `bundle.js`: "Unexpected reserved word". Chrome 34 (webOS 2) failed on `class` of libjass. Moreover, it does not support `let` and arrow function syntax. But site is loaded.

Next, I played with emulator RAM amount and instead of "Web page was unexpectedly terminated" started to get "medianotsupported". After several attempts (triggered Play one time), got an error.
jellyfin-web_issue07_02-noplay-webos2
jellyfin-web_issue07_03-noplay-webos2
But some videos give no error, but stand still.

As I stated in ScrollManager PR, built-in browsers in webOS 2/3 are not detected as webOS. Supported formats may need to be tuned.

Or maybe playback issue is related to emulator and/or server configuration. So I asked what was happening on real TV.

dkanada:
@thornbill the bundle issues above might indicate an issue with ES6, does webpack have anything to transpile the dependencies right now?

dmitrylyzo:
It seems that this is `flv.js` _(not libjass, as I thought)_.
<details>
  <summary>Unminified bundle.js fragment</summary>

  ```javascript
        class o {
            static e(e, t) {
                e && !o.FORCE_GLOBAL_TAG || (e = o.GLOBAL_TAG);
                let r = `[${e}] > ${t}`;
                o.ENABLE_CALLBACK && o.emitter.emit("log", "error", r), o.ENABLE_ERROR && (console.error ? console.error(r) : console.warn ? console.warn(r) : console.log(r))
            }
            static i(e, t) {
                e && !o.FORCE_GLOBAL_TAG || (e = o.GLOBAL_TAG);
                let r = `[${e}] > ${t}`;
                o.ENABLE_CALLBACK && o.emitter.emit("log", "info", r), o.ENABLE_INFO && (console.info ? console.info(r) : console.log(r))
            }
            static w(e, t) {
                e && !o.FORCE_GLOBAL_TAG || (e = o.GLOBAL_TAG);
                let r = `[${e}] > ${t}`;
                o.ENABLE_CALLBACK && o.emitter.emit("log", "warn", r), o.ENABLE_WARN && (console.warn ? console.warn(r) : console.log(r))
            }
            static d(e, t) {
                e && !o.FORCE_GLOBAL_TAG || (e = o.GLOBAL_TAG);
                let r = `[${e}] > ${t}`;
                o.ENABLE_CALLBACK && o.emitter.emit("log", "debug", r), o.ENABLE_DEBUG && (console.debug ? console.debug(r) : console.log(r))
            }
            static v(e, t) {
                e && !o.FORCE_GLOBAL_TAG || (e = o.GLOBAL_TAG);
                let r = `[${e}] > ${t}`;
                o.ENABLE_CALLBACK && o.emitter.emit("log", "verbose", r), o.ENABLE_VERBOSE && console.log(r)
            }
        }
        o.GLOBAL_TAG = "flv.js", o.FORCE_GLOBAL_TAG = !1, o.ENABLE_ERROR = !0, o.ENABLE_INFO = !0, o.ENABLE_WARN = !0, o.ENABLE_DEBUG = !0, o.ENABLE_VERBOSE = !0, o.ENABLE_CALLBACK = !1, o.emitter = new s.a;
  ```
</details>

JustAMan:
@dmitrylyzo I cannot check for WebOS compat now, don't have time due to various reasons.
I'm still planning on working on WebOS 2.0 compatibility, so if your changes work for _some_ recent enough WebOS - don't feel blocked on older versions for now, go ahead.

src/components/emby-slider/emby-slider.js
```diff
@@ -177,6 +185,116 @@ define(['browser', 'dom', 'layoutManager', 'css!./emby-slider', 'registerElement
         }
     };
 
+    /**
+     * Keyboard dragging timeout.
+     * After this delay "change" event will be fired.
+     */
+    var KeyboardDraggingTimeout = 1000;
```
JustAMan: why this one starts with capital `K` while other starts with lower-case `k`?

src/components/emby-slider/emby-slider.js
```diff
@@ -177,6 +185,116 @@ define(['browser', 'dom', 'layoutManager', 'css!./emby-slider', 'registerElement
...
+     */
+    var keyboardDraggingTimer;
+
+    /**
+     * Start keyboard dragging.
+     *
+     * @param {Object} elem slider itself
+     */
+    function startKeyboardDragging(elem) {
+
```
JustAMan:
```suggestion
```

src/components/emby-slider/emby-slider.js
```diff
@@ -177,6 +185,116 @@ define(['browser', 'dom', 'layoutManager', 'css!./emby-slider', 'registerElement
...
+        elem.keyboardDragging = true;
+
+        clearTimeout(keyboardDraggingTimer);
+
```
JustAMan:
```suggestion
```

src/components/emby-slider/emby-slider.js
```diff
@@ -177,6 +185,116 @@ define(['browser', 'dom', 'layoutManager', 'css!./emby-slider', 'registerElement
...
+        }, KeyboardDraggingTimeout);
+    }
+
+    /**
+     * Finish keyboard dragging.
+     *
+     * @param {Object} elem slider itself
+     */
+    function finishKeyboardDragging(elem) {
+
```
JustAMan:
```suggestion
```

src/components/emby-slider/emby-slider.js
```diff
@@ -177,6 +185,116 @@ define(['browser', 'dom', 'layoutManager', 'css!./emby-slider', 'registerElement
...
+    }
+
+    /**
+     * Do step by delta.
+     *
+     * @param {Object} elem slider itself
+     * @param {number} delta step amount
+     */
+    function stepKeyboard(elem, delta) {
+
```
JustAMan:
```suggestion
```

src/components/emby-slider/emby-slider.js
```diff
@@ -177,6 +185,116 @@ define(['browser', 'dom', 'layoutManager', 'css!./emby-slider', 'registerElement
...
+     */
+    function onKeyDown(e) {
+
+        switch (e.key) {
+            case 'ArrowLeft':
+            case 'Left':
+                stepKeyboard(this, -this.keyboardStepDown || -1);
+                e.preventDefault();
+                e.stopPropagation();
+            break;
```
JustAMan:
```suggestion
                break;
```

src/components/emby-slider/emby-slider.js
```diff
@@ -177,6 +185,116 @@ define(['browser', 'dom', 'layoutManager', 'css!./emby-slider', 'registerElement
...
+
+            case 'ArrowRight':
+            case 'Right':
+                stepKeyboard(this, this.keyboardStepUp || 1);
+                e.preventDefault();
+                e.stopPropagation();
+            break;
```
JustAMan:
```suggestion
                break;
```

src/components/emby-slider/emby-slider.js
```diff
@@ -177,6 +185,116 @@ define(['browser', 'dom', 'layoutManager', 'css!./emby-slider', 'registerElement
...
+        }
+    }
+
+    /**
+     * Enable keyboard dragging.
+     */
+    EmbySliderPrototype.enableKeyboardDragging = function () {
+
```
JustAMan:
```suggestion
```

src/components/emby-slider/emby-slider.js
```diff
@@ -177,6 +185,116 @@ define(['browser', 'dom', 'layoutManager', 'css!./emby-slider', 'registerElement
...
+    /**
+     * Handle KeyDown event
+     */
+    function onKeyDown(e) {
+
+        switch (e.key) {
+            case 'ArrowLeft':
+            case 'Left':
+                stepKeyboard(this, -this.keyboardStepDown || -1);
```
JustAMan: what is the reason for having different steps for up and down stepping?

src/components/emby-slider/emby-slider.js
```diff
@@ -177,6 +185,116 @@ define(['browser', 'dom', 'layoutManager', 'css!./emby-slider', 'registerElement
...
+    }
+
+    /**
+     * Set steps for keyboard input.
+     *
+     * @param {number} stepDown step to reduce
+     * @param {number} stepUp step to increase
+     */
+    EmbySliderPrototype.setKeyboardSteps = function (stepDown, stepUp) {
+
```
JustAMan:
```suggestion
```

src/controllers/videoosd.js
```diff
@@ -770,6 +770,10 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
             var isProgressClear = state.MediaSource && null == state.MediaSource.RunTimeTicks;
             nowPlayingPositionSlider.setIsClear(isProgressClear);
 
+            if (nowPlayingItem.RunTimeTicks) {
+                nowPlayingPositionSlider.setKeyboardSteps(userSettings.skipBackLength() * 1000000 / nowPlayingItem.RunTimeTicks, userSettings.skipForwardLength() * 1000000 / nowPlayingItem.RunTimeTicks);
```
JustAMan: can this long line be wrapped?

src/videoosd.html
```diff
@@ -31,7 +31,7 @@ <h3 class="osdTitle"></h3>
                 <div class="osdTextContainer endTimeText" style="margin: 0 0 0 .25em;"></div>
             </div>
 
-            <div class="buttons">
+            <div class="buttons focuscontainer-x">
```
JustAMan: what does `focuscontainer-x` mean?

src/components/emby-slider/emby-slider.js
dmitrylyzo: This is const.

src/components/emby-slider/emby-slider.js
dmitrylyzo: User has such settings, for rewind and fast-forward.

src/components/emby-slider/emby-slider.js
dkanada: Not sure we need to differentiate since we cant migrate this codebase to use const or let variables for quite a while.

src/videoosd.html
dmitrylyzo: User navigates horizontally in such container. On Up/Down Arrow button, focus moves out from container.

src/components/emby-slider/emby-slider.js
dmitrylyzo: Looking at the rest of the code, most of the blocks (functions) are separated from their beginning. Trying to understand the code style. So, there is no need to separate the beginning of the block from its body, isn't it?

JustAMan: that's how I like to write stuff... but I mostly review JF web code for now, let's ask @jellyfin/web

src/videoosd.html
JustAMan: is this class already defined somewhere?

dmitrylyzo: <details>
  <summary>focusManager</summary>

  https://github.com/jellyfin/jellyfin-web/blob/17ea143b6b32b2d8de6d4ed6e77988017f3df71e/src/components/focusManager.js#L165-L208
</details>

src/components/emby-slider/emby-slider.js
dkanada: That's an inherited style :p most of us hate it but we just barely deminified most of the code, stylistic changes will be next. Generally, newlines should probably follow the rules below.

* no empty lines at the beginning of methods
* always have an empty line after code blocks (methods, if-else, switch, etc)
* never more than one empty line
* single newlines to break up logical sections of code as desired

EraYaN:
I know I'm a bit late but here I go anyway.

Also if you want to test this in the "app-engine" of webOS (for early versions this is actually different).

You can use this wrapper app: https://github.com/jellyfin/jellyfin-webos Just install the LG SDK to give it a go (this also allows for the debugger (Chrome) to be attached to a real TV or the emulator)

Also for the keyCodes (for webOS4 but they seem to be universal I made list: https://github.com/jellyfin/jellyfin-web/issues/303

dmitrylyzo:
@EraYaN, yes, I know about app. Actually, this app is the only option for webOS 4 emulator - no built-in browser. In webOS 2 I never got it worked (but built-in browser works). My bad, I say it so late. At current moment, webos app fails on `bundle.js` loading, as well as built-in browser - on `class`.

@JustAMan, the first attempt was unacceptable, so I added WIP. In general, now it is ready, but there is some inconvenience with LiveTV (see below).

Tested this PR on some LiveTV. Since the content is initially short in duration, 10/30 seconds seems too big for seeking.
Should the step be limited by content? For example, a percentage of the duration.

dkanada:
Does the seek completely fail or just move to the beginning or end of the progress? I'd say the latter is perfectly fine.

dmitrylyzo:
> Does the seek completely fail or just move to the beginning or end of the progress? I'd say the latter is perfectly fine.

On rewind: moves to the beginning.
On forward: moves to the end, throws an error in the console `HLS Error: Type: mediaError Details: bufferStalledError Fatal: false`, and after a delay continues that on the air.

dkanada:
I think the best fix would just be a check to make sure the progress doesn't seek past 100% during keyboard dragging, and then we can merge this.

dmitrylyzo:
> I think the best fix would just be a check to make sure the progress doesn't seek past 100% during keyboard dragging, and then we can merge this.

I limit value by min/max property in `stepKeyboard`, but I think the slider also limits value by itself.

Could this error be due to the fact that the buffer is not ready or something like that?

dmitrylyzo:
> It turned out that the steps are not updated with content, but they are in percentage, so rewinding/forwarding happens so abruptly.

I could move `setKeyboardSteps` call to `updateTimeDisplay`. This helps, but for local content it seems redundantly.
<details>
 
  ```javascript
                nowPlayingPositionText.innerHTML = "";
                nowPlayingDurationText.innerHTML = "";
            } else {
                if (nowPlayingPositionSlider && !nowPlayingPositionSlider.dragging) {
                    if (runtimeTicks) {
                        var pct = positionTicks / runtimeTicks;
                        pct *= 100;
                        nowPlayingPositionSlider.value = pct;
                    } else {
                        nowPlayingPositionSlider.value = 0;
                    }

                    if (runtimeTicks && null != positionTicks && currentRuntimeTicks && !enableProgressByTimeOfDay && currentItem.RunTimeTicks && "Recording" !== currentItem.Type) {
                        endsAtText.innerHTML = "&nbsp;&nbsp;-&nbsp;&nbsp;" + mediaInfo.getEndsAtFromPosition(runtimeTicks, positionTicks, true);
                    } else {
                        endsAtText.innerHTML = "";
                    }
                }

                if (nowPlayingPositionSlider) {
                    nowPlayingPositionSlider.setBufferedRanges(bufferedRanges, runtimeTicks, positionTicks);

>>>
                    if (runtimeTicks) {
                        nowPlayingPositionSlider.setKeyboardSteps(userSettings.skipBackLength() * 1000000 / runtimeTicks,
                            userSettings.skipForwardLength() * 1000000 / runtimeTicks);
                    }
<<<
                }

                updateTimeText(nowPlayingPositionText, positionTicks);
                updateTimeText(nowPlayingDurationText, runtimeTicks, true);
            }
  ```
</details>

Also, not sure for `enableProgressByTimeOfDay` - it seems that it shouldn't be seekable. Do not know how to check it.

> Also, if I rewind out of buffered area (if buffer is not at 0:00), playback returns to position where it was initially. The same for mouse.

Some channels (m3u tuner) have buffered range at 0.183% (still not perfect zero), and they are seekable to the beginning, others - at 40+%, and they are fail to seek to the beginning.

dkanada:
I think only enabling keyboard dragging for TV layouts is fine for now. Then we can merge this PR and you can make improvements in a separate one if you'd like.

dmitrylyzo:
If I understand correctly, we will consider this the first step - navigation for local content.
LiveTV navigation support is the next step.

src/components/emby-slider/emby-slider.js
```diff
@@ -177,6 +185,110 @@ define(['browser', 'dom', 'layoutManager', 'css!./emby-slider', 'registerElement
...
+    }
+
+    /**
+     * Do step by delta.
+     *
+     * @param {Object} elem slider itself
+     * @param {number} delta step amount
+     */
+    function stepKeyboard(elem, delta) {
+        startKeyboardDragging(elem);
```
dkanada: I would add logging here for debugging in the future.

src/components/emby-slider/emby-slider.js
```diff
@@ -177,6 +185,110 @@ define(['browser', 'dom', 'layoutManager', 'css!./emby-slider', 'registerElement
...
+                e.stopPropagation();
+                break;
+
+            case 'ArrowRight':
+            case 'Right':
+                stepKeyboard(this, this.keyboardStepUp || 1);
+                e.preventDefault();
+                e.stopPropagation();
+                break;
+        }
```
dkanada: There are two empty lines you could remove here.

src/controllers/videoosd.js
```diff
@@ -1070,12 +1075,21 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
...
+
         function onWindowKeyDown(e) {
             if (!currentVisibleMenu && 32 === e.keyCode) {
                 playbackManager.playPause(currentPlayer);
                 return void showOsd();
             }
 
+            if (layoutManager.tv && NavigationKeys.indexOf(e.key) != -1) {
+                return void showOsd();
+            }
```
dkanada: We should just call `showOsd()` in these two blocks and then return normally, unless this provides some benefit. I think it's just left over from the minified code.

JustAMan:
@dkanada my approval was stamped above and still stands :)

JustAMan:
@dmitrylyzo regardless of our approval please address what @dkanada said: https://github.com/jellyfin/jellyfin-web/pull/566#pullrequestreview-318234950, as his suggestions seem solid

src/components/emby-slider/emby-slider.js
dmitrylyzo: Like this?
```
function stepKeyboard(elem, delta) {
    console.log("EmbySlider.stepKeyboard delta: " + delta);

    startKeyboardDragging(elem);
```
But it will spam on each left/right `keydown` event when slider is focused. Therefore, I doubt a little.

dkanada: It would only log the steps at least, not the KeyDown events, but yeah you're right. We can probably do without this change.

PR #571, opened 2019-11-10
Fix select server redirect and remove more dependencies

dkanada:
We only have the API client and two modified dependencies left to go now. There are technically three left if you count the scroller library which is currently in the components directory.

This pull request is blocking the latest Android release because of a bug I mentioned [here](https://github.com/jellyfin/jellyfin-apiclient-javascript/pull/18) that affects the standalone client. @grafixeyehero you can just apply the changes from my first commit here to fix the issue.


grafixeyehero:
I tested on android for server redirect it work

PR #576, opened 2019-11-13
Remove the leading /web/ from the manifest to avoid PWA titles

### LINKED ISSUE #575: Edit: Having a baseurl makes the title bar show for the web page added to the home screen on iOS
Without a base url or back on 1.4.0
17B59BBA-4644-4627-B66D-074CAB9EA1DC

Since updating to 1.4.1 and having the baseurl set to /jellyfin it looks like this.
332437E7-73C2-431D-82B0-A35C040A6FFC

original report assumed that /jellyfin was permanent and unchangeable (my bad). Issue edited to reflect that this is related to the baseurl feature.

anthonylavado:
Remove the leading /web/ from the manifest URL to avoid PWA titles.

After this is added, whenever someone changes their base URL, they will have to make a new shortcut on their homescreen. Tested on iOS, have not tested on Android. Desktop Web remains unaffected.

**Changes**
Change the start url in manifest.json from:
```json
"start_url": "/web/index.html#!/home.html"
```
to
```json
"start_url": "index.html#!/home.html"
```
**Issues**
Fixes #575.

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -4,7 +4,7 @@
     "description": "Jellyfin: the Free Software Media System.",
     "lang": "en-US",
     "short_name": "Jellyfin",
-    "start_url": "/web/index.html#!/home.html",
+    "start_url": "index.html#!/home.html",
     "theme_color": "#101010",
     "background_color": "#101010",
     "display": "standalone",


anthonylavado:
For posterity, I have tested this with both `/jellyfin` and without any base URL set. If you do change it, you must make a new home screen shortcut on iOS though.

joshuaboniface:
Backported into 10.4.2.

PR #577, opened 2019-11-13
Fix iOS platform detection #572

### LINKED ISSUE #572: iPad OS with Chrome 79.0 cannot load web page
**Describe the bug**
iPad OS with Chrome 79 cannot view jellyfin, it just display a black and blank page, but the Safari can display and use Jellyfin web page normally.

**To Reproduce**

1. Use my iPad Pro 2018 with iPad OS 13, and open Chrome 79 which is from the Testflight
2. Input URL of my own Jellyfin
3. The error dispalys. Just a black and blank page. Which actually already happened in 10.4.0, now is 10.4.1

**Expected behavior**
It should be displaying log in page properly as Safari does

**Logs**
iPad Chrome cannot export logs

**Screenshots**
No need to upload screenshots because it is just black and blank page

 - OS: iPad OS 13
 - Browser: Chrome (Also tried with Alook browser which is not working as the Chrome)
 - Jellyfin Version: 10.4.1
 - Reverse proxy: nginx

dmitrylyzo:
Fix #572 

Apple iPad Pro 11
iOS 13.1.?
Chrome 78.0.3904.84
```
userAgent: Mozilla/5.0 (iPad; CPU OS 13_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) CriOS/78.0.3904.84 Mobile/15E148 Safari/604.1
platform: MacIntel
appVersion: 5.0 (iPad; CPU OS 13_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) CriOS/78.0.3904.84 Mobile/15E148 Safari/604.1
```

It seems that there are more Apple platforms [(What is the list of possible values for navigator.platform as of today?).](https://stackoverflow.com/questions/19877924/what-is-the-list-of-possible-values-for-navigator-platform-as-of-today), and this is a pretty old question.

### DIFF
--- a/src/scripts/browser.js
+++ b/src/scripts/browser.js
@@ -122,7 +122,7 @@ define([], function () {
     }
 
     function iOSversion() {
-        if (/iP(hone|od|ad)/.test(navigator.platform)) {
+        if (/iP(hone|od|ad)|MacIntel/.test(navigator.platform)) {
             // supports iOS 2.0 and later: <http://bit.ly/TJjs1V>
             var v = (navigator.appVersion).match(/OS (\d+)_(\d+)_?(\d+)?/);
             return [parseInt(v[1], 10), parseInt(v[2], 10), parseInt(v[3] || 0, 10)];
--- a/src/scripts/browser.js
+++ b/src/scripts/browser.js
@@ -122,6 +122,7 @@ define([], function () {
     }
 
     function iOSversion() {
+        // MacIntel: Apple iPad Pro 11 iOS 13.1
         if (/iP(hone|od|ad)|MacIntel/.test(navigator.platform)) {
             // supports iOS 2.0 and later: <http://bit.ly/TJjs1V>
             var v = (navigator.appVersion).match(/OS (\d+)_(\d+)_?(\d+)?/);
@@ -301,7 +302,10 @@ define([], function () {
 
     if (browser.iOS) {
         browser.iOSVersion = iOSversion();
-        browser.iOSVersion = browser.iOSVersion[0] + (browser.iOSVersion[1] / 10);
+
+        if (browser.iOSVersion && browser.iOSVersion.length >= 2) {
+            browser.iOSVersion = browser.iOSVersion[0] + (browser.iOSVersion[1] / 10);
+        }
     }
 
     browser.chromecast = browser.chrome && userAgent.toLowerCase().indexOf('crkey') !== -1;


dmitrylyzo:
Safari works because it is just not detected as iOS.

Same Apple iPad Pro 11
iOS 13.2.2 (just updated)
Safari
```
userAgent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/13.0.3 Safari/605.1.15
platform: MacIntel
appVersion: 5.0 (Macintosh; Intel Mac OS X 10_15) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/13.0.3 Safari/605.1.15
```

src/scripts/browser.js
```diff
@@ -122,7 +122,7 @@ define([], function () {
     }
 
     function iOSversion() {
-        if (/iP(hone|od|ad)/.test(navigator.platform)) {
+        if (/iP(hone|od|ad)|MacIntel/.test(navigator.platform)) {
```
dkanada: Maybe add a comment and also fix the crash itself from happening again with a quick check?

dmitrylyzo: The check will protect against crushing, but it can shadow the issue: everything works fine, but in fact iOSVersion is undefined, and no one knows about it.
_Once I argued about the same with the boss. :smile: His position was that it is better not to crush during the presentation._

JustAMan:
This doesn't look completely right - I think it would detect desktop Safari running on MacOS as iOS... see this: https://stackoverflow.com/questions/57765958/how-to-detect-ipad-and-ipad-os-version-in-ios-13-and-up

dmitrylyzo:
> This doesn't look completely right

I agree, this is not perfect.

> I think it would detect desktop Safari running on MacOS as iOS...

It does not call `iOSversion` because `browser.iOS` is `false` on Safari (no "ipad" in `userAgent`). So, it detects non-iOS browser.

For "perfect" match, condition from stackoverflow should be placed somewhere in `uaMatch` function.
And version of OS will be wrong anyway.

dmitrylyzo:
> Also, I will check which layout it loads in Safari on iPad.

On iPad Pro 11, Safari is detected as desktop.

anthonylavado:
This kind of makes sense now actually, they did talk about iPad OS brining desktop-class Safari to the iPad. This is probably part of the reason - if it reported itself as anything other than macOS, it would probably get shifted to a mobile version of a lot of pages online.

PR #578, opened 2019-11-14
Add autocomplete attributes to login fields

### LINKED ISSUE #574: Login page inputs missing autofill attributes
**Describe the bug**
On a Jellyfin server login page, the password field incorrectly prompts Safari on iOS and MacOS to suggest new generated passwords rather than suggesting saved passwords. The user field should contain the HTML attribute `autocomplete="username"` and the password field should contain `autocomplete="current-password"`.

**To Reproduce**
1. Go to login page for a Jellyfin server
2. Click on password field
3. Safari recommends a new generated password

**Expected behavior**
When selecting user field on login page, browser suggests stored user/password combos if applicable.

**Screenshots**

jellyfin_password_prompt_ios

jellyfin_password_prompt_macos

 - OS: iOS 12, MacOS Mojave
 - Browser: Safari
 - Jellyfin Version: 10.0.2.0

ryan-hartzell:
Fixes #574

### DIFF
--- a/src/login.html
+++ b/src/login.html
@@ -6,17 +6,17 @@
 
             <h1 style="margin-top:1em;text-align: left;">${HeaderPleaseSignIn}</h1>
 
-            <div style="height:0; overflow: hidden;"><input type="text" name="fakeusernameremembered"
-                    tabindex="-1" /><input type="password" name="fakepasswordremembered" tabindex="-1" /></div>
-
-            <div class="inputContainer">
-                <input is="emby-input" type="text" id="txtManualName" required="required" label="${LabelUser}"
-                    autocomplete="username" />
+            <div style="height:0; overflow: hidden;">
+                <input type="text" name="fakeusernameremembered" tabindex="-1" />
+                <input type="password" name="fakepasswordremembered" tabindex="-1" />
             </div>
 
             <div class="inputContainer">
-                <input is="emby-input" id="txtManualPassword" type="password" label="${LabelPassword}"
-                    autocomplete="current-password" />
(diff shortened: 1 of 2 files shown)


dkanada:
@thornbill do you mean the newlines?

thornbill:
@dkanada yes that’s what I was meaning.

dkanada:
It's used in a few other places as well. I agree it's a bit odd but until we add proper linting here I'm not going to get too worried about style issues.

Speaking of which, I think we can add linting soon. There are only three dependencies left in the source tree, and at least two of them are deminified. When I have some time I'll run eslint and see what we can start validating.

src/login.html
```diff
@@ -6,14 +6,17 @@
 
             <h1 style="margin-top:1em;text-align: left;">${HeaderPleaseSignIn}</h1>
 
-            <div style="height:0; overflow: hidden;"><input type="text" name="fakeusernameremembered" tabindex="-1" /><input type="password" name="fakepasswordremembered" tabindex="-1" /></div>
+            <div style="height:0; overflow: hidden;"><input type="text" name="fakeusernameremembered"
+                    tabindex="-1" /><input type="password" name="fakepasswordremembered" tabindex="-1" /></div>
```
dkanada: Actually, this change has particularly odd indentation. I would nest both input elements inside the div element with newlines, and place the closing tag for the div on its own line as well.

ryan-hartzell: This one I agree should definitely be changed to match nesting. To match previous style, I'll change the other lines to remove the newlines as well. They were added by my text editor formatting as those lines were over 120 chars long.

joshuaboniface:
Backported into 10.4.2.

PR #583, opened 2019-11-20
Fixed set aspect ratio option in the player

Wunax:


joshuaboniface:
Backported into 10.4.2.

PR #584, opened 2019-11-20
Enable remote control navigation

Nickbert7:
Enables remote controll navigation and moves text sections into collapse
image

### DIFF
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -3860,7 +3860,6 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
                 "SetMaxStreamingBitrate",
                 "DisplayContent",
                 "GoToSearch",
-                "DisplayMessage",
                 "SetRepeatMode",
                 "PlayMediaSource",
                 "PlayTrailers"
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -3860,6 +3860,7 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
                 "SetMaxStreamingBitrate",
                 "DisplayContent",
                 "GoToSearch",
+                "DisplayMessage",
                 "SetRepeatMode",
                 "PlayMediaSource",
                 "PlayTrailers"
--- a/src/nowplaying.html
+++ b/src/nowplaying.html
@@ -74,38 +74,35 @@
                 </div>
             </div>
         </div>
-
         <div class="navigationSection">
+            <br />
             <div>
-                <button is="paper-icon-button-light" class="btnArrowUp btnCommand autoSize" title="${ButtonArrowUp}" data-command="MoveUp">
+                <button is="paper-icon-button-light" class="btnArrowUp btnCommand autoSize raised" title="${ButtonArrowUp}" data-command="MoveUp">
                     <i class="md-icon">keyboard_arrow_up</i>
                 </button>
             </div>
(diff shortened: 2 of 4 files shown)


Nickbert7:
So, I have it basically working
It would be possible to add some more buttons, but for the basic navigation it is enough.

image

src/components/playback/playbackmanager.js
```diff
@@ -3860,7 +3860,6 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
                 "SetMaxStreamingBitrate",
                 "DisplayContent",
                 "GoToSearch",
-                "DisplayMessage",
```
JustAMan: why?

dkanada: We really need to document the client remote control feature set, right now everything is just getting passed around blindly.

Nickbert7: These are the features of the player passed to the server during local playback. (I think so)
And I think I would never send me a message to the same client...
For the KODI addon it is listed here https://github.com/jellyfin/jellyfin-kodi/blob/f44efb8e4cb99de6371eb6be21d9a6ddd2db180d/jellyfin_kodi/monitor.py#L305

I have just seen that I might have to remove three other features:
image

Maybe I'm able to hide the navigation section and both text sections completely if I can check on the now playing screen if it is local playback.
Then we could keep the features for the local player as it is.

Nickbert7: Now it should be better

dkanada:
Do you have a screenshot of what the desktop layout will look like? Also, I'd rather have an option in the user settings to turn this off for those that don't want too many buttons on the remote control page.

Nickbert7:
Cutrently on desktop it looks the same. It could also be moved into a collapse.
Nevertheless the layout also needs further improvements.
But at first I'd like to have the functionality merged.

src/nowplaying.html
```diff
@@ -74,60 +74,54 @@ <h2 class="nowPlayingPageTitle"></h2>
...
-        <div class="hide">
+        <div class="navigationSection">
+            <br />
             <div>
-                <button is="paper-icon-button-light" class="btnArrowUp btnCommand autoSize" title="${ButtonArrowUp}" data-command="MoveUp">
+                <button is="paper-icon-button-light" class="btnArrowUp btnCommand autoSize raised" title="${ButtonArrowUp}" data-command="MoveUp">
                     <i class="md-icon">keyboard_arrow_up</i>
                 </button>
             </div>
+                <br />
```
dkanada:
```suggestion
            <br />
```

src/nowplaying.html
```diff
@@ -74,60 +74,54 @@ <h2 class="nowPlayingPageTitle"></h2>
...
                 </button>
-
-                <button is="paper-icon-button-light" class="btnOk btnCommand autoSize" title="${ButtonOk}" data-command="Select">
-                    <i class="md-icon">check</i>
+                <button is="paper-icon-button-light" class="btnOk btnCommand autoSize raised" title="${ButtonOk}" data-command="Select">
+                    <i class="md-icon">keyboard_return</i>
                 </button>
-
-                <button is="paper-icon-button-light" class="btnArrowRight btnCommand autoSize" title="${ButtonArrowRight}" data-command="MoveRight">
+                <button is="paper-icon-button-light" class="btnArrowRight btnCommand autoSize raised" title="${ButtonArrowRight}" data-command="MoveRight">
```
dkanada: Don't think I'll be a fan of the raised buttons but I'll give them a shot.

PR #587, opened 2019-11-21
Add some linting rules for the CI

dkanada:


src/components/metadataeditor/metadataeditor.js
```diff
@@ -439,7 +443,6 @@ define(['itemHelper', 'dom', 'layoutManager', 'dialogHelper', 'datetime', 'loadi
 
         var html = metadataInfo.ContentTypeOptions.map(function (i) {
```
JustAMan: maybe remove those, too?

src/components/navdrawer/navdrawer.js
```diff
@@ -21,13 +21,13 @@ define(["browser", "dom", "css!./navdrawer", "scrollStyles"], function(browser,
...
-                deltaX = endX - (menuTouchStartX || 0),
-                deltaY = endY - (menuTouchStartY || 0);
+            var isOpen = self.visible;
+            var touches = getTouches(e);
+            var touch = touches[0] || {};
+            var endX = touch.clientX || 0;
+            var endY = touch.clientY || 0;
+            var deltaX = endX - (menuTouchStartX || 0);
+            var deltaY = endY - (menuTouchStartY || 0);
             setVelocity(deltaX), isOpen && 1 !== dragMode && deltaX > 0 && (dragMode = 2), 0 === dragMode && (!isOpen || Math.abs(deltaX) >= 10) && Math.abs(deltaY) < 5 ? (dragMode = 1, scrollContainer.addEventListener("scroll", disableEvent), self.showMask()) : 0 === dragMode && Math.abs(deltaY) >= 5 && (dragMode = 2), 1 === dragMode && (newPos = currentPos + deltaX, self.changeMenuPos())
```
JustAMan: why is not this de-uglified?..

src/components/navdrawer/navdrawer.js
```diff
@@ -100,41 +100,47 @@ define(["browser", "dom", "css!./navdrawer", "scrollStyles"], function(browser,
...
                 passive: !0
             }))
         };
-        var menuTouchStartX, menuTouchStartY, menuTouchStartTime, edgeContainer = document.querySelector(".mainDrawerHandle"),
-            isPeeking = !1;
+        var menuTouchStartX;
+        var menuTouchStartY;
+        var menuTouchStartTime;
+        var edgeContainer = document.querySelector(".mainDrawerHandle");
+        var isPeeking = !1;
```
JustAMan:
```suggestion
        var isPeeking = false;
```

src/components/metadataeditor/metadataeditor.js
dkanada: the newlines?

src/components/navdrawer/navdrawer.js
dkanada: There are still a few stray files that haven't been completely deuglified, I'm guessing less than ten left in the codebase.

dkanada:
@grafixeyehero the semicolon rule was picking up some minified files so I will leave it off for now. When we fix those issues I'll turn it on with the rest of the ignored rules.

src/components/metadataeditor/metadataeditor.js
JustAMan: yeah, they seem excessive

dkanada: That will be a future change, it requires some fancy changes in the eslint rules to work properly and I don't feel like figuring them out right now.

PR #592, opened 2019-11-23
Add Screen saver

grafixeyehero:


src/components/logoscreensaver/plugin.js
```diff
@@ -0,0 +1,191 @@
...
+            var keyframes = [
+                { transform: "translate3d(3000px, 0, 0)", opacity: "0", offset: 0 },
+                { transform: "translate3d(-25px, 0, 0)", opacity: "1", offset: 0.6 },
+                { transform: "translate3d(100px, 0, 0)", offset: 0.75 },
+                { transform: "translate3d(-5px, 0, 0)", offset: 0.9 },
+                { transform: "none", opacity: "1", offset: 1 }];
+            var timing = { duration: 900, iterations: iterations, easing: "cubic-bezier(0.215, 0.610, 0.355, 1.000)" };
+            return elem.animate(keyframes, timing);
+        }
+        function shake(elem, iterations) {
```
JustAMan:
```suggestion

        function shake(elem, iterations) {
```

src/components/screensavermanager.js
```diff
@@ -0,0 +1,134 @@
+define(["events", "playbackManager", "pluginManager"], function (events, playbackManager, pluginManager) {
+
+    function getMinIdleTime() {
+        // Returns the minimum amount of idle time required before the screen saver can be displayed
```
JustAMan: pls add a notion of time units here, is it in milliseconds or what?

src/components/screensavermanager.js
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    var lastFunctionalEvent = 0;
+    function getFunctionalEventIdleTime() {
+        return new Date().getTime() - lastFunctionalEvent;
+    }
+
+    events.on(playbackManager, "playbackstop", function (e, stopInfo) {
+        var state = stopInfo.state;
+        if (state.NowPlayingItem && state.NowPlayingItem.MediaType == "Video") {
```
JustAMan: what about series and other potential media types having inner video?..

src/components/screensavermanager.js
```diff
@@ -0,0 +1,134 @@
...
+
+            if (currentServer) {
+
+                var loggedInScreenSavers = screensavers.filter(function (screensaver) {
+                    return !screensaver.supportsAnonymous;
+                });
+
+                if (loggedInScreenSavers.length) {
+                    screensavers = loggedInScreenSavers;
+                }
```
JustAMan: why you disable anonymous screensavers when someone is not logged in?.. I get the idea of disabling non-anonymous when there's no one logged in, but why the other way around?

src/scripts/site.js
```diff
@@ -710,6 +715,9 @@ var AppInfo = {};
 
         require(["css!css/site"]);
 
+        // alias
+        define("inputmanager", ["inputManager"], returnFirstDependency);
```
JustAMan: why?

grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/592/commits/4038c9bcaef22a5adabc1d375c8dc79f4d54da1b

src/components/logoscreensaver/plugin.js
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/592/commits/4038c9bcaef22a5adabc1d375c8dc79f4d54da1b

src/components/screensavermanager.js
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/592/commits/4038c9bcaef22a5adabc1d375c8dc79f4d54da1b

src/components/screensavermanager.js
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/592/commits/4038c9bcaef22a5adabc1d375c8dc79f4d54da1b

src/components/screensavermanager.js
```diff
@@ -18,6 +20,42 @@ define(["events", "playbackManager", "pluginManager"], function (events, playbac
...
+            }
+        }();
+
+        var defaultOption = function (isLoggedIn) {
+            if (isLoggedIn) {
+                return "backdropscreensaver";
+            }
+
+            return "logoscreensaver";
+        }(isLoggedIn);
```
JustAMan: Can you explain why you wanted to do it so complex instead of
```js
var option;
var defaultOption = isLoggedIn ? "backdropscreensaver" : "logoscreensaver";

try {
    option = userSettings.get("screensaver", false);
} catch (err) {
    option = null;
}
```
?

src/components/screensavermanager.js
```diff
@@ -18,6 +20,42 @@ define(["events", "playbackManager", "pluginManager"], function (events, playbac
...
+
+        if ((option = option || defaultOption) === "none") {
```
JustAMan: what this `option = option` does? look like a hack, maybe just
```js
if (option === "none" || defaultOption === "none") {
```

src/components/screensavermanager.js
```diff
@@ -18,6 +20,42 @@ define(["events", "playbackManager", "pluginManager"], function (events, playbac
...
+            return null;
+        }
+
+        var plugin = pluginManager.ofType("screensaver").filter(function (i) {
+            return i.id === option;
+        })[0];
```
JustAMan: where is error handling of "user wants a screensaver we don't have"?

src/components/screensavermanager.js
```diff
@@ -18,6 +20,42 @@ define(["events", "playbackManager", "pluginManager"], function (events, playbac
...
+
+        return plugin = plugin || pluginManager.ofType("screensaver").filter(function (i) {
+            return i.id === defaultOption;
+        })[0];
```
JustAMan: eh, this looks like uglified code we fought before :) please don't, write slightly more verbose code, no one is going to object

Nickbert7:
Just a question as the file naming is quite similar to the files of recently introduced Emby screensaver feature.
Is it done by you or ported from them?
Asking whether there might be copyright / licensing issues or not...

grafixeyehero:
> 
> 
> Just a question as the file naming is quite similar to the files of recently introduced Emby screensaver feature.
> Is it done by you or ported from them?
> Asking whether there might be copyright / licensing issues or not...

https://github.com/grafixeyehero/Emby.Web.Theater/blob/flexheader/components/screensavermanager.js

Nickbert7:
Ah, ok. Then also Emby just reimplemented their existing code.
I just wanted to avoid unnecessary complications...
Please just keep going 👍

src/components/screensavermanager.js
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/592/commits/f28074fb9ed25b6566c0dd58ee8b3ccec96e4adf

src/components/screensavermanager.js
```diff
@@ -18,6 +20,37 @@ define(["events", "playbackManager", "pluginManager"], function (events, playbac
         }
     });
 
+    var isLoggedIn;
+    var defaultOption = isLoggedIn ? "backdropscreensaver" : "logoscreensaver";
```
JustAMan: this won't work, as you determine the value of `isLoggedIn` down below

grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/592/commits/1a63de77e16d59abc9bff889f1007a0b69fa8f91

src/components/screensavermanager.js
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/592/commits/f28074fb9ed25b6566c0dd58ee8b3ccec96e4adf

src/components/screensavermanager.js
```diff
@@ -0,0 +1,133 @@
...
+            option = null;
+        }
+
+        if (option === "none" || defaultOption === "none") {
+            return null;
+        }
+
+        var plugin = pluginManager.ofType("screensaver").filter(function (i) {
+            return i.id === option;
+        })[0];
```
JustAMan: where is error handling of "user wants a screensaver we don't have"?

src/components/screensavermanager.js
```diff
@@ -0,0 +1,133 @@
...
+
+        return plugin = plugin || pluginManager.ofType("screensaver").filter(function (i) {
+            return i.id === defaultOption;
+        })[0];
```
JustAMan: eh, this looks like uglified code we fought before :) please don't, write slightly more verbose code, no one is going to object

src/components/screensavermanager.js
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/592/commits/06b42973f401b5d93b4f692c762145d40cffb6c0

src/components/screensavermanager.js
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/592/commits/06b42973f401b5d93b4f692c762145d40cffb6c0

src/components/screensavermanager.js
```diff
@@ -20,27 +20,26 @@ define(["events", "playbackManager", "pluginManager", "inputManager", "connectio
...
 
-    function getScreensaverPlugin(isLoggedIn) {
+    function getScreensaverPlugin() {
 
         var option;
-        var defaultOption = isLoggedIn ? "backdropscreensaver" : "logoscreensaver";
         try {
             option = userSettings.get("screensaver", false);
         } catch (err) {
             option = null;
```
JustAMan: maybe we should define it here using info if user is logged in?..

grafixeyehero: why?

JustAMan: in line 89 you're calling it like this:
```js
var screensaver = getScreensaverPlugin(isLoggedIn);
```
but your latest implementation of `getScreensaverPlugin()` doesn't take any parameters

grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/592/commits/0821d0ad9edffa9570723750a547f0cb764df73f

dkanada:
I'd rather not add this until we have a setting in the user preferences. Actually, you did mention it already exists as a hidden setting right?

grafixeyehero:
@dkanada I already enable that.

Issue #465, opened 2019-09-13
"Enable backdrops" doesn't effect backdrop display


ThatNerdyPikachu:
**Describe the bug**
If you attempt to disable backdrops, the checkbox in Display doesn't do anything

**To Reproduce**
1. Try to disable backdrops via Display settings

**Expected behavior**
Backdrops should be enabled or disabled based on the checkbox

grafixeyehero:
The reason why this happening that in itemdetailpage the backdrop doesn't pass through the checking point. 
there are two background file one is autobackdrop and the other is  backdrop. the checking was made in atuobackdrop and pass to backdrop. from itemdetailpage it only pass through backdrop.
here is the example if you remove this code u will see the result
https://github.com/jellyfin/jellyfin-web/blob/79cd6a75523f250f3fc428e3194105708ccf9911/src/controllers/itemdetailpage.js#L524-L525

dkanada:
@grafixeyehero I suppose the proper fix would be to remove the backdrop call in the item details page? This has actually been bothering me for ages, glad someone finally looked into the problem.

grafixeyehero:
if we remove this when user enable the backdrop it will get random backdrop. so for itemdetailpage i will set to clear backdrop.

PR #594, opened 2019-11-24
Fix some issues with the backdrop image

grafixeyehero:
**Issues**
"Enable backdrops" doesn't effect backdrop display #465

### DIFF
--- a/src/components/themes/blueradiance/theme.css
+++ b/src/components/themes/blueradiance/theme.css
@@ -49,7 +49,7 @@ html {
 }
 
 .backgroundContainer.withBackdrop {
-    opacity: .60
+    opacity: .86
 }
 
 @media (orientation:portrait) {
--- a/src/components/themes/blueradiance/theme.css
+++ b/src/components/themes/blueradiance/theme.css
@@ -49,7 +49,7 @@ html {
 }
 
 .backgroundContainer.withBackdrop {
-    opacity: .93
+    opacity: .60
 }
 
 @media (orientation:portrait) {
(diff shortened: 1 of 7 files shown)


src/components/themes/blueradiance/theme.css
```diff
@@ -49,7 +49,7 @@ html {
 }
 
 .backgroundContainer.withBackdrop {
-    opacity: .93
+    opacity: .60
```
JustAMan: why?

grafixeyehero: more visible for backdrop

grafixeyehero: Before
ba

After
tr

dkanada: Not a fan of this change, I think it should only go in the Blue Radiance theme.

src/controllers/itemdetailpage.js
```diff
@@ -521,11 +521,7 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "cardBuild
         renderDetails(page, item, apiClient, context);
         renderTrackSelections(page, instance, item);
 
-        if (dom.getWindowSize().innerWidth >= 1000) {
-            backdrop.setBackdrops([item]);
-        } else {
-            backdrop.clear();
-        }
+        backdrop.clear();
```
dkanada: Why not just check the backdrop setting here as well for the time being? We can combine the code later but at least then it would be configurable if people like the backgrounds.

src/components/themes/blueradiance/theme.css
grafixeyehero: why backdrop option needed then?

src/controllers/itemdetailpage.js
grafixeyehero: i am thinking on detailpage like plex

grafixeyehero: detail
detail-2

src/components/themes/blueradiance/theme.css
dkanada: It actually looks a lot different on my machine, I'm not sure why yours is so dark by default.

src/controllers/itemdetailpage.js
dkanada: I wonder if we need the backdrops at all if we go that route. Not sure how many people actually like them but I honestly prefer the Plex interface a lot better, which has no shaded backdrops.

src/components/themes/blueradiance/theme.css
grafixeyehero: the opacity allmost 93%

src/controllers/itemdetailpage.js
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/594/commits/7afdc4b2b1dd9f3f5f98342a434ee9a225a76b59

src/components/themes/blueradiance/theme.css
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/594/commits/81b78035c060ea72b63224a4d44bee12e092f454

joshuaboniface:
Cherry-picked into 10.4.3.

PR #596, opened 2019-11-25
Fix input field user avatar in a wrong position

Wunax:
chrome_cfUCWFmVao
chrome_E1dk8Dmqng


joshuaboniface:
Cherry-picked into 10.4.3.

PR #520, opened 2019-10-11
Cleanup connect.emby.media

grafixeyehero:
**To Do**
Cleanup connect.emby.media

know i have just do commit out the code before completely remove. 
Any opinion or input is Appreciated

PR #600, opened 2019-11-26
Fix can't connect to server if enableAutoLogin is false

### LINKED ISSUE #599: Disabling Remember Me causes infinite Spinning Circle
**Describe the bug**
*Not* enabling the Remember Me feature during one session will make the next time fail by showing a spinning circle indefinitely until you clear your browser cache. 

**To Reproduce**

1. Go to the login page, enter your credentials.
2. Disable remember me and log in.
3. Close the browser or tab.
4. Reopen it and attempt to return to Jellyfin's webpage.
5. Immediately see spinning circle which does not disapear.

To clear the error, clear your browser cache for the site (in Firefox, click lock next to URL, click "Clear Cookies and Site Data") and re-login.

**Expected behavior**
I'm met with either a login prompt or am immediately brought to the dashboard.

**Logs**
No logs on the server-side. In the browser I don't see anything out of the ordinary, but these might be helpful; this is what I see when the spinning circle is present:
```
returning instance from getOrAddApiClient connectionmanager.js:555:21
Begin connect connectionmanager.js:741:17
Begin getAvailableServers connectionmanager.js:633:21
Begin connectToServers, with 1 servers connectionmanager.js:649:21
begin connectToServer connectionmanager.js:665:21
tryReconnect: https://media.example.com connectionmanager.js:394:21
getTryConnectPromise https://media.example.com connectionmanager.js:336:21
ConnectionManager requesting url: https://media.example.com/system/info/public connectionmanager.js:121:17
fetchWithTimeout: timeoutMs: 20000, url: https://media.example.com/system/info/public connectionmanager.js:98:17
fetchWithTimeout: succeeded connecting to url: https://media.example.com/system/info/public connectionmanager.js:105:25
ConnectionManager response status: 200, url: https://media.example.com/system/info/public connectionmanager.js:123:21
Reconnect succeeded to https://media.example.com connectionmanager.js:345:29
```

**Screenshots**
I won't waste the bytes on a spinning circle.

 - OS: Host is Docker, client is Arch Linux
 - Browser: Firefox 70.0.1
 - Jellyfin Version: 10.4.2

**Additional context**
I've tested this in Iridium and the problem did not seem to occur there.

Wunax:
**Issues**
Fix #599

### DIFF
--- a/src/libraries/apiclient/connectionmanager.js
+++ b/src/libraries/apiclient/connectionmanager.js
@@ -410,9 +410,7 @@ define(["events", "apiclient", "appStorage"], function (events, apiClientFactory
             var credentials = credentialProvider.credentials();
             options = options || {};
 
-            if (false !== options.enableAutoLogin) {
-                afterConnectValidated(server, credentials, systemInfo, connectionMode, serverUrl, true, options, resolve);
-            }
+            afterConnectValidated(server, credentials, systemInfo, connectionMode, serverUrl, true, options, resolve);
         }
 
         function afterConnectValidated(server, credentials, systemInfo, connectionMode, serverUrl, verifyLocalAuthentication, options, resolve) {


dkanada:
We are actually in the process of migrating the API client over to [this](https://github.com/jellyfin/jellyfin-apiclient-javascript) repository, since it makes more sense as a separate project. Any chance you could also add a pull request there so we don't forget it? Any ideas on the best way to migrate over cleanly would also be welcome, the initialization process is slightly different so code changes will be needed in the web client.

JustAMan:
This basically reverts https://github.com/jellyfin/jellyfin-web/pull/520, ping @grafixeyehero

dkanada:
The logic looks fine though, we should always call the method here even if the auto-login option isn't checked.

JustAMan:
It does, but @grafixeyehero added that check for some reason (it was without this check before he made his PR), that's why I'm asking what was the reason for adding it.

grafixeyehero:
this is my mistake thanks Wunax for fix

joshuaboniface:
File does not exist in 10.4.z release chain; not backporting.

PR #597, opened 2019-11-25
Minor changes to improve Firefox support

dkanada:
Firefox is really screwed up right now, and I'm not sure exactly why but it might have been a recent update. Can someone try an older version and see if the same issues exist? I added the charset but another ten issues were hiding below that.

PR #601, opened 2019-11-26
Fix webcomponents-lite bundling 

grafixeyehero:
i don't know why firefox not working. but this well fix.

### DIFF
--- a/src/bundle.js
+++ b/src/bundle.js
@@ -4,12 +4,6 @@
 // Use define from require.js not webpack's define
 var _define = window.define;
 
-// document-register-element
-var docRegister = require("document-register-element");
-_define("document-register-element", function() {
-    return docRegister;
-});
-
 // fetch
 var fetch = require("whatwg-fetch");
 _define("fetch", function() {
@@ -76,12 +70,6 @@ _define("swiper", function() {
 var sortable = require("sortablejs");
 _define("sortable", function() {
     return sortable;
-});
-
-// webcomponents
-var webcomponents = require("webcomponents.js-2");
-_define("webcomponents", function() {
-    return webcomponents
 });
 
 // libjass
--- a/src/bundle.js
+++ b/src/bundle.js
@@ -4,6 +4,12 @@
 // Use define from require.js not webpack's define
 var _define = window.define;
 
+// document-register-element
+var docRegister = require("document-register-element");
+_define("document-register-element", function() {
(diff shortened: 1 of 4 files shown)


src/bundle.js
```diff
@@ -4,12 +4,6 @@
 // Use define from require.js not webpack's define
 var _define = window.define;
 
-// document-register-element
-var docRegister = require("document-register-element");
-_define("document-register-element", function() {
-    return docRegister;
-});
-
```
JustAMan: why remove this and other piece?

dkanada:
@dmitrylyzo is document-register-element also causing issues or is it just the webcomponents library?

dmitrylyzo:
> @dmitrylyzo is document-register-element also causing issues or is it just the webcomponents library?

`document-register-element` seems good. At least, there are no errors in the console, and Firefox is loading.

src/bundle.js
grafixeyehero: i don't know why this happening on firefox. removeing form bundle fix the problem.

dmitrylyzo: :confused: For me, only `webcomponents` caused an issue in Firefox.

dkanada:
My charset pull request does fix another console error on Firefox so we should merge that one as well.

src/bundle.js
grafixeyehero: ahh i have found the way to fix

dkanada:
> Module not found: Error: Can't resolve 'webcomponents.js/webcomponents-lite'

The build is failing for me on this branch. How about using the dependency @dmitrylyzo mentioned in [this](https://github.com/jellyfin/jellyfin-web/pull/597#issuecomment-558779051) comment?

dmitrylyzo:
> The build is failing for me on this branch.

I tried `fix` branch (+merged `master` to it) and have no errors (`npm update` + `npm run build`).
Resulting `bundle.js` has `webcomponents-lite` code.
Firefox loaded.
IE11 fails on `bundle.js` (on `class` keyword I suppose) - according to `site.js`, `webcomponents` is for `msie`.

> How about using the dependency @dmitrylyzo mentioned in [this](https://github.com/jellyfin/jellyfin-web/pull/597#issuecomment-558779051) comment?

Mentioned by me is older.

grafixeyehero:
Firefox load correctly form my side.

PR #604, opened 2019-11-28
Fix collection padding in TV layout

grafixeyehero:
Before
Capture bef

After
Capture c


joshuaboniface:
Cherry-picked into 10.4.3.

Issue #471, opened 2019-09-17
[Feature Request] [Documentation] Getting started developing


majora2007:
**Description**
I think it would be extremely helpful if there was a CONTRIBUTING.md or README.md that explained clear steps on how to develop the jellyfin-web with a local server. Do we copy the main directory? The dist folder? Where do we copy? Etc.

This should help attract new developers as the setup is daunting with no documentation (except for the good docs on the main jellyfin server setup). 

I figured this would fit better here rather than the feathub.

thornbill:
I really want to make some improvements to simplify this going forward, but here is my current process to run an instance of the website locally and connect to an existing Jellyfin server.

0. Fork and clone this repository.
1. Run `yarn install`.
2. Apply this [patch](https://gist.github.com/thornbill/069f5a979ffbd77a9068bd63cfb4f0a1) to enable multiserver support and make the necessary changes to `index.html` to load the main scripts.
3. Start a web server hosting the static files in `dist/`. PHP and Python include servers that you can use for this anything that can server static files will work. I use `cd dist && sudo python3 -m http.server 80` (`sudo` is needed to bind to port 80 on my operating system).
4. Open the site in a browser. In the case above this would just be at `http://localhost`. The loading spinner will continue to display, but proceed to select "Change Server" then "Add Server" and add your Jellyfin server url.
5. Make changes to the web source.
6. Run `yarn build`.
7. Refresh to view changes in browser.

grafixeyehero:
my setup with vsc
1.Fork and clone jellyfin repository
2.open it with vsc workspace
3.add jellyfin-web submodule in to workspace
4. jellyfin.Server/Program.cs change this
`webDir = Path.Combine(AppContext.BaseDirectory, "jellyfin-web", "src");` to `webDir = Path.Combine(AppContext.BaseDirectory, "jellyfin-web", "dist");`
and this to 
MediaBrowser.WebDashboard/MediaBrowser.WebDashboard.csproj change this
`<None Include="jellyfin-web\src\**\*.*">` to `<None Include="jellyfin-web\dist\**\*.*">`
5.open jellyfin-web in vsc terminal 
6.yarn install 
7.yarn run build 
note: every time you change code must run yarn run build
8.start debugging

ringmatter:
[...]

PR #605, opened 2019-11-29
Add support for local webpack devserver development

ringmatter:
**Changes**
Added html-webpack-plugin and webpack-concat-plugin to be able to inject the appLoader script using webpack dev server. This enabled developers to use the webpack dev server for local development and eliminates the need for constant rebuilding when changing something in the source code.

Background: The index.html file itself does not contain any scripts. The appLoader script will be [injected by the server](https://github.com/jellyfin/jellyfin/blob/master/MediaBrowser.WebDashboard/Api/PackageCreator.cs#L140) during startup.

~~This PR needs an PR in jellyfin server which removes the template tag from index.html during startup.~~

**Issues**
Refs #471

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -102,5 +102,6 @@
     <div class="skinHeader focuscontainer-x"></div>
     <div class="mainAnimatedPages skinBody"></div>
     <div class="mainDrawerHandle"></div>
+    <%= htmlWebpackPlugin.options.appLoader %>
 </body>
 </html>
--- a/src/index.html
+++ b/src/index.html
@@ -102,6 +102,5 @@
     <div class="skinHeader focuscontainer-x"></div>
     <div class="mainAnimatedPages skinBody"></div>
     <div class="mainDrawerHandle"></div>
-    <%= htmlWebpackPlugin.options.appLoader %>
 </body>
 </html>
--- a/webpack.dev.js
+++ b/webpack.dev.js
@@ -1,7 +1,8 @@
 const path = require("path");
 const common = require("./webpack.common");
 const merge = require("webpack-merge");
-var HtmlWebpackPlugin = require('html-webpack-plugin');
+const HtmlWebpackPlugin = require('html-webpack-plugin');
+const ConcatPlugin = require('webpack-concat-plugin');
 
 module.exports = merge(common, {
     mode: "development",
@@ -24,9 +25,12 @@ module.exports = merge(common, {
     },
     plugins: [
         new HtmlWebpackPlugin({
-            appLoader: '<script>window.dashboardVersion="0.0.0";</script><script src="scripts/apploader.js?v=0.0.0" defer></script></body>',
             filename: 'index.html',
             template: 'index.html'
+        }),
+        new ConcatPlugin({
+            name: 'scripts/apploader.js',
+            filesToConcat: ['./scripts/apploader.js']
         })
     ],
     devServer: {
--- a/webpack.dev.js
+++ b/webpack.dev.js
@@ -30,14 +30,7 @@ module.exports = merge(common, {
         }),
         new ConcatPlugin({
             name: 'scripts/apploader.js',
-            filesToConcat: ['./scripts/apploader.js']
+            filesToConcat: ['./standalone.js', './scripts/apploader.js']
         })
-    ],
-    devServer: {
-        proxy: {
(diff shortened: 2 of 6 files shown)


src/index.html
```diff
@@ -102,5 +102,6 @@
     <div class="skinHeader focuscontainer-x"></div>
     <div class="mainAnimatedPages skinBody"></div>
     <div class="mainDrawerHandle"></div>
+    <%= htmlWebpackPlugin.options.appLoader %>
```
JustAMan: Won't this break the (future) ability to serve those files as static ones by e.g. nginx? Can we change the logic so that it's possible to do so, like this?

```suggestion
    <!-- <%= htmlWebpackPlugin.options.appLoader %> -->
```

then you won't need to change the server...

dkanada: Yeah, I am 100% in favor of serving the web client with webpack, but not in a way that requires more server interaction.

ringmatter: I understand, and i've also tried to use the regular mechanism of the html-webpack-plugin. 
However, i don't have enough insight on how the current build-process is implemented.

In a "standard" webpack setup (at least from projects i have been working on), the entry-point is compiled into a single js filed and injected into the final html file.

This does not seem to be the case here. The "dist" folder has basically a copy of all the assets in the "src" folder after a "npm build" (with the webpack entry-point bundle.js being the only minified file").

Basically, the only thing to make webpack-dev serving work is to add the apploader.js script into the index.html. I think the only built-in webpack way is to do it with the html-webpack-plugin, but not with the standard mechanism which injects the output bundle.js, because we need a different script in the index.html.

The other option is to create a webpack plugin which simply injects the script tag hard-coded.

Do you have any other suggestions? I'd be happy to work on them, but i think without a proper local developer setup, contributors will have a hard time working on this project.

dkanada: > The other option is to create a webpack plugin which simply injects the script tag hard-coded.

This seems like a decent solution until we can build the entire source tree with webpack, which is still a work in progress.

ringmatter: @dkanada found a way to inject the apploader unmanipulated using the webpack-concat-plugin plugin.

webpack.dev.js
```diff
@@ -21,4 +23,21 @@ module.exports = merge(common, {
...
+        new HtmlWebpackPlugin({
+            filename: 'index.html',
+            template: 'index.html'
+        }),
+        new ConcatPlugin({
+            name: 'scripts/apploader.js',
+            filesToConcat: ['./scripts/apploader.js']
+        })
+    ],
+    devServer: {
```
dkanada: Any chance you could update `readme.md` with some information about building and running the client with your new changes?

webpack.dev.js
```diff
@@ -21,4 +23,21 @@ module.exports = merge(common, {
...
+        proxy: {
+            "!/**/*.html" : "http://localhost:8096",
+            "!/**/*.css" : "http://localhost:8086",
+            "!/**/*.js" : "http://localhost:8086"
```
dkanada: We have images and other assets that need to get proxied as well, would it be insecure to just pass everything in the source folder?

ringmatter: I don't think that's possible with glob-patterns (that's what webpack-dev-server uses here). I think we can just add those file patterns on demand if it starts to be a problem.

webpack.dev.js
ringmatter: Done. I'm not sure regarding the note `yarn OR npm` as i think it should be either or, and since `yarn.lock`was committed to the repo i assumed yarn is the way to go.

webpack.dev.js
dkanada: Actually, this proxy shouldn't even be required. If you add the following line to the index file the web client doesn't need to proxy any requests.

```
<script>window.appMode='standalone';</script>
```

ringmatter: Yes, that brings up a server selection dialog. If i put it in there, this will always be shown, also in the production bundle. Is that okay for you?

ringmatter: Nevermind, i added `standalone.js`which will be in injected in dev mode.

hammerings:
This is great, made it much easier to get started. But why was npm instructions removed from the readme? I use npm, works fine.

dkanada:
I use npm too, we can always add a note that it works fine.

PR #610, opened 2019-12-01
Fix existing library folder path returning undefined

YouKnowBlom:
getDefaultPath in directorybrowser.js now returns a promise if options.path exists

**Issues**
#588

### DIFF
--- a/src/components/directorybrowser/directorybrowser.js
+++ b/src/components/directorybrowser/directorybrowser.js
@@ -223,9 +223,9 @@ define(['loading', 'dialogHelper', 'dom', 'listViewStyle', 'emby-input', 'paper-
 
     function getDefaultPath(options) {
         if (options.path) {
-            Promise.resolve(options.path);
+            return Promise.resolve(options.path);
         } else {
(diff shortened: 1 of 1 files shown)


src/components/directorybrowser/directorybrowser.js
```diff
@@ -223,7 +223,7 @@ define(['loading', 'dialogHelper', 'dom', 'listViewStyle', 'emby-input', 'paper-
 
     function getDefaultPath(options) {
         if (options.path) {
-            Promise.resolve(options.path);
+            return Promise.resolve(options.path);
         } else {
             ApiClient.getJSON(ApiClient.getUrl("Environment/DefaultDirectoryBrowser")).then(
```
JustAMan: shouldn't this one be returned as well?

YouKnowBlom:
Good catch! I added a return to line 228 and tested without issue.

As a sidenote I should really work on my git skills 😄. I'm pretty sure I've reverted my mess up. Please do let me know if you notice anything I should've done differently

joshuaboniface:
Cherry-picked into 10.4.3.

Issue #552, opened 2019-10-26
Scroll arrows missing when searching


ManfredRichthofen:
There are no scroll arrows when searching for a movie

1. Go to the search
2. Type a term where you would see multiple movies
3. Notice there are no scroll arrows 

To be able to scroll through multiple movies in the search

Expected: https://imgur.com/Ya719eo 
When searching: https://imgur.com/GWFMS6s

 - OS: Ubuntu 18.4
 - Browser: Firefox Chrome and Edge
 - Jellyfin Version: 10.4
 - Reverse proxy: Same with nginx and without

ManfredRichthofen:
fixed with #612

PR #612, opened 2019-12-02
Fix Scroll arrows missing when searching

grafixeyehero:
**Issues**
Scroll arrows missing when searching #552


joshuaboniface:
Cherry-picked into 10.4.3.

PR #613, opened 2019-12-02
reset subtitle offset on next episode play

redSpoutnik:
Reset the subtitle offset (and the flag to display the subtitle offset slifing bar) when next episode is automatically played.

**Changes**

To avoid playing the next episode with subtitle offset from the previous episode, when the current htmlvideoplayer start to play a new stream:

* subtitleOffset (current subtitle offset) is set to **0**.
* showTrackOffset (flag to show the sliding bar) is set to **false**.

**Issues**

#560

### DIFF
--- a/src/components/htmlvideoplayer/plugin.js
+++ b/src/components/htmlvideoplayer/plugin.js
@@ -280,6 +280,8 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
 
             self._currentTime = null;
 
+            self.resetSubtitleOffset();
+
             return createMediaElement(options).then(function (elem) {
 
                 return updateVideoUrl(options, options.mediaSource).then(function () {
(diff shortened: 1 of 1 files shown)


src/components/htmlvideoplayer/plugin.js
```diff
@@ -556,6 +558,15 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
             setCurrentTrackElement(index);
         };
 
+        self.resetSubtitleOffset = function() {
+            if (currentTrackOffset) {
```
cvium: Is it necessary to check this? I think this should be fine?
```
            currentTrackOffset = 0;
            showTrackOffset = false;
```

redSpoutnik: You are right, I will add a commit.

joshuaboniface:
Cherry-picked into 10.4.3.

PR #615, opened 2019-12-05
Add option to randomly sort music albums

AndreCarvalho:
Add option to randomly sort music albums

**Changes**
Add music albums menu option to sort by random


src/strings/en-gb.json
```diff
@@ -1358,6 +1358,7 @@
     "RefreshQueued": "Refresh queued.",
     "Play": "Play",
     "PasswordResetProviderHelp": "Choose a Password Reset Provider to be used when this user requests a password reset",
+    "OptionRandom": "Random",
```
JustAMan: if the value is equal to "base language" (`en-us`) there's no need to duplicate it

AndreCarvalho:
> LGTM assuming it was tested and proven to be working

Correct

PR #618, opened 2019-12-07
Fix a bug that hides children elements in the videoosd

### LINKED ISSUE #530: Disappearing Buttons in Playback Menu
There is a bug on the Jellyfin Android app and also on the web version of Jellyfin on Chrome for Android. When pausing the video and letting the buttons menu fade away, it causes the last pressed button to also disappear. Best shown using screenshots:

1. All buttons are present. In this screenshot I have paused the episode.
All_Buttons

2. ~~I let the menu with all the buttons fade away~~ I tap the screen to hide the menu/controls bar, and after tapping on the screen again the menu reappears with one less button (the last button I pressed, in this case the pause/play)
Pause_Play_Missing

3. I was able to replicate this with all of the buttons in the bottom menu, and they've all now disappeared.
All_Missing

The screenshots are from the latest Jellyfin app on Android, however I was able to replicate the same issues via Chrome on my Android Device.

EDIT: Better (smaller) screenshots..
EDIT2:
**How To Replicate:**
1. Start playback
2. Press pause/play, and let controls auto fade
3. Tap screen to show controls
4. Hit play, and hide the controls manually by tapping the screen again
5. Tap the screen once more to show controls. Pause/play is now missing

Wunax:
This bug seems to be only present in the layout **mobile**

the event `onHideAnimationComplete` in the file videoosd.js also seems to be triggered by children elements
chrome_2019-12-07_13-11-01

**Changes**
checks that the element is the parent element

**Issues**
Fix #530

### DIFF
--- a/src/controllers/videoosd.js
+++ b/src/controllers/videoosd.js
@@ -402,6 +402,8 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
 
         function onHideAnimationComplete(e) {
             var elem = e.target;
+            if (elem != osdBottomElement)
+                return;
             elem.classList.add("hide");
             dom.removeEventListener(elem, transitionEndEventName, onHideAnimationComplete, {
                 once: true


dkanada:
You just missed a minor release! We can always get this in Android right away though, which is good.

thornbill:
Looks like this should fix https://github.com/jellyfin/jellyfin-web/issues/530

JustAMan:
@dkanada @anthonylavado @joshuaboniface 

> You just missed a minor release! We can always get this in Android right away though, which is good.

I think this PR alone warrants Android and iOS sub-releases, as they both should've been affected by that.

thornbill:
iOS does not bundle the web source, so it will need a server update for the fix to be available there.

dkanada:
I already released a new Android build, not sure if @anthonylavado has pushed it anywhere yet because Amazon requires a test server.

anthonylavado:
@dkanada I have not, but it can always be pushed to Google. Do we want to make a sub-release and then go ahead?

PR #619, opened 2019-12-07
Fix for card image scaling problem

DaveChild:
Card images were losing aspect, change to "background-size: cover" to fix.

**Changes**

**Issues**

### DIFF
--- a/src/components/cardbuilder/card.css
+++ b/src/components/cardbuilder/card.css
@@ -247,7 +247,7 @@ button {
 }
 
 .coveredImage {
-    background-size: 100% 100%;
+    background-size: cover;
     background-position: center center;
 }
 


dkanada:
Since this is a style change, could you provide before and after screenshots?

MrTimscampi:
Before:

Screenshot_2020-01-04 Jellyfin

After:

Screenshot_2020-01-04 Jellyfin(1)

It helps a lot to make previews and non-standard posters look nice by fitting only the height to the container and letting the width get sized to whatever it needs to keep the ratio of the image (while hiding the overflow).

DaveChild:
> Since this is a style change, could you provide before and after screenshots?

Sorry, missed this reply before. Thanks @MrTimscampi for adding the screenshots.

PR #507, opened 2019-10-08
Fix subtitle sync

redSpoutnik:
Quick PR to fix two problems I've experienced since 10.4.0 upgrade:

* Moving subtitle-sync slider doesn't apply offset to vtt subtitles.
* After closing or fading, subtitle-sync doesn't recover previous subtitle offset.

**Changes**

* Update code to recover subtitle-track.
* The wrong function was tested in "getPlayerSubtitleOffset" in playbackmanager.js (test if actual player support subtitle-sync): it always returned false and subtitleOffset was interpreted as 0.

**Issues**

PR #624, opened 2019-12-09
Fix subtitle sync firefox

redSpoutnik:
After 10.4.0 upgrade the subtitle sync didn't worked. This was partially solved in #507 for ass subtitles (chrome and firefox) and vtt subtitles (chrome only).

This PR allow vtt subtitles offset to work on firefox as well.

This PR will also reset the subtitle-offset on track changes to fix the following bug:

1. play a video with multiple subtitle tracks.
2. select a track, set an offset.
3. select a second track (track play without offset but the previous offset is still displayed).
4. change the second track offset (offset is applied by substracting the displayed offset, wrong offset is applied).

**Changes**

* vtt subtitle offset mechanism in ```setSubtitleOffset``` function is simplified with:
   * unncessary ```customTrackIndex``` check is removed.
   * the vtt subtitle offset is applied to ```_mediaElement.textTracks``` for chrome and chromium-based web browsers.
  * the vtt subtitle offset is applied to ```currentTrackEvents``` for firefox-based web browsers.
* reset subtitle offset on subtitle-track changes (when switching subtitle on the same video).

**Tested on**

* Brave 1.0.1
* Chrome 78.0.3904.108
* Firefox 70.0.1

**Issues**

#669

### DIFF
--- a/src/components/htmlvideoplayer/plugin.js
+++ b/src/components/htmlvideoplayer/plugin.js
@@ -579,10 +579,10 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
             var videoElement = self._mediaElement;
             if (videoElement) {
                 return Array.from(videoElement.textTracks)
-                .find(function(trackElement) {
-                    // get showing .vtt textTack
-                    return trackElement.mode === 'showing';
-                });
+                    .find(function(trackElement) {
+                        // get showing .vtt textTack
+                        return trackElement.mode === 'showing';
+                    });
             } else {
                 // get track events
                 return currentTrackEvents;
@@ -603,7 +603,7 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
                     offsetValue = updateCurrentTrackOffset(offsetValue);
                     setVttSubtitleOffset(trackElement, offsetValue);
                 } else {
-                    console.log("No available track, cannot apply offset : " + offsetValue);
+                    console.log("No available track, cannot apply offset : ", offsetValue);
                 }
             }
         };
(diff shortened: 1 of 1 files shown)


src/components/htmlvideoplayer/plugin.js
```diff
@@ -583,28 +597,14 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
-                        }
-
-                    });
+                var trackElement = getTextTrack();
+                // if .vtt currently rendering
+                if (trackElement) {
+                    offsetValue = updateCurrentTrackOffset(offsetValue);
+                    setVttSubtitleOffset(trackElement, offsetValue);
+                } else {
+                    console.log("No available track, cannot apply offset : ", offsetValue);
```
dkanada:
```suggestion
                    console.log("No available track, cannot apply offset: ", offsetValue);
```

src/components/htmlvideoplayer/plugin.js
```diff
@@ -583,28 +597,14 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
-                            offsetValue = updateCurrentTrackOffset(offsetValue);
-                            setVttSubtitleOffset(trackElement, offsetValue);
-                        } else {
-                            console.log("No available track, cannot apply offset : " + offsetValue);
-                        }
-
-                    });
+                var trackElement = getTextTrack();
+                // if .vtt currently rendering
+                if (trackElement) {
```
JustAMan: What happens here if "events" are returned?

redSpoutnik: If ```getTextTrack();``` return **currentTrackEvents**, **currentTrackEvents**  is an array of events, which is handled by:
* ```if (Array.isArray(currentTrack))``` (line 631, function **setVttSubtitleOffset**).

JustAMan: But isn't it truethy, thus going into 'then' clause while it shouldn't?

redSpoutnik: hum... ok, I will try to rewrite this part in a more explicit way to avoid mistakes.

JustAMan:
@dkanada can you have a look again?

Issue #569, opened 2019-11-07
Focused cards in single-row itemContainers clipped by container (in TV layout on PC)


dmitrylyzo:
**Describe the bug**
Focused cards in single-row itemContainers clipped by container (in TV layout on PC).

There are two cases:
1. `emby-scroller` wraps `itemContainer`.
Here it got `overflow: hidden`
https://github.com/jellyfin/jellyfin-web/blob/c10f3feb3149d761c993a0c3b007c09c8a848f14/src/components/scroller.js#L776
2. Standalone `itemContainer`
With `scrollX` class it got `overflow-y: hidden`

In addition, the height of `itemContainer` depends on the size of the window, and cards can also be clipped at the top. In standalone `itemContainer`, cards are always clipped at the top.

**To Reproduce**
1. Turn on TV layout
2. Go to 'Movies/_SomeMovie_'
3. Scroll down to cast section with a keyboard
4. Focus on 1st card
5. Card will be zoomed and clipped at the left

Or _(easier)_
1. Turn on TV layout
2. Go to 'Home'
3. Focus on 1st card of library section
4. Card will be zoomed and clipped at the left

And _(standalone itemContainer)_
1. Turn on TV layout
2. Go to 'Movies/Suggestions'
3. Focus on 1st card of Latest Movies section
4. Card will be zoomed and clipped at the left

**Expected behavior**
Card should not be clipped.

**Screenshots**
I added blue border for `emby-scroller`, green border for `itemContainer`, and red border for `cardBox`.
Look at 1st card.
![jellyfin-web_issue05_02-embyscroller](https://user-images.githubusercontent.com/56478732/68420965-765d8880-01ae-11ea-8c4e-d7e40f9a0100.png)
![jellyfin-web_issue05_01-embyscroller](https://user-images.githubusercontent.com/56478732/68420963-75c4f200-01ae-11ea-8065-f06f559c1d1c.png)

**System (please complete the following information):**
 - OS: Kubuntu 19.04
 - Browser: Firefox 70.0.1, Chrome 78.0.3904.70
 - Jellyfin Version: 10.4.1+ (jellyfin/jellyfin-web@77a88280187a05ab0c9ace84cd20a290e119f1b9)

dmitrylyzo:
Remained "Continue Watching" and "Latest Movies" on "Suggestions" tab. They are just not wrapped with `emby-scroller`.

PR #626, opened 2019-12-11
Fix context menu margin and user preference buttons

dkanada:
The original goal was refactoring a few things but I ended up fixing some css bugs along the way. Most of the changes are explained below.


src/components/actionsheet/actionsheet.css
```diff
@@ -37,11 +37,12 @@
     box-shadow: none;
     flex-shrink: 0;
     border-radius: 0;
+    margin: 0;
```
dkanada: This fixes a slight margin on the context menu items.

src/components/cardbuilder/card.css
```diff
@@ -627,7 +627,7 @@ button {
 
 @media (min-width: 43.75em) {
     .overflowSquareCard, .overflowPortraitCard {
-        width: 23.3vw;
+        width: 23.1vw;
```
dkanada: These are all magic numbers, and they were changed to fit within the padding better.

src/mypreferencesmenu.html
```diff
@@ -3,7 +3,7 @@
         <div class="readOnlyContent" style="margin: 0 auto;">
             <div class="verticalSection verticalSection-extrabottompadding">
                 <h2 class="sectionTitle headerUsername" style="padding-left:.25em;"></h2>
-                <a is="emby-linkbutton" data-ripple="false" href="#" class="lnkMyProfile listItem-border">
+                <a is="emby-linkbutton" data-ripple="false" href="#" style="display:block;padding:0;margin:0;" class="lnkMyProfile listItem-border">
```
dkanada: This was causing the weird button issues in the user preference menu. The two elements that were always working had inline styles, which isn't ideal but now at least everything is working.

src/components/emby-scrollbuttons/emby-scrollbuttons.css
```diff
@@ -6,6 +6,7 @@
     justify-content: center;
     min-width:104px;
     min-height:24px;
+    padding-top: 1.25em;
```
dkanada: I aligned the scroll buttons with the scroller header.

src/components/emby-scroller/emby-scroller.css
```diff
@@ -25,4 +20,4 @@
         margin-left: 0;
         margin-right: 0;
     }
```
grafixeyehero:
```
.layout-tv .emby-scroller {
    padding-left: 3.3%;
    padding-right: 3.3%;
    margin-left: 0;
    margin-right: 0;
}
```
this will fix https://github.com/jellyfin/jellyfin-web/issues/569

dkanada: What if we just remove the media selector here? I forget why that was originally added, might have something to do with the mobile layout.

grafixeyehero: yes indeed mobile layout used that too

grafixeyehero: or you can change it to `.layout-mobile .emby-scroller`

src/mypreferencesmenu.html
thornbill: Is this the issue where sometimes the menu items are too large?

dkanada: This might fix that as well but it definitely fixes the vertical alignment, which might be what you're talking about.

PR #627, opened 2019-12-12
add missing string and translation

s0urcelab:
add missing string in "itemcontextmenu.js"

**Changes**

**Issues**

### DIFF
--- a/src/strings/zh-cn.json
+++ b/src/strings/zh-cn.json
@@ -888,6 +888,7 @@
     "Monday": "星期一",
     "MoreFromValue": "更多来自 {0}",
     "MoreUsersCanBeAddedLater": "稍后可以在控制台中添加更多用户。",
+    "MoreMediaInfo": "媒体信息",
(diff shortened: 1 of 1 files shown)


dkanada (approved):

dkanada:
By the way, we have a Weblate instance if you want a more convenient method to edit strings, it doesn't require dealing with GitHub.

s0urcelab:
> By the way, we have a Weblate instance if you want a more convenient method to edit strings, it doesn't require dealing with GitHub.

i tried, but seem like without a origin string(English version) there is no way to add translation.

anthonylavado (approved):

dkanada:
Hmm that might be an issue then, every string should exist in the source language.

anthonylavado:
Looks like it does exist?
https://github.com/jellyfin/jellyfin-web/blob/e108997376ef75d999398599ce0bec844afcbbcc/src/strings/en-us.json#L1002

PR #630, opened 2019-12-14
Fixed Jellyfin Link

Artiume:
Corrected the URL so it redirects to https automatically and the correct TLD.

With that said, the link feels out of place and lonely at the bottom of the dashboard, maybe 'Jellyfin Website' or something, the single word in the wild nags at me

**Changes**

**Issues**


dkanada:
I've been meaning to add a card on that page with links to all our different stuff, so it's definitely going to happen as soon as I decide on a design.

PR #632, opened 2019-12-14
Library URL Path Corrected

Artiume:
URL was working properly using redirects. Corrected Path to new URL.

**Changes**

**Issues**


oddstr13:
I believe the plan was to move documentation over to the subdomain eventually?
It just hasn't been done yet.

JustAMan:
Ping @joshuaboniface or @anthonylavado for clarification

anthonylavado:
So the plan is indeed to have a sub domain for docs, and this redirect is standing until we get that finished.

Artiume:
So this PR should be canceled. With the redirect, the current link works and my fix is going in the wrong direction.

PR #633, opened 2019-12-15
Correct URL Link, discuss new networking page

Artiume:
Redirect from old archive URL. This URL is the closest option currently available for the networking page. A PR needs to be made to create the networking page or perhaps convert this link into the new page. discussion welcome

**Changes**

**Issues**


dkanada:
That section is supposed to be different from the hosting settings, which are the actual settings you can change related to the network in Jellyfin.

PR #634, opened 2019-12-15
Move more controllers to a sane directory structure

dkanada:


src/scripts/routes.js
```diff
@@ -16,52 +16,80 @@ define([
...
     defineRoute({
         path: "/addserver.html",
         autoFocus: false,
         anonymous: true,
         startup: true,
-        controller: "addserver"
+        controller: "auth/addserver"
+    });
+    defineRoute({
+        path: "/mypreferencesdisplay.html",
```
JustAMan: why `.html` templates were not moved?

src/scripts/routes.js
```diff
@@ -229,57 +252,21 @@ define([
...
-        autoFocus: false,
-        transition: "fade",
-        controller: "user/subtitles"
-    });
     defineRoute({
         path: "/notificationsetting.html",
         autoFocus: false,
         roles: "admin",
-        controller: "notificationsetting"
+        controller: "dashboard/notifications/notification"
```
JustAMan: wtf, why there are "notification" and "notifications"?..

src/scripts/routes.js
```diff
@@ -391,17 +373,18 @@ define([
         roles: "admin",
         controller: "userprofilespage"
     });
+
     defineRoute({
         path: "/wizardremoteaccess.html",
         autoFocus: false,
         anonymous: true,
-        controller: "wizardremoteaccess"
+        controller: "wizard/access"
```
JustAMan: I would name the controller `wizard/remoteaccess`, as it's a different thing than just `access` IMHO

src/scripts/routes.js
```diff
@@ -191,19 +219,14 @@ define([
...
     });
     defineRoute({
         path: "/login.html",
         autoFocus: false,
         anonymous: true,
         startup: true,
-        controller: "loginpage"
-    });
-    defineRoute({
-        path: "/metadataadvanced.html",
```
JustAMan: why was this removed?

src/scripts/routes.js
```diff
@@ -16,52 +16,80 @@ define([
...
 
     defineRoute({
         path: "/addplugin.html",
         autoFocus: false,
         roles: "admin",
-        controller: "addpluginpage"
+        controller: "dashboard/plugins/add"
     });
     defineRoute({
-        path: "/autoorganizelog.html",
```
JustAMan: why was this removed?

src/scripts/routes.js
```diff
@@ -16,52 +16,80 @@ define([
...
-        roles: "admin"
+        path: "/mypreferencesmenu.html",
+        autoFocus: false,
+        transition: "fade",
+        controller: "user/menu"
     });
     defineRoute({
-        path: "/channelsettings.html",
```
JustAMan: why was this removed?

src/scripts/routes.js
```diff
@@ -345,11 +332,6 @@ define([
         roles: "admin",
         controller: "streamingsettings"
     });
-    defineRoute({
-        path: "/support.html",
```
JustAMan: why this was removed?

src/videoosd.html
```diff
@@ -1,27 +1,19 @@
...
+        <div class="osdPoster"></div>
         <div class="osdControls">
-
             <div class="osdTextContainer osdMainTextContainer">
                 <h3 class="osdTitle"></h3>
                 <div class="osdMediaInfo"></div>
                 <div class="osdMediaStatus hide">
-                    <i class="md-icon animate">autorenew</i> <span>${FetchingData}</span>
+                    <i class="md-icon animate">autorenew</i>
+                    <span>${FetchingData}</span>
```
JustAMan: is indentation correct here?

src/scripts/routes.js
dkanada: None of the templates have been moved so far, only controllers.

src/scripts/routes.js
dkanada: The latter is the list showing all of them while the former is the settings page for a single notification.

src/scripts/routes.js
dkanada: doesn't exist

src/scripts/routes.js
dkanada: doesn't exist

src/scripts/routes.js
dkanada: doesn't exist afaik

src/scripts/routes.js
dkanada: definitely doesn't exist

src/videoosd.html
dkanada: It is now, it definitely wasn't before though.

JustAMan: Right, I didn't notice closing `</i>`

PR #635, opened 2019-12-15
Some general refactoring for assets

dkanada:
I pulled out the refactoring commits from another pull request into this one. All of the static assets were moved into their own folder, and themes were given a new folder since I wanted more people to see them any maybe write more. I also started the migration of custom elements to a new folder, although this isn't done yet.


src/scripts/site.js
```diff
@@ -753,11 +753,26 @@ var AppInfo = {};
         // TODO remove these libraries
         // all three have been modified so we need to fix that first
```
grafixeyehero: // all five have been modified so we need to fix that first

src/index.html
```diff
@@ -22,44 +22,44 @@
     <link rel="apple-touch-icon" sizes="180x180" href="touchicon.png">
 
     <!-- iPhone 5 -->
```
thornbill: Looks like you added extra `/` characters here.

Issue #591, opened 2019-11-23
Highlight username field during setup


anthonylavado:
**Describe the bug**
During setup, it’s not clear to users that the username is an important field that they will need to remember. Since all users are now hidden by default, they may forget what the username is when trying to log in for the first time.

This results in a lot of extra support work and people having to reset their accounts.

**Expected behavior**
The username field could have some different shading behind it to make it stand out, and better descriptive text - maybe with some of it in red (e.g. “Please select a username. **You will need this in order to login.**”)

PR #640, opened 2019-12-17
Add a warning to user wizard reflecting the importance of usernames

YouKnowBlom:
Me and @sparky8251 discussed the importance of this earlier in matrix but decided to move the discussion to a pr. Personally I think the red text might be a little over the top but I do believe the other changes are positive for the user experience

**Changes**
Replace "Your first name:" with "Username:"
Replace original username description with one more relevant to the field
Move original username description to bottom
userwizard-text-edit

**Issues**
Resolves: #591

### DIFF
--- a/src/strings/en-us.json
+++ b/src/strings/en-us.json
@@ -1282,6 +1282,7 @@
     "SearchForMissingMetadata": "Search for missing metadata",
     "SearchForSubtitles": "Search for Subtitles",
     "SearchResults": "Search Results",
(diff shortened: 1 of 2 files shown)


dkanada:
I think the two fields can be combined and we probably don't need the text to be red. The clarification should be enough for new users.

YouKnowBlom:
Sorry I haven't gotten around to changing this yet, holidays and sickness has kept me from doing much. Will update the pr once I get a chance!

YouKnowBlom:
@dkanada

PR #644, opened 2019-12-20
More accurate jump to time on slide bar

Elypha:
Hello. I notice that it is difficult to do a precise jump to time if the slider icon displays right under the cursor, which often happens when you just want to skip or go back a few seconds.
It seems that the slider icon covers the control bar so that the cursor clicks on the icon, not the bar. I try to fix it by scaling down the size of the slider to 0 while hovering and it works, though it may not look fancy.
I have checked some streaming sites like youtube and twitch and found that they did not have such problem. But I failed to figure out how they make it, as I am really new to web coding. It would be much better if someone else can fix it elegantly.

**Changes**

**Issues**

### DIFF
--- a/src/components/emby-slider/emby-slider.css
+++ b/src/components/emby-slider/emby-slider.css
@@ -84,7 +84,10 @@ _:-ms-input-placeholder {
 }
 
 .mdl-slider:hover::-webkit-slider-thumb {
-    transform: scale(1.6);
+    transform: scale(0);
+    transition: 0.4s;
+    width: 0;
+    height: 0;
 }
 
 .mdl-slider.show-focus:focus::-webkit-slider-thumb {


dkanada:
I would want to compare this with other common video platforms (Netflix or Youtube) since it would affect all slider elements in the project.

JustAMan:
Feels awkward... I don't think it is the proper way to go, can we make the slider icon pass mouse clicks through itself down to the bar?

Elypha:
Sorry about that, I should have thought that this would be a global influence on this project.
Youtube's slider is really ideal I think, but I just failed to figure out how it works...
The slider in this project seems to be controlled by an "input" element (if I'm not mistaken). It was also my first idea to pass mouse clicks. I then checked some related documents but didn't find any native parameter that can be added to an "input" tag to solve this... Do you know how this can be done?

dkanada:
Don't worry about it! That's what pull requests are for, lots of discussion on how to improve software.

dmitrylyzo:
> Do you know how this can be done?

It looks like Firefox 72 already passes clicks through.
For Chrome, adding `pointer-events: none;` to `.mdl-slider::-webkit-slider-thumb` in `src/elements/emby-slider/emby-slider.css` will do the job. But popup indicator is a bit off (+3 sec while hover, but it updates on click) - but that's another story.
Maybe other browsers also need CSS mod.

Elypha:
Thanks for that! `pointer-events: none;` works well in my chrome. 

> But popup indicator is a bit off (+3 sec while hover, but it updates on click)

I have a guess of why this happens and here's a sketch:

image

When cursor clicks, what happens is that point A (centre of the slider button) is moved right under the cursor, while it seems to be point B (left edge of the button) which really controls the playback bar.

If the size of the thumb is set to "0", which means point A is at the same position with point B, then this problem is gone.

```
    .mdl-slider::-webkit-slider-thumb {
        width: 0em;
        height: 0em;
    }
```

Maybe this is going to be an adjustment of the slider thumb's position? Hope this could be helpful for someone experienced in this…

dmitrylyzo:
@Elypha you were close.
Thumb issue due to error in calculating `bubbleValue` here:
https://github.com/jellyfin/jellyfin-web/blob/ca06c47b557caaaf99faa6b8007497d6e67459ab/src/elements/emby-slider/emby-slider.js#L157
As you stated, zeroing the size of the thumb fixes this. I tried the opposite - made a thumb 40px in size and changed the calculation of `bubbleValue`.
```js
var bubbleValue = (clientX - rect.left - 20) / (rect.width - 40);
```
So, in fact, thumb takes some of slider range.

dmitrylyzo:
slider-offset1
Hell of a perfectionist :weary:
The same at the beginning.

Elypha:
Thanks again! I want to bring the solution to code but there's a problem...

The slider is defined by `1.2em` for width and height but it is not a fixed size. I was thinking of calculating the size in a way like (I'm not really good at javascript):

```javascript
var sliderWidth = Math.round(parseFloat(getComputedStyle(document.querySelector('input.mdl-slider')).fontSize) * 1.2);
var bubbleValue = (clientX - rect.left - sliderWidth / 2) / (rect.width - sliderWidth);
```

I was trying to replace the fixed value `1.2` with a function that can get the value from `.mdl-slider::-webkit-slider-thumb` but had no idea where to get it. Also could not find it using Chrome's DevTools...

dmitrylyzo:
`-webkit-slider-thumb` is pseudo-element and, AFAIK, has no access from JS. Besides, there are more such pseudo-elements - for Firefox, Internet Explorer, and handling all of them will look clumsy.

<details>
<summary>It has just dawned upon me, but this is not a solution.</summary>

Add inside tree of slider some invisible element with same size as thumb (controlled by CSS).
Take its size like you did.
</details>

**Option 1**
_I did not check the calculation, but should work_
Taking into account [this](https://github.com/jellyfin/jellyfin-web/pull/644#issuecomment-582944768) I thought about adding `padding: 0 0.6em` (half of thumb size) to `.mdl-slider-background-flex-container` and take `rect` of `.mdl-slider-background-flex` - no need to subtract thumb size, but need to add clamping of `bubbleValue` to range [0, 1] (actually, this will need in calculations above too).
But this requires to think about layout, because all sliders will changed
from this
slider2
to this
slider1

**Option 2**
_Just idea - (over?) complicated, and not sure about compatibility with older browsers_
Make real thumb zero size.
Add new custom thumb with CSS `transform: translation(-50%, 0)` and switch thumb-related CSS on it except "zero size" ones.
Add event handlers `focus`/`blur` to add some class to slider container (`focused`).
Add event handlers `mouseover`/`mouseout` (or pointer ones) to add some class to slider container (`hovered`).
Change CSS from `.mdl-slider.show-focus:focus` to `.mdl-slider-container.show-focus.focused .mdl-slider` and from `.mdl-slider:hover` to `.mdl-slider-container.hovered .mdl-slider` or something like that.
Add custom thumb position update - similar to bubble.
Also, `.show-focus` should be copied to slider container.
Then will be something like this
slider3
Also, no need to subtract thumb size (clamp of `bubbleValue` to range [0, 1] just for sure). This also slightly affects the look of the site, but is easier to position.

What will the Jellyfin team tell us?

dkanada:
Option one looks like the best fix to me.

Elypha:
I have no idea which one of the two is technically more compatible with the current structure though both look like satisfying solutions…

@dmitrylyzo BTW, a little bit off-topic but may I consult something about Youtube's slider? It's a 13x13 size outer div contains a 0 size div inside, as following:

```html
<div class="ytp-scrubber-container" style="transform: translateX(34.5544px);">  <!-- 13px*13px -->
    <div class="ytp-scrubber-button ytp-swatch-background-color" style="height: 13px;">  <!-- 0px -->
        <div class="ytp-scrubber-pull-indicator"></div>  <!-- 0px -->
    </div>
</div>
```

I see a structure like this (use divs only, and 0 size div inside) on many popular streaming sites. For this one, is it the inner 0 size div that controls the playback bar and the outer div is used to display a thumb?

dmitrylyzo:
> I have no idea which one of the two is technically more compatible with the current structure though both look like satisfying solutions…

I am trying **Option 1** now. Bubble position should be corrected too - I just added `sliderBubbleTrack` div where `sliderBubble` will live. Chrome is already "perfect". But Firefox and Edge :man_facepalming:
* Firefox uses own style. I think to make slider to be the same everywhere.
* Edge crops scaled thumb, so slider should be larger - this breaks calculations. But negative margins will help.

Also, now I think **Option 2** can be done by negative margins on `input`, i.e. without custom thumb.

As for YoutTube,
IMO, element `ytp-progress-bar` controls playback - it has event handlers.
`ytp-scrubber-container` controls thumb position.
`ytp-scrubber-button` is thumb itself - it is scaled to 0 and unscaled on bar hover.
`ytp-scrubber-pull-indicator` - some indicator, I dunno what is for.

PR #645, opened 2019-12-20
Rework device icons + add Kodi icon

mcarlton00:
**Changes**

Adjusts the function providing device icons to match first on `DeviceName` and then checks against `AppName` before returning the default icon.  Also adds an icon for Kodi devices

### DIFF
--- a/src/controllers/dashboardpage.js
+++ b/src/controllers/dashboardpage.js
@@ -623,7 +623,7 @@ define(["datetime", "events", "itemHelper", "serverNotifications", "dom", "globa
             }
         },
         getClientImage: function (connection) {
-            var iconUrl = imageHelper.getDeviceIcon(connection.DeviceName);
+            var iconUrl = imageHelper.getDeviceIcon(connection);
             return "<img src='" + iconUrl + "' />";
         },
         getNowPlayingImageUrl: function (item) {
--- a/src/controllers/devices.js
+++ b/src/controllers/devices.js
@@ -74,7 +74,7 @@ define(["loading", "dom", "libraryMenu", "globalize", "scripts/imagehelper", "hu
             deviceHtml += '<div class="cardScalable">';
             deviceHtml += '<div class="cardPadder cardPadder-backdrop"></div>';
             deviceHtml += '<a is="emby-linkbutton" href="' + (canEdit ? "device.html?id=" + device.Id : "#") + '" class="cardContent cardImageContainer">';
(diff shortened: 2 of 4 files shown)


dkanada:
Did you get this icon from the same source @thornbill mentioned in the original issue? Maybe we should add a license file or link somewhere so people know where they're from in the future.

mcarlton00:
No, I just made this one replicating one of the Kodi logos I found online.  I had already finished making it by the time i heard back about the original source of the others, and I never went to find where it was

dkanada:
The issue with this logic is that most clients use DeviceName for the model or hostname and AppName for the client name, but the web client always uses Jellyfin Web for AppName and the browser model for DeviceName. We should really change one or the other for consistency, but I haven't thought about which is "more" correct in this case.

mcarlton00:
We've gone back and forth about DeviceName vs AppName several times, but we've never really reached a conclusion.  I figure doing it this way we can account for both, until a decision is made and other clients are updated to match.  Since the web client is probably still one of the most used interfaces, we probably shouldn't break that one here if possible.

The web browsers are also a unique situation where the AppName is technically "Jellyfin Web" in all of them, but we want to find the underlying system.  So maybe it makes more sense to match everything else based on AppName, and then do a second switch statement for the browsers only:

```
function getDeviceIcon(device) {
    var baseUrl = "img/devices/";
    switch (device.AppName) {
        case "Samsung Smart TV":
            return baseUrl + "samsung.svg";
        case "Xbox One":
            return baseUrl + "xbox.svg";
        case "Sony PS4":
            return baseUrl + "playstation.svg";
        case "Web Browser":
            return baseUrl + "html5.svg";
        case "Kodi":
            return baseUrl + "kodi.svg";
        case "Jellyfin Web":
            switch (device.Name) {
                case "Opera":
                case "Opera TV":
                    return baseUrl + "opera.svg";
                case "Chrome":
                    return baseUrl + "chrome.svg";
                case "Firefox":
                    return baseUrl + "firefox.svg";
                case "Edge":
                    return baseUrl + "edge.svg";
                case "Internet Explorer":
                    return baseUrl + "msie.svg";
                default:
                    return baseUrl + "other.svg";
            }
        default:
            return baseUrl + "other.svg";
    }
}
```

dkanada:
@mcarlton00 that sounds great. I don't think we have a `Web Browser` client though, so you should remove that. The generic browser icon should also be `html5.svg` instead of `other.svg` so it's unique.

PR #648, opened 2019-12-22
Fix coveredImage not being cover-sized

MrTimscampi:
**Changes**

Changed a background-size property for coveredImage from "100% 100%" to "cover".

Most of these had been fixed a year ago, but I noticed some of my posters being deformed and tracked it down to this. Turns out one background-size property hadn't been changed.

**Issues**
 
n/a


MrTimscampi:
Noticed another PR already fixing this, so I'm closing.

PR #649, opened 2019-12-22
Allow WEBP Uploading

XVicarious:
**Changes**

Allow WEBP images to be uploaded in imageuploader.

2/2 (jellyfin/jellyfin-apiclient-javascript#24)

### DIFF
--- a/src/components/imageuploader/imageuploader.template.html
+++ b/src/components/imageuploader/imageuploader.template.html
@@ -22,7 +22,7 @@
                 <div class="imageEditor-dropZone fieldDescription">
                     <div>${LabelDropImageHere}</div>
                     <output id="imageOutput" class="flex align-items-center justify-content-center" style="position: absolute;top:0;left:0;right:0;bottom:0;width:100%;"></output>
-                    <input type="file" accept="image/png,image/x-png,image/jpeg" id="uploadImage" name="uploadImage" style="position: absolute;top:0;left:0;right:0;bottom:0;width:100%;opacity:0;" />
+                    <input type="file" accept="image/png,image/x-png,image/jpeg,image/webp" id="uploadImage" name="uploadImage" style="position: absolute;top:0;left:0;right:0;bottom:0;width:100%;opacity:0;" />
                 </div>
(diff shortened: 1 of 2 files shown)


dkanada:
Maybe we should just allow adding any image file from the frontend and let the backend handle errors with unsupported formats. @Bond-009 I know we have quite a few formats listed in the skia project, what do you think?

Bond-009:
@dkanada Sounds good to me

dkanada:
@XVicarious what about using something like `file.type.contains("image")` instead to avoid other formats getting ignored in the future?

XVicarious:
How does that work? I'm not too versed in the specific frameworks, but I assumed that the vanilla JavaScript `String.prototype.startsWith` would both work and be appropriate for this application.

src/components/imageuploader/imageuploader.template.html
```diff
@@ -22,7 +22,7 @@ <h2 style="margin:0;">${HeaderAddUpdateImage}</h2>
                 <div class="imageEditor-dropZone fieldDescription">
                     <div>${LabelDropImageHere}</div>
                     <output id="imageOutput" class="flex align-items-center justify-content-center" style="position: absolute;top:0;left:0;right:0;bottom:0;width:100%;"></output>
-                    <input type="file" accept="image/png,image/x-png,image/jpeg" id="uploadImage" name="uploadImage" style="position: absolute;top:0;left:0;right:0;bottom:0;width:100%;opacity:0;" />
+                    <input type="file" accept="image/png,image/x-png,image/jpeg,image/webp" id="uploadImage" name="uploadImage" style="position: absolute;top:0;left:0;right:0;bottom:0;width:100%;opacity:0;" />
```
dkanada:
```suggestion
                    <input type="file" accept="image/*" id="uploadImage" name="uploadImage" style="position: absolute;top:0;left:0;right:0;bottom:0;width:100%;opacity:0;" />
```

anthonylavado: @XVicarious are you able to commit the suggestion from @dkanada?

PR #650, opened 2019-12-24
Align the page title with other text and icons in the header

MrTimscampi:
**Changes**

Changes the top margin for pageTitle from 0 to .3em in order to properly align the baseline of the title with the other text and icons in the headers.

Before:
Screenshot_2019-12-24 Jellyfin(2)

After:
Screenshot_2019-12-24 Jellyfin(3)

**Issues**

None

### DIFF
--- a/src/css/librarybrowser.css
+++ b/src/css/librarybrowser.css
@@ -93,7 +93,7 @@
     display: -webkit-inline-box;
     display: -webkit-inline-flex;
     display: inline-flex;
-    margin: 0 0 0 .5em;
+    margin: .3em 0 0 .5em;
     height: 1.7em;
     -webkit-box-align: center;
     -webkit-align-items: center;


dkanada:
Good catch! Did you notice any rules that could have replaced this one entirely? I've been trying to slim down the custom styling lately.

MrTimscampi:
I didn't notice any, but I'm not an expert in CSS typography. We could theoretically use the same styling for the buttons and the title, for alignment at least. But that'd require a bigger rework than this (And I'm not comfortable enough with the code yet :p )

To me, the issue seems to be that the container was aligned so that the base of the pageTitle div aligned with the baseline of the rest. However, the height of that div is set to 1.7em, making the font misaligned with the baseline (because it's not aligned with the base of the div to begin with).

We can probably get "better" numbers with the div set to 1em, then properly adjusting the top margin. I'll look into it to make this cleaner, because I'm not fond of using an odd number for margins...

PR #652, opened 2019-12-28
Show Year in Movie collections

phrac:
I am not sure if this is the proper way to do this as it seems Collections are only available in folder view currently.

**Changes**

**Issues**

### DIFF
--- a/src/components/cardbuilder/cardBuilder.js
+++ b/src/components/cardbuilder/cardBuilder.js
@@ -882,7 +882,7 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
                     }
                 }
 
-                if (options.showYear || options.showSeriesYear) {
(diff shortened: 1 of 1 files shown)


Larvitar:
This does not seem to be a good idea:
 a) This change will affect all views not just the collections and we might not want that.
 b) TV series in collections will still be displayed without year.

I've looked into this myself. I think the right way to do it is to make a change to [itemdetailpage.js](https://github.com/jellyfin/jellyfin-web/blob/master/src/controllers/itemdetailpage.js) and use `options.showYear`. I'll test it and make a PR later if it works properly.

Larvitar:
This one seems to be working correctly: [itemdetailpage.js](https://github.com/jellyfin/jellyfin-web/compare/master...Larvitar:master#diff-41db3f441cf79fd796d222e6202f767b)

phrac:
Yes, your patch works for me. Thank you.

PR #654, opened 2019-12-30
Add autoprefix to postcss

MrTimscampi:
**Changes**

Adds the postcss-loader plugin to webpack and enables the autoprefix plugin.

This allows us to remove all vendor prefixes from the CSS sources, since they'll be added automatically (in both Dev and Prod).

It should also prevent accidental compatibility oversight in future PRs.

**Important note:** We should add a browserlist key into package.json to define which browsers we officially support. We can specify specific browsers for kiosk environments such as WebOS.

**Security note:** This also updates a webpack plugin to fix potential XSS vulnerabilities, as reported by the NPM Audit tool.

**Issues**

None


MrTimscampi:
I added a starter browserslist which covers the following:

* All browsers with more than 0.5% use on the general web
* Last 2 versions of every browser
* NO dead browser

And then I add the following explicitly:

* Chrome 38, 47, 53, 56 and 63 (This covers WebOS TV 3.0+ and Tizen 2017+)
* Firefox ESR

More browsers are likely required, such as a Chrome equivalent for the PS4 browser (Since it uses WebKit) but I can't find much information aside from official Sony compatibility documents (https://www.sie.com/content/dam/corporate/jp/guideline/PS4_Web_Content-Guidelines_e.pdf) and the PS4 Developers Wiki (https://www.psdevwiki.com/ps4/Internet_Browser)

PR #655, opened 2019-12-30
Remove vendor prefixes in CSS

MrTimscampi:
**Changes**

Removes all vendor prefixes in CSS files to cleanup the code.

Note: Some selector prefixes remain. ~~These will be cleaned manually in further commits on this branch.~~ These should be handled separately after the style resets are cleaned up.

~~Note 2: Obviously, this requires extensive testing. If a rule is only defined as a vendor prefix, it WILL break stuff.~~

Edit: I tested this a bit and fixed what I found broken. It should be pretty good for now, as I was rather conservative in the changes made. Specific tests for WebOS and such should still be done, but autoprefixer should handle these with the config in #654 

Depends on #654 

**Issues**

None


MrTimscampi:
Some comments on the fixes applied in the second commit and the first commit:

* `-webkit-border-fit`: Re-added it, since I don't know what it does. It's non-standard, though, and an alternative should be found in the future.
* `-webkit-tap-highlight-color`: Removed since not standard and has a warning on MDN. I can see why it would be useful so can be re-added if needed on TVs and such, but all it did was set the highlight to black, so I don't think it's vital to have.
* `-webkit-font-smoothing`: Non-standard and not recommended for use in production on MDN. This shouldn't be handled by us but should be properly done by the browser in the first place, imo.
* `-webkit-touch-callout`: Deprecated, shouldn't be used. There's a `touch-action` property that is supported from Chrome 36 onwards.
* `-webkit-box-orient`, `-webkit-box-direction`: Replaced by the standard property.
* `-webkit-background-clip`: Replaced by the standard property.
* `-webkit-overflow-scrolling`: Removed because non standard, only supported on Safari.
* `-ms-overflow-style`: Replaced with the standard `scrollbar-width`.
* Material icons have been left alone to not screw things up. These should be removed from the codebase and imported from an NPM package like `material-design-icons` instead.

There's currently an issue with the alignment of headerLeft, which I'm trying to sort out.

dkanada:
The card height issue is fixed in one of my open pull requests.

src/components/cardbuilder/card.css
```diff
@@ -81,7 +80,6 @@ button {
     transition: none;
     border: 0 solid transparent;
     /* These both are needed in case cardBox is a button */
```
dkanada: I would also remove any comments related to the vendor prefixes like this one.

MrTimscampi: Done.

There's still a few prefixed stuff left, like animations, but I think we need to do more cleanup and refactoring before removing these safely.

src/components/emby-button/emby-button.css
```diff
@@ -25,7 +22,6 @@
     position: relative;
     font-weight: 600;
     /* Disable webkit tap highlighting */
```
dkanada: +1

src/components/emby-button/emby-button.css
```diff
@@ -119,7 +112,6 @@
     overflow: hidden;
     border-radius: 50%;
     /* Disable webkit tap highlighting */
```
dkanada: +2

src/components/emby-input/emby-input.css
```diff
@@ -10,10 +10,8 @@
     font-weight: inherit;
     padding: .4em .25em;
     /* Prevent padding from causing width overflow */
```
dkanada: +3

src/components/emby-slider/emby-slider.css
```diff
@@ -6,26 +6,17 @@ _:-ms-input-placeholder {
...
     user-select: none;
     outline: 0;
     color: #00a4dc;
-    -webkit-align-self: center;
-    -ms-flex-item-align: center;
     align-self: center;
     z-index: 1;
     cursor: pointer;
     margin: 0;
     /* Disable webkit tap highlighting */
```
dkanada: +4

src/components/remotecontrol/remotecontrol.css
```diff
@@ -70,25 +50,15 @@
...
-    -webkit-box-shadow: 0 0 1.9vh #000;
     box-shadow: 0 0 1.9vh #000;
     border: .1em solid #222;
     user-drag: none;
-    user-select: none;
-    -moz-user-select: none;
-    -webkit-user-drag: none;
-    -webkit-user-select: none;
-    -ms-user-select: none
+    user-select: none
```
dkanada: This could use a semicolon, it seems to have been removed for some reason.

src/components/themes/blueradiance/theme.css
```diff
@@ -451,7 +446,6 @@ html {
 }
 
 ::-webkit-scrollbar-track {
-    -webkit-box-shadow: inset 0 0 6px rgba(0, 0, 0, .3)
 }
```
dkanada: Looks like this rule can be removed.

src/components/themes/dark/theme.css
```diff
@@ -428,7 +424,6 @@ html {
 }
 
 ::-webkit-scrollbar-track {
-    -webkit-box-shadow: inset 0 0 6px rgba(0, 0, 0, .3)
 }
```
dkanada: Same here.

src/components/themes/purple-haze/theme.css
```diff
@@ -24,7 +24,6 @@ html {
 }
 
 .skinHeader.semiTransparent {
-    -webkit-backdrop-filter: none !important;
     backdrop-filter: none !important;
     background-color: rgba(0, 0, 0, .3);
     background: -webkit-gradient(linear, left top, left bottom, from(rgba(0, 0, 0, .6)), to(rgba(0, 0, 0, 0)));
```
dkanada: This should be a gradient right? Might need to run a search for this as well.

src/css/site.css
```diff
@@ -20,18 +20,12 @@ html {
...
-    -moz-user-select: none;
-    -ms-user-select: none;
     user-select: none
 }
 
 body {
     overflow-x: hidden;
-    background-color: transparent !important;
-    -webkit-font-smoothing: antialiased
+    background-color: transparent !important
```
dkanada: Another missing semicolon.

src/components/themes/purple-haze/theme.css
```diff
@@ -550,7 +545,7 @@ a[data-role=button] {
 }
 
 ::-webkit-scrollbar-track {
-    -webkit-box-shadow: inset 0 0 6px rgba(0, 0, 0, .3)
+    box-shadow: inset 0 0 6px rgba(0, 0, 0, .3)
 }
```
dkanada: I'm wondering if maybe these rules are actually still necessary, I wouldn't think so.

src/components/cardbuilder/card.css
dkanada: Oops, I hadn't read you're entire comment about the leftover prefixes. We can leave them for now, but I definitely want to remove the rest of the related comments since they won't make sense without the original context.

dkanada: Oops, I hadn't read your entire comment about the leftover prefixes. We can leave them for now, but I definitely want to remove the rest of the related comments since they won't make sense without the original context.

src/components/remotecontrol/remotecontrol.css
MrTimscampi: There's a bunch of semicolons missing all over the place, from what I've seen. The CSS cleanup PR mostly takes care of it, but I'll add them for the lines affected by this PR.

src/components/themes/purple-haze/theme.css
MrTimscampi: We should try to be as conservative as possible, I think. We can look into removing these eventually after the prefix stuff is taken care of.

I'm mainly trying to not make stuff break and have the change be as transparent as possible in the frontend :p

src/components/cardbuilder/card.css
MrTimscampi: I ended up going over the keyframes and some background gradients to remove the vendor-prefixed parts, since autoprefixer takes care of that.

I also cleaned up the few @supports statements I found.

src/components/themes/blueradiance/theme.css
MrTimscampi: I ended up leaving these in, but non-prefixed (since autoprefixer takes care of that if needed).

It still works in webkit but, for some reason, autoprefixed didn't re-add the rule.

src/components/themes/purple-haze/theme.css
MrTimscampi: I generally cleaned up the prefixed gradients. I don't know what the original idea was for adding a background color AND a gradient. Perhaps fallback for an old IE version that was never removed ?

I think we can safely get rid of it.

dkanada: I think it was indeed support for older browsers, so we can get rid of it for now.

src/css/librarybrowser.css
```diff
@@ -582,7 +531,7 @@
         display: none
     }
 }
-
+justify-content: left;
```
dkanada: here's another stray entry

src/css/librarybrowser.css
```diff
@@ -149,27 +133,20 @@
 }
 
 .headerLeft {
-    display: flex;
-    -webkit-align-items: center;
+    display: flex;justify-content: left;
```
dkanada: newline would be useful here

src/components/themes/dark/theme.css
```diff
@@ -437,7 +430,6 @@ html {
 
 ::-webkit-scrollbar-thumb:horizontal,
 ::-webkit-scrollbar-thumb:vertical {
-    -webkit-border-radius: 2px;
```
dkanada: There were a few places where you removed this and at least one where you just removed the vendor prefix instead.

src/css/librarybrowser.css
dkanada: actually the same attribute exists at the bottom of this rule

dkanada: The justify-content attribute can be removed here, it is also defined at the end of this rule.

src/components/themes/dark/theme.css
MrTimscampi: I re-added the ones that were removed as generic properties.

MrTimscampi:
I'll rebase this tonight (GMT+1) to fix the merge conflicts.

src/components/themes/appletv/theme.css
```diff
@@ -23,14 +23,20 @@ html {
...
+<<<<<<< HEAD
+@supports (backdrop-filter:blur(1.5em)) {
+    .skinHeader-blurred {
+        background: linear-gradient(to right, rgba(188, 188, 188, .7), rgba(167, 180, 183, .7), rgba(190, 181, 165, .7), rgba(173, 190, 194, .7), rgba(185, 199, 203, .7));
+        backdrop-filter: blur(1.5em)
+    }
+}
+
+=======
+>>>>>>> master
```
Bond-009: Don't think this is supposed to be here @dkanada

PR #656, opened 2019-12-30
Add linting for JS and CSS using browserslist

MrTimscampi:
**Changes**

The aim of this PR is to add linting for both JS and CSS that will point out the incompatibilities that can trigger problems on the browsers targeted via Browserslist.

This should help us find what needs to be polyfilled or when not to something.

Depends on #654 due to the need for browserslist.

**Issues**

None

### DIFF
--- a/src/assets/css/librarybrowser.css
+++ b/src/assets/css/librarybrowser.css
@@ -200,6 +200,7 @@
 }
 
 @media all and (min-width:40em) {
+
     .dashboardDocument .adminDrawerLogo,
     .dashboardDocument .mainDrawerButton {
         display: none !important
@@ -380,6 +381,8 @@
 
 .itemBackdrop {
     background-size: cover;
+    background-repeat: no-repeat;
+    background-position: center;
     height: 50vh;
     position: relative
 }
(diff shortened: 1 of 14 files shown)


MrTimscampi:
As of 29340db, everything that triggers an error is stuff that I either can't find a Polyfill already in use in the code or which has no Polyfill used in the code already.

Namely:

* `navigator.connection()` for Safari 12.1 and Chrome 38
* `Array.from()` for Chrome 38
* `window.requestIdleCallback()` for Safari 12.1, Edge 17 and Chrome 38
* `IntersectionObserver` for Chrome 38
* `Notification.requestPermission()` for Chrome 38
* `navigator.serviceWorker()` for Chrome 38
* `screen.orientation()` for Safari 12.1
* `navigator.getGamepads()` for Safari 12.1
* `document.hasFocus()` for Chrome 38

MrTimscampi:
With ce88355, stylelint is added with the following:

* Extends the "standard" stylelint config (These will return errors by default.)
* Throws warning for features that are not supported by all the browsers we target

The reason for throwing a warning instead of an error is because we want to see what needs a fallback, but the use of the syntax shouldn't be blocking in itself.

Currently stylelint throws a LOT of errors due to inconsistent style (Multiple thousands).

MrTimscampi:
Some explanations about the changes as of 14078a4 :

* **Added rules for JS deprecation**
  ESLint will now warn if functions, expressions or imports are marked for deprecation, allowing to refactor their use.
* **Added JQuery to environment**
  It's a dependency of the project, so adding it to the environment makes sense. We can also refactor a lot of code down the line to use JQuery.
* **Added RequireJS plugin**
  Check if `define()` and `require()` are used properly and in a consistent style.
* **Added JSDoc plugin**
  This only outputs warning as a means of seeing what needs to be documented properly. JSDoc was chosen as it's standardized correctly and supported by a ton of tools.

Issue #346, opened 2019-05-28
AirPlay support


dprandzioch:
**Describe the feature you'd like**
If you are on a Mac, many web players like YouTube support streaming the currently active video to an AirPlay compatible device. It would be great if the Jellyfin web interface would support this since the device-support, at least in the Apple ecosystem, is very limited (no iOS/tvOS app). Since the implementation of AirPlay in smart TVs is also increasing, it could be a good fit.

**Additional context**
<img width="357" alt="Bildschirmfoto 2019-05-28 um 21 12 57" src="https://user-images.githubusercontent.com/4499821/58505520-6d8f7200-818d-11e9-9e45-8f25bee2758a.png">

anthonylavado:
On iOS you can airplay if you fullscreen the video, but I agree - adding a dedicated button is a good way to go.

I tried to make it work a while ago and didn’t get anywhere. I’ll try again.

vaheed:
+1

Issue #132, opened 2019-02-12
Add support for ButterChurn visualizer


LogicalPhallacy:
When playing music it would be cool to have a toggle-able visualizer. Milkdrop is objectively the best visualizer ever made, and a web port exists for it. We should add support (with a toggle) for https://github.com/jberg/butterchurn

sparky8251:
As mentioned, this is likely non-trivial to add. The current web project isn't well structured and it is non-trivial to add existing dependencies to existing project files.

I'm more or less of the mind that "minor" UX enhancements and bug fixes are all we will do for `jellyfin-web` so the new `jellyfin-webng` isn't trying to catch up to a moving target.

PR #659, opened 2020-01-02
Add initial AirPlay support

anthonylavado:
**Changes**
Adds an AirPlay button to the web OSD controls. When clicked, it will populate with all available AirPlay hardware that is in range, using a system picker.

**Issues**
#346, and the corresponding [Fider #132](https://features.jellyfin.org/posts/132/airplay-support
) (kind of).

**Notes**
The Apple doc reference is here: https://developer.apple.com/documentation/webkitjs/adding_an_airplay_button_to_your_safari_media_controls

I am not actually listening for availability as they recommend, instead, I'm just displaying the button. Improvements welcome.

If testing, you can't be using `localhost`, because the Apple TV can't reach that. Also, subtitles do not render at this point in time.

### DIFF
--- a/src/components/htmlvideoplayer/plugin.js
+++ b/src/components/htmlvideoplayer/plugin.js
@@ -1640,7 +1640,7 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
         if (document.AirPlayEnabled) {
             return document.AirplayElement ? true : false;
         }
-        
+
         return false;
     };
 
--- a/src/components/htmlvideoplayer/plugin.js
+++ b/src/components/htmlvideoplayer/plugin.js
@@ -1658,7 +1658,6 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
             }
         } else {
             video.webkitShowPlaybackTargetPicker();
-
         }
     };
 
--- a/src/components/htmlvideoplayer/plugin.js
+++ b/src/components/htmlvideoplayer/plugin.js
@@ -1639,14 +1639,8 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
 
         if (document.AirPlayEnabled) {
             return document.AirplayElement ? true : false;
-        } else {
-            var video = this._mediaElement;
-            if (video) {
-                return video.webkitShowPlaybackTargetPicker();
-
-            }
         }
-
+        
         return false;
     };
 
--- a/src/components/htmlvideoplayer/plugin.js
+++ b/src/components/htmlvideoplayer/plugin.js
@@ -1522,7 +1522,7 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
                 list.push('PictureInPicture');
             }
         }
-        
+
         if (browser.safari || browser.iOS || browser.iPad) {
             list.push('AirPlay')
         }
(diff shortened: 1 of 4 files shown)


src/components/htmlvideoplayer/plugin.js
```diff
@@ -1591,6 +1595,39 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
     };
 
+    HtmlVideoPlayer.prototype.isAirPlayEnabled = function () {
+
+        if (document.AirPlayEnabled) {
+            return document.AirplayElement ? true : false;
+        } else {
+            var video = this._mediaElement;
+            if (video) {
+                return video.webkitShowPlaybackTargetPicker();
```
dkanada: I'm not sure we want this else statement since this method is only used to check the AirPlay status, not trigger a selection menu.

yarn.lock
```diff
@@ -4924,10 +4924,10 @@ wbuf@^1.1.0, wbuf@^1.7.3:
   dependencies:
     minimalistic-assert "^1.0.0"
 
-webcomponents.js-2@^0.7.24:
+webcomponents.js@^0.7.24:
```
dkanada: Another open pull request already has this change so if you have time it would be great to rebase without this file to prevent possible merge issues.

src/components/htmlvideoplayer/plugin.js
anthonylavado: Will adjust. It was a late night :-p

yarn.lock
anthonylavado: Will do. Must have been updated when I was building things.

EraYaN:
We could add the event listener separately and just track some state with it, and use it in the display condition. The availability of that event is also the give away that the browser supports AirPlay (if other browsers ever end up doing it).

src/components/htmlvideoplayer/plugin.js
```diff
@@ -1523,6 +1523,10 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
             }
         }
 
+        if (browser.safari || browser.iOS || browser.iPad) {
+            list.push('AirPlay')
```
JustAMan: I think we settled on double quotes as new standard, right @dkanada ?

```suggestion
            list.push("AirPlay")
```

src/components/htmlvideoplayer/plugin.js
```diff
@@ -1631,6 +1635,33 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
+            if (video) {
+                if (isEnabled) {
+                    video.requestAirPlay().catch(onAirPlayError);
+                } else {
+                    document.exitAirPLay().catch(onAirPlayError);
+                }
+            }
+        } else {
+            video.webkitShowPlaybackTargetPicker();
+
```
JustAMan:
```suggestion
```

src/components/htmlvideoplayer/plugin.js
```diff
@@ -1631,6 +1635,33 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
+        if (document.AirPlayEnabled) {
+            return document.AirplayElement ? true : false;
+        }
+
+        return false;
+    };
+
+    HtmlVideoPlayer.prototype.setAirPlayEnabled = function (isEnabled) {
+
```
JustAMan:
```suggestion
```

src/components/playback/playbackmanager.js
```diff
@@ -3854,6 +3859,9 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
                 if (player.supports('PictureInPicture')) {
                     list.push('PictureInPicture');
                 }
+                if (player.supports('AirPlay')) {
+                    list.push('AirPlay');
```
JustAMan: same about double quotes

src/components/htmlvideoplayer/plugin.js
dkanada: Nope we will eventually move to single quotes, just haven't changed the old code because it's a huge change.

anthonylavado:
@EraYaN From my testing, it’s handled at the system level and only supported in Safari, so it should never appear as a button unless you’re using Safari on a Mac, iOS, or iPadOS.

anthonylavado:
If all good, pls squash

PR #660, opened 2020-01-02
 Display year in "Collection" and "More like this"

Larvitar:
1. Show year for movies and series in "Collection" view
example_1
2. Show year for series in "More like this" (now it shows only for movies)
example_2


Larvitar:
#652

PR #663, opened 2020-01-03
Enable headroom only on mobile

MrTimscampi:
**Changes**

If using the desktop layout, don't make the header disappear on scroll.

Desktops should have way more than enough headroom nowadays for this not to be necessary, and it's actually pretty annoying to have part of the page disappear when scrolling, especially since it has the search button.

**Issues**

None

### DIFF
--- a/src/components/emby-scroller/emby-scroller.js
+++ b/src/components/emby-scroller/emby-scroller.js
@@ -155,7 +155,7 @@ define(['scroller', 'dom', 'layoutManager', 'inputManager', 'focusManager', 'bro
             initCenterFocus(this, this.scroller);
         }
 
-        if (bindHeader) {
+        if (bindHeader && layoutManager.mobile) {
             initHeadroom(this);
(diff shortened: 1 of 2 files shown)


dkanada:
Would also prefer to disable for TV layout and only use on mobile. As I mentioned on Matrix, I think we really need to nail down the navigation elements between the three different layouts. Right now they all just kind of feel tacked on.

MrTimscampi:
@dkanada The change properly disables the headroom on TV as well, but navigation to the top is still iffy, as it doesn't account for the headroom being there and, when going to the top of the page, it still doesn't go all the way to the top.

grafixeyehero:
> 
> 
> Have you tested this with the new item details page? What happens to the detail header when you scroll down all the way?

@dkanada good to go i just fix the style i will updated here https://github.com/jellyfin/jellyfin-web/pull/694
Desktop-mode-header

PR #665, opened 2020-01-04
Fix JS linting errors

MrTimscampi:
**Changes**

Depends on #656 

Work on fixing as much JS linting errors as reasonably possible before the parent PR is merged.

**Issues**

None


MrTimscampi:
So, after spending a day bashing my head against this, unfortunately this will require quite a bit of refactoring and cleaning up.

As such, I'll close this and get to refactoring stuff bit by bit. For now, I think we'll have to disable the advanced linting and simply adjust the few rules we have now to correspond to what the Standard config uses, then slowly improve over it from there.

PR #667, opened 2020-01-06
Update remotecontrol layout

Nickbert7:
I have changed the layout of the remote control buttons (bigger on mobile devices) and moved them into a collapse

Old mobile
image

New mobile
image
image

Old desktop
image

New desktop
image

image

### DIFF
--- a/src/components/remotecontrol/remotecontrol.css
+++ b/src/components/remotecontrol/remotecontrol.css
@@ -18,23 +18,23 @@
 }
 
 .btnArrowUp{
-border-radius: 40% 40% 10% 10%;
+    border-radius: 40% 40% 10% 10%;
 }
 
 .btnArrowLeft{
-border-radius: 40% 10% 10% 40%;
+    border-radius: 40% 10% 10% 40%;
 }
 
 .btnArrowRight{
-border-radius: 10% 40% 40% 10%;
+    border-radius: 10% 40% 40% 10%;
 }
 
 .btnArrowDown{
-border-radius: 10% 10% 40% 40%;
+    border-radius: 10% 10% 40% 40%;
 }
 
 .btnOk{
-border-radius: 10%;
+    border-radius: 10%;
 }
 
 .nowPlayingPageTitle {
--- a/src/components/remotecontrol/remotecontrol.css
+++ b/src/components/remotecontrol/remotecontrol.css
@@ -12,11 +12,6 @@
     text-align: center
 }
 
-.btnArrowUp, .btnArrowLeft, .btnArrowRight, .btnArrowDown, .btnOk{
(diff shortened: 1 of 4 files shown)


src/components/remotecontrol/remotecontrol.css
```diff
@@ -12,6 +12,31 @@
     text-align: center
 }
 
+.btnArrowUp, .btnArrowLeft, .btnArrowRight, .btnArrowDown, .btnOk{
+    border: 2.7px solid rgba(255, 255, 255, .6);
+    background-color: rgba(0, 0, 0, .5);
+}
+
+.btnArrowUp{
+border-radius: 40% 40% 10% 10%;
```
JustAMan: I think we want indentation here and in similar places

src/components/remotecontrol/remotecontrol.js
```diff
@@ -215,7 +215,7 @@ define(["browser", "datetime", "backdrop", "libraryBrowser", "listView", "imageL
                 context.querySelector(".sendTextSection").classList.add("hide");
             }
 
-            if (!currentPlayer.isLocalPlayer) {
+            if (-1 != supportedCommands.indexOf("MoveUp" || "MoveDown" || "MoveRight" || "MoveRight" || "MoveLeft" || "Select" || "Back" || "ToggleContextMenu" || "GoToSearch" || "GoToSettings" || "GoHome") && !currentPlayer.isLocalPlayer) {
```
JustAMan: are you sure this is doing what you think it does? :)
to me it seems you're actually only checking for `.indexOf("MoveUp")` ;)

JustAMan:
Disagreed. We _already_ have those buttons, making them bigger on mobile is good for UX.

dkanada:
It's not about making them bigger, it's the style itself.

src/components/remotecontrol/remotecontrol.js
Nickbert7: Hm, yes it seems you're right... 😒
Do you have a hint for me how do solve it without checking for them seperate?
Or should I just keep the collapse visible with only disabled buttons?

I found out that I dont have to hide each button, they get disabled if not supported:
image

Nickbert7:
I thought on getting them similar to the play button on the mobile layout.
image

dkanada:
Ah, I was actually migrating the play buttons to a flat style. You can see the one I added on the media cards with no border and a nice transition effect.

Nickbert7:
Ok, I'll have a look on it tomorrow or next week.
Maybe I should also check some other remote control apps to get some impressions.

Nickbert7:
I have removed the border and use the color of the submit button now
image

PR #672, opened 2020-01-06
Hide Cast button on TV layout

MrTimscampi:
**Changes**

I fail to see a proper use case for the Cast button on the TV UI, since it'll mainly be used on TVs (Either directly on the TV or on a box connected to it). So this commit keeps the cast button hidden in the TV Layout.

**Issues**

None

### DIFF
--- a/src/scripts/librarymenu.js
+++ b/src/scripts/librarymenu.js
@@ -47,7 +47,9 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
                 headerSearchButton.classList.remove("hide");
             }
 
-            headerCastButton.classList.remove("hide");
+            if (!layoutManager.tv) {
+                headerCastButton.classList.remove("hide");
+            }
         } else {
             headerHomeButton.classList.add("hide");
             headerCastButton.classList.add("hide");
--- a/src/scripts/librarymenu.js
+++ b/src/scripts/librarymenu.js
@@ -103,7 +103,10 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
 
         headerUserButton.addEventListener("click", onHeaderUserButtonClick);
         headerHomeButton.addEventListener("click", onHeaderHomeButtonClick);
-        headerCastButton.addEventListener("click", onCastButtonClicked);
+
+        if (!layoutManager.tv) {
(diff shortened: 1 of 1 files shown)


anthonylavado:
Holding as per request my @MrTimScampi

MrTimscampi:
@anthonylavado Change made in d963fe5, good to go on my end :)

JustAMan:
I wonder why we should actually hide it... what if a user has some headless (but speaker-ful) system which can play music, and uses the TV to cast music to it?

MrTimscampi:
Do the TV browsers even support the Cast API ? I'd figure they would pretty restricted.

Wouldn't they use a smartphone to cast instead, which makes more sense ? Usually you'd control a headless system through a phone or a tablet, for which the TV layout isn't adapted anyway.

Part of my decision for removing it is that real estate is pretty limited in the TV layout, as everything is pretty big. So hiding a non-essential feature made sense in order to gain some visual cleanliness.

See this screenshot for an example of the busyness of the header in TV layout:

anthonylavado:
The button does have the remote control interface behind it, but I agree that it shouldn’t be in the TV layout

JustAMan:
Again, I think there are different "cast" things here. While I agree that average user isn't likely casting from a TV to a Chromecast, casting a playlist to a headless player seems more viable to me (however this is actually a future option for "web ui config" @dkanada wants to create).

dkanada:
@JustAMan has a point, you can control other clients from the same cast menu.

MrTimscampi:
I think there's a better discussion about this button's purpose to be had, here.

In every other app, the "Cast" icon means "send content to a Chromecast" or a similar action. To me, and I think to most users, a cast button leading to a remote control doesn't really make sense, especially since nobody else uses that icon this way.

I think a better pattern would be to have this icon handle Chromecast and Airplay, then handle playback to other clients from the details page, for example.  
Have the play button be a dropdown button (A button with a dropdown arrow on the right). By default, clicking on the Play button leads to playback on the current client, and the dropdown would show "Play on <client name>". If no other clients are paired to the server, it simply shows a "No other client available for playback".

This makes more sense for discovery of features and for the iconography used.

JustAMan:
Absolutely agree, this makes much more sense that way! :)

PR #680, opened 2020-01-09
Use scrollManager for TV layout in all browser and improve scroll feel

### LINKED ISSUE #675: Missing Scrolling support on WebOS
Hello, I'm not sure if this is the right place to report this. So I installed the IPK and it works just fine including streaming to the device from the jellyfin server, but the interface will not scroll. Essentially you are stuck with only the content that fits on screen making the interface pretty much unusable.

MrTimscampi:
**Changes**

* Enable scollManager for the TV layout, regardless of browser.
* Use smooth scrolling in scrollManager on WebOS, Tizen, Firefox and Chrome. More browsers are likely supported but need testing.
* Use smooth scrolling for the scroller element on Firefox, even in TV layout.

**Issues**

Fix #675


src/components/scrollManager.js
```diff
@@ -372,7 +372,7 @@ define(["dom", "browser", "layoutManager"], function (dom, browser, layoutManage
      */
     function useSmoothScroll() {
 
-        if (browser.tizen) {
+        if (browser.tizen || browser.web0s || browser.firefox || browser.chrome) {
```
dkanada: I know lots of people don't like smooth scroll, so we can't just enable it everywhere by default.

MrTimscampi:
@dkanada I removed the changes other than enabling the scroll manager on TV layouts. This way, the PR fixes the wonky scrolling in TV layout.

The original code, imo, made no sense:

* Disable it if not in TV layout
* Enable it for Tizen and WebOS
* But disable it for everything else

Imo it makes sense to have the scroll manager always enabled in the layout that uses mainly keyboard navigation. The other changes can be discussed once we get to them later on :)

PR #681, opened 2020-01-10
Migrate fully from RequireJS to Webpack

MrTimscampi:
**Changes**

Also known as "The Humongous Webpack Fix Pull Request™".

In advance, sorry for whoever has to review this, I couldn't really divide it in anyway since it's basically following the trail of errors until something starts working...

Most of the code changes are moving define() statements to their own files in order to be able to alias them in Webpack.

It does the following:

* Move EVERY SINGLE define() statement into their own file (one per statement)
* Add aliases for EVERY SINGLE RequireJS module used into webpack.common.js
* Change the entry point to site.js (Since bundle.js doesn't exist anymore)
* Change the target to "web" in order to build for browsers
* Alias "css!" to "style-loader" and "text!" to raw-loader
* Divide the bundle into chunks
* Add hashes to every bundle in order to force cache busting
* Create a vendor bundle so that app dependencies don't need to be downloaded again by the user between versions if no change were made to them
* Add Babel to transpile everything into whatever is supported by the lowest browsers in our browserslist

That's it for now. PR is WIP and there's still a lot to fix, so I'll edit this when there's an update.

**Update 1 - 2020/01/10 - 7PM (GMT+1)**

This is now at the point where the app is bootstrapping properly and tries to start. However, a few issues arise with this part of the process.  
Webpack complains about not being to statically analyze a bunch of require() statements in the viewContainer and the homesections. This analysis is needed by Webpack and can't be disabled.

As such, the next step is to rewrite how viewContainer and homesections load their modules, by removing the automatic loading of things and loading everything statically on initialization.

The plugins are also disabled at the moment due to needing some work in regards to loading as well.

**Update 2 - 2020/01/10 - 10PM (GMT+1)**

The few problematic require() statements have been fixed. Most of these were switch...case statements that set a variable that was then added to the require. The require() statement were duplicated and moved in order to make them static. Duplicate code, but we can't fix that without some major refactoring and I'm not doing that here (I'll break enough stuff with this process as-is :) )

I'm now encountering an error in the next part of the app starting process (Inside onAppReady, in the main require() block). I'm trying to track it down, but RequireJS debugging is particularly unhelpful.

**Update 3 - 2020/01/11 - 4PM (GMT+1)**

As of this update, the client builds and bootstraps properly, with only a few warnings pertaining to the use of expressions as dependencies in require() statements.

The issue with document-register-element and an issue with lazyLoader were fixed. In the process, the "scroll" version of lazyLoader has been removed and [a polyfill for intersectionObserver](https://github.com/w3c/IntersectionObserver/tree/master/polyfill) will be added in a future commit.

**Update 4 - 2020/01/12 - 2AM (GMT+1)**

The router is now working and you can navigate pages somewhat. They are incredibly broken and lack most of the CSS since a lot of things still aren't loading properly.  
The ConnectionManager is also still an issue, which means you can't connect to a server.

**Update 5 - 2020/01/15 - 6:40PM (GMT+1)**

Theme loading has been fixed, as well as images and icons. The client now starts fully without issues and server connection works.

Remaining issues currently are:

* Impossible to connect as a user
* Error dialog doesn't show up (Error in console)
* Playback plugins are currently disabled

**Issues**

To complete once the migration is done.


MrTimscampi:
Deprecated in favor of #862

Issue #623, opened 2019-12-08
Event Listeners are not registered correctly inside "inputManager"


nerxses:
**Describe the bug**
I came across this bug by searching for the reason, why i cannot control a photo slideshow with my arrowkeys. As it turns out the "exists" check for counter "eventListenerCount" inside the methods  "on" and "off"  of  "src/scripts/inputManager" is  wrong. 

''fix on the current master branch''

**To Reproduce**
1. Go to a photo libary
2. start slideshow
3. wait till bottombar is closed after x seconds (why? this "feature" is annyoing see method "onInput" inside "slideshow.js")
4. press right/left Arrowkey (nothing will happen)

**Expected behavior**
right/left Arrowkey should move you to the next/previous picture

**Additional context**
this is the original code of function "on"/"off" inside "src/scripts/inputManager"

    var eventListenerCount = 0;
    function on(scope, fn) {
        if (eventListenerCount) {
            eventListenerCount++;
        }
        dom.addEventListener(scope, 'command', fn, {});
    }

    function off(scope, fn) {
        if (eventListenerCount) {
            eventListenerCount--;
        }
        dom.removeEventListener(scope, 'command', fn, {});
    }

this code will result in the expected behaveior.

    var eventListenerCount = 0;
    function on(scope, fn) {
        if (typeof eventListenerCount != "undefined") {
            eventListenerCount++;
        }
        dom.addEventListener(scope, 'command', fn, {});
    }
    function off(scope, fn) {
        if (typeof eventListenerCount != "undefined") {
            eventListenerCount--;
        }
        dom.removeEventListener(scope, 'command', fn, {});
    }

dmitrylyzo:
IMHO you are right - the code is incorrect.

But...

`eventListenerCount` is defined in `var eventListenerCount = 0;`

So, your code is equal to:
```
var eventListenerCount = 0;
function on(scope, fn) {
    eventListenerCount++;
    dom.addEventListener(scope, 'command', fn, {});
}
function off(scope, fn) {
    eventListenerCount--;
    dom.removeEventListener(scope, 'command', fn, {});
}
```
which is good for the ideal case: `off` calls correspond to `on` ones.

Ideally, `eventListenerCount` should be incremented on successful `addEventListener` and decremented on successful `removeEventListener`.

nerxses:
[...]

Issue #621, opened 2019-12-08
LG WebOS Transport Controls not accessable


Narfinger:
Using jellyfin-webos as wrapper, the transport controls while playing a movie seem inaccessable.
Ok button does nothing and back button returns to the info screen of a file.

The expected behavior is that the ok button brings up the transport controls.

WebOS: 5.50.55
Jellyfin-web: 398b399ec60247ffa659424231eaa05dce2b7b45
Jellyfin: 6f283d80dc00fe63b59dfb7a134528f295615da7

JustAMan:
What are "transport controls" precisely?

Narfinger:
Transport controls is the bar that contains play, pause, forward, backward etc.

JustAMan:
So, when you start the playback, does it already miss the controls, or are they present but you cannot use them?

Narfinger:
Starting playback shows them but after it vanishes the ok button (binding
to enter on a keyboard, I think) does not bring it back up.

JustAMan:
@dmitrylyzo any ideas? Is this broken in the same way on Tizen?

@Narfinger can you test if it works on desktop browser but with TV mode enabled (not sure how to enable it, but reading the source could give pointers) by e.g. pressing Enter?

dmitrylyzo:
@JustAMan yes, on `Enter` OSD don't show up both in Tizen 4 (real) and Chrome in TV mode. I look git log - `Enter` handling (as `Pause`) was removed in 658aedf96a4e337f7e1e3980ba9cb9d0b53095b6.

Currently, for TV layout, OSD shows up on navigation keys (in addition to other shortcuts).Not totally sure, but I think it will be safe to catch `Enter` in TV layout and just `showOsd`. To do this, it can be added to the `NavigationKeys` array.

JustAMan:
I personally feel that in TV layout Enter should pause/unpause _and_ show OSD controls at the same time, I think this is also a way WebOS YouTube app behaves.

dmitrylyzo:
In Tizen 4 YouTube:
* `OK`/`Enter` - show OSD
* `Space` - show OSD and pause
* `Back`/`Esc` - hide OSD (if visible) or back to previous page.

JustAMan:
Yeah, I think on WebOS YouTube uses same scheme (though I didn't test a `Space` as I don't have a keyboard hooked to my TV). Looks like a solid keyboard scheme to follow.

Narfinger:
[...]

PR #702, opened 2020-01-17
Add playback control from TV remote

dmitrylyzo:
Fix event subscription in `inputManager` (required by next change). (#623)

Add key handling to support LG and Samsung remotes (#303):
* Play
* Pause
* PlayPause
* Stop
* Rewind
* FastForward
* Previous Track
* Next Track
* Back _(Tizen app do not handle "Back" by default; for WebOS app it is better to disable built-in handling)_

Add shortcuts to video playback (#621):
* `Enter` - show OSD
* `Escape`/`Back` - hide OSD / go to back
* I left `Space` as is - pause and show OSD only if the latter is hidden

In addition, `Escape` on TV is now "Go to Back".

**Remarks**
I found out how to disable `Back` built-in handling in WebOS.
Need to add `disableBackHistoryAPI: true` to `appinfo.json` of `jellyfin-webos`.
[http://webostv.developer.lge.com/develop/app-developer-guide/back-button/](http://webostv.developer.lge.com/develop/app-developer-guide/back-button/)

**Limitations**
In TV browser (at least Tizen 4), `Back` and `Escape` are handled as "Go to Back" by browser itself - webapp doesn't receive event. So, `Back`/`Escape` functionality doesn't work in TV browser.

### DIFF
--- a/src/components/keyboardnavigation.js
+++ b/src/components/keyboardnavigation.js
@@ -1,35 +1,108 @@
-define(['inputManager', 'focusManager'], function(inputManager, focusManager) {
-    'use strict';
+define(["inputManager", "layoutManager"], function (inputManager, layoutManager) {
+    "use strict";
 
     console.log("keyboardnavigation");
 
+    /**
+     * Key name mapping.
+     */
+    // Add more to support old browsers
+    var KeyNames = {
+        13: "Enter",
+        19: "Pause",
+        27: "Escape",
+        32: "Space",
+        37: "ArrowLeft",
+        38: "ArrowUp",
+        39: "ArrowRight",
+        40: "ArrowDown",
+        412: "MediaRewind", // MediaRewind (Tizen/WebOS)
+        413: "MediaStop", // MediaStop (Tizen/WebOS)
+        415: "MediaPlay", // MediaPlay (Tizen/WebOS)
+        417: "MediaFastForward", // MediaFastForward (Tizen/WebOS)
+        461: "Back", // Back (WebOS)
+        10009: "Back", // Back (Tizen)
+        10232: "MediaTrackPrevious", // MediaTrackPrevious (Tizen)
+        10233: "MediaTrackNext", // MediaTrackNext (Tizen)
+        10252: "MediaPlayPause" // MediaPlayPause (Tizen)
+    };
+
+    /**
+     * Returns key name from event.
+     *
+     * @param {KeyboardEvent} keyboard event
+     * @return {string} key name
+     */
+    function getKeyName(event) {
+        return KeyNames[event.keyCode] || event.key;
+    }
+
     function enable() {
-        document.addEventListener('keydown', function(e) {
+        document.addEventListener("keydown", function (e) {
             var capture = true;
 
-            switch (e.keyCode) {
-                case 37: // ArrowLeft
-                    inputManager.handle('left');
+            switch (getKeyName(e)) {
+                case "ArrowLeft":
+                    inputManager.handle("left");
                     break;
-                case 38: // ArrowUp
-                    inputManager.handle('up');
+                case "ArrowUp":
+                    inputManager.handle("up");
                     break;
-                case 39: // ArrowRight
-                    inputManager.handle('right');
+                case "ArrowRight":
+                    inputManager.handle("right");
                     break;
-                case 40: // ArrowDown
-                    inputManager.handle('down');
+                case "ArrowDown":
+                    inputManager.handle("down");
                     break;
+
+                case "Back":
+                    inputManager.handle("back");
+                    break;
+
+                case "Escape":
+                    if (layoutManager.tv) {
+                        inputManager.handle("back");
+                    } else {
+                        capture = false;
+                    }
+                    break;
+
+                case "MediaPlay":
+                    inputManager.handle("play");
+                    break;
+                case "Pause":
+                    inputManager.handle("pause");
+                    break;
+                case "MediaPlayPause":
+                    inputManager.handle("playpause");
+                    break;
+                case "MediaRewind":
+                    inputManager.handle("rewind");
+                    break;
+                case "MediaFastForward":
+                    inputManager.handle("fastforward");
+                    break;
+                case "MediaStop":
+                    inputManager.handle("stop");
+                    break;
+                case "MediaTrackPrevious":
+                    inputManager.handle("previoustrack");
+                    break;
+                case "MediaTrackNext":
+                    inputManager.handle("nexttrack");
+                    break;
+
                 default:
                     capture = false;
             }
+
             if (capture) {
                 console.log("Disabling default event handling");
                 e.preventDefault();
             }
         });
-
     }
+
     return {
         enable: enable
     };
--- a/src/components/keyboardnavigation.js
+++ b/src/components/keyboardnavigation.js
@@ -27,6 +27,20 @@ define(["inputManager", "layoutManager"], function (inputManager, layoutManager)
         10252: "MediaPlayPause" // MediaPlayPause (Tizen)
     };
 
+    var hasFieldKey = false;
+    try {
+        hasFieldKey = "key" in new KeyboardEvent("keydown");
+    } catch (e) {
+        console.log("error checking 'key' field");
+    }
+
+    if (!hasFieldKey) {
+        // Add [a..z]
+        for (var i = 65; i <= 90; i++) {
+            KeyNames[i] = String.fromCharCode(i).toLowerCase();
+        }
+    }
+
     /**
      * Returns key name from event.
      *
@@ -104,6 +118,7 @@ define(["inputManager", "layoutManager"], function (inputManager, layoutManager)
     }
 
     return {
-        enable: enable
+        enable: enable,
+        getKeyName: getKeyName
     };
 });
--- a/src/components/keyboardnavigation.js
+++ b/src/components/keyboardnavigation.js
@@ -16,15 +16,24 @@ define(["inputManager", "layoutManager"], function (inputManager, layoutManager)
         38: "ArrowUp",
         39: "ArrowRight",
         40: "ArrowDown",
-        412: "MediaRewind", // MediaRewind (Tizen/WebOS)
-        413: "MediaStop", // MediaStop (Tizen/WebOS)
-        415: "MediaPlay", // MediaPlay (Tizen/WebOS)
-        417: "MediaFastForward", // MediaFastForward (Tizen/WebOS)
-        461: "Back", // Back (WebOS)
-        10009: "Back", // Back (Tizen)
-        10232: "MediaTrackPrevious", // MediaTrackPrevious (Tizen)
-        10233: "MediaTrackNext", // MediaTrackNext (Tizen)
-        10252: "MediaPlayPause" // MediaPlayPause (Tizen)
+        // MediaRewind (Tizen/WebOS)
+        412: "MediaRewind",
+        // MediaStop (Tizen/WebOS)
+        413: "MediaStop",
+        // MediaPlay (Tizen/WebOS)
+        415: "MediaPlay",
+        // MediaFastForward (Tizen/WebOS)
+        417: "MediaFastForward",
+        // Back (WebOS)
+        461: "Back",
+        // Back (Tizen)
+        10009: "Back",
+        // MediaTrackPrevious (Tizen)
+        10232: "MediaTrackPrevious",
+        // MediaTrackNext (Tizen)
+        10233: "MediaTrackNext",
+        // MediaPlayPause (Tizen)
+        10252: "MediaPlayPause"
     };
 
     var hasFieldKey = false;
--- a/src/controllers/playback/videoosd.js
+++ b/src/controllers/playback/videoosd.js
@@ -1105,7 +1105,6 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
                 case "Enter":
                     showOsd();
                     break;
-
                 case "Escape":
                 case "Back":
                     // Ignore key when some dialog is opened
@@ -1114,52 +1113,45 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
                         e.stopPropagation();
                     }
                     break;
-
                 case "k":
                     playbackManager.playPause(currentPlayer);
                     showOsd();
                     break;
-
                 case "l":
                 case "ArrowRight":
                 case "Right":
                     playbackManager.fastForward(currentPlayer);
                     showOsd();
                     break;
-
                 case "j":
                 case "ArrowLeft":
                 case "Left":
                     playbackManager.rewind(currentPlayer);
                     showOsd();
                     break;
-
                 case "f":
                     if (!e.ctrlKey && !e.metaKey) {
                         playbackManager.toggleFullscreen(currentPlayer);
                         showOsd();
                     }
                     break;
-
                 case "m":
                     playbackManager.toggleMute(currentPlayer);
                     showOsd();
                     break;
-
                 case "NavigationLeft":
                 case "GamepadDPadLeft":
                 case "GamepadLeftThumbstickLeft":
-                // Ignores gamepad events that are always triggered, even when not focused.
+                    // Ignores gamepad events that are always triggered, even when not focused.
                     if (document.hasFocus()) {
                         playbackManager.rewind(currentPlayer);
                         showOsd();
                     }
                     break;
-
                 case "NavigationRight":
                 case "GamepadDPadRight":
                 case "GamepadLeftThumbstickRight":
-                // Ignores gamepad events that are always triggered, even when not focused.
+                    // Ignores gamepad events that are always triggered, even when not focused.
                     if (document.hasFocus()) {
                         playbackManager.fastForward(currentPlayer);
                         showOsd();
--- a/src/controllers/playback/videoosd.js
+++ b/src/controllers/playback/videoosd.js
@@ -437,6 +437,11 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
                 });
                 currentVisibleMenu = null;
                 toggleSubtitleSync("hide");
+
+                // Firefox does not blur by itself
+                if (document.activeElement) {
+                    document.activeElement.blur();
+                }
             }
         }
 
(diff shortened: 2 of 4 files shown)


src/controllers/playback/videoosd.js
```diff
@@ -1100,6 +1100,20 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
             }
 
             switch (e.key) {
+                case "Enter":
+                    showOsd();
+                    break;
+
+                case "Escape":
+                case "RCUBack": // WebOS back
+                case "XF86Back": // Tizen back
```
dmitrylyzo: In `keyboardnavigation`, I added `getKeyName`, which can be used here to map keys - this will also fix lack of `key` field in `KeyboardEvent` on old browsers (WebOS 2/3).

JustAMan:
> I found out how to disable `Back` built-in handling in WebOS.

What is wrong with `Back` handled by WebOS?..

src/controllers/playback/videoosd.js
JustAMan: Why did't you use it here already, is there a reason for it?

src/controllers/playback/videoosd.js
```diff
@@ -1280,7 +1294,7 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
                 showOsd();
                 inputManager.on(window, onInputCommand);
                 dom.addEventListener(window, "keydown", onWindowKeyDown, {
-                    passive: true
+                    capture: true
```
JustAMan: why?

dmitrylyzo:
> What is wrong with `Back` handled by WebOS?..

Hide OSD functionality will not work (as I said for TV browsers).

src/controllers/playback/videoosd.js
dmitrylyzo: to `stopPropagation` after OSD hide before `keyboardnavigation` make "go to back"

src/controllers/playback/videoosd.js
dmitrylyzo: Just not sure. Does it (import `keyboardnavigation` here) look good?

JustAMan: For me it looks okay, @MrTimscampi is now working on re-ordering imports, maybe he have a say about that?

src/controllers/playback/videoosd.js
JustAMan: Oh, right, I thought `passive` did the opposite thing, like "this element should be passive, JS function does all the stuff" LOL

JustAMan:
If we disable it being handled we'd have to manually navigate the history when it's being pressed, right?

dmitrylyzo:
> If we disable it being handled we'd have to manually navigate the history when it's being pressed, right?

This is done already in `keyboardnavigation` calling `inputManager.handle("back");`

JustAMan:
Upon reading the link about WebOS, it seems we'd have to change Back handling a bit, as they require to quit to "Home" view when app has no "back history":

> If the current page is the entry page to your app, move back to the TV's Home screen.

So we have to somehow handle it a bit differently as stated here: http://webostv.developer.lge.com/develop/app-developer-guide/back-button/#platformBack

dmitrylyzo:
> they require to quit to "Home" view when app has no "back history":

You are right, I missed it.
Moreover, Samsung recommends asking the user to exit.

From "web" we cannot do this because `webOS` object is not available. Only app has access. The same for Tizen.

In Tizen app, I just have to do this - no built-in handling.
I count `historyDepth` on each `pushstate`/`popstate` and reset counter on some pages: login, home, selectserver - so as not to unwind the whole history.
https://github.com/dmitrylyzo/jellyfin-tizen/blob/ea271573c1342276f63f13b9ee04aa5cadd5d54f/tizen.js#L130-L163
And then ask user
https://github.com/dmitrylyzo/jellyfin-tizen/blob/ea271573c1342276f63f13b9ee04aa5cadd5d54f/tizen.js#L48-L68

But I chose "fat client" - have access to `tizen` object.

Earlier, I tried "thin client", but open Jellyfin in frame to keep `tizen` object (spied somewhere). Now, this is left incomplete. I think the main reason is that I wanted Jellyfin server select style, but didn't want to write it myself.

For testing purpose, I also have simplest (window redirect) "thin client" with hardcoded server url.

dmitrylyzo:
Decided to update Tizen app (the "fat" one; I abandoned it "slightly") for `jellyfin-web:master`. It turns out that in my case only the application should process "Go back" (still manually). Otherwise, exiting the application does not work.

src/controllers/playback/videoosd.js
jma-odoo: No worries with imports on my end, it's only module definitions that were an issue :)

MrTimscampi: No issues on my end, the changes to modules are only related to define() (One per file, have to respect relative path, extensions needed for css! etc)

dmitrylyzo:
I played around with "Go back" handling and seemed to find a solution for #675. But we have to handle "Go back" manually.

In webos app, I open JF in `iframe` and pass `webOS` object into `contentWindow` - this solves two issues:
* iframe scrolls by wheel and scrollManager (#675)
* we can call `webOS.platformBack` from JF

_I can create PR in `jellyfin-webos`._

To handle "Go back":
In `inputManager`, on `back` command, I check `appRouter.canGoBack` and call `appHost.exit` if can not.
In `appHost` I ask user for exit (Tizen app only) and exit in platform-specific way.
In `appRouter.canGoBack`, I added test for opened dialog to be able to "go back" from it.

<details>
<summary>Changes</summary>

```diff
diff --git a/src/components/appRouter.js b/src/components/appRouter.js
index 8643e3f47..88c8ae0e6 100644
--- a/src/components/appRouter.js
+++ b/src/components/appRouter.js
@@ -551,7 +551,7 @@ define(['loading', 'globalize', 'events', 'viewManager', 'layoutManager', 'skinM
             return false;
         }
 
-        if (curr.type === 'home') {
+        if (!document.querySelector(".dialogContainer") && curr.type === 'home') {
             return false;
         }
         return page.canGoBack();
diff --git a/src/components/apphost.js b/src/components/apphost.js
index b1b1c30dc..9050aa944 100644
--- a/src/components/apphost.js
+++ b/src/components/apphost.js
@@ -311,6 +311,52 @@ define(["appSettings", "browser", "events", "htmlMediaHelper"], function (appSet
         });
     }
 
+    /**
+      * Do exit according to platform
+     */
+    function doExit() {
+        try {
+            if (window.NativeShell) {
+                window.NativeShell.AppHost.exit();
+            } else if (browser.tizen) {
+                tizen.application.getCurrentApplication().exit();
+            } else if (browser.web0s) {
+                webOS.platformBack();
+            } else {
+                window.close();
+            }
+        } catch (err) {
+            console.log("error closing application: " + err);
+        }
+    }
+
+    var exitPromise;
+
+    /**
+      * Ask user fo exit
+     */
+    function askForExit() {
+        if (!!exitPromise) {
+            return;
+        }
+
+        require(["actionsheet"], function (actionsheet) {
+            exitPromise = actionsheet.show({
+                    title: Globalize.translate('Exit?'),
+                    items: [
+                        {id: 'yes', name: Globalize.translate('Yes')},
+                        {id: 'no', name: Globalize.translate('No')}
+                    ]
+                }).then(function (value) {
+                    if (value === 'yes') {
+                        doExit();
+                    }
+                }).finally(function () {
+                    exitPromise = null;
+                });
+        });
+    }
+
     var deviceId;
     var deviceName;
     var appName = "Jellyfin Web";
@@ -326,16 +372,10 @@ define(["appSettings", "browser", "events", "htmlMediaHelper"], function (appSet
             alert("setWindowState is not supported and should not be called");
         },
         exit: function () {
-            if (window.NativeShell) {
-                window.NativeShell.AppHost.exit();
-            } else if (browser.tizen) {
-                try {
-                    tizen.application.getCurrentApplication().exit();
-                } catch (err) {
-                    console.log("error closing application: " + err);
-                }
+            if (!!window.appMode && browser.tizen) {
+                askForExit();
             } else {
-                window.close();
+                doExit();
             }
         },
         supports: function (command) {
diff --git a/src/scripts/inputManager.js b/src/scripts/inputManager.js
index 6839b0b6f..2db8aaa56 100644
--- a/src/scripts/inputManager.js
+++ b/src/scripts/inputManager.js
@@ -1,4 +1,4 @@
-define(['playbackManager', 'focusManager', 'appRouter', 'dom'], function (playbackManager, focusManager, appRouter, dom) {
+define(['playbackManager', 'focusManager', 'appRouter', 'dom', 'apphost'], function (playbackManager, focusManager, appRouter, dom, appHost) {
     'use strict';
 
     var lastInputTime = new Date().getTime();
@@ -96,7 +96,11 @@ define(['playbackManager', 'focusManager', 'appRouter', 'dom'], function (playba
                 appRouter.showSettings();
                 break;
             case 'back':
-                appRouter.back();
+                if (appRouter.canGoBack()) {
+                    appRouter.back();
+                } else if (appHost.supports('exit')) {
+                    appHost.exit();
+                }
                 break;
             case 'forward':
                 break;
```
</details>

The question is, should I add all this into this PR? Or made the separate one?

JustAMan:
I would say create a separate one...

dmitrylyzo:
> Also maybe ask about quitting in WebOS version as well as in Tizen?.. if that is not forbidden by LG UX guidelines...

In opposite to Tizen, `webOS.platformBack` does not close app - it shows home menu over app. So, I think there is no need to ask user about quitting.

MrTimscampi:
> @MrTimscampi is this not something that webpack could fix?

In what sense ? I don't really see anything here that bundling, minifying and transpiling could help, and the polyfills provided by Babel are more for Javascript API polyfilling.

I didn't read the code in detail, but this all seem fairly sensible as a way to handle weird platforms.

src/components/keyboardnavigation.js
```diff
@@ -1,36 +1,124 @@
-define(['inputManager', 'focusManager'], function(inputManager, focusManager) {
-    'use strict';
+define(["inputManager", "layoutManager"], function (inputManager, layoutManager) {
+    "use strict";
```
dkanada: We're migrating to single quote as the default.

src/controllers/playback/videoosd.js
```diff
@@ -1088,18 +1088,33 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
         var NavigationKeys = ["ArrowLeft", "ArrowRight", "ArrowUp", "ArrowDown"];
 
         function onWindowKeyDown(e) {
```
dkanada: Have you checked for any other usages of `onWindowKeyDown` for other purposes?

src/controllers/playback/videoosd.js
```diff
@@ -1088,18 +1088,33 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
...
                 showOsd();
                 return;
             }
 
-            switch (e.key) {
+            switch (key) {
+                case "Enter":
+                    showOsd();
+                    break;
+
```
dkanada: don't need a newline after breaks

src/scripts/inputManager.js
```diff
@@ -22,9 +22,7 @@ define(['playbackManager', 'focusManager', 'appRouter', 'dom'], function (playba
 
     var eventListenerCount = 0;
     function on(scope, fn) {
-        if (eventListenerCount) {
-            eventListenerCount++;
-        }
+        eventListenerCount++;
```
dkanada: I seem to recall this being the cause of a bug at some point.

src/components/keyboardnavigation.js
```diff
@@ -1,36 +1,124 @@
...
+        40: "ArrowDown",
+        412: "MediaRewind", // MediaRewind (Tizen/WebOS)
+        413: "MediaStop", // MediaStop (Tizen/WebOS)
+        415: "MediaPlay", // MediaPlay (Tizen/WebOS)
+        417: "MediaFastForward", // MediaFastForward (Tizen/WebOS)
+        461: "Back", // Back (WebOS)
+        10009: "Back", // Back (Tizen)
+        10232: "MediaTrackPrevious", // MediaTrackPrevious (Tizen)
+        10233: "MediaTrackNext", // MediaTrackNext (Tizen)
+        10252: "MediaPlayPause" // MediaPlayPause (Tizen)
```
dkanada: Maybe put the comments on the line above each value so we can expand on them if we need to.

src/scripts/inputManager.js
dmitrylyzo: With condition, subscription is useless - `eventListenerCount` is always "0".
I think the goal of the counter is not to spam events if there are no listeners.
https://github.com/jellyfin/jellyfin-web/blob/ec6ce5aa14384f51ff4f97a749bb4e0985aed773/src/scripts/inputManager.js#L63-L77
Possible flaws:
* subscribers do not unsubscribe
* subscribers unsubscribe more than they subscribe
* some subscriber mistakenly cancels event

src/controllers/playback/videoosd.js
dmitrylyzo: In Chrome most of cases (keyboard and one pair of gamepad) works. "Left"/"Right" and other gamepad alternatives are remain untested. But `getKeyName` returns the same `e.key` for buttons that are not listed in the array, and `onWindowKeyDown` for them remained the same.

I did not check Firefox. Now I checked, keyboard works, except for `Space`. It looks like the videoplayer receives a `click` from `Space` and pauses. Then videoosd catches `Space` and unpauses. Tried on clean JF - same behavior.
https://github.com/jellyfin/jellyfin-web/blob/ec6ce5aa14384f51ff4f97a749bb4e0985aed773/src/controllers/playback/videoosd.js#L1093-L1097

src/controllers/playback/videoosd.js
dmitrylyzo: What about empty lines below? Also remove?
https://github.com/jellyfin/jellyfin-web/blob/ec6ce5aa14384f51ff4f97a749bb4e0985aed773/src/controllers/playback/videoosd.js#L1118-L1142
and so on

dkanada: sure

src/controllers/playback/videoosd.js
dkanada: Can you add a browser check there so we can stop the behavior then? Maybe with a comment explaining why.

dmitrylyzo: I looked again and found a way to repeat similar to this in Chrome.
* We at the state when OSD is hidden.
* Press and hold `Space`.
* `videoosd` pauses player and shows OSD.
* Focus moves to `Play/Pause` button.
* Release `Space` - this generates `click` on `Play/Pause` button, and playback resumes.

It seems that in Firefox, hidden button still has focus (CSS styles `:focus` while hidden). That's why it is more "sensitive", I think.

I tried to add `document.activeElement.blur()` in `hideMainOsdControl` - Firefox started working similarly to Chrome (I notice how the focus goes away). But an issue with `Space` holding still remains in both.

dmitrylyzo: Previous solution was still unstable.
<details>
<summary>The new one</summary>

```diff
diff --git a/src/controllers/playback/videoosd.js b/src/controllers/playback/videoosd.js
index 6392c3a1a..f192abbb5 100644
--- a/src/controllers/playback/videoosd.js
+++ b/src/controllers/playback/videoosd.js
@@ -437,6 +437,11 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
                 });
                 currentVisibleMenu = null;
                 toggleSubtitleSync("hide");
+
+                // Firefox does not blur by itself
+                if (document.activeElement) {
+                    document.activeElement.blur();
+                }
             }
         }
 
@@ -1087,7 +1092,15 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
          */
         var NavigationKeys = ["ArrowLeft", "ArrowRight", "ArrowUp", "ArrowDown"];
 
+        /**
+         * Clicked element.
+         * To skip `click` handling on Firefox/Edge.
+         */
+        var clickedElement;
+
         function onWindowKeyDown(e) {
+            clickedElement = e.srcElement;
+
             var key = keyboardnavigation.getKeyName(e);
 
             if (!currentVisibleMenu && 32 === e.keyCode) {
@@ -1159,6 +1172,14 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
             }
         }
 
+        function onWindowMouseDown(e) {
+            clickedElement = e.srcElement;
+        }
+
+        function onWindowTouchStart(e) {
+            clickedElement = e.srcElement;
+        }
+
         function getImgUrl(item, chapter, index, maxWidth, apiClient) {
             if (chapter.ImageTag) {
                 return apiClient.getScaledImageUrl(item.Id, {
@@ -1289,6 +1310,12 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
                 dom.addEventListener(window, "keydown", onWindowKeyDown, {
                     capture: true
                 });
+                dom.addEventListener(window, window.PointerEvent ? "pointerdown" : "mousedown", onWindowMouseDown, {
+                    passive: true
+                });
+                dom.addEventListener(window, "touchstart", onWindowTouchStart, {
+                    passive: true
+                });
             } catch (e) {
                 require(['appRouter'], function(appRouter) {
                     appRouter.showDirect('/');
@@ -1303,6 +1330,12 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
             dom.removeEventListener(window, "keydown", onWindowKeyDown, {
                 capture: true
             });
+            dom.removeEventListener(window, window.PointerEvent ? "pointerdown" : "mousedown", onWindowMouseDown, {
+                passive: true
+            });
+            dom.removeEventListener(window, "touchstart", onWindowTouchStart, {
+                passive: true
+            });
             stopOsdHideTimer();
             headerElement.classList.remove("osdHeader");
             headerElement.classList.remove("osdHeader-hidden");
@@ -1472,7 +1505,9 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
             playbackManager.previousTrack(currentPlayer);
         });
         view.querySelector(".btnPause").addEventListener("click", function () {
-            playbackManager.playPause(currentPlayer);
+            if (this.contains(clickedElement)) {
+                playbackManager.playPause(currentPlayer);
+            }
         });
         view.querySelector(".btnNextTrack").addEventListener("click", function () {
             playbackManager.nextTrack(currentPlayer);
```
</details>

The idea is to skip `click` if keydown/mousedown/pointerdown/touchstart was on different element.

dkanada: That looks good to me, can you apply it here?

Issue #696, opened 2020-01-13
[Bug] [Regression] Android client suggestion tile swipe broken


samuel9554:
**Describe the bug**
In the suggestion tab of the android client swipe from left to right will only work on the first row. The next row while cause a tab change.
Here's a video for easy understanding.
[https://www.youtube.com/watch?v=9Jd1pupS47U](https://www.youtube.com/watch?v=9Jd1pupS47U)

**To Reproduce**
1. Open the android client
2. Click on your movie library
3. Go to tab suggestion
4. Try to swipe left or right some of the row

**Expected behavior**
The row should scroll to show the content

**System (please complete the following information):**
 - OS: [e.g. Docker]
 - Browser: [Android Client]
 - Jellyfin Version: [10.4.3]
 - Reverse proxy: [nginx]

JustAMan:
Definitely a web issue so moved, not sure if it's a regression on our part or a change in browsers tho.

MrTimscampi:
Closing, as the fix is merged into master.

PR #704, opened 2020-01-17
Fix scroll for movie recommendation

grafixeyehero:
**Issues**
[Bug] [Regression] Android client suggestion tile swipe broken #696

### DIFF
--- a/src/controllers/movies/moviesrecommended.js
+++ b/src/controllers/movies/moviesrecommended.js
@@ -124,7 +124,10 @@ define(["events", "layoutManager", "inputManager", "userSettings", "libraryMenu"
             shape: getPortraitShape(),
             scalable: true,
             overlayPlayButton: true,
-            allowBottomPadding: allowBottomPadding
+            allowBottomPadding: allowBottomPadding,
+            showTitle: true,
+            showYear: true,
+            centerText: true
         });
 
         if (enableScrollX()) {
--- a/src/controllers/movies/moviesrecommended.js
+++ b/src/controllers/movies/moviesrecommended.js
@@ -133,7 +133,7 @@ define(["events", "layoutManager", "inputManager", "userSettings", "libraryMenu"
         if (enableScrollX()) {
             html += '</div>';
         }
-
+        html += "</div>";
         html += "</div>";
         return html;
     }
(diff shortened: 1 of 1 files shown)


grafixeyehero:
focuse tv layout
f
swipe mobile layout
scroll

samuel9554:
@grafixeyehero Tested on android client and it work.

Would you considare adding those 3 line after line 127 to show title and year in the recommendation.

`			showTitle: true,
            showYear: true,
            centerText: true
`

Screenshot_20200120-124232_Jellyfin

grafixeyehero:
> 
> 
> @grafixeyehero Tested on android client and it work.
> 
> Would you considare adding those 3 line after line 127 to show title and year in the recommendation.
> 
> ` showTitle: true, showYear: true, centerText: true `
> 
added https://github.com/jellyfin/jellyfin-web/pull/704/commits/779e3782884da0dcfe11b43c7e7c597abe559a59

src/controllers/movies/moviesrecommended.js
```diff
@@ -114,19 +114,26 @@ define(["events", "layoutManager", "inputManager", "userSettings", "libraryMenu"
...
             shape: getPortraitShape(),
             scalable: true,
             overlayPlayButton: true,
-            allowBottomPadding: allowBottomPadding
+            allowBottomPadding: allowBottomPadding,
+            showTitle: true,
+            showYear: true,
+            centerText: true
         });
-        html += "</div>";
```
dkanada: I think this was used to close the `verticalSection` div element on line 112, so we need to keep it.

grafixeyehero: > 
> 
> I think this was used to close the `verticalSection` div element on line 112, so we need to keep it.

the verticalSection was closed at line 137

dkanada: Doesn't 134 close the `itemsContainer` and 137 close the scroller?

src/controllers/movies/moviesrecommended.js
```diff
@@ -114,18 +114,25 @@ define(["events", "layoutManager", "inputManager", "userSettings", "libraryMenu"
...
             overlayPlayButton: true,
-            allowBottomPadding: allowBottomPadding
+            allowBottomPadding: allowBottomPadding,
+            showTitle: true,
+            showYear: true,
+            centerText: true
         });
+
+        if (enableScrollX()) {
+            html += '</div>';
```
samuel9554: I think you should close it inside this if because if you open 2 div you need to close 2 div. Else we only open one div.

dkanada: We actually need to close either two or three divs in this method, so this seems correct to me. A few lines above there is a third div element.

PR #705, opened 2020-01-18
Fix scroll for movie recommendation #704

samuel9554:
#704 This changed worked for me. I tested it on multiple platform without a issue.

Fix for #696

### DIFF
--- a/src/controllers/movies/moviesrecommended.js
+++ b/src/controllers/movies/moviesrecommended.js
@@ -110,9 +110,9 @@ define(["events", "layoutManager", "inputManager", "userSettings", "libraryMenu"
         }
 
         html += '<div class="verticalSection">';
-		html += '<div class="sectionTitleContainer sectionTitleContainer-cards">';
+        html += '<div class="sectionTitleContainer sectionTitleContainer-cards">';
         html += '<h2 class="sectionTitle sectionTitle-cards padded-left">' + title + "</h2>";
-		html += "</div>";
+        html += "</div>";
         var allowBottomPadding = true;
 
         if (enableScrollX()) {
@@ -128,7 +128,7 @@ define(["events", "layoutManager", "inputManager", "userSettings", "libraryMenu"
             allowBottomPadding: allowBottomPadding
         });
 
-		html += "</div>";
+        html += "</div>";
 
         html += "</div>";
         return html;


samuel9554:
> 
> 
> please fix the lint errors reported by azure :)

Done.

grafixeyehero:
@ samuel9554 thanks for your support. their was an open PR https://github.com/jellyfin/jellyfin-web/pull/704 that was fix this may u can cheack it

samuel9554:
> 
> 
> @samuel9554 thanks for your support. their was an open PR #704 that was fix this may u can cheack it

@grafixeyehero I did check you PR. But there's in my opinion there's a bug with it. At line 117 and 118 you open 2 DIV if condition is true. But you only close it one time. Causing the issue in my case when testing. So that's why I create this PR. I test it on the android apps, web... and for me it work fine. Would you considerate modify your pull or removing it?
Also at line 113 in my pull I added a missing div.

grafixeyehero:
> 
> 
> > @samuel9554 thanks for your support. their was an open PR #704 that was fix this may u can cheack it
> 
> @grafixeyehero I did check you PR. But there's in my opinion there's a bug with it. At line 117 and 118 you open 2 DIV if condition is true. But you only close it one time. Causing the issue in my case when testing. So that's why I create this PR. I test it on the android apps, web... and for me it work fine. Would you considerate modify your pull or removing it?
> Also at line 113 in my pull I added a missing div.

i did't see any issue here and for the div it will be colsed automaticaly

samuel9554:
> 
> 
> > > @samuel9554 thanks for your support. their was an open PR #704 that was fix this may u can cheack it
> > 
> > 
> > @grafixeyehero I did check you PR. But there's in my opinion there's a bug with it. At line 117 and 118 you open 2 DIV if condition is true. But you only close it one time. Causing the issue in my case when testing. So that's why I create this PR. I test it on the android apps, web... and for me it work fine. Would you considerate modify your pull or removing it?
> > Also at line 113 in my pull I added a missing div.
> 
> i didn't see any issue here and for the div it will be closed automatically

You have test on android apk? Because I really see a issue with the div. All the verticalSection div open inside the last one. I could send you a pic later today.

samuel9554:
Fix by pr #704

PR #707, opened 2020-01-18
Fix playedIndicator color; close #706

### LINKED ISSUE #706: Visually distinguish between fully watched series and series with unwatched episodes.
Right now, fully watched series have a blue check mark and series with new episodes have a blue number:

image

I keep getting that small moment of excitement thinking there is a new episode, and upon closer inspection it's just the same blue circle with a check mark. I've ignored this for a while but I can't get used to it. I guess there are two groups of people: Those sensitive to textual communication (number versus check mark) and those sensitive to visual communication for lack of a better word (like red versus green in traffic lights).

Can we make the circle with a checkbox have a different color, like green?

image

Redsandro:
**Changes**
Change `playedIndicator` color to green as it was colored ambiguously similar to `countIndicator`.

**Issues**
Fixes #706

### DIFF
--- a/src/themes/appletv/theme.css
+++ b/src/themes/appletv/theme.css
@@ -292,9 +292,12 @@ html {
 }
 
 .countIndicator,
-.fullSyncIndicator,
-.playedIndicator {
+.fullSyncIndicator {
     background: #00a4dc
+}
+
+.playedIndicator {
+    background: #5dd000
 }
 
 .fullSyncIndicator {
--- a/src/themes/blueradiance/theme.css
+++ b/src/themes/blueradiance/theme.css
@@ -295,9 +295,12 @@ html {
 }
 
 .countIndicator,
-.fullSyncIndicator,
-.playedIndicator {
+.fullSyncIndicator {
     background: #00a4dc
+}
+
+.playedIndicator {
+    background: #5dd000
 }
 
 .fullSyncIndicator {
--- a/src/themes/dark/theme.css
+++ b/src/themes/dark/theme.css
@@ -276,9 +276,12 @@ html {
 }
 
 .countIndicator,
-.fullSyncIndicator,
-.playedIndicator {
+.fullSyncIndicator {
     background: #00a4dc
+}
+
+.playedIndicator {
+    background: #5dd000
 }
 
 .fullSyncIndicator {
--- a/src/themes/light/theme.css
+++ b/src/themes/light/theme.css
@@ -293,9 +293,12 @@ html {
 }
 
 .countIndicator,
-.fullSyncIndicator,
-.playedIndicator {
+.fullSyncIndicator {
     background: #00a4dc
+}
+
+.playedIndicator {
+    background: #5dd000
 }
 
 .fullSyncIndicator {
--- a/src/themes/wmc/theme.css
+++ b/src/themes/wmc/theme.css
@@ -273,9 +273,12 @@ html {
 }
 
 .countIndicator,
-.fullSyncIndicator,
-.playedIndicator {
+.fullSyncIndicator {
     background: #00a4dc
+}
+
+.playedIndicator {
+    background: #5dd000
 }
 
 .fullSyncIndicator {


Redsandro:
image

Redsandro:
@nvllsvm

nvllsvm:
The green looks out of place with the rest of the default theme. Inverting the number indicator might be better.

dkanada:
We could also just remove it entirely for completed shows.

Redsandro:
@dkanada commented:
> We could also just remove it entirely for completed shows.

Please don't! It's convenient to easily distinguish between unwatched shows and completed shows.

@nvllsvm commented:
> The green looks out of place with the rest of the default theme. Inverting the number indicator might be better.

I like it like this. The saturation is the same as the blue one. I'm happily using it. Although I can see it clearly on my monitor, we don't really need to read what's in the green circle, because there will only ever be a check-mark. The color communicates it's meaning.

Feel free to adjust it to your liking. This will surely benefit others.
image

JustAMan:
Maybe we can keep it coloured the same but change its shape, like make it a square with a check?..

grafixeyehero:
maybe like this
Annotation 2020-01-22 1850027

Redsandro:
@JustAMan I don't like it. The shape difference is too subtle. A letter is a shape. This PR is about communicating without using letters, but something more easily distinguishable, which includes looking towards the TV while not wearing glasses.

@grafixeyehero I prefer the color, but if the community prefers this, as long as it's that radically different, I approve.

@dkanada about the label; be advised that I am running with this small change and it makes it so much better. This PR commits that change back to the community in the hope that it will make others happy, too. I'm swamped at work so I don't have time to discuss back and forth and make different versions. Options are:

- Take it, merge it now, possibly improve later
- Leave it, let me run this version on my own
- Adjust it to your liking yourself (I've given maintainers access), then take it
- Let someone else create a new PR that obsoletes this one.

sparky8251:
While these changes will impact all themes, it would be nice for the images showcasing the changes to be done only in the default theme.

The theme in this image is purple haze, and its radically different looking compared to the default dark theme.

For drive by viewers and commenters (who we want on UI changes), we should keep the potential for confusion to a minimum.

Can you please post these images exclusively from the perspective of the default dark theme from here on? Thanks!

dkanada:
@Redsandro we're not implying you have to change this pull request at all, we are just considering different options for the icon design.

Redsandro:
@dkanada sorry, I might have gotten the wrong idea because this is being discussed in the PR review rather than in #706.

MrTimscampi:
So, a bit of an accessibility pass on this issue (Since I'm directly concerned :p)

A good color difference is always better. "Subtle" changes in shape like the proposal from @JustAMan can lead to confusion when someone with bad eyesight is looking for something but skimming the items.

That said, I find @grafixeyehero's change not consistent with the rest of the layout, visually. It's also harder to recognize numbers at a glance when they're rotated like that.

The proposal by @Redsandro is good, but it may need a slight colour change (Maybe a bit darker and slightly less saturated ?)

Generally, I'm not a fan of the lack of contrast between the blue circles and the white text inside of it either, but that's another issue.

Artiume:
My concern for making green the default is for those who are color blind (deuteranomalous). I think a better idea is to introduce community css for allowing customization and provide the css templates on the JF website.

Redsandro:
> My concern for making green the default is for those who are color blind (deuteranomalous). I think a better idea is to introduce community css for allowing customization and provide the css templates on the JF website.

I recommend against over-engineering this.

About 4.5% of the population is colorblind. The majority of them are red-green colorblind. There is no red or green in blue, so these colors are compatible as can be seen in a red-green-colorblindness emulator:

image

Besides, it's not like the fraction of people that would be affected would _lose_ something because of this. For those people that aren't sensitive to color communication, they can still see the checkmark.

I think the simplest version, the initial PR, is an improvement compared to not having the distinction at all. User customization for theme elements would be a different feature.

Remember, perfect is the enemy of good.

Artiume:
Just giving my 2 cents. Over engineering it would be to build in the customization directly into JF. Having freely available css scripts isn't exactly "over engineering" lol

Artiume:
https://www.toptal.com/designers/colorfilter/

Testing out the green image, the only one that was really impacted are those with tritanopia, which just leaves the green to be basically a light blue. 

@Redsandro you're coming off a little salty about concerns people may have, if we can't have open discussion and ideas, what's the point of discussing things? We all want the same thing in the end, right?

MrTimscampi:
> My concern for making green the default is for those who are color blind (deuteranomalous). I think a better idea is to introduce community css for allowing customization and provide the css templates on the JF website.

Indeed, that's very valid. Perhaps we can combine shape and color to cater to both disabilities ?

DMouse10462:
Just to throw some other ideas into the pot, we could take a page from Sonarr's Series page and use a bottom border (maybe with countIndicator popping up from it with the number of unwatched episodes) that is color coded (with some friendly striping to assist visually) to distinguish between the two. 

In the case of a partially watched series, the bottom border could act similarly to the "Continue Watching" bar that shows progress in an episode, but with progress over a whole series. This provides (1) much more information at a glance and (2) an additional contrast between the partially watched and completed series (partial border vs. complete border).

Here are some examples (ripped straight from Sonarr, sometimes with my own ugly edits):

### Partially Watched Series (countIndicator in bottom-left):
countIndicator
Note the striped progress bar that contrasts with the solid bar for the watched series below. 
With this implementation, there are multiple points of distinction:
- Color (blue for in-progress) with optional (user-configurable like Sonarr?) stripes for color-impaired users
- Incomplete border (like "Continue Watching" on the home page)
- Numeric indicator tag above bar

### Watched Series (Ignore the red flag in the upper-right):
playedIndicator
Note the lack of a playedIndicator; the colored border at the bottom would identify a completed series.
With the above bar for incomplete series, this may not even be necessary and we could keep the checkmark.

A completely unwatched series would probably remain unchanged (no lower border).

nvllsvm:
Lets continue discussion in the issue itself as there is no agreement on the changes here. We can reopen if we change our mind.

Redsandro:
> @Redsandro you're coming off a little salty

@Artiume being salty means _"feeling agitated, feeling out of place or feeling attacked",_ right?

I might be a bit agitated (as in nervous) because this gives me flashbacks of projects where I commit something small, and then it is discussed, scope widened, complexity added for 6 to 9 months until everyone loses interest and the change is dangling, never committed.

I'm new to Jellyfin and I don't know the process for something as simple as a color change, so yes feel out of place too and I should sit back and see what happens before making assumptions. 

That said I'm very very very much in favor of projects with CI, i.e. merging something good now and _then_ discuss how it can be perfected, rather than do nothing now and discuss how it can be perfected.

> as someone who is visually impaired, I actually take offense

You can't be serious @MrTimscampi. While it shouldn't be relevant, __I'm visually impaired too. It was the main motivation for this PR.__ I've had multiple eye surgeries. I __had given the choice of colors thought__ before requesting a pull. I'm taking something that hasn't been given thought, and improve it greatly for visually impaired people. I recommend against over-engineering _this initial PR which we're discussing._ Once a first iteration is available for all, _it can be gradually improved_ based on a larger pool of feedback. _Feel free to submit your own PR._

Yes I guess I'm salty.

nvllsvm:
@Redsandro I truly appreciate the time and energy you're putting into this change - thank you.

Jellyfin already has custom CSS configuration in  **Server -> General -> Branding**. You just have to use `!important` until https://github.com/jellyfin/jellyfin-web/issues/742 is resolved.

I've confirmed this works:
```
.playedIndicator { background: #5dd000 !important; }
```

image

Artiume:
So I made a quick draft page so we can have community scripts. i'm running the green patch now :). Not sure which color I'll eventually land on haha.

Check it out
https://github.com/jellyfin/jellyfin-docs/pull/215

Redsandro:
@nvllsvm 

> I truly appreciate the time and energy you're putting into this change

:thinking: 

> You just have to use !important until #742 is resolved. I've confirmed this works:
> ```
> .playedIndicator { background: #5dd000 !important; }
> ```

I hadn't thought about that. It works this way. :+1: 

> this color change negatively impacts the viewing experience for me. Ideally, these indicators should not be placed over the poster to avoid risk of blending into a similarly covered poster.

Fair enough. _Extra_ indicators, while staying within the same color and being chromatically peaceful, add visual load, too. When visually impaired because of severe astigmatism or strabismus, adding extra elements will cause visual overload, especially when contrasty (e.g. extra indicator on the black background rather than the thumbnail).

I can use these CSS rules and I'm happy about them. Problem solved, for me. As for the rest of the users however, some people don't know what CSS is. 

It's probably fair to say that some people prefer a simple change like the one proposed, and others find that it messes with the zen of the layout. So while I was thinking about this community scripts by @Artiume on the one hand, and the usability for everyone (AKA non-tech people) on the other hand, perhaps one interesting idea would be to have "skin options".

Somewhere in the settings for the skin, there can be some checkboxes that enable certain CSS rules. In this particular example:

:ballot_box_with_check: Turn fully played media indicator green.

(Or some other consensus-driven simple alternative that suits the average needs for a global cross-section of users, where more complex solutions can be added through custom CSS.)

Just a thought. Writing CSS on the settings works too, as long as you `!important` everything.

Artiume:
I like your idea, how hard do you think it'd be to make skins? My current idea is to make the community css page, providing visual examples and walk throughs on how to add the code inside the webui. This'll allow us to add additional customization with low maintenance on our behalf. 

Discussing it with the team, there's a few gotchas we have to watch out for. The css scripts may behave differently between the various apps, doing something such as checking the screen width could help provide filtering so that different versions of the same script can exist for each client type.

Doing some quick searching of community css already available on the interwebs, there looks to be plenty available to test out :)

Redsandro:
> So I made a quick draft page so we can have community scripts. i'm running the green patch now :). Not sure which color I'll eventually land on haha.
> 
> Check it out
> [jellyfin/jellyfin-docs#215](https://github.com/jellyfin/jellyfin-docs/pull/215)

Turned out really nice @Artiume 
https://jellyfin.org/docs/general/clients/css-customization.html

Do you know if there are ways to do something similar on the Android TV client?

thornbill:
The next release of the Android TV client will include a theme option in the app settings. These are managed in a resource file in the app: https://github.com/jellyfin/jellyfin-androidtv/blob/master/app/src/main/res/values/themes.xml. There isn't currently a way of modifying the colors via a server setting.

PR #708, opened 2020-01-19
Move material icons to external package

MrTimscampi:
**Changes**

Removes material icons from the source and use an external package to handle them instead.

I opted to use [material-design-icons-iconfont](https://www.npmjs.com/package/material-design-icons-iconfont) since it's a newer version of Google's official package, with icons that are in the repository but not packaged by Google for some reason, faster checkout (due to not including thousand of source files that we won't use) and supports SCSS/SASS for when we eventually move to that.

This PR basically only adds the package via yarn, defines it in bundle.js then replaces all instances of md-icon with material-icons in the source.

**Issues**

None


src/components/medialibrarycreator/medialibrarycreator.js
```diff
@@ -95,7 +95,7 @@ define(["loading", "dialogHelper", "dom", "jQuery", "components/libraryoptionsed
...
-        return html += '<div class="listItem listItem-border lnkPath" style="padding-left:.5em;">', html += '<div class="' + (pathInfo.NetworkPath ? "listItemBody two-line" : "listItemBody") + '">', html += '<div class="listItemBodyText">' + pathInfo.Path + "</div>", pathInfo.NetworkPath && (html += '<div class="listItemBodyText secondary">' + pathInfo.NetworkPath + "</div>"), html += "</div>", html += '<button type="button" is="paper-icon-button-light"" class="listItemButton btnRemovePath" data-index="' + index + '"><i class="md-icon">remove_circle</i></button>', html += "</div>"
+        return html += '<div class="listItem listItem-border lnkPath" style="padding-left:.5em;">', html += '<div class="' + (pathInfo.NetworkPath ? "listItemBody two-line" : "listItemBody") + '">', html += '<div class="listItemBodyText">' + pathInfo.Path + "</div>", pathInfo.NetworkPath && (html += '<div class="listItemBodyText secondary">' + pathInfo.NetworkPath + "</div>"), html += "</div>", html += '<button type="button" is="paper-icon-button-light"" class="listItemButton btnRemovePath" data-index="' + index + '"><i class="material-icons">remove_circle</i></button>', html += "</div>"
```
JustAMan: seems we missed de-uglifying this one

yarn.lock
```diff
@@ -3039,6 +3039,11 @@ map-visit@^1.0.0:
   dependencies:
     object-visit "^1.0.0"
 
+material-design-icons-iconfont@^5.0.1:
```
JustAMan: I presume this thing would pull in the font, right? Where would it activate the font?

src/components/medialibrarycreator/medialibrarycreator.js
dkanada: still haunting us to this day

MrTimscampi: Ouch, I'll give that a look once I'm home. Let's take the opportunity to clean this monstrosity.

yarn.lock
MrTimscampi: It pulls in the font, then gets imported in the bundle like hls.js and other external modules.

The current Webpack process will copy it to the dist directory.

src/components/medialibrarycreator/medialibrarycreator.js
dkanada: @grafixeyehero was doing more deminification so he probably included this as well.

ferferga:
I don't understand very well this PR and I'm not by any means a JS expert, so probably I'm missing something, but what's the problem with the actual approach with CSS? In my opinion, doing this with a npm package is overcomplicating such a simple thing.

We can't just update the woff files?

MrTimscampi:
It's generally good practice to not have outside dependencies in the source if they're unmodified.

For this, there is a maintained, widely used (50k+ downloads per week) package available on NPM. So it makes sense to move it out of the tree.

This also allows us to update it easier. Either we can use Yarn to check for updates, or we can, if we wish to at some point, setup some bot to notify when a dependency is updated.

Once we move to SASS for the stylesheets, we don't even need to add it as a JS module anymore, since we'll be able to import it directly from our main stylesheet.

It's the same as for Bootstrap, for example. You usually wouldn't add all the bootstrap files to your code. You'd add it as a dependency via NPM, then import it.

PR #710, opened 2020-01-20
Fix some issues with theme resources

dkanada:


src/components/skinManager.js
```diff
@@ -34,15 +34,12 @@ define(['apphost', 'userSettings', 'browser', 'events', 'pluginManager', 'backdr
             id: "dark",
             isDefault: true,
             isDefaultServerDashboard: true
-        }, {
-            name: "Emby",
-            id: "emby"
```
JustAMan: why not rename to `emerald`?

dkanada: This theme is almost exactly the same as the dark theme and maintaining them both will be a pain until we have better theme support. Working on custom themes is a better alternative to leaving this in the web source and having to deal with it.

PR #715, opened 2020-01-20
Remove libjass

MrTimscampi:
**Changes**

Removes libjass since it's not supported anymore and has been replaced by JavascriptSubtitleOctopus anyway, which works with pretty much everything we target.

**Issues**

None.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,6 @@
     "jquery": "^3.4.1",
     "jstree": "^3.3.7",
     "libass-wasm": "https://github.com/jellyfin/JavascriptSubtitlesOctopus",
-    "libjass": "^0.11.0",
     "material-design-icons-iconfont": "^5.0.1",
     "native-promise-only": "^0.8.0-a",
     "page": "^1.11.5",
--- a/src/bundle.js
+++ b/src/bundle.js
@@ -84,13 +84,6 @@ _define("webcomponents", function() {
     return webcomponents;
 });
 
-// libjass
-var libjass = require("libjass");
-require("libjass/libjass.css");
-_define("libjass", function() {
-    return libjass;
-});
-
 // libass-wasm
 var libass_wasm = require("libass-wasm");
 _define("JavascriptSubtitlesOctopus", function() {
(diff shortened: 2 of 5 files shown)


Artiume:
I could check out the Ps4, just let me know what you need

MrTimscampi:
> I could check out the Ps4, just let me know what you need

We need to check if playing a file with .ass/.ssa subs work with this PR (It removes the non-Web Assembly library that is no longer supported by its developer).  
You'd likely need @JustAMan 's fix of SubtitleOctopus though, since I doubt the PS4 run Web Assembly and thus needs the fallback to asm.js.

Artiume:
Ok, I'll pull in the PR. This will be using the built in browser on the Ps4, right?

dkanada:
@Artiume any updates here?

Artiume:
I can do it tomorrow, I had forgotten about it 😅

Artiume:
So I'm trying to verify functionality but I never use my ps4 browser. I'm still on normal nightly, not the PR yet. I can't seem to click on any media on the home page. I can't tell if it's me not knowing how to properly navigate the Ps4 ui or if it's just being buggy

Artiume:
So I'm not sure if it's the Ps4 being slow or maybe an issue with the webui. If I click on anything, I can't change pages. I can see the url attempts to update and I get the waiting icon like it's trying to load the new page but nothing occurs. If I exit the browser and reenter the browser, I'll be on the new page.

I feel like this is a new issue. I did try the Ps4 about a month or so ago and i was able to play content.

Artiume:
Unable to verify functionality on branch either. Same sort of issue, refreshing the browser is required to move to a new page but I am unable use that trick to get media to play.

MrTimscampi:
@JustAMan @dkanada To bring this back to life, according to [this blog post](http://blog.greggant.com/posts/2016/04/07/exploring-and-developing-for-the-ps4-browser.html), the PS4 supports Canvas and Web Workers, which is the only check performed right now for whether to enable Octopus or not.

In my opinion, we should remove libjass and, if needed for performance or compatibility reasons, force transcoding for the PS4 [here](https://github.com/jellyfin/jellyfin-web/blob/master/src/components/apphost.js#L52). The reasoning being that it's better to render fully accurate subtitles by transcoding, rather than falling back on an unmaintained library that has known rendering issues.

MrTimscampi:
> Why so many lines that (seemingly) are unrelated to libjass are removed from lockfile? @MrTimscampi

Due to changes in SubOctopus, I had to regenerate a full lockfile after removing the yarn cache. This one is the result of a fresh `yarn install`, basically.

Issue #703, opened 2020-01-17
Update BrowserDeviceProfile.js


anthonylavado:
**Describe the bug**
The copy of BrowserDeviceProfile.js we have is largely untouched, with only minor changes since the fork. As a result, there are many supported codecs/formats that are marked as unsupported for certain browsers.

**To fix**
Review each platform/browser that BrowserDeviceProfile can identify, and ensure that the correct capabilities are reported back.

**Additional context**
Resources to use:
* https://caniuse.com/
* https://jellyfin.org/docs/general/clients/codec-support.html

the-r0cket06:
I can help anyone that is planning on tackling this. I started to poke around a little bit today, and might have some stuff ready for a PR in the coming days.

JustAMan:
@the-r0cket06 Go ahead, and make a draft PR when you have at least something, so others know your approximate progress (and the fact that you're working on it).

timo95:
Related:
https://github.com/jellyfin/jellyfin-expo/issues/106

thornbill:
This file received regular updates now 🙃

PR #716, opened 2020-01-20
Safari supports AC-3

Bond-009:
#703

### DIFF
--- a/src/scripts/browserdeviceprofile.js
+++ b/src/scripts/browserdeviceprofile.js
@@ -77,10 +77,6 @@ define(['browser'], function (browser) {
             return true;
         }
 
-        if (browser.osx || browser.iOS) {
-            return false;
-        }
-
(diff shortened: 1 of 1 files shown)


anthonylavado:
So I started on this (haven’t made the PR yet), but I stopped because I needed to figure something out. Does removing this check mean that it will test if support is available, and thus use it?

Are the checks just hard coded supported states?

anthonylavado:
Okay, so the statement below does query correctly for support.

Want to remove the restriction on eac3 @bond-009? I’ve tested that it works as well, on both desktop and mobile Safari.

Bond-009:
EAC-3 doesn't have the same restriction, right?
https://github.com/jellyfin/jellyfin-web/blob/361a6acc96d1d5b0cff926ceea4ce7e9d4ee9c6a/src/scripts/browserdeviceprofile.js#L83

anthonylavado:
Yep all good ty @Bond-009

PR #717, opened 2020-01-20
Possible Fix Custom Video Controls

grafixeyehero:
Tested 
Firefox: 73.0b6 (64-bit)
Chrome: 79.0.3945.117
Microsoft Edge 44.18362.449.0

### DIFF
--- a/src/components/htmlvideoplayer/plugin.js
+++ b/src/components/htmlvideoplayer/plugin.js
@@ -1378,6 +1378,10 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
             }
         }
 
+        function enableCustomVideoControls() {
+            return true;
+        }
+
         function createMediaElement(options) {
 
             if (browser.tv || browser.iOS || browser.mobile) {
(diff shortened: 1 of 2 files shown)


MrTimscampi:
I can confirm this works on Firefox 72.0.1 (64bit) and Chrome 79.0.3945.130 (64bit).

src/components/htmlvideoplayer/plugin.js
```diff
@@ -1378,6 +1378,10 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
             }
         }
 
+        function enableCustomVideoControls() {
+            return true;
+        }
```
JustAMan: heh, I want some explanation here... you're basically always default to `else` clause, which (as I judge from comments) could break on WebOS which have old-ish Chrome bundled.

dkanada: Seems like this might be resolved by simply removing the CSS rule.

JustAMan: Those things there were for a reason, so I'm sensing that we'd need them for some edge cases...

dkanada: I specifically mean the four lines of CSS he removed, which look like they would globally hide all custom video controls on firefox.

grafixeyehero: > 
> 
> heh, I want some explanation here... you're basically always default to `else` clause, which (as I judge from comments) could break on WebOS which have old-ish Chrome bundled.

i am limited here

dkanada:
resolved in #726 for now

Issue #553, opened 2019-10-26
alphaPicker: page padding on the wrong side in TV mode


dmitrylyzo:
**Describe the bug**
In TV mode, the page with alphaPicker has padding on the opposite to the alphaPicker side. As a result, on a real TV (and in device mode of browsers) keyboard navigation jumps out from the alphaPicker to an overlaped card.

**To Reproduce**
1. Turn on the TV display mode in the settings (or activate device mode with TV UserAgent in browser).
2. Go to Music/Albums (or other pages with the alphaPicker).
3. The alphaPicker on the right, but the padding (the wider one) on the left.

**Expected behavior**
In Desktop and Mobile display modes padding is on the right (as the alphaPicker). So I suppose the same should be in TV display mode.

**Screenshots**
TV (Firefox)
![jellyfin-web_issue04_03-tv](https://user-images.githubusercontent.com/56478732/67625895-ff8acc00-f84c-11e9-8819-0dfe8686680a.png)

TV (Firefox)
![jellyfin-web_issue04_02-tv](https://user-images.githubusercontent.com/56478732/67625899-07e30700-f84d-11e9-9659-9ff64f2777bc.png)

Desktop (Firefox)
![jellyfin-web_issue04_01-desktop](https://user-images.githubusercontent.com/56478732/67625897-0580ad00-f84d-11e9-80df-1284061a3932.png)

**System**
 - OS: Kubuntu 19.04 x64
 - Browser: Firefox 70, Chrome 76.0.3809.100
 - Jellyfin Version: 10.4.z (0d8bc6fb46f0f105b34869d82d0ab826b0e78251)

JustAMan:
This is intentional (don't know why tho), see in unminify PR: https://github.com/jellyfin/jellyfin-web/pull/509/files#diff-1300564f83a1c0f9bc3221666118b2eaR219-R223

dkanada:
We should fix it regardless, not sure why it does that in the first place.

JustAMan:
Wasn't saying we shouldn't, was just giving a reference to a potential place where this thing happens.

JustAMan:
shoo bot, it should be fixed soon

dmitrylyzo:
Fixed in `master`

PR #721, opened 2020-01-21
Fix alphaPicker padding 

grafixeyehero:
**Issues**
alphaPicker: page padding on the wrong side in TV mode #553

Tv mode
TV-mode

Desktop mode
Desktop-mode

Mobile mode
Mobile-mode


dkanada:
Is there any particular design standard that requires the picker to be on the left for TV layouts? Maybe we can just remove this logic entirely.

dkanada:
Somewhat related but can you figure out how to center this vertically next to the list?

a

JustAMan:
> Maybe we can just remove this logic entirely.

Yeah, maybe we should just show picker on the right (or on the left)?
It being on the left might be because it was intended to be selectable by D-pad remote?..

JustAMan:
cc @dmitrylyzo @EraYaN for TV insights

dmitrylyzo:
Personaly, I'm used to having this on the right, because I look on page from left to right, and alphaPicker is not "distracting".

Also, if the left menu will be (_someday_) navigable with keyboard and should be on library page, then right alphaPicker is preferable.

JustAMan:
In this case let's move it to the right on all layouts I think.

grafixeyehero:
@dkanada the alphapicker it's already centered or did i miss some think

Annotation 2020-01-22 185027

EraYaN:
I think they mean the filter and sort buttons.

Also yes it is selectable with the D-Pad, although the position of it is very stupid. But the right seems like the popular choice.

dkanada:
I meant that on large portrait screens, the alpha picker isn't centered vertically next to the item container as seen above.

grafixeyehero:
it happen when you take the screen shot the whole page

src/controllers/movies/movies.js
```diff
@@ -169,11 +169,8 @@ define(["loading", "layoutManager", "userSettings", "events", "libraryBrowser",
                     valueChangeEvent: "click"
                 });
 
-                if (layoutManager.desktop || layoutManager.mobile) {
-                    alphaPickerElement.classList.add("alphabetPicker-right");
-                    itemsContainer.classList.remove("padded-left-withalphapicker");
-                    itemsContainer.classList.add("padded-right-withalphapicker");
-                }
+                alphaPickerElement.classList.add("alphaPicker-fixed-right");
+                itemsContainer.classList.add("padded-right-withalphapicker");
```
dkanada: Does this one not need the `tabContent` object used in the other files?

PR #724, opened 2020-01-23
Update and fix some canPlayType

nyanmisaka:
**Changes**
Update and fix some canPlayType in BrowserDeviceProfile.js

**Issues**
https://github.com/jellyfin/jellyfin-web/issues/703

### DIFF
--- a/src/scripts/browserdeviceprofile.js
+++ b/src/scripts/browserdeviceprofile.js
@@ -253,7 +253,7 @@ define(['browser'], function (browser) {
             case 'ts':
                 supported = testCanPlayTs();
                 videoCodecs.push('h264');
-                if (canPlayH265(videoTestElement)) {
+                if (canPlayH265(videoTestElement, options)) {
                     videoCodecs.push('h265');
                     videoCodecs.push('hevc');
                 }
@@ -477,7 +477,7 @@ define(['browser'], function (browser) {
             hlsVideoCodecs.push('h264');
         }
 
-        if (canPlayH265(videoTestElement)) {
+        if (canPlayH265(videoTestElement, options)) {
             mp4VideoCodecs.push('h265');
             mp4VideoCodecs.push('hevc');
 
--- a/src/scripts/browserdeviceprofile.js
+++ b/src/scripts/browserdeviceprofile.js
@@ -241,7 +241,7 @@ define(['browser'], function (browser) {
                 supported = browser.tizen || browser.orsay;
                 break;
             case 'mov':
-                supported = browser.tizen || browser.orsay || browser.web0s || browser.edgeUwp;
+                supported = browser.tizen || browser.orsay || browser.web0s || browser.chrome || browser.edgeUwp;
                 videoCodecs.push('h264');
                 break;
             case 'm2ts':
--- a/src/scripts/browserdeviceprofile.js
+++ b/src/scripts/browserdeviceprofile.js
@@ -180,6 +180,10 @@ define(['browser'], function (browser) {
 
             return true;
         }
+		
+        if (browser.edgeUwp) {
+            return true;
+        }
 
         return false;
     }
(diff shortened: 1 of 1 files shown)


src/scripts/browserdeviceprofile.js
```diff
@@ -18,12 +18,27 @@ define(['browser'], function (browser) {
...
-        // Unfortunately haven't yet found a canPlayType for proper detection
-        if (browser.iOS && (browser.iOSVersion || 0) >= 11) {
-            return true;
+        if (browser.ps4) {
+            return false;
+        }
+
+        if (protocol === 'hls') {
+
+            //safari seems to be lying about this
```
JustAMan: this is really strange, I would feel that the only browsers which in reality _can_ play hls with h265 are safari-based, have you tested this?

src/scripts/browserdeviceprofile.js
```diff
@@ -18,12 +18,27 @@ define(['browser'], function (browser) {
...
+            (videoTestElement.canPlayType('video/mp2t; codecs="hvc1.1.L0.0"').replace(/no/, '') ||
+            videoTestElement.canPlayType('video/mp2t; codecs="hev1.1.L0.0"').replace(/no/, '') ||
+            videoTestElement.canPlayType('video/mp2t; codecs="hev1.1.2.L150"').replace(/no/, ''));
         }
 
-        return !!(videoTestElement.canPlayType && videoTestElement.canPlayType('video/hevc; codecs="hevc, aac"').replace(/no/, ''));
+        return !!videoTestElement.canPlayType &&
+        (videoTestElement.canPlayType('video/mp4; codecs="hvc1.1.L0.0"').replace(/no/, '') ||
+        videoTestElement.canPlayType('video/mp4; codecs="hev1.1.L0.0"').replace(/no/, '') ||
+        videoTestElement.canPlayType('video/mp4; codecs="hev1.1.2.L150"').replace(/no/, ''));
```
JustAMan: 1) this seems to be a direct copy-paste of lines above - there's no need for that
2) where did you get these levels? can you explain their magic here?

src/scripts/browserdeviceprofile.js
```diff
@@ -750,8 +773,26 @@ define(['browser'], function (browser) {
         var maxH264Level = browser.chromecast ? 42 : 51;
         var h264Profiles = 'high|main|baseline|constrained baseline';
 
-        if (maxH264Level >= 51 && browser.chrome && !browser.osx) {
-            h264Profiles += '|high 10';
+        if (browser.tizen || browser.orsay || browser.web0s ||
+            videoTestElement.canPlayType('video/mp4; codecs="avc1.640833"').replace(/no/, '')) {
+            maxH264Level = 51;
```
JustAMan: it is already `51` if it's not a chromecast, and if it is this test won't probably pass... something strange is going on here

src/scripts/browserdeviceprofile.js
nyanmisaka: H265 HLS doesn't seem to work on iOS13 but it did work on my iOS12 device. It's wired so I will remove that part now.

src/scripts/browserdeviceprofile.js
nyanmisaka: `video/mp4; codecs="hev1.1.0.L120"` for new Edge Chromium and iOS.
`canPlayType('video/hevc; codecs="hevc,aac"')` output NO on new Edge and iOS.
`hvc1` and `hev1` are two different names of writing video tags.

refer:
https://cconcolato.github.io/media-mime-support/
https://chromium.googlesource.com/chromium/src.git/+/master/media/base/video_codecs_unittest.cc#450

src/scripts/browserdeviceprofile.js
nyanmisaka: I will let `maxH264Level = 42` at first.
I passed the test `canPlayType('video/mp4; codecs="avc1.640833"')` on `Chrome` `Firefox` `Safari` `Edge` `new Edge`.
This [doc](https://developers.google.com/cast/docs/media) says that only Chromecast Ultra can play up to level 5.2 but I dont have an Ultra
 now.

PR #726, opened 2020-01-23
Remove useless autoplay script and add folder for settings

dkanada:


src/components/apphost.js
```diff
@@ -167,24 +167,24 @@ define(["appSettings", "browser", "events", "htmlMediaHelper"], function (appSet
         if (browser.mobile) {
             return false;
         }
-
-        var savedResult = appSettings.get(htmlMediaAutoplayAppStorageKey);
-        return "true" === savedResult || "false" !== savedResult && null;
```
thornbill: This function will no longer return any value by default. That could cause issues for anything that calls this expecting a return value.

src/scripts/site.js
```diff
@@ -865,8 +878,6 @@ var AppInfo = {};
         define("toast", [componentsPath + "/toast/toast"], returnFirstDependency);
         define("scrollHelper", [componentsPath + "/scrollhelper"], returnFirstDependency);
         define("touchHelper", [componentsPath + "/touchhelper"], returnFirstDependency);
-        define("appSettings", [componentsPath + "/appSettings"], returnFirstDependency);
-        define("userSettings", [componentsPath + "/usersettings/usersettings"], returnFirstDependency);
         define("userSettingsBuilder", [componentsPath + "/usersettings/usersettingsbuilder", "layoutManager", "browser"], returnFirstDependency);
```
grafixeyehero: ??

dkanada: good catch

PR #727, opened 2020-01-23
Removes the login required notification and red highlight

ManfredRichthofen:
I understand if you don't want to change this it just annoys me that the username login is constantly surrounded by red

**Changes**
Removes the login required notification and red highlight

### DIFF
--- a/src/login.html
+++ b/src/login.html
@@ -8,7 +8,7 @@
             </div>
 
             <div class="inputContainer">
-                <input is="emby-input" type="text" id="txtManualName" required="required" label="${LabelUser}" autocomplete="username" />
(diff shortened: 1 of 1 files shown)


MrTimscampi:
I think it would be better if we find out why it's constantly red instead of removing the required property, since that attribute is good to have for validating input.

ManfredRichthofen:
If you add `input:required {
    box-shadow:none;
}` it keeps the please fill out this field but remove the red border witch i think is caused from the autocomplete being weird on Firefox as chrome doesn't have the same issue
**Chrome**
ch
**Firefox**

ManfredRichthofen:
Please see pr #730

PR #731, opened 2020-01-24
Fix for live tv channel thumbnails and horizontal scroll on trackpad

dannymichel:
'smushed'/bad aspect ratio images on the live tv channel thumbnails
horizontal scrolling through containers on trackpads.

without my fix: https://i.imgur.com/dMxt4zr.png - with my fix: https://i.imgur.com/qrLh6E1.png 

[css.mp4.zip](https://github.com/jellyfin/jellyfin-web/files/4110869/css.mp4.zip)

**Changes**

**Issues**

### DIFF
--- a/src/components/cardbuilder/card.css
+++ b/src/components/cardbuilder/card.css
@@ -36,6 +36,11 @@ button {
     margin-right: -0.6em;
 }
 
+.itemsContainer.scrollSlider.focuscontainer-x.animatedScrollX {
+    overflow-x: scroll;
+    margin-bottom: -15px
+}
+
 /* TODO replace this with a proper fix */
 /* doesnt work on mobile devices */
 /* negative margin fixes annoying misalignment with cards and title */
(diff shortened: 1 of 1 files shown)


src/components/cardbuilder/card.css
```diff
@@ -36,6 +36,11 @@ button {
     margin-right: -0.6em;
 }
 
+.itemsContainer.scrollSlider.focuscontainer-x.animatedScrollX {
```
dmitrylyzo: @dkanada Maybe we should not use `focuscontainer-x` (and other focusManager-related classes) for styling, since they are used to customize keyboard navigation?

dkanada: Shouldn't be an issue because that section will be removed anyways in favor of the pull request @thornbill has open.

dkanada:
The `visualCardBox` class will actually affect a lot of images, so we need to make sure it doesn't negatively affect anything first.

Issue #487, opened 2019-09-29
Can't trackpad scroll


dannymichel:
**Describe the bug**
You can't horizontally trackpad scroll through .emby-scroller .emby-itemscontainer

dannymichel:
https://www.dropbox.com/s/tc50r669fsbz1g5/scrolling.mp4 vs jellyfin no trackpad scrolling. https://www.dropbox.com/s/tec9x8bcwivaf1e/no-scrolling.mp4 
this is something you can do on plex and emby perfectly
its beed broken on Jellyfin since 10.4. it was fine in 10.3
and its a glaring issue that just has me thinking i can never fully switch to jellyfin

ferferga:
@dannymichel Can you confirm that this is fixed in nightly?

dannymichel:
@ferferga yes it is fixed but https://github.com/jellyfin/jellyfin-web/issues/867

ferferga:
@dannymichel I will leave that one opened then, as the main point stated in this issue has been resolved already :)

PR #733, opened 2020-01-25
[WIP] Fix horizontal scrolling of scrollbutton items

thornbill:
**Changes**
* Fix horizontal scrolling of scrollbutton items
* Update button state when clicking the scrollbuttons
* Disable the left button by default since we always start at the beginning of the list

There is still an issue where the scroll button should be hidden initially if they are not needed. When the buttons are initialized, the data has not been fetched yet though, so we would need a callback somehow once the data is populated.

**Issues**
#487

### DIFF
--- a/src/components/emby-scrollbuttons/emby-scrollbuttons.js
+++ b/src/components/emby-scrollbuttons/emby-scrollbuttons.js
@@ -54,14 +54,21 @@ define(['layoutManager', 'dom', 'css!./emby-scrollbuttons', 'registerElement', '
     }
 
     function onScroll(e) {
-        var scrollButtons = this;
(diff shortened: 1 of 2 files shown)


thornbill:
Well it seems this does not work well in Chrome as the list scrolls too quickly. I will work on improving it there.

dkanada:
There's actually a callback in the item container element but I think I had some issues when I was trying to use it. I couldn't make the changes in a single location and had to pass it manually for every instance of the scroller element in the code.

grafixeyehero:
@thornbill @dkanada here is may solution https://github.com/grafixeyehero/jellyfin-web/commits/enable-horizontal-scroll

PR #735, opened 2020-01-25
Change font family

ferferga:
As I mentioned one day in Jellyfin's matrix rooms, Jellyfin uses a different font family in each system, depending on the available fonts in the system. This PR aims to unify all of that and bundle the font used in Jellyfin's installation, so it no longer depends on what's available in the client's system.

I choose Quicksand in the commit, as it's the font used by the Jellyfin's logo. However, I tried many of them, as I think that Quicksand doesn't fit well. As this is subject to personal taste, I choose some of the most famous fonts in Google Fonts, so the one with most votes gets finally merged into Jellyfin's Web repository.

- Quicksand

Quicksand

- Roboto

Roboto

- Montserrat

Montserrat

- Noto

Noto

- Raleway

Raleway

- Poppins

Poppins

My favourites are Poppins and Raleway, as they look modern and different (unlike Roboto, which is already seen everywhere).

Other suggestions are welcome as well, as long as they can be distributed without issues

### DIFF
--- a/src/assets/css/fonts.css
+++ b/src/assets/css/fonts.css
@@ -1,29 +1,30 @@
 html {
-    font-family: -apple-system, "Helvetica", system-ui, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen-Sans", "Ubuntu", "Cantarell", "Helvetica Neue", 'Open Sans', sans-serif;
+    font-family: "Noto Sans", sans-serif;
     font-size: 93%;
     -webkit-text-size-adjust: 100%;
     text-size-adjust: 100%;
+    -webkit-font-smoothing: antialiased;
+    text-rendering: optimizeLegibility;
 }
 
 h1,
 h2,
 h3 {
-    /* For better bolding, since Helvetica does not support 500 weight, and 600 is too thick */
-    font-family: -apple-system, system-ui, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen-Sans", "Ubuntu", "Cantarell", "Helvetica Neue", 'Open Sans', sans-serif;
+    font-family: "Noto Sans", sans-serif;
 }
 
 h1 {
-    font-weight: 500;
+    font-weight: 400;
     font-size: 1.8em;
 }
 
 h2 {
-    font-weight: 500;
+    font-weight: 400;
     font-size: 1.5em;
 }
 
 h3 {
-    font-weight: 500;
+    font-weight: 400;
     font-size: 1.17em;
 }
 
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "flv.js": "^1.5.0",
     "hls.js": "^0.12.4",
     "howler": "^2.1.2",
+    "jellyfin-noto": "https://github.com/jellyfin/jellyfin-noto",
     "jquery": "^3.4.1",
     "jstree": "^3.3.7",
     "libass-wasm": "^2.1.1",
--- a/src/assets/css/fonts.sized.css
+++ b/src/assets/css/fonts.sized.css
@@ -1,5 +1,5 @@
 h1 {
-    font-weight: 500;
+    font-weight: 400;
     font-size: 1.8em;
 }
 
@@ -8,12 +8,12 @@ h1 {
 }
 
 h2 {
-    font-weight: 500;
+    font-weight: 400;
     font-size: 1.5em;
 }
 
 h3 {
-    font-weight: 500;
+    font-weight: 400;
     font-size: 1.17em;
 }
 
--- a/src/bundle.js
+++ b/src/bundle.js
@@ -102,3 +102,8 @@ var material_icons = require("material-design-icons-iconfont/dist/material-desig
 _define("material-icons", function() {
     return material_icons;
 });
+
+var jellyfin_noto = require("jellyfin-noto");
+_define("jellyfin-noto", function () {
+    return jellyfin_noto;
+});
--- a/src/scripts/site.js
+++ b/src/scripts/site.js
@@ -732,7 +732,8 @@ var AppInfo = {};
                     "sortable",
                     "libjass",
                     "webcomponents",
-                    "material-icons"
+                    "material-icons",
+                    "jellyfin-noto"
                 ]
             },
             urlArgs: urlArgs,
@@ -747,6 +748,7 @@ var AppInfo = {};
         });
 
         require(["css!assets/css/site"]);
+        require(["jellyfin-noto"]);
 
         // define styles
         // TODO determine which of these files can be moved to the components themselves
--- a/yarn.lock
+++ b/yarn.lock
@@ -675,6 +675,24 @@ atob@^2.1.1:
   resolved "https://registry.yarnpkg.com/atob/-/atob-2.1.2.tgz#6d9517eb9e030d2436666651e86bd9f6f13533c9"
   integrity sha512-Wm6ukoaOGJi/73p/cl2GvLjTI5JM1k/O14isD73YML8StrH/7/lRFgmg8nICZgD3bZZvjwCGxtMOD3wWNAu8cg==
 
+autoprefixer@^9.0.0, autoprefixer@^9.7.3:
+  version "9.7.4"
+  resolved "https://registry.yarnpkg.com/autoprefixer/-/autoprefixer-9.7.4.tgz#f8bf3e06707d047f0641d87aee8cfb174b2a5378"
+  integrity sha512-g0Ya30YrMBAEZk60lp+qfX5YQllG+S5W3GYCFvyHTvhOki0AEQJLPEcIuGRsqVwLi8FvXPVtwTGhfr38hVpm0g==
+  dependencies:
+    browserslist "^4.8.3"
+    caniuse-lite "^1.0.30001020"
+    chalk "^2.4.2"
+    normalize-range "^0.1.2"
+    num2fraction "^1.2.2"
+    postcss "^7.0.26"
+    postcss-value-parser "^4.0.2"
+
+bail@^1.0.0:
+  version "1.0.5"
+  resolved "https://registry.yarnpkg.com/bail/-/bail-1.0.5.tgz#b6fa133404a392cbc1f8c4bf63f5953351e7a776"
+  integrity sha512-xFbRxM1tahm08yHBP16MMjVUAvDaBMD38zsM9EMAUN61omwLmKlOpB/Zku5QkjZ8TZ4vn53pj+t518cH0S03RQ==
+
 balanced-match@^1.0.0:
   version "1.0.0"
   resolved "https://registry.yarnpkg.com/balanced-match/-/balanced-match-1.0.0.tgz#89b4d199ab2bee49de164ea02b89ce462d71b767"
@@ -856,6 +874,23 @@ browserify-zlib@^0.2.0:
   dependencies:
     pako "~1.0.5"
 
+browserslist@^1.1.3:
+  version "1.7.7"
+  resolved "https://registry.yarnpkg.com/browserslist/-/browserslist-1.7.7.tgz#0bd76704258be829b2398bb50e4b62d1a166b0b9"
+  integrity sha1-C9dnBCWL6CmyOYu1Dkti0aFmsLk=
+  dependencies:
+    caniuse-db "^1.0.30000639"
+    electron-to-chromium "^1.2.7"
+
+browserslist@^4.8.3:
+  version "4.8.6"
+  resolved "https://registry.yarnpkg.com/browserslist/-/browserslist-4.8.6.tgz#96406f3f5f0755d272e27a66f4163ca821590a7e"
+  integrity sha512-ZHao85gf0eZ0ESxLfCp73GG9O/VTytYDIkIiZDlURppLTI9wErSM/5yAKEq6rcUdxBLjMELmrYUJGg5sxGKMHg==
+  dependencies:
+    caniuse-lite "^1.0.30001023"
+    electron-to-chromium "^1.3.341"
+    node-releases "^1.1.47"
+
 buffer-from@^1.0.0:
   version "1.1.1"
   resolved "https://registry.yarnpkg.com/buffer-from/-/buffer-from-1.1.1.tgz#32713bc028f75c02fdb710d7c7bcec1f2c6070ef"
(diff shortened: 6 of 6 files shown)


anthonylavado:
I vote for Noto, which seems to be the consensus in chat so far. Quicksand was only ever meant to be the logo, and we used it on the website because it stayed “on brand”.

ferferga:
I like Noto as well. It also doesn't differ a lot from the other system fonts that Jellyfin was using, which I think it's also a factor to consider.

Let's keep this open here for a while for everyone choosing and explaining the reasons for their selection. Once it's more less clear that everyone likes Noto, I will add another commit so it can be reviewed and merge. Right for you, @anthonylavado ?

AndreasGB:
I'd also vote for Noto, it looks super clean in comparison to some of the others

ferferga:
Everyone agreed that Noto Sans should be the way to go. With the merge conflicts fixed, this is ready to be merged now.

CC @anthonylavado

src/assets/css/fonts.css
```diff
@@ -1,36 +1,110 @@
...
+  font-family: 'Noto Sans';
+  font-style: normal;
+  font-weight: 400;
+  font-display: swap;
+  src: url(fonts/notosans/latin.woff2) format('woff2');
+  unicode-range: U+0000-00FF, U+0131, U+0152-0153, U+02BB-02BC, U+02C6, U+02DA, U+02DC, U+2000-206F, U+2074, U+20AC, U+2122, U+2191, U+2193, U+2212, U+2215, U+FEFF, U+FFFD;
+}
+
 html {
-    font-family: -apple-system, "Helvetica", system-ui, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen-Sans", "Ubuntu", "Cantarell", "Helvetica Neue", 'Open Sans', sans-serif;
```
MrTimscampi: It's weird that the linting didn't trip on this, but the standard (for now anyway) is 4 spaces.

src/assets/css/fonts.css
```diff
@@ -1,36 +1,110 @@
+/* cyrillic-ext */
```
MrTimscampi: Bit of a personal request (and the approval isn't dependent on this), but would it be possible to add the CJK fonts as well ?

The unicode ranges should be as follow:

Japanese

`unicode-range: U+25e56, U+25e62, U+25e65, U+25ec2, U+25ed8, U+25ee8, U+25f23, U+25f5c, U+25fd4, U+25fe0, U+25ffb, U+2600c, U+26017, U+26060, U+260ed, U+26222, U+2626a, U+26270, U+26286, U+2634c, U+26402, U+2667e, U+266b0, U+2671d, U+268dd, U+268ea, U+26951, U+2696f, U+26999, U+269dd, U+26a1e, U+26a58, U+26a8c, U+26ab7, U+26aff, U+26c29, U+26c73, U+26c9e, U+26cdd, U+26e40, U+26e65, U+26f94, U+26ff6-26ff8, U+270f4, U+2710d, U+27139, U+273da-273db, U+273fe, U+27410, U+27449, U+27614-27615, U+27631, U+27684, U+27693, U+2770e, U+27723, U+27752, U+278b2, U+27985, U+279b4, U+27a84, U+27bb3, U+27bbe, U+27bc7, U+27c3c, U+27cb8, U+27d73, U+27da0, U+27e10, U+27fb7, U+2808a, U+280bb, U+28277, U+28282, U+282f3, U+283cd, U+2840c, U+28455, U+2856b, U+285c8-285c9, U+286d7, U+286fa, U+28946, U+28949, U+2896b, U+28987-28988, U+289ba-289bb, U+28a1e, U+28a29, U+28a43, U+28a71, U+28a99, U+28acd, U+28add, U+28ae4, U+28bc1, U+28bef, U+28cdd, U+28d10, U+28d71, U+28dfb, U+28e0f, U+28e17, U+28e1f, U+28e36, U+28e89, U+28eeb, U+28ef6, U+28f32, U+28ff8, U+292a0, U+292b1, U+29490, U+295cf, U+2967f, U+296f0, U+29719, U+29750, U+29810, U+298c6, U+29a72, U+29d4b, U+29ddb, U+29e15, U+29e3d, U+29e49, U+29e8a, U+29ec4, U+29edb, U+29ee9, U+29fce, U+29fd7, U+2a01a, U+2a02f, U+2a082, U+2a0f9, U+2a190, U+2a2b2, U+2a38c, U+2a437, U+2a5f1, U+2a602, U+2a61a, U+2a6b2, U+2a9e6, U+2b746, U+2b751, U+2b753, U+2b75a, U+2b75c, U+2b765, U+2b776-2b777, U+2b77c, U+2b782, U+2b789, U+2b78b, U+2b78e, U+2b794, U+2b7ac, U+2b7af, U+2b7bd, U+2b7c9, U+2b7cf, U+2b7d2, U+2b7d8, U+2b7f0, U+2b80d, U+2b817, U+2b81a, U+2f804, U+2f80f, U+2f815, U+2f818, U+2f81a, U+2f822, U+2f828, U+2f82c, U+2f833, U+2f83f, U+2f846, U+2f852, U+2f862, U+2f86d, U+2f873, U+2f877, U+2f884, U+2f899-2f89a, U+2f8a6, U+2f8ac, U+2f8b2, U+2f8b6, U+2f8d3, U+2f8db-2f8dc, U+2f8e1, U+2f8e5, U+2f8ea, U+2f8ed, U+2f8fc, U+2f903, U+2f90b, U+2f90f, U+2f91a, U+2f920-2f921, U+2f945, U+2f947, U+2f96c, U+2f995, U+2f9d0, U+2f9de-2f9df, U+2f9f4;`

Simplified Chinese

`unicode-range: U+1f1e9-1f1f5, U+1f1f7-1f1ff, U+1f21a, U+1f232, U+1f234-1f237, U+1f250-1f251, U+1f300, U+1f302-1f308, U+1f30a-1f311, U+1f315, U+1f319-1f320, U+1f324, U+1f327, U+1f32a, U+1f32c-1f32d, U+1f330-1f357, U+1f359-1f37e;`

Traditional Chinese

`unicode-range: U+1f921-1f930, U+1f932-1f935, U+1f937-1f939, U+1f940-1f944, U+1f947-1f94a, U+1f950-1f95f, U+1f962-1f967, U+1f969-1f96a, U+1f980-1f981, U+1f984-1f98d, U+1f990-1f992, U+1f994-1f996, U+1f9c0, U+1f9d0, U+1f9d2, U+1f9d4, U+1f9d6, U+1f9d8, U+1f9da, U+1f9dc-1f9dd, U+1f9df-1f9e2, U+1f9e5-1f9e6, U+20024, U+20487, U+20779, U+20c41, U+20c78, U+20d71, U+20e98, U+20ef9, U+2107b, U+210c1, U+22c51, U+233b4, U+24a12, U+2512b, U+2546e, U+25683, U+267cc, U+269f2, U+27657, U+282e2, U+2898d, U+29d5a, U+f0001-f0005, U+f0019, U+f009b, U+f0101-f0104, U+f012b, U+f01ba, U+f01d6, U+f0209, U+f0217, U+f0223-f0224, U+fc355, U+fe327, U+fe517, U+feb97, U+fffb4;`

Korean

`unicode-range: U+f9ca-fa0b, U+ff03-ff05, U+ff07, U+ff0a-ff0b, U+ff0d-ff19, U+ff1b, U+ff1d, U+ff20-ff5b, U+ff5d, U+ffe0-ffe3, U+ffe5-ffe6;`

They're the three other scripts that aren't supported properly currently, so it'd be nice to have them in as well, if only for consistency.

src/assets/css/fonts.css
```diff
@@ -1,36 +1,110 @@
...
+  font-weight: 400;
+  font-display: swap;
+  src: url(fonts/notosans/greek-ext.woff2) format('woff2');
+  unicode-range: U+1F00-1FFF;
+}
+/* greek */
+@font-face {
+  font-family: 'Noto Sans';
+  font-style: normal;
+  font-weight: 400;
```
MrTimscampi: We also likely need at least a bold version of fonts (font-weight: 700) as well. Not much variation in font-weight at the moment in the UI (Though there's one <b> tag and a handful of 500 and 700 in there), but we likely want to change that down the line to emphasis some visual elements.

src/assets/css/fonts.css
MrTimscampi: Also note that we officially ship Arabic and Hebrew translations as well, which need their own fonts.

ferferga: I added all the Noto Sans typefaces available at Google Fonts + Hebrew and Arabic in early access!

ferferga:
Added all Noto Sans typefaces available at Google Fonts + arabic and hebrew (available at early acces for the moment, will keep taking a look to update them in case Google updates them too)

Screenshot with Japanese test using Noto Sans:

japanese

ferferga:
@dkanada Another solution would be to merge all the 100 files of the Japanese, Simplified Chinese and Korean in a single file (one file per font family), although that means that browsers will load more characters than the ones that it needs to be displayed, as it has to download the whole charset.

That's how I started to do it yesterday (just merge all of them in one file and forget about unicode ranges), but looking for information about the purpose of unicode-range css property, Mozilla and W3 states that it's preferrable to have all the fonts as splitted as possible to save bandwidth and resources on the systems. That's why Google Fonts split them a lot, to only serve what's needed at each moment.

I did a quick search in npm in mobile and all modules I found do what I say, merge them all in one file and load the whole charset. So it really depends in what we prefer, development convenience over speed. The fonts however shouldn't need further modifications (except Arabic and Hebrew maybe), so this is a merge and forget.

CC @MrTimscampi for ideas as well in how to handle this.

MrTimscampi:
> I think we should go searching for a package to pull in through yarn rather than include them all in the source.

Agree with this. Merging them doesn't make sense since we have users using mobile network through the Android/iOS clients and a huge font will kill their bandwidth.

We should also be keeping the unicode-range property, since it helps with load times.

I found a few packages on NPM: https://www.npmjs.com/search?q=%40openfonts%2Fnoto-sans

Main issues I see with these is: very few downloads (people usually use remote fonts...), not all charsets are available, should we eventually need them.

Our specific requirements make this kind of a hard problem to solve, I think.

We can't use web fonts, even if that'd be the easiest.  
We could ship only the main font, but then a whole bunch of users aren't covered (As an aside, we should test layouts of other scripts more often. Specifically thinking about Arabic here, since it reads the other way around)

So if we're okay with these package, I say go for it. Otherwise, we'll need to see what can be done...

dkanada:
We could always start hosting assets like these with a webserver. I'd be willing to set that up and we probably need something like that anyways.

AndreasGB:
I'd also recommend packaging them, otherwise we have an indirect phone-home via the fonts server. Could also be nice to get an estimate about the install base, but probably not aligned with the project values.

dkanada:
@AndreasGB we could also track metrics by scraping download counts, logging plugin requests, checking documentation usage, or any number of methods. Users could always choose to ignore remote fonts if they are worried about it. That doesn't even include the fact that anyone who installs through Google Play, Amazon, or any other future platforms is forced to give us extremely exhaustive metrics through usage of those services.

The one argument I can see for packaging is to avoid external assets and keep everything as local as possible. We still have lots of remote images and other libraries littered throughout the code though, so it's a work in progress. I do agree with you, but not for the same reasons.

Another factor to consider is the state of package management for Javascript. I might go so far as to say I trust us to host the assets more than I trust a random dependency that sees a few hundred downloads a week.

MrTimscampi:
We could provide a Display setting to disable remote font loading ?  Include it in the initial setup so it's fair to users, perhaps default to false if not set.

We can provide a graceful fallback through CSS (Have it fallback to Helvetica, Roboto and others) and enable remote loading of fonts only if the user setting is enabled.

Provide a list of fonts in the docs, linked in the help text for the option, to allow users to install local versions of the fonts if they want to have the "official" look of text without pinging to us.

It's a bit much, especially since we also ping back for plugins and so on, but at least it gives the option to choose.

dkanada:
That honestly sounds like an option for the web client global configuration file I'm planning on adding for admin usage.

ferferga:
I created a ``yarn`` package in [this repository](https://github.com/ferferga/jellyfin-noto). However, I'm not sure what's the best method to load the font-faces.css file properly, so the font is applied globally.

AndreasGB:
If the yarn package is the way to go, it should probably be transferred to the Jellyfin organization.

MrTimscampi:
> I created a `yarn` package in [this repository](https://github.com/ferferga/jellyfin-noto). However, I'm not sure what's the best method to load the font-faces.css file properly, so the font is applied globally.

You can use `yarn add https://github.com/ferferga/jellyfin-noto` to add the package to the project.

To use the module in the code, take a look at `bundle.js`. Add the package there, then to the "bundles" array in `site.js`. Then in the same file, you can do `require(["jellyfin-noto"]);` and that will import the CSS file. I think the best place to set it is before the require of site.css (a little below requirejs.config), that way it's loaded before everything.

dkanada:
@AndreasGB just forked it at jellyfin/jellyfin-noto if that works for @ferferga.

ferferga:
@AndreasGB That's my idea, but I have no write access to the organization. I suppose we will move it when this PR is merged

@MrTimscampi Just testing, was missing the ``bundle.js`` to make it work before. I will push the changes in a moment

@dkanada I think it's better to move it, so I can directly make PRs and merge fonts if needed there. I just made a commit to make it work (two assets were misplaced) and I think it's a bigger cumbersome syncing the history changes for you, right?. 
Will however link to the jellyfin organization in both ``package.json`` files (jellyfin-not and jellyfin-web) if all reviews are positive about this PR, so when this get merged, my repo can simply be moved without issues.

dkanada:
@ferferga that sounds good, and this pull request will definitely getting merged, we need a standard font throughout the web client.

dkanada:
Although actually, we should also consider the effects on other clients. The Android app will package all these files and nearly quadruple in size.

ferferga:
Everything working correctly now, except that ``yarn install`` is throwing ``RangeError: Maximum call stack size exceeded`` if I add the lines in ``bundle.js`` before installing the package. If I add & install the package first, and I add the line in ``bundle.js`` afterwards it works perfectly. 
This is why I think that checks are failing (also because jellyfin/jellyfin-noto repository doesn't exist yet). I'm ready to move the repository whenever the current ``jellyfin/jellyfin-noto`` is deleted.

@dkanada That's a good point. Is there any way to target an specific package for mobile only? Because as resources in mobile doesn't need to be fetched from the network, I think it makes total sense in this case to merge the font packages in as few files possible, reducing the font's footprint from **50 MBs** to maybe **10 MB**

ferferga:
> Everything working correctly now, except that `yarn install` is throwing `RangeError: Maximum call stack size exceeded` if I add the lines in `bundle.js` before installing the package. If I add & install the package first, and I add the line in `bundle.js` afterwards it works perfectly. Any idea why @MrTimscampi ?
> This is why I think that checks are failing (also because jellyfin/jellyfin-noto repository doesn't exist yet). I'm ready to move the repository whenever the current `jellyfin/jellyfin-noto` is deleted.
> 
> @dkanada That's a good point. Is there any way to target an specific package for mobile only? Because as resources in mobile doesn't need to be fetched from the network, I think it makes total sense in this case to merge the font packages in as few files possible, reducing the font's footprint from **50 MBs** to maybe **10 MB**

The ``RangeError: Maximum call stack size exceeded`` was fixed in the latest commit of jellyfin-noto repository.

ferferga:
@dkanada Btw, we agreed in removing ``woff`` files from the package, right?

dkanada:
yeah, just the woff files

ferferga:
@dkanada PR for removing woff files at jellyfin/jellyfin-noto#1 .

PR #736, opened 2020-01-25
Remove "Automatically update this server" checkbox

ferferga:
Removed the "Automatically update this server" checkbox, as it actually does nothing


ferferga:
@dkanada See referenced PR

dkanada:
It should generally be fine. I'll get access to the Weblate instance though so I can resolve those issues in the future.

PR #737, opened 2020-01-25
Miscellaneous CSS fixes

MrTimscampi:
**Changes**

*First image is before, second is after*

Properly center the server selection cards

image

Screenshot_2020-01-25 Jellyfin

Rework the margins and alignments in the headerbar

image

image

Rework text alignment in item lists

image

image

Aligned section text with chevron

image

image

Align sidebar menu option text with icon

image

image

**Issues**

None

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "serve": "webpack-dev-server --config webpack.dev.js --open",
     "build": "webpack --config webpack.prod.js",
     "lint": "eslint \"src\"",
-    "stylelint": "stylelint src/**/*.css",
+    "stylelint": "stylelint \"src/**/*.css\"",
     "prepare": "webpack --config webpack.prod.js"
   }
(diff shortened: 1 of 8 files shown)


dkanada:
1. Does the new padding on the edges of the header look alright with the main body?
2. I think the login page looked like that to accomodate servers with enough users to fill two or three rows, what does that page look like now?

MrTimscampi:
Here's what it looks like with multiple servers:

image

Originally, on that page, all the margins were on top and left. This change divides the left margin by two then adds a right margin of the same value to the right to keep the spacing correct.

As for the rest of the page after the header change, here's how it looks:

image

MrTimscampi:
Went a bit over the tabs to get them closer to the Material Design guidelines (since the rest of the client follows it somewhat)

image

PR #738, opened 2020-01-25
Switched OS media controls

ferferga:
Jellyfin was the only app I knew that displayed the artist's name as the title of the item in the OS media controls of all the devices I have. I don't know if that was intended that way, but how I left it makes much more sense to me:

* **Before**

Screenshot_20200126_000327_com android chrome
Screenshot_20200126_000332_com android keyguard
media_controls2

* **After**

Screenshot_20200126_000150_com huawei android launcher
Screenshot_20200126_000155_com android keyguard
media_controls

### DIFF
--- a/src/components/playback/mediasession.js
+++ b/src/components/playback/mediasession.js
@@ -97,11 +97,8 @@ define(['playbackManager', 'nowPlayingHelper', 'events', 'connectionManager'], f
     }
 
     function updatePlayerState(player, state, eventName) {
-        var now = new Date().getTime();
-        lastUpdateTime = now;
-
         // Don't go crazy reporting position changes
-        if (eventName == 'timeupdate' && (now - lastUpdateTime) < 5000) {
+        if (eventName == 'timeupdate') {
             // Only report if this item hasn't been reported yet, or if there's an actual playback change.
             // Don't report on simple time updates
             return;
--- a/src/components/playback/mediasession.js
+++ b/src/components/playback/mediasession.js
@@ -160,8 +160,8 @@ define(['playbackManager', 'nowPlayingHelper', 'events', 'connectionManager'], f
 
         if (navigator.mediaSession) {
             navigator.mediaSession.metadata = new MediaMetadata({
-                title: title,
-                artist: artist,
+                title: artist,
+                artist: title,
                 album: album,
                 artwork: getImageUrls(item),
                 albumArtist: albumArtist,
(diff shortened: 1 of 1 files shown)


src/components/playback/mediasession.js
```diff
@@ -160,8 +160,8 @@ define(['playbackManager', 'nowPlayingHelper', 'events', 'connectionManager'], f
 
         if (navigator.mediaSession) {
             navigator.mediaSession.metadata = new MediaMetadata({
-                title: title,
-                artist: artist,
+                title: artist,
+                artist: title,
```
dkanada: I would rather find the code that inverts the logic and remove that.

ferferga:
Logic fixed. I also noticed that updates were received, processed and discarded with a return if there were no changes. I moved that check to the top of the logic, so it does check it first, which seems more logical to me.

src/components/playback/mediasession.js
```diff
@@ -118,19 +127,9 @@ define(['playbackManager', 'nowPlayingHelper', 'events', 'connectionManager'], f
...
 
         var playState = state.PlayState || {};
-
         var parts = nowPlayingHelper.getNowPlayingNames(item);
-
-        var artist = parts.length === 1 ? '' : parts[0].text;
-        var title = parts[parts.length - 1].text;
-
-        // Switch these two around for video
-        if (isVideo && parts.length > 1) {
```
JustAMan: do you have any clue on why was this ever needed?..

ferferga: No idea. It was right, however, given how the variables were first declared, don't know why they didn't do the same for music 😆.

anthonylavado:
Maybe this is what was happening to @Artiume in this issue/PR: https://github.com/jellyfin/jellyfin-android/pull/234

ferferga:
99% sure this is the solution. For all devices that support media controls, not only for Bluetooth audio though.

src/components/playback/mediasession.js
```diff
@@ -97,6 +97,15 @@ define(['playbackManager', 'nowPlayingHelper', 'events', 'connectionManager'], f
     }
 
     function updatePlayerState(player, state, eventName) {
+        var now = new Date().getTime();
+        lastUpdateTime = now;
+
+        // Don't go crazy reporting position changes
+        if (eventName == 'timeupdate' && (now - lastUpdateTime) < 5000) {
```
marius-luca-87: lastUpdateTime = now;
now - lastUpdateTime = 0

Artiume:
Thanks for finding the fix :)

src/components/playback/mediasession.js
ferferga: Wow, completely missed that. It's crazy to see what we got from Emby codebase!

Thanks for pointing it out!

ferferga:
@marius-luca-87 For some reason, clicking on the re-request button doesn't do anything, so I can't notify you. Can you review this again, please?

PR #740, opened 2020-01-26
Enable imageLoader fade-in effect

MrTimscampi:
**Changes**

Enables the fade-in effect for the imageLoader component and adds a new Display setting to enable faster animations.

image

**Issues**

None


MrTimscampi:
> Can we make animation speed tunable instead of a boolean fast/slow choice?

We can, yes. I need to see what I can whip up that would be compatible we every browser we support.

Otherwise, this is a good first step until that is done.

PR #741, opened 2020-01-26
Added color while hovering over buttons in the new layout

ferferga:
Probably this is pointless as I suppose the new layout is still a WIP and this was already planned, but hovering over buttons in the new layout used in item's details page needed some improvement in my opinion, as with 50% of the opacity texts were barely readable. The color used is the same used in the "Play" material icon that appears while hovering over a poster.

* **Before**

old

* **After**

new

### DIFF
--- a/src/elements/emby-button/emby-button.css
+++ b/src/elements/emby-button/emby-button.css
@@ -47,7 +47,7 @@
 }
 
 .button-flat:hover {
-    opacity: .5;
+    background-color: rgba(0,164,220, .4);
 }
 
 .button-link {


JustAMan:
@ferferga can we make the button use the colour of enclosing theme somehow?

dkanada:
I would actually rather highlight the text and icon rather than wrap it in a box, but that might require larger changes.

ferferga:
@JustAMan I'm not exactly sure I understand what you mean, but the blue tone used is the exact same color value of all the "accented" elements used in the UI. It's the same blue tone used for buttons, playback sliders, selection menus... And that blue tone doesn't change at all if you change your UI theme, it's the same blue in Light theme, dark theme and all the rest included with Jellyfin.

Thought exactly about this when making the PR and started to wonder where I could code the color change depending on the theme chosen, when I recalled that all of the themes had the same color scheme. I checked again with Blue Radiance, Light and Dark to confirm this.

Anyway, I thought during this last month I've been using Jellyfin intensively for listening music while studying that an accent color picker should be added to the Web config, in a Windows-alike style. New UI overhauls like this should just take the color value chosen by the user and forget about themes, while also allowing greater customization in a familiar way for users. But for now, nothing like this still exists. It may be my next PR though :P

@dkanada It should only be one line change in the CSS, so I suppose no problem? That's how I did it first, but it didn't seem that "noticeable" to me that the button was hovered, so went for the "box" design. I will try to attach an screenshot tomorrow so you can see the differences.

dkanada:
An accent color would be very hard to include with the theme support so I'm not sure it's a good idea with the current codebase. It also doesn't exactly make sense when we can just write entire themes to customize the interface, unlike platforms like Android where basic theme support is the only thing available.

ferferga:
@dkanada I disagree with you. I already took a look when I first thought about this and it should be as simple as overriding the color of each element in the JS script that's involved for that specific element. An interface like this is miles ahead of CSS hacks for normal and not technically-inclined users, while still providing a little of customization.

Btw, here is the screenshot of the colored button.
Sin título

dkanada:
@ferferga a text color change was added on hover, would that be sufficient for now?

ferferga:
@dkanada Sorry for my delayed response, as you know, I've been very busy the last two weeks. Now I'm fully back on track!

I pulled the latest master branch and just saw the color for the buttons. Much better as it is right now, I think that we can close this then!

Reopen if you disagree.

Issue #525, opened 2019-10-13
Controllers should not act on UI when not in focus


xkronusx:
**Describe the bug**

Jellyfin Web always accepts control inputs from xbox one controllers even when not in focus.

**To Reproduce**

1. Have controller connected
2. Click any face buttons
3. Web page will go back, stop, start, attempt to open NULL pages etc.

**Expected behavior**
Xbox Controller should not have control of client by default (web). Should be a toggle option to disable.

**System (please complete the following information):**
 - OS: Docker
 - Browser: Chrome
 - Jellyfin Version:  10.4.0
 - Reverse proxy: Traefik

anthonylavado:
_moved to web_

anthonylavado:
This has also come up again with someone else on the forums, using a joystick: https://forum.jellyfin.org/t/weird-problem-setting-up-one-pw-field-full-of-chars-one-empty/891/10

And also with a steering wheel: https://github.com/jellyfin/jellyfin/issues/1762

t3knical:
for the time being till a button is made in the settings, navigate to \jellyfin-web\components\serverNotifications\gamepadtokey.js and comment out //runInputLoop(); near the end.

djpbessems:
I'm having the exact same issue. I have a X55 Rhino HOTAS.

I don't think a toggle-preference is going to be good enough though, the gamepadtokey.js should probably be made aware of the concept of more than 2 axis and if it encounters more, than it should disregard any axis that is not at a zero-state or near zero (deadzone).

djpbessems:
I'm having the exact same issue. I have a X55 Rhino HOTAS.

I don't think a toggle-preference is going to be good enough though, the `gamepadtokey.js` should probably be made aware of the concept of more than 2 axis and if it encounters more, than it should disregard any axis that is not at a zero-state or near zero (deadzone).

For now, a better solution than to edit the .js file is to go to `about:config` in either Firefox or Chrome and toggle the setting `dom.gamepad.enabled` to `false`

mijofa:
I've run into this a couple times and only just figured out this is what's actually causing it, there was *so much* confusion!
Consider this a +1/upvote/whatever for this issue

t3knical:
> I'm having the exact same issue. I have a X55 Rhino HOTAS.
> 
[...]

PR #746, opened 2020-01-29
Load gamepadtokey.js only when is necessary

### LINKED ISSUE #602: GamepadToKey.js Causes Display Lag
@JustAMan I already found that the source of the lag is caused by the 'gamepadtokey.js' script under /components/serverNotifications. I removed it just fine and I've been using it on all my devices without issues. Although I would like to investigate the root of it (and completely remove scripts that call it in case it's obsolote and no longer doing a useful job), I think that it would be a good idea to remove the file in 10.4.2 packages, as a quick and temporary hotfix. It's really annoying how it lags right now.

CC @joshuaboniface as well

@dkanada Thanks to you as well for approving it 😊. It's great to see that the contributions are considered to be useful for some people!

_Originally posted by @ferferga in https://github.com/jellyfin/jellyfin-web/pull/545#issuecomment-557919246_

ferferga:
This PR aims to fix all the problems and lags caused by the resource-hungry ``gamepadtokey.js`` script, only loading it when a gamepad is connected and detected.

This is a WIP because I don't have a controller at hand, so everything here has been done based in assumptions and hasn't been thoroughly tested. All the help that can be provided for testing this with a gamepad is more than appreciated

**Changes**

The logic that started the ``gamepadtokey.js`` in ``serverNotifications.js`` (L201) was always returning undefined, so the function ``enableNativeGamepadKeyMapping`` was always false. The second check done in ``isGamepadSupported`` was always true, as all modern browsers support gamepads, although that doesn't mean that a gamepad is connected. 

Now, these checks have been replaced in ``L216``, calling the ``getGamepads()`` method. If the browser supports that function, it will be true, else undefined, not triggering the gamepadtokey.js script. That way, two checks have been replaced by one.

In that if statement, we also check the first index of the array returned, as [described here](https://developer.mozilla.org/en-US/docs/Web/API/Gamepad_API/Using_the_Gamepad_API#querying), so we know if a gamepad is connected or not.

Also, I removed all those checks in ``gamepadtokey.js``, as they're useless there because all the gamepad checks have been made in ``serverNotifications.js``. We also removed the function ``webkitGetGamepads()`` as it's completely deprecated.

**Concerns**

I have some doubts that I want to cover/discuss
* I also disabled ``mouseManager.js`` because it doesn't seem to do anything, at least that's what I found. Any clue on what it does?
* Any reason the proposed solution in #525 wasn't merged in the code? We can add that change to this PR as well.

**Issues**
Fixes #602 and should improve #545, as well as overall performance of the web interface

**To test**

* Connecting a gamepad and entering Jellyfin
* Connecting a gamepad while Jellyfin is running and check that is handled correctly and the gamepadtokey.js script is started.
* Disconnect a gamepad while Jellyfin is running and check that the gamepad script is killed.


dkanada:
Regarding #525, do you mean the user preference? Also, I hate that this file is loaded by the server notifications file, it should be placed with the rest of the navigation files like the keyboard manager and such. Server notifications shouldn't depend on a gamepad library.

ferferga:
Interesting... I rolled the changes into my production server and asked a friend who owns a Xbox to try the changes.

He received the popup that "This page uses game controls instead of a cursor. Press - for changing between the different modes" correctly, so it seems that my changes trigger the script correctly when a gamepad is connected. However, the UI in gamepad mode didn't work at all, he described it to me with the following sentence: "is like the gamepad doesn't know where each card is located".

I rolled back the original code and asked him to reboot and load Jellyfin again and... Same popup and same issue.
So the problem wasn't related to my changes. Seems that ``gamepadtokey.js`` is not only a CPU hog, but it has been broken for a while now?

@dkanada In ``keyboardnavigation.js`` you mean right? That should be fairly easy, I will try to do it this afternoon. No clue about ``mousemanager.js`` right?

MrTimscampi:
> "is like the gamepad doesn't know where each card is located"

This is likely the Focus Manager. It likes to not follow logic when deciding where to move the focus.

Pretty much every keyboard-like navigation has that issue currently (It's a pain on WebOS, for example)

src/components/serverNotifications/gamepadtokey.js
```diff
@@ -254,11 +254,7 @@ require(['apphost'], function (appHost) {
     function runInputLoop() {
         // Get the latest gamepad state.
         var gamepads;
-        if (navigator.getGamepads) {
-            gamepads = navigator.getGamepads();
-        } else if (navigator.webkitGetGamepads) {
-            gamepads = navigator.webkitGetGamepads();
-        }
+        gamepads = navigator.getGamepads();
         gamepads = gamepads || [];
```
JustAMan:
```js
var gamepads = navigator.getGamepads() || [];
```
?

src/components/serverNotifications/serverNotifications.js
```diff
@@ -197,30 +197,30 @@ define(['connectionManager', 'playbackManager', 'events', 'inputManager', 'focus
         events.on(apiClient, "message", onMessageReceived);
     }
 
-    function enableNativeGamepadKeyMapping() {
-        if (window.navigator && "string" == typeof window.navigator.gamepadInputEmulation) {
-            window.navigator.gamepadInputEmulation = "keyboard";
-            return true;
+    function attachgamepad(e) {
```
JustAMan:
```suggestion
    function attachGamepad(e) {
```

src/components/serverNotifications/serverNotifications.js
```diff
@@ -197,30 +197,30 @@ define(['connectionManager', 'playbackManager', 'events', 'inputManager', 'focus
...
+        if (navigator.getGamepads.length > 0) {
```
JustAMan:
```suggestion
        if (navigator.getGamepads().length > 0) {
```
?

src/components/serverNotifications/serverNotifications.js
```diff
@@ -197,30 +197,30 @@ define(['connectionManager', 'playbackManager', 'events', 'inputManager', 'focus
...
+            require(["components/serverNotifications/gamepadtokey"]);
+            console.log("Gamepad connected! Attaching gamepadtokey.js script");
         }
-
-        return false;
     }
 
-    function isGamepadSupported() {
-        return "ongamepadconnected" in window || navigator.getGamepads || navigator.webkitGetGamepads;
+    function dettachgamepad(e) {
```
JustAMan:
```suggestion
    function dettachGamepad(e) {
```

src/components/serverNotifications/serverNotifications.js
```diff
@@ -197,30 +197,30 @@ define(['connectionManager', 'playbackManager', 'events', 'inputManager', 'focus
...
+        console.log("Gamepad connected! Attaching gamepadtokey.js script");
+        window.addEventListener("gamepaddisconnected", dettachgamepad);
+    } else {
+        window.addEventListener("gamepadconnected", attachgamepad);
+        console.log("No gamepad connected to this device");
     }
-
-    require(["components/serverNotifications/mouseManager"]);
-
+    // require(["components/serverNotifications/mouseManager"]);
```
JustAMan: why?

ferferga: As I specify in the PR, it doesn't seem to me that mousemanager.js does something useful. Do you know its purpose?

If we are sure that it's useful, we can keep it, otherwise I think we should get rid of it in this PR.

src/components/serverNotifications/serverNotifications.js
ferferga: Thanks for noticing, wrote that ``if`` statement before the one in L216 and I changed the L216 one but forgot about the event handler. However, with that, you are receiving an array that contains 4 arrays, which are empty if there are no controllers attached. That means that the condition will always be true.

I'm still exploring ways in checking the number of gamepads connected, the proposed solution in Mozilla docs (the one I use in L216) doesn't convince me at all for our use case.

Will expand this information whenever I push the next commit to this PR.

src/components/serverNotifications/serverNotifications.js
dkanada: cc @dmitrylyzo

dmitrylyzo: `mouseManager` detects mouse inactivity - to hide the cursor (adds CSS classes). Also, this beast autofocus on hover in Tizen - it is pretty annoying (it’s good that I don’t use the mouse on the TV).

ferferga: Worth removing then? Or perhaps just check if the browser is a TV before firing it like what I want to do with ``gamepadtokey.js``?

dmitrylyzo: I think this is also essential for non-TV devices.
From what I noticed (if you disable `mouseManager`):
* The cursor does not disappear (even in browser in non-TV layout) - it’s bad when you watch a video or pictures. In the Tizen emulator, cursor is hided by system itself.
* In TV layout in browser, volume control is displayed (normally it is available only on mouse move) - not so bad, but the slider itself does not focus, so as not to get stuck in it when keyboard navigating.
* In TV layout, when viewing pictures, the navigation buttons do not disappear.

src/components/serverNotifications/serverNotifications.js
dmitrylyzo: at `addEventListener`

src/components/serverNotifications/serverNotifications.js
dmitrylyzo: at `addEventListener`

src/components/serverNotifications/serverNotifications.js
```diff
@@ -197,30 +197,30 @@ define(['connectionManager', 'playbackManager', 'events', 'inputManager', 'focus
...
+            console.log("Gamepad connected! Attaching gamepadtokey.js script");
         }
-
-        return false;
     }
 
-    function isGamepadSupported() {
-        return "ongamepadconnected" in window || navigator.getGamepads || navigator.webkitGetGamepads;
+    function dettachgamepad(e) {
+        delete require.cache[require(["components/serverNotifications/gamepadtokey"])];
```
dmitrylyzo: I didn't tested this PR - just with my eyes. Should remained gamepads be checked before module unloading?

Another option: instead of loading/unloading the module, you can start/stop `runInputLoop`. Then all gamepad-related logic can be moved to `gamepad` module.

dkanada:
@ferferga I'd like to get this pull request in, but it seems that we should keep the mouse manager. If you could also move the gamepad input file to `src/scripts/input` as well that would be excellent.

ferferga:
@dkanada That's already done in my testing branch. I'm trying to get my hands on my friend's Xbox console for checking some stuff before making the commit, although so far so good in desktop and mobile browsers.

ferferga:
Pushed my latest local changes. Still waiting for my friend to test as currently I'm not sure if just adding the handler works for gamepads that are previously connected to the system, or we need to manually check first and then add the handler, see [here](https://github.com/jellyfin/jellyfin-web/pull/746/files#diff-56b9612ccb19d654f1b875a6598c51a0R166).

If he says that the gamepad is connected (probably [he will still face what @MrTimscampi said](https://github.com/jellyfin/jellyfin-web/pull/746#issuecomment-579804242), but unrelated to this PR) and detected successfully, and all reviews are positive, we can merge this.

Note also how I removed the gamepad check in ``gamepadtokey.js`` file, as we expect the script to be ran only when a gamepad is connected, so the check is unnecessary there.

ferferga:
My friend confirmed that the gamepad is detected, even with only the event handler, so seems that no checks are necessary to be performed at start.

src/components/input/gamepadtokey.js
```diff
@@ -254,101 +254,93 @@ require(['apphost'], function (appHost) {
     function runInputLoop() {
         // Get the latest gamepad state.
         var gamepads;
-        if (navigator.getGamepads) {
-            gamepads = navigator.getGamepads();
-        } else if (navigator.webkitGetGamepads) {
-            gamepads = navigator.webkitGetGamepads();
-        }
-        gamepads = gamepads || [];
+        gamepads = navigator.getGamepads();
```
dkanada: `var gamepads = navigator.getGamepads();`

src/components/input/gamepadtokey.js
```diff
@@ -254,101 +254,93 @@ require(['apphost'], function (appHost) {
...
         var i;
         var j;
         var len;
         for (i = 0, len = gamepads.length; i < len; i++) {
             var gamepad = gamepads[i];
-            if (gamepad) {
```
dmitrylyzo: I just tested on PC. I have Xbox Wireless Adapter, and it always reports 4 gamepads, even only one is connected. So this condition is required.

src/components/keyboardnavigation.js
```diff
@@ -126,6 +126,41 @@ define(["inputManager", "layoutManager"], function (inputManager, layoutManager)
...
+    function attachGamepad(e) {
+        if (isGamepadConnected()) {
+            require(["components/input/gamepadtokey"]);
+            console.log("Gamepad connected! Attaching gamepadtokey.js script");
+        }
+    }
+
+    function dettachGamepad(e) {
+        if (!isGamepadConnected()) {
+            delete require.cache[require(["components/input/gamepadtokey"])];
```
dmitrylyzo: `require.cache` is undefined - I got an error, and `gamepadtokey` stays and loops through 4 virtual gamepads endlessly.

src/components/keyboardnavigation.js
```diff
@@ -126,6 +126,41 @@ define(["inputManager", "layoutManager"], function (inputManager, layoutManager)
...
+            console.log("Gamepad disconnected! No other gamepads are connected, dettaching gamepadtokey.js");
+        } else {
+            console.log("Gamepad disconnected! There are gamepads still connected.");
+        }
+    }
+
+    // No need to check for gamepads manually at load time, the eventhandler will be fired for that
+    window.addEventListener("gamepaddisconnected", dettachGamepad);
+    window.addEventListener("gamepadconnected", attachGamepad);
+    require(["components/input/mouseManager"]);
```
dmitrylyzo: Since `mouseManager` is on its own, maybe this line should be moved to `site.js`.
https://github.com/jellyfin/jellyfin-web/blob/bf4207cc3657b28498bdabc8ee7052ab0ada6c2b/src/scripts/site.js#L485-L491

src/components/input/gamepadtokey.js
ferferga: @dmitrylyzo This is why [I made this loop](https://github.com/jellyfin/jellyfin-web/pull/746/files#diff-56b9612ccb19d654f1b875a6598c51a0R129-R141) in ``keyboardnavigation.js``. The ``gamepadlist`` will always return 4 ``gamepad`` objects. All of them will be empty if no gamepad is connected. However, they will be populated once a gamepad is connected.

Afaik, the problem is in ``keyboardnavigation.js``, I will describe it to you in your comment there.

src/components/keyboardnavigation.js
ferferga: This is where the problem you describe above happens. Whenever a gamepad is disconnected, the ``gamepadtokey.js`` should be killed no matter what, so the "check" for connected gamepads in ``gamepadtokey.js`` rely solely in ``keyboardnavigation.js``, which is the responsible for killing the script whenever no gamepads are connected.

This is the most common approach I found for "unrequiring" a previously required module and kill the execution of ``gamepadtokey.js``. Couldn't test it as I have no gamepads, so thank you for pointing that out, will need to investigate that more.

Do you know what's the best way and how I can "unrequire" a module?

src/components/keyboardnavigation.js
ferferga: Hmm, for me no problem, but maybe makes more sense to have all the input controllers in one place?

ferferga: Btw by seeing this... Maybe we can invoke ``gamepadtokey`` just like ``keyboardnavigation`` and ``autofocuser`` are invoked? So, whenever we want to kill it, we can call ``disable()``

Pretty noob with JavaScript packages as you can see, so probably this makes no sense, so any feedback about this is appreciated! :)

src/components/keyboardnavigation.js
dmitrylyzo: I am absolute dumb in "requirejs" :neutral_face: Like you, I found this approach to remove the "required" module. I may be wrong, but problem is that `require` is provided by `amelada` library, so it may not be fully implemented.

As I mention before, if we have no option to remove "required" module, we can start/stop `runInputLoop`: `requestAnimationFrame`/`cancelAnimationFrame`. In this case, we can (or have to) move all gamepad-related code into `gamepadtokey` and just require it in `site.js` (near `keyboardnavigation`).

src/components/keyboardnavigation.js
dkanada: The one place in this case would be `site.js` and the input folder rather than the `keyboardnavigation.js` file.

dmitrylyzo: > Pretty noob with JavaScript packages as you can see

I am not better :stuck_out_tongue: 

> Maybe we can invoke `gamepadtokey` just like `keyboardnavigation` and `autofocuser` are invoked? So, whenever we want to kill it, we can call `disable()`

Actually, these two can be used without dedicated `enable`. I think this makes sense in case of delayed activation and/or external control. Won't do worse, but gamepad does not need this.

ferferga:
@dmitrylyzo Please, test with this new changes. Perhaps a simple return stops the loop until we call ``require`` again in reconnection (or it fails because require can only be called once). Worth doing some trial/error here.

Connect your gamepad to your computer, see ``console`` and check if the gamepadtokey.js has been loaded, a message will be displayed. Once you see it is loaded, go to "Performance" tab and start recording. Go to a movie library (for example) and scroll up and down (with the mouse, no need to use the gamepad). Stop recording and see if gamepadtokey was running in background. Disconnect gamepad and repeat same process. Is it running in the background without the gamepad?

What also if you reconnect it afterwards?

dmitrylyzo:
1. Module is attached on gamepad input (move stick). On connect even no message.
2. `return` stops `runInputLoop` totally because `requestAnimationFrame` (at the end) is not called. Changed to `break`.
3. `runInputLoop` running in background even without gamepad.

ferferga:
@dmitrylyzo 
1. Strange. Did you filter messages from certain modules in your browser's console? However, did you connect before starting Jellyfin or after?
2. ``return`` should only be invoked when a gamepad is not connected. Perhaps is fired by the ``AnimationFrame``, though.
3. After having it connected first, right?
4. Expected. ``gamepadtokey.js`` running with the connected gamepad right?

dmitrylyzo:
>     Strange. Did you filter messages from certain modules in your browser's console? However, did you connect before starting Jellyfin or after?

No filters. Who knows how Linux work with this adapter, but I think this won't cause problems. Gamepad was turned on after start Jellyfin (+disabled cache).

>     `return` should only be invoked when a gamepad is not connected. Perhaps is fired by the `AnimationFrame`, though.

`gamepads` array always has 4 items. One of them filled with connected gamepad. Other are `undefined` - on second item -> return.

>     After having it connected first, right?

Yes. Module is attached properly - on gamepad attach.

>     Expected. `gamepadtokey.js` running with the connected gamepad right?

`gamepadtokey.js` runs single iteration in case of `return` and are not restarted after gamepad reconnect. Or runs forever in case of `break`.

This one is fun:
https://github.com/jellyfin/jellyfin-web/blob/bf4207cc3657b28498bdabc8ee7052ab0ada6c2b/src/components/serverNotifications/gamepadtokey.js#L266
and
https://github.com/jellyfin/jellyfin-web/blob/bf4207cc3657b28498bdabc8ee7052ab0ada6c2b/src/components/serverNotifications/gamepadtokey.js#L289
use same `len` variable, and first loop gets more gamepads after first iteration. This is from the birth of this file. `break` will help a little, but I dunno if `gamepads` array will be without gaps. `continue` will be safer, but `len` should be splitted.

dmitrylyzo:
This starts/stops on connect/disconnect and reconnect :smile: The only drawback I see is that module will be in RAM but "sleep". 
<details>
<summary>Start/stop runInputLoop</summary>

```diff
diff -ruN /home/user/tmp/jellyfin-web~fer/dist/components/input/gamepadtokey.js /usr/lib/jellyfin/bin/jellyfin-web/components/input/gamepadtokey.js
--- /home/user/tmp/jellyfin-web~fer/dist/components/input/gamepadtokey.js	2020-02-07 16:41:48.757287594 +0300
+++ /usr/lib/jellyfin/bin/jellyfin-web/components/input/gamepadtokey.js	2020-02-07 19:25:51.738564756 +0300
@@ -251,13 +251,15 @@
         }
     }
 
+    var inputLoopTimer;
+
     function runInputLoop() {
         // Get the latest gamepad state.
         var gamepads = navigator.getGamepads();
-        for (var i = 0, len = gamepads.length; i < len; i++) {
+        for (var i = 0; i < gamepads.length; i++) {
             var gamepad = gamepads[i];
             if (!gamepad) {
-                return;
+                continue;
             }
             // Iterate through the axes
             var axes = gamepad.axes;
@@ -279,8 +281,7 @@
             }
             // Iterate through the buttons to see if Left thumbstick, DPad, A and B are pressed.
             var buttons = gamepad.buttons;
-            var j;
-            for (j = 0, len = buttons.length; j < len; j++) {
+            for (var j = 0; j < buttons.length; j++) {
                 if (ProcessedButtons.indexOf(j) !== -1) {
                     if (buttons[j].pressed) {
                         switch (j) {
@@ -347,10 +348,50 @@
             }
         }
         // Schedule the next one
-        requestAnimationFrame(runInputLoop);
+        inputLoopTimer = requestAnimationFrame(runInputLoop);
+    }
+
+    function startInputLoop() {
+        if (!inputLoopTimer) {
+            runInputLoop();
+        }
+    }
+
+    function stopInputLoop() {
+        cancelAnimationFrame(inputLoopTimer);
+        inputLoopTimer = undefined;
+    }
+
+    function isGamepadConnected() {
+        var gamepads = navigator.getGamepads();
+        for (var i = 0, len = gamepads.length; i < len; i++) {
+            var gamepad = gamepads[i];
+            if (gamepad) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    function attachGamepad(e) {
+        if (isGamepadConnected()) {
+            console.log("Gamepad connected! Starting input loop");
+            startInputLoop();
+        }
+    }
+
+    function dettachGamepad(e) {
+        if (!isGamepadConnected()) {
+            console.log("Gamepad disconnected! No other gamepads are connected, stopping input loop");
+            stopInputLoop();
+        } else {
+            console.log("Gamepad disconnected! There are gamepads still connected.");
+        }
     }
 
-    runInputLoop();
+    // No need to check for gamepads manually at load time, the eventhandler will be fired for that
+    window.addEventListener("gamepaddisconnected", dettachGamepad);
+    window.addEventListener("gamepadconnected", attachGamepad);
 
     // The gamepadInputEmulation is a string property that exists in JavaScript UWAs and in WebViews in UWAs.
     // It won't exist in Win8.1 style apps or browsers.
diff -ruN /home/user/tmp/jellyfin-web~fer/dist/components/input/keyboardnavigation.js /usr/lib/jellyfin/bin/jellyfin-web/components/input/keyboardnavigation.js
--- /home/user/tmp/jellyfin-web~fer/dist/components/input/keyboardnavigation.js	2020-02-07 16:41:48.705288240 +0300
+++ /usr/lib/jellyfin/bin/jellyfin-web/components/input/keyboardnavigation.js	2020-02-07 19:24:30.915625562 +0300
@@ -126,37 +126,6 @@
         });
     }
 
-    function isGamepadConnected() {
-        var gamepads = navigator.getGamepads();
-        for (var i = 0, len = gamepads.length; i < len; i++) {
-            var gamepad = gamepads[i];
-            if (gamepad) {
-                return true;
-            }
-        }
-        return false;
-    }
-
-    function attachGamepad(e) {
-        if (isGamepadConnected()) {
-            require(["components/input/gamepadtokey"]);
-            console.log("Gamepad connected! Attaching gamepadtokey.js script");
-        }
-    }
-
-    function dettachGamepad(e) {
-        if (!isGamepadConnected()) {
-            console.log("Gamepad disconnected! No other gamepads are connected, dettaching gamepadtokey.js");
-        } else {
-            console.log("Gamepad disconnected! There are gamepads still connected.");
-        }
-    }
-
-    // No need to check for gamepads manually at load time, the eventhandler will be fired for that
-    window.addEventListener("gamepaddisconnected", dettachGamepad);
-    window.addEventListener("gamepadconnected", attachGamepad);
-    require(["components/input/mouseManager"]);
-
     return {
         enable: enable,
         getKeyName: getKeyName
diff -ruN /home/user/tmp/jellyfin-web~fer/dist/scripts/site.js /usr/lib/jellyfin/bin/jellyfin-web/scripts/site.js
--- /home/user/tmp/jellyfin-web~fer/dist/scripts/site.js	2020-02-07 16:41:48.765287496 +0300
+++ /usr/lib/jellyfin/bin/jellyfin-web/scripts/site.js	2020-02-07 19:24:16.899809796 +0300
@@ -485,6 +485,8 @@
                 require(["keyboardnavigation"], function(keyboardnavigation) {
                     keyboardnavigation.enable();
                 });
+                require(["components/input/gamepadtokey"]);
+                require(["components/input/mouseManager"]);
                 require(["focusPreventScroll"]);
                 require(["autoFocuser"], function(autoFocuser) {
                     autoFocuser.enable();

```
</details>

MrTimscampi:
> The only drawback I see is that module will be in RAM but "sleep".

Performance-wise, an unused module does nothing anyway. It'll sit in RAM, but that's not really an issue 99% of the time.

ferferga:
@dmitrylyzo That's not an issue for me as well, it sits in RAM, but will be sitting only if you previously had a gamepad connected, users that don't have gamepads aren't affected by this, and users that do have them will only be affected in sessions where they use them.

Btw, thank you very much for your extensive tests and comments, they're really useful, will get my hands on this again whenever I have some time.

JustAMan:
@ferferga do we have it working now? Is it review-ready?

ferferga:
Sorry for the huge rebase, thought that GitHub would handle it more gracefully, knowing that those commits are already in master)

This should be ready for review now!

dmitrylyzo:
@ferferga your commits were not in master. It seems that you rebased and then merged with your branch again.

ferferga:
@dmitrylyzo Had to do it due to merge conflicts, didn't know a better way without pulling all the newest changes, rebasing them and merging conflicts. Could you please show me how it should be done? I'm not that pro in git 😅.

If that's a problem, perhaps it's a better idea to close this PR and open a new one? Or it can be done just right with the "rebase" merge option?

Sorry again for the mess 😓.

dmitrylyzo:
@ferferga You could do this (_I tried on your branch_)
```sh
git reset HEAD~
```
_you should move to your pre-merge commit with changes of last (merge) commit on `keyboardnavigation.js`_
Then fix this part because there is code dublication
https://github.com/jellyfin/jellyfin-web/blob/ae27e55f477ac943acbe96e45c4f8d2d77891dc0/src/components/input/keyboardnavigation.js#L148-L204
Make new commit (_actually, I removed all of above and new commit will not required_).
Check the beauty of git-tree with some gui.
Force-push.

It will be good if you have a backup to compare with it.

ferferga:
@dmitrylyzo Your suggestions helped me to redo my branch, now it's all cleaned up, thanks!

src/components/input/keyboardnavigation.js
```diff
@@ -145,6 +145,32 @@ define(["inputManager", "layoutManager"], function (inputManager, layoutManager)
             }
         });
     }
+    // For hiding the mouse while playbackç
```
JustAMan:
```suggestion
    // For hiding the mouse while playback
```

src/components/input/keyboardnavigation.js
```diff
@@ -145,6 +145,32 @@ define(["inputManager", "layoutManager"], function (inputManager, layoutManager)
...
+    require(["components/input/mouseManager"]);
+
+    // Gamepad initialization. No script is required if no gamepads are at boot time, saving a bit of resources._10k
```
JustAMan:
```suggestion
    // Gamepad initialisation. No script is required if no gamepads are present at init time, saving a bit of resources.
```

src/components/input/keyboardnavigation.js
```diff
@@ -145,6 +145,32 @@ define(["inputManager", "layoutManager"], function (inputManager, layoutManager)
...
+            var gamepad = gamepads[i];
+            if (gamepad) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    function attachGamepadScript(e) {
+        if (isGamepadConnected()) {
```
JustAMan: why do we want to check that, if that is fired on the `gamepadconnected` event?

src/components/input/keyboardnavigation.js
```diff
@@ -145,6 +145,32 @@ define(["inputManager", "layoutManager"], function (inputManager, layoutManager)
             }
         });
     }
+    // For hiding the mouse while playbackç
+    require(["components/input/mouseManager"]);
```
JustAMan: why is this component not present in `site.js`?

src/components/input/keyboardnavigation.js
ferferga: True, thanks!

src/components/input/keyboardnavigation.js
ferferga: My PR was published before many modules were moved to ``site.js`` I think. Moving this now.

src/components/input/keyboardnavigation.js
```diff
@@ -146,6 +146,17 @@ define(["inputManager", "layoutManager"], function (inputManager, layoutManager)
         });
     }
 
+    // Gamepad initialisation. No script is required if no gamepads are present at init time, saving a bit of resources.
+    // Whenever the gamepad is connected, we hand all the control of the gamepad to gamepadtokey.js by removing the event handler
+    function attachGamepadScript(e) {
+        console.log("Gamepad connected! Attaching gamepadtokey.js script");
+        window.removeEventListener("gamepadconnected", attachGamepadScript);
+        require(["components/input/gamepadtokey"]);
```
JustAMan: oh, wait, would this start input loop? maybe it would only load the script but not start the loop?

ferferga: That's why I need @dmitrylyzo and @MrTimscampi to test it.

dmitrylyzo: @JustAMan you are right.
* when I connect gamepad - no action
* press button "A" - `gamepadtokey.js` loaded, loop does not start
* any input - loop does not start

We have 2 options (both tested):
1. Load `gamepadtokey.js` in `site.js` - as in my variant somewhere above. Module will sleep until gamepad is connected.
2. Start loop automatically (at the very end of `gamepadtokey.js`) - as it was, but keep `startInputLoop` handler in case of reconnect gamepad.

dmitrylyzo: Tried this (not perfect but works):
```js
    function attachGamepad(e) {
        if (isGamepadConnected() && document.hasFocus()) {     <-------
            console.log("Gamepad connected! Starting input loop");
            startInputLoop();
        }
    }

    function dettachGamepad(e) {
        if (!isGamepadConnected() || !document.hasFocus()) {     <-------
            console.log("Gamepad disconnected! No other gamepads are connected, stopping input loop");
            stopInputLoop();
        } else {
            console.log("Gamepad disconnected! There are gamepads still connected.");
        }
    }

    // Event Listeners for any change in gamepads' state.
    window.addEventListener("gamepaddisconnected", dettachGamepad);
    window.addEventListener("gamepadconnected", attachGamepad);
    window.addEventListener("blur", dettachGamepad);     <-------
    window.addEventListener("focus", attachGamepad);    <-------
```
I think there was some issue with response to gamepad input, even if window is not focused.

JustAMan: Cannot we call `attachGamepad()` from `gamepadtokey.js` when requiring the script upon first gamepad attach instead of options 1 and 2 above?

dmitrylyzo: > Cannot we call `attachGamepad()` from `gamepadtokey.js` when requiring the script upon first gamepad attach instead of options 1 and 2 above?

Yes, we can export it, but "2" is simpler. If we add what I posted above (focus-related), then we must run `attachGamepad` instead of `startInputLoop`.

dmitrylyzo:
(Minimal) working changes.
```diff
--- a/src/components/input/gamepadtokey.js	2020-02-28 16:45:59.558504090 +0300
+++ b/src/components/input/gamepadtokey.js	2020-02-28 18:44:45.731141523 +0300
@@ -373,14 +375,14 @@
     }
 
     function attachGamepad(e) {
-        if (isGamepadConnected()) {
+        if (isGamepadConnected() && document.hasFocus()) {
             console.log("Gamepad connected! Starting input loop");
             startInputLoop();
         }
     }
 
     function dettachGamepad(e) {
-        if (!isGamepadConnected()) {
+        if (!isGamepadConnected() || !document.hasFocus()) {
             console.log("Gamepad disconnected! No other gamepads are connected, stopping input loop");
             stopInputLoop();
         } else {
@@ -391,6 +393,8 @@
     // Event Listeners for any change in gamepads' state.
     window.addEventListener("gamepaddisconnected", dettachGamepad);
     window.addEventListener("gamepadconnected", attachGamepad);
+    window.addEventListener("blur", dettachGamepad);
+    window.addEventListener("focus", attachGamepad);
 
     // The gamepadInputEmulation is a string property that exists in JavaScript UWAs and in WebViews in UWAs.
     // It won't exist in Win8.1 style apps or browsers.
@@ -401,4 +405,5 @@
         window.navigator.gamepadInputEmulation = "gamepad";
     }
 
+    attachGamepad();
 });
```

dmitrylyzo:
@ferferga `attachGamepad` at the end of `gamepadtokey.js` is required to start loop on module attach

ferferga:
@dmitrylyzo Thank you very much for your input!

However, is really necessary to call attachGamepad function now that we have also a listener for window focus? Seems redundant to me, take a look that we already load the module when a gamepad is connected (not actually firing the input loop, as I divided the code you first suggested between gamepadtokey and keyboardnavigation), and also whenever the window gets in focus, which should be fired as well when the gamepad is connected (I suppose)?

ferferga:
@dmitrylyzo You were so much faster than me when writing your comment :). Already pushed. Do the final test with your gamepad whenever you can please, so we can merge.

Thanks again for your input!

dmitrylyzo:
Chrome - OK

Firefox:
`gamepads` array after disconnect gamepad:
```
1:  true 
(1) […]
​    0: Gamepad { id: "045e-0291-Xbox 360 Wireless Receiver (XBOX)", index: 0, connected: false, … }
    ​length: 1
​<prototype>: Array []
```
So, need to check `gamepad.connected`.

src/components/input/gamepadtokey.js
```diff
@@ -251,114 +251,150 @@ require(['apphost'], function (appHost) {
...
+    function stopInputLoop() {
+        cancelAnimationFrame(inputLoopTimer);
+        inputLoopTimer = undefined;
+    }
+
+    function isGamepadConnected() {
+        var gamepads = navigator.getGamepads();
+        for (var i = 0, len = gamepads.length; i < len; i++) {
+            var gamepad = gamepads[i];
+            if (gamepad) {
```
dmitrylyzo:
```suggestion
            if (gamepad && gamepad.connected) {
```

ferferga: So I suppose that you checked that the field exists as well in Chrome, right?

dmitrylyzo: Yes. At least, Chrome 79.0.3945.79 has it.
Also, I checked Opera 66.0.3515.115 - OK
Edge does not receive gamepads.

ferferga:
@dmitrylyzo Done! Confirm if everything is alright now whenever you have a chance! :)

src/components/input/gamepadtokey.js
```diff
@@ -347,10 +347,54 @@ require(['apphost'], function (appHost) {
...
+        for (var i = 0, len = gamepads.length; i < len; i++) {
+            var gamepad = gamepads[i];
+            if (gamepad && gamepad.connected) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    function attachGamepad(e) {
```
JustAMan:
```suggestion
    function attachGamepadOrFocus(e) {
```
or something similar, so we don't confuse future devs on why `attachGamepad` is called upon focusing a page...

dmitrylyzo:
At least it works as I expected.
* Chrome - OK (module, input loop, stick, D-pad, A/B buttons)
* Firefox - OK (module, input loop, stick, A/B buttons)

_I have not tested the D-pad before._

In Chrome, gamepad has 16 buttons. D-pad layout: `left`, `right`, `up`, `down` -> 14, 15, 12, 13.
In Firefox, gamepad has 14 buttons. D-pad layout: `left`, `right`, `up`, `down` -> 11, 12, 13, 14.
_This for Linux._

JustAMan:
@dmitrylyzo I feel that fixing gamepad mapping keys should be done in a separate PR.

src/components/input/gamepadtokey.js
ferferga: @JustAMan I suppose you mean ``attachGamepadOnFocus``, right?

ferferga:
@dmitrylyzo Yeah, I agree with @JustAMan . That's not strictly related to the load of the module based on hardware support, which was the main aim of this PR. Wait until this PR is merged to open a new one with your proposal, as if you do it now you will face merge conflicts if mine gets merged before.

I can also open it if you want, as long as you're there helping me out with testing :P. Hopefully I get a gamepad soon =D.

src/components/input/gamepadtokey.js
dmitrylyzo: No, he means that same handler is attached to `attach` and `focus` and `detach` and `blur`. So name of function and its use may confuse devs.

dmitrylyzo: As option for names:
`attachGamepad` -> `tryStartInputLoop`
`detachGamepad` (_with single 't'_) -> `tryStopInputLoop`

On the other hand, `attachGamepad` could be treated as "attach gamepad to document".

dmitrylyzo:
> That's not strictly related to the load of the module based on hardware support, which was the main aim of this PR.

:+1: 

I just stated not to forget. Besides, what if we have different gamepads/joysticks/steeringwheels connected at the same time. Then my approach may not work, and we need some mapping function called for each gamepad (or each gamepad button).

src/components/input/gamepadtokey.js
ferferga: @dmitrylyzo Yeah I know, but ``attachGamepadOrFocus`` doesn't make any sense: with that name seems that the function or attach the gamepad, or focus the page, which is not true at all.
The function has an ``and`` logic, meaning that it runs when a gamepad exists **and** the document is focused. That's why I think that @JustAMan meant ``attachGamepadOnFocus``, because it's executed as long as the document is focused.

About the other names, let's see @JustAMan opinion's.

dkanada: I think it's fine for now, this is a vast improvement over the current behavior.

JustAMan: Maybe then `onFocusOrGamepadAttach`?..

ferferga: @JustAMan But why the ``or``? Both should happen at the same time for i to be run 😅😅.

JustAMan: Because either of those events fires off the callback. It doesn't matter which conditions are expected inside, but rather _when_ it is called, IMHO.

ferferga: I understand your point now. I renamed to ``onFocusOrGamepadAttach`` and ``onFocusOrGamepadDetach``, which I think that it's really clear for upcoming devs who are used to expect to name things expecting the result of the conditions (like) and the events that runs the callback.

ferferga:
This is now completely ready to merge now after applying all the proposed suggestions. @dkanada Whenever you can, please, review this so it can be merged.

dkanada:
For what it's worth, I've been keeping a tab on this pull request :sweat_smile: there were a lot of good changes though, so I kept quiet until now.

ferferga:
@dkanada No worries, there was a lot of conversation over here, so it's normal! =D Thank you very much for merging this, finally another long-standing issue solved!

dmitrylyzo:
In general, #525 is also fixed. The only thing that can be done is option to completely ignore gamepads.

PR #747, opened 2020-01-29
Add placeholder image to detail pages and more placeholder icons

### LINKED ISSUE #18: set default album thumbnail if no cover art is found
This applies to the `itemdetails` view for both mobile and desktop, which show an empty section where the cover art would normally be shown. Ideally it would display the same default thumbnail that shows up for other album views. The same fix could eventually be applied to all media types in `itemdetails` since I believe this is an issue for all of them.

related to #15

MrTimscampi:
**Changes**

This adds a placeholder "image" (it's actually either the `<a>` to edit the image or a `<div>`) that uses  the color and icon from the card views.

The div is square-ish due to the width being dynamic and the height being calculated somewhat on the fly through CSS viewport units (No worries about compatibility, see here: https://caniuse.com/#feat=viewport-units)

The icon is dynamically sized depending on the size of its parent and icons have been added to handle the item types "Movie" and "Series".

**Screenshots**

image

image

image

image

image

**Issues**

Fixes #18


ferferga:
Would you mind also adding one placeholder image to the "Active devices" view while nothing is playing in the dashboard?

Captura

Bond-009:
Does this also work for the image on the actor page? Currently it's just a white plane

MrTimscampi:
> Does this also work for the image on the actor page? Currently it's just a white plane

Didn't think about that, but it does !

image

YouKnowBlom:
> Would you mind also adding one placeholder image to the "Active devices" view while nothing is playing in the dashboard?
> 
> Captura

If there isn't an appropriate icon I think just a darker background on `.sessionNowPlayingInnerContent` would work

Something like this:
image

ferferga:
@YouKnowBlom There isn't an specific one but [here](https://material.io/resources/icons/?style=baseline) there are a couple ones that I think that could be used

* layers
* filter_none
* blur_circular or blur_on
* phonelink_off
* surround_sound
* art_track
* watch_later

Some non-so-serious ones that (however) fit perfectly as well imo: 
* weekend
* spa
* all_inclusive
* explore
* event_seat

YouKnowBlom:
@ferferga good idea, I like blur_on a lot personally.

I might be mistaken here but aren't active devices supposed to have a browser/OS icon instead of that generic one in the top left? Maybe something could be done using those icons.

ferferga:
Mine favourite is weekend one hahaha, although I know it's informal.

MrTimscampi:
@YouKnowBlom @ferferga I think that request would be better suited to a separate PR, since the code is in a completely different place than what's done here :)

That said, these device cards do need a big rework at some point.

ferferga:
@MrTimscampi hmm but I wasn't talking about the devices icons and I think that @YouKnowBlom either in his first comment (he commented that as a side note in the second comment as something that should be improved as well, but not necessarily in this PR)

We were referring to the area filled by poster images when something is playing. When nothing is playing, is completely black and we think that a placeholder icon fits there too.

MrTimscampi:
In my opinion, it'd look too busy to simply put an icon there, with the device icon already there and the info on the card, as well as the lack of contrast between the icon and the overlaid text:

image

It's also not the same code as the one used for the media cards, so I maintain that I don't think it fits in this PR.

ferferga:
@MrTimscampi Hmmm, in the "mockup" I made in my mind, it had some transparency (0.5-0.8 opacity values maybe?), but forgot to mention.

Didn't remember that the codebase there was different from the media cards one, will make my own PR in this case whenever I have some time (if you don't do it before, you're damn fast! :P). That way we can also discuss better, in a separate thread, the pros and cons of this.

MrTimscampi:
> Didn't remember that the codebase there was different from the media cards one, will make my own PR in this case whenever I have some time (if you don't do it before, you're damn fast! :P). That way we can also discuss better, in a separate thread, the pros and cons of this.

For sure, go ahead :) It'll be easier to discuss it in another PR with a visual of what you have in mind :)  
I'm looking at stuff that needs fixing for 10.5.0 and we're trying to get a bunch of stuff done to make a release, so I'm trying to not make too much PRs about new stuff. It's mostly about fixing inconsistencies, making the CI pass again and making sure everything is in working order.

After 10.5.0, my main plan is to go back to the site.js cleanup and the Webpack migration, to hopefully get that done (Learning a lot about Webpack along the way :p )  
I'd like to get us to a point where we can move to SASS/SCSS and TypeScript sooner rather than later. Also ideally get some testing in place to prepare for taking ApiClient out of tree. But we might have to wing that one without any tests :p

JustAMan:
@MrTimscampi you have a merge conflict

PR #751, opened 2020-01-30
Checks if stream URL is successfully copied and displays an appropriate toast

ThibaultNocchi:
**Changes**
Displays an error message in a toast when the browser fails copying the stream URL to the clipboard. It used to say it successfully copied the link.
**Issues**
#557 UX improvement for this issue

### DIFF
--- a/src/components/itemcontextmenu.js
+++ b/src/components/itemcontextmenu.js
@@ -351,14 +351,15 @@ define(["apphost", "globalize", "connectionManager", "itemHelper", "appRouter", 
                     document.body.appendChild(textArea);
                     textArea.focus();
                     textArea.select();
-                    try {
-                        document.execCommand("copy");
-
+                    if (document.execCommand("copy")) {
                         require(["toast"], function (toast) {
                             toast(globalize.translate("CopyStreamURLSuccess"));
                         });
-                    } catch (err) {
+                    } else {
                         console.error("Failed to copy to clipboard");
+                        require(["toast"], function (toast) {
+                            toast(globalize.translate("CopyStreamURLError"));
+                        });
                     }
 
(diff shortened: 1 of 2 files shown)


src/components/itemcontextmenu.js
```diff
@@ -351,14 +351,15 @@ define(["apphost", "globalize", "connectionManager", "itemHelper", "appRouter",
...
                     textArea.select();
-                    try {
-                        document.execCommand("copy");
-
+                    if (document.execCommand("copy")) {
                         require(["toast"], function (toast) {
                             toast(globalize.translate("CopyStreamURLSuccess"));
                         });
-                    } catch (err) {
+                    } else {
```
dkanada: Is there a situation that would actually break the page if this isn't wrapped in a try-catch block? Otherwise, I think this is already good to merge.

ThibaultNocchi: From the [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand), it doesn't seem to ever return an exception here.
If we merge this, shouldn't we rename the PR as it doesn't fix the original problem? Finally, is the way I added a new string the correct one? Thanks!

dkanada: Renaming the pull request sounds good, and the string addition is indeed correct.

ThibaultNocchi: Thanks!

dkanada: No problem!

ThibaultNocchi:
@MrTimscampi I've also noticed this, and I'm looking into it at this issue: #557 
I'm hitting some limits with the Clipboard API (on FF at least) that I explained there.

PR #754, opened 2020-02-01
Clipboard API & fallback method to copy stream URL

ThibaultNocchi:
**Changes**
* Removed deprecated copy method to use Clipboard API
* Added a prompt to let the user copy the URL if the Clipboard API failed

**Issues**
#557 Added workaround to let Firefox users (and other browsers not supporting Clipboard API or not interacting well with it) copy stream URL


thornbill:
Browser support for this looks [really bad](https://caniuse.com/#feat=mdn-api_clipboard_write) right now. Perhaps we should detect if the browser supports this api and fallback to the current method if it does not.

ThibaultNocchi:
You are right, I added back the current method of copying. So now it tries with Clipboard API, then the current method and finally resorts to a prompt. I also updated my branch and it added commits to the PR, I don't know if I did something wrong. Thanks.

src/components/itemcontextmenu.js
```diff
@@ -346,23 +346,28 @@ define(["apphost", "globalize", "connectionManager", "itemHelper", "appRouter",
...
+                        textArea.select();
+                        if (document.execCommand("copy")) {
+                            require(["toast"], function (toast) {
+                                toast(globalize.translate("CopyStreamURLSuccess"));
+                            });
+                        } else {
+                            prompt(globalize.translate("CopyStreamURL"), downloadHref);
+                            require(["toast"], function (toast) {
+                                toast(globalize.translate("CopyStreamURLSuccess"));
+                            });
```
dkanada: Not sure we need a toast if a prompt is already used here.

ThibaultNocchi: You may be right, shall I remove it?

dkanada: Go for it!

ThibaultNocchi: Done!

PR #758, opened 2020-02-02
Fix padding in header

dannymichel:
**Changes**

**Issues**

### DIFF
--- a/src/assets/css/librarybrowser.css
+++ b/src/assets/css/librarybrowser.css
@@ -264,7 +264,7 @@
 
 @media all and (max-width: 84em) {
     .withSectionTabs .headerTop {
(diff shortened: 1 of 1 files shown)


dannymichel:
fix

MrTimscampi:
The screenshot seems to be the same as in #759 , what are the differences between the two PRs ?

Shadowghost:
One for the header text and the other for the logo

MrTimscampi:
The text is intended to be level with the icons, though (Everything should be centered vertically).

Do also note that #737 already fixes the padding and margins around the headerbar element, aside from the logo fixed in #759 .

Here's how it currently looks:

image

Issue #742, opened 2020-01-27
Custom CSS (branding) Not Taking Priority


nvllsvm:
**Describe the bug**
Custom CSS is not taking priority when rendering a page. It briefly flashes, but is then replaced by the original theme's styling.

**To Reproduce**

![image](https://user-images.githubusercontent.com/6550543/73145445-a0229f00-407b-11ea-9777-431112ef09ab.png)
1. Set custom branding 
2. Navigate to home page
3. Notice flicker

**Expected behavior**
Custom CSS takes priority.

**Workaround**
Add `!important` to the property.

**Logs**
![image](https://user-images.githubusercontent.com/6550543/73145548-2939d600-407c-11ea-9de9-15cc1fe57a63.png)

**System (please complete the following information):**
 - OS: Docker
 - Browser: Firefox 72
 - Jellyfin Version: 10.4.3

dkanada:
I assume the CSS is just getting loaded normally. Unless you think we should append `!important` to every rule in the custom CSS then I wouldn't consider this a bug in most cases.

thornbill:
I would agree that this is a bug. The custom css should be applied last so it will properly override the theme without needing to increase the specificity of the custom rules.

YouKnowBlom:
I believe this can be closed now that #763 has been merged

PR #763, opened 2020-02-04
Fix custom css precedence

thornbill:
**Changes**
* Adds support for custom css in native apps with web wrappers
* Injects custom css into the body so it will take precedence over themes

**Issues**
https://github.com/jellyfin/jellyfin-android/issues/153
#742 

sunglasses


YouKnowBlom:
if any reviewers need a theme to test it out, here you go ;)
[hotdogtheme.txt](https://github.com/jellyfin/jellyfin-web/files/4156548/hotdogtheme.txt)

PR #764, opened 2020-02-04
Fix device images on main dashboard page

thornbill:
**Changes**
Device images were falling back to the default for all clients on the main dashboard page due to inconsistent naming of the properties in the device object passed to this function. This allows both cases to be supported.

**Issues**
N/A

doublemint

### DIFF
--- a/src/scripts/imagehelper.js
+++ b/src/scripts/imagehelper.js
@@ -3,7 +3,7 @@ define(["browser"], function (browser) {
 
     function getDeviceIcon(device) {
         var baseUrl = "assets/img/devices/";
-        switch (device.AppName) {
+        switch (device.AppName || device.Client) {
             case "Samsung Smart TV":
                 return baseUrl + "samsung.svg";
             case "Xbox One":
@@ -15,7 +15,7 @@ define(["browser"], function (browser) {
             case "Jellyfin Android":
                 return baseUrl + "android.svg";
             case "Jellyfin Web":
-                switch (device.Name) {
+                switch (device.Name || device.DeviceName) {
                     case "Opera":
                     case "Opera TV":
                         return baseUrl + "opera.svg";


src/scripts/imagehelper.js
```diff
@@ -3,7 +3,7 @@ define(["browser"], function (browser) {
 
     function getDeviceIcon(device) {
         var baseUrl = "assets/img/devices/";
-        switch (device.AppName) {
+        switch (device.AppName || device.Client) {
```
dkanada: the first one is correct

src/scripts/imagehelper.js
```diff
@@ -15,7 +15,7 @@ define(["browser"], function (browser) {
             case "Jellyfin Android":
                 return baseUrl + "android.svg";
             case "Jellyfin Web":
-                switch (device.Name) {
+                switch (device.Name || device.DeviceName) {
```
dkanada: the new one is correct

src/scripts/imagehelper.js
thornbill: Both are used within the web ui though. It’s not a matter of what clients are sending, there is different object types being passed to this function depending on where it’s being called. The section on the main dashboard that lists the current active devices is broke and always uses the default device icon. The dedicated device page works as expected though.

dkanada: Agh that's annoying, I guess we need this for now then. Why can't the endpoints just report the same values.

dmitrylyzo:
I also looked at it and it seemed to me that everything was somewhat turned inside out (I mean `DeviceName` as `AppName`).

This is fallback in case of non-NativeShell apps and browsers:
https://github.com/jellyfin/jellyfin-web/blob/8d02d05441ad630ac4feebe734d2b0b23033dc9c/src/components/apphost.js#L105-L122
i.e. we have
* For Samsung
```
deviceName: 'Samsung Smart TV'
appName: 'Jellyfin Web'   <- In app, I changed this to 'Jellyfin for Tizen'
```
* For LG
```
deviceName: 'LG Smart TV'
appName: 'Jellyfin Web'   <- In app, I changed this to 'Jellyfin for WebOS'
```
* For most
```
deviceName: 'Chrome' / 'Edge' / 'Firefox' / ...
appName: 'Jellyfin Web'
```

I thought about this.
```javascript
switch (device.AppName || device.Client) {
    case "Kodi":
        return baseUrl + "kodi.svg";
    case "Jellyfin Android":
        return baseUrl + "android.svg";
    // "Jellyfin Web" and other
    default:
        switch (device.Name || device.DeviceName) {
            case "Opera":
            case "Opera TV":
                return baseUrl + "opera.svg";
            case "Chrome":
                return baseUrl + "chrome.svg";
            case "Firefox":
                return baseUrl + "firefox.svg";
            case "Edge":
                return baseUrl + "edge.svg";
            case "Internet Explorer":
                return baseUrl + "msie.svg";
            case "Samsung Smart TV":
                return baseUrl + "samsungtv.svg";
            case "Xbox One":
                return baseUrl + "xbox.svg";
            case "Sony PS4":
                return baseUrl + "playstation.svg";
            default:
                return baseUrl + "other.svg";
        }
}
```

thornbill:
I think ideally if we have a dedicated app for a device (Samsung, WebOS, Xbox?) then the app should report a different app name and not use "Jellyfin Web". For devices we don't have a dedicated app for it probably would make sense to move those under "Jellyfin Web".

dmitrylyzo:
> I think ideally if we have a dedicated app for a device (Samsung, WebOS, Xbox?) then the app should report a different app name and not use "Jellyfin Web". For devices we don't have a dedicated app for it probably would make sense to move those under "Jellyfin Web".

True. The above is more universal because it takes into account that someone can open a browser on TV (which also has app), and then `AppName` = `JellyfinWeb`, or we just do not handle name of new app yet.

PR #762, opened 2020-02-04
Adjust details page ribbon margins and colors

MrTimscampi:
**Changes**

Adjusts the top margin of the details page ribbon from 8% to 6.6% to fit the new header bar margins.

Also goes over the colors of the ribbon in the various themes to fix visual issues and make it look more integrated.

image

image

image

image

image

**Issues**

PR #766, opened 2020-02-05
Fix detail page positioning

dannymichel:
Fix this gap

Screen Shot 2020-02-05 at 3 54 06 AM

### DIFF
--- a/src/assets/css/librarybrowser.css
+++ b/src/assets/css/librarybrowser.css
@@ -499,7 +499,7 @@
     align-items: center;
     align-content: center;
     position: sticky;
-    top: 8%;
+    top: 7.7%;
     z-index: 2;
 }
 


MrTimscampi:
@dannymichel This is already covered as part of #762, sorry.

dannymichel:
He’s got the wrong %

dkanada:
Heh @MrTimscampi sounds like you might have got the wrong percent. I feel like it has to do with using percent as a value here though.

dannymichel:
Yeah should be pixels but you guys do your thing. I was just fixing it

dannymichel:
done https://github.com/jellyfin/jellyfin-web/pull/769

jma-odoo:
I'll double check it, but yeah, I think the issue is with it being percent. I simply adjusted the value to fix it on my end (As can be seen in the screenshots on my PR, it's fixed there)

Pixels isn't a solution though (We're generally trying to get rid of pixels as a unit in the source).

As an aside, in this sort of cases where a fix is already awaiting review and/or merging, it's better if you use the review feature of Github to suggest a change on the existing PR.  
It makes it easier for us to keep track of what's awaiting to be merged (since sometimes the merge order matters) and it's generally better form to suggest changes.

dannymichel:
@jma-odoo someone change it to ``3.8em`` then cause if you're doing em thats what it's gonna be

Issue #539, opened 2019-10-17
Impossible to change H264 preset


CTalvio:
It is impossible to change H264 preset, saving does not complete and refreshing simply resets it

**To Reproduce**
1. Go to 'playback'
2. Change "CH264 encoding preset
3. Save (does not complete successfully)
4. Refresh page

**Expected behavior**
Changing the setting should change the setting

**System (please complete the following information):**
 - OS: [Mint 19.1]
 - Browser: [Vivaldi]
 - Jellyfin Version: [e.g. 10.4.0]

nerxses:
i have kinda the same problem also version 10.4.0 i cannot change the H264 encoding preset because it resets itself to "auto" on page reload.

I also noticed that everything i hit the save button my Transcode path changes.... in a weird way

it adds /transcodes to the path everything i hit save so after 5 saves my path looked like that:
/var/lib/jellyfin/transcoding-temp/transcodes/transcodes/transcodes/transcodes/transcodes/transcodes
which should only be
/var/lib/jellyfin/transcoding-temp/transcodes

CTalvio:
@nerxses 
Note that there is already a bug report for the transcoding appending issue [here](https://github.com/jellyfin/jellyfin-web/issues/527).

And yes, it resets to Auto. But for example, if the preset is changed in the config file it will work and other settings besides auto can be used. Though the setting in webUI will still say auto. I would guess this is also true if a preset was applied in a previous working config.

Bottom line, the H264 preset cannot be changed from the webUI settings, while the setting itself is functioning.

mooninite:
On 10.4.3 I cannot get the setting to take effect by writing it to encoding.xml. Also, upon a restart of Jellyfin the encoding.xml file is re-written so the preset setting is erased.

JustAMan:
Could be fixed by https://github.com/jellyfin/jellyfin/pull/1991 (should be available in 10.5.0 when it hits).

PR #767, opened 2020-02-05
Fix encoder preset option not working

nyanmisaka:
**Changes**
Fix encoder preset option not working in jellyfin-web.

**Issues**
Fixes: https://github.com/jellyfin/jellyfin-web/issues/539


dkanada:
@cvium updating them manually usually doesn't cause issues.

PR #769, opened 2020-02-05
Fix detail page

dannymichel:
fix from % to pixels for every browser window size


dannymichel:
someone change it to ``3.8em`` then cause if you're doing em thats what it's gonna be

PR #770, opened 2020-02-05
Disable keyboard navigation for non-TV

### LINKED ISSUE #504: Unable to select saved login+password in login form using arrow

### LINKED ISSUE #682: Left+Right arrow keys don't work on text fields
**Describe the bug**
When a text field has focus, the left and right arrow keys do not change the cursor position like they should. When editing metadata, nothing happens when the keys are pressed. In the settings dashboard, left arrow seems to defocus the field and right arrow refocuses it.

**To Reproduce**
1. Go to any text field, ex editing a media's metadata
2. Enter some text into the field
3. Use the mouse to position the cursor in the middle of the text
4. Attempt to navigate via the left and right arrow keys

**Expected behavior**
Cursor changes position, one character per press.

**Logs**
N/A

**Screenshots**
N/A

 - OS: Arch Linux
 - Browser: Firefox
 - Jellyfin Version: 10.4.3

**Additional context**
None

dmitrylyzo:
I asked about disabling the keyboard on non-TV "between this and then", and I can’t say for sure what we decided.

I look at it like this.

**Desktop layout**
This is mostly for browsers on PC (with mouse and keyboard) - just leave them to do the job in their own way. There is no focus indication, so users will be disoriented.

**Mobile layout**
This is for those with a touch screen - I can hardly imagine using a physical keyboard. And, like previous, there is no focus indication, so users will be disoriented too.

**TV layout**
Most TV users are doomed to use the keyboard/remote - this is our "clients".

**Changes**
Disable keyboard navigation for non-TV.

**Issues**
Partially fixes #504 (at least for `Desktop`/`Mobile` users).
Partially fixes #682 (at least for `Desktop`/`Mobile` users).

For TV users (as I am), I tried (a little luck) to grab keyboard by `emby-input` like [here](https://github.com/jellyfin/jellyfin-web/issues/504#issuecomment-574233711), but this is for another PR.

Also, I heard that you are back to single-quotes. Should I change to single-quotes the touched files? Or touched lines?

### DIFF
--- a/src/components/keyboardnavigation.js
+++ b/src/components/keyboardnavigation.js
@@ -36,6 +36,11 @@ define(["inputManager", "layoutManager"], function (inputManager, layoutManager)
         10252: "MediaPlayPause"
     };
 
+    /**
+     * Keys used for keyboard navigation.
+     */
+    var NavigationKeys = ["ArrowLeft", "ArrowRight", "ArrowUp", "ArrowDown"];
+
     var hasFieldKey = false;
     try {
         hasFieldKey = "key" in new KeyboardEvent("keydown");
@@ -60,11 +65,28 @@ define(["inputManager", "layoutManager"], function (inputManager, layoutManager)
         return KeyNames[event.keyCode] || event.key;
     }
 
+    /**
+     * Returns _true_ if key is used for navigation.
+     *
+     * @param {string} key name
+     * @return {boolean} _true_ if key is used for navigation
+     */
+    function isNavigationKey(key) {
+        return NavigationKeys.indexOf(key) != -1;
+    }
+
     function enable() {
         document.addEventListener("keydown", function (e) {
+            var key = getKeyName(e);
+
+            // Ignore navigation keys for non-TV
+            if (!layoutManager.tv && isNavigationKey(key)) {
+                return;
+            }
+
             var capture = true;
 
-            switch (getKeyName(e)) {
+            switch (key) {
                 case "ArrowLeft":
                     inputManager.handle("left");
                     break;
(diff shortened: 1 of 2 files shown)


MrTimscampi:
> Also, I heard that you are back to single-quotes. Should I change to single-quotes the touched files? Or touched lines?

We're going to handle this in an automated fashion Soon™. So usually, you can keep the same style as the code around where you're editing.

src/components/keyboardnavigation.js
```diff
@@ -60,11 +65,28 @@ define(["inputManager", "layoutManager"], function (inputManager, layoutManager)
...
+     *
+     * @param {string} key name
+     * @return {boolean} _true_ if key is used for navigation
+     */
+    function isNavigationKey(key) {
+        return NavigationKeys.indexOf(key) != -1;
+    }
+
     function enable() {
         document.addEventListener("keydown", function (e) {
```
MrTimscampi: Shouldn't we be adding the event listener only if in TV layout, instead of registering an event simply to perform the check and return ?

dmitrylyzo: At first I did it as you said - just return from `enable`. But then I thought about the possibility of using media buttons in other layouts, therefore cut off only those that interfere. I have no media buttons on my keyboard. Emulated through global hotkeys are not worked - separate buttons are captured. `Pause` works because everyone has it.

Do you suggest to completely disable the keyboard?

MrTimscampi: Fair enough, I hadn't though about media buttons.
We can leave it as-is then.

PR #775, opened 2020-02-07
Alter hardware acceleration message

joshuaboniface:
**Changes**
Don't say this is "experimental" since it's not. Mention the documentation on configuring hardware acceleration and that it requires additional configuration.

**Issues**
N/A

### DIFF
--- a/src/strings/en-us.json
+++ b/src/strings/en-us.json
@@ -652,7 +652,7 @@
     "LabelH264Crf": "H264 encoding CRF:",
     "LabelH264EncodingPreset": "H264 encoding preset:",
     "LabelHardwareAccelerationType": "Hardware acceleration:",
-    "LabelHardwareAccelerationTypeHelp": "This is an experimental feature only available on supported systems.",
+    "LabelHardwareAccelerationTypeHelp": "Hardware acceleration requires additional configuration - please see the documentation at https://docs.jellyfin.org/general/administration/hardware-acceleration.html",
     "LabelHomeNetworkQuality": "Home network quality:",
     "LabelHomeScreenSectionValue": "Home screen section {0}:",
     "LabelHttpsPort": "Local HTTPS port number:",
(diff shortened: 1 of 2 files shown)


src/strings/en-us.json
```diff
@@ -652,7 +652,7 @@
     "LabelH264Crf": "H264 encoding CRF:",
     "LabelH264EncodingPreset": "H264 encoding preset:",
     "LabelHardwareAccelerationType": "Hardware acceleration:",
-    "LabelHardwareAccelerationTypeHelp": "This is an experimental feature only available on supported systems.",
+    "LabelHardwareAccelerationTypeHelp": "Hardware acceleration requires additional configuration - please see the documentation at https://docs.jellyfin.org/general/administration/hardware-acceleration.html",
```
dkanada:
```suggestion
    "LabelHardwareAccelerationTypeHelp": "Hardware acceleration requires additional configuration.",
```

joshuaboniface:
Sure, I've been lazy about it :D

Artiume:
I was thinking we could revamp the remux edition too. And we should introduce remuxing as container swapping only. Direct streaming would be container + Audio 😎

src/strings/en-us.json
```diff
@@ -652,7 +652,7 @@
     "LabelH264Crf": "H264 encoding CRF:",
     "LabelH264EncodingPreset": "H264 encoding preset:",
     "LabelHardwareAccelerationType": "Hardware acceleration:",
-    "LabelHardwareAccelerationTypeHelp": "This is an experimental feature only available on supported systems.",
+    "LabelHardwareAccelerationTypeHelp": "Hardware acceleration requires additional configuration.",
```
thornbill: Should copy changes like this not be handled with weblate?

JustAMan: What's going to be handled with Weblate? This is _source_ language...

Artiume: well, we keep doing something to break weblate. it isn't us making changes to these strings on git rather than weblate?

thornbill: Looks like source strings can be edited with Weblate unless I am missing something.

Screenshot_2020-02-18 Jellyfin Jellyfin Web - English (United States)

Just trying to avoid conflicts due to changes being made in Weblate **and** via PRs although that is probably less likely to happen in source language files.

dkanada: Oh I thought we had the source languages locked. This shouldn't cause any issues but I'll make an account over there just in case.

PR #745, opened 2020-01-27
Cleanup site.js

MrTimscampi:
**Changes**

Moving this out of #681 to ease review and work on that PR.

It's a general cleanup of site.js with some refactoring along the way. Due to changes in how the module names are shortened, a few other files had to be adapted.

**Issues**

None

PR #780, opened 2020-02-09
Move confirm, prompt and chromecast modules definitions outside of site.js

MrTimscampi:
**Changes**

Per request of @dkanada , this is moved out of #745.

**Issues**

None


src/components/prompt/browserprompt.js
```diff
@@ -0,0 +1,104 @@
...
+        dlg.querySelector('form').addEventListener('submit', function (e) {
+
+            submitValue = dlg.querySelector('#txtInput').value;
+            e.preventDefault();
+            e.stopPropagation();
+
+            // Important, don't close the dialog until after the form has completed submitting, or it will cause an error in Chrome
+            setTimeout(function () {
+                dialogHelper.close(dlg);
+            }, 300);
```
JustAMan: can we use `}, 0);` here?

src/components/prompt/browserprompt.js
```diff
@@ -0,0 +1,104 @@
...
+                scrollHelper.centerFocus.off(dlg.querySelector('.formDialogContent'), false);
+            }
+
+            var value = submitValue;
+
+            if (value) {
+                return value;
+            } else {
+                return Promise.reject();
+            }
```
JustAMan:
```js
if (submitValue) {
    return submitValue;
} else {
    return Promise.reject();
}
```
?

src/components/prompt/browserprompt.js
dkanada: This seems like a hack to make sure the submission process finishes first.

src/components/prompt/browserprompt.js
MrTimscampi: Yep, I still need to go back on this and do a bit more (The chromecast bit isn't done).

I was sort of waiting on 10.5.0... :p

src/components/prompt/browserprompt.js
JustAMan: But that 300ms is an arbitraty thing, a server could be slower.
I was hoping it was just a technique to make this function execute at next possible timeframe (like normal `setTimeout(foo, 0)` does)

src/components/prompt/browserprompt.js
JustAMan: I'm just thinking of "why save a value to just return it - why extra local variable"?

MrTimscampi:
> Can you describe why you want to move the checks outside of `site.js`?

Good practice (and Webpack, from my tests) requires on `define()` statement per file.

The way it works currently is that it dynamically defines what these modules are depending on the environment. My goal is to make it have only one define() while keeping the different module content depending on the environment.

MrTimscampi:
Summary of the changes:

Prompt, Confirm and SendCastAPI were all moved into one module each.  
Originally, these were conditionally defined in site.js. The previous approach of returning a require() didn't work, so I merged the modules, as they were rather small and serve the same function in different cases.

Prompt and Confirm were tested (Deletion confirmation dialog and Send Message prompt), but I don't have anything to test SendCastAPI with.

PR #785, opened 2020-02-09
Update Rotten Tomatoes icons

MrTimscampi:
**Changes**

Brings the icons up to date with the current Rotten Tomatoes designs. Replaces PNG icons with SVG ones to allow better scaling.

image

image

**Issues**

None


MrTimscampi:
> What is the license for those assets?

Same as those we used before: https://www.rottentomatoes.com/help_desk/licensing

It's merely a refresh of the icons to use the new ones.

JustAMan:
So it says:

> If you would like to use any of our trademarks or logos, please submit a request with details of your planned usage via our Business Proposal Form for review and access to assets.

I am unaware of us receiving their permission... we probably should, pinging @anthonylavado

anthonylavado:
Where do we even get these scores? Is it from them?

PR #786, opened 2020-02-09
Fix slider positioning on click and (try to) make it look the same in different browsers

dmitrylyzo:
Option 1 from https://github.com/jellyfin/jellyfin-web/pull/644#issuecomment-583559307

**Changes**
* Make pointer click pass through.
* Fix font on `emby-input` (details below)
* Fix slider bubble range
* Fix progress bar range
* Make slider in Edge same as in Chrome
* Make slider in Firefox same as in Chrome
* Fix progress bar position on WebOS 3
* Limit slider bubble position to its track
* Fix vertical position of slider progress bar

**Font size**
In Firefox and Chrome, `font-size` goes from browser ignoring `fonts.css`. This makes hard to align progress bar and bubble to slider range. I use `font-size: inherit`, and this helps. But I am not sure that this will work everywhere.
In Edge, font goes from `fonts.css` as expected.

**Screenshots**
New slider structure (add bubble track)
slider-structure
yellowish background - slider container
red background - slider (`input` itself)
red line - `sliderBubble` track

Firefox before
slider1-firefox-old
Firefox after
slider1-firefox-new

Chrome before
slider1-chrome-old
Chrome after
slider1-chrome-new

Edge before
slider1-edge-old
Edge after
slider1-edge-new

On TV, thumb became much bigger :smile: It was small due to uninherited font size.

Firefox TV before
slider1-firefox_tv-old
Firefox TV after
slider1-firefox_tv-new

Chrome TV before
slider1-chrome_tv-old
Chrome TV after
slider1-chrome_tv-new

On iPad Pro 11, looks OK.
IE11 does not load - stop loading on random file with no errors :confused: 
Opera 66.0.3515.72 looks OK.

**To discuss**
- [x] 1. Bubble has no limits:
simplescreenrecorder7-2020-02-09_15 49 27 mkv
I can fix it (with JS, but maybe someone know how it could be fixed with plain CSS)
simplescreenrecorder7-2020-02-09_15 50 23 mkv
Does it look good when bubble changes its size (being clamped)?
simplescreenrecorder7-2020-02-09_15 51 31 mkv

- [x] 2. I left condition `supportsNativeProgressStyle`. Should I remove it?

- [x] 3. There is historical class `mdl-slider-ie-container`. No reference. I could switch to it instead of new `slider-browser-edge`. Or just remove it.

- [x] 4. I can set `transition: 0.2s` like `emby-button` has. There are much more "transitions" - historical.

- [x] 5. There is `mdl-slider-hoverthumb`. It is added for non-Mobile. Not sure about its purpose. We could use it to make thumb non-scalable on Mobile.

- [x] 6. `mdl-slider-background-flex-container` and `mdl-slider-background-flex` have negative `margin-top` (-0.13em in total). To be mathematically "perfect" it must be -0.1em. Also, this will be good for TV layout since it has bigger font. But due to rounding errors we will get:
Firefox
slider1-firefox-new2
Chrome - no change
Edge
slider1-edge-new2
Firefox TV
slider1-firefox_tv-new2
Chrome TV - thicker volume bar
slider1-chrome_tv-new2

### DIFF
--- a/src/elements/emby-slider/emby-slider.js
+++ b/src/elements/emby-slider/emby-slider.js
@@ -48,8 +48,13 @@ define(['browser', 'dom', 'layoutManager', 'keyboardnavigation', 'css!./emby-sli
     function updateBubble(range, value, bubble, bubbleText) {
 
         requestAnimationFrame(function () {
+            var bubbleTrackRect = range.sliderBubbleTrack.getBoundingClientRect();
+            var bubbleRect = bubble.getBoundingClientRect();
 
-            bubble.style.left = value + '%';
+            var bubblePos = bubbleTrackRect.width * value / 100;
+            bubblePos = Math.min(Math.max(bubblePos, bubbleRect.width / 2), bubbleTrackRect.width - bubbleRect.width / 2);
+
+            bubble.style.left = bubblePos + 'px';
 
             if (range.getBubbleHtml) {
                 value = range.getBubbleHtml(value);
--- a/src/elements/emby-slider/emby-slider.js
+++ b/src/elements/emby-slider/emby-slider.js
@@ -84,6 +84,9 @@ define(['browser', 'dom', 'layoutManager', 'keyboardnavigation', 'css!./emby-sli
         if (browser.noFlex) {
             this.classList.add('slider-no-webkit-thumb');
         }
+        if (browser.edge || browser.msie) {
+            this.classList.add('slider-browser-edge');
+        }
         if (!layoutManager.mobile) {
             this.classList.add('mdl-slider-hoverthumb');
         }
--- a/src/elements/emby-slider/emby-slider.js
+++ b/src/elements/emby-slider/emby-slider.js
@@ -75,7 +75,8 @@ define(['browser', 'dom', 'layoutManager', 'keyboardnavigation', 'css!./emby-sli
     function updateValues() {
 
         // Do not update values when dragging with keyboard to keep current progress for reference
-        if (!!this.keyboardDragging) {
+        // Do not update values when touched to keep current progress for reference
+        if (!!this.keyboardDragging || !!this.touched) {
             return;
         }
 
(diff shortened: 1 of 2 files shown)


JustAMan:
@MrTimscampi @dkanada feedback?

dkanada:
1. Why isn't the bubble showing up when you drag it along the slider in those videos?

2. It depends on what the value is for, but it sounds like you might be able to remove it.

3. I'm always a fan of removing old classes or replacing them with better names.

4. Definitely a good idea.

5. I'm guessing that was the intent, and it also sounds like a good idea.

6. The photos are a bit unclear, are you referencing the element size?

dmitrylyzo:
>     1. Why isn't the bubble showing up when you drag it along the slider in those videos?

"Bubble" is box with time, not the thumb. Just hover, no mouse press.

>     2. It depends on what the value is for, but it sounds like you might be able to remove it.

It was used only for Firefox to use native slider progress. But in this case there was no buffered progress.

>     6. The photos are a bit unclear, are you referencing the element size?

Progress line of sliders goes lower a bit. More noticeable on TV due to larger font (`em` units). _You can open "after"-shots and shots from 6._

dkanada:
1. I like this change.

2. Sounds like we can remove it then.

6. I vote to make it perfect.

dmitrylyzo:
Looked deeper for `mdl-slider-hoverthumb`.
* It was used to make thumb scaled up on hover.
* But it was intended that on mobile the thumb must always be scaled up. Or, more precise, on non-mobile the thumb was scaled down in its normal state.

At current moment, on iPhone/iPad, clicking on progress bar does not work, and I have to grab the thumb, but it is pretty small.

JustAMan:
@dmitrylyzo what makes it a draft now? What's not ready?

dmitrylyzo:
> what makes it a draft now? What's not ready?

Only `mdl-slider-hoverthumb`. Look my message above.

dkanada:
Ler's keep that class then and use it for the mobile layout.

dkanada:
Let's keep that class then and use it for the mobile layout.

dmitrylyzo:
_Maybe last two commits are too much for one PR_

* Make `mdl-slider-hoverthumb` as "hover enabled".

* Now, in theory, `emby-slider` may have a range other than [0, 100].
I tried [-100, 100], [50, 100]
I had to add `toLocaleString` for a more adequate display of numbers like 10.000000000005.

* I added HACK to fix [issue](https://github.com/jellyfin/jellyfin-web/pull/786#issuecomment-585610977) on iPhone/iPad: catch `touchstart` to set slider value.
Tested in Safari/Chrome on iPhone 5S, Chrome iPad Pro 11.
Unfortunately this does not work in Safari on iPad Pro 11 due to an incorrect browser detection - not detected as iOS.
In Android Emulator, this hack is not required.

src/elements/emby-slider/emby-slider.js
```diff
@@ -17,10 +16,67 @@ define(['browser', 'dom', 'layoutManager', 'keyboardnavigation', 'css!./emby-sli
...
+        var fraction = valueRange !== 0 ? (value - range.min) / valueRange : 0;
+        return Math.min(Math.max(fraction, 0), 1);
+    }
+
     function updateValues() {
 
         // Do not update values when dragging with keyboard to keep current progress for reference
-        if (!!this.keyboardDragging) {
+        // Do not update values when touched to keep current progress for reference
+        if (!!this.keyboardDragging || !!this.touched) {
```
dkanada: we need to add an option for this

dmitrylyzo: Is this what you are looking for?
simplescreenrecorder7-2020-02-14_19 44 13 mkv

dkanada: We had to add several hacks for the volume slider to update during the drag events, but I have no clue where those changes were made at this point.

dmitrylyzo: I found some of them.
f948541e683087400c5e0698dabffd6453698679

dkanada: There are three volume sliders in the codebase, and I think we had to revisit the custom solution and add a timeout to prevent spamming the server with volume changes during remote control sessions.

Issue #589, opened 2019-11-18
Stop Playback/Fullscreen like Youtube


platinum55:
currently when doubleclicking the movie, for fullscreen, it stops for a moment... youtube handles it likes this:

when a click is recognized, it waits for 500ms - if no other click followed, then the playback stops. if a second click follows within 500ms, then fullscreen is triggered.

would be nice if someone could implement this behavior into jellyfin

JustAMan:
Major feature requests are accepted here: https://features.jellyfin.org/
This one doesn't look major, plus it belongs to a Web UI, so I transferred it.

PR #787, opened 2020-02-09
Add delay on play/pause to prevent hiccup when entering/exiting fullscreen

YouKnowBlom:
**To Discuss**
I found 300ms to be the optimal delay for me but some might like a longer delay

**Changes**
Adds a 300ms delay before pause/unpause when left-clicking on the player. This allows for uninterrupted playback when toggling fullscreen. Does not affect pausing via the OSD button or media keys.

**Issues**
Resolves: #589

### DIFF
--- a/src/controllers/playback/videoosd.js
+++ b/src/controllers/playback/videoosd.js
@@ -1230,7 +1230,7 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
                     shell.disableFullscreen();
                 });
 
-                if(playerPauseClickTimeout) {
+                if (playerPauseClickTimeout) {
                     clearTimeout(playerPauseClickTimeout);
                 }
                 var player = currentPlayer;
@@ -1392,9 +1392,9 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
 
                 case "mouse":
                     if (!e.button) {
-                        if(playerPauseClickTimeout){
+                        if (playerPauseClickTimeout) {
                             clearTimeout(playerPauseClickTimeout);
-                            playerPauseClickTimeout = 0;   
+                            playerPauseClickTimeout = 0;
                         } else {
                             playerPauseClickTimeout = setTimeout(() => {
                                 playbackManager.playPause(currentPlayer);
--- a/src/controllers/playback/videoosd.js
+++ b/src/controllers/playback/videoosd.js
@@ -1230,6 +1230,9 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
                     shell.disableFullscreen();
                 });
 
+                if(playerPauseClickTimeout) {
+                    clearTimeout(playerPauseClickTimeout);
+                }
                 var player = currentPlayer;
                 view.removeEventListener("viewbeforehide", onViewHideStopPlayback);
                 releaseCurrentPlayer();
@@ -1369,6 +1372,7 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
             destroySubtitleSync();
         });
         var lastPointerDown = 0;
+        let playerPauseClickTimeout;
         dom.addEventListener(view, window.PointerEvent ? "pointerdown" : "click", function (e) {
             if (dom.parentWithClass(e.target, ["videoOsdBottom", "upNextContainer"])) {
                 return void showOsd();
@@ -1388,8 +1392,16 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
 
                 case "mouse":
                     if (!e.button) {
-                        playbackManager.playPause(currentPlayer);
-                        showOsd();
+                        if(playerPauseClickTimeout){
+                            clearTimeout(playerPauseClickTimeout);
+                            playerPauseClickTimeout = 0;   
+                        } else {
+                            playerPauseClickTimeout = setTimeout(() => {
+                                playbackManager.playPause(currentPlayer);
+                                showOsd();
+                                playerPauseClickTimeout = 0;
+                            }, 300);
+                        }
                     }
 
                     break;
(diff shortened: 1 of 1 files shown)


src/controllers/playback/videoosd.js
```diff
@@ -1230,6 +1230,9 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
                     shell.disableFullscreen();
                 });
 
+                if(playerPauseClickTimeout) {
```
anthonylavado:
```suggestion
                if (playerPauseClickTimeout) {
```

src/controllers/playback/videoosd.js
```diff
@@ -1388,8 +1392,16 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
 
                 case "mouse":
                     if (!e.button) {
-                        playbackManager.playPause(currentPlayer);
-                        showOsd();
+                        if(playerPauseClickTimeout){
```
anthonylavado:
```suggestion
                        if (playerPauseClickTimeout) {
```

src/controllers/playback/videoosd.js
```diff
@@ -1388,8 +1392,16 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
...
+                            clearTimeout(playerPauseClickTimeout);
+                            playerPauseClickTimeout = 0;
```
anthonylavado:
```suggestion
                            playerPauseClickTimeout = 0;
```

src/controllers/playback/videoosd.js
```diff
@@ -1230,6 +1230,9 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
                     shell.disableFullscreen();
                 });
 
+                if (playerPauseClickTimeout) {
+                    clearTimeout(playerPauseClickTimeout);
```
thornbill: The declaration `let playerPauseClickTimeout;` should happen before it's usage here. Also it's not necessary to check for it to have a value, `clearTimeout()` will not through an exception on an invalid input. https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/clearTimeout#Notes

src/controllers/playback/videoosd.js
```diff
@@ -1388,8 +1392,16 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
 
                 case "mouse":
                     if (!e.button) {
-                        playbackManager.playPause(currentPlayer);
-                        showOsd();
+                        if (playerPauseClickTimeout) {
```
thornbill: Again here this check shouldn't be necessary.

src/controllers/playback/videoosd.js
```diff
@@ -1388,8 +1392,16 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
...
+                            clearTimeout(playerPauseClickTimeout);
+                            playerPauseClickTimeout = 0;
+                        } else {
+                            playerPauseClickTimeout = setTimeout(() => {
+                                playbackManager.playPause(currentPlayer);
+                                showOsd();
+                                playerPauseClickTimeout = 0;
```
thornbill: This shouldn't be required either. The ids should not be reused, so we shouldn't need to reset it when complete. https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout#Return_value

YouKnowBlom: oh wow, I've been living a lie lol. Thanks for letting me know thornbill!

src/controllers/playback/videoosd.js
YouKnowBlom: It would still need a way to check if the user has clicked though. Is there a better way?

YouKnowBlom: @thornbill

dmitrylyzo: `touch` has something similar
https://github.com/jellyfin/jellyfin-web/blob/637480230959217fade1c90542357bf0939f5b61/src/controllers/playback/videoosd.js#L1385-L1389

dmitrylyzo: Actually, I got almost the same
```js
case "mouse":
    if (!e.button) {
        if (now - lastPointerDown < 300) {
            clearTimeout(playerPauseClickTimeout);
        } else {
            playerPauseClickTimeout = setTimeout(function () {
                showOsd();
            }, 300);
        }
        lastPointerDown = now;
    }

    break;
 ```

File: src/controllers/playback/videoosd.js [hunk 7]
```diff
@@ -1224,12 +1224,14 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
             return null;
         }
 
+        let playPauseClickTimeout;
```
JustAMan: are we aware that this breaks older browsers like WebOS? :) they don't handle `let`

YouKnowBlom:
Should I open a new pr without let and arrow functions?

src/controllers/playback/videoosd.js
```diff
@@ -1388,8 +1390,16 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
 
                 case "mouse":
                     if (!e.button) {
-                        playbackManager.playPause(currentPlayer);
-                        showOsd();
+                        if (playPauseClickTimeout) {
+                            clearTimeout(playPauseClickTimeout);
+                            playPauseClickTimeout = 0;
+                        } else {
+                            playPauseClickTimeout = setTimeout(() => {
```
dmitrylyzo: `=>` is  also not supported by WebOS 3

dmitrylyzo:
@YouKnowBlom could you also fix `=>` in `homesections.js`?

YouKnowBlom:
Will do👍

JustAMan:
Would be good if it's easy.

PR #797, opened 2020-02-13
Fix some minor style issues

dkanada:


src/components/cardbuilder/card.css
```diff
@@ -371,12 +371,12 @@ button::-moz-focus-inner {
 }
 
 .cardImageIcon {
-    font-size: 5em;
+    font-size: 5em !important;
```
MrTimscampi: This is okay for now, but we should look into a proper fix once we re-organize the CSS. I'm not too fond of fixing things with the CSS equivalent of a nuclear bomb :p

dkanada: I agree, this was mainly to fix it for the release.

PR #801, opened 2020-02-14
Update some strings

dkanada:


Artiume:
Think this'll impact weblate?

JustAMan:
@dkanada how can this be fixing #804 if all I see is a lot of removed messages?..

dkanada:
The `update` part of the title involves adding a missing string to the source language and removing any strings that don't exist in the source language.

JustAMan:
> The `update` part of the title involves adding a missing string to the source language

@dkanada 
But I don't see any additions to `en-us.json`?..

dkanada:
@Artiume I added a basic restart warning to the UPnP section as well, I didn't want to make drastic changes for the generic restart message yet.

Artiume:
Yeah, that can be a project for another day. You want me to cancel my other PR? :P

src/strings/en-us.json
```diff
@@ -1223,6 +1223,7 @@
     "PlayFromBeginning": "Play from beginning",
     "PlayNext": "Play next",
     "PlayNextEpisodeAutomatically": "Play next episode automatically",
+    "PlaybackErrorNoCompatibleStream": "There are no available formats that are compatible with this device.",
```
JustAMan: Probably suggest a user to look for errors converting? As "no streams" should never actually be seen, it should either be "transcoding required but disabled" or "transcoding failed"...

src/components/libraryoptionseditor/libraryoptionseditor.js
```diff
@@ -107,7 +107,7 @@ define(["globalize", "dom", "emby-checkbox", "emby-select", "emby-input"], funct
         if (!plugins.length) return html;
 
         html += '<div class="metadataFetcher" data-type="' + availableTypeOptions.Type + '">';
-        html += '<h3 class="checkboxListLabel">' + globalize.translate("LabelTypeMetadataDownloaders", availableTypeOptions.Type) + "</h3>";
+        html += '<h3 class="checkboxListLabel">' + globalize.translate("LabelTypeMetadataDownloaders", globalize.translate(availableTypeOptions.Type)) + "</h3>";
```
thornbill: This does not look right...

thornbill: Okay nevermind I think this is fine, but it looks really odd to translate the fallback value also... do you also need a fallback for the translation of the fallback? :joy:

dkanada: You'd think that, but the type is also a translatable string, which was causing the issue for music libraries. Shows and movies were only one word so they never looked odd but music types were always untranslated.

PR #805, opened 2020-02-14
Add Space-Key as PlayPause-Key

screeny05:
This PR adds the spacebar as a key to toggle playback in the videoplayer.

**Changes**
Add additional key-shortcut in videoplayer

### DIFF
--- a/src/controllers/playback/videoosd.js
+++ b/src/controllers/playback/videoosd.js
@@ -1122,6 +1122,7 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
                     }
                     break;
                 case "k":
+                case "Space":
                     playbackManager.playPause(currentPlayer);
                     showOsd();
                     break;


dmitrylyzo:
In TV layout, `Space` triggers focused buttons.
It could be ignored here
https://github.com/jellyfin/jellyfin-web/blob/3c49f10b4061d80f06d13963b6242a442c96f620/src/controllers/playback/videoosd.js#L1107

At current moment, `Space` is handled when OSD is hidden.
https://github.com/jellyfin/jellyfin-web/blob/3c49f10b4061d80f06d13963b6242a442c96f620/src/controllers/playback/videoosd.js#L1101

src/controllers/playback/videoosd.js
```diff
@@ -1122,6 +1122,7 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
                     }
                     break;
                 case "k":
```
MrTimscampi: Not directly related to this PR, but why is this K instead of something sensible like P ? (I get that JKL are homerow, next to each other and all, but for discovery by most users, K makes no sense)

dmitrylyzo: As on Youtube

dkanada: If is was going to be something else it would just be the arrow keys.

JustAMan: Because `vim`?..

dmitrylyzo:
Just tested this PR changes on master:
* When OSD is visible, focus is on button `Play/Pause`, and we get playPause + `click` on button `Play/Pause` (_playPause again_). So this does not work as expected.
* The only way to make it work is to first click on videoplayer. In this case, focus will be on player instead of button.

src/controllers/playback/videoosd.js
MrTimscampi: I had no idea Youtube used that :p

Nevermind, then.

dkanada:
@screeny05 did you have plans to fix the issues mentioned earlier?

dkanada:
I'm closing this for now from lack of movement, but I agree there are some inconsistencies that need to be resolved with the space key. Feel free to open this back up again if you plan to fix the compatibility issues mentioned above!

PR #810, opened 2020-02-15
WIP-Add home horizontal view for tv

grafixeyehero:


grafixeyehero:
tv-home

dkanada:
There seem to be a few lint issues with the changes.

grafixeyehero:
> 
> 
> There seem to be a few lint issues with the changes.

i run fix lint but i get lint problem what should i do to fix

MrTimscampi:
Here's the error:

```
/home/vsts/work/1/s/src/controllers/tvhome/home.js
  8:16  error  'html' is assigned to itself  no-self-assign
```

src/controllers/tvhome/home.js
```diff
@@ -5,7 +5,7 @@ define(['connectionManager', 'loading', 'globalize', 'tabbedPage', 'libraryMenu'
 
         var homeScrollContent = page.querySelector('.contentScrollSlider');
 
-        html = html;
+        self.html = html;
```
MrTimscampi: Why?

grafixeyehero: i think i should remove this line ` self.html = html;`?

MrTimscampi: `self.html` isn't used anywhere in the file, so I'm struggling to understand what you were trying to do :)

grafixeyehero: i mean like this https://github.com/jellyfin/jellyfin-web/pull/810/commits/57e4287c8301be8335de4f1c9966ca093736a8f9

dkanada:
Shoot, did I merge something to cause more conflicts? We should get this in next because it's been here a while.

Larvitar:
I've tried running this in Tizen OS (using emulator) and it's working correctly but it does not seem to be optimized for HighRes (4K)
tizen-screen-horizontal

MrTimscampi:
@grafixeyehero I updated the branch from master, to help a little with the review (We've enforces semicolons an a few other places have been cleaned up since this was opened, which is why there were conflicts).

grafixeyehero:
@MrTimscampi no luck for me https://github.com/jellyfin/jellyfin-web/issues/868#issuecomment-613824335.

src/layouts/tv/home/views.livetv.html
```diff
@@ -7,7 +7,7 @@
                     <div class="cardPadder-square"></div>
                     <div class="cardContent cardContent-shadow">
                         <div class="cardImageContainer coveredImage">
-                            <i class="cardImageIcon material-icons">menu</i>
+                            <em class="cardImageIcon material-icons">menu</em>
```
MrTimscampi, line 10: Don't use `<em>` for icons, use `<span>`.

Sonarqube's parser doesn't know about icons, so they're recommending the wrong element for this case.

The change is meant to better support screen readers, which treat emphasis differently than normal text. What we want to do is inform these screen readers that the element is "a bunch of content" that isn't really anything specific.

Then we describe it more through `aria-label` and `aria-hidden`. The first one is used in cases where the icon represents an action to be performed (for example an "Add" button). The second one is to be used when an element is decorative (So not necessary for the comprehension of the screen).

grafixeyehero:
#1350

PR #812, opened 2020-02-15
Enhance volume slider

dmitrylyzo:
_Requested by_ @dkanada

**Changes**
* Add `data-slider-keep-progress` attribute. At current moment, this attribute is set for position sliders.
* Make delayed volume update (remotecontrol, nowplayingbar).

`data-slider-keep-progress="true"` freezes slider progress update as it was before on position sliders (i.e. keep progress as reference). To be as on "after" GIF, need to set `data-slider-keep-progress="false"` or leave it unset.

**Screenshots**
Before
simplescreenrecorder7-2020-02-15_13 44 55 mkv
After
simplescreenrecorder7-2020-02-15_13 43 14 mkv

At current moment, volume sliders are observed by `mousemove`/`touchmove` events. I think this is for IE11 - no 'input' event. I tried to listen only 'input' - works in Chrome/Firefox.


dkanada:
The VideoOsd page might also have custom changes to the volume slider.

dmitrylyzo:
> The VideoOsd page might also have custom changes to the volume slider.

I removed external update in first commit. Delayed update is already there (maybe not so pretty).
https://github.com/jellyfin/jellyfin-web/blob/bb1ba5a29ed28222ef49b4d314b830a9289bc45c/src/controllers/playback/videoosd.js#L1411-L1440

src/components/nowplayingbar/nowplayingbar.js
```diff
@@ -405,10 +419,6 @@ define(['require', 'datetime', 'itemHelper', 'events', 'browser', 'imageLoader',
             muteButton.querySelector('i').innerHTML = 'volume_up';
         }
 
-        if (progressElement) {
-            progressElement.style.width = (volumeLevel || 0) + '%';
-        }
-
```
JustAMan: why this is no longer needed?

dmitrylyzo: Slider updates itself: I call `updateValues` in `input` event handler (if `data-slider-keep-progress` is not `true`).

dkanada:
> this attribute is set for position sliders

Why not enable for all sliders?

dkanada:
Any chance you could update the VideoOSD with a setVolumeDelayed method as well?

dmitrylyzo:
> Why not enable for all sliders?

Others do not need it.
_I added description in first message_

> Any chance you could update the VideoOSD with a setVolumeDelayed method as well?

Will do

PR #817, opened 2020-02-16
Add basic web client configuration

dkanada:


src/scripts/settings/webSettings.js
```diff
@@ -0,0 +1,30 @@
+define(['appStorage', 'events'], function (appStorage, events) {
+    'use strict';
+
+    function readConfig(path, callback) {
+        var file = new XMLHttpRequest();
```
thornbill: Should use fetch and promises instead of callbacks.

src/components/skinManager.js
```diff
@@ -1,4 +1,4 @@
-define(['apphost', 'userSettings', 'browser', 'events', 'pluginManager', 'backdrop', 'globalize', 'require', 'appSettings'], function (appHost, userSettings, browser, events, pluginManager, backdrop, globalize, require, appSettings) {
+define(['apphost', 'userSettings', 'browser', 'events', 'pluginManager', 'backdrop', 'globalize', 'require', 'appSettings', 'webSettings'], function (appHost, userSettings, browser, events, pluginManager, backdrop, globalize, require, appSettings, webSettings) {
```
JustAMan: why add?

src/scripts/settings/webSettings.js
```diff
@@ -0,0 +1,17 @@
...
+
+    var data = {};
+
+    function WebSettings() {
+        fetch("/config.json").then(function (response) {
+            data = response.json();
+        })
+    }
+
+    WebSettings.prototype.getMultiServer = function () {
```
JustAMan:
```suggestion
    WebSettings.prototype.allowMultiServer = function () {
```

src/strings/en-us.json
```diff
@@ -15,7 +15,7 @@
     "Alerts": "Alerts",
     "All": "All",
     "AllChannels": "All channels",
-    "AllComplexFormats": "All complex formats (ASS, SSA, VOBSUB, PGS, SUB/IDX, etc.)",
+    "AllComplexFormats": "All Complex Formats (ASS, SSA, VOBSUB, PGS, SUB, IDX)",
```
JustAMan: why do you edit language files when adding configuration? don't mix unrelated stuff please :)

src/scripts/settings/webSettings.js
dkanada: I figured we could use get here because we will never add a set method to these options.

src/strings/en-us.json
dkanada: I made these changes because I was originally adding some configuration options to the subtitles options page, I wouldn't normally do that.

src/components/skinManager.js
dkanada: good catch

src/scripts/settings/webSettings.js
JustAMan: I still don't like the verb `get` here, because it looks a bit confusing to me what it "gets".
Maybe rename to `isMultiServerEnabled` to be extra-clear?

src/strings/en-us.json
JustAMan: I'll believe you... this time 👅

src/scripts/settings/webSettings.js
```diff
@@ -0,0 +1,17 @@
...
+    var data = {};
+
+    function WebSettings() {
+        fetch("/config.json").then(function (response) {
+            data = response.json();
+        })
+    }
+
+    WebSettings.prototype.enableMultiServer = function () {
+        return data.multiServer || false;
```
thornbill: Since getting the config is async this could randomly return the default if the fetch has not completed when this function is called which doesn’t seem ideal. I wonder if it would be better to have a single function `getConfig()` or something that returns a `Promise` that will resolve with the complete config object.

thornbill: Since getting the config is async this could randomly return the default if the fetch has not completed when this function is called which doesn’t seem ideal. I wonder if it would be better to have a single function `getConfig()` or something that returns a `Promise` that will resolve with the complete config object.

src/scripts/settings/webSettings.js
```diff
@@ -1,16 +1,26 @@
...
+        if (data) {
+            return data;
+        }
 
-    function WebSettings() {
         fetch("/config.json").then(function (response) {
             data = response.json();
         })
+
+        return data;
```
JustAMan: this feels wrong... if anything your `getConfig` should be a promise or something...

src/components/displaysettings/displaysettings.template.html
```diff
@@ -63,7 +63,7 @@ <h2 class="sectionTitle">
 
     <div class="selectContainer fldDateTimeLocale hide">
         <select is="emby-select" class="selectDateTimeLocale" label="${LabelDateTimeLocale}">
-            <option value="">${AutoBasedOnLanguageSetting}</option>
+            <option value="">${Auto}</option>
```
MrTimscampi, line 66: I feel like it's clearer to let it be "Auto (based on language setting)", though I'd remove the "Auto" part.

src/scripts/settings/appSettings.js
```diff
@@ -20,6 +20,14 @@ define(['appStorage', 'events'], function (appStorage, events) {
         return this.get('enableAutoLogin') !== 'false';
     };
 
+    AppSettings.prototype.enableSystemExternalPlayers = function (val) {
+        if (val != null) {
```
MrTimscampi, line 24:
```suggestion
        if (val !== null) {
```

A strict comparison is usually better.

dmitrylyzo:
@dkanada why did you refuse config caching (`data`)? You could simply return `Promise.resolve(data)` if `data` is fetched. At current moment, if there will be more `getConfig` calls, it will be fetched each time.

dkanada:
The intent is to apply config changes immediately, which won't work if any kind of browser cache is used during the fetch.

src/scripts/settings/appSettings.js
dkanada: good catch, missed that

src/components/displaysettings/displaysettings.template.html
dkanada: I was mainly trying to reduce the amount of strings I think, because `auto` already implies automatic detection using some metric and we don't explain any similar instances.

src/scripts/settings/webSettings.js
```diff
@@ -1,27 +1,11 @@
...
-    }
-
-    WebSettings.prototype.enableMultiServer = function () {
-        return getConfig().multiServer || false;
-    };
-
-    return new WebSettings();
-});
+function getConfig() {
+    return fetch("/config.json?nocache=" + new Date().getUTCMilliseconds()).then(function (response) {
```
JustAMan, line 2: Any reason for not storing the value of response to a local variable?
IMHO it should request settings once and use them until a page is completely refreshed.

dkanada: > The intent is to apply config changes immediately, which won't work if any kind of browser cache is used during the fetch.

JustAMan: I see the intent, but it would make running the UI slower by adding one more request.
The case when you update the settings is usually rare, you don't go changing them all the time.

So refreshing a page manually when settings are changed isn't a big deal, but adding more lag is IMHO.

dkanada: Refreshing the page manually will always hit the browser cache with the same URL and we can't ask every user to update their browser cache whenever they change these settings.

dmitrylyzo: > You could simply return Promise.resolve(data) if data is fetched.

```js
let data;

function getConfig() {
    if (data) {
        return Promise.resolve(data);
    }

    return fetch("/config.json?nocache=" + new Date().getUTCMilliseconds()).then(function (response) {
        data = response.json();
        return data;
    });
}
...
```

dkanada: That would be a nice middle ground, but it depends on if we want to refresh the settings during a user session or not. This wouldn't refresh until they fully reloaded the website.

dkanada: I went with this method, I think once per session should be enough now that you mention it.

PR #821, opened 2020-02-17
#2407: Prefer embedded MP4-metadata for episode and season numbers

viaregio:
**Changes**
New `LbraryOption` `EnableEmbeddedEpisodeInfos` for TV-Shows.

**Issues**
#2407


Artiume:
https://github.com/jellyfin/jellyfin/pull/2421

src/strings/en-gb.json
```diff
@@ -597,6 +597,8 @@
     "PreferredNotRequired": "Preferred, but not required",
     "PreferEmbeddedTitlesOverFileNamesHelp": "This determines the default display title when no internet metadata or local metadata is available.",
     "PreferEmbeddedTitlesOverFileNames": "Prefer embedded titles over filenames",
+    "PreferEmbeddedEpisodeInfosOverFileNamesHelp": "This uses the episode informations from the embedded metadata if available.",
+    "PreferEmbeddedEpisodeInfosOverFileNames": "Prefer embedded episode informations over filenames",
```
JustAMan: If en-gb version doesn't differ from en-us you don't need to specify it, only specify translation parts which differ.

dkanada: Well we already have so many duplicate strings in the British translation I don't think it matters at this point. I assume the same thing happens for all similar languages.

JustAMan: That doesn't mean we should keep piling dupes tho ;)

PR #822, opened 2020-02-17
Add restart warning to upnp

Artiume:
**Changes**

**Issues**

https://github.com/jellyfin/jellyfin/issues/2401

### DIFF
--- a/src/strings/en-gb.json
+++ b/src/strings/en-gb.json
@@ -1031,7 +1031,7 @@
     "LabelEnableDlnaServerHelp": "Allows UPnP devices on your network to browse and play content.",
     "LabelEnableDlnaDebugLoggingHelp": "Create large log files and should only be used as needed for troubleshooting purposes.",
     "LabelEnableDlnaClientDiscoveryIntervalHelp": "Determines the duration in seconds between SSDP searches performed by Jellyfin.",
-    "LabelEnableAutomaticPortMapHelp": "Attempt to automatically map the public port to the local port via UPnP. This may not work with some router models.",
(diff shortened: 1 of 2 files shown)


PrplHaz4:
Is it really only restart to disable or is it restart the server for changes to take effect.

Artiume:
Not sure

Artiume:
There's a lot of settings that require restart to take effect. Do we want the warning to list them all? And personally, I like the warning there instead of a pop up. If I know it requires a system restart, chances are I'm less likely to mess with the setting.

PookaMustard:
@Artiume @PrplHaz4 I used the issue linked in the OP to find out which scenarios required a restart to trigger or stop the issue and which didn't. Here's what I found:

- **DLNA Play To** triggers #2401 without restart. Restarting required for stopping it.
- **Automatic port forwarding** requires a restart for triggering AND stopping #2401.

dkanada:
It's not a dialog, it's just stored in a separate string file so we wouldn't need to list them all. All we need to do is append the generic `This option requires a restart to take effect.` for any applicable settings.

Artiume:
Does that string already exist or would we need to add its implementation?

dkanada:
I think it exists somewhere, but if it doesn't we should make it now.

Artiume:
https://github.com/jellyfin/jellyfin-web/blob/7e369da67b7b0965258e0ba1d9310f4bd3182bf1/src/networking.html#L88 

What'd be the best way to add the new variable?

dkanada:
added to #801

PR #823, opened 2020-02-17
Details page artwork ribbon revamp

### LINKED ISSUE #820: Header art is too big on movie details page
**Describe the bug**
The header art is too big on the movie details page. It obscures the poster and other details (cast and similar movies)

**Expected behavior**
Solution is not clear. In my opinion, it should be removed and displayed only as a backdrop.

**Screenshots**
out

 - Browser: Firefox 73
 - Jellyfin Version: 20200216 nightly

MrTimscampi:
**Changes**

Rework of the artwork on the details page.

ded19b1 - Hides the artwork ribbon (backdrop and logo) if the user has disabled the backdrops. Adjusts the position of the details and the poster to compensate for the backdrop ribbon not being shown.

e0dc32c  - Moves the poster and sticky bar, removes the stickyness.

Further commits fix up the layout when no backdrop is there.

image

image

**Issues**

Fixes #820

### DIFF
--- a/src/assets/css/librarybrowser.css
+++ b/src/assets/css/librarybrowser.css
@@ -439,8 +439,19 @@
     background-size: cover;
     background-repeat: no-repeat;
     background-position: center;
+    background-attachment: fixed;
     height: 50vh;
     position: relative;
+}
+
+.layout-desktop .itemBackdrop::after,
+.layout-tv .itemBackdrop::after {
+    content: "";
+    width: 100%;
+    height: 100%;
+    background: #000000ad;
+    background: rgba(0, 0, 0, 0.65);
+    display: block;
 }
 
 .itemBackdrop.noBackdrop {
@@ -525,8 +536,6 @@
     display: flex;
     align-items: center;
     align-content: center;
-    position: sticky;
-    top: 3.85em;
     z-index: 2;
 }
 
@@ -536,13 +545,19 @@
     top: 0;
 }
 
-.layout-tv .detailPagePrimaryContainer {
+.layout-tv .detailPagePrimaryContainer:not(.noBackdrop),
+.layout-desktop .detailPagePrimaryContainer:not(.noBackdrop) {
     position: relative;
     top: 0;
+    padding-left: 32.45vw;
 }
 
 .detailSticky {
-    background-color: #101010;
+    margin-top: -7.3em;
+}
+
+.detailSticky.noBackdrop {
+    margin-top: 0;
 }
 
 .infoWrapper {
@@ -572,15 +587,15 @@
 }
 
 .detailImageContainer {
-    position: sticky;
-    top: 25%;
+    position: relative;
+    margin-top: -25vh;
     float: left;
-    width: 22.786458333333332vw;
+    width: 25vw;
+    z-index: 3;
 }
 
-.layout-mobile .detailImageContainer,
-.layout-tv .detailImageContainer {
-    position: relative;
+.detailImageContainer.noBackdrop {
+    margin-top: 0;
 }
 
 .detailPagePrimaryContent {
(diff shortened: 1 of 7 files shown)


src/controllers/itemdetailpage.js
```diff
@@ -457,20 +457,25 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "cardBuild
         var imgUrl;
         var screenWidth = screen.availWidth;
         var hasbackdrop = false;
+        console.debug(page);
```
dkanada: maybe remove this

src/controllers/itemdetailpage.js
```diff
@@ -1047,6 +1055,8 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "cardBuild
     }
 
     function renderDetails(page, item, apiClient, context, isStatic) {
+        var taglineElement = page.querySelector(".detailPageContent");
```
dkanada: Is this used somewhere below?

MrTimscampi:
> It might be better to keep the same detail sticky placement without a backdrop and simply remove the background color change.

The issue is that then, we're losing a ton of space to what is essentially blank space. For example, it doesn't make sense for actors or media types where there will pretty much never be a backdrop, then it's simply wasted space.

As an aside, I removed all the dead code pointed out by ESLint in itemdetailspage.js. It was doing API calls to get user details that were never used afterwards, as well as passing completely useless parameters to some of the methods.

src/assets/css/librarybrowser.css
```diff
@@ -439,10 +439,43 @@
...
     position: relative;
 }
 
+.layout-desktop .itemBackdrop::after,
+.layout-tv .itemBackdrop::after {
+    content: "";
+    width: 100%;
+    height: 100%;
+    background: #000000ad;
+    background: rgba(0, 0, 0, 0.65);
```
JustAMan: uh?

src/controllers/itemdetailpage.js
```diff
@@ -464,13 +462,17 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "cardBuild
             item.Type === "MusicArtist" ||
             item.Type === "Person";
 
+        if (!layoutManager.mobile && !userSettings.enableBackdrops()) {
+            return hasbackdrop;
```
JustAMan:
```suggestion
            return false;
```
for clarity?

src/controllers/itemdetailpage.js
```diff
@@ -506,15 +508,15 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "cardBuild
                 index: 0,
                 tag: item.ImageTags.Thumb
             });
-            itemBackdropElement.classList.remove("noBackdrop");
+            page.classList.remove("noBackdrop");
             imageLoader.lazyImage(itemBackdropElement, imgUrl, false);
             hasbackdrop = true;
         } else {
-            itemBackdropElement.classList.add("noBackdrop");
```
JustAMan: why?

src/controllers/itemdetailpage.js
```diff
@@ -506,15 +508,15 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "cardBuild
...
             itemBackdropElement.style.backgroundImage = "";
         }
 
         if ("Person" === item.Type) {
+            page.classList.add('noBackdrop');
```
JustAMan: why?

src/assets/css/librarybrowser.css
MrTimscampi: It's from aivit's code for this.

Not really needed nowadays to have the hex fallback, but it doesn't hurt (I can remove it tonight if needed).

src/controllers/itemdetailpage.js
MrTimscampi: The noBackdrop class is now at the very top of the page, so everything below it has access to it. No need to add it elsewhere.

src/controllers/itemdetailpage.js
MrTimscampi: Person's seldom have a backdrop. There's likely a better way to handle this case, though.

JustAMan: add a `//FIXME` comment explaining this case then, please

PR #837, opened 2020-02-21
Close dialogs when container background is right clicked

h1nk:
**Changes**
* Close application right click menus when background container is right clicked and also prevent browser's default context menu from appearing.

### DIFF
--- a/src/components/dialogHelper/dialogHelper.js
+++ b/src/components/dialogHelper/dialogHelper.js
@@ -169,6 +169,15 @@ define(['appRouter', 'focusManager', 'browser', 'layoutManager', 'inputManager',
         }, {
             passive: true
         });
+
+        dom.addEventListener((dlg.dialogContainer || backdrop), 'contextmenu', function (e) {
+            if (e.target === dlg.dialogContainer) {
+                // Close the application dialog menu
+                close(dlg);
+                // Prevent the default browser context menu from appearing
+                e.preventDefault();
+            }
+        });
     }
 
     function isHistoryEnabled(dlg) {


dkanada:
I think this might be a good change.

JustAMan:
I don't see how it is an expected behaviour that right-clicking anywhere in a dialog closes it.
If I saw this in some software I would consider it a bug TBH...

dkanada:
It's not inside the dialog, I thought he meant outside the dialog.

h1nk:
@JustAMan & @dkanada this does apply to when the background container of the dialog (the space outside the dialog) is right clicked, not the context menu itself. This behavior is very common in web-based and Electron UI applications such as Discord, Spotify etc. When a user makes any sort of action outside of the application provided dialog/context menu (left or right clicking etc.) the dialog should close. Additionally the default browser context menu should be completely hidden from the user which this does.

JustAMan:
I still don't see how is this a useful behaviour. I very much dislike a dialog disappearing because of a misclick. There should be buttons to explicitly close the dialog.

At worst make it configurable, as @dkanada added the web-ui configuration object. I personally would like to disable such behaviour on my server...

h1nk:
> There should be buttons to explicitly close the dialog.

@JustAMan Aren't these dialogs just supposed to emulate right click context menus? Do you mean add a button DOM element in the corner of the context menu to close it when it's open or are you talking about just having the left mouse click button be the only mouse button able to close right click context menus when you click the surrounding space of the context menu?

I still think it's important to have special handling of right clicks while a context menu is already open, otherwise you end up seeing the browser context menu:
4f85b450-565b-4aeb-8170-a92d3cb3b643

In the Spotify web player client right clicking while a context menu is already open will just reopen the menu at that the new cursor position:
c369a510-e64b-4e0e-a3e7-7b30bab0db74

It's also how the default right click works in Firefox:
b1eba276-3067-4635-999f-30a0b20d2e33

Perhaps this is how it should instead of just closing the menu for a second right click? I think it'd be nice to remove the long open and close animations for context menus and also remove the background darkening and just have the context menu appear at the cursor position instead of centered at the clicked item (which I find to be a little awkward and unexpected). How does that sound for an improvement?

dkanada:
I agree the background dim should be removed, but I think at least some kind of animation should be kept for the open and close actions. @JustAMan the right click action seems like an improvement with this pull request to me, especially because it's a standard on other popular sites.

MrTimscampi:
The menu should work differently for desktop than it does now.

Right now, it feels very cumbersome, with the long fade-in, the dim background, the bad positioning.

It feels alright for mobile or TV, but it feels really clunky for desktop. So I'm for this, as a first step towards changing how this thing behaves.

JustAMan:
I thought it was for full-blown dialog with Save/Cancel buttons etc. Is this only for popup menu?

dkanada:
This is for any dialog, including the ones with save buttons.

JustAMan:
> This is for any dialog, including the ones with save buttons.

Can we apply auto-close to only menus (and hide browser menu everywhere if that is needed)?

dkanada:
Not without reworking the dialog options, but that isn't within the scope of this pull request in my opinion.

h1nk:
It was my intention for this PR to just apply to the way right-click context menus close and not other UI elements. I can expand this PR to rework context menu handling to cover more issues as discussed previously:
- Opening context menus at the cursor position rather than centered on the clicked item
- Remove the background darkening for context menus
- Shortening the animation time for context menus
- Decoupling the close logic for dialogs and right-click context menus

dkanada:
Those all sound like excellent changes, but they can definitely be in separate pull requests. Splitting the context menus into a separate component should help with some of the issues @JustAMan was mentioning.

JustAMan:
Yeah, I just messed with dialogs two days ago, and I support this change - it's doing better than it was.
Still would love to see a distinction between a dialog and a context menu later on.

PR #838, opened 2020-02-21
Add missing final newlines

h1nk:
I noticed that EditorConfig ended up making changes in a lot of files that I opened and viewed because they did not include final newlines, which is annoying since I didn't wish to commit changes in those files.

**Changes**
* Added final newlines to source files that did not previously include them

Perhaps there's a good way to enforce final newlines for future CI linter checks?


MrTimscampi:
> Perhaps there's a good way to enforce final newlines for future CI linter checks?

This is indeed something that needs to be added at some point.

The JS linter checks generally need a lot of improvements, but a lot of the style changes we want to make require a decent amount of work to fix them manually in the existing source.

I do intend to go back over some rules and make them tighter after 10.5.0 releases, though, if only to be consistent in every file (Single quotes, empty line at EOF, etc)

joshuaboniface:
Why would we not want trailing newlines? My editor (vim) creates them, and they're not bad.

dkanada:
Changing [this](https://github.com/jellyfin/jellyfin-web/blob/master/.eslintrc.yml#L11) line will enforce the rule for JavaScript files.

h1nk:
@thornbill, oops, looks like I missed that one gif file. There should be no changes in any binary files now.

PR #843, opened 2020-02-22
Webos icon fix

grafixeyehero:


src/components/nowplayingbar/nowplayingbar.js
```diff
@@ -54,7 +54,7 @@ define(['require', 'datetime', 'itemHelper', 'events', 'browser', 'imageLoader',
 
         html += '<div class="nowPlayingBarRight">';
 
-        html += '<button is="paper-icon-button-light" class="muteButton mediaButton"><i class="material-icons volume_up"></i></button>';
+        html += '<button is="paper-icon-button-light" class="muteButton mediaButton"><i class="material-icons"></i></button>';
```
MrTimscampi: I think a better fix would be performed here: https://github.com/jellyfin/jellyfin-web/blob/3b5a63f930b4452dafd04bce57b5875f19c2cece/src/components/nowplayingbar/nowplayingbar.js#L416

Seems like that needs to be changed to remove the volume_up class from the `<i>` and add the other corresponding class.

src/components/playback/volumeosd.js
```diff
@@ -11,7 +11,7 @@ define(['events', 'playbackManager', 'dom', 'browser', 'css!./iconosd', 'materia
     function getOsdElementHtml() {
         var html = '';
 
-        html += '<i class="material-icons iconOsdIcon volume_up"></i>';
+        html += '<i class="material-icons iconOsdIcon"></i>';
```
MrTimscampi: Same thing here, I think the proper fix is elsewhere.

src/nowplaying.html
```diff
@@ -63,7 +63,7 @@ <h2 class="nowPlayingPageTitle"></h2>
                         </button>
 
                         <button is="paper-icon-button-light" class="buttonMute autoSize" title="${Mute}">
-                            <i class="xlargePaperIconButton material-icons volume_up"></i>
```
MrTimscampi: This one as well likely needs a JS change instead of a class removal.

src/videoosd.html
```diff
@@ -76,7 +76,7 @@ <h3 class="osdTitle"></h3>
 
                 <div class="volumeButtons hide-mouse-idle-tv">
                     <button is="paper-icon-button-light" class="buttonMute autoSize" title="${Mute} (m)">
-                        <i class="xlargePaperIconButton material-icons volume_up"></i>
```
MrTimscampi: Same for this.

src/components/nowplayingbar/nowplayingbar.js
grafixeyehero: yap that's why i remove those class b/c it will add  it <i>-here-</i>

PR #844, opened 2020-02-23
Fix horizontal scrolling on emby-scrollers

thornbill:
**Changes**
* Removes data attributes disabling wheel event handling by scroller.js
* Modifies scroller.js so that when horizontal scrolling is enabled only deltaX is used

Without the modification to scroller.js, vertical scrolling in emby-scroller components would be handled like the user was horizontal scrolling. This would prevent the user from scrolling vertically through the page until the end of the items in the emby-scroller component had been reached.

**Issues**
#487 

rollin

### DIFF
--- a/src/components/scroller.js
+++ b/src/components/scroller.js
@@ -575,7 +575,9 @@ define(['browser', 'layoutManager', 'dom', 'focusManager', 'ResizeObserver', 'sc
 		 */
         function normalizeWheelDelta(event) {
             // wheelDelta needed only for IE8-
-            scrolling.curDelta = ((o.horizontal ? event.deltaY || event.deltaX : event.deltaY) || -event.wheelDelta);
+            // JELLYFIN MOD: Only use deltaX for horizontal scroll instead of `event.deltaY || event.deltaX`
+            scrolling.curDelta = ((o.horizontal ? event.deltaX : event.deltaY) || -event.wheelDelta);
+            // END JELLYFIN MOD
 
             if (transform) {
                 scrolling.curDelta /= event.deltaMode === 1 ? 3 : 100;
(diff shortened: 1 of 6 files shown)


dkanada:
Is there any way we can add a config option somewhere? It sounds like this would block vertical scroll for every single scroller on the home page, which isn't ideal.

src/components/scroller.js
```diff
@@ -575,7 +575,9 @@ define(['browser', 'layoutManager', 'dom', 'focusManager', 'ResizeObserver', 'sc
 		 */
         function normalizeWheelDelta(event) {
             // wheelDelta needed only for IE8-
-            scrolling.curDelta = ((o.horizontal ? event.deltaY || event.deltaX : event.deltaY) || -event.wheelDelta);
+            // JELLYFIN MOD: Only use deltaX for horizontal scroll instead of `event.deltaY || event.deltaX`
+            scrolling.curDelta = ((o.horizontal ? event.deltaX : event.deltaY) || -event.wheelDelta);
```
dmitrylyzo: Firefox 72.0.1 on mouse wheel reports
```
event.deltaX = 0;
event.deltaY = 3;
```
In result, `scrolling.curDelta = -event.wheelDelta <[undefined] = NaN`

I tried to leave as it was, replace `passive: true` with `capture: true` for `scrollHandler` and add `event.preventDefault()` in `scrollHandler` after this block
https://github.com/jellyfin/jellyfin-web/blob/5fb2c8bd82b3dae4ad629d38712f683935ad06cf/src/components/scroller.js#L600-L602
In this case, emby-scroller scrolls when hovered, window does not scroll (even scroller at very start/end). To scroll window, need to place cursor over empty place of page.

thornbill: I removed the fallback to `wheelDelta` since that was only present to support IE8 based on the comment. It is working well for me now on Firefox 73 and Chrome 80.

thornbill:
@dkanada my change to scroller.js was to prevent it blocking the window scrolling when inside an emby-scroller.

There was an issue where sometimes the `wheelDelta` would be used when `deltaX` was `0`. Based on the comment in the file that was only present to support IE8, so I removed it. It is working well for me on Firefox 73 and Chrome 80 now.

PR #849, opened 2020-02-24
Catch some errors and remove unused method

dkanada:


src/controllers/streamingsettings.js
```diff
@@ -30,24 +31,6 @@ define(["jQuery", "libraryMenu", "loading"], function ($, libraryMenu, loading)
     }
 
     $(document).on("pageinit", "#streamingSettingsPage", function () {
-        var page = this;
-        $("#btnSelectTranscodingTempPath", page).on("click.selectDirectory", function () {
```
JustAMan: why is this no longer used? is transcoding path no longer controllable?

src/controllers/dashboard/general.js
```diff
@@ -56,6 +56,12 @@ define(["jQuery", "loading", "fnchecked", "emby-checkbox", "emby-textarea", "emb
                         }
                     });
                 });
+            }, function () {
```
JustAMan: don't we have to make this being called in `.then()` of inner calls, too? right now it only fires on error of `.updateServerConfiguration()` if I understand correctly

src/controllers/streamingsettings.js
dkanada: It's controllable in a separate page, this must have been left over from some other time.

src/controllers/dashboard/general.js
dkanada: Technically yes, but I was only adding this in response to a specific error related to the server configuration, so I didn't want to track down every other API call and add an error handler.

JustAMan: So looks like we need a ticket for tracking that thing... could fix later.
Do we have a ticket for your "specific error"?

dkanada: No ticket, I just found and fixed it in like fifteen minutes.

PR #856, opened 2020-02-25
Add eslint:recommended to eslint config

thornbill:
**Changes**
Updates eslint config to use eslint:recommended

**Issues**
N/A


src/components/cardbuilder/cardBuilder.js
```diff
@@ -1383,7 +1380,7 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
             }
 
             if (item.Type === 'CollectionFolder' || item.CollectionType) {
-                var refreshClass = item.RefreshProgress || (item.RefreshStatus && virtualFolder.item !== 'Idle') ? '' : ' class="hide"';
+                var refreshClass = item.RefreshProgress ? '' : ' class="hide"';
```
JustAMan: why?

src/controllers/dashboard/dashboard.js
```diff
@@ -741,21 +741,15 @@ define(["datetime", "events", "itemHelper", "serverNotifications", "dom", "globa
...
-                renderHasPendingRestart(view, apiClient, true);
-            }
+            console.debug('onServerShuttingDown not implemented', evt, apiClient);
         }
 
         function onServerRestarting(evt, apiClient) {
-            if (apiClient.serverId() === serverId) {
-                renderHasPendingRestart(view, apiClient, true);
-            }
+            console.debug('onServerRestarting not implemented', evt, apiClient);
```
JustAMan: why?

src/controllers/itemdetailpage.js
```diff
@@ -647,7 +647,7 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "cardBuild
     function setTitle(item, apiClient) {
         var url = logoImageUrl(item, apiClient, {});
 
-        if (url = null) {
+        if (url == null) {
```
JustAMan: LOL

src/controllers/dashboard/dashboard.js
MrTimscampi: `renderHasPendingRestart()` doesn't exist.

src/controllers/itemdetailpage.js
dmitrylyzo: Since 11994657af29ad8f30eafe93b82f5ad66b12a0a8
But I think that `!=` (or just `!url`) should be used here, because `url` is used in the block.

src/components/cardbuilder/cardBuilder.js
thornbill: `virtualFolder` is undefined and going back to the original fork commit it has never been defined, so this could never be true.

src/controllers/dashboard/dashboard.js
thornbill: `renderHasPendingRestart` is undefined. It used to be just an empty function, but that was removed at some point. Thought we should just log it in case this is used somewhere.

src/controllers/itemdetailpage.js
JustAMan: yes, I think it should be `url != null`

thornbill: Good catch! That is definitely the case.

src/components/htmlvideoplayer/plugin.js
```diff
@@ -740,7 +742,9 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
                 }
             }
 
+            // No idea what this is possibly trying to workaround
```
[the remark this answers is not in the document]
thornbill: :wastebasket: removed.

src/controllers/itemdetailpage.js
thornbill: Fixed :sweat_smile:

PR #871, opened 2020-03-01
Move page.js to node

grafixeyehero:
**Note: need test for tv and mobile platform**

And this still need some clean up


src/components/dialogHelper/dialogHelper.js
```diff
@@ -172,7 +172,7 @@ define(['appRouter', 'focusManager', 'browser', 'layoutManager', 'inputManager',
     }
 
     function isHistoryEnabled(dlg) {
-        return dlg.getAttribute('data-history') === 'true';
+        return dlg.getAttribute('data-history') === 'false';
```
JustAMan: wtf? :)

src/components/appRouter.js
```diff
@@ -666,7 +669,8 @@ define(['loading', 'globalize', 'events', 'viewManager', 'layoutManager', 'skinM
 
     function pushState(state, title, url) {
         state.navigate = false;
-        page.pushState(state, title, url);
+        history.pushState(state, title, url);
```
JustAMan: where is `history` defined?

grafixeyehero: https://developer.mozilla.org/en-US/docs/Web/API/History/pushState

src/components/dialogHelper/dialogHelper.js
grafixeyehero: What's wrong with this?

dmitrylyzo: > What's wrong with this?

logic reverse

grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/871/commits/4fca1c15638e9f0cd4fc8317fe7d8baef115da66

src/components/appRouter.js
```diff
@@ -562,7 +562,10 @@ define(['loading', 'globalize', 'events', 'viewManager', 'layoutManager', 'skinM
         if (!document.querySelector('.dialogContainer') && startPages.indexOf(curr.type) !== -1) {
             return false;
         }
-        return page.canGoBack();
+        if (enableHistory) {
```
dmitrylyzo:
```suggestion
        if (enableHistory()) {
```

grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/871/commits/08afa4c0ca07487ee3763201726af6472335ded1

dmitrylyzo:
`enableHistory` and `enableNativeHistory` are still questionable.
I looked how it was before. Result of `enableHistory` function is passed to `page` constructor, then to `page.start`. Then it becomes the result of `page.enableNativeHistory`. So, `enableHistory` = `enableNativeHistory`. :confused:

MrTimscampi:
Do note that page.js doesn't seem to have any parameter relating to history anymore, so I think it was dropped (and can thus be entirely removed).

All our clients support the HTML5 history API anyway.

In fact, the official documentation explicitly states that the history API should be polyfilled if the clients don't support it.

JustAMan:
> `enableHistory` and `enableNativeHistory` are still questionable.
> I looked how it was before. Result of `enableHistory` function is passed to `page` constructor, then to `page.start`. Then it becomes the result of `page.enableNativeHistory`. So, `enableNativeHistory` = `enableHistory`. 😕



> Do note that page.js doesn't seem to have any parameter relating to history anymore, so I think it was dropped (and can thus be entirely removed).
> 
> All our clients support the HTML5 history API anyway.

it might have something to do with TVs being the clients... @dmitrylyzo your say?

dmitrylyzo:
> it might have something to do with TVs being the clients...

`enableHistory` (and therefore `enableNativeHistory` - before PR) returned `true` for almost every device. So old `page.js` used `history` almost always. New `page.js` uses `history` if one exists. It seems that Xbox One has forcibly disabled history.
https://github.com/jellyfin/jellyfin-web/blob/9162f53b98872c60af923bccd06f761d48872639/src/components/appRouter.js#L359-L362

Above, I mean that now `enableHistory` and `enableNativeHistory` are not in sync.

Here is the only place where it is used:
https://github.com/jellyfin/jellyfin-web/blob/9162f53b98872c60af923bccd06f761d48872639/src/components/dialogHelper/dialogHelper.js#L378-L380
And now it is turned off, so dialogs do not use history, but navigation uses.

PR #872, opened 2020-03-02
Update ISO playback warning

Artiume:
Tinkering with ISO backups. I've noticed this behavior. I'm actually not sure how to play DVDs/Blurays via the webui.


src/components/playback/experimentalwarnings.js
```diff
@@ -46,21 +46,21 @@ define(['connectionManager', 'globalize', 'userSettings', 'apphost'], function (
     function showBlurayMessage() {
 
         var message =
-            'Playback of Bluray folders in this app is experimental. Some titles may not work at all. For a better experience, consider converting to mkv video files, or use an Jellyfin app with native Bluray folder support.';
+            'Playback of Bluray folders in this app is experimental. Jellyfin cannot decrypt DRM. The current behavior is to play all files in series, this can include a two minute black intro video.';
```
JustAMan: I don't like `two minute long black video`. Also I'm not sure this is how it works, let's ask @Bond-009 as IIRC he was the last one to rework VOB support.

ferferga:
@Artiume @dkanada This is now translatable

I brought back the 'titles' word, as I don't understand why it was removed. 'Titles' in DVD, ISO and Blu-Ray are the different parts that compose a DVD, such as the menus, the extras, etc. From my understanding (although I don't know the context because I never saw that message), I think this warning was intended for those kinds of menus, that don't really fit in a media server.

I also removed some redundant stuff like "this app" (we already know that we're running Jellyfin) and make the message a bit clearer, so it covers those strange menus and DRM content without any confusion (imho)

src/strings/en-us.json
```diff
@@ -1492,5 +1492,8 @@
     "EveryXMinutes": "Every {0} minutes",
     "EveryHour": "Every hour",
     "EveryXHours": "Every {0} hours",
-    "OnApplicationStartup": "On application startup"
+    "OnApplicationStartup": "On application startup",
+    "ExperimentalBlurayPlayback": "Blu-Ray playback is experimental. Jellyfin cannot decrypt content protected by DRM. All the titles in the Blu-Ray will be played anyway, including the protected ones. Some parts might be in black",
```
dkanada, line 1496:
```suggestion
    "ExperimentalBlurayPlayback": "Jellyfin cannot decrypt content protected by DRM but all content will be attempted regardless, including protected titles. This may cause some files to appear completely black during the encrypted sections.",
```

dkanada: I would change all three to match this warning instead.

ferferga: @dkanada But what if one title is interactive and not encrypted? They can't be played back, but they're not encrypted. Black Playback ≠ Content Encrypted

 Also, I would use **titles** instead of **files**

MrTimscampi: Interactive DVD titles will simply show up as short videos. Blu-ray menus aren't really titles per se, so nothing interactive will show up (afaik all interactive content on Blu-ray discs uses Java)

ferferga: @MrTimscampi You tested that? Because I know VLC and so on plays them as short videos, but from what I guessed from the original message (not @Artiume's) is that those titles doesn't work at all in Jellyfin.

dkanada:
I clicked the wrong button, whoops.

src/strings/en-us.json
```diff
@@ -1492,5 +1492,8 @@
...
+    "ExperimentalDVDPlayback": "DVD playback is experimental. Jellyfin cannot decrypt content protected by DRM. All the titles in the DVD will be played anyway, including the protected ones. Some parts might be in black",
+    "ExperimentalISOPlayback": "ISO playback is experimental. All the titles in the ISO will be played anyway. Some parts might be in black"
```
dkanada, line 1498:
```suggestion
    "UnsupportedPlayback": "Jellyfin cannot decrypt content protected by DRM but all content will be attempted regardless, including protected titles. Some files may appear completely black due to encryption or other unsupported features, such as interactive titles."
```

dkanada, line 1498:
```suggestion
    "UnsupportedPlayback": "Jellyfin cannot decrypt content protected by DRM but all content will be attempted regardless, including protected titles. Some files may appear completely black due to encryption or other unsupported features, such as interactive titles."
```

src/components/playback/experimentalwarnings.js
```diff
@@ -44,24 +44,15 @@ define(['connectionManager', 'globalize', 'userSettings', 'apphost'], function (
     }
 
     function showBlurayMessage() {
-
-        var message =
-            'Playback of Bluray folders in this app is experimental. Some titles may not work at all. For a better experience, consider converting to mkv video files, or use an Jellyfin app with native Bluray folder support.';
-        return showMessage(message, 'blurayexpirementalinfo', 'nativeblurayplayback');
+        return showMessage(globalize.translate("ExperimentalBluRayPlayback"), 'blurayexpirementalinfo', 'nativeblurayplayback');
```
dkanada, line 47:
```suggestion
        return showMessage(globalize.translate("UnsupportedPlayback"), 'blurayexpirementalinfo', 'nativeblurayplayback');
```

src/components/playback/experimentalwarnings.js
```diff
@@ -44,24 +44,15 @@ define(['connectionManager', 'globalize', 'userSettings', 'apphost'], function (
...
     }
 
     function showDvdMessage() {
-
-        var message =
-            'Playback of Dvd folders in this app is experimental. Some titles may not work at all. For a better experience, consider converting to mkv video files, or use an Jellyfin app with native Dvd folder support.';
-        return showMessage(message, 'dvdexpirementalinfo', 'nativedvdplayback');
+        return showMessage(globalize.translate("ExperimentalDVDPlayback"), 'dvdexpirementalinfo', 'nativedvdplayback');
```
dkanada, line 51:
```suggestion
        return showMessage(globalize.translate("UnsupportedPlayback"), 'dvdexpirementalinfo', 'nativedvdplayback');
```

src/components/playback/experimentalwarnings.js
```diff
@@ -44,24 +44,15 @@ define(['connectionManager', 'globalize', 'userSettings', 'apphost'], function (
...
     }
 
     function showIsoMessage() {
-
-        var message =
-            'Playback of ISO files in this app is experimental. Some titles may not work at all. For a better experience, consider converting to mkv video files, or use an Jellyfin app with native ISO support.';
-        return showMessage(message, 'isoexpirementalinfo', 'nativeisoplayback');
+        return showMessage(globalize.translate("ExperimentalISOPlayback"), 'isoexpirementalinfo', 'nativeisoplayback');
```
dkanada, line 55:
```suggestion
        return showMessage(globalize.translate("UnsupportedPlayback"), 'isoexpirementalinfo', 'nativeisoplayback');
```

PR #881, opened 2020-03-04
Fix for incorrect button positioning when hovering card

ferferga:
Follow-up of #866 but using flex, which works better when scaling the content

### DIFF
--- a/src/components/cardbuilder/cardBuilder.js
+++ b/src/components/cardbuilder/cardBuilder.js
@@ -1457,7 +1457,7 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
                 html += '<button is="paper-icon-button-light" class="' + btnCssClass + ' cardOverlayFab-primary" data-action="resume"><i class="material-icons cardOverlayButtonIcon cardOverlayButtonIcon-hover play_arrow"></i></button>';
             }
 
-            html += '<div class="cardOverlayButton-br">';
+            html += '<div class="cardOverlayButton-br flex align-items-center">';
 
             var userData = item.UserData || {};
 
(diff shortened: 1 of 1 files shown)


src/components/cardbuilder/cardBuilder.js
```diff
@@ -1457,7 +1457,7 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
                 html += '<button is="paper-icon-button-light" class="' + btnCssClass + ' cardOverlayFab-primary" data-action="resume"><i class="material-icons cardOverlayButtonIcon cardOverlayButtonIcon-hover play_arrow"></i></button>';
             }
 
-            html += '<div class="cardOverlayButton-br">';
+            html += '<div class="cardOverlayButton-br flex align-items-center">';
```
dkanada: We probably want the items on the bottom right corner.

ferferga: They are still in the right corner in the rest of the cards. There is some weird styling that keeps them in that place when there is enough space.

In fact, the overlapping only happened in the itemDetails page, I couldn't reproduce it in the ``Recently Added`` card's, for example.

ferferga: @dkanada Just pulled the changes I did in Github editor to my local branch and no issues at all, everything is in place.

dmitrylyzo: Buttons are not centered because button container is shrunk to its content and absolutely positioned in right bottom corner of card.
But `flex` alone is enough - it just forces buttons to line up. But still tiny part (~1px of hover indicator) of left-most button does not fit.

_Actually, I get button wrapping even with #866 in collection (Chrome and Firefox)._

ferferga: @dmitrylyzo That's been commented in Matrix, that would need a card resize out of the scope of this PR and 10.5, maybe for 10.6 imo

dmitrylyzo: @ferferga `align-items-center` for what? For me, `flex` only is enough

ferferga: @dmitrylyzo True, I forgot it

PR #883, opened 2020-03-04
Fix tabs hopping

### LINKED ISSUE #829: Menu hops
with the 2px border added to the active menus, the menu hops when you switch menu items in 10.5

MrTimscampi:
**Changes**

Fix tab hopping by removing the bottom border and adjusts margins to not look borked.

**Issues**

Fixes #829

### DIFF
--- a/src/components/emby-tabs/emby-tabs.css
+++ b/src/components/emby-tabs/emby-tabs.css
@@ -13,11 +13,11 @@
     vertical-align: middle;
     flex-shrink: 0;
     margin: 0;
-    padding: 1.5em;
+    padding: 1.5em 1.5em;
     position: relative;
     height: auto;
     min-width: initial;
-    line-height: initial;
+    line-height: 1.25;
     border-radius: 0;
     overflow: hidden;
     font-weight: 600;
--- a/src/assets/css/librarybrowser.css
+++ b/src/assets/css/librarybrowser.css
@@ -854,7 +854,7 @@ div.itemDetailGalleryLink.defaultCardBackground {
         width: auto;
         align-items: center;
         justify-content: center;
-        margin-top: -4.3em;
+        margin-top: -4.2em;
         position: relative;
     }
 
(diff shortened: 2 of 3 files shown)


src/components/emby-tabs/emby-tabs.css
```diff
@@ -13,11 +13,11 @@
     vertical-align: middle;
     flex-shrink: 0;
     margin: 0;
-    padding: 1.5em;
+    padding: 1.5em 1.5em;
     position: relative;
     height: auto;
     min-width: initial;
-    line-height: initial;
+    line-height: 1.25;
```
ferferga: Btw (somewhat unrelated to this PR and nitpicking) I think this property can be completely removed, it prevents the text from centering in the vertical axis in smaller screens

dmitrylyzo:
For vertical hopping, I was thinking of making some invisible filler 2px height. Maybe by pseudo-element.

dmitrylyzo:
Something like this
```css
.emby-tab-button-active::after {
    content: "";
    display: block;
    width: 100%;
    height: 2px;
    background: white;
    bottom: 0;
    position: absolute;
    left: 0;
}
```

ferferga:
@dmitrylyzo I tried your class and, either I applied it wrong, or it doesn't do anything at all

src/assets/css/site.css
```diff
@@ -5,6 +5,10 @@ html {
     height: 100%;
 }
 
+body.withSectionTabs {
+    overflow-y: scroll;
```
thornbill: Do we want to force the scrollbars to display even if they aren't needed?

MrTimscampi: Crap, I left it, good catch.
Can you push to the branch and remove the "Fix horizontal hop" commit? (Latest one).

I can't do it for another 13-14 hours.

thornbill: Removed.

dmitrylyzo:
@ferferga I put that class in `emby-tabs.js` in addition to other changes. It should restore bottom line but keep tabs vertically in-place when switching. _Just checked in Firefox 72 and Chrome 79_

Without `overflow-y: scroll;`, tab buttons are hopping horizontally.
I have not trailers, so when I switch between trailers tab and some filled tab, I get this behavior - top bar width changes.

PR #884, opened 2020-03-04
Fix padding in itemdetails' selectors

ferferga:
Fixes the overlapping and the small separation between the selectors in the video elements (version, subtitles, audio, etc...)

**Before:**
Sin título_old

**After:**
Sin título

### DIFF
--- a/src/itemdetails.html
+++ b/src/itemdetails.html
@@ -116,7 +116,7 @@
                         <div class="genres hide" style="margin: .7em 0;font-size:92%;"></div>
                         <div class="directors hide" style="margin: .7em 0;font-size:92%;"></div>
 
-                        <form class="trackSelections flex align-items-center flex-wrap-wrap hide focuscontainer-x" style="padding:1em 0">
+                        <form class="trackSelections flex align-items-center flex-wrap-wrap hide focuscontainer-x">
                             <div class="selectContainer selectContainer-inline selectSourceContainer hide trackSelectionFieldContainer flex-shrink-zero">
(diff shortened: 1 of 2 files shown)


ferferga:
Switched to classes and margins to get the exact same effect. 

I thought that only adding the padding to the video would fix all the spacing issues, but in mobile layout there was still spacing issues between the audio and subtitles selection tracks. Using padding would misplace the dropdown arrow, so I switched to margins as well.

With margins there are no issues in TV, Desktop and Mobile

src/itemdetails.html
```diff
@@ -116,14 +116,14 @@
                         <div class="genres hide" style="margin: .7em 0;font-size:92%;"></div>
                         <div class="directors hide" style="margin: .7em 0;font-size:92%;"></div>
 
-                        <form class="trackSelections flex align-items-center flex-wrap-wrap hide focuscontainer-x" style="margin: .5em 0;font-size:92%; padding: 0;">
-                            <div class="selectContainer selectContainer-inline selectSourceContainer hide trackSelectionFieldContainer flex-shrink-zero" style="margin-right:1.5em;">
+                        <form class="trackSelections flex align-items-center flex-wrap-wrap hide focuscontainer-x" style="padding:1em 0">
```
dkanada: Why does only this one have padding?

ferferga: Good catch! @dkanada

ferferga:
Saw that the text was noticeable misaligned from the other texts in TV layout (a little bit in the others as well) , so removed left margin and now LGTM.

PR #886, opened 2020-03-05
Repositioned 'played on' message on item details

ferferga:
Changes the position of the "Played on" message on item details to the upper right corner, in a position where fits better imo.

**Before**
OLD

**After**
image

I placed it in the right corner to avoid loosing it in the crowd of all the text, so it's more relevant and noticeable for users. However, I think that it's also good just at the top, without the  ``text-align: right``.

I leave that for discussion :D

### DIFF
--- a/src/assets/css/librarybrowser.css
+++ b/src/assets/css/librarybrowser.css
@@ -28,6 +28,10 @@
     padding-top: 0 !important;
 }
 
+.itemLastPlayed {
+    text-align: right;
+}
+
 .standalonePage {
     padding-top: 4.5em !important;
 }
@@ -854,7 +858,7 @@ div.itemDetailGalleryLink.defaultCardBackground {
         width: auto;
         align-items: center;
         justify-content: center;
-        margin-top: -4.2em;
+        margin-top: -4.3em;
         position: relative;
     }
 
--- a/src/itemdetails.html
+++ b/src/itemdetails.html
@@ -113,6 +113,7 @@
                 <div class="detailPagePrimaryContent padded-left padded-right">
                     <div class="detailSection" style="margin-bottom: 0;">
                         <div class="itemMiscInfo nativeName hide"></div>
+                        <div class="itemLastPlayed hide"></div>
                         <div class="genres hide" style="margin: .7em 0;font-size:92%;"></div>
                         <div class="directors hide" style="margin: .7em 0;font-size:92%;"></div>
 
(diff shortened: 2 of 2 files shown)


JustAMan:
When we push it to top-right, can we make it so it doesn't push "Categories" element down a line?

MrTimscampi:
> When we push it to top-right, can we make it so it doesn't push "Categories" element down a line?

We have to be careful with overlap then, since there's (afaik) no limit on the amount of categories shown.

dkanada:
I actually like it this way.

ferferga:
Same as @dkanada , I think that makes it more noticeable than placing it in the same line. Take in mind that it also only appears and take the space there when it needs to be shown (stuff that already has been reproduced)

dkanada:
Wait, I got the images mixed up. I prefer it on the left side.

ferferga:
@dkanada But at the top, before ``Genres``, right?

In fact, I would put the "Played On" text and the track selectors at the top.

MrTimscampi:
I'd put it below the external links, as it's not an important piece of information for most users.

When browsing, the genres, director, synopsis, track selection and all that are more important.

Basically everything that helps you choose media and the way to play it. A "Last played on" date isn't relevant to choosing media, really, so it gets the lowest priority and goes on the bottom.

ferferga:
> I'd put it below the external links, as it's not an important piece of information for most users.
> 
> When browsing, the genres, director, synopsis, track selection and all that are more important.
> 
> Basically everything that helps you choose media and the way to play it. A "Last played on" date isn't relevant to choosing media, really, so it gets the lowest priority and goes on the bottom.

I'm for this as well

ferferga:
@dkanada I suppose this PR is deprecated because #949 , although I don't know if @MrTimscampi did something about the 'Last played' message in his PR. If he didn't, we might want to keep both PRs. Either way, I think #949 should be merged first.

MrTimscampi:
> @dkanada I suppose this PR is deprecated because #949 , although I don't know if @MrTimscampi did something about the 'Last played' message in his PR. If he didn't, we might want to keep both PRs. Either way, I think #949 should be merged first.

I removed it from my branch because I didn't really see the benefit in having that info there (It'd makke more sense in the media info, I think ?).

I'm opened to re-adding it below the links in my PR, but what's the use case for having the "Last played" time in the main details ? Isn't there a better spot we could find for it ?

ferferga:
@MrTimscampi That's why I think it should be on the right side: It isn't relevant at all to have it mixed up with the content imho. There are no submenus where it makes sense either.

dkanada:
I'm fine with removing it entirely for now and seeing how we feel about it.

ferferga:
@dkanada Then, I'll hand it to @MrTimscampi 's PR

PR #890, opened 2020-03-06
Fix itemdetails poster overlapping now playing bar

shemanaev:
**Changes**

Before -> after
img

**Issues**

Introduced in #823


ferferga:
Thank you very much, currently testing your branch. This should be done before 10.5 release

PR #898, opened 2020-03-07
Fix icon doubling

dmitrylyzo:
**WARNING about recording button**
This is blindly fixed. I checked that there is no more doubling.

**Changes**
* Fix recording button.
* Fix brightness indicator (OSD).
* Fix sort button.

**Issue**
recordingbutton-1

### DIFF
--- a/src/components/playback/brightnessosd.js
+++ b/src/components/playback/brightnessosd.js
@@ -98,15 +98,22 @@ define(['events', 'playbackManager', 'dom', 'browser', 'css!./iconosd', 'materia
         }
     }
 
+    function setIcon(iconElement, icon) {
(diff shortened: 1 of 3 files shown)


Nickbert7:
Screenshot_20200307-143121_Chrome
In the photo library it seems to be the same

dmitrylyzo:
Brightness OSD is also doubled.

dmitrylyzo:
Probably, `itemstab.js` also has doubled sort, but it seems like it is not used :confused:

And now that I'm done, I think we could just create icon dictionary (`name` -> `hex`) and drop that classList manipulation.

Nickbert7:
Sometimes in the tasks page I also get the doubled icons after a task has finished

PR #903, opened 2020-03-07
Reduced OSD timeout from 5 seconds to 3

sparky8251:
Dropped the video OSD timeout down from 5 seconds to 3.

Youtube uses a 4 second timeout and their OSD is much smaller. Our OSD tends to obscure any onscreen subtitles and we can't fix the problem for burned in subs. This is an attempt at mitigating that frustration without introducing usability issues.


Artiume:
Can we backport?

PR #904, opened 2020-03-07
Clear player data after stop

### LINKED ISSUE #800: Live TV Metadata Stale Cache
**Describe the bug**

When using LiveTV, the tv channel will sometimes pull from the previous channel's metadata. In this case, it's Doctor Who from my library that is displaying the last channel's information. 

**Screenshots**

image

 - OS: [e.g. Docker, Debian, Windows] Docker 
 - Browser: [e.g. Firefox, Chrome, Safari] Firefox
 - Jellyfin Version: [e.g. 10.0.1] Nightly

**Additional context**

There is another ticket that discussed this and how it affected DVR as well, but I can't find it.

dmitrylyzo:
_**Some warning at the end**_

When player stops, `playerData` still keeps `streamInfo` of previous item.
`videoosd` updates on bind to player and on play start.

In case of LiveTV then local movie:
* [on bind]: async request of previous item is sent
https://github.com/jellyfin/jellyfin-web/blob/2a99df8365993c0d0f02f80f11f6bd20fb697829/src/controllers/playback/videoosd.js#L290
https://github.com/jellyfin/jellyfin-web/blob/2a99df8365993c0d0f02f80f11f6bd20fb697829/src/controllers/playback/videoosd.js#L112-L126
* [on play start]: update poster (and other data) with current item
https://github.com/jellyfin/jellyfin-web/blob/2a99df8365993c0d0f02f80f11f6bd20fb697829/src/controllers/playback/videoosd.js#L290
* then response of the first request overrides poster (and other data) with previous item

With this PR, there is no update on bind.

**Changes**
* Clear player data after stop

**Issues**
Fixes #800

**Warning**
During the test I set breakpoint at
https://github.com/jellyfin/jellyfin-web/blob/2a99df8365993c0d0f02f80f11f6bd20fb697829/src/controllers/playback/videoosd.js#L273
In Firefox, for LiveTV, I got a reference to `null` - `timeupdate` events are handled as if they were in a separate thread. :confused: 
In Chrome, everything is fine.

### DIFF
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -3162,7 +3162,8 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
 
             // User clicked stop or content ended
             var state = self.getPlayerState(player);
-            var streamInfo = getPlayerData(player).streamInfo;
+            var data = getPlayerData(player);
+            var streamInfo = data.streamInfo;
 
             var nextItem = self._playNextAfterEnded ? self._playQueueManager.getNextItemInfo() : null;
 
@@ -3210,6 +3211,9 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
                 showPlaybackInfoErrorMessage(self, displayErrorCode, nextItem);
             } else if (nextItem) {
                 self.nextTrack();
+            } else {
+                // Nothing more to play - clear data
+                data.streamInfo = null;
             }
         }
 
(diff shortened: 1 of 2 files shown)


dkanada:
@dmitrylyzo if you end up fixing this before we release a patch we should definitely include it!

dmitrylyzo:
I checked Firefox 72, Chrome 79, Edge. Only Firefox has that strange behavior with breakpoint (see first message).

dkanada:
@dmitrylyzo well that sounds like a problematic bug so I think we'll have to wait then. I'll run some tests if I have time this week.

PR #907, opened 2020-03-08
Improve image loading speed and sizes

### LINKED ISSUE #858: Android App Laggs when watching Photo folder
Hello,
i noticed today when i watched my photos on my phone, that the phone was extremly lagging and need extremly much time to load my photos.

The only thing i can find in the log is:
[2020-02-25 20:18:27.510 +00:00] [INF] Getting image size for item "Photo" "/media/****.jpg"
[2020-02-25 20:18:32.324 +00:00] [WRN] HTTP Response 200 to "192.168.1.1". Time (slow): 0:00:02.0143251. "http://sub.domain.com/Items/***/Images/Primary?tag=53945fae259f872bcdbab281986eec13&quality=90"

Jellyfin Server with Docker 10.4.3 and Android 0.9.9

### LINKED ISSUE #911: Thumbnails are not used in Poster and Poster card views
I just upgraded to v10.5 and images started loading in slow. I don't remember this happening before the upgrade. I used the browser inspector to see that jellyfin is pulling in the original full size album/cover art for my music and movies in the Poster and Poster card views. The list view is correctly using thumbnails.

MrTimscampi:
**Changes**

Improves loading times and image sizes for most cards, backdrops and the details page.

**Issues**

Fixes #911 
Fixes #858 

Edit: I'm marking items done with :+1:

### DIFF
--- a/src/components/cardbuilder/cardBuilder.js
+++ b/src/components/cardbuilder/cardBuilder.js
@@ -233,9 +233,9 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
 
         function getImageWidth(shape, screenWidth, isOrientationLandscape) {
             var imagesPerRow = getPostersPerRow(shape, screenWidth, isOrientationLandscape);
-            var shapeWidth = screenWidth / imagesPerRow;
+            var shapeWidth = Math.round(screenWidth / imagesPerRow) * 2;
 
-            return Math.round(shapeWidth);
+            return shapeWidth;
         }
 
         function setCardData(items, options) {
@@ -463,6 +463,7 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
 
                 imgUrl = apiClient.getScaledImageUrl(item.Id, {
                     type: "Thumb",
+                    maxWidth: width,
                     tag: item.ImageTags.Thumb
                 });
 
@@ -470,6 +471,7 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
 
                 imgUrl = apiClient.getScaledImageUrl(item.Id, {
                     type: "Banner",
+                    maxWidth: width,
                     tag: item.ImageTags.Banner
                 });
 
@@ -477,6 +479,7 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
 
                 imgUrl = apiClient.getScaledImageUrl(item.Id, {
                     type: "Disc",
+                    maxWidth: width,
                     tag: item.ImageTags.Disc
                 });
 
@@ -484,6 +487,7 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
 
                 imgUrl = apiClient.getScaledImageUrl(item.Id, {
                     type: "Logo",
+                    maxWidth: width,
                     tag: item.ImageTags.Logo
                 });
 
@@ -491,6 +495,7 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
 
                 imgUrl = apiClient.getScaledImageUrl(item.ParentLogoItemId, {
                     type: "Logo",
+                    maxWidth: width,
                     tag: item.ParentLogoImageTag
                 });
 
@@ -498,6 +503,7 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
 
                 imgUrl = apiClient.getScaledImageUrl(item.SeriesId, {
                     type: "Thumb",
+                    maxWidth: width,
                     tag: item.SeriesThumbImageTag
                 });
 
@@ -505,6 +511,7 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
 
                 imgUrl = apiClient.getScaledImageUrl(item.ParentThumbItemId, {
                     type: "Thumb",
+                    maxWidth: width,
                     tag: item.ParentThumbImageTag
                 });
 
@@ -512,6 +519,7 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
 
                 imgUrl = apiClient.getScaledImageUrl(item.Id, {
                     type: "Backdrop",
+                    maxWidth: width,
                     tag: item.BackdropImageTags[0]
                 });
 
@@ -521,6 +529,7 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
 
                 imgUrl = apiClient.getScaledImageUrl(item.ParentBackdropItemId, {
                     type: "Backdrop",
+                    maxWidth: width,
                     tag: item.ParentBackdropImageTags[0]
                 });
 
@@ -530,6 +539,8 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
 
                 imgUrl = apiClient.getScaledImageUrl(item.Id, {
                     type: "Primary",
+                    maxHeight: height,
+                    maxWidth: width,
                     tag: item.ImageTags.Primary
                 });
 
@@ -550,6 +561,8 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
 
                 imgUrl = apiClient.getScaledImageUrl(item.PrimaryImageItemId || item.Id || item.ItemId, {
                     type: "Primary",
+                    maxHeight: height,
+                    maxWidth: width,
                     tag: item.PrimaryImageTag
                 });
 
@@ -567,12 +580,14 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
 
                 imgUrl = apiClient.getScaledImageUrl(item.ParentPrimaryImageItemId, {
                     type: "Primary",
+                    maxWidth: width,
                     tag: item.ParentPrimaryImageTag
                 });
             } else if (item.SeriesPrimaryImageTag) {
 
                 imgUrl = apiClient.getScaledImageUrl(item.SeriesId, {
                     type: "Primary",
+                    maxWidth: width,
                     tag: item.SeriesPrimaryImageTag
                 });
             } else if (item.AlbumId && item.AlbumPrimaryImageTag) {
@@ -581,6 +596,8 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
 
                 imgUrl = apiClient.getScaledImageUrl(item.AlbumId, {
                     type: "Primary",
+                    maxHeight: height,
+                    maxWidth: width,
                     tag: item.AlbumPrimaryImageTag
                 });
 
@@ -594,6 +611,7 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
 
                 imgUrl = apiClient.getScaledImageUrl(item.Id, {
                     type: "Thumb",
+                    maxWidth: width,
                     tag: item.ImageTags.Thumb
                 });
 
@@ -601,6 +619,7 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
 
                 imgUrl = apiClient.getScaledImageUrl(item.Id, {
                     type: "Backdrop",
+                    maxWidth: width,
                     tag: item.BackdropImageTags[0]
                 });
 
@@ -608,6 +627,7 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
 
                 imgUrl = apiClient.getScaledImageUrl(item.Id, {
                     type: "Thumb",
+                    maxWidth: width,
                     tag: item.ImageTags.Thumb
                 });
 
@@ -615,6 +635,7 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
 
                 imgUrl = apiClient.getScaledImageUrl(item.SeriesId, {
                     type: "Thumb",
+                    maxWidth: width,
                     tag: item.SeriesThumbImageTag
                 });
 
@@ -622,6 +643,7 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
 
                 imgUrl = apiClient.getScaledImageUrl(item.ParentThumbItemId, {
                     type: "Thumb",
+                    maxWidth: width,
                     tag: item.ParentThumbImageTag
                 });
 
@@ -629,6 +651,7 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
 
                 imgUrl = apiClient.getScaledImageUrl(item.ParentBackdropItemId, {
                     type: "Backdrop",
+                    maxWidth: width,
                     tag: item.ParentBackdropImageTags[0]
                 });
 
--- a/src/components/cardbuilder/cardBuilder.js
+++ b/src/components/cardbuilder/cardBuilder.js
@@ -2,7 +2,6 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
     function (datetime, imageLoader, connectionManager, itemHelper, focusManager, indicators, globalize, layoutManager, appHost, dom, browser, playbackManager, itemShortcuts, imageHelper) {
         'use strict';
 
-        var devicePixelRatio = window.devicePixelRatio || 1;
         var enableFocusTransform = !browser.slow && !browser.edge;
 
         function getCardsHtml(items, options) {
@@ -592,7 +591,7 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
                 });
             } else if (item.AlbumId && item.AlbumPrimaryImageTag) {
 
-                width = primaryImageAspectRatio ? Math.round(height * primaryImageAspectRatio) : null;
+                height = width && primaryImageAspectRatio ? Math.round(width / primaryImageAspectRatio) : null;
 
                 imgUrl = apiClient.getScaledImageUrl(item.AlbumId, {
                     type: "Primary",
--- a/src/components/dom.js
+++ b/src/components/dom.js
@@ -118,9 +118,7 @@ define([], function () {
         var height = window.innerHeight;
 
         if (height > width) {
-            /* If we're in portrait, compute the proper width for this height,
-               taking 16:9 as the most common ratio */
-            width = height * 1.78;
+            width = height * (16.0 / 9.0);
         }
 
         var closest = standardWidths.sort(function (a, b) {
--- a/src/components/dom.js
+++ b/src/components/dom.js
@@ -114,8 +114,8 @@ define([], function () {
 
     var standardWidths = [480, 720, 1280, 1440, 1920, 2560, 3840, 5120, 7680];
     function getScreenWidth() {
-        var width = window.innerWidth * window.devicePixelRatio;
-        var height = window.innerHeight * window.devicePixelRatio;
+        var width = window.innerWidth;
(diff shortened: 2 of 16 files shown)


dmitrylyzo:
Person cards are limited with 160px - looks ugly on TV.
https://github.com/jellyfin/jellyfin-web/blob/9a4d992c7ce5080d6f3f09d62b680d618a42c760/src/controllers/itemdetailpage.js#L1862
On Desktop, person card width is 182px, but again image is limited by 160px

dmitrylyzo:
Removed that string - much better: requests 446px.

MrTimscampi:
This needs to have videoosd.js looked at before any merge (As reported by JustJane on Matrix)

@jellyfin/web If you find more examples of pages that are super slow, please mention it here.

Already included at this points:

* Details page
* Everything that uses CardBuilder.js

Nickbert7:
I'm not sure if it is already covered, but the images within the libraries settings are not scaled.
image

MrTimscampi:
Playlist thumbnails also need a pass.

dmitrylyzo:
"Active Devices" in Dashboard

ferferga:
> 
> 
> This needs to have videoosd.js looked at before any merge (As reported by JustJane on Matrix)
> 
> @jellyfin/web If you find more examples of pages that are super slow, please mention it here.
> 
> Already included at this point:
> 
>     * Details page
> 
>     * Everything that uses CardBuilder.js

@MrTimscampi I need to deeply check everything that re-scales and test your branch, so I'm not sure if it's affected by CardBuilder, but I think that posters of the elements in the itemdetails page should be at full resolution, no matter what. Ignore me if they are.

MrTimscampi:
> I think that posters of the elements in the item details page should be at full resolution, no matter what

I respectfully disagree, because images can be MASSIVE. For example, my poster for Blade Runner 2049 is 3000x2000 and is 1MB in size. This is massive, especially on a mobile connection where data is at a premium. Even on a desktop, is significantly increased page size, which means longer loading times. We've already had a few users that have noted the negative change.

And not only does it make pages heavier, but it also creates significant aliasing, which simply looks bad.

> I'm also wondering what's your solution for window re-scaling

Resizing happens so rarely that it's honestly not worth it to optimize for that case, in my opinion. It would mean reloading all the images as soon as a resize happens, which, again, would generate a ton of data usage, when people will likely navigate away from the page.

> What if I open the window side by side another one and then, I max it out?

In this specific case, your window would be taller than it is wide, so we recalculate the full width of your window in portrait mode at a 16:9 resolution (Chosen because it's the most common for artwork). Essentially, you'd get the same artwork with two windows side-by-side than you would with one window maximized.

The issue is that it's a difficult balance to strike. We need to keep both artwork quality AND performance/data usage in mind. We cannot get rid of one for the other, hence the solution chosen here to have most artwork be 2x their resolution on the screen, in order to sidestep Skia's bad rescaling while ensuring that images don't have an enormous resolution unless they need it.

dmitrylyzo:
@MrTimscampi I think float numbers are not supported. At current moment, all is fine, but maybe round them just for sure.

MrTimscampi:
@dmitrylyzo Good point! Done ;)

dmitrylyzo:
@MrTimscampi Sorry, I guess I misled you. :sweat_smile: 
https://github.com/jellyfin/jellyfin-web/blob/76b085982b2ca0183f946d1cd5b4c3f2d00cc57e/src/libraries/apiclient/apiclientcore.js#L220-L223
And it looks like `devicePixelRatio` is also taken into account.

MrTimscampi:
@dmitrylyzo Hm, interesting.

Should we remove it from the API Client and handle it for each request? (Would be better if the user changes the window)

We can also remove it from all individual requests and make the one in ApiClient dynamic.

dmitrylyzo:
> Should we remove it from the API Client and handle it for each request? (Would be better if the user changes the window)

Thinking about possible purpose, having it in `ApiClient` might be useful for forced pixel ratio. If we ever need this.

> We can also remove it from all individual requests and make the one in ApiClient dynamic.

I have no argument against. This will make the rest of the code simpler and pixel ratio will be in single place. Dunno if someone wants pixel ratio not for images.

At current moment, we have `devicePixelRatio` at:
* your `dom.getScreenWidth`
* bunch of `maxWidth`/`maxHeight`
* `cardbuilder` but it is unused https://github.com/jellyfin/jellyfin-web/blob/76b085982b2ca0183f946d1cd5b4c3f2d00cc57e/src/components/cardbuilder/cardBuilder.js#L5

src/controllers/itemdetailpage.js
```diff
@@ -1806,7 +1818,6 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "userSetti
             require(["chaptercardbuilder"], function (chaptercardbuilder) {
                 chaptercardbuilder.buildChapterCards(item, chapters, {
                     itemsContainer: scenesContent,
-                    width: 400,
```
dmitrylyzo: I cannot check chapters but I believe that 400px will still be here
https://github.com/jellyfin/jellyfin-web/blob/76b085982b2ca0183f946d1cd5b4c3f2d00cc57e/src/components/cardbuilder/chaptercardbuilder.js#L83

dmitrylyzo:
`Music / Recommended / RecentlyPlayed` and `Music / Recommended / TopPlayed` are not limited
`width` become `null`
https://github.com/jellyfin/jellyfin-web/blob/76b085982b2ca0183f946d1cd5b4c3f2d00cc57e/src/components/cardbuilder/cardBuilder.js#L578-L587

ferferga:
> I respectfully disagree, because images can be MASSIVE. For example, my poster for Blade Runner 2049 is 3000x2000 and is 1MB in size. This is massive, especially on a mobile connection where data is at a premium. Even on a desktop, it significantly increases page size, which means longer loading times. We've already had a few users that have noted the negative change.

The negative change is obviously noticeable when there are more than 10 images that we're loading at once. But in the new detail's layout, poster images are the heart of it, they're no longer of the same size as the cards in the library view as it used to be in 10.4.3.
The same applies to backdrops. 

I don't see any issues loading 1 MB images, that's the usual size of images that are that big in webpages.

That for Desktop and TV, of course mobile can have some constraints (as that's where the data caps are really). 

> Resizing happens so rarely that it's honestly not worth it to optimize for that case, in my opinion. It would mean reloading all the images as soon as a resize happens, which, again, would generate a ton of data usage, when people will likely navigate away from the page.

That is one of my main workflows, where I have something in the left and Jellyfin's playlist or TV Show going on. I don't see any reason why anybody doesn't use it that way, I'm sure that I'm not that alone in this.

> In this specific case, your window would be taller than it is wide, so we recalculate the full width of your window in portrait mode at a 16:9 resolution (Chosen because it's the most common for artwork). Essentially, you'd get the same backdrops with two windows side-by-side than you would with one window maximized. Cards artwork size is usually computed based on their display size, but we have 2x the size displayed as a margin if the user resizes the window, which should be more than enough.

Okay, that's great, thanks for clarifying! However, this couldn't be solved ignoring completely the size of the window and taking into account the screen size?
 
> The issue is that it's a difficult balance to strike. We need to keep both artwork quality AND performance/data usage in mind. We cannot get rid of one for the other, hence the solution chosen here to have most artwork be 2x their resolution on the screen, in order to sidestep Skia's bad rescaling while ensuring that images don't have an enormous resolution unless they need it.

But, as said above, it's way too overkill to simply do it everywhere (as it was an overkill doing full everywhere). We should mix both in my opinion, just like I mentioned above: Backdrops, Poster images in item details and profile pics should be at max res imo.

I didn't say that before, but resizing profile photos is probably one of the worst ideas, at least how it was implemented before. It's better to download the full res image once and rely in browser's cache everywhere (just like the current implementation) that generating multiple sizes that can change everywhere, across one session in fact.

ferferga:
I've been trying your branch, and the only thing I can say is thank you very much for achieving this balance. It's impressive how it retains the quality while loading fast. You didn't try Jellyfin before I made the full-res PR, right? You will understand why this is so important:

However, I still don't agree with the re-scaling of the images in details section:

**Scale**
before
**After**
after

Although it still looks great and I agree this is fine for 95% of the people, it's not for the 5% of us who are so focused in the tiny details. Notice how, in the scaled version, Pennywise's face is almost unnoticeable in the globe. Blood splatters in the name are also a blurried mess. And that's without diving more in other aspects of the image.

Again, that can be good for a layout where the poster takes 50% less of space, but in this layout, it's the heart of everything. And this has been tested in Desktop with relatively high DPI, not sure what would happen with large screens with little resolution, didn't had the time to test it yet.

It's not that much of a problem as I told before, because 95% of the users will be happy about it. However, it's a single image and, in fact, I feel that re-scaling the main "showcase" for the poster would simply make choosing higher resolution in the image editor absolutely worthless.

src/controllers/itemdetailpage.js
```diff
@@ -765,41 +770,48 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "userSetti
...
             url = apiClient.getScaledImageUrl(item.Id, {
                 type: "Primary",
+                maxWidth: Math.round(dom.getScreenWidth() * 0.5),
                 tag: item.ImageTags.Primary
             });
             detectRatio = true;
         } else if (item.BackdropImageTags && item.BackdropImageTags.length) {
             url = apiClient.getScaledImageUrl(item.Id, {
                 type: "Backdrop",
+                maxWidth: Math.round(dom.getScreenWidth() * 0.5),
```
JustAMan: why `0.5`?

src/controllers/playback/videoosd.js
```diff
@@ -334,18 +334,24 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
 
             if (item) {
                 var imgUrl = seriesImageUrl(item, {
+                    maxWidth: Math.round(dom.getScreenWidth() * 0.20),
```
JustAMan: why `0.2`?

MrTimscampi: Posters on the video osd are 10% width. With the 2x size to counter Skia's blurry resize, it's 20%, or 0.2 *screen width.

src/controllers/itemdetailpage.js
MrTimscampi: The poster is 25% of the page, doubled to counter Skia.

dmitrylyzo:
Looked again why `Music / Suggestions / RecentlyPlayed` and `Music / Suggestions / TopPlayed` are unlimited. Only first item has `PrimaryImageAspectRatio` field - something wrong on server side.

At least we could make this part the same as the others.
```diff
--- a/src/components/cardbuilder/cardBuilder.js	2020-03-09 22:55:31.460475765 +0300
+++ b/src/components/cardbuilder/cardBuilder.js	2020-03-10 20:13:57.060552674 +0300
@@ -592,7 +592,7 @@
                 });
             } else if (item.AlbumId && item.AlbumPrimaryImageTag) {
 
-                width = primaryImageAspectRatio ? Math.round(height * primaryImageAspectRatio) : null;
+                height = width && primaryImageAspectRatio ? Math.round(width / primaryImageAspectRatio) : null;
 
                 imgUrl = apiClient.getScaledImageUrl(item.AlbumId, {
                     type: "Primary",

```

src/controllers/playback/videoosd.js
JustAMan: I don't like that is hardcodes some internal knowledge in a way like this.
In a month or so no one would understand this magic number at all. Also a theme (or custom CSS) can change this poster width.

Can we query real poster width instead (and multiply it by DPI) instead of doing this funny workaround thing?..

src/controllers/itemdetailpage.js
JustAMan: continues down below, closing this one as they're effectively same

src/controllers/playback/videoosd.js
dmitrylyzo: > Can we query real poster width instead (and multiply it by DPI) instead of doing this funny workaround thing?..

I think `cardbuilder` first builds urls - no card exists.

Not sure if this is possible:
* generate empty cards (with placeholders)
* set image urls

MrTimscampi: This is indeed an oversight and fixable (at least in this case, I need to check the others).

My suggestion is:

```suggestion
                    maxWidth: Math.round(dom.getScreenWidth() * osdPoster.clientWidth),
```

**However**, this needs to be tested AND it will eventually need a polyfill for early WebOS.

dmitrylyzo: Huge poster?

MrTimscampi: Details page build the poster before, so we can use the same technique. 

CardBuilder computes the number of cards per rows here: https://github.com/jellyfin/jellyfin-web/blob/039728359914d18765a78f22b2aaf73fc9a1aebf/src/components/cardbuilder/cardBuilder.js#L17

It then uses that to find out the size of a card, so it is already dynamic (But it also uses a ton of hardcoded values)

MrTimscampi: More info on the support for the solution proposed: https://caniuse.com/#feat=mdn-api_element_clientwidth

MrTimscampi: > > `dom.getScreenWidth() * osdPoster.clientWidth`
> 
> Huge poster?

Indeed, I'm dumb and I'm working on something else, so I only gave this a cursory glance :smile: 

It'd be:

```suggestion
                    maxWidth: osdPoster.clientWidth),
```

src/components/cardbuilder/cardBuilder.js
```diff
@@ -530,6 +538,8 @@ define(['datetime', 'imageLoader', 'connectionManager', 'itemHelper', 'focusMana
 
                 imgUrl = apiClient.getScaledImageUrl(item.Id, {
                     type: "Primary",
+                    maxHeight: height,
+                    maxWidth: width,
```
dkanada: Why do some of them get the height as well? Shouldn't one of them be enough?

src/controllers/dashboard/dashboard.js
```diff
@@ -636,20 +636,23 @@ define(["datetime", "events", "itemHelper", "serverNotifications", "dom", "globa
...
                 return ApiClient.getScaledImageUrl(item.Id, {
+                    maxWidth: Math.round(dom.getScreenWidth() * 0.20),
                     type: "Backdrop",
                     tag: item.BackdropImageTags[0]
                 });
             }
 
             if (item && item.ParentBackdropImageTags && item.ParentBackdropImageTags.length) {
                 return ApiClient.getScaledImageUrl(item.ParentBackdropItemId, {
+                    maxWidth: Math.round(dom.getScreenWidth() * 0.20),
```
dkanada: Almost feel like we should include this in the API client to avoid so much duplication.

src/components/cardbuilder/cardBuilder.js
MrTimscampi: I think it's for everything that doesn't fall into anything else. Doesn't fall into any specific category, so just display whatever, I think ?

src/controllers/dashboard/dashboard.js
MrTimscampi: Not everything uses the screen width as a base though. Ideally, we'd get rid of all this for some sensible logic once moving to a framework.

MrTimscampi:
@dmitrylyzo @JustAMan @dkanada I made the changes in ApiClient (making the pixel scale dynamic) and adjusted the cards, but I couldn't manage to reliably get the size of the poster on the item details page.

If you guys find a way, I'm all for it.

src/components/dom.js
```diff
@@ -112,6 +112,24 @@ define([], function () {
...
 
+    var standardWidths = [480, 720, 1280, 1440, 1920, 2560, 3840, 5120, 7680];
+    function getScreenWidth() {
+        var width = window.innerWidth;
+        var height = window.innerHeight;
+
+        if (height > width) {
+            /* If we're in portrait, compute the proper width for this height,
+               taking 16:9 as the most common ratio */
+            width = height * 1.78;
```
JustAMan:
```suggestion
            width = height * 16.0 / 9.0;
```
for clarity?

src/controllers/dashboard/dashboard.js
JustAMan: Why this `0.20` is still there? I thought you said you changed all the things...

src/libraries/apiclient/apiclientcore.js
```diff
@@ -58,7 +58,7 @@ define(["events", "appStorage"], function(events, appStorage) {
         return request.data && ("string" == typeof request.data ? fetchRequest.body = request.data : (fetchRequest.body = paramsToString(request.data), contentType = contentType || "application/x-www-form-urlencoded; charset=UTF-8")), contentType && (headers["Content-Type"] = contentType), request.timeout ? fetchWithTimeout(request.url, fetchRequest, request.timeout) : fetch(request.url, fetchRequest)
     }
 
-    function ApiClient(serverAddress, appName, appVersion, deviceName, deviceId, devicePixelRatio) {
+    function ApiClient(serverAddress, appName, appVersion, deviceName, deviceId) {
```
JustAMan: this should probably be ported over to standalone js client?..

src/components/dom.js
MrTimscampi: I think 1.78 is fairly standard.

What does @jellyfin/web think ?

src/libraries/apiclient/apiclientcore.js
MrTimscampi: Yep, will do once this is merged.

MrTimscampi:
@JustAMan Seems like I can't answer one of your review comments.

I didn't remove all of it, as said in my comment above.

For the details screen, I couldn't reliably get the size of of the poster and asked if anyone had an idea.

For the playback card there, it's impossible to get a size, because at that point in the code, the element doesn't exist yet. It needs a major refactor to be able to find its size dynamically, I believe.

src/components/dom.js
dkanada: I wouldn't have known, so this makes more sense to me. You could probably remove the comment with this change.

src/controllers/dashboard/dashboard.js
JustAMan: If it's currently impossible to compute, at least add a comment explaining what this `0.20` means and how it is generated, and a `// TODO` to fix it later.

src/controllers/dashboard/dashboard.js
```diff
@@ -634,6 +634,8 @@ define(["datetime", "events", "itemHelper", "serverNotifications", "dom", "globa
             return "<img src='" + iconUrl + "' />";
         },
         getNowPlayingImageUrl: function (item) {
+            /* Screen width is multiplied by 0.2, as the there is now way to get the width of
+            elements that aren't crated yet. */
```
Shadowghost: Some mistakes sneaked into your sentence 😄 
```
            /* Screen width is multiplied by 0.2 as there is no way to get the width of
            elements that aren't created yet. */
```

src/components/listview/listview.js
```diff
@@ -72,7 +72,7 @@ define(['itemHelper', 'mediaInfo', 'indicators', 'connectionManager', 'layoutMan
         var apiClient = connectionManager.getApiClient(item.ServerId);
 
         var options = {
-            width: width,
+            maxWidth: width,
```
dmitrylyzo: We get 80px for playlist icons and 500px for TV show episodes (in TV layout, 83px and 576px).
```suggestion
            maxWidth: width * 2,
```
But this won't help much - images for episodes are 400-600px.

dmitrylyzo:
Another one https://github.com/jellyfin/jellyfin-web/blob/291f0b1cac0a059624d1c411243db2b4c8ffbc30/src/components/remotecontrol/remotecontrol.js#L124-L128

src/components/listview/listview.js
```diff
@@ -105,7 +105,6 @@ define(['itemHelper', 'mediaInfo', 'indicators', 'connectionManager', 'layoutMan
         var apiClient = connectionManager.getApiClient(item.ServerId);
 
         var options = {
-            width: width,
```
dmitrylyzo: `maxWidth: width * 2,`
Potentially, this will be the same as `getImageUrl`: 80px or 500px.
I can’t verify this - need for normal LiveTV tuner.

PR #908, opened 2020-03-08
Make item detail overview expandable (Fix #888)

### LINKED ISSUE #888: Limit person bio length
Currently, opening an artist page shows the full bio, which can be quite long and is especially noticeable on narrow screens. A correct way of handling this would be to limit bio length to a sane number and add a `Read more` button, that expands the bio to full length (or maybe shows a nice pop-up).

**Screenshots**
image

 - OS: Arch Linux
 - Browser: Qutebrowser
 - Jellyfin Version: 10.5.0 (nightly)

masterkoppa:
**Changes**
Adds a `onClick` action that toggles expanding the item detail Overview. By default it crops the text to 12 lines on mobile, 6 on larger viewports.

**Issues**
Fixes #888

**Notes**
This is my first PR for jellyfin, so any additional feedback is welcome :)

I would love some guidance on the location of the callback or CSS classes.

`-webkit-line-clamp` is supported on all Webkit browsers and Firefox as of 3/2020. A fallback with max-height is provided as well, just missing the ellipsis. This might be confusing without a better UX or hints.

**Collapsed Screenshot**
image

**Expanded Screenshot**
image

**Known issues**
1) Show More / Show Less does not dynamically change visibility after a window resize event. This might be problematic.

### DIFF
--- a/src/assets/css/librarybrowser.css
+++ b/src/assets/css/librarybrowser.css
@@ -1145,6 +1145,11 @@ div:not(.sectionTitleContainer-cards) > .sectionTitle-cards {
     font-size: 1.4em;
 }
 
+.overview-controls {
+    display: flex;
+    justify-content: flex-end;
+}
+
 .detail-clamp-text {
     overflow: hidden;
     display: -webkit-box;
--- a/src/assets/css/librarybrowser.css
+++ b/src/assets/css/librarybrowser.css
@@ -1150,16 +1150,10 @@ div:not(.sectionTitleContainer-cards) > .sectionTitle-cards {
     display: -webkit-box;
     -webkit-line-clamp: 12;
     -webkit-box-orient: vertical;
-
-    /* Fallback for older browsers: line-height * number of lines */
-    max-height: calc(1.35 * 12em);
 }
 
 @media all and (min-width: 40em) {
     .detail-clamp-text {
         -webkit-line-clamp: 6;
-
-        /* Fallback for older browsers: line-height * number of lines */
-        max-height: calc(1.35 * 6em);
     }
 }
--- a/src/assets/css/librarybrowser.css
+++ b/src/assets/css/librarybrowser.css
@@ -1144,3 +1144,22 @@ div:not(.sectionTitleContainer-cards) > .sectionTitle-cards {
     margin-top: 0;
     font-size: 1.4em;
 }
+
+.detail-clamp-text {
+    overflow: hidden;
+    display: -webkit-box;
+    -webkit-line-clamp: 12;
+    -webkit-box-orient: vertical;
+
+    /* Fallback for older browsers: line-height * number of lines */
+    max-height: calc(1.35 * 12em);
+}
+
+@media all and (min-width: 40em) {
+    .detail-clamp-text {
+        -webkit-line-clamp: 6;
+
+        /* Fallback for older browsers: line-height * number of lines */
+        max-height: calc(1.35 * 6em);
+    }
+}
--- a/src/controllers/itemDetails.js
+++ b/src/controllers/itemDetails.js
@@ -972,14 +972,16 @@ define(['loading', 'appRouter', 'layoutManager', 'connectionManager', 'userSetti
         }
     }
 
-    function toggleLineClamp(e) {
-        var target = e.target;
+    function toggleLineClamp(clampTarget, e) {
+        var expandButton = e.target;
         var clampClassName = 'detail-clamp-text';
 
-        if (target.classList.contains(clampClassName)) {
-            target.classList.remove(clampClassName);
+        if (clampTarget.classList.contains(clampClassName)) {
+            clampTarget.classList.remove(clampClassName);
+            expandButton.innerHTML = 'Show Less';
         } else {
-            target.classList.add(clampClassName);
+            clampTarget.classList.add(clampClassName);
+            expandButton.innerHTML = 'Show More';
         }
     }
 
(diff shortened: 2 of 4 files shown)


dkanada:
We have a custom element called `emby-accordion` or something that you could maybe use here. We could also make a new custom element, but I would rather do something like that.

src/assets/css/librarybrowser.css
```diff
@@ -1122,3 +1122,22 @@ div:not(.sectionTitleContainer-cards) > .sectionTitle-cards {
 .itemsViewSettingsContainer > .button-flat {
     margin: 0;
 }
+
+.detail-clamp-text {
+    overflow: hidden;
+    display: -webkit-box;
```
JustAMan: why prefixed `display`? we have `gulp` building now that should add all needed prefixes

dkanada: It hasn't happened yet, that's all.

dmitrylyzo:
Personally, I worry about how to use it with only keyboard (on TV layout).
Open dialog with scrollable textarea. :thinking:

src/assets/css/librarybrowser.css
dkanada: What I meant is that we haven't quite updated our build process to remove all these again since the changes.

src/assets/css/librarybrowser.css
```diff
@@ -1122,3 +1122,22 @@ div:not(.sectionTitleContainer-cards) > .sectionTitle-cards {
...
+    /* Fallback for older browsers: line-height * number of lines */
+    max-height: calc(1.35 * 12em);
+}
+
+@media all and (min-width: 40em) {
+    .detail-clamp-text {
+        -webkit-line-clamp: 6;
+
+        /* Fallback for older browsers: line-height * number of lines */
+        max-height: calc(1.35 * 6em);
```
MrTimscampi, line 1141: It shouldn't need a fallback, as it's supported by everything we target: https://caniuse.com/#search=line-clamp (See the Browserslist key in package;json)

src/controllers/itemdetailpage.js
```diff
@@ -982,6 +993,10 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "userSetti
             if (overview) {
                 elem.innerHTML = overview;
                 elem.classList.remove("hide");
+
+                elem.classList.add("detail-clamp-text");
+                elem.addEventListener('click', toggleLineClamp);
```
MrTimscampi, line 998: It does not work on TV, as the element is not focusable.

Ideally, there should be a "Show more" link-button that unveils the element.

masterkoppa:
Apologies for the radio silence here, with everything going on this fell by the wayside.

I've removed the unneeded fallback, and I'm looking to see what it would take to make the interaction work with a `link-button`.

masterkoppa:
I've implemented the same behavior, using a `emby-linkbutton` that is adjacent to the overview paragraph.

Collapsed Screenshot
image

Expanded Screenshot
image

There's is only two issues I see with the change so far:

1) I noticed this debug log entry that might be important,
```
inputManager: tried to process command with no action assigned: unknown
```

Not exactly sure what it means, or how to address it. But it seems important to get right from what I could see. (Related to keyboard / tv controls?)

2) The Show More / Show Less button is not responsive to window size changes. I suspect it's not a big deal, but wanted to check before I find a solution.

I was thinking of using something like [`ResizeObserver`](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) to listen to changes in the dimensions of the overview and recalculating overflow that way.

dmitrylyzo:
>     1. I noticed this debug log entry that might be important,

Debug message was added recently. I think you can ignore it.
`inputManager` watches for user actions to determine idle time. At current moment, screensaver uses it.
Probably, this event (`unknown`) could be used by listeners that connected with `inputManager.on`.

https://github.com/jellyfin/jellyfin-web/blob/d5c7078c3f25ef95b9fcbba9fbd24df026a63a37/src/scripts/inputManager.js#L11-L14
https://github.com/jellyfin/jellyfin-web/blob/d5c7078c3f25ef95b9fcbba9fbd24df026a63a37/src/scripts/inputManager.js#L241-L243

masterkoppa:
Rebased against the lastest master to resolve conflicts.

MrTimscampi:
I'm still not able to select the button when using the TV layout.

dmitrylyzo:
> I'm still not able to select the button when using the TV layout.

Need to add attribute `href="#"`

masterkoppa:
@MrTimscampi 
What's the best way to test the TV Layout, I meant to ask earlier how to do so?

Also, the label does get updated dynamically to `Show Less` should we update that to be translated as well? (Not sure how the translation tool you're using works)

MrTimscampi:
> What's the best way to test the TV Layout, I meant to ask earlier how to do so?

Sorry for the late response. You can go to Settings > Display > Layout and set that to TV.

Make sure to refresh the page, as currently everything doesn't reload properly.

> Also, the label does get updated dynamically to Show Less should we update that to be translated as well? (Not sure how the translation tool you're using works)

Yes. You can call `globalize.translate()` with the translation key as a parameter, that will handle the translation for you.

masterkoppa:
@MrTimscampi 

Thanks for the advice on globalize and the TV layout. I'll keep it in mind next time!

I've rebased against the latest master and rechecked things, everything still looks good. Feel free to merge whenever you feel it's appropriate and thanks for being patient with all the back and forth!

src/controllers/itemDetails.js
```diff
@@ -972,6 +972,19 @@ define(['loading', 'appRouter', 'layoutManager', 'connectionManager', 'userSetti
         }
     }
 
+    function toggleLineClamp(clampTarget, e) {
+        var expandButton = e.target;
+        var clampClassName = 'detail-clamp-text';
+
+        if (clampTarget.classList.contains(clampClassName)) {
+            clampTarget.classList.remove(clampClassName);
+            expandButton.innerHTML = globalize.translate('Show Less');
```
dmitrylyzo, line 981: as in `en-us.json`
```suggestion
            expandButton.innerHTML = globalize.translate('ShowLess');
```

src/controllers/itemDetails.js
```diff
@@ -972,6 +972,19 @@ define(['loading', 'appRouter', 'layoutManager', 'connectionManager', 'userSetti
...
+        } else {
+            clampTarget.classList.add(clampClassName);
+            expandButton.innerHTML = globalize.translate('Show More');
```
dmitrylyzo, line 984: as in `en-us.json`
```suggestion
            expandButton.innerHTML = globalize.translate('ShowMore');
```

src/strings/en-us.json
```diff
@@ -1373,6 +1373,7 @@
     "Share": "Share",
     "ShowAdvancedSettings": "Show advanced settings",
     "ShowIndicatorsFor": "Show indicators for:",
+    "ShowMore": "Show more",
```
dmitrylyzo, line 1376:
```suggestion
    "ShowLess": "Show less",
    "ShowMore": "Show more",
```

masterkoppa:
Good catch @dmitrylyzo, thanks for the suggestions.

PR #909, opened 2020-03-08
[WIP] Add loading placeholders to cards - Part 1

MrTimscampi:
**Changes**

Adds a loading placeholder to cards. Icon is SVG to be scalable.

image

image

image

Some cards don't have a perfectly centered icon and need a bit more refactoring, but this is a nice start.

**Issues**

None

### DIFF
--- a/src/assets/img/icon-transparent-grey.svg
+++ b/src/assets/img/icon-transparent-grey.svg
@@ -1,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   version="1.1"
+   viewBox="0 0 512 512"
+   id="icon-transparent-white">
+  <metadata
+     id="metadata12">
+    <rdf:RDF>
+      <cc:Work
+         rdf:about="">
+        <dc:format>image/svg+xml</dc:format>
+        <dc:type
+           rdf:resource="http://purl.org/dc/dcmitype/StillImage" />
+        <dc:title>icon-transparent-white</dc:title>
+      </cc:Work>
+    </rdf:RDF>
+  </metadata>
+  <defs
+     id="defs10" />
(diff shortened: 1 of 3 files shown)


dkanada:
I think we should match the media type cards used in other sections.

KristupasSavickas:
How does this work with your endless scroll PR?

ferferga:
@dkanada I completely disagree with you. With the Jellyfin icon, you can distinguish quickly if the item has an image (but hasn't loaded yet) between items that don't have an image at all.

@MrTimscampi Why is this WIP btw? I tried it back in the day you published it, but I remember seeing it pretty finished.

@KristupasSavickas That PR only changes the amount of items that will be displayed on screen. Images still would need to be downloaded from the server.

KristupasSavickas:
@ferferga Yeah, I get that, what I wondered was if library scrolling was faster, i.e. cards with placeholder graphics were loaded before images were as you kept scrolling.

dmitrylyzo:
> I completely disagree with you. With the Jellyfin icon, you can distinguish quickly if the item has an image (but hasn't loaded yet) between items that don't have an image at all.

I think @dkanada means darkened version of media type icon.
test1
_there is already a marker on the photo item_

test2

MrTimscampi:
> @MrTimscampi Why is this WIP btw? I tried it back in the day you published it, but I remember seeing it pretty finished.

There's a few cases where it looks wrong and I haven't taken the time to go back to it.

> what I wondered was if library scrolling was faster, i.e. cards with placeholder graphics were loaded before images were as you kept scrolling.

This is already the case currently. Only the visible images will load when a library is shown. As you scroll, more will load as they get into view.  
The big issue currently is that images won't unload when they leave the viewport. The PR to add limit setting (that enables you to load the full library if set to 0) will work the same way in that regard, for now. The team as a whole wants to improve the image loader, though, but no one has gotten to it at the moment.

> I think @dkanada meant darkened version of media type icon.

That does look really good.

dkanada:
I didn't mean that but I like your idea better, so let's roll with it!

ferferga:
@dmitrylyzo I like yours more than the original idea. You're still able to distinguish items that aren't fully loaded while also being able to know the type of their item, Both things at once!

I vote for it as well :)

MrTimscampi:
After looking a bit into this, there's a simple-ish way to do it, but it required more refactoring to the card building code than I'm willing to do right now. And with the large refactoring planned, it's probably not the best to spend time rewriting the card builder...

Here's what it'd look like with the current code:

image

PR #920, opened 2020-03-10
Fix slideshow 2

dmitrylyzo:
Another problems with slideshow.

**1. There is no escape**
* Turn on TV layout
* Run slideshow
* Wait for OSD (at bottom) disappeared
* Push `Escape` (in TV, it fires `back` command)
* `appRouter` goes back but slideshow is still on top

The problem is that `inputManager` sends events from activeElement which is `body` after OSD disappears.

**2. Nested slideshow**
* Turn on TV layout
* Run slideshow by keyboard (OSD should not appear)
* Push `Enter` again to run nested slideshow

The problem is that focus stays on card if OSD does not appear.

**Changes**
* Add opened dialog as inputManager event source
* Prevent nested slideshow

### DIFF
--- a/src/components/slideshow/slideshow.js
+++ b/src/components/slideshow/slideshow.js
@@ -188,11 +188,6 @@ define(['dialogHelper', 'inputManager', 'connectionManager', 'layoutManager', 'f
                 stopInterval();
             });
 
-            // Blur foreign element to prevent starting of "nested" slideshow
-            if (document.activeElement && !dlg.contains(document.activeElement)) {
-                document.activeElement.blur();
-            }
-
             inputManager.on(window, onInputCommand);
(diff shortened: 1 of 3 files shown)


src/components/slideshow/slideshow.js
```diff
@@ -188,6 +188,11 @@ define(['dialogHelper', 'inputManager', 'connectionManager', 'layoutManager', 'f
                 stopInterval();
             });
 
+            // Blur foreign element to prevent starting of "nested" slideshow
+            if (document.activeElement && !dlg.contains(document.activeElement)) {
+                document.activeElement.blur();
+            }
```
dkanada: This can happen with any dialog. Maybe the best place to add this change would be the dialog component instead.

PR #929, opened 2020-03-11
Fix Network Settings Page

mark-monteiro:
**Changes**
Do not validate HTTPS certificate settings when remote connections are disabled.

This fixes a bug where network settings cannot be saved when remote connections are disabled, because the certificate validation check always fails. When remote connections are disabled, external HTTPS settings are not required or relevant (they're also not even shown on the configuration page unless the remote connection checkbox is checked), so they do not need to be validated.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3519,11 +3519,6 @@ is-wsl@^1.1.0:
   resolved "https://registry.yarnpkg.com/is-wsl/-/is-wsl-1.1.0.tgz#1f16e4aa22b04d1336b66188a66af3c600c3a66d"
   integrity sha1-HxbkqiKwTRM2tmGIpmrzxgDDpm0=
 
-isarray@0.0.1:
-  version "0.0.1"
-  resolved "https://registry.yarnpkg.com/isarray/-/isarray-0.0.1.tgz#8a18acfca9a8f4177e09abfc6038939b05d1eedf"
-  integrity sha1-ihis/Kmo9Bd+Cav8YDiTmwXR7t8=
-
 isarray@1.0.0, isarray@^1.0.0, isarray@~1.0.0:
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -3519,6 +3519,11 @@ is-wsl@^1.1.0:
   resolved "https://registry.yarnpkg.com/is-wsl/-/is-wsl-1.1.0.tgz#1f16e4aa22b04d1336b66188a66af3c600c3a66d"
```
mark-monteiro: I included the changes to `lock.json` that should have been added with PR #871. Can remove this if necessary. Maybe this fix could be backported if this is removed?

dkanada:
Is this remote as in WAN or anything outside localhost, because that would make a difference.

mark-monteiro:
> Is this remote as in WAN or anything outside localhost, because that would make a difference.

The setting is this one:
image

From what I can tell from the server-side code, setting this to false disables UPNP and also rejects any requests that appear to be from outside the local network.

dkanada:
@mark-monteiro those two features definitely need to be split into separate options at some point.

yarn.lock
dkanada: I think you're right so I would remove the changes in this file.

mark-monteiro: Reverted the changes to the lock file

mark-monteiro:
> @mark-monteiro those two features definitely need to be split into separate options at some point.

UPnP is already a separate setting, but both settings are used to check if port forwarding should execute: https://github.com/jellyfin/jellyfin/blob/master/Emby.Server.Implementations/EntryPoints/ExternalPortForwarding.cs#L91

In any case, I think making a change to how that config works would be out scope for this PR

PR #932, opened 2020-03-11
Show hamburger menu on mobile and fix title padding

MrTimscampi:
**Changes**

Fix title padding on mobile and shows the hamburger menu icon even on mobile.

**Issues**

### DIFF
--- a/src/assets/css/librarybrowser.css
+++ b/src/assets/css/librarybrowser.css
@@ -115,7 +115,7 @@
     display: -webkit-inline-box;
     display: -webkit-inline-flex;
     display: inline-flex;
-    margin: 0.3em 0 0 0.5em;
+    margin: 0 0 0 0.5em;
     height: 1.7em;
     -webkit-box-align: center;
     -webkit-align-items: center;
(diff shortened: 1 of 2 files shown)


hamburglar2160:
With the hamburger menu visible on all interfaces the Home button can be hidden/removed as that button is present in the top of the menu.

dkanada:
We might want to remove the icon for mobile layouts when the hamburger menu is visible.

MrTimscampi:
> We might want to remove the icon for mobile layouts when the hamburger menu is visible.

Which icon ?

dkanada:
The project logo and name, which both take up a lot of space on the header.

MrTimscampi:
This is what it looks like on the smallest phone available in Firefox:

image

I think it looks fine for now, but I'm open to having only the icon without "Jellyfin" if needed.

dkanada:
What does this look like on other pages that would normally include a back button and home button?

MrTimscampi:
It's a little janky on regular screens:

image

I'd probably remove the home button (There isn't a lot of UI depth anyway) and hide the hamburger on screens other than home, like most mobile apps do.

dkanada:
@MrTimscampi that sounds fine to me. I'm not sure it will be ideal but let's test it out and see what it feels like.

PR #936, opened 2020-03-12
Fix "Copy Stream URL" for iOS.

### LINKED ISSUE #935: Copy Stream url not workin in iOS safari/chrome/firefox
**Describe the bug**
Tapping "Copy Stream URL" does nothing.

**To Reproduce**

1. Use Safari/Chrome/Firefox using an iPad/iPhone
2. Click on any media menu
3. Click on "Copy Stream URL"

**Expected behavior**
It should present a dialogue to copy the Stream URL 

 - OS: iPadOS 13.3.1
 - Browser: Firefox, Chrome, Safari
 - Jellyfin Version: 10.5.0

**Additional context**
This bug was introduced in 10.5.0, I believe from commit https://github.com/jellyfin/jellyfin-web/commit/0d0f210c6a2c737e0bc892106a6d2c59b8f1215d .
`navigator.clipboard` is `undefined` on [iOS safari][1]  so the fallback code never gets run.
Moving the fallback code outside `navigator.clipboard.writeText` fixes the issue.

[1]: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard

macr:
`navigator.clipboard` is `undefined` on iOS safari so the fallback code inside `navigator.clipboard.writeText`  will not trigger on iOS.

**Changes**
Added checking `undefined` `navigator.clipboard` and then running the clipboard copy fallback code there as well.
**Issues**
Fixes #935

### DIFF
--- a/src/components/itemcontextmenu.js
+++ b/src/components/itemcontextmenu.js
@@ -346,7 +346,7 @@ define(["apphost", "globalize", "connectionManager", "itemHelper", "appRouter", 
                     break;
                 case "copy-stream":
                     var downloadHref = apiClient.getItemDownloadUrl(itemId);
-                    var textareaCopy = function () {
+                    var textAreaCopy = function () {
                         var textArea = document.createElement("textarea");
                         textArea.value = downloadHref;
                         document.body.appendChild(textArea);
--- a/src/components/itemcontextmenu.js
+++ b/src/components/itemcontextmenu.js
@@ -346,11 +346,7 @@ define(["apphost", "globalize", "connectionManager", "itemHelper", "appRouter", 
                     break;
                 case "copy-stream":
                     var downloadHref = apiClient.getItemDownloadUrl(itemId);
-                    navigator.clipboard.writeText(downloadHref).then(function () {
-                        require(["toast"], function (toast) {
-                            toast(globalize.translate("CopyStreamURLSuccess"));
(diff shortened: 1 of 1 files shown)


src/components/itemcontextmenu.js
```diff
@@ -346,11 +346,7 @@ define(["apphost", "globalize", "connectionManager", "itemHelper", "appRouter",
                     break;
                 case "copy-stream":
                     var downloadHref = apiClient.getItemDownloadUrl(itemId);
-                    navigator.clipboard.writeText(downloadHref).then(function () {
-                        require(["toast"], function (toast) {
-                            toast(globalize.translate("CopyStreamURLSuccess"));
-                        });
-                    }, function () {
+                    var textareaCopy = function () {
```
MrTimscampi:
```suggestion
                    var textAreaCopy = function () {
```

Use proper lowerCamelCase please.

Otherwise, this looks okay.

MrTimscampi:
```
/home/vsts/work/1/s/src/components/itemcontextmenu.js
  349:25  warning  'textAreaCopy' is assigned a value but never used  no-unused-vars
  365:25  error    'textareaCopy' is not defined                      no-undef
  371:28  error    'textareaCopy' is not defined                      no-undef
```

The other references need to be fixed as well :)

PR #937, opened 2020-03-12
Fix radio style

### LINKED ISSUE #536: No focus marker on unchecked radiobuttons
**Description**
No focus marker on unchecked radiobuttons.

**To Reproduce**
1. Turn on TV layout.
2. Open page "TV Shows"/"Movies".
3. Click on "Sort".
4. Navigate with keyboard to change focus.

**System**
Browser/TV: Firefox 69.0.2, Tizen 4.0

**My Investigations**
CSS style for focus marker is shadow for inner circle. When the radiobutton is unchecked, inner circle has _transform: scale3d(0,0,0)_, and shadow not shown.
jellyfin-web_issue02_01
jellyfin-web_issue02_02

dmitrylyzo:
:radio: :notes: 

I tried to use `em` units, but due to rounding errors, inner circle was off-centered. So I turned to inline SVG. There is still some warping (Desktop NEW "Parental rating"), but centered at least.

I had to place focus circle outside because WebOS 3 does not support `overflow` for SVG.

Desktop
radio

TV
radio_tv

Checkbox (TV)
checkbox

**Changes**
* Fix radiobutton focus marker
* Fix checkbox border on TV (_just `em` units_)

**Issues**
Fixes #536

### DIFF
--- a/src/elements/emby-radio/emby-radio.css
+++ b/src/elements/emby-radio/emby-radio.css
@@ -46,6 +46,7 @@
     top: 0;
     left: 0;
     z-index: 1;
+    overflow: visible;
 }
 
 .mdl-radio__button:disabled + .mdl-radio__circles {
--- a/src/elements/emby-radio/emby-radio.css
+++ b/src/elements/emby-radio/emby-radio.css
@@ -4,7 +4,6 @@
     display: inline-block;
     box-sizing: border-box;
     margin: 0;
-    padding-left: 24px;
 }
 
 .radio-label-block {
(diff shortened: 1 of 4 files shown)


dmitrylyzo:
Not sure if I should remove focus on non-TV.

dkanada:
I think focus should only show up on TV layouts. We could utilize that global class again here, forget what the name was.

MrTimscampi:
> I think focus should only show up on TV layouts. We could utilize that global class again here, forget what the name was.

Disagree a little, for accessibility.

Ideally, we should strive to improve the accessibility of the UI for impaired users, and I think having focus for elements could be a good thing.

We can eventually build an "Accessibility" preference menu in the user settings to allow enabling/disabling it, too.

dmitrylyzo:
Focus on Desktop activates radiobutton https://github.com/jellyfin/jellyfin-web/issues/536#issuecomment-588517767

On the other hand, if you navigate with Tab then focus marker is required :confused: 
In example above, with arrow keys you switch between options in the group.

dmitrylyzo:
At least I can turn off "show focus" in Mobile.
mobile_no_focus

tbh, I am still not satisfied how it looks - "Date Played" is squashed in Chrome.

Firefox beats Chrome.
mobile_no_focus-firefox

dmitrylyzo:
I decided to turn off focus on Mobile since it disappears after tap.

MrTimscampi:
@dmitrylyzo Is this ready for review?

dmitrylyzo:
@MrTimscampi yes, formally it is ready. I just don't know how to workaround rounding error issues with `em` units: squashed and stretched circles.

_I rebased it onto recent master_

MrTimscampi:
> @MrTimscampi yes, formally it is ready. I just don't know how to workaround rounding error issues with `em` units: squashed and stretched circles.
> 
> _I rebased it onto recent master_

Thanks :)

I don't know how to fix rounding errors either (aside from using px, but I really dislike that), I had some issues with it when reworking the details page as well.

We can always improve things once we find a fix.

dmitrylyzo:
The fun is that Firefox has a perfect look. :heart_eyes:

src/elements/emby-radio/emby-radio.css
```diff
@@ -31,67 +30,82 @@
...
-    transition-timing-function: cubic-bezier(0.4, 0, 0.2, 1);
+    transition-duration: 0.2s;
     transition-property: -webkit-transform;
     transition-property: transform;
     transition-property: transform, -webkit-transform;
-    -webkit-transform: scale3d(0, 0, 0);
-    transform: scale3d(0, 0, 0);
-    border-radius: 50%;
-    background: #00a4dc;
+    -webkit-transform: scale(0);
```
dkanada: This is unrelated but we should be able to remove these soon, right @MrTimscampi?

MrTimscampi: We should be able to remove them without issue now, yeah.

Needs a bit of testing to be sure (don't want to have to revert it again :p) but it should work since every CSS file goes through postcss now..

PR #938, opened 2020-03-12
Remove query string from source

dkanada:


anthonylavado:
@dkanada Merge conflict now. Also, what did this do? 🙃

MrTimscampi:
> @dkanada Merge conflict now. Also, what did this do? upside_down_face

I believe it's the polyfill added in the move to Gulp :p

We both edited the same file at the same point, adding one more block of code.

PR #939, opened 2020-03-12
Fix Dashboard OS for German

Artiume:
https://www.reddit.com/r/jellyfin/comments/fh0d8z/1005_admin_panel_bug_operating_system_definition/

**Changes**

**Issues**


Artiume:
I updated it via weblate if we want to approve it via weblate instead

joshuaboniface:
I can't seem to backport this one, not sure what's going on there. It might have to wait until 10.6.0.

PR #949, opened 2020-03-15
Improve details page informations and selects

MrTimscampi:
**Changes**

Changes the way labels and info works on the details screen.

image

**Issues**

None


ferferga:
THANK YOU VERY MUCH FOR DOING THIS! Really, thank you! The attempt I did with the padding before 10.5 was a quick one which turned to be even worse for mobile layouts (selectors were moved to the right and misaligned from the text when wrapped), so I'm very grateful that you finally solved what was driving me that crazy!

If I would change one thing, it would be the width of those selectors, they are really big imo. Which turns to be strange when you have as many selectors as me in a lot of items:

image

Also, notice that some padding is needed between them.

Take this out from draft soon please, it's a much needed thing :)

Artiume:
Would it be feasible to add the `bit depth` to the codec? stating it's 8bit or 10bit.

YouKnowBlom:
They're a bit too wide for my taste, what do you think @dkanada?. Aside from that, all looks good 👍

MrTimscampi:
> They're a bit too wide for my taste, what do you think @dkanada?. Aside from that, all looks good +1

This is to make subtitles with titles like this readable:

`Commentary by co-producers Michael Wilson, Barbara Broccoli and various crew members`

dkanada:
What we really need for the select menus is to more closely match the Plex style, which looks very nice even for wider elements. That can be for another time though.

MrTimscampi:
> What we really need for the select menus is to more closely match the Plex style, which looks very nice even for wider elements. That can be for another time though.

Agreed, the main issue is that the select simply suck. I wrestled them into submission somewhat, here, but they're not ideal.

ferferga:
@MrTimscampi I see your point (and in fact I was aware beforehand that you were thinking about long track names). However, don't you think it's better to reduce them a little bit and expand all of the selectors when one of them has a longer name than the other?

I think that a balanced default width would end in the ``Marie Colvin`` phrase in the description of your screenshot. Few track names will be bigger than this. And if they are, all of the selectors will expand accordingly:

76702057-02387180-66c7-11ea-980c-a9d6a489967f

PR #953, opened 2020-03-16
Remove tar from published artifacts

dkanada:


.ci/azure-pipelines.yml
```diff
@@ -16,44 +16,61 @@ jobs:
...
 
-    - script: 'yarn install'
+    - task: Cache@2
+      displayName: 'Check Cache'
+      inputs:
+        key: 'yarn | yarn.lock'
+        path: 'node_modules'
+        cacheHitVar: CACHE_RESTORED
+
+    - script: 'yarn install --pure-lockfile'
```
MrTimscampi:
```suggestion
    - script: 'yarn install --frozen-lockfile'
```

Otherwise, the lockfile is essentially useless...

.ci/azure-pipelines.yml
```diff
@@ -67,8 +84,16 @@ jobs:
...
```
MrTimscampi: Same here

MrTimscampi:
```suggestion
    - script: 'yarn install --frozen-lockfile'
```

PR #954, opened 2020-03-16
Add standalone Debian package configuration

joshuaboniface:
**Changes**
Adds a standalone `.deb` package configuration to the jellyfin-web repository. This allows us to build the `.deb` for web on its own, independent from the main server.

The resulting package has a `Recommends` on a package called `jellyfin-server`, which will eventually be the renamed version of the original `jellyfin` package. `jellyfin` will then become a metapackage containing both the web and server components, allowing an administrator to pick-and-choose what to install.

This package changes the location of the installed files from the original `jellyfin` package, putting them in `/usr/share/jellyfin/web` rather than `/usr/lib/jellyfin/bin/jellyfin-web`. This reflects both the separation (the main package will continue to install binaries to `/usr/lib/jellyfin/bin`) as well as the architecture-independent nature of the web files, while avoiding conflicts with older package versions should that arise.

**Issues**
N/A (I thought there was one, but guess not)


dkanada:
@joshuaboniface the standalone version doesn't work with plugins yet, which would be a blocker.

.gitignore
```diff
@@ -576,3 +576,5 @@ healthchecksdb
 
 # dist for webpack output
 dist
+web
```
dkanada: What is the web folder?

debian/control
```diff
@@ -0,0 +1,16 @@
...
+Standards-Version: 3.9.4
+Homepage: https://jellyfin.media/
+Vcs-Git: https://github.org/jellyfin/jellyfin.git
+Vcs-Browser: https://github.org/jellyfin/jellyfin
+
+Package: jellyfin-web
+Recommends: jellyfin-server
+Architecture: any
+Description: Jellyfin is the Free Software Media System.
+ This package provides the Jellyfin server WebUI.
```
dkanada:
```suggestion
 This package provides the web client for Jellyfin.
```

debian/source.lintian-overrides
```diff
@@ -0,0 +1,3 @@
+# This is an override for the following lintian errors:
+jellyfin source: license-problem-md5sum-non-free-file Emby.Drawing/ImageMagick/fonts/webdings.ttf*
+jellyfin source: source-is-missing
```
dkanada: Are these actual problems here?

debian/source/options
```diff
@@ -0,0 +1,11 @@
...
+tar-ignore='**/.git'
+tar-ignore='**/.hg'
+tar-ignore='**/.vs'
+tar-ignore='**/.vscode'
+tar-ignore='deployment'
+tar-ignore='**/bin'
+tar-ignore='**/obj'
+tar-ignore='**/.nuget'
+tar-ignore='*.deb'
+tar-ignore='ThirdParty'
```
dkanada: Most of these don't apply to this repository.

.gitignore
joshuaboniface: `dist` gets renamed to `web` as part of the build process; this just prevents it showing up if the build fails mid-way through.

debian/control
joshuaboniface: Sounds good.

debian/source.lintian-overrides
joshuaboniface: Probably not, I just didn't touch this.

debian/source/options
joshuaboniface: ibid.

joshuaboniface:
Closing this one off - I'm going to include it as a part of the larger standalone build PR.

PR #955, opened 2020-03-16
Fix schedules direct buttons being hidden by default

thornbill:
**Changes**
Fixes an issue reported on [Reddit](https://old.reddit.com/r/jellyfin/comments/fiprd5/guide_data_for_live_tv/) where the submit and cancel buttons were hidden by default when adding Schedules Direct guide provider.

can't see me

**Issues**
N/A

### DIFF
--- a/src/components/tvproviders/schedulesdirect.js
+++ b/src/components/tvproviders/schedulesdirect.js
@@ -264,18 +264,20 @@ define(["jQuery", "loading", "emby-checkbox", "listViewStyle", "emby-input", "em
         self.init = function () {
             options = options || {};
 
-            // Show cancel button by default
-            if (options.showCancelButton !== false) {
-                page.querySelector(".btnCancel").classList.remove("hide");
-            } else {
+            // Only hide the button if explicitly set to false; default to showing if undefined or null
+            // FIXME: rename this option to clarify logic
+            if (options.showCancelButton === false) {
                 page.querySelector(".btnCancel").classList.add("hide");
+            } else {
+                page.querySelector(".btnCancel").classList.remove("hide");
             }
 
-            // Show submit button by default
-            if (options.showSubmitButton !== false) {
-                page.querySelector(".btnSubmitListings").classList.remove("hide");
-            } else {
+            // Only hide the button if explicitly set to false; default to showing if undefined or null
+            // FIXME: rename this option to clarify logic
+            if (options.showSubmitButton === false) {
                 page.querySelector(".btnSubmitListings").classList.add("hide");
+            } else {
+                page.querySelector(".btnSubmitListings").classList.remove("hide");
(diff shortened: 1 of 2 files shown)


dmitrylyzo:
`Cancel` button is only visible when second form is available.

I moved `Cancel` to the end of template and moved break line element into second form div (at the beginning).
screen1
screen2

src/components/tvproviders/schedulesdirect.js
```diff
@@ -264,13 +264,15 @@ define(["jQuery", "loading", "emby-checkbox", "listViewStyle", "emby-input", "em
         self.init = function () {
             options = options || {};
 
-            if (options.showCancelButton) {
+            // Show cancel button by default
+            if (options.showCancelButton !== false) {
```
JustAMan: Wtf, can you please explain?..

mark-monteiro: Not 100% sure what's happening here but it feels like it would be better to rename this option to `hideCancelButton`. So that when the option unset/null the cancel button is still shown.

thornbill: This changes so that it is explicitly checking for a `false` value being set instead of hiding by default if the option is "falsey" (`undefined`, `null`, `0`, ...). This was the original behavior, but was broken in a previous commit as a suggested change in a PR. Changing options seems like it would be beyond the scope of a bugfix that should be backported.

https://github.com/jellyfin/jellyfin-web/commit/36fae11a38aeb1840a51496aa2334c8cce7c8c36#diff-e515aef7db17535709492337ab84cb24L269-L276

JustAMan: At least add a //FIXME, I think Mark is right here in terms of renaming the option.

src/components/tvproviders/schedulesdirect.js
```diff
@@ -264,18 +264,20 @@ define(["jQuery", "loading", "emby-checkbox", "listViewStyle", "emby-input", "em
...
-            // Show cancel button by default
-            if (options.showCancelButton !== false) {
-                page.querySelector(".btnCancel").classList.remove("hide");
-            } else {
+            // Only hide the button if explicitly set to false; default to showing if undefined or null
+            // FIXME: rename this option to clarify logic
+            if (options.showCancelButton === false) {
                 page.querySelector(".btnCancel").classList.add("hide");
+            } else {
+                page.querySelector(".btnCancel").classList.remove("hide");
```
mark-monteiro: What about this:

```
const hideCancelButton = showCancelButton === false;
page.querySelector(".btnCancel").classList.toggle("hide", hideCancelButton);
`````

This seems pretty clear to me and will also make the transition to rename the option simpler later on

PR #958, opened 2020-03-17
Show missing indicator in ListView

MrTimscampi:
**Changes**

For some reason, the "missing" indicator was hidden in ListViews, making it more difficult than necessary to see what's missing in a TV show season.

The entire block of CSS was removed since it did pretty much nothing visible.

Screenshot_2020-03-17 Jellyfin

**Issues**

### DIFF
--- a/src/components/listview/listview.css
+++ b/src/components/listview/listview.css
@@ -238,13 +238,6 @@
     background-color: transparent !important;
 }
 
-.listItemMediaInfo {
-    /* Don't display if flex not supported */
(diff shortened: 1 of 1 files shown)


MrTimscampi:
> If it's missing, why the play button is still on the item's thumbnail? We shouldn't adress that as well in this PR?

I do plan on getting to it.

This was done quickly to show it, because it was annoying :p I still need to see how that indicator get the "missing" status for the item and add the check to the play button.

PR #959, opened 2020-03-18
Make header icons bigger

ferferga:
I think that the icons should be bigger in the header for the sake of accessibility, as currently there is a lot of wasted space imo.

**Before**
*Desktop*
before
*Mobile*
mobile_before

**After**
*Desktop*
after
*Mobile*
mobile_after

### DIFF
--- a/src/assets/css/librarybrowser.css
+++ b/src/assets/css/librarybrowser.css
@@ -95,6 +95,10 @@
     justify-content: flex-start;
 }
 
+.headerLeft > button {
+    transform: scale(1.2);
+}
+
 .headerRight {
     display: -webkit-box;
     display: -webkit-flex;
@@ -109,6 +113,10 @@
 .noHeaderRight .headerRight,
 .noHomeButtonHeader .headerHomeButton {
     display: none !important;
+}
+
+.headerRight > * {
+    transform: scale(1.2);
 }
 
 .pageTitle {


hamburglar2160:
These sizes are in line with other apps. Perhaps upping the systemwide dpi is a better solution?

ferferga:
@hamburglar2610 So the icons are bigger in desktop and not in Mobile? I'm not sure of what you mean and I don't know if there is any difference between scaling the icons and upping the DPI as you suggest.

dkanada:
I'm not sold on this change, @thornbill and @YouKnowBlom what do you guys think? The footer element for music playback has a similar rule, if we do keep this they should maybe have the same class.

YouKnowBlom:
I kinda like it for mobile but for desktop I'm not entirely sure

YouKnowBlom:
In my opinion the icons are too large on desktop but perfect for mobile

MrTimscampi:
Reviving this.

Testing on TV, Desktop and Mobile, it generally feels too big.

I also think 24px is pretty standard for all icons everywhere, while 28.8px is very much not. It's also no pixel-perfect, which causes blurriness:

image

I also don't see how adding 4 pixels for height makes them more accessible. If anything, what we should do accessibility-wise is:

* Replace `<i>` with `<span>` to avoid confusing screen readers
* Add `aria-label` or `aria-hidden` depending if the icon is indicating an action or is decorative

dkanada:
We definitely can't merge this with the clipping issue. I never thought this was an issue before, but I'd like to make a decision on this pull request at some point either way because it's been sitting for a while.

ferferga:
@dkanada I would simply close it or just change the size in desktop layout, because as @hamburglar2160 pointed out, the sizes on mobile seem to be in-par with all other (Android at least) applications. We can experiment a lot on Desktop because there isn't an uniform interface everywhere, but I don't think we should in mobile after noticing it.

We def need to make something about the wasted space, I still think header is really big for what it's being used.

YouKnowBlom:
So I've done some testing on my phone. I'm not sure which variables may change this between phones other than resolution, but mine has a 1080p screen.

Comparing the hamburger menu to some of my commonly used apps:
Jellyfin 10.5.4: 49x33px
Google Play: 49x33px
Telegram: 48x34px
Discord: 50x34px

Comparing the Google Cast icon:
Jellyfin 10.5.4: 59x49px
Youtube: 58x48px
Google Play Movies: 59x49px

I see no real inconsistency in size with my current setup.

MrTimscampi:
> We def need to make something about the wasted space

But why? It actually has very little wasted vertical space, and it's not really wasted space, but is allowing the content inside of it to breathe.

I join @YouKnowBlom in pointing out that our icons use a very standard icon size and it doesn't make sense to override it simply for the sake of making it bigger.

I still think the accessibility concerns are non-existent. As someone who uses accessibility features every day (because my eyesight is so bad that I can't drive), I don't want to have website make their stuff bigger. I'd rather have them respect my scaling choices, whether in the browser or system-wide, which we do. For even more impaired users, the move to `<span>` instead of `<i>` for icons already helps screen-readers give context to the elements. For helping them describe the page, we should add `aria-*` properties to elements that need them.

I honestly don't see any rationale for merging this.

PR #960, opened 2020-03-18
Translate missing items - Part 1

ferferga:
I still need to get rid of the ``humanedate`` component and translate all the dates, but here there are already some things that were forgotten.


src/scripts/librarybrowser.js
```diff
@@ -83,7 +83,7 @@ define(["userSettings"], function (userSettings) {
 
             if (html += '<div class="listPaging">', showControls) {
                 html += '<span style="vertical-align:middle;">';
-                html += (totalRecordCount ? startIndex + 1 : 0) + "-" + recordsEnd + " of " + totalRecordCount;
+                html += Globalize.translate("ListPaging").replace("{0}", (totalRecordCount ? startIndex + 1 : 0) + "-" + recordsEnd).replace("{1}", totalRecordCount);
```
dkanada: The translate library has a method to replace the variables.

ferferga: That's how is done in other parts of the code, so I used that way to keep that consistent. Will replace that then with the built-in method in every part where it's used across the entire web code. Any hint at the function and an example so I can accomplish this?

Thanks for pointing that out!

Btw, adding the strings at `en-US` should be enough for Weblate to populate it across all languages, right?

dkanada: The media type provider selectors implement this, I think you just have to overload the translate method with the extra strings and it will fill them in properly.

ferferga:
You were right @dkanada , just overloading the translate method did the trick. I replaced all the parts in the code that used ``.replace`` in conjunction with translate.

ferferga:
Before I dive into handling translations for dates better, I saw another thing in the code about translations that was incoherent and I would like the input of everybody in @jellyfin/web .

Some scripts define the ``globalize`` function, while others call it directly with ``Globalize``, without defining it in the ``define`` statement. Which approach do you think it's better, so we leave everything in one way or in the other (but not both)?

dkanada:
@thornbill as our resident standards guru, how commonly should global variables be used for important scripts? Is there any reason at all to use globals now that we're migrating to ES6 or should we try to remove them all? I suppose the one good use would be important scripts that also need to keep their state, like the API client.

MrTimscampi:
> I suppose the one good use would be important scripts that also need to keep their state, like the API client.

Those should like be classes with a singleton pattern, then.

I'd advocate for removing all globals, because globals are confusing and can generally be done another way.

thornbill:
👍 I agree. Globals should be avoided.

ferferga:
Will replace all occurences that use ``Globalize`` globally then in my next push. Thanks for your input @MrTimscampi @thornbill

dkanada:
@ferferga any chance you can also move the strings from #872 into the translation files so we can localize them?

ferferga:
@dkanada Of course. I assigned it to myself so I don't forget it

ferferga:
@dkanada I'm thinking that maybe it's better if that PR gets merged first, as is right now, and I merge it into my branch, so I don't end up with merge conflicts. Can this be possible?

ferferga:
I need the expertise of @jellyfin/web as a whole for implementing ``date-fns`` (a module proposed by @MrTimscampi for translating dates to every locale we support), as I'm stuck with it.

I'm not sure how I can implement date-fns in a way that's clean and that also allows the user to change between locales in the same web session without showing dates in the locale that was used at start time.

Take a look at [1656e73 commit](https://github.com/jellyfin/jellyfin-web/pull/960/commits/1656e73adf3dfc3906d9cab7abf5a22266e90935) and ignore the rest of them. According to [date-fns](https://date-fns.org/v2.11.0/docs/I18n)'s documentation, you can import locales using an ``import`` statement in each of the places where do you want to use locales. According to date-fns/date-fns#816 , it's impossible to set a default locale for ``date-fns``. That means that I would need to import all of the languages to each of the scripts that will use it (like I do with ``scheduledtasks.js``) and then, make a function that retrieves from globalize the current locale and pass the date-fns locale object to the dict in the ``date-fns``'s functions. But there should be another way to do it by just importing the locale we need, right? Any idea?

You can see an example in ``scheduledtasks.js`` for Spanish only in the above commit.

Thank you very much in advance for your input.

dkanada:
This pull request is reaching the edges of what I would consider an easy review so if possible I'd move the `date-fns` library to another one. We definitely can't break CI with the linting changes though, because CI always needs to pass on master or it's basically useless.

dkanada:
I'm looking at the lint errors now, how did all of that start showing up randomly? Was the code just convoluted enough that even eslint couldn't make sense of the codebase?

src/controllers/dlnaprofile.js
```diff
@@ -476,11 +476,11 @@ define(["jQuery", "loading", "fnchecked", "emby-select", "emby-button", "emby-in
 
             html += "<div>";
             html += '<a is="emby-linkbutton" href="#" class="lnkEditSubProfile" data-profileindex="' + i + '">';
-            html += "<p>" + Globalize.translate("ValueCodec").replace("{0}", profile.Codec || allText) + "</p>";
```
dkanada: I would keep these changes and revert the global variable removal for now because you can't be expected to fix all those linting errors at the same time. Hopefully you can just revert to just before the variable change.

MrTimscampi:
> I'm looking at the lint errors now, how did all of that start showing up randomly? Was the code just convoluted enough that even eslint couldn't make sense of the codebase?

There's really only these:

```
  166:63   error    'globalize' is not defined           no-undef
  195:31   error    'globalize' is not defined           no-undef
  198:31   error    'globalize' is not defined           no-undef
  201:31   error    'globalize' is not defined           no-undef
  204:31   error    'globalize' is not defined           no-undef
  207:31   error    'globalize' is not defined           no-undef
```

The warnings have been there for a few weeks, they're due to changes in the linting config by @thornbill , iirc.

It's stuff we need to fix, but isn't breaking.

dkanada:
There's also no point to reporting warnings in the CI because no one will fix warnings by looking at the Azure logs, they'll just run eslint on their own machine and fix things there.

ferferga:
I agree with you @dkanada, it's giving me a lot of headaches. I think I will discard the global commit and leave this only with date-fns (which I have somewhat sorted out right now) and the translation to all the places and just leave the globals for another one.

dmitrylyzo:
@ferferga you just need to put `globalize` after some module that has a placeholder (argument).

src/components/accessschedule/accessschedule.js
```diff
@@ -1,4 +1,4 @@
-define(["dialogHelper", "datetime", "emby-select", "paper-icon-button-light", "formDialogStyle"], function (dialogHelper, datetime) {
+define(["dialogHelper", "datetime", "emby-select", "paper-icon-button-light", "formDialogStyle", "globalize"], function (dialogHelper, datetime, globalize) {
```
dmitrylyzo: for example
```suggestion
define(["dialogHelper", "datetime", "globalize", "emby-select", "paper-icon-button-light", "formDialogStyle"], function (dialogHelper, datetime, globalize) {
```

dkanada: @ferferga I was misreading the warnings as errors, there are only a few issues that @MrTimscampi has linked above that can be fixed with this change :+1: no need to revert the globalize rename.

ferferga:
@dmitrylyzo Wasn't sure if I could place it everywhere in the define, I googled it and didn't get any good conclusion, so thanks for pointing me in the right direction.

However, I will apply them to another branch and make a new PR, because although I wanted to do a pretty big revision of Globalize in this PR, this isn't really in the scope of translating things. Will request you for review when it's done if you don't mind, okay?

dkanada:
> maybe it's better if that PR gets merged first

If you translate those items right now we can merge this very soon, so you won't have to fix conflicts.

src/scripts/dfnshelper.js
```diff
@@ -0,0 +1,103 @@
+import * as locale from 'date-fns/locale'
+define(["globalize"], function (globalize) {
```
MrTimscampi, line 2:
```suggestion
import { ar, be, bg, ca, cs, da, de, el, enGB, enUS, es, faIR, fi, fr, frCA, he, hi, hr, hu, id, it, kk, ko, lt, ms, nb, nl, pl, ptBR, pt, ro, ru, sk, sl, sv, tr, uk, vi, zhCN, zhTW } from 'date-fns/locale';
import globalize from 'globalize';

```

Mixing ES6 imports and AMD imports is a bad idea. You can import an AMD module in ES6 in jellyfin-web, for now. We have a Babel plugin that will convert your imports to AMD.

Importing an entire package with `*` is also bad practice. We don't want to import thing we don't want and it's clearer as to what is used or not.

src/scripts/site.js
```diff
@@ -647,12 +648,12 @@ var AppInfo = {};
             inputManager: "scripts/inputManager",
             datetime: "scripts/datetime",
             globalize: "scripts/globalize",
+            dfnshelper: "scripts/dfnshelper",
```
MrTimscampi, line 651: It'd rename the module to something more explicit and generic, like `DateLocaleHelper`.

src/scripts/dfnshelper.js
MrTimscampi: > We have a Babel plugin that will convert your imports to AMD.

Do note that your IDE might complain that it can't find the module. This is normal for now, as we need to keep the AMD module paths the same. It'll be fixed in a final PR at the end of the migration.

ferferga:
I think that this PR has reached a fair enough work for getting merge, as fixing the current mess completely in one PR will make this unreviewable.

**Things Done**

* Everything I'm aware of is translated right now, no more hardcoded strings
* Implementation of date-fns library (as suggested by @MrTimscampi in Matrix) and a ``datefnshelper`` script that returns the correct locale object from the date-fns library based on the current locale of the user. Not used in a lot of places right now, but the plan is replace ``datetime`` with it completely. But that's for another PR.
* The locale of the dates also changes if you change the language of the interface while it's running - no need to reload

**Known issues**

* It seems that dates close to now are translated incorrectly. For example, if an user is active right now, the text "Last seen in a minute" might be displayed instead of "Last seen a minute ago". I'm not sure if this is a date-fns error or it's some kind of time desynchronization between the server and the client
* I also attempted to translate ``scheduledtask.js`` file. However, the current implementation is a complete mess and would need a refactor. Again, a really big thing for this PR.
This means that some sentences will appear in other languages like "Sundays a las 21:00" (in Spanish)

All of this is commented in the code, so it can be found easily later.

**Screenshots**

*This is what I mean with the ``scheduledtask.js`` thing*
image

image

Tasks and descriptions are already translated on jellyfin/jellyfin#2678

image

Sin título

*Activity log is also improved. It displays dates like "today at X", "yesterday at X", etc etc*
*My IP is hidden, but that's untouched*
Sin título

*Some CSS fixes for showing the whole text should be made here*
Sin título

I'm sorry as well for the huge number of commits. After previously breaking my branch because of the changes that were moved to #1004 , I started being really careful of everything I made, so it was easy to come back. I hope this is also good for reviewing and cherry-picking if necessary.

ferferga:
@dkanada I think that #872 can be merged once this gets approved, so I can fetch and merge master and make a final commit with that sentence also translated.

src/scripts/site.js
ferferga: @MrTimscampi I agree that dfnshelper it's not the best name, but it's short and you can correlate it to date-fns. ``DateLocaleHelper`` doesn't mention at all date-fns, so it might be confusing in the long-term. It  only handles the locale objects of date-fns, nothing to do with globalize or anything else related to localization, so I think it's a good idea to leave some kind of mention to date-fns

src/scripts/dfnshelper.js
ferferga: @MrTimscampi Indeed! Thank you very much! 

Although I commited this today, this was written way back before the ES6 migration started, between the breakage I made with the first attempt to use globalize locally (now moved to #1004).

MrTimscampi:
Do remember to add the new module to the "overrides" in package.json as explained here: https://github.com/jellyfin/jellyfin-web/issues/868#issuecomment-593068027

src/scripts/dfnshelper.js
```diff
@@ -0,0 +1,104 @@
...
+            case 'zh-tw':
+                return zhTW;
+            default:
+                return enUS;
+        }
+    }
+    return {
+        getLocale: getLocale
+    };
+});
```
dmitrylyzo, line 104:
```suggestion
```

src/scripts/dfnshelper.js
```diff
@@ -0,0 +1,104 @@
+import { ar, be, bg, ca, cs, da, de, el, enGB, enUS, es, faIR, fi, fr, frCA, he, hi, hr, hu, id, it, kk, ko, lt, ms, nb, nl, pl, ptBR, pt, ro, ru, sk, sl, sv, tr, uk, vi, zhCN, zhTW } from 'date-fns/locale';
+import globalize from 'globalize';
+
+    "use strict";
+
+    function getLocale()
+    {
```
dmitrylyzo, line 7: @MrTimscampi should it be like this?
```suggestion
    export function getLocale() {
```

src/scripts/dfnshelper.js
```diff
@@ -0,0 +1,104 @@
...
+                return zhCN;
+            case 'zh-hk':
+                return zhCN;
+            case 'zh-tw':
+                return zhTW;
+            default:
+                return enUS;
+        }
+    }
+    return {
```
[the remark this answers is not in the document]
ferferga: Afaik, this would make the module incompatible with the other that are still using define. And it's not the purpose of this PR to move those to ES6.

In fact, I got ``async`` errors in gulp when I used export statements.

If you can confirm that there are no issues with the current state, I will commit it

src/scripts/dfnshelper.js
ferferga: Same thing I said to @MrTimscampi applies here

src/controllers/dashboard/plugins/available.js
```diff
@@ -16,7 +16,7 @@ define(["loading", "libraryMenu", "globalize", "cardStyle", "emby-button", "emby
     }
 
     function getHeaderText(category) {
-        category = category.replace(" ", "");
+        category = category(" ", "");
```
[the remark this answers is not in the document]
ferferga: Hmm, no idea why this is here 🤔. Probably I introduced it in a merge commit.

Let me check the history and discard non-relevant changes.

src/scripts/dfnshelper.js
ferferga: Same thing I said to @MrTimscampi applies here

src/scripts/dfnshelper.js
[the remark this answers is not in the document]
dmitrylyzo: You should add module here
https://github.com/jellyfin/jellyfin-web/blob/d41163d4b9ee8bb4993f370475b7a38f0d2986c6/package.json#L80-L89
@MrTimscampi right?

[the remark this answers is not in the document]
dmitrylyzo: I think, since this is a new file, you can remove indent for one level.

src/scripts/dfnshelper.js
```diff
@@ -0,0 +1,103 @@
+import * as locale from 'date-fns/locale'
+define(["globalize"], function (globalize) {
+    "use strict";
+
+    function getLocale()
+    {
+        switch (globalize.getCurrentLocale()) {
+            case 'ar':
```
JustAMan, line 9: I would rather see this is a dictionary, rather than this giant switch

src/scripts/dfnshelper.js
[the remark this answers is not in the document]
MrTimscampi: @dmitrylyzo Exactly :) That array of files tells the ES2015 to AMD plugin to do its thing on the file.

And since it's a new file, no need to keep a weird indent :)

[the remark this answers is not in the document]
ferferga: Okay, thanks both!

src/controllers/dashboard/plugins/available.js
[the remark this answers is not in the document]
ferferga: Completely my fault!

src/scripts/dfnshelper.js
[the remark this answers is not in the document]
ferferga: Okay, thanks both!

src/scripts/dfnshelper.js
ferferga: Hmm I've done it this way because I thought it would be more readable than the globalize implementation (which uses a dict). I'm open to whatever it's decided, I guess it's a matter of discussing this.

ferferga:
@MrTimscampi @dmitrylyzo With those commits, I get ``exports is not defined`` everytime datefns retrieves from dfnshelper the ``locale`` object

src/scripts/dfnshelper.js
JustAMan: dict storing looks more compact without losing the readability for my taste. Though by no means count this as a blocker, if others are fine with the PR they can go ahead and merge :)

package.json
```diff
@@ -86,7 +87,8 @@
           "src/components/input/keyboardnavigation.js",
           "src/components/sanatizefilename.js",
           "src/scripts/settings/webSettings.js",
-          "src/components/scrollManager.js"
+          "src/components/scrollManager.js",
+          "scripts/dfnshelper.js"
```
dmitrylyzo, line 91:
```suggestion
          "src/scripts/dfnshelper.js"
```

ferferga: Glad there is code review 😫

ferferga: Works 👌 now thanks!

ferferga:
I'll tag this as blocked as this PR needs to translate the strings added in #872 

cc @dkanada

src/components/notifications/notifications.js
```diff
@@ -173,15 +173,15 @@ define(['serverNotifications', 'playbackManager', 'events', 'globalize', 'requir
...
             } else if (status === 'cancelled') {
-                notification.title = globalize.translate('PackageInstallCancelled').replace('{0}', installation.Name + ' ' + installation.Version);
+                notification.title = globalize.translate('PackageInstallCancelled', installation.Name + ' ' + installation.Version);
             } else if (status === 'failed') {
-                notification.title = globalize.translate('PackageInstallFailed').replace('{0}', installation.Name + ' ' + installation.Version);
+                notification.title = globalize.translate('PackageInstallFailed', installation.Name + ' ' + installation.Version);
                 notification.vibrate = true;
             } else if (status === 'progress') {
-                notification.title = globalize.translate('InstallingPackage').replace('{0}', installation.Name + ' ' + installation.Version);
+                notification.title = globalize.translate('InstallingPackage', installation.Name + ' ' + installation.Version);
```
MrTimscampi, line 184: We probably want to let the translation library handle the string replacement for these as well.

src/controllers/dashboard/scheduledtasks/scheduledtasks.js
```diff
@@ -66,7 +66,8 @@ define(["jQuery", "loading", "events", "globalize", "serverNotifications", "huma
         var html = "";
         if (task.State === "Idle") {
             if (task.LastExecutionResult) {
-                html += globalize.translate("LabelScheduledTaskLastRan").replace("{0}", humaneDate(task.LastExecutionResult.EndTimeUtc)).replace("{1}", humaneElapsed(task.LastExecutionResult.StartTimeUtc, task.LastExecutionResult.EndTimeUtc));
+                html += globalize.translate("LabelScheduledTaskLastRan", datefns.formatDistanceToNow(Date.parse(task.LastExecutionResult.EndTimeUtc), { addSuffix: true, locale: dfnshelper.getLocale() }),
+                    datefns.formatDistance(Date.parse(task.LastExecutionResult.StartTimeUtc), Date.parse(task.LastExecutionResult.EndTimeUtc), { addSuffix: false, locale: dfnshelper.getLocale() }));
```
MrTimscampi, line 70:
```suggestion
                const lastExecutionResultToNow = datefns.formatDistanceToNow(Date.parse(task.LastExecutionResult.EndTimeUtc), { addSuffix: true, locale: dfnshelper.getLocale() })
                const lastExecutionResultInterval = datefns.formatDistance(Date.parse(task.LastExecutionResult.StartTimeUtc), Date.parse(task.LastExecutionResult.EndTimeUtc), { addSuffix: false, locale: dfnshelper.getLocale() });
                html += globalize.translate("LabelScheduledTaskLastRan", lastExecutionResultToNow, lastExecutionResultInterval);
```

This is a bit long and complex, so feel free to separate it into variables.

You could also probably move the following to the helper module, as it seems to be quite common:
```
{ addSuffix: true, locale: dfnshelper.getLocale() }
```
Export it as something like:
```javascript
export const localeWithSuffix = { addSuffix: true, locale: dfnshelper.getLocale() }
```

src/components/cardbuilder/cardBuilder.js
```diff
@@ -1082,11 +1082,7 @@ import 'programStyles';
 
                 if (options.showPersonRoleOrType) {
                     if (item.Role) {
-                        lines.push('as ' + item.Role);
-                    } else if (item.Type) {
```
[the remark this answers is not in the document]
ferferga: @MrTimscampi If you take a look at [this line](https://github.com/jellyfin/jellyfin-web/pull/960/files/60329961b52737ad182a76940065f6cf264c5bcc#diff-265725aca33ce25b93965819e407eb05L1087), you will see that the Types weren't actually used at all in the interface.

``Type`` is what the person does (e.g Composer, Actor, Producer, etc), just as you said.
``Role`` is what the person does in the movie. See this for example:

image

In this case, ``Type`` is *Actor* and ``Role`` is *Buster Moon (voice)*

This is completely redundant, because most of the times Role = Type. See this one:

image

``Producer`` is the role and the type. So we would have the same thing twice in the UI. That's why I think this line was removed, because in your example, Beethoven's Type (Composer) = Role (Composer)

If you think that the Type should be back, let's label this as discussion, but I don't think we should: It's easily deductable already.

This is what you mean? If not, can you please attach an screenshot showcasing it? I'm not sure I understand 100% the context.

src/components/cardbuilder/cardBuilder.js
```diff
@@ -1082,11 +1082,7 @@ import 'programStyles';
...
-                        lines.push(globalize.translate('' + item.Type));
-                    } else {
```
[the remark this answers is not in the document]
ferferga: @MrTimscampi Same as above. I did not found anything different.

src/components/notifications/notifications.js
ferferga: Do you have an screenshot of those notifications? Not 100% sure of the context

src/components/cardbuilder/cardBuilder.js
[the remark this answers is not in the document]
MrTimscampi: Then we should probably change how the non-actor roles are handled, but no need to do so right now :)

src/components/notifications/notifications.js
MrTimscampi: I don't, but there's no need to concatenate the strings manually here, in my opinion. But it's a non-blocker.

src/components/itemidentifier/itemidentifier.js
```diff
@@ -309,7 +309,7 @@ define(["dialogHelper", "loading", "connectionManager", "require", "globalize",
                     fullName = idInfo.Name + " " + globalize.translate(idInfo.Type);
                 }
 
-                var idLabel = globalize.translate("LabelDynamicExternalId").replace("{0}", fullName);
+                var idLabel = globalize.translate("LabelDynamicExternalId", idInfo.Name);
```
dmitrylyzo, line 312:
```suggestion
                var idLabel = globalize.translate("LabelDynamicExternalId", fullName);
```
#971

src/components/metadataeditor/metadataeditor.js
```diff
@@ -470,7 +470,7 @@ define(['itemHelper', 'dom', 'layoutManager', 'dialogHelper', 'datetime', 'loadi
                 fullName = idInfo.Name + " " + globalize.translate(idInfo.Type);
             }
 
-            var labelText = globalize.translate("LabelDynamicExternalId").replace("{0}", fullName);
+            var labelText = globalize.translate('LabelDynamicExternalId', idInfo.Name);
```
dmitrylyzo, line 473:
```suggestion
            var labelText = globalize.translate('LabelDynamicExternalId', fullName);
```
#971

PR #962, opened 2020-03-19
Add method to open client settings

dkanada:
This will let people control things like notification persistence and battery optimizations on android and other clients. @thornbill do you think this could be useful for iOS as well?

### DIFF
--- a/src/components/shell.js
+++ b/src/components/shell.js
@@ -10,12 +10,6 @@ define([], function () {
             }
 
         },
-        canExec: false,
-        exec: function (options) {
-            // options.path
-            // options.arguments
-            return Promise.reject();
-        },
         enableFullscreen: function () {
             if (window.NativeShell) {
                 window.NativeShell.enableFullscreen();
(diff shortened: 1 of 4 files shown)


src/components/shell.js
```diff
@@ -8,23 +8,6 @@ define([], function () {
...
-        exec: function (options) {
-            // options.path
-            // options.arguments
-            return Promise.reject();
-        },
-        enableFullscreen: function () {
-            if (window.NativeShell) {
-                window.NativeShell.enableFullscreen();
-            }
-        },
```
dkanada: These were only added for Android to work around a bug in Chrome. The fix is handled within Cordova now.

src/strings/en-us.json
```diff
@@ -147,6 +147,7 @@
     "ChannelNumber": "Channel number",
     "Channels": "Channels",
     "CinemaModeConfigurationHelp": "Cinema mode brings the theater experience straight to your living room with the ability to play trailers and custom intros before the main feature.",
+    "ClientSettings": "Client Settings",
```
thornbill: Maybe “App Settings” would be more clear?

dkanada: I tried to keep it generic since some platforms don't use the term app, such as desktops.

PR #963, opened 2020-03-19
Don't scale logo on details page

MrTimscampi:
**Changes**

Background size is set as "contain" on the logo, which makes some logos overly big. Enlarging these isn't a good idea anyway, so this simply lets them use their normal size.

Before
Screenshot_2020-03-19 Jellyfin

After
Screenshot_2020-03-19 Jellyfin(1)

**Issues**


dmitrylyzo:
Does not work for tall logos.
logo-1

MrTimscampi:
> Does not work for tall logos.
> logo-1

Good catch, will adjust the size of the container.

PR #968, opened 2020-03-22
Add controls for DeinterlaceMethod

alset333:
Add controls and US-English text for DeinterlaceMethod in encodingsettings. Uses the values referenced here: https://github.com/jellyfin/jellyfin/pull/2639

**Changes**
- Add a section to EncodingSettings to control DeinterlaceMethod.
- Add en-us text for aforementioned new section.


dkanada:
@alset333 is this ready for review?

alset333:
Yes, this should be ready for review now. The code it depends on was committed at https://github.com/jellyfin/jellyfin/commit/5acd0521a26cbc89c66cc20cad3213b4d2a111ee

alset333:
Yes, @dkanada, this should be ready for review now. The code it depends on was committed at https://github.com/jellyfin/jellyfin/commit/5acd0521a26cbc89c66cc20cad3213b4d2a111ee

dmitrylyzo:
I don't see where `selectDeinterlaceMethod` option is sent to server. :confused:

src/strings/en-us.json
```diff
@@ -174,6 +174,7 @@
     "DefaultErrorMessage": "There was an error processing the request. Please try again later.",
     "DefaultMetadataLangaugeDescription": "These are your defaults and can be customized on a per-library basis.",
     "DefaultSubtitlesHelp": "Subtitles are loaded based on the default and forced flags in the embedded metadata. Language preferences are considered when multiple options are available.",
+    "DeinterlaceMethodHelp": "Select the deinterlacing method to use when transcoding.",
```
MrTimscampi, line 177:
```suggestion
    "DeinterlaceMethodHelp": "Select the deinterlacing method to use when transcoding interlaced content.",
```

MrTimscampi:
> I don't see where `selectDeinterlaceMethod` option is sent to server. confused

Indeed, there's no getter or setter for the setting. As-is, it will do nothing but add the non-functionning input @alset333 .

src/encodingsettings.html
```diff
@@ -136,6 +136,14 @@ <h3 class="checkboxListLabel">${LabelEnableHardwareDecodingFor}</h3>
                     <div class="fieldDescription">${H264CrfHelp}</div>
                 </div>
 
+                <div class="selectContainer">
+                    <select is="emby-select" id="selectDeinterlaceMethod" label="${LabelDeinterlaceMethod}">
+                        <option value="yadif">Yadif</option>
+                        <option value="yadif_bob">Yadif Bob</option>
```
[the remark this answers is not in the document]
alset333: Good point.
Frankly I'm not sure what/how these should be translated. "Yadif" is a program name (originally comes as an acronym from Yet Another Deinterlacing Filter) and "bob" typically indicates interpolation to get 60p (instead of throwing out frames to get 30p)

alset333:
> 
> 
> > I don't see where `selectDeinterlaceMethod` option is sent to server. confused
> 
> Indeed, there's no getter or setter for the setting. As-is, it will do nothing but add the nonfunctioning input @alset333 .

Hm... I'll take another look but it seemed to be working on my end; possible I missed something with the repo.

src/encodingsettings.html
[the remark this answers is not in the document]
MrTimscampi: It's mostly if translators prefer to write it in their local script. Like, say, a Japanese translator decides to write it as `ヤディフ` for some reason.

It gives people options if they want them, otherwise it'll use the defautl en_US version.

alset333:
I think this is working correctly now. The getters and setters for "selectDeinterlaceMethod" have been added to encodingsettings.js based on the structure of "selectEncoderPreset"'s getters/setters. Now testing good on my end again; surprised I didn't catch this issue myself so thanks.

src/controllers/encodingsettings.js
```diff
@@ -14,6 +14,7 @@ define(["jQuery", "loading", "globalize", "dom", "libraryMenu"], function ($, lo
         $("#txtVaapiDevice", page).val(config.VaapiDevice || "");
         page.querySelector("#selectEncoderPreset").value = config.EncoderPreset || "";
         page.querySelector("#txtH264Crf").value = config.H264Crf || "";
+        page.querySelector("#selectDeinterlaceMethod").value = config.DeinterlaceMethod || "";
```
dmitrylyzo, line 17: If `config.DeinterlaceMethod` is not set, `selectDeinterlaceMethod` is empty. Should there be a default method?

alset333: It should always be set, the default server-side value is "yadif" per this code: https://github.com/jellyfin/jellyfin/commit/5acd0521a26cbc89c66cc20cad3213b4d2a111ee#diff-d0f215ed6b3deb7472d977cd0fe0ff7e

We could add a default in encodingsettings, but there isn't one for most of the other settings (EncoderPreset, H264Crf, etc) which are already there.

PR #969, opened 2020-03-22
Fix XviD playback on Tizen

dmitrylyzo:
I received a message from Tizen user https://github.com/dmitrylyzo/jellyfin-tizen/issues/5

I can reproduce this with
```
General
Complete name                            : xvid_test.avi
Format                                   : AVI
Format/Info                              : Audio Video Interleave
File size                                : 44.1 MiB
Duration                                 : 5 min 0 s
Overall bit rate                         : 1 233 kb/s
Writing application                      : Lavf58.20.100

Video
ID                                       : 0
Format                                   : xvid
Codec ID                                 : xvid
Duration                                 : 5 min 0 s
Bit rate                                 : 1 092 kb/s
Width                                    : 1 280 pixels
Height                                   : 720 pixels
Display aspect ratio                     : 16:9
Frame rate                               : 29.970 (30000/1001) FPS
Scan type                                : Progressive
Bits/(Pixel*Frame)                       : 0.040
Stream size                              : 39.1 MiB (89%)

Audio
ID                                       : 1
Format                                   : MPEG Audio
Format version                           : Version 1
Format profile                           : Layer 3
Format settings                          : Joint stereo / MS Stereo
Codec ID                                 : 55
Codec ID/Hint                            : MP3
Duration                                 : 5 min 0 s
Bit rate mode                            : Constant
Bit rate                                 : 128 kb/s
Channel(s)                               : 2 channels
Sampling rate                            : 44.1 kHz
Compression mode                         : Lossy
Stream size                              : 4.58 MiB (10%)
Alignment                                : Aligned on interleaves
Interleave, duration                     : 26  ms (0.78 video frame)
Writing library                          : LAME3.100
```

Currently `avi` uses `DirectPlay` but new Samsung TV don't support XviD - this can be verified with video-file on USB-flash.
On play, I got just black screen with no errors anywhere, and webui just borked.

**Changes**
Explicitly add supported codecs for 'avi' for Tizen to make other codecs be transcoded.

**Issues**
https://github.com/dmitrylyzo/jellyfin-tizen/issues/5

### DIFF
--- a/src/scripts/browserdeviceprofile.js
+++ b/src/scripts/browserdeviceprofile.js
@@ -216,7 +216,7 @@ define(['browser'], function (browser) {
                 supported = browser.tizen || browser.orsay || browser.web0s || browser.edgeUwp;
                 // New Samsung TV don't support XviD/DivX
                 // Explicitly add supported codecs to make other codecs be transcoded
-                if (browser.tizen) {
+                if (browser.tizenVersion >= 4) {
                     videoCodecs.push('h264');
                     if (canPlayH265(videoTestElement, options)) {
                         videoCodecs.push('h265');
--- a/src/scripts/browserdeviceprofile.js
+++ b/src/scripts/browserdeviceprofile.js
@@ -215,6 +215,7 @@ define(['browser'], function (browser) {
             case 'avi':
                 supported = browser.tizen || browser.orsay || browser.web0s || browser.edgeUwp;
                 // New Samsung TV don't support XviD/DivX
+                // Explicitly add supported codecs to make other codecs be transcoded
                 if (browser.tizen) {
                     videoCodecs.push('h264');
                     if (canPlayH265(videoTestElement, options)) {
--- a/src/scripts/browserdeviceprofile.js
+++ b/src/scripts/browserdeviceprofile.js
@@ -214,6 +214,14 @@ define(['browser'], function (browser) {
                 break;
             case 'avi':
                 supported = browser.tizen || browser.orsay || browser.web0s || browser.edgeUwp;
+                // New Samsung TV don't support XviD/DivX
+                if (browser.tizen) {
+                    videoCodecs.push('h264');
+                    if (canPlayH265(videoTestElement, options)) {
+                        videoCodecs.push('h265');
+                        videoCodecs.push('hevc');
+                    }
+                }
                 break;
             case 'mpg':
             case 'mpeg':
--- a/src/scripts/browser.js
+++ b/src/scripts/browser.js
@@ -271,6 +271,9 @@ define([], function () {
 
     if (!browser.tizen) {
         browser.orsay = userAgent.toLowerCase().indexOf('smarthub') !== -1;
+    } else {
+        var v = (navigator.appVersion).match(/Tizen (\d+).(\d+)/);
+        browser.tizenVersion = parseInt(v[1]);
     }
 
     if (browser.edgeUwp) {


src/scripts/browserdeviceprofile.js
```diff
@@ -214,6 +214,14 @@ define(['browser'], function (browser) {
...
             case 'avi':
                 supported = browser.tizen || browser.orsay || browser.web0s || browser.edgeUwp;
+                // New Samsung TV don't support XviD/DivX
+                if (browser.tizen) {
+                    videoCodecs.push('h264');
+                    if (canPlayH265(videoTestElement, options)) {
+                        videoCodecs.push('h265');
+                        videoCodecs.push('hevc');
+                    }
+                }
```
dkanada: Why not just remove Tizen from the `supported` variable?

MrTimscampi: Because AVI can contain other things than XVID, like H.264 and H.265.

No need to transcode an entire container because one codec it can use isn't supported.

dkanada: Wouldn't we still need to report false or something so the regular behavior isn't enabled for this case at the end of the function?

MrTimscampi: As-is, I believe it will report:

```
Container: "avi"
Type: 'Video',
VideoCodec: "h264,h265,hevc"
AudioCodec: ""
```

Which should have the desired effect. Hopefully it's tested, though.

dmitrylyzo: it supported H.264 in `avi`. H.265 and HEVC possibly but not tested

dkanada: I would add a comment at least to explain the solution.

PrplHaz4:
Any idea if the dlna profile for these tvs should be updated too or does it have different playback support in app/browser?

MrTimscampi:
> Any idea if the dlna profile for these tvs should be updated too or does it have different playback support in app/browser?

I'd update the DLNA profile too, since often times they enable browser/app support for the same formats (ie all formats the TV can decode with hardware).

dmitrylyzo:
UE65NU7400 (Tizen 4) does not support XviD though DLNA.
But I also have LT27H390 (Tizen 3), and it supports XviD from USB-flash and DLNA.

dmitrylyzo:
I can add version checking.
Any suggestions how to handle it?
* Floating point number (as iOS version)? Worried about epsilon.
* String "3.0", "2.4"? Then what to use for comparison? https://www.npmjs.com/package/compare-versions or something else or write myself :stuck_out_tongue:

dkanada:
Integer should work fine as long as you keep it consistent.

dmitrylyzo:
> Integer should work fine as long as you keep it consistent.

OK. I will make only major version since `>= 4` for now is enough.

dmitrylyzo:
For Tizen 3, I had to check on pre-gulp version: `query-string` have syntax error (unexpected `[`). Not transpiled?

PrplHaz4:
> 
> 
> UE65NU7400 (Tizen 4) does not support XviD though DLNA.
> But I also have LT27H390 (Tizen 3), and it supports XviD from USB-flash, browser and DLNA.

Well it's going to be interesting maintaining profiles: https://developer.samsung.com/smarttv/develop/specifications/tv-model-groups.html

No mention of XviD on the individual tv spec page either 👎 

UniversalMediaServer only maintains a couple Samsung profiles I think...not sure what format support is significant enough to break them at...also not sure if support is more dependent on Tizen version or actual hardware :-/

dmitrylyzo:
> also not sure if support is more dependent on Tizen version or actual hardware :-/

Very likely. I think some system info can be obtained from app, but not from browser.
Maybe we could make configurable device profile. Don't know if there was a request for this somewhere.

JustAMan:
> Any suggestions how to handle it?
> 
> * Floating point number (as iOS version)? Worried about epsilon.
> * String "3.0", "2.4"? Then what to use for comparison? https://www.npmjs.com/package/compare-versions or something else or write myself 😛

Late to the party, but you can do the way h264 levels are stored - "5.1" is stored as number 51, etc. So multiply major by 10 and add minor and you have your version.

PR #976, opened 2020-03-23
Freeze lockfile by default

MrTimscampi:
**Changes**

Makes builds more reproducible by freezing the lockfile by default.

Should make dependency versions more consistent across builds.

**Issues**

### DIFF
--- a/.yarnrc
+++ b/.yarnrc
@@ -1,0 +1,1 @@
+--frozen-lockfile true


YouKnowBlom:
I'm getting an error when running `yarn install` on a fresh clone.

```
yarn install v1.21.1
[1/4] Resolving packages...
error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.
```

dmitrylyzo:
_Repost from matrix_
I tested native build under Ubuntu 18.04 with `--fronzen-lockfile`. Mentioned (backuping) lines are not required - lock file remains untouched. But it seems that with `--fronzen-lockfile` we cannot build with fixed versions: need update -> fail

MrTimscampi:
@YouKnowBlom @dmitrylyzo Rebasing on master fixes it. Seems like the lockfile became stale at some point.

We're likely to have a small transition period where PRs will need to be rebased to work as expected.

MrTimscampi:
Looking into it, I wonder if we should be doing this for dev.

Reproducible builds are good for CI (we want all the CI branches to be handled the same) but I'm not sure if it'll be useful to have to general use.

Closing this for now while I research more.

PR #984, opened 2020-03-25
Fix Browsersync JavaScript reload error

MrTimscampi:
**Changes**

Moves a few things around to properly handle live reloading for Javascript.

**Issues**

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -18,14 +18,14 @@ const stream = require('webpack-stream');
 const inject = require('gulp-inject');
 const postcss = require('gulp-postcss');
 const sass = require('gulp-sass');
- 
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -18,14 +18,14 @@ const stream = require('webpack-stream');
 const inject = require('gulp-inject');
 const postcss = require('gulp-postcss');
 const sass = require('gulp-sass');
- 
-sass.compiler = require('node-sass')
 
+sass.compiler = require('node-sass');
 
+var config;
```
thornbill: Why not `let`?

MrTimscampi: No particular reason :p

I made the change in a new commit.

PR #985, opened 2020-03-26
Add browser lazy-sync

dmitrylyzo:
Only added/changed files will be processed and synced.
Now `apploader.js` uses same javascript pipeline.

_Maybe I did it too hacky way_


dmitrylyzo:
Wrong dependency list. :relaxed:

PR #999, opened 2020-03-28
ES6 migration: autoFocuser, dom, scrollManager

dmitrylyzo:
**Changes**
* Migrate ScrollManger to ES6
* Migrate AutoFocuser to ES6
* Migrate DOM-module to ES6
* Add default export for compatibility `import` <-> `require`

**Issues**
#868


src/components/autoFocuser.js
```diff
@@ -1,22 +1,29 @@
...
 
     /**
      * Previously selected element.
      */
-    var activeElement;
+    let activeElement;
 
     /**
-     * Returns true if AutoFocuser is enabled.
+     * Returns _true_ if AutoFocuser is enabled.
```
dkanada: I don't think we want to wrap this in underscores.

src/components/autoFocuser.js
```diff
@@ -28,24 +35,19 @@ define(["focusManager", "layoutManager"], function (focusManager, layoutManager)
...
-    /**
-     * Create an array from some source.
-     */
-    var arrayFrom = Array.prototype.from || function (src) {
-        return Array.prototype.slice.call(src);
-    }
-
     /**
      * Set focus on a suitable element, taking into account the previously selected.
+     * @param {HTMLElement} [container] - element to limit scope
```
dkanada: I also don't think we were wrapping parameters in brackets.

src/components/scrollManager.js
```diff
@@ -433,39 +443,39 @@ define(["dom", "browser", "layoutManager"], function (dom, browser, layoutManage
     /**
      * Scrolls the document to a given element.
      *
-     * @param {HTMLElement} element target element of scroll task
-     * @param {boolean} [smooth=false] smooth scrolling
+     * @param {HTMLElement} element - target element of scroll task
+     * @param {boolean} [smooth=false] - smooth scrolling
```
dkanada: Will this value always be false?

src/components/dom.js
```diff
@@ -104,46 +158,60 @@ define([], function () {
...
-    var _animationEvent;
-    function whichAnimationEvent() {
+    /**
+     * Name of animation end event.
+     */
+    let _animationEvent;
 
+    /**
+     * Returns name of animation end event.
+     * @returns {string} Name of animation end event
```
dkanada: @MrTimscampi period or no period?

src/components/autoFocuser.js
dmitrylyzo: if you turn on `plugins/markdown`, this makes _italic_

src/components/autoFocuser.js
dmitrylyzo: this marks optional argument

src/components/scrollManager.js
dmitrylyzo: this marks optional + default is _false_

src/components/dom.js
dmitrylyzo: Currently it looks like this
doc1

MrTimscampi: I'd add a period (that's what I've been doing on every JSDoc comment).

I feel like we should be writing these both with in-source legibility (hence my use of "-" to separate parameter names and descriptions) and periods (to look nicer in the compiled docs).

Eventually, we probably want the JSDoc to be available in the developer doc to help people understand the web client before diving in.

dmitrylyzo: doc2

MrTimscampi: As discussed on Matrix with @dkanada, adopting the Google Style Guide for JSDoc is probably a good idea: https://google.github.io/styleguide/jsguide.html#jsdoc

We're nearly there anyway, so...

We'll make an addition to the developer documentation about it at some point.

dkanada: Yeah, we can ignore this change for now and then fix all issues with the linting.

PR #1000, opened 2020-03-28
Support H264 Level 52 (Tizen 5.0)

Nazar78:
**Changes**
Support H264 Level 52 (Tizen 5.0)
Direct play support on Tizen 5.0+ for 100 Mbps H264 videos i.e. from action cameras.

image


src/scripts/browserdeviceprofile.js
```diff
@@ -355,6 +355,31 @@ define(['browser'], function (browser) {
         // Not sure how to test for this
         var supportsMp2VideoAudio = browser.edgeUwp || browser.tizen || browser.orsay || browser.web0s;
 
+        // Get Tizen version & support
+        var getTizenVersion = self.tizen && self.tizen.systeminfo ? 
+            parseFloat(tizen.systeminfo.getCapability('http://tizen.org/feature/platform.version')) : null;
```
dmitrylyzo: _tbh, I see this code section first time_ :relaxed:

I know that it was before, but it will work in app only.
I was told that we should support browser users as possible. Personally, I see no reason to use browser when you have app, but why not if we can.

I recently added getting Tizen version.
https://github.com/jellyfin/jellyfin-web/blob/d0c2c1c8e0fe605f5ddc416cdbff81056faa2f5b/src/scripts/browser.js#L275-L276
We could use that. `browser.tizenVersion` is integer (enough for now).

Also, there are lint errors:
```
  359:68  error    Trailing spaces not allowed                             no-trailing-spaces
  364:21  warning  'version' is already defined                            no-redeclare
  365:17  error    Expected space(s) after "switch"                        keyword-spacing
```

dkanada: If we release an app compatible with all tizen devices I would say we can ignore tizen browser support from that point. Depending on how old tizen can be I would even say just the latest versions are fine.

dkanada: I would say browser support is required until we can apply for an official release though, and I have no idea what the timeline is for that.

dmitrylyzo: > If we release an app compatible with all tizen devices I would say we can ignore tizen browser support from that point.

Not "if" but "when" :wink: 
iirc, this is a second place where Tizen API is used, and we could remove both.
Btw, I already checked what I suggest.

>  Depending on how old tizen can be I would even say just the latest versions are fine.

Currently, we already support Tizen 3+ (with `query-string` being "babeled"). There was issue from a guy who owned 2.3 or 2.4, but he didn't publish results or just had no time, because building a Tizen app is a pain.

Nazar78:
Ok noted.

So far I tested the H264 level 5.2 all ok on latest version of Chrome, Chrome mobile and Edge. Not sure though it's ok on other browsers because seems even on my TV the app mode it can play but not it's browser. Also on Firefox it's not smooth, kept buffering sometimes even stall at start of video.

So I was focusing on getting the H264 level 5.2 running my Tizen 5.0 2019 TV app mode, hence this PR. Any further advise? Thank you!

dmitrylyzo:
> but not it's browser.

How did you check that? I mean, did you force it with `maxH264Level = 52` or used my suggestion? Because `tizen` object is not available in browser.

Nazar78:
I tried the following two ways both it won't play:

1. Point the Samsung native browser to Jellyfin-web with forced `maxH264Level = 52`.

2. Point the Samsung native browser directly to the video served in nginx.

dmitrylyzo:
> I tried the following two ways both it won't play:
> 
>     1. Point the Samsung native browser to Jellyfin-web with forced `maxH264Level = 52`.
> 
>     2. Point the Samsung native browser directly to the video served in nginx.

OK. So we should disable DTS both in app and browser for Tizen 4+, and enable 52 level only in app for Tizen 5+.

I would do it simplest way :p
```js
// DTS audio not supported in 2018 models (Tizen 4.0)
if (browser.tizenVersion >= 4) {
    supportsDts = false;
}
```
and
```js
// Support H264 Level 52 (Tizen 5.0) - app only
if (browser.tizenVersion >= 5 && window.NativeShell) {
    maxH264Level = 52;
}
```

Nazar78:
Thanks a lot @dmitrylyzo. The `browser.tizenVersion` and `window.NativeShell` works fine on my TV app.

PR #816, opened 2020-02-16
Make the console quieter by default

MrTimscampi:
**Changes**

Changes most/all the `console.log()` to use the proper channel depending on the message.

Makes the client a lot quieter for most users, while keeping debug messages and outputting errors as actual errors.

**Issues**

None

PR #1004, opened 2020-03-29
Don't use Globalize globally

ferferga:
Follow-up from #960 

Globalize was used globally mostly everywhere. Now it's used locally.


ferferga:
For some unknown reason ``yarn serve`` doesn't work in my branch (``connectionManager`` undefined``, as I had while working on #960), although I don't see why, the changes in the commits are pretty straightforward. Any feedback on this would be really appreciated!

dmitrylyzo:
`appHost` requires `globalize` requires `connectionManager` which is not set yet in `defineConnectionManager`.

ferferga:
Wow, thank you very much @dmitrylyzo, I've been stuck with this three days in my other branch.

The only way I see to use globalize locally in ``appHost`` is to move the [event handler](https://github.com/jellyfin/jellyfin-web/blob/master/src/scripts/globalize.js#L256-L261) in ``globalize`` to ``site.js`` or somewhere else.

Although it seems more reasonable to me to simply leave ``appHost`` as ``site.js``, using globals, as they're the first chain in the startup process.

dmitrylyzo:
`globalize` will be replaced.

I think event handlers should be taken out.
Could be connected here:
https://github.com/jellyfin/jellyfin-web/blob/0afe5f62ef5f709db5710174aa5b68f6e7f5888b/src/scripts/site.js#L441-L445
Exactly after connection manager is created.

cc @MrTimscampi

ferferga:
@dmitrylyzo The problem is that the createConnectionManager function also requires appHost. They're indivisible each other.

Or I'm missing something...

dmitrylyzo:
`globalize` should not require `connectionManager` (at least in `define`).
We need to extract at least this
https://github.com/jellyfin/jellyfin-web/blob/0afe5f62ef5f709db5710174aa5b68f6e7f5888b/src/scripts/globalize.js#L256

I think we could just export `updateCurrentCulture` and connect it to `connectionManager` where I said like this
```js
require(['globalize', 'connectionManager', 'events'], function (globalize, connectionManager, events) {
    events.on(connectionManager, 'localusersignedin', globalize.updateCurrentCulture);
});
```
_bigger view_
https://github.com/jellyfin/jellyfin-web/blob/0afe5f62ef5f709db5710174aa5b68f6e7f5888b/src/scripts/site.js#L438-L455

ferferga:
Okay, now I understand what do you mean. Will try to do it tomorrow and let's see if it works! :)

Thank you very much for your input and your patience @dmitrylyzo! 😊

dmitrylyzo:
In fact, after we replace `globalize` with some cool library, all modules will require it instead of `globalize`. But we still need controller to configure language, locale (culture).

Currently, `globalize` just combines both library and controller. So we could split it. But I am not sure if it is worth now.

MrTimscampi:
> with some cool library

Speaking of which, I was thinking about moving to this for translations: https://www.i18next.com/

ferferga:
@MrTimscampi Will take a look at it!

MrTimscampi:
@ferferga This is a bit large lol You might want to cherry-pick the commits to different branches and open a bunch of PRs.

ferferga:
This builds but webSettings is undefined (#816)

ferferga:
@MrTimscampi That's why I told you that this would be a monster on Matrix 🙃.

The problem is that the new PR I open will be blocked until this one is merged (as this depends on the changes made to globalize)

MrTimscampi:
> The problem is that the new PR I open will be blocked until this one is merged (as this depends on the changes made to globalize)

That's not an issue, I do it all the time :) We can merge this one, then a simple rebase on master will get rid of the commits and then review the other one.

It'll be easier to figure out where errors come from as well (which is why this was planned as a progressive move since a few modules need to be rewritten as classes, the plugin system and themes need a touchup, etc.

ferferga:
@MrTimscampi if you don't mind, I will use this PR as a base to cherry-pick the commits to other branches. I will remove the cherry-picked commits from this branch once they're on another PR and, once that PR is merged, go for another batch, so I don't get crazy and I still have a reference. Is that okay?

The changes will be carried over to the new PRs. I will leave this PR open (no way to mark it as a draft again) so I can have an overview of what's left to be merged and we all can track them easily on #868. So the main aim of this PR is basically dead from now on.

MrTimscampi:
@ferferga Sure, no problem :+1: 

Maybe add "[WIP]" to the issue so it's clear that nobody should review this for now :)

MrTimscampi:
@ferferga Could remove the ES6 commits from this and rebase to fix the eventual merge conflicts so we can get this merged?

ferferga:
@MrTimscampi You mean the use of Globalize globally right? Because the removal of connectionManager from Globalize was already done in another commit.

MrTimscampi:
> @MrTimscampi You mean the use of Globalize globally right? Because the removal of connectionManager from Globalize was already done in another PR.

Yep, make this a mergeable PR so we can get rid of the global Globalize :)

PR #1005, opened 2020-03-29
Switch to new version of JavascriptSubtitlesOctopus, enable new options

JustAMan:
**Changes**

First of all, this makes JavascriptSubtitlesOctopus be used from a fixed tag, so whatever happens later with `master` is irrelevant. Reproducibility yay!

Second, I'm updating it to our new version of renderer with vastly improved performance, but setting the limits relatively low (total memory consumption should stay around ~200 MiB at worst), should be fine even for heavy-animated subtitles on recent enough laptop/desktop clients but requires tuning for low-power devices like Smart TV. Current (pre-update) state is much worse though, as those animated subtitles would most likely hang those TVs to the point of unresponsiveness, so I'm not doing anything bad really :)

**Issues**

Not sure if there are any issues made, but this is to improve responsiveness of our web player when displaying subtitles. It also should make them more in sync with video.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "jellyfin-noto": "https://github.com/jellyfin/jellyfin-noto",
     "jquery": "^3.4.1",
     "jstree": "^3.3.7",
-    "libass-wasm": "https://github.com/jellyfin/JavascriptSubtitlesOctopus",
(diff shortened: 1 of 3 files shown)


JustAMan:
> one for medium ones (phones ?)

Fixing octopus even starting on our Android build is yet to be done :)

PR #1007, opened 2020-03-31
Remove API Client from the tree

MrTimscampi:
**Changes**

Branch to parallel my cleanup of the API Client in order for it to be removed from this repository.

The API Client side is at https://github.com/jellyfin/jellyfin-apiclient-javascript/pull/36

**Issues**

PR #1011, opened 2020-04-01
Add syncplay feature

OancaAndrei:
**Changes**
This PR adds playback syncing of different sessions that are watching the same content.
- Users that are playing something can create a group
- Multiple users can join a group to sync the same media
- Playback (play, pause, seek) is synced on all players of a group

The implementation tries to be time and network resistant, meaning that syncing is not based on merely sending the _playback command_ but also tells the clients _when_ to actually run said command. This means shifting the network latency problem to a time synchronization one.

Ideally, the client's time should be correct but, in the event this isn't true, a server time estimation is made by the client (clients always use this at the moment).

A major problem faced was the delay caused by issuing a play command on the web-player and it actually starting the playback. This issue has been solved by speeding up the playback to catch up.
In other words, to implement this feature on other players, the "playback rate" (or speed) should be supported by the player to achieve optimal results.

Feedback and change proposals are welcome.
I'm curious to see whether the current UX is ideal or if some things should be revised.

**Planned features for this PR**
- [x] Option to toggle feature for specific user [User edit > Media Playback > Syncplay access]
- [x] Filter groups by library access and parental rating (needs review)

**Proposed features for future PRs**
- Share the transcoding process with the members of the group
- Manually adjust delay
- Time sync between devices in the same network
- Group types (public, private, personal)
- Group join by PIN

**Features being tackled in next PR**
- Sync playlist as well (more discussion needed)
- Group wait when one or more members of a group are buffering

**Issues**
Addresses some requested features of https://github.com/jellyfin/jellyfin/issues/227.


anthonylavado:
Link to other issue: https://github.com/jellyfin/jellyfin/pull/2733

dkanada:
This isn't a requirement but we have recently added [ES6 support](https://github.com/jellyfin/jellyfin-web/issues/868) so you could potentially port the new files in this pull request as an improvement.

src/libraries/apiclient/apiclientcore.js
```diff
@@ -1530,6 +1530,15 @@ define(["events", "appStorage"], function(events, appStorage) {
...
         })
+    }, ApiClient.prototype.sendSyncplayCommand = function (sessionId, command, options) {
+        if (!sessionId) throw new Error("null sessionId");
+        if (!command) throw new Error("null command");
+        var url = this.getUrl("Syncplay/" + sessionId + "/" + command, options || {});
+        this.ensureWebSocket();
+        return this.ajax({
+            type: "POST",
+            url: url
+        })
```
dkanada, line 1541: We are also extremely close to migrating our API client to a [separate repository](https://github.com/jellyfin/jellyfin-apiclient-javascript) so it can be used in other projects, courtesy of @MrTimscampi. It should be much easier to work with than the mess we have here.

OancaAndrei: I definitely missed that repository, ops. I'll look into it.

dkanada: Don't worry about it :smile: we haven't exactly been advertising the change yet.

MrTimscampi: It's still being heavily messed with since it's not 100% compatible with the web client yet...

No ETA on the move though, so definitely keep it in there :) Once your PR here is merged, feel free to open one over there to add it, or I'll port it there myself if you don't have time ;)

Artiume:
Seeing how it ties into everything, I'm not sure it's possible, but can you make it to where you can make a Group before you start playback? If not, perhaps we could add a warning or something to say that you have to create a Group during playback. Took it for a test run, it looks awesome.

src/libraries/apiclient/apiclientcore.js
OancaAndrei: No worries, plenty of time on my end as well at the moment :wink:
I'll keep polishing it here then.

OancaAndrei:
@Artiume, I'm glad you're liking it. What you're proposing is totally doable, that's how it worked when I was exploring the implementation. Then I changed it, mainly to avoid extra checks that ensured the same content was playing on all clients. So yeah, I kinda cheated :laughing:. 

But again, I'm really curious about the different use-cases you all have in mind. We should gather some, so that we can eventually make a mix of them.

So, to expand on your idea, a user creates a group, people join it and as soon as anyone starts playing something, all players are told to load the content. Right?
Also, what should happen in case anyone hits the back button? Shall everyone stop the playback and return to the home screen? Or shall they continue, and that single user just leaves the group?

Nickbert7:
Can you please also add an option in the usersettings for admins to disable/enable this for specific users?
It is currently also possible to join a group playing a video of a library the user has no access to.

OancaAndrei:
> Can you please also add an option in the usersettings for admins to disable/enable this for specific users?

@Nickbert7, should this option just disable the ability to create a group but still allow the user to join one? Or you mean like disabling the whole feature?

Right now I'm imagining two checkboxes:
- :ballot_box_with_check: Allow joining a syncplay group
- :ballot_box_with_check: Allow creating a syncplay group

where the second option is available only if the first one is enabled.

MrTimscampi:
> Right now I'm imagining two checkboxes

This is the best option for handling rights around this feature, I think.

dkanada:
I think this deserves its own section in the user settings like transcoding.

> It is currently also possible to join a group playing a video of a library the user has no access to.

I'm pretty sure this is what we want, because the alternative sounds somewhat useless for users and difficult to handle on the backend. We could potentially add a third setting allowing access to any content during sync regardless of permissions, but I don't think it's a blocker.

We might want an option for private groups that have to invite users, although I'm not sure about this yet. It might be too complex for my taste.

OancaAndrei:
> I'm pretty sure this is what we want, because the alternative sounds somewhat useless for users and difficult to handle on the backend. We could potentially add a third setting allowing access to any content during sync regardless of permissions, but I don't think it's a blocker.

Just to make sure I read this correctly haha, you're saying that a group filtering should be done, based on what the user has access to. If this is the case then I totally agree.

dkanada:
How are active groups stored on the backend? I'm still trying to think of the best structure for group access control to be honest, haven't looked at the server changes yet. We also might want to think of a better name than groups now that I think of it.

OancaAndrei:
At the moment they're only stored in memory, in a Dictionary, and are lost on a server restart.
What about of "rooms"? Now that you bring up the topic of names, "Syncplay" is open for changes as well.

Nickbert7:
> > Can you please also add an option in the usersettings for admins to disable/enable this for specific users?
> 
> @Nickbert7, should this option just disable the ability to create a group but still allow the user to join one? Or you mean like disabling the whole feature?
> 
> Right now I'm imagining two checkboxes:
> 
> * ☑️ Allow joining a syncplay group
> * ☑️ Allow creating a syncplay group
> 
> where the second option is available only if the first one is enabled.

Those options are fine for me.
I just want a way to prevent someone joining a session with content they should not see. Either by disable it complete or via parental control.

dkanada:
Does Netflix have exclusive use of party? That seems like a good one because it can be used for both the feature and the individual instances, but I'm not sure we can use it. If we keep the existing name I think `SyncPlay` or `Sync Play` is preferable.

MrTimscampi:
> Does Netflix have exclusive use of party? That seems like a good one because it can be used for both the feature and the individual instances, but I'm not sure we can use it. If we keep the existing name I think `SyncPlay` or `Sync Play` is preferable.

It seems a bit too generic for a trademark. I'd avoid using "Jellyfin Party", but calling the feature "Party Mode" and the groups "Party" (Like "Join party") should be good.

dkanada:
@OancaAndrei could this feature potentially work with any media type, or is it somewhat limited to videos? This seems like it would be perfect for multi-room speakers and other things besides videos.

OancaAndrei:
> @OancaAndrei could this feature potentially work with any media type, or is it somewhat limited to videos? This seems like it would be perfect for multi-room speakers and other things besides videos.

No limitations, music works as well. The only requirement for an almost perfect match is the playback speed support by the player.

One of my end goals, eventually, is to watch a movie on a TV screen and have the headphones of my phone play only the audio for that movie. I know, it's a rather peculiar setup :sweat_smile:, but should be feasible.

dkanada:
> One of my end goals, eventually, is to watch a movie on a TV screen and have the headphones of my phone play only the audio for that movie. I know, it's a rather peculiar setup sweat_smile, but should be feasible.

I just realized this would be a possible addition and am even more excited! We'll need to think carefully about the UI for all these options though, it sounds like a lot of complexity.

sparky8251:
On the topic of the name for this, I'd vote against anything "party" if the eventual end goal is to add synced audio playback between your TV and phone.

I'd lean more to sync, but... There are already sync features planned. Offline playback with sync of watch status etc.

Casting is already taken too...

Naming is not east lol

KeepBotting:
https://syncplay.pl/

Syncplay already exists

Jellyfin Sync?

Multifin?

dkanada:
I just realized we implement most of the "Session" features such as remote control, display mirroring, and casting in a single location. After we add this feature, we could unify all of them under a single umbrella term and think of a good way to display everything.

dkanada:
Some options for the unified session icon include [device hub](https://material.io/resources/icons/?icon=device_hub&style=baseline), [an interesting upload-ish icon](https://material.io/resources/icons/?icon=keyboard_capslock&style=baseline), [this cloud](https://material.io/resources/icons/?icon=wb_cloudy&style=baseline), or a [generic share icon](https://material.io/resources/icons/?icon=share&style=baseline) to replace the cast icon that currently resides on the header.

MrTimscampi:
> Some options for the unified session icon include [device hub](https://material.io/resources/icons/?icon=device_hub&style=baseline), [an interesting upload-ish icon](https://material.io/resources/icons/?icon=keyboard_capslock&style=baseline), [this cloud](https://material.io/resources/icons/?icon=wb_cloudy&style=baseline), or a [generic share icon](https://material.io/resources/icons/?icon=share&style=baseline) to replace the cast icon that currently resides on the header.

What about [ondemand_video](https://material.io/resources/icons/?icon=ondemand_video&style=baseline) ? Or a cheeky [fireplace](https://material.io/resources/icons/?icon=fireplace&style=baseline).

[dynamic_feed](https://material.io/resources/icons/?icon=dynamic_feed&style=baseline) might be a great choice. [subscriptions](https://material.io/resources/icons/?icon=subscriptions&style=baseline) also fits.

For the name, "Multi-play" fits every use case: play on multiple devices to enjoy remote audio, play on multiple devices with friends, ...

dkanada:
I don't think those icons match the functionality enough, and I'm also not a fan of MultiPlay yet, but we have lots of time to consider everything.

YouKnowBlom:
I just tried this out with a friend and I'm blown away by how well it works! The timings are near perfect if not perfect. Great work man!

iwalton3:
When the API for this is finalized I definitely want to get this working in the MPV Shim client.

Artiume:
> One of my end goals, eventually, is to watch a movie on a TV screen and have the headphones of my phone play only the audio for that movie. I know, it's a rather peculiar setup , but should be feasible.

Brilliant 

> Also, what should happen in case anyone hits the back button? Shall everyone stop the playback and return to the home screen? Or shall they continue, and that single user just leaves the group?

> This is the best option for handling rights around this feature, I think


So I was thinking. there's a couple different use cases. Personal use of wanting to use two devices at once. Having a private group they can join with a pin. Or a public group. Perhaps three check boxes for each scenario? One box to share within your signed in devices for that user. One box to make private sessions that could have names or be nameless and you join using a pin. Then public sessions that anyone on the server can see and join (maybe enable pins as an option as well?) 

> I just want a way to prevent someone joining a session with content they should not see. Either by disable it complete or via parental control 

I like this idea. Parental controls would definitely be a plus

TheNamelessWonderer:
This looks great! I wonder if it would possible to set a max bitrate for the group. This would ensure that one user can’t choose a bitrate so high that the connection doesn’t have enough bandwidth for other users. Furthermore some users might be on a local network and thus wouldn’t need this restriction. I understand you can limit it in the user settings but this means that when not in a group they can’t take advantage of the extra bandwidth and use a higher bitrate.

Artiume:
> This looks great! I wonder if it would possible to set a max bitrate for the group. This would ensure that one user can’t choose a bitrate so high that the connection doesn’t have enough bandwidth for other users. Furthermore some users might be on the local network and thus wouldn’t need this restriction. I understand you can limit it in the user settings but this means that when not in a group they can’t take advantage of the extra bandwidth and use a higher bitrate.
> 
> Also does this require the user to have an account or would joining via a public link be possible?

@OancaAndrei can correct me where I'm wrong, but as is, I want to say every stream in bitrate independent already (that's actually one of its current issues we want to resolve, allow two people to share a stream to reduce resource load).

I would also say that you probably do need an account for this. Until we do the database rebuild, a lot of the dynamic features requested around user accounts and libraries is gonna be on hold most likely.

dkanada:
It will be extremely hard to get the streams to share files as far as I know, so it might be better to leave that for another pull request. Although, I think one of the major issues was clients requesting different positions, so it might be possible with some changes to how the backend handles the stream chunks.

@JustAMan is probably the best person to ask about those problems.

Artiume:
> It will be extremely hard to get the streams to share files as far as I know, so it might be better to leave that for another pull request

Yeah. Another issue for not just figuring out how to have clients share a stream will be when more than just the browsers have the sync play capability. Once Android TV has it, the backend will have to assess a profile compatible for both clients (saying one is browser and one is the atv for instance)

OancaAndrei:
I've explored the code a bit, but I'm not ready to make changes yet.

The idea on optimizing the transcoding, I think, would be to group the clients with similar capabilities and same bitrate settings and make them use the same stream.
Also, instead of accesing the ffmpeg stream directly, some sort of middleware might be needed, so that parts that have already been encoded can be fed directly to the requesting client.
This extra layer would have to keep track of clients using a TranscodingJob (in order to stop unused jobs), and cache those encoded segments (with proper information ).

But I might be missing stuff :sweat_smile:, still wrapping my head around the code.
Also, I don't mind leaving this optimization for a future PR.

JustAMan:
I think if clients in a group aren't far apart from each other and are using HLS as transport for transcoded data (and are using identical settings for dimensions, bitrate, etc.) then there would probably be only one transcoding job, and clients would get files delivered from it. So you don't have to optimize transcoding (at least not now). Remember, "premature optimization is the root of evil" ;)

As for naming... a play around "Bonfire" comes to my mind, like if everyone gathers around a bonfire and share a view of something.

Artiume:
So there's two different features here. One is for an individual to share amongst devices for themselves such as the TV and phone situation. The other is sharing a show with friends. Perhaps we should seperate their naming and how we present them?

JustAMan:
The only difference I see here is different users joining from multiple devices vs single user joining from multiple devices. Not sure we need to draw a line here.

hamburglar2160:
This is neat! Perhaps with the multiple devices for a single user they can get checkboxes in the "Play On" menu to select which devices/Google Cast devices they want to join with. Then there'd be another subsection in the "Play On" menu for user groups?

How would this go with music if playlist sync is added? Would it broadcast the same metadata sort of as Shoutcast does?

if the multiple devices thing for a single user is pursued it'd be worth considering that Google Cast can device and sync multiple audio assistant devices but not video.

MayeulC:
> Does Netflix have exclusive use of party? That seems like a good one because it can be used for both the feature and the individual instances, but I'm not sure we can use it. If we keep the existing name I think `SyncPlay` or `Sync Play` is preferable.

For naming, "Multiplayer" comes to my mind. Short and to the point. Ampache has a "democratic play" option as well, where users can vote for the next title.

Regarding implementation, it looks like you are just reading the time value out of the server. This is mostly fine, but will not correct for latency, which might or might not be an issue. Using something like [NTP](https://en.wikipedia.org/wiki/Network_Time_Protocol#Clock_synchronization_algorithm), you can also extract this offset: the basic idea is to ping the server to measure latency. Unsure if the extra precision is worth the additional work?
Also unsure if multicast could be done across the Internet on IPv6 networks?

OancaAndrei:
@MayeulC, the "democratic play" could be an interesting feature to add.

> Regarding implementation, it looks like you are just reading the time value out of the server.

The code might not be commented enough just yet but the current implementation does the following:
- Client saves its current time (`pingStartTime`);
- Client asks the server for its time (`serverTimeUTC`);
- Clients saves the time when request is complete (`pingEndTime`);
- Client computes `ping` (using `pingEndTime` and `pingStartTime`);
- Client estimates `time offset` using `ping` and `serverTimeUTC`;
- Client repeats after a while and updates `time offset` only if ping was faster.

```
S:         [2]
C: [1]<----> <---->[3]
       ping   ping

[1] pingStartTime
[2] serverTimeUTC
[3] pingEndTime

delay = pingEndTime - pingStartTime
ping = delay / 2
offset = serverTimeUTC - (pingEndTime - ping)
```

It's surely not the NTP protocol, but it's still something that attempts to estimate the time offset between the client and the server. 
Nonetheless, I should've definitely given NTP a look, maybe I'll rewrite that part (it's not a big change anyway). I think that NTP takes into consideration server load as well (like execution time).

> This is mostly fine, but will not correct for latency, which might or might not be an issue.

Events are then timed according to the server's time. This way clients know when a certain event needs to be executed (because they know the `time offset` and the `event time` relative to the server). Clients also report to the server their last `ping`.

This means that the server can pick a suitable time in the future (the `event time`) that accounts for latency and also gives space to the clients to schedule the event (currently the highest `ping` in a group is picked and doubled).

> Also unsure if multicast could be done across the Internet on IPv6 networks?

At this stage I'm not considering the streaming part just yet, only syncing playback events, as suggested previously :smile:.

MrTimscampi:
Is the ping generic? If so it could be re-used for #1067

OancaAndrei:
For anyone out there using Safari, I'm aware of the rather peculiar issues that exist.
Looks like Safari, after a while and for no apparent reason (at least to me), just stops calling the callback set for the `timeupdate` event. Using `setTimeout` or `setInterval` doesn't seem to help either. Maybe because it gets called to often? I have no clue.
The result is that sometimes syncing during playback might not work.

I'm posting this here just to let you know I haven't given up :sweat_smile:.

Oh, and `playbackRate` looks broken as well on Safari.

OancaAndrei:
> Is the ping generic? If so it could be re-used for #1067

@MrTimscampi, I would say that it is, meaning that it could be placed in its own module, something like `KeepAliveManager`.

At the moment clients send a `KeepAlive` to the server by making a `GET` request, to which the server responds back with a `KeepAlive` event through the WebSocket for that specific session.

In other words, as differently proposed in #1067, the `KeepAlive` is initiated by the client, mainly because it has a new ping to report. But this can be changed to accommodate the other issue as well, I guess.

For optimal results I would propose the following approach.
Clients initiate a `KeepAlive` every `ClientLostTimeout / 0.25` seconds, that refreshes their `last activity timestamp` on the server.
The server can then check the `last activity timestamp` of all clients every `ClientLostTimeout / 0.5` seconds.
Clients found exceeding a certain threshold (lets say `ClientLostTimeout * 0.75`) are directly contacted by the server and urged to update their timestamp.
Clients found exceeding the `ClientLostTimeout` are considered to be lost.

This solution has the advantage of distributing the requests more evenly over time, instead of having all the clients make their request at the same time.

MrTimscampi:
> > Is the ping generic? If so it could be re-used for #1067
> 
> @MrTimscampi, I would say that it is, meaning that it could be placed in its own module, something like `KeepAliveManager`.
> 
> At the moment clients send a `KeepAlive` to the server by making a `GET` request, to which the server responds back with a `KeepAlive` event through the WebSocket for that specific session.
> 
> In other words, as differently proposed in #1067, the `KeepAlive` is initiated by the client, mainly because it has a new ping to report. But this can be changed to accommodate the other issue as well, I guess.
> 
> For optimal results I would propose the following approach.
> Clients initiate a `KeepAlive` every `ClientLostTimeout / 0.25` seconds, that refreshes their `last activity timestamp` on the server.
> The server can then check the `last activity timestamp` of all clients every `ClientLostTimeout / 0.5` seconds.
> Clients found exceeding a certain threshold (lets say `ClientLostTimeout * 0.75`) are directly contacted by the server and urged to update their timestamp.
> Clients found exceeding the `ClientLostTimeout` are considered to be lost.
> 
> This solution has the advantage of distributing the requests more evenly over time, instead of having all the clients make their request at the same time.

Sounds good :) We could kill two birds with one stone.

MayeulC:
@OancaAndrei, thank you for the explanation! I must admit I had a very cursory look at the diff from Github, and didn't find the relevant section :sweat_smile: 

Your algorithm sounds sane, though it is true that looking at some NTP implementations could give you ideas about how to better handle synchronization and stay synced. :)

----
I guess it would be better to start simple, and improve the precision later if needed, but
<details><summary>here are a few ideas.</summary>
<ul>
<li><p>The javascript implementation could give you some variance due to how the code is scheduled, I wonder if using <code>event.timeStamp</code> or a similar "native" timestamping feature couldn't get better results.</p></li>

<li><p>The browser can <a href=https://developer.mozilla.org/en-US/docs/Web/API/Event/timeStamp#Reduced_time_precision>reduce timestamping precision</a> to resist fingerprinting. Various signal processing techniques can be applied to extract the value from the quantization noise introduced, like sending packets every millisecond or so, and see after how many packets the timestamp changes.</p></li>
</ul>
</details>

Such techniques might not be necessary if the goal isn't to perfectly sync audio (<a href=https://github.com/badaix/snapcast>snapcast</a> does that, it could be another inspiration source).

MayeulC:
> Clients initiate a `KeepAlive` every `ClientLostTimeout / 0.25` seconds, that refreshes their `last activity timestamp` on the server.

I initially read that as `ClientLostTimeout = 0.25s`, and was about to say it's overkill. That sounds fine (only it should be eithte `/4` or `*0.25`, likewise for `*0.5`). How are you planning to handle one of the clients buffering, though?

OancaAndrei:
@MayeulC I knew it! I knew somebody else had implemented the same idea before me :laughing:, Snapcast has roughly the same algorithm to sync the audio (an almost perfect time synchronization and speed control of playback to resync).

I must also thank you for the link on the _Reduced time precision_, interesting to know.
In the previous post I misused the term `events` (I was referring to the _playback command_ that the server sends to all clients), but I guess that the rounding happens also when creating  a `new Date()`. From the specs I get that a 100ms error could be present (in case the user enabled that privacy feature), which is definitely notable, especially when listening to music.

OancaAndrei:
> That sounds fine (only it should be eithte `/4` or `*0.25`, likewise for `*0.5`). 

Ops, what a typo, thanks for spotting it.

> How are you planning to handle one of the clients buffering, though?

The initial idea was to have two ways of handling it:
- Group wait, where all members of a group wait for the buffering ones;
- Silent sync, where a single member resyncs itself either by skipping to the correct position or by speeding up the playback to catch up.

The second method is more suitable in cases where a client lags behind others just for a couple of seconds, while the first method is preferred when buffering takes longer.

Currently only the second method is implemented.

dkanada:
I guess some new plans for this pull request would include making the heartbeat generic so we can use it for other things, and also I'd like to get the interface stamped before we merge this so there are no UI changes in the future. That might depend on how your group feature is implemented. Is this a master-slave scenario with the original device handling the play state and current media, or can any device in the group control those things?

OancaAndrei:
> Is this a master-slave scenario with the original device handling the play state and current media, or can any device in the group control those things?

I'm leaning more towards a peer scenario, where everyone (any _device_) can control the play state.

Regarding the media that's playing (or the playlist) I would instead restrict it to the _user_ that created the group (as to avoid mayhem where everyone changes the current media continuously or even by mistake).

dkanada:
We have quite a few options piling up so I'll try and list everything below in a nice list. One thing to keep in mind is that this feature will be useless if the design is so complex no clients actually implement everything.

## Interface

There will be a button on header which will be the main entrypoint for any non-local playback. A secondary button will exist on the video player, which currently uses the header but in the future will be a separate element.

## AirPlay

I have no idea what icon this uses or if it currently uses the chromecast menu, but we should move all these devices there as well. Can multiple devices control a single AirPlay device at the same time?

## Chromecast

Same exact behavior as now. This currently functions similar to AirPlay as a remote control, but both are unique because they can have multiple controlling devices.

## Remote Control

Similar to the above but without multiple controllers.

## Sync Play

There are two possible methods to display this feature. We could either treat them separately from devices, or try and mirror Chromecast behavior by simply displaying the device itself as a "public" session to certain users. This will heavily depend on how we want to deal with the related settings, which I have ranked in order of how annoying they are when abused.

**Shared Controls:** Modify the playback state for the current item

**Shared Queue:** Changing the current item or otherwise modifying the queue

This becomes even more complicated because we could split "Shared Queue" into two categories. I would consider modifying the playlist totally fine but changing the current item would be annoying.

**Shared Display:** The new name for display mirroring

## Broadcast

A long requested feature has been to flag devices as public to enable multiple controllers, which is extremely relevant here. @iwalton3 is one of the developers who suggested it, and the goal would be to mirror Chromecast or AirPlay functionality. This would be useful for public devices like Android TV or the MPV Shim Client. I haven't figured out an elegant way to integrate this with the current plans yet, but @OancaAndrei might have ideas since this will probably use SyncPlay behind the scenes.

## More Settings

I don't think any of these are required for the initial implementation because they are all somewhat complex. We should consider if these are useful or just introduce bloat.

**Sync Method:** This would be either `Wait` or `Skip` as defined above

**Accessible Content:** Only allowed content or any synced content

When a user tries to join a SyncPlay session with content they can't access and this is enabled they would get blocked. However, if someone tries to play content in a session that someone can't access, we could either kick them or throw an error on the request.

**Semi-Public Sessions:** Only share this content with certain users

**Passwords:** This could be a good alternative to the semi-public sessions

dkanada:
Another concern I forgot to mention is deciding which settings will be permissions available on the dashboard for administrators. Basically all of the things I listed could be toggled for individual users.

ladrua:
Couldnt this be solved with a "Streaming" solution? So a user creates a stream that others can join? Also would solve multiple transcoding of same content, and also could be used to create sort of a linear "tv channel". Disclamer, I just found Jellyfin today, and was researching these posibilities.

OancaAndrei:
> Couldnt this be solved with a "Streaming" solution? So a user creates a stream that others can join? Also would solve multiple transcoding of same content, and also could be used to create sort of a linear "tv channel". Disclamer, I just found Jellyfin today, and was researching these posibilities.

@ladrua, you mean something like IPTV? I think this would force transcoding even when not needed.

I would say that the problem is not creating a single stream for all devices, but rather making sure that the media gets played at the same time on all clients. Would this "streaming" approach really ensure it?

Quality would also be affected, which is not strictly a bad thing, but still something to consider.

BTW, I'm fairly new to Jellyfin as well, being this my first PR :smile:.

iwalton3:
> Couldn't this be solved with a "Streaming" solution?

This isn't as attractive as a solution because if several people are streaming the media, it imposes the worst quality on all of the people. If one person has a slow internet connection and can only stream 480p, then the server owner who has a 1080p copy is forced to watch the video in 480p as well.

> Would this "streaming" approach really ensure it?

There are definitely buffering and latency considerations. HLS typically works by splitting the video into several files, usually about a second or so in length (depending on keyframes). There may be differences in how quickly some client load the stream due to latency and other factors, which this would need to handle.

If I remember right, one of the goals of this would be to be able to sync audio on a separate device, so for instance, you could use your phone connected to headphones and have it sync with the TV, and you would almost certainly need more robust syncing to do this.

OancaAndrei:
> ## Split Play
> 
> @OancaAndrei this was your eventual goal, but I have absolutely no clue how to display it in the interface. This would only be for video files, but it would route the audio to a separate device during playback.

I should probably clarify this goal of mine :smile:. This "split play" feature is more of a specific use-case really, that can be obtained by combining manually this "Syncplay" feature with the ["Audio only"](https://features.jellyfin.org/posts/533/audio-only-option-on-video-playback) one.

This could be achieved by the user by making the screen and the headphones join the same group, then the screen device would play the media with its settings (quality, subtitles, etc.) while the headphones would play the media by having the quality set to "Audio only" (or something similar).

So it doesn't really need a specific place in the UI, or even a name.
At the same time I'm not saying that, if this use-case was to become mainstream, it should never be incorporated in some kind of shortcut.

OancaAndrei:
> It's surely not the NTP protocol, but it's still something that attempts to estimate the time offset between the client and the server.

So today I was separating the time sync part into its own module as also rewriting it following the NTP protocol.

In doing so I realized that what I have is already the NTP protocol, only simplified. This is because NTP was designed for a state-less type of connection, while in this case HTTP is used.
Taking the _offset_ formula `((t1 - t0) + (t2 - t3)) / 2` on such a high level results in `t1` and `t2` both having the same value `tx`, which leaves us with:
```
offset = ((t1 - t0) + (t2 - t3)) / 2
       = ((tx - t0) + (tx - t3)) / 2
       = (2 * tx - (t0 + t3)) / 2
       = tx - (t0 + t3) / 2
```
which can be renamed to:
```
tx -> serverTimeUTC
t0 -> pingStartTime
t3 -> pingEndTime

offset = serverTimeUTC - (pingStartTime + pingEndTime) / 2
       = serverTimeUTC - (pingStartTime + pingEndTime + pingEndTime - pingEndTime) / 2
       = serverTimeUTC - (pingStartTime + 2 * pingEndTime - pingEndTime) / 2
       = serverTimeUTC - (2 * pingEndTime + pingStartTime - pingEndTime) / 2
       = serverTimeUTC - (pingEndTime - (pingEndTime - pingStartTime) / 2)
```
and then introducing:
```
delay = pingEndTime - pingStartTime
ping = delay / 2
```
we obtain:
```
offset = serverTimeUTC - (pingEndTime - ping)
```
The implementation will treat `t1` and `t2` like two separate values nonetheless.

#

On a more mattering note, I was considering to piggy-back on this new `TimeSync` service the `KeepAlive` one. The time polling is already there, might just use it to send the _KeepAlive_ message as well. What do you think?

jfu-jfu:
This feature looks so good. I can't wait for it to land.

On the name (if it needs one) I liked the direction @JustAMan was going with 'Bonfire' analogy. Along a similar trend I thought 'Aquarium' fits with the whole Jellyfin portuguese man-of-war fish thing.

Looking forward to it!

dkanada:
> Along a similar trend I thought 'Aquarium' fits with the whole Jellyfin portuguese man-of-war fish thing.

Huh, that's actually pretty good!

@OancaAndrei I think the groups are initiated in a separate button on the header right now, but we want SyncPlay integrated into the Cast button. We also wanted to pick a different icon for that button and since this is your PR you can pick one for the pull request, anything would be an improvement for now.

JustAMan:
> I think the groups are initiated in a separate button on the header right now, but we want SyncPlay integrated into the Cast button.

Let's not add _more_ to this PR than is needed for it to function. Moving icon, adding pretty icon etc. could all be done after core feature starts working.

MayeulC:
> Taking the _offset_ formula `((t1 - t0) + (t2 - t3)) / 2` on such a high level results in `t1` and `t2` both having the same value `tx`, which leaves us with:

Right, this is assuming the server time doesn't change during the round-trip, if I am reading this correctly?

What you did here looks fine, but I'm wondering if this couldn't use a true NTP server somewhere **at a later point**, to piggyback on the robust implementations that already exist in various languages, that use complex phase-locking mechanisms to eliminate jitter and achieve microsecond-level precision in some cases. The android app could [use this](https://github.com/wlach/ntpclient-android), for instance :)

JustAMan:
> achieve microsecond-level precision

For synced video playback? Seriously? 🤣 I mean, we're not launching space rockets here, so we need to stay in sync within human perception, I would say a dozen ms or so is more than acceptable.

src/components/syncplay/timeSyncManager.js
```diff
@@ -0,0 +1,209 @@
...
+     * @param {Date} t1 Server's timestamp of the request reception
+     * @param {Date} t2 Server's timestamp of the response transmission
+     * @param {Date} t3 Client's timestamp of the response reception
+     */
+    constructor(t0, t1, t2, t3) {
+        this.t0 = t0.getTime();
+        this.t1 = t1.getTime();
+        this.t2 = t2.getTime();
+        this.t3 = t3.getTime();
+    }
```
JustAMan, line 35: it's better to have descriptive names instead of relying on docs only, like
```suggestion
    /**
     * Creates a new measurement.
     * @param {Date} requestSent Client's timestamp of the request transmission
     * @param {Date} requestReceived Server's timestamp of the request reception
     * @param {Date} responseSent Server's timestamp of the response transmission
     * @param {Date} responseReceived Client's timestamp of the response reception
     */
    constructor(requestSent, requestReceived, responseSent, responseReceived) {
        this.requestSent = requestSent.getTime();
        this.requestReceived = requestReceived.getTime();
        this.responseSent = responseSent.getTime();
        this.responseReceived = responseReceived.getTime();
    }
```

src/components/syncplay/timeSyncManager.js
```diff
@@ -0,0 +1,209 @@
...
+
+        this.startPing();
+    }
+
+    /**
+     * Gets status of time sync.
+     * @returns {boolean} _true_ if a measurement has been done, _false_ otherwise.
+     */
+    isReady() {
+        return this.measurement ? true : false;
```
JustAMan, line 79:
```suggestion
        return !!this.measurement;
```

src/libraries/apiclient/apiclientcore.js
```diff
@@ -144,14 +144,31 @@ define(["events", "appStorage"], function(events, appStorage) {
...
+            instance.sendWebSocketMessage("KeepAlive");
+            scheduleKeepAlive(instance, msg.Data);
         }
         events.trigger(instance, "message", [msg])
     }
 
+    function scheduleKeepAlive(instance, timeout) {
+        if (instance.keepAliveInterval) {
+            clearInterval(instance.keepAliveInterval);
+            instance.keepAliveInterval = null;
```
JustAMan, line 165:
```suggestion
```
you're going to overwrite it next line, no need to clean it before that

src/scripts/librarymenu.js
```diff
@@ -12,6 +12,7 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
         html += "</div>";
         html += '<div class="headerRight">';
         html += '<span class="headerSelectedPlayer"></span>';
+        html += '<button is="paper-icon-button-light" class="headerSyncButton syncButton headerButton headerButtonRight hide"><i class="material-icons">sync_disabled</i></button>';
```
JustAMan, line 15:
```suggestion
        html += '<button is="paper-icon-button-light" class="headerSyncButton syncButton headerButton headerButtonRight hide"><span class="material-icons sync_disabled"></span></button>';
```
we're going to replace `<i>` for `<span>` for icons, let's use `<span>` here from the get go. cc @dmitrylyzo

src/scripts/librarymenu.js
```diff
@@ -84,6 +86,12 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
...
                 headerCastButton.classList.remove("hide");
             }
+
+            var policy = user.Policy ? user.Policy : user.localUser.Policy;
+
+            if (headerSyncButton && policy && policy.SyncplayAccess !== "None") {
+                headerSyncButton.classList.remove("hide");
+            }
         } else {
             headerHomeButton.classList.add("hide");
```
JustAMan, line 96: you should add `headerSyncButton.classList.add("hide");` somewhere here

src/scripts/librarymenu.js
```diff
@@ -177,6 +186,32 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
         });
     }
 
+    function onSyncButtonClicked() {
+        var btn = this;
+
+        require(["groupSelectionMenu"], function (groupSelectionMenu) {
```
JustAMan, line 192: can we `require` all statically on top? cc @MrTimscampi

src/scripts/librarymenu.js
```diff
@@ -177,6 +186,32 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
...
+            groupSelectionMenu.show(btn);
+        });
+    }
+
+    function updateSyncplayIcon(event, enabled) {
+        var icon = headerSyncButton.querySelector("i");
```
JustAMan, line 198: please no, icons should be `<span>`-s now, and the image should be selected by class, not by `.innerHTML` because innerHTML breaks on certain platforms like WebOS. cc @MrTimscampi @dmitrylyzo

src/scripts/librarymenu.js
```diff
@@ -177,6 +186,32 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
...
+        if (enabled) {
+            icon.innerHTML = "sync";
+        } else {
+            icon.innerHTML = "sync_disabled";
+        }
+    }
+
+    function updateSyncplayErrorIcon(event, show_error) {
+        var icon = headerSyncButton.querySelector("i");
```
JustAMan, line 207: same

src/scripts/librarymenu.js
```diff
@@ -147,6 +155,7 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
         }
 
         headerAudioPlayerButton.addEventListener("click", showAudioPlayer);
+        headerSyncButton.addEventListener("click", onSyncButtonClicked);
```
JustAMan, line 158: where's paired `.removeEventListener()`?

src/scripts/librarymenu.js
```diff
@@ -925,6 +961,8 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
         updateUserInHeader();
     });
     events.on(playbackManager, "playerchange", updateCastIcon);
+    events.on(syncplayManager, "SyncplayEnabled", updateSyncplayIcon);
+    events.on(syncplayManager, "SyncplayError", updateSyncplayErrorIcon);
```
JustAMan, line 965: where's paired `events.off()`?

src/strings/en-us.json
```diff
@@ -841,6 +843,17 @@
     "LabelSubtitlePlaybackMode": "Subtitle mode:",
     "LabelSubtitles": "Subtitles",
     "LabelSupportedMediaTypes": "Supported Media Types:",
+    "LabelSyncplayTimeOffset": "Time offset with server:",
```
JustAMan, line 846:
```suggestion
    "LabelSyncplayTimeOffset": "Time offset with the server:",
```

src/strings/en-us.json
```diff
@@ -1003,6 +1016,17 @@
     "MessageUnableToConnectToServer": "We're unable to connect to the selected server right now. Please ensure it is running and try again.",
     "MessageUnsetContentHelp": "Content will be displayed as plain folders. For best results use the metadata manager to set the content types of sub-folders.",
     "MessageYouHaveVersionInstalled": "You currently have version {0} installed.",
+    "MessageSyncplayEnabled": "Syncplay enabled.",
+    "MessageSyncplayDisabled": "Syncplay disabled.",
+    "MessageSyncplayUserJoined": "<b>{0}</b> joined group.",
```
JustAMan, line 1021:
```suggestion
    "MessageSyncplayUserJoined": "<b>{0}</b> has joined the group.",
```

src/strings/en-us.json
```diff
@@ -1003,6 +1016,17 @@
...
+    "MessageSyncplayUserLeft": "<b>{0}</b> left group.",
```
JustAMan, line 1022:
```suggestion
    "MessageSyncplayUserLeft": "<b>{0}</b> has left the group.",
```

gulpfile.js
```diff
@@ -47,7 +47,7 @@ const options = {
         query: ['src/**/*.png', 'src/**/*.jpg', 'src/**/*.gif', 'src/**/*.svg']
     },
     copy: {
-        query: ['src/**/*.json', 'src/**/*.ico']
+        query: ['src/**/*.json', 'src/**/*.ico', 'src/**/*.wav']
```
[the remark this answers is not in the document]
xnaas: Also voting for something other than WAV. 👍

MayeulC:
> > achieve microsecond-level precision
> 
> For synced video playback? Seriously? rofl I mean, we're not launching space rockets here, so we need to stay in sync within human perception, I would say a dozen ms or so is more than acceptable.

Of course :) But if you can achieve that much, you can certainly achieve less by dropping some complexity and constraints, I just said that to emphasize the capabilities of NTP (and there's [PTP](https://en.wikipedia.org/wiki/Precision_Time_Protocol) if you need sub-microsecond :stuck_out_tongue:, also interesting in that article is a mention of [Audio Video Bridging](https://en.wikipedia.org/wiki/Audio_Video_Bridging), which I didn't know about).

 if your eventual goal is to lip-sync streams, you indeed need~12 to 20ms. The current implementation might achieve that on high-quality links, but I think some basic filtering is needed if you want to eliminate some noise (latency jitter), so why not repurpose what already exists? **at a later point**, of course, I don't want to bikeshed this.

Well, about the microsecond value, I'll admit I pulled it out of my hat, as that seemed a reasonable approximation. It turns out [I'm not really far off](https://kb.meinbergglobal.com/kb/time_sync/time_synchronization_accuracy_with_ntp#example_time_synchronization_results_for_linux), confirmation bias might apply :)

I'm quite excited about this PR, I wanted to work on this idea for kodi and submitted a proposal [a number of times in the past](https://forum.kodi.tv/showthread.php?tid=220078) (retrospectively, I should have started work on this outside of the GSoC...).

OancaAndrei:
PTP looks fascinating, if I understood it correctly. It would mean that devices that are on the same network would be better synced together.

This idea of distributed syncing does make a lot of sense in our case: we would have physically distant devices synced under, let's say, those ~100ms, which would be hardly noticeable by the end users, and local devices synced under ~1ms, where delays between audio/video are much more annoying.

This future PR, that focuses on syncing local devices, would look amazing :smile: (inspired by PTP, I'm thinking about picking a _master-local device_ that speaks directly to the _server_ and then syncs _local devices_ according to its time).

So I would say that we can definitely implement the ideas behind the protocols, I just hope that following them precisely, without - in some cases - having the conditions to do so, does not become counter-productive.

MrTimscampi:
@OancaAndrei FYI, #1007 (which removes the API client from this tree completely) is pretty much ready now.

We can probably move the API client part of this PR to https://github.com/jellyfin/jellyfin-apiclient-javascript if you want.

src/components/syncplay/groupSelectionMenu.js
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    apiClient.sendSyncplayCommand(sessionId, "ListGroups").then(function (response) {
+        response.json().then(function (groups) {            
+            var menuItems = groups.map(function (group) {
+                // TODO: update running time if group is playing?
+                var name = datetime.getDisplayRunningTime(group.PositionTicks);
+                if (!inSession) {
+                    name = group.PlayingItemName;
+                }
```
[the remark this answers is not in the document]
OancaAndrei: The idea is to show two types of lists, based on whether the user is playing something:
- if nothing is playing, the user will see a list with all the available groups;
- if an item is playing, the user will only see groups that are playing the same item.

That's why for the second case I thought of showing the "running time" to identify a group.

src/components/syncplay/groupSelectionMenu.js
```diff
@@ -0,0 +1,198 @@
...
+
+    // TODO: should feature be disabled if playback permission is missing?
+    playbackPermissionManager.check().then(() => {
+        console.debug("Playback is allowed.");
+    }).catch((error) => {
+        console.error("Playback not allowed!", error);
+        toast({
+            text: globalize.translate("MessageSyncplayPlaybackPermissionRequired")
+        });
+    });
```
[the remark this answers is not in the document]
OancaAndrei: This `PlaybackPermissionManager` should be adopted also when starting a normal playback.

Currently what happens is that the user clicks on an item they wish to play, the client sends a request to the server which in turn sends the play command asynchronously.

The problem is that in some browsers (mainly Safari I guess) the async play command on the media is seen as an autoplay initiated by the code and not by the user itself. So the play command is rejected with `NotAllowedError`.

The solution is to grab the permission inside a click event generated by the user, in this case when he clicks the `Sync` icon. Ideally this is needed once.

Grabbing permission simply means starting playing something inside the click event. This then allows the code to control the playback later on.

src/components/syncplay/syncplayManager.js
```diff
@@ -0,0 +1,737 @@
...
+        this.roundTripDuration = 0;
+        this.notifySyncplayReady = false;
+        
+        events.on(playbackManager, "playerchange", () => {
+            this.onPlayerChange();
+        });
+
+        this.bindToPlayer(playbackManager.getCurrentPlayer());
+
+        events.on(this, "TimeUpdate", (event) => {
```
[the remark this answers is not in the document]
OancaAndrei: I really find it much harder to read when it's all lower-case, that's it :sweat_smile:.

src/components/syncplay/syncplayManager.js
```diff
@@ -0,0 +1,737 @@
...
+                });
+                break;
+            case 'JoinGroupDenied':
+                toast({
+                    text: globalize.translate('MessageSyncplayJoinGroupDenied', cmd.Data)
+                });
+                break;
+            case 'LibraryAccessDenied':
+                toast({
+                    text: globalize.translate('MessageSyncplayLibraryAccessDenied', cmd.Data)
```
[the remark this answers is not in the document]
OancaAndrei: Ops, some of those are actually copy/paste typos. At the moment only `MessageSyncplayUserJoined`, `MessageSyncplayUserLeft` and `MessageSyncplayGroupWait` receive the name of the interested user.

src/components/syncplay/syncplayManager.js
```diff
@@ -0,0 +1,737 @@
...
+                } catch (error) {
+                    playingItemId = "";
+                }
+                // Sometimes JoinGroup fails, maybe because server hasn't been updated yet
+                setTimeout(() => {
+                    apiClient.sendSyncplayCommand(sessionId, "JoinGroup", {
+                        GroupId: groupId,
+                        PlayingItemId: playingItemId
+                    });
+                }, 500);
```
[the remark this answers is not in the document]
OancaAndrei: It is unfortunately. Without it the client doesn't have the time to update its status server-side, so when joining, even if the media is already loaded, the server will send another `PrepareSession` command thus making the client reload the media.

Now that I think about it, I can probably solve it by ignoring the second `PrepareSession` if the `PlayingItemId` matches the required one.

src/components/syncplay/syncplayManager.js
```diff
@@ -0,0 +1,737 @@
...
+                this.syncEnabled = false;
+                this.syncAttempts++;
+                this.showSyncIcon("SpeedToSync (x" + speed + ")");
+
+                this.syncTimeout = setTimeout(() => {
+                    this.currentPlayer.setPlaybackRate(1);
+                    this.syncEnabled = true;
+                    this.clearSyncIcon();
+                }, SpeedToSyncTime);
+            } else if (absDiffMillis > MaxAcceptedDelaySkipToSync) {
```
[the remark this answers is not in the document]
OancaAndrei: I forgot to mention it in the code as well.

Seeking takes some time, during which playback is halted. This time is also unknown and variable between consecutive calls. Being it so unpredictable, trying to sync by seeking doesn't really work that well (in browsers at least).

During testing I found out that insisting on syncing when delay is less than 300ms makes the playback stutter, which is really annoying. In some cases it did eventually sync under ~ 50ms, but it was rare.

src/components/syncplay/syncplayManager.js
```diff
@@ -0,0 +1,737 @@
...
+     * @param {Object} player The player.
+     */
+    bindToPlayer (player) {
+        if (player !== this.currentPlayer) {
+            this.releaseCurrentPlayer();
+            this.currentPlayer = player;
+            if (!player) return;
+        }
+        
+        // TODO: remove this extra functions
```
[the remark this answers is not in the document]
OancaAndrei: Because the `events` module changes the scope when calling a callback.
Calling for instance `onPlayPauseStateChanged` from the wrong scope breaks things, because `this` points to `player` (the event emitter) instead of pointing to the `SyncplayManager` singleton.

This was a quick fix to avoid spending to much time on it. Is there a cleaner way or should the `events` module be rethinked?

src/components/syncplay/syncplayManager.js
```diff
@@ -0,0 +1,737 @@
...
+        this.syncplayEnabledAt = enabledAt;
+        this.injectPlaybackManager();
+        events.trigger(this, "SyncplayEnabled", [true]);
+
+        waitForEvent(this, "SyncplayReady").then(() => {
+            this.processCommand(this.queuedCommand, apiClient);
+            this.queuedCommand = null;
+        });
+        this.syncplayReady = false;
+        this.notifySyncplayReady = true;
```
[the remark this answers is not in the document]
OancaAndrei: I feel like it's not that relevant as the event is called asynchronously (this function has to terminate before the event can be triggered). But even if it wasn't, it makes more sense to first register the event listener and then tell the trigger code that the event can be triggered, because the listener has been set.

That's my view on it :sweat_smile:. The code can be rearranged but the flow is clearer this way IMO.

src/scripts/librarymenu.js
OancaAndrei: The whole file has no `events.off()`. I'm not sure there's a place where I can unregister those callbacks.

src/scripts/librarymenu.js
OancaAndrei: I have been following the "style" I found. Looks like the callbacks are "removed" in `renderHeader()` by this line:
```
skinHeader.innerHTML = html;
```

It's ugly, but again, I don't see a place where to remove the event listeners. Rewriting how the whole `librarymenu` works seems out of the scope of this PR.

Am I missing some other hidden file? :sweat_smile:

src/components/syncplay/syncplayManager.js
```diff
@@ -0,0 +1,737 @@
...
+        var pauseAtTimeLocal = timeSyncManager.serverDateToLocal(pauseAtTime);
+
+        if (pauseAtTimeLocal > currentTime) {
+            var pauseTimeout = pauseAtTimeLocal - currentTime;
+
+            this.scheduledCommand = setTimeout(() => {
+                playbackManager.syncplay_pause();
+                setTimeout(() => {
+                    playbackManager.syncplay_seek(positionTicks);
+                }, 800);
```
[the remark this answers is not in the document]
OancaAndrei: Hacky way to give the player some time to pause before seeking, otherwise it just breaks. I should probably try to see if replacing it with `waitForEventOnce` works in a reliable way.

gulpfile.js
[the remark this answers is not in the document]
OancaAndrei: I had the belief that a `.wav` file would be more supported than other formats.
`.mp3` [(external link)](https://caniuse.com/#feat=mp3) seems to be the most promising.

src/components/playerstats/playerstats.js
```diff
@@ -327,6 +327,28 @@ define(['events', 'globalize', 'playbackManager', 'connectionManager', 'playMeth
         return sessionStats;
     }
 
+    function getSyncplayStats() {
+        var syncStats = [];
+        var stats = syncplayManager.getStats();
+
+        syncStats.push({
+            label: globalize.translate("LabelSyncplayTimeOffset"),
+            value: stats.TimeOffset + "ms"
```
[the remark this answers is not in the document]
OancaAndrei: Because seconds are part of the SI, is localization really needed?

src/components/syncplay/groupSelectionMenu.js
```diff
@@ -0,0 +1,198 @@
...
+                title: globalize.translate('HeaderSyncplaySelectGroup'),
+                items: menuItems,
+                positionTo: button,
+                resolveOnClick: true,
+                border: true
+            };
+
+            actionsheet.show(menuOptions).then(function (id) {
+                if (id == "new-group") {
+                    apiClient.sendSyncplayCommand(sessionId, "NewGroup");
```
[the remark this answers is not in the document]
OancaAndrei: The server has that data already, it's sent by the client with `reportPlaybackStart`.

src/components/playerstats/playerstats.js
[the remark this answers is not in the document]
JustAMan: Unit is SI, but its writing is different in different languages, for example in Russian you write `мс` :)

src/components/syncplay/groupSelectionMenu.js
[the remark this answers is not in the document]
JustAMan: Hmm, I think it would be better to show all groups but warn that currently played item would be changed.

src/components/syncplay/groupSelectionMenu.js
```diff
@@ -0,0 +1,198 @@
...
+function getActivePlayerId () {
+    var info = playbackManager.getPlayerInfo();
+    return info ? info.id : null;
+}
+
+/**
+ * Used to avoid console logs about uncaught promises
+ */
+function emptyCallback () {
+    // avoid console logs about uncaught promises
```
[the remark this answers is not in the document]
JustAMan: bump

src/components/syncplay/groupSelectionMenu.js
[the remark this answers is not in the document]
JustAMan: this looks like a stuff to be put in separate PR...

src/components/syncplay/syncplayManager.js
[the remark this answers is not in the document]
JustAMan: All other places we did used lowercase... so I propose to conform to single standard. cc @MrTimscampi

src/components/syncplay/syncplayManager.js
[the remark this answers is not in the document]
JustAMan: Please add this explanation a `//FIXME` comment

src/scripts/librarymenu.js
JustAMan: Ohm, what a mess... but for a separate PR certainly.

src/scripts/librarymenu.js
JustAMan: This means we would need to refactor this whole thing later on...

src/scripts/librarymenu.js
```diff
@@ -962,7 +962,7 @@ define(["dom", "layoutManager", "inputManager", "connectionManager", "events", "
     });
     events.on(playbackManager, "playerchange", updateCastIcon);
     events.on(syncplayManager, "SyncplayEnabled", updateSyncplayIcon);
-    events.on(syncplayManager, "SyncplayError", updateSyncplayErrorIcon);
+    events.on(syncplayManager, "SyncplayProblem", updateSyncplayStatusIcon);
```
JustAMan: what's the difference between `updateSyncplayIcon` and `updateSyncplayStatusIcon`?

src/components/syncplay/groupSelectionMenu.js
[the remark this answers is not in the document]
MrTimscampi: Agreed, a lot of promises are already uncaught in the code and I've been (very) slowly adding .catch() blocks to them.

We should always either recover from a failed promise, warn the user or log the error, depending on the situation.

src/components/playerstats/playerstats.js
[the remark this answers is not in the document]
OancaAndrei: Thanks, I really thought that the writing was the same too.

src/components/syncplay/groupSelectionMenu.js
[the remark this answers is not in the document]
OancaAndrei: > Agreed, a lot of promises are already uncaught in the code and I've been (very) slowly adding .catch() blocks to them.

I got my inspiration from `playerSelectionMenu.js` and I thought that this was the way to work with `actionsheet` :sweat_smile:.

I'll check the other promises as well.

src/components/syncplay/groupSelectionMenu.js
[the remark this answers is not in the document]
OancaAndrei: > this looks like a stuff to be put in separate PR...

Definitely.

For now `PlaybackPermissionManager` is used only when trying to enable SyncPlay, otherwise all the commands sent by the server through the WebSocket are blocked.

src/components/syncplay/syncplayManager.js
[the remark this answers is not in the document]
OancaAndrei: Sure, I'll make the changes then.

src/scripts/librarymenu.js
OancaAndrei: One is called when SyncPlay is enabled/disabled (user joins or leaves a group), the other is called when the media is being synced (while SyncPlay is enabled).

I'll try to give them a better naming. Or should I join them together?

src/components/syncplay/groupSelectionMenu.js
```diff
@@ -102,15 +90,17 @@ function showNewJoinGroupSelection (button, user, apiClient) {
                         PlayingItemId: playingItemId
                     });
                 }
-            }, emptyCallback);
+            }).catch((error) => {
+                console.error(error);
```
JustAMan, line 94:
```suggestion
                console.error("Syncplay: unexpected error listing groups:", error);
```

src/components/syncplay/groupSelectionMenu.js
```diff
@@ -153,14 +142,16 @@ function showLeaveGroupSelection (button, user, apiClient) {
         if (id == "leave-group") {
             apiClient.sendSyncplayCommand(sessionId, "LeaveGroup");
         }
-    }, emptyCallback);
+    }).catch((error) => {
+        console.error(error);
```
JustAMan, line 146:
```suggestion
        console.error("Syncplay: error leaving group:", error);
```

JustAMan:
Adding `blocked` label because server part https://github.com/jellyfin/jellyfin/pull/2733 should be merged first.

JustAMan:
@OancaAndrei could you please stop rebasing a single commit? :) it forces reviewers to re-read a lot of stuff which didn't change...

src/components/playerstats/playerstats.js
```diff
@@ -332,17 +332,17 @@ define(['events', 'globalize', 'playbackManager', 'connectionManager', 'syncplay
         var stats = syncplayManager.getStats();
 
         syncStats.push({
-            label: globalize.translate("LabelSyncplayTimeOffset"),
-            value: stats.TimeOffset + "ms"
+            label: globalize.translate('LabelSyncplayTimeOffset'),
+            value: stats.TimeOffset + globalize.translate('LabelSyncplayMillisecondsUnit')
```
JustAMan, line 336: Why not just name it `MillisecondsUnit`? I mean, it's highly unlikely milliseconds in other places (not syncplay) would have different localization :)

Also where is the related localization files change?

OancaAndrei: > Also where is the related localization files change?

Should be this one in [`en-us.json`](https://github.com/OancaAndrei/jellyfin-web/commit/84a9e6e15e35d04c3a5607429895342b98c79a18#diff-c3a474b438fc441c65a39d4ce2b1357cR859).

Also, am I supposed to put all new strings in alphabetical order? Or is it going to be fixed automatically?

JustAMan: dunno about that, let's ask @dkanada

src/libraries/apiclient/apiclientcore.js
```diff
@@ -150,27 +150,51 @@ define(["events", "appStorage"], function(events, appStorage) {
             if (messageIdsReceived[messageId]) return;
             messageIdsReceived[messageId] = !0
         }
-        if ("UserDeleted" === msg.MessageType) instance._currentUser = null;
-        else if ("UserUpdated" === msg.MessageType || "UserConfigurationUpdated" === msg.MessageType) {
+        if ("UserDeleted" === msg.MessageType) {
```
MrTimscampi, line 153: All the API client stuff should be mirrored to https://github.com/jellyfin/jellyfin-apiclient-javascript, as #1007 removes all the API client files from this repository.

src/components/syncplay/timeSyncManager.js
```diff
@@ -0,0 +1,207 @@
...
+     * @returns {Date} Server time.
+     */
+    localDateToServer(local) {
+        // server - local = offset
+        return new Date(local.getTime() + this.getTimeOffset());
+    }
+}
+
+/** TimeSyncManager singleton. */
+export default new TimeSyncManager();
```
MrTimscampi, line 207: I think this would also return a new instance with each import.

src/components/syncplay/syncplayManager.js
```diff
@@ -0,0 +1,841 @@
...
+    /**
+     * Signals an error state, which disables and resets Syncplay for a new session.
+     */
+    signalError () {
+        this.disableSyncplay();
+    }
+}
+
+/** SyncplayManager singleton. */
+export default new SyncplayManager();
```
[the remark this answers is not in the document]
OancaAndrei: By exporting an instance of the class and not the class itself, I limit external code from being able to create new objects. Also, modules are already themselves singletons, as the code is executed once when imported somewhere.

I adapted a quick example to match the export used here in `SyncplayManager` and `TimeSyncManager`.

Basically, it shows how importing the same module in different files won't create new instances. If that were the case, the `secret` number would be different as it is generated with `Math.random()`.

Hope this makes sense :grin:.

[Edit es6-singletons](https://codesandbox.io/s/es6-singletons-lk7wp?fontsize=14&hidenavigation=1&theme=dark)

src/components/playerstats/playerstats.js
dkanada: It's not automatic but you don't need to, we can always fix it with some scripts in the root directory if we decide to start sorting them.

src/components/playback/playbackmanager.js
```diff
@@ -3777,6 +3782,24 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
         }
     };
 
+    PlaybackManager.prototype.setPlaybackRate = function (value, player) {
+        player = player || this._currentPlayer;
```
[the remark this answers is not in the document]
OancaAndrei: The whole file should be refactored this way :laughing:.

src/libraries/apiclient/apiclientcore.js
OancaAndrei: Done, [related PR](https://github.com/jellyfin/jellyfin-apiclient-javascript/pull/44).

src/components/playerstats/playerstats.js
OancaAndrei: > Do note that using [date-fns](https://date-fns.org/v2.12.0/docs/) for time (including durations) is preferred to using `globalize` since we're already using it elsewhere in the code and the plan is to replace every instance of dates and times with date-fns.

So, I gave a quick look at `date-dns` and, from what I could see, it doesn't seem to handle milliseconds properly. I tried the following:
```
var helperDate = datefns.addMilliseconds(new Date(0), milliseconds);
var formatted = datefns.format(helperDate, 'TT', {
    locale: dfnshelper.getLocale()
});
```
but the unit is missing.

Were you proposing `date-fns` for formatting only or for operations on dates as well?

JustAMan:
> Please use merge commits for updating instead of rebasing, since rebasing forces us to re-review the entire PR if commit hashes change.

@MrTimscampi you can mark file diffs as viewed, and github would keep them collapsed even if commit hashes changed but file content didn't.

joshuaboniface:
Now that the core server portion is in, let's get this one going. It's got the required review feedback but needs some merge conflicts fixed, if you can @OancaAndrei

OancaAndrei:
> Now that the core server portion is in, let's get this one going. It's got the required review feedback but needs some merge conflicts fixed, if you can @OancaAndrei

Sure.

Quick question, is it ok if I add myself to the `CONTRIBUTORS.md` list? (forgot to ask for server side as well)

Artiume:
Of course!

JustAMan:
200w_d

mdic:
Can't wait to try this feature :smiley:

PR #1014, opened 2020-04-02
Remove duplicate implementations of parentWithClass()

MrTimscampi:
**Changes**

I spotted some re-implementation of a pretty much identical parentWithClass() function dotted around the source.

This removes those identical blocks of code and uses the one in the dom module instead.

**Issues**


dmitrylyzo:
>  dkanada merged commit d41163d into jellyfin:master **1 minute ago**

:laughing:

PR #1016, opened 2020-04-02
Touch up ESLint config

MrTimscampi:
**Changes**

* Changes parser support to ES2020
* Explicitly enables ES2017 and ES2020 environments
* Removes globals already defined in environments
* Removes a deprecated non-standard way to detect touch
* Hides warnings in CI
* Enforces good practices on imports
* Adds some eslint-enable back into the ES6 files
* Adds warnings to Promises to encourage good practices and help to find potentially fixable areas
* Adds ESLint comments linting (For rule disabling)

**Issues**


src/scripts/browser.js
```diff
@@ -292,7 +292,7 @@ define([], function () {
     }
 
     if (typeof document !== 'undefined') {
-        if (('ontouchstart' in window) || window.DocumentTouch && document instanceof DocumentTouch) {
+        if (('ontouchstart' in window) || (navigator.MaxTouchPoints > 0)) {
```
dmitrylyzo, line 295: I think it should be `navigator.maxTouchPoints`, but didn't check.
https://developer.mozilla.org/en-US/docs/Web/API/Navigator/maxTouchPoints

src/components/cardbuilder/cardBuilder.js
```diff
@@ -1854,6 +1854,8 @@ import 'programStyles';
             }
         }
 
+/* eslint-enable indent */
+
 export default {
     getCardsHtml: getCardsHtml,
     getDefaultBackgroundClass: getDefaultBackgroundClass,
```
dmitrylyzo, line 1861: Did you forget to remove indent?
And how does ESLint react to this?

MrTimscampi: It's to re-enable indent after the badly indented section. You disable it at the start of the wrongly indented content, then re-enable it after. It's good practice when doing this.

ESLint will now complain about stray ESLint comments that disable stuff for the file without re-enabling them after, so this removes that error from the lint.

.ci/azure-pipelines.yml
```diff
@@ -95,7 +95,7 @@ jobs:
       displayName: 'Install Dependencies'
       condition: ne(variables.CACHE_RESTORED, 'true')
 
-    - script: 'yarn run lint'
+    - script: 'yarn run lint --quiet'
```
dkanada, line 98: Don't forget we're switching to npm for everything.

PR #1022, opened 2020-04-04
Fix mobile layout for itemdetails

ferferga:
Just a quick fix for an alignment problem I saw some time ago but completely forgot about it.

**Before**
before1
before

**After**
after
after2

### DIFF
--- a/src/assets/css/librarybrowser.css
+++ b/src/assets/css/librarybrowser.css
@@ -529,6 +529,10 @@
     align-items: center;
 }
 
+.parentName > a:nth-child(3) {
+    font-size: 0.7em;
+}
+
 .layout-mobile .parentName,
 .layout-mobile .itemName,
 .layout-mobile .itemMiscInfo,
--- a/src/assets/css/librarybrowser.css
+++ b/src/assets/css/librarybrowser.css
@@ -516,6 +516,7 @@
 
 .itemName {
     margin: 0.5em 0;
+    font-weight: 600;
 }
 
 .nameContainer {
--- a/src/assets/css/librarybrowser.css
+++ b/src/assets/css/librarybrowser.css
@@ -533,7 +533,6 @@
 .layout-mobile .itemName,
 .layout-mobile .itemMiscInfo,
 .layout-mobile .mainDetailButtons {
-    display: flex;
     align-items: center;
     justify-content: center;
     text-align: center;
--- a/src/assets/css/librarybrowser.css
+++ b/src/assets/css/librarybrowser.css
@@ -518,6 +518,12 @@
     margin: 0.5em 0;
 }
 
+.nameContainer {
+    display: flex;
+    flex-direction: column;
+    flex-wrap: wrap;
+}
+
 .itemMiscInfo {
     display: -webkit-box;
     display: -webkit-flex;
@@ -527,10 +533,6 @@
     -webkit-box-align: center;
     -webkit-align-items: center;
     align-items: center;
-}
-
-.parentName > a:nth-child(3) {
-    font-size: 0.7em;
 }
 
 .layout-mobile .parentName,
@@ -578,7 +580,6 @@
 
 .infoText {
     white-space: nowrap;
-    overflow: hidden;
     text-overflow: ellipsis;
     text-align: left;
 }
--- a/src/controllers/itemdetailpage.js
+++ b/src/controllers/itemdetailpage.js
@@ -336,7 +336,7 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "userSetti
 
         return html = html.join(" / ");
     }
-
+    // TODO: Clean this and the CSS rules for mobile layout. See #1022
     function renderName(item, container, isStatic, context) {
         var parentRoute;
         var parentNameHtml = [];
@@ -406,9 +406,9 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "userSetti
 
         if (parentNameHtml.length) {
             if (parentNameLast) {
-                html = '<h3 class="parentName" style="margin: .25em 0;">' + parentNameHtml.join(" - ") + "</h3>";
+                html = '<h3 class="parentName" style="margin: .25em 0;">' + parentNameHtml.join("</br>") + "</h3>";
             } else {
-                html = '<h1 class="parentName" style="margin: .1em 0 .25em;">' + parentNameHtml.join(" - ") + "</h1>";
+                html = '<h1 class="parentName" style="margin: .1em 0 .25em;">' + parentNameHtml.join("</br>") + "</h1>";
             }
         }
 
--- a/src/controllers/itemdetailpage.js
+++ b/src/controllers/itemdetailpage.js
@@ -336,7 +336,6 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "userSetti
 
         return html = html.join(" / ");
     }
-    // TODO: Clean this and the CSS rules for mobile layout. See #1022
     function renderName(item, container, isStatic, context) {
         var parentRoute;
         var parentNameHtml = [];
@@ -406,9 +405,18 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "userSetti
 
         if (parentNameHtml.length) {
             if (parentNameLast) {
-                html = '<h3 class="parentName" style="margin: .25em 0;">' + parentNameHtml.join("</br>") + "</h3>";
+                // Music
+                if (layoutManager.mobile) {
+                    html = '<h3 class="parentName" style="margin: .25em 0;">' + parentNameHtml.join("</br>") + "</h3>";
+                } else {
+                    html = '<h3 class="parentName" style="margin: .25em 0;">' + parentNameHtml.join(" - ") + "</h3>";
+                }
             } else {
-                html = '<h1 class="parentName" style="margin: .1em 0 .25em;">' + parentNameHtml.join("</br>") + "</h1>";
+                if (layoutManager.mobile) {
+                    html = '<h1 class="parentName" style="margin: .1em 0 .25em;">' + parentNameHtml.join("</br>") + "</h1>";
+                } else {
+                    html = '<h1 class="parentName" style="margin: .1em 0 .25em;">' + parentNameHtml[0] + "</h1>";
+                }
             }
         }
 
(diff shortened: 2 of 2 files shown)


ferferga:
I think maybe we should remove the dash in mobile layout?

MrTimscampi:
I would definitely remove the `-` on mobile and maybe make the season text a little smaller, to make it less important than the title of the show.

ferferga:
A bit hacky but sincerely I am lazy right now for refactoring all that huge JS. I added a comment so we can keep it tracked for further design improvements for 10.6. Far from perfect (in fact it overflows sometimes) but enough for a backport imo

Final state:
image
image
image
image

src/assets/css/librarybrowser.css
```diff
@@ -529,11 +529,14 @@
     align-items: center;
 }
 
+.parentName > a:nth-child(3) {
```
JustAMan, line 532: this feels like a dirty hack...

src/controllers/itemdetailpage.js
```diff
@@ -336,7 +336,7 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "userSetti
 
         return html = html.join(" / ");
     }
-
+    // TODO: Clean this and the CSS rules for mobile layout. See #1022
```
JustAMan, line 339:
```suggestion
    // TODO: Clean this and the CSS rules for mobile layout. See GH #1022
```
IIRC github has dpecial handling for such comments

src/assets/css/librarybrowser.css
ferferga: Any other quick & non-hacky solution without rewriting a lot of the JS? As I said before, I would like this quick for 10.5 (so later on we can focus on it deeper) but if everyone doesn't agree, I don't have any problem in doing that fine for 10.6.

src/controllers/itemdetailpage.js
ferferga: How?

src/assets/css/librarybrowser.css
MrTimscampi: See my other review comment. Separate both (only on mobile) and add some class or id to the new element if needed.

Here, you're changing the rule for all layouts, where the goal was to change it only for mobile :)

As a general rule of thumb, :nth-child is to be avoided since any new element or layout change will completely break the rule :/

src/controllers/itemdetailpage.js
```diff
@@ -406,9 +406,9 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "userSetti
 
         if (parentNameHtml.length) {
             if (parentNameLast) {
-                html = '<h3 class="parentName" style="margin: .25em 0;">' + parentNameHtml.join(" - ") + "</h3>";
+                html = '<h3 class="parentName" style="margin: .25em 0;">' + parentNameHtml.join("</br>") + "</h3>";
             } else {
-                html = '<h1 class="parentName" style="margin: .1em 0 .25em;">' + parentNameHtml.join(" - ") + "</h1>";
+                html = '<h1 class="parentName" style="margin: .1em 0 .25em;">' + parentNameHtml.join("</br>") + "</h1>";
```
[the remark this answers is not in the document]
ferferga: requiring browser and checking if browser.mobile will do the trick?

src/assets/css/librarybrowser.css
ferferga: The problem is that if we add the <br> tag, in Desktop would overflow a lot, that's why I applied it to all layouts. If I have an 100% reliable way of checking the target device in JS, I can easily remove this.

src/controllers/itemdetailpage.js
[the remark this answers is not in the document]
MrTimscampi: That's pretty much all you'd need, yeah :)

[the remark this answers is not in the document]
MrTimscampi: You can require LayoutManager and check layoutManager.mobile :)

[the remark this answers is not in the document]
MrTimscampi: For CSS, there is a .layout-mobile class that you can use before your selector.

src/controllers/itemdetailpage.js
ferferga: I'll remove this completely as in my next commit I'll handle this with layoutManager, so no hacky stuff involved anymore.

ferferga:
@MrTimscampi Was bout to push my commit when I discovered this in desktop layout:

image

MrTimscampi:
> @MrTimscampi Was about to push my commit when I discovered this in desktop layout:
> 
> image
> 
> I think that this should be adressed here as well

What if we move the season indicator to the next line (with the episoe title) ?

It'd go like:

```
Star Trek: The Next Generation
Season 7 - 25. All good things...
```

Then keep it as three lines on mobile ?

ferferga:
@MrTimscampi I like your idea. I think we can display it with the same weight and font-size as the original titles.

MrTimscampi:
@ferferga I'd make it slightly smaller, perhaps. For weight, we're a tad limited since we only ship two, but heh.

ferferga:
I'm not 100% happy with the result because the code changes add a bit more of mess to the itemdetails script, but I think it's good enough for an stable backport (and even for a longer term, if we don't feel ambitiuous enough to rewrite this for 10.6)

Here is a GIF I made:

<p align="center">
  
</p>

This is mobile:
image

I also gave myself some rights 😜, and, as I liked the titles in bold, I put them in bold everywhere, which looks really good (easy to distinguish between aired dates, rating and so on), but that specific commit can be cherry picked if you don't agree with me.

src/controllers/itemdetailpage.js
```diff
@@ -406,9 +405,18 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "userSetti
...
+                    html = '<h3 class="parentName" style="margin: .25em 0;">' + parentNameHtml.join("</br>") + "</h3>";
+                } else {
+                    html = '<h3 class="parentName" style="margin: .25em 0;">' + parentNameHtml.join(" - ") + "</h3>";
+                }
             } else {
-                html = '<h1 class="parentName" style="margin: .1em 0 .25em;">' + parentNameHtml.join("</br>") + "</h1>";
+                if (layoutManager.mobile) {
+                    html = '<h1 class="parentName" style="margin: .1em 0 .25em;">' + parentNameHtml.join("</br>") + "</h1>";
+                } else {
+                    html = '<h1 class="parentName" style="margin: .1em 0 .25em;">' + parentNameHtml[0] + "</h1>";
```
JustAMan, line 418: why use only `parentNameHtml[0]` on desktop but all (joined by `<br>`) on mobile? doesn't make much sense IMHO

src/controllers/itemdetailpage.js
```diff
@@ -418,13 +426,17 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "userSetti
         var offset = parentNameLast ? ".25em" : ".5em";
 
         if (html && !parentNameLast) {
-            html += '<h3 class="itemName infoText" style="margin: .25em 0 .5em;">' + name + '</h3>';
+            if (!layoutManager.mobile && parentNameHtml[1]) {
```
JustAMan, line 429: why use specific index (`1` in this case)?..

src/controllers/itemdetailpage.js
```diff
@@ -418,13 +426,17 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "userSetti
...
+                html += '<h3 class="itemName infoText" style="margin: .25em 0 .5em; font-weight: 600;">' + parentNameHtml[1] + ' - ' + name + '</h3>';
```
JustAMan, line 430: I'd rather we have font weights specified via css classes, not inline

MrTimscampi: I agree, avoid inline CSS in all cases. It's a mess to maintain.

src/controllers/itemdetailpage.js
ferferga: @JustAMan That's what @MrTimscampi proposed earlier: Keep Show, season and episode name in different lines for mobile, while keeping TV Show in one line and Episode and Season in the same line but in a different one.

``parentNameHTML`` array contains the html code for each of the elements that are parent of an specific one. I tested it in all kinds of elements and it can only have two entries, no more, so 0 and 1 should be always indexes inside the ranges (unless I missed something).

``0`` is the TV Show. ``1`` is the season. See the GIF I attached for reference.

And yes, probably I can remove the ``<br>``, as I added the flex container afterwards, which should handle the spacing. Let me verify that.

src/controllers/itemdetailpage.js
ferferga: @JustAMan Explained above. I check that it's not undefined there because not all elements have two parents.

src/controllers/itemdetailpage.js
ferferga: @JustAMan @MrTimscampi I tried writing the rules in the ``itemName`` and ``infoText`` classes but for some reason they weren't applied, even when using ``!important``, so I thought that it wasn't that important. I'm not sure though if that was caused by gulp not reloading the page correctly, so let me check that. But I suspect that this might need its own class or some kind of refactoring

src/controllers/itemdetailpage.js
MrTimscampi: It'd probably be nicer if parentNameHTML was divided into variables with a name related to their content. It makes it easier to read and understand what's going on.

ferferga: @JustAMan Now that I tested I know why I used ``<br>``. Flex already does the job correctly on mobile devices, but not in tablets. See iPad emulator in Firefox:

**Without**
image

**With**
image

Take in mind that the array doesn't include ``Como conocí a vuestra madre`` or ``Temporada 1``. It includes ``<a style="color:inherit;" class="button-link emby-button" is="emby-linkbutton" href="itemdetails.html?id=bb0ec83....&serverId=606da8b....">Cómo conocí a vuestra madre</a>`` and ``<a style="color:inherit;" class="button-link emby-button" is="emby-linkbutton" href="itemdetails.html?id=173ad87.....&serverId=606da8b77....">Temporada 1</a>`` which are generated before. That's why I say that this need a huge refactoring, because it's a mess :)

src/controllers/itemdetailpage.js
ferferga: @JustAMan @MrTimscampi Seems that it was indeed my browser's view that didn't reload the preview in real time. Fixed in my new commit.

src/controllers/itemdetailpage.js
MrTimscampi: > Take in mind that the array doesn't include Como conocí a vuestra madre or Temporada 1. It includes <a style="color:inherit;" class="button-link emby-button" is="emby-linkbutton" href="itemdetails.html?id=bb0ec83....&serverId=606da8b....">Cómo conocí a vuestra madre</a> and <a style="color:inherit;" class="button-link emby-button" is="emby-linkbutton" href="itemdetails.html?id=173ad87.....&serverId=606da8b77....">Temporada 1</a> instead, which are generated before. That's why I say that this need a huge refactoring, because it's a mess :)

It's still easier if it's like:
```javascript
var showNameHtml = a style="color:inherit;" class="button-link emby-button" is="emby-linkbutton" href="itemdetails.html?id=bb0ec83....&serverId=606da8b....">Cómo conocí a vuestra madre</a>;
var seasonNameHtml = <a style="color:inherit;" class="button-link emby-button" is="emby-linkbutton" href="itemdetails.html?id=173ad87.....&serverId=606da8b77....">Temporada 1</a>;
```

It's easier to understand than having to go back up to look at .append() then inferring the position in the array to know what's in there :)

ferferga: @MrTimscampi Okay, now I understand what you mean

src/controllers/itemdetailpage.js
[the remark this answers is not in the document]
dkanada: Is this PR still on hold?

PR #1024, opened 2020-04-04
Throw linting errors if using non-polyfilled APIs

MrTimscampi:
**Changes**

This adds an ESLint plugin that provides rules to detect whenever a non-polyfilled API is used and throws an error.

It uses the Browserslist targets to detect the minimal versions of the browsers, then uses caniuse to lookup feature availability.

**Issues**


src/components/input/gamepadtokey.js
```diff
@@ -401,7 +403,7 @@ require(['apphost'], function (appHost) {
     if (window.navigator && typeof window.navigator.gamepadInputEmulation === "string") {
         // We want the gamepad to provide gamepad VK keyboard events rather than moving a
         // mouse like cursor. Set to "keyboard", the gamepad will provide such keyboard events
-        // and provide input to the DOM navigator.getGamepads API.
+        // and provide input to tonFocusOrGamepadDetachhe DOM navigator.getGamepads API.
```
dkanada, line 406:
```suggestion
        // and provide input to onFocusOrGamepadDetachhe DOM navigator.getGamepads API.
```

src/components/input/keyboardnavigation.js
```diff
@@ -159,7 +159,9 @@ function attachGamepadScript(e) {
 }
 
 // No need to check for gamepads manually at load time, the eventhandler will be fired for that
-window.addEventListener("gamepadconnected", attachGamepadScript);
+if (navigator.getGamepads) { /* eslint-disable-line compat/compat */
```
dmitrylyzo, line 162: `attachGamepadScript` above can also be placed inside.

MrTimscampi: ESLint yells at me if I move the function inside (https://eslint.org/docs/rules/no-inner-declarations).

dmitrylyzo: Oh, right. `var foo = function (...)` should work. Or we could insert it directly as `addEventListener` callback.

MrTimscampi: I don't think it's really necessary to get it inside of there, it'll never get executed anyway if there is no gamepad support.

If you really feel like we should, would something like this work?

```javascript
window.addEventListener("gamepadconnected", function () {
    console.log("Gamepad connected! Attaching gamepadtokey.js script");
    window.removeEventListener("gamepadconnected", this);
    require(["components/input/gamepadtokey"]);
});
```

dmitrylyzo: > would something like this work?

Nope. `this` is window. :disappointed: I didn't thought about `removeEventListener`.

OK. Leave it as it is.

src/components/appRouter.js
```diff
@@ -268,6 +268,7 @@ define(['loading', 'globalize', 'events', 'viewManager', 'layoutManager', 'skinM
     }
 
     function getMaxBandwidth() {
+        /* eslint-disable compat/compat */
```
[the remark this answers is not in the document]
MrTimscampi: ESLint throws an error on `navigator.connection`, which is an API that allows to detect connection information (speed, type of conneciton, roaming status, etc).

I haven't found a polyfill and the feature is checked anyway, so I'm disabling the compat check for this block.

src/components/apphost.js
```diff
@@ -453,6 +453,7 @@ define(["appSettings", "browser", "events", "htmlMediaHelper", "webSettings"], f
 
     if (doc) {
         doc.addEventListener(visibilityChange, function () {
+            /* eslint-disable-next-line compat/compat */
```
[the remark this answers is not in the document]
MrTimscampi: ESLint throws an error on Document.visibilityState, which I didn't find a polyfill for.

As the feature is conditional anyway, I disabled the warning.

src/components/htmlMediaHelper.js
```diff
@@ -31,7 +31,7 @@ define(['appSettings', 'browser', 'events'], function (appSettings, browser, eve
     }
 
     function enableHlsShakaPlayer(item, mediaSource, mediaType) {
-
+        /* eslint-disable-next-line compat/compat */
```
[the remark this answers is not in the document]
MrTimscampi: window.MediaSource is not available on Chrome 27. As the feature is conditional anyway, I disabled the compat check to prevent ESLint from throwing an error.

src/components/input/gamepadtokey.js
```diff
@@ -184,7 +184,7 @@ require(['apphost'], function (appHost) {
     function allowInput() {
 
         // This would be nice but always seems to return true with electron
-        if (!isElectron && document.hidden) {
+        if (!isElectron && document.hidden) { /* eslint-disable-line compat/compat */
```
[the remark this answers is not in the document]
MrTimscampi: The Page Visibility API (document.hidden) isn't available on Chrome 27. As it's a conditional feature, I disabled the ESLint check.

src/components/input/gamepadtokey.js
```diff
@@ -373,13 +373,15 @@ require(['apphost'], function (appHost) {
     }
 
     function onFocusOrGamepadAttach(e) {
+        /* eslint-disable-next-line compat/compat */
```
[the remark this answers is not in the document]
MrTimscampi: document.hasFocus() is throwing an error for Chrome 27.

It is supported [according to CanIUse](https://caniuse.com/#feat=mdn-api_document_hasfocus).

I opened an issue here: https://github.com/amilajack/eslint-plugin-compat/issues/304

src/scripts/browser.js
```diff
@@ -292,6 +292,7 @@ define([], function () {
     }
 
     if (typeof document !== 'undefined') {
+        /* eslint-disable-next-line compat/compat */
```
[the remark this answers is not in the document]
MrTimscampi: navigator.maxTouchPoints isn't available on Chrome 27.

It's ignored because the feature is conditional.

src/components/input/gamepadtokey.js
```diff
@@ -254,7 +254,7 @@ require(['apphost'], function (appHost) {
     var inputLoopTimer;
     function runInputLoop() {
         // Get the latest gamepad state.
-        var gamepads = navigator.getGamepads();
+        var gamepads = navigator.getGamepads(); /* eslint-disable-line compat/compat */
```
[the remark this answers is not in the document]
MrTimscampi: navigator.getGamepads isn't polyfill-able. This file is loaded conditionally anyway, now, do no issue in disabling the compat lint here.

src/controllers/playback/videoosd.js
```diff
@@ -1152,7 +1152,7 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
                 case "GamepadDPadLeft":
                 case "GamepadLeftThumbstickLeft":
                     // Ignores gamepad events that are always triggered, even when not focused.
-                    if (document.hasFocus()) {
+                    if (document.hasFocus()) { /* eslint-disable-line compat/compat */
```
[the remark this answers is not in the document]
MrTimscampi: Same linting issue as the other document.hasFocus(). An issue has been filed upstream to the plugin's repo.

src/components/apphost.js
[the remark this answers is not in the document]
JustAMan: but the condition is on `doc`, not `doc.visibilityState`?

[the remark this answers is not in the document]
MrTimscampi: `document[visibilityState]` resolves to `document.visibilityState`.

I changed it to make it clearer.

src/components/apphost.js
```diff
@@ -453,7 +453,8 @@ define(["appSettings", "browser", "events", "htmlMediaHelper", "webSettings"], f
 
     if (doc) {
         doc.addEventListener(visibilityChange, function () {
-            if (document[visibilityState]) {
+            /* eslint-disable-next-line compat/compat */
+            if (document.visibilityState) {
```
dmitrylyzo, line 457: `document[visibilityState]` was use `visibilityState` variable that set above. `document.visibilityState` is supposed be polyfilled.

MrTimscampi: Hm, this entire thing seems a little screwy and convoluted.

It should work in a different way, probably. I'll try refactoring it all tonight.

MrTimscampi: Should be better now :)

src/components/apphost.js
```diff
@@ -426,35 +426,27 @@ define(["appSettings", "browser", "events", "htmlMediaHelper", "webSettings"], f
...
         }
     }
 
-    if (doc) {
-        doc.addEventListener(visibilityChange, function () {
-            /* eslint-disable-next-line compat/compat */
-            if (document.visibilityState) {
+    if (self.document) {
+        document.addEventListener(visibilityChange, function () {
+            if (visibilityState) {
```
dmitrylyzo, line 449: If `visibilityState` is set, `onAppHidden` is called in all cases. I tested by switching tabs in browser.
```suggestion
            if (document[visibilityState]) {
```
But it is strange: are all vendor-members boolean (as `document.hidden`)?

dmitrylyzo: It seems that `mozHidden`, `msHidden` and `webkitHidden` must be used.
https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API in **Example**

src/components/apphost.js
```diff
@@ -426,33 +426,26 @@ define(["appSettings", "browser", "events", "htmlMediaHelper", "webSettings"], f
         }
     };
 
-    var doc = self.document;
     var isHidden = false;
 
-    if (doc) {
-        if (void 0 !== doc.visibilityState) {
+    if (self.document) {
+        if (self.document.visibilityState !== undefined) {
```
dmitrylyzo, line 432:
```suggestion
        if (self.document.hidden !== undefined) {
```

src/components/apphost.js
```diff
@@ -426,33 +426,26 @@ define(["appSettings", "browser", "events", "htmlMediaHelper", "webSettings"], f
...
-                } else {
-                    if (void 0 !== doc.webkitHidden) {
-                        visibilityChange = "webkitvisibilitychange";
-                        visibilityState = "webkitVisibilityState";
-                    }
-                }
-            }
+        } else if (self.document.mozHidden !== undefined) {
+            visibilityChange = "mozvisibilitychange";
+            visibilityState = "mozVisibilityState";
```
dmitrylyzo, line 437:
```suggestion
            visibilityState = "mozHidden";
```

src/components/apphost.js
```diff
@@ -426,33 +426,26 @@ define(["appSettings", "browser", "events", "htmlMediaHelper", "webSettings"], f
...
+        } else if (self.document.msHidden !== undefined) {
+            visibilityChange = "msvisibilitychange";
+            visibilityState = "msVisibilityState";
```
dmitrylyzo, line 440:
```suggestion
            visibilityState = "msHidden";
```

src/components/apphost.js
```diff
@@ -426,33 +426,26 @@ define(["appSettings", "browser", "events", "htmlMediaHelper", "webSettings"], f
...
+        } else if (self.document.webkitHidden !== undefined) {
+            visibilityChange = "webkitvisibilitychange";
+            visibilityState = "webkitVisibilityState";
```
dmitrylyzo, line 443:
```suggestion
            visibilityState = "webkitHidden";
```

src/components/apphost.js
MrTimscampi: We're checking for API support here, not status, I believe?

src/components/apphost.js
```diff
@@ -426,33 +426,26 @@ define(["appSettings", "browser", "events", "htmlMediaHelper", "webSettings"], f
...
         }
     }
 
-    if (doc) {
-        doc.addEventListener(visibilityChange, function () {
+    if (self.document) {
+        document.addEventListener(visibilityChange, function () {
             if (document[visibilityState]) {
```
dmitrylyzo, line 449: ESLint considers `document.visibilityState` to be requested. :confused: 
```suggestion
            /* eslint-disable-next-line compat/compat */
            if (document[visibilityState]) {
```

src/components/apphost.js
MrTimscampi: Alright, I stumbled on the source for that part of the code on MDN, it's a copy-paste from the "Example" section here: https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API

dmitrylyzo: I think we are just choosing proper field to check visibility state.

MrTimscampi:
@dmitrylyzo I switched the whole thing to the version on MDN, since that's likely more correct than whatever was in there :)

I also removed the MS-specific case, since Edge 12+ supports it and that version is now ancient.

src/components/apphost.js
```diff
@@ -426,39 +424,27 @@ define(["appSettings", "browser", "events", "htmlMediaHelper", "webSettings"], f
...
+        visibilityChange = "visibilitychange";
+    } else if (typeof document.webkitHidden !== "undefined") {
+        hidden = "webkitHidden";
+        visibilityChange = "webkitvisibilitychange";
     }
 
-    if (doc) {
-        doc.addEventListener(visibilityChange, function () {
-            if (document[visibilityState]) {
+    if (self.document) {
```
dmitrylyzo, line 439: I wonder why `self.document` is even checked - `document` is used everywhere.

MrTimscampi: Good point. I removed it since it shouldn't matter anyway.

PR #1026, opened 2020-04-04
Babel query-string to support older browsers

dmitrylyzo:
Babel `query-string` to support older browsers (Tizen 3 ~ Chrome 47).

**Changes**
- Include `query-string` for "babeling".
- Make global babel config.

_If there is a more correct way to do this, feel free to replace this PR_

### DIFF
--- a/babel.config.json
+++ b/babel.config.json
@@ -1,0 +1,3 @@
+{
+  "presets": ["@babel/preset-env"]
+}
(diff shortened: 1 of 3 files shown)


dmitrylyzo:
It seems that all `presets` can be removed - I get same `dist` folder content with global babel config.

JustAMan:
I wonder why we don't babelize the whole `node_modules`...

MrTimscampi:
> I wonder why we don't babelize the whole `node_modules`...

I think most packages, by default, set their export to the transpiled, bundled file in their dist directory.

Setting your default export to the entry point of the source version seems to be the exeption.

Issue #913, opened 2020-03-09
All CJK characters are displayed in Traditional Chinese fonts regardless of language setting


lldsolitude:
**Describe the bug**
Some CJK characters of different varieties (Simplified Chinese, Traditional Chinese, Japanese Kanji, Korean Hanja) share the same unicode code points but actually have different forms and therefore should use different fonts. Currently `jellyfin-noto/font-faces.css` imports all five CJK-specific CSSs without taking this into account. Since `TC.css` is imported last, all CJK characters are displayed in Traditional Chinese fonts regardless of the language setting.

**Expected behavior**
When Simplified Chinese/Japanese is selected in language settings, characters should be displayed in the corresponding SC/JP font instead of TC font.

**System (please complete the following information):**
 - OS: macOS 10.15.3
 - Browser: Safari
 - Jellyfin Version: 10.5.0

MrTimscampi:
Related:  https://github.com/jellyfin/JavascriptSubtitlesOctopus/issues/7

JustAMan:
@lldsolitude can you provide an example character that looks differently in different varieties? It's hard to reproduce when someone doesn't know any CJK at all (like me) without some external help.

@MrTimscampi @dkanada can we import a `css` with correct fonts _again_ (so that some css ends up imported twice) to override `TC.css` with a correct one based on language setting?

lldsolitude:
Here are three examples:
(U+542F 启; U+63A7 控; U+7247 片)
<img width="530" alt="SC/TC difference" src="https://user-images.githubusercontent.com/11805475/78239498-55f0ea80-7510-11ea-8526-a2ac3052caa0.png">

MrTimscampi:
Here's the proposal talked about on Matrix:

* Separate the CJK fonts in their own CSS file, one per language (KO, JP, TC and SC)
* Add a display option below the language selector for "Preferred Chinese characters variation"
* By default, if the culture (language) is set to Simplified Chinese, Traditional Chinese, Japanese or Korean, that display option is set to the corresponding character variation
* If not set to one of these, we decide for the user what to show
* During bootstrap of the client, as soon as user settings are available, we load the relevant CSS file


Some more info about CJK in Unicode, for reference: https://www.unicode.org/faq/han_cjk.html

ferferga:
[...]

PR #1027, opened 2020-04-04
Quick fix for CJK character display issue

ferferga:
This is a quick fix for #913, so it can be backported quickly to 10.5 until we figure out a better solution like the one described [here](https://github.com/jellyfin/jellyfin-web/issues/913#issuecomment-607813875)

This PR requires jellyfin/jellyfin-noto#2. Labeling this as blocked until that one is merged

### DIFF
--- a/src/scripts/globalize.js
+++ b/src/scripts/globalize.js
@@ -68,6 +68,9 @@ define(['connectionManager', 'userSettings', 'events'], function (connectionMana
                 ensureTranslation(allTranslations[i], fallbackCulture);
             }
         }
+        if (culture == 'zh-tw' || culture == 'zh-hk') {
+            require(["css!jellyfin-noto/css/TC"]);
+        }
     }
 
     function ensureTranslation(translationInfo, culture) {
--- a/src/scripts/globalize.js
+++ b/src/scripts/globalize.js
@@ -57,6 +57,12 @@ define(['connectionManager', 'userSettings', 'events'], function (connectionMana
             currentDateTimeCulture = currentCulture;
         }
         ensureTranslations(currentCulture);
+        // FIXME: See GH #1027 and #913. This should be configurable and not strictly based on locale.
+        if (currentCulture == 'zh-tw' || currentCulture == 'zh-hk') {
+            require(["css!jellyfin-noto/css/TC"]);
+        } else {
+            require(["css!jellyfin-noto/css/SC"]);
+        }
     }
 
     function ensureTranslations(culture) {
@@ -67,9 +73,6 @@ define(['connectionManager', 'userSettings', 'events'], function (connectionMana
             for (var i in allTranslations) {
                 ensureTranslation(allTranslations[i], fallbackCulture);
             }
-        }
-        if (culture == 'zh-tw' || culture == 'zh-hk') {
-            require(["css!jellyfin-noto/css/TC"]);
         }
     }
 
--- a/yarn.lock
+++ b/yarn.lock
@@ -6448,8 +6448,8 @@ isurl@^1.0.0-alpha5:
     is-object "^1.0.1"
 
 "jellyfin-noto@https://github.com/jellyfin/jellyfin-noto":
-  version "1.0.2"
-  resolved "https://github.com/jellyfin/jellyfin-noto#a441b179c833288fc372cadb408d32a76c5479f1"
+  version "1.0.3"
+  resolved "https://github.com/jellyfin/jellyfin-noto#b784602db063734c721a46563ae5d6577ec2b35d"
 
 jquery@>=1.9.1, jquery@^3.4.1:
   version "3.4.1"


yarn.lock
```diff
@@ -6447,9 +6447,9 @@ isurl@^1.0.0-alpha5:
     has-to-string-tag-x "^1.2.0"
     is-object "^1.0.1"
 
-"jellyfin-noto@https://github.com/jellyfin/jellyfin-noto":
-  version "1.0.2"
-  resolved "https://github.com/jellyfin/jellyfin-noto#a441b179c833288fc372cadb408d32a76c5479f1"
+"jellyfin-noto@https://github.com/ferferga/jellyfin-noto":
+  version "1.0.3"
+  resolved "https://github.com/ferferga/jellyfin-noto#3f4b90ccaff4fcbe2d68557a2b6d5f40028df8e8"
```
dmitrylyzo, line 6452: Why is your fork here?

ferferga: I thought I discarded that change, sorry 🤦‍♂️

ferferga:
If #1029 is backported as well, we can make the require statement in ``site.js``, which is probably cleaner than doing it in ``globalize.js``

dmitrylyzo:
If do this in `globalize`, I would load fonts in `updateCurrentCulture` (directly or add some extra function `loadFonts(culture)`).

src/scripts/globalize.js
```diff
@@ -68,6 +68,9 @@ define(['connectionManager', 'userSettings', 'events'], function (connectionMana
                 ensureTranslation(allTranslations[i], fallbackCulture);
             }
         }
+        if (culture == 'zh-tw' || culture == 'zh-hk') {
+            require(["css!jellyfin-noto/css/TC"]);
+        }
```
JustAMan, line 73: What about other CJK variants? At least SC?

ferferga: Loaded by default

ferferga:
@dmitrylyzo I was thinking in doing it in site.js instead (which would require #1029), but I didn't thought about the other way around (changing lang from traditional to simplified), so maybe it's a better idea to leave it in ``globalize.js``. I think that with my last commit, pretty much all the use-cases are covered.

src/scripts/globalize.js
MrTimscampi: I think Japanese should be loaded by default (Because of Hiragana and Katakana). Korean as well, since the Korean font includes only Hangul.

Both variants of Chinese should be loaded only on their respective cultures for, as we're not really sure what the usages are compared to each other and to Japanese.

ferferga: @MrTimscampi That's the current situation. All of them are loaded by jellyfin-noto package (also TC before, I thought it used other ranges, so I imported all of them at once). Now TC is not imported (only when required) and SC is used as a fallback for all the other languages (so users with other languages will see characters in SC by default).

The behaviour of the other charsets are unchanged, so everything it's still the same in that matter.

src/scripts/globalize.js
```diff
@@ -57,6 +57,11 @@ define(['connectionManager', 'userSettings', 'events'], function (connectionMana
             currentDateTimeCulture = currentCulture;
         }
         ensureTranslations(currentCulture);
+        if (currentCulture == 'zh-tw' || currentCulture == 'zh-hk') {
```
JustAMan, line 60: at least put in a `// FIXME` comment please

src/scripts/globalize.js
MrTimscampi: No, what I mean is that Japanese and TC/SC share characters, but with different variations :)

So, without being sure, we should load the Chinese fonts only if explicitly required by the user. We don't know if Simplified Chinese replaces characters from Japanese with versions that don't look native to Japanese people. But as Japanese has characters that Chinese doesn't have (Hiragana and Katakana), it should be loaded by default (like now).

Neither Simplified Chinese nor Traditional Chinese should be loaded by default. These should be applied only if the user has selected their respective languages.

ferferga:
@JustAMan Done! Shall I squash some of the commits? 3 for such a simple fix seem a lot to me lol.

JustAMan:
Yeah, go squash if you want :)

ferferga:
This is now ready to be merged

src/scripts/globalize.js
```diff
@@ -57,6 +57,12 @@ define(['connectionManager', 'userSettings', 'events'], function (connectionMana
             currentDateTimeCulture = currentCulture;
         }
         ensureTranslations(currentCulture);
+        // FIXME: See GH #1027 and #913. This should be configurable and not strictly based on locale.
+        if (currentCulture == 'zh-tw' || currentCulture == 'zh-hk') {
+            require(["css!jellyfin-noto/css/TC"]);
+        } else {
+            require(["css!jellyfin-noto/css/SC"]);
```
MrTimscampi, line 64: I still have issues with it loading Simplified Chinese by default over Japanese, since they also share characters and we're not sure if characters are different or not.

As-is, this fixes Traditional Chinese and Simplified Chinese, but we're not sure if it'd still screw up Japanese text or not.

I maintain that, in my opinion, Japanese should be the only script of this kind loaded by default (Because of Hiragana and Katakana, that aren't shared with any other script) and either variant of Chinese should be loaded only in their respective case.

MrTimscampi: @jellyfin/web This is, of course, open to discussion if anyone can reliably confirm that all shared characters of Simplified Chinese and Japanese are exactly identical.

frankyifei: There are some difference between SC and JP. bt most people from east asian can read all the different writings. 
for example 
<span lang= ja>门</span> and <span lang=zh-cn>门</span> are the same code point. you can copy and paste to see they are the same.

MrTimscampi: Thanks for the info :)

Then I'm positive we want to give people the best experience, so loading their relevant language is the best approach. We ideally want everyone to be a first-class citizen and, while they certainly are able to read variations, it's not much work to make sure it uses the proper ones, so might as well do it :)

JustAMan: This means we'll have to make [jf-noto](https://github.com/jellyfin/jellyfin-noto) to _not_ include all CJK by default as well.

ferferga: This can be done with an additional check for Japanese. What's going to happen with those with other languages though?

JustAMan: I think non-CJK languages don't have conflicting code points, so they're handled by the browser just fine.

ferferga: @JustAMan But I have items with Japanese text and my UI is in Spanish. How we should handle that?

JustAMan: We should pick one variant as the default probably and make the logic do like
```
switch (locale) {
    case tradidional chinese:
          load TC.css
    case simplified chinese:
          load SC.css
// etc.
    default (no matching locale)
         load someDefault.css
```

frankyifei: each CJK country select their own version of the same character. but they are the same character. most people who can read Han characters would understand theses variants. Noto CJK fonts also include latin fonts so displaying Spanish and Jp at the same time is not an issue. 
Html attribute "lang" and CSS selector lang() could be used to display these variant correctly. I did that with  <span lang=ja>门</span> and <span lang=zh-cn>门</span>. 
my own media library includes SC, TC, and JP contents. I can read most of the variants so it has never been an issue. But to make things perfect, there should be an option for users to specify the display language of the contents. Maybe this can be done with custom css file.

MrTimscampi: What we thought about as a more permanent solution (in a further PR for 10.6) is to add a Display option to choose between Japanese, Traditional Chinese, and Simplified Chinese.

The default would be set to Japanese, which would then always be loaded (Due to hiragana and katakana) but would set itself automatically to the relevant option when a Chinese variant is selected. The user would also be free to override the setting to whatever they please afterward.

This seems to give the best support for multi-lingual libraries, while respecting local expectations for the look of various characters.

frankyifei: Both hiragana and katakana should be included in any variants of noto sans CJK as the same. I checked windows Character Map on NotoSansCJKsc OTF. Hiragana and katakana are from U+3042 to U+30FF. But I have no idea how jellyfin-noto was subsetted.  The differences of NotoCJKsc, tc, jp are their Han characters.

ferferga: @frankyifei You can see how they are subsetted [here](https://github.com/jellyfin/jellyfin-noto/blob/master/font-faces.css).

As you can see, it's a replica of Google Fonts: we set in CSS a reference to each of the ranges a woff2 file support and the browser loads only the woff files that contains those characters, so we save bandwidth.

In [these lines](https://github.com/jellyfin/jellyfin-noto/blob/b784602db063734c721a46563ae5d6577ec2b35d/font-faces.css#L194-L213) you can see where the CJK fonts are loaded. [Before](https://github.com/jellyfin/jellyfin-noto/blob/9baaa18d15e644e682a9fc0fec9a5e81b3eb3caa/font-faces.css), all of them (JP, HK, KR, TC, SC) were loaded. Now, TC was removed and it's loaded by jellyfin-web once the client is set to traditional Chinese. So, by default, SC, HK, KR and JP are loaded, so users who have foreign languages but have items in CJK characters are covered.

frankyifei: I looked the 5 .css files roughly.
Everyone of them includes hiragana and katakana.
Only KR includes hangual, other don't. 
Except KR, other .css share quite large amount of code points.

But it does not really matter because I just found out most of the text such as movie titles and descriptions are displayed in chrome are rendered with local fonts. Only a few of the ui text use webfonts.

ferferga: @frankyifei This is happening in Chrome Windows/Linux/Mac v80? Any sample text where the font isn't applied? Can you attach an screenshot?

frankyifei: I guess it should work. but HK is not necessary.

frankyifei: firefox, chrome on windows. I dont know why is this happenning.  Look the font style is different when I set local font to serif Noto CJK. 
Screenshot_2020-04-10 Jellyfin(1)

ferferga:
I'm going to close this because, although this issue makes CJK users not first class citizens because not all chars are the most appropiated for their languages, I learnt along the way that it's not such a big deal because the UI is still understandable for CJK users - I thought at first that it wasn't understandable at all. I didn't know anything about CJK, so I suppose those are the things that happens when you start learning :)

**We will directly go with a definitive solution for 10.6, instead of this hacky one**, as the one described in #913: [See the reply here](https://github.com/jellyfin/jellyfin-web/issues/913#issuecomment-607813875) 

Any feedback about this should go to that issue.

PR #1029, opened 2020-04-04
Globalize shouldn't require ConnectionManager

ferferga:
Original commit made in #1004


dmitrylyzo:
Export `updateCurrentCulture` by placing it in `return` statement.

ferferga:
@dmitrylyzo Shall I add it to package.json as well?

dmitrylyzo:
> @dmitrylyzo Shall I add it to package.json as well?

`globalize`? no. Just add `updateCurrentCulture: updateCurrentCulture` at the end of `return {...}`

PR #1034, opened 2020-04-05
Set autocaptialize to 'off' on the login username form

rexbron:
Set autocaptialize to 'off' on the login username form to improve mobile UX

**Changes**
Set the autocapitalize switch to off on the login form as user names are case sensitive and mobile keyboards default to autocapitalizing the start of sentences.

### DIFF
--- a/src/assets/css/librarybrowser.css
+++ b/src/assets/css/librarybrowser.css
@@ -314,7 +314,6 @@
 
     .dashboardDocument .mainDrawer-scrollContainer {
         margin-top: 4.6em !important;
-
     }
 }
 
--- a/src/assets/css/librarybrowser.css
+++ b/src/assets/css/librarybrowser.css
@@ -242,7 +242,7 @@
 }
 
 .mainDrawer-scrollContainer {
-    padding-bottom: 10vh;
+    margin-bottom: 10vh;
 }
 
 @media all and (min-width: 40em) {
@@ -314,6 +314,7 @@
 
     .dashboardDocument .mainDrawer-scrollContainer {
         margin-top: 4.6em !important;
+
     }
 }
 
--- a/src/login.html
+++ b/src/login.html
@@ -8,7 +8,7 @@
             </div>
 
             <div class="inputContainer">
-                <input is="emby-input" type="text" id="txtManualName" required="required" label="${LabelUser}" autocomplete="username" />
+                <input is="emby-input" type="text" id="txtManualName" required="required" label="${LabelUser}" autocomplete="username" autocapitalize="off" />
             </div>
 
             <div class="inputContainer">


src/login.html
```diff
@@ -8,7 +8,7 @@ <h1 style="margin-top:1em;text-align: left;">${HeaderPleaseSignIn}</h1>
             </div>
 
             <div class="inputContainer">
-                <input is="emby-input" type="text" id="txtManualName" required="required" label="${LabelUser}" autocomplete="username" />
+                <input is="emby-input" type="text" id="txtManualName" required="required" label="${LabelUser}" autocomplete="username" autocapitalize="off" />
```
ferferga, line 11: Perhaps we can also make this non autocorrectable as well?

```suggestion
                <input is="emby-input" type="text" id="txtManualName" required="required" label="${LabelUser}" autocomplete="username" autocapitalize="off" autocorrect="off" />
```

MrTimscampi: We could add `autocapitalize="off"` to the password field right after, as well.

rexbron: As the password field is tagged as such, Android at lease defaults to no autocaptialization. Probably won't hurt but also not required I think.

ferferga: I tested and it seems that, with the ``autocomplete`` label set to username, Chrome doesn't autocorrect the field. It seems that ``autocomplete`` is the standard label to rule them all, so no specific labels for each situation (aside from autocapitalize). This means that passwords are not set to capitalize by default (as @MrTimscampi suggested to do as well) because it's implicit in the autocomplete tag. Usernames won't be autocorrected either because of that tag.

I approve the original commit. Thanks for your contribution! @rexbron

src/assets/css/librarybrowser.css
```diff
@@ -242,7 +242,7 @@
 }
 
 .mainDrawer-scrollContainer {
-    padding-bottom: 10vh;
+    margin-bottom: 10vh;
```
MrTimscampi, line 245: Why ?

MrTimscampi: Nevermind, saw the issue and the comment.

rexbron: I'm a git hub noob, was fixing a different issue now and was trying to make a new PR. 
Not sure how to split it into a different request. See bug 1035.

MrTimscampi: No worries :) It's so small that we can roll it into this one.

Your issue here is that you made your original PR with the `master` branch. That's the main branch of your fork, from which you should base all your future PRs.

Once this PR is merged, you can do the following to reset your branch to the state of the `master` on this repository:

```bash
git remote add upstream https://github.com/jellyfin/jellyfin-web.git
git fetch upstream
git reset --hard upstream/master
git push origin HEAD --force-with-lease
```

This will reset your repository's master branch both locally and on Github.

Then whenever you want to make some new development, do the following to create a new branch:

```bash
git checkout master
git checkout -b <your new branch>
```

To update your master branch periodically, you can do the following;

```bash
git checkout master
git fetch upstream
git merge upstream/master --ff-only
git push origin HEAD
```

Do wait before this is merged though. Otherwise, we will lose the code in this PR.

PR #1039, opened 2020-04-06
Sonarqube bug fixin' - Part 1: Phantom Bugs

MrTimscampi:
**Changes**

Going through Sonarqube items to fix them.

**Issues**


src/components/appRouter.js
```diff
@@ -577,7 +577,8 @@ define(['loading', 'globalize', 'events', 'viewManager', 'layoutManager', 'skinM
 
     function showDirect(path) {
         return new Promise(function(resolve, reject) {
-            resolveOnNextShow = resolve, page.show(baseUrl()+path);
+            resolveOnNextShow = resolve;
+            resolve = page.show(baseUrl() + path);
```
thornbill, line 581: This doesn’t seem to match the function of the original.

MrTimscampi: It's what I found through research :/ I honestly never use comma operators because I find them insanely dirty, so I'm not sure how to translate this to "proper" code.

What would you use ?

MrTimscampi: a1132e9 should be better.

It's a damn confusing operator.

thornbill: It is really weird. I’m sure that’s a holdover from minification. I’ve never seen anyone actually use it in practice lol

MrTimscampi: There's a few big these dotted around the source, so I'd bet on leftover minification as well.

src/scripts/dfnshelper.js
```diff
@@ -1,99 +1,56 @@
...
-        case 'zh-cn':
-            return zhCN;
-        case 'zh-hk':
-            return zhCN;
-        case 'zh-tw':
-            return zhTW;
-        default:
-            return enUS;
-    }
+    return dateLocales(globalize.getCurrentLocale() || enUS);
```
dmitrylyzo, line 53: _visually_
```suggestion
    return dateLocales(globalize.getCurrentLocale()) || enUS;
```

dmitrylyzo: Btw, does `globalize.getCurrentDateTimeLocale` make sense now?

src/scripts/inputManager.js
```diff
@@ -84,168 +89,77 @@ define(['playbackManager', 'focusManager', 'appRouter', 'dom', 'apphost'], funct
...
-            case 'settings':
-                appRouter.showSettings();
-                break;
-            case 'back':
+        const keyActions = (command) => ({
+            'up': focusManager.moveUp(sourceElement),
+            'down': focusManager.moveDown(sourceElement),
+            'left': focusManager.moveLeft(sourceElement),
+            'right': focusManager.moveRight(sourceElement),
+            'home': appRouter.goHome(),
```
dmitrylyzo, line 97: This approach calls all of these functions.
Maybe something like `bind` will save us, but I was happy with `switch`. :p

MrTimscampi: I'll investigate the function calls :)

Visually, I find this nicer. It also seems like this is more optimized in general, especially in cases where there are a lot of cases in the switch statement.

See here for more details: https://sonarcloud.io/organizations/mrtimscampi/rules?open=javascript%3AS1479&rule_key=javascript%3AS1479

src/scripts/dfnshelper.js
MrTimscampi: All the dates aren't converted yet (in fact, the only thing using date-fns currently is the activity log) so it still has uses :)

src/scripts/inputManager.js
MrTimscampi: Fixed the potential issues by wrapping them in functions and adding `.call()` after the result is returned from the table.

It now works as expected. I also added error handling by sending a message to the debug channel when an unhandled event is sent.

src/scripts/inputManager.js
```diff
@@ -84,168 +89,163 @@ define(['playbackManager', 'focusManager', 'appRouter', 'dom', 'apphost'], funct
...
-                break;
-            default:
-                break;
+            }
+        })[command];
+
+        try {
+            keyActions(commandName).call();
+        } catch (error) {
+            console.debug(`inputManager: tried to process command with no action assigned: ${commandName}`);
```
dmitrylyzo, line 233: Could it be that we have caught the exception that was thrown in the call and no one will know about it (even in browser console)?

I am not sure (don't know) about performance side of `switch` vs generating such object.

MrTimscampi: I added that part after testing, when noticing that an "unknown" event is sent to the inputManager when focusing the window. The input manager works properly.

According to what I read, lookup tables are a lot faster on older engines and around the same nowadays. Very large switch statements (30+ items) are apparently significantly faster due to most interpreters assigning them to two arrays instead of one.

dmitrylyzo: > I added that part after testing, when noticing that an "unknown" event is sent to the inputManager when focusing the window. The input manager works properly.

I mean, if you throw an exception in `focusManager.moveUp` (or deeper) and don't catch it there. Will it be at least in console?

src/components/slideshow/slideshow.js
```diff
@@ -307,8 +307,7 @@ define(['dialogHelper', 'inputManager', 'connectionManager', 'layoutManager', 'f
          */
         function getSwiperSlideHtmlFromItem(item) {
             return getSwiperSlideHtmlFromSlide({
-                imageUrl: getImgUrl(item),
-                originalImage: getImgUrl(item, true),
+                originalImage: getImgUrl(item),
```
JustAMan, line 310: you sure?

src/controllers/auth/selectserver.js
```diff
@@ -95,7 +95,7 @@ define(["loading", "appRouter", "layoutManager", "appSettings", "apphost", "focu
     }
 
     function showServerConnectionFailure() {
-        alertText(globalize.translate("MessageUnableToConnectToServer"), globalize.translate("HeaderConnectionFailure"));
+        alertText(globalize.translate("MessageUnableToConnectToServer"));
```
JustAMan, line 98: you sure?

src/legacy/dashboard.js
```diff
@@ -3,16 +3,20 @@ Dashboard.confirm = function(message, title, callback) {
     require(["confirm"], function(confirm) {
         confirm(message, title).then(function() {
             callback(!0);
-        }, function() {
+        }).catch(function() {
```
JustAMan, line 6: lol, good find

src/scripts/dfnshelper.js
```diff
@@ -1,99 +1,59 @@
...
-            return zhCN;
-        case 'zh-hk':
-            return zhCN;
-        case 'zh-tw':
-            return zhTW;
-        default:
-            return enUS;
-    }
+    console.warn(globalize.getCurrentLocale());
+    console.warn(dateLocales(globalize.getCurrentLocale()));
```
JustAMan, line 55: wtf? 😛

src/scripts/inputManager.js
JustAMan: I support, better to check for `keyActions(commandName)` being `undefined` explicitly and _not_ swallow the error within the handler silently

src/controllers/auth/selectserver.js
MrTimscampi: The `alertText()` defined in this module only has one parameter.

src/scripts/dfnshelper.js
MrTimscampi: I need to stop leaving these everywhere lol

src/components/slideshow/slideshow.js
MrTimscampi: imageUrl isn't used in the file anymore since we don't use the resized image and only use the originalImage.

src/scripts/inputManager.js
```diff
@@ -61,191 +66,186 @@ define(['playbackManager', 'focusManager', 'appRouter', 'dom', 'apphost'], funct
...
+            },
+            'repeatone': () => {
                 playbackManager.setRepeatMode('RepeatOne');
-                break;
-            default:
-                break;
+            }
+        })[command];
+
+        if (keyActions(commandName) === undefined) {
```
dmitrylyzo, line 230:
```suggestion
        const action = keyActions(commandName);

        if (action === undefined) {
```

src/scripts/inputManager.js
```diff
@@ -61,191 +66,186 @@ define(['playbackManager', 'focusManager', 'appRouter', 'dom', 'apphost'], funct
...
+            console.debug(`inputManager: tried to process command with no action assigned: ${commandName}`);
+        } else {
+            keyActions(commandName).call();
```
dmitrylyzo, line 233:
```suggestion
            action.call();
```

src/scripts/inputManager.js
```diff
@@ -1,44 +1,49 @@
...
     }
 
-    var commandTimes = {};
+    let commandTimes = {};
 
     function checkCommandTime(command) {
 
-        var last = commandTimes[command] || 0;
-        var now = new Date().getTime();
+        let last = commandTimes[command] || 0;
```
dmitrylyzo, line 45:
```suggestion
        const last = commandTimes[command] || 0;
```

Issue #899, opened 2020-03-07
Chrome Mobile / Android Resume not working on non-Direct Play


Artiume:
Behavior observed on mobile browsers and android. You can only start a partially played episode from the beginning, there is no resume option. Related note, on home page, you can only resume an episode using the play button. 

![Screenshot_20200307-050341_Jellyfin](https://user-images.githubusercontent.com/20715731/76141984-0e715e80-6037-11ea-82c8-803177648093.jpg)

eltiel:
Having this problem as well.

JustAMan:
Apparently this only happens for transcoded items, direct play items should be fine.
Right @Artiume ?

Artiume:
Yup, direct play resumed perfectly

![Screenshot_20200401-172106_Chrome](https://user-images.githubusercontent.com/20715731/78187889-42c11a80-743d-11ea-80ac-e1eb20fe75f7.jpg)

bnichs5:
When can we expect to see the update. Just tried with the newest nightly version. Still having the same problems.

JustAMan:
@bnichs5 this was released as part of 10.5.4 (only as web part), Android client is yet to be released.
See https://github.com/jellyfin/jellyfin-android/pull/309

If you're seeing seek issues in mobile Chrome on 10.5.4 please open a new ticket.

Issue #815, opened 2020-02-16
Android app v0.9.8 media resume not working with transcoding


4luk:
Just updated to the latest Android mobile version 0.9.8 and now any time I transcode a movie it will start from the start. With no transcoding the movie will resume as normal.
Uninstall and reinstalled didn't help
Android tv works fine

JustAMan:
@dkanada this feels specific to Android, not jf-web, any reasons you moved it here?

Artiume:
I just attempted to replicate the issue on my android. I did not have issues so either the fix is in nightly or in my PR. I can try vanilla nightly to see if it's my PR

Artiume:
So I was unable to produce the issue using a nightly server.

dkanada:
I've seen this with a nightly server and the latest Android client, but the resume code isn't handled by Android so I doubt the issue is in that repository.

Artiume:
@dkanada i tried it with a vanilla nightly and it seemed fixed.

dkanada:
@Artiume can you give it a try once the new Android build gets pushed to GitHub? If resume and subtitles are both working we will release it on Google Play. Can you think of any other regressions mentioned in the last few days?

Artiume:
Hevc playback is broken but I think that issue was already present

PR #1040, opened 2020-04-07
Fix native hls resume

JustAMan:
**Changes**

Remove the hack that was used to resume HLS from previous position on Safari (actually it was used to resume everything, even direct plays, so Safari should have played 2.5 seconds of the beginning always) and implement a proper event-driven way to seek the stream only when it's really ready to be sought (should usually be faster than 2.5 seconds, but in reality depends on the bandwidth and the server).

**Issues**

Fixes the broken resume of non-direct play on mobile Chrome and Android app:
* Fixes https://github.com/jellyfin/jellyfin-web/issues/899 (only the resuming part; in current design you're actually expected to _resume_ from mobile, there's no option to start from the beginning, it's starting from start prior to my fix due to the bug I'm fixing).
* Fixes https://github.com/jellyfin/jellyfin-web/issues/815

### DIFF
--- a/src/components/htmlMediaHelper.js
+++ b/src/components/htmlMediaHelper.js
@@ -171,13 +171,29 @@ define(['appSettings', 'browser', 'events'], function (appSettings, browser, eve
 
             // Appending #t=xxx to the query string doesn't seem to work with HLS
             // For plain video files, not all browsers support it either
-            var delay = browser.safari ? 2500 : 0;
-            if (delay) {
-                setTimeout(function () {
-                    setCurrentTimeIfNeeded(element, seconds);
-                }, delay);
-            } else {
+
+            if (element.duration >= seconds) {
+                // media is ready, seek immediately
                 setCurrentTimeIfNeeded(element, seconds);
+            } else {
+                // update video player position when media is ready to be sought
+                var events = ["durationchange", "loadeddata", "play", "loadedmetadata"];
+                var onMediaChange = function(e) {
+                    if (element.currentTime === 0 && element.duration >= seconds) {
+                        // seek only when video position is exactly zero,
+                        // as this is true only if video hasn't started yet or
+                        // user rewound to the very beginning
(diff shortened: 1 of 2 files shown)


Artiume:
Addressing the missing UX button of #899 is non-issue. I can make a new ticket to address that specific issue. Having proper resume has been the focus of #899 at this point.

Artiume:
What else should we look into fixing for an android release?

dkanada:
We just need to add the GitHub and webhook releases to the CI for a new Android release.

src/components/htmlMediaHelper.js
```diff
@@ -171,13 +171,29 @@ define(['appSettings', 'browser', 'events'], function (appSettings, browser, eve
...
+            } else {
+                // update video player position when media is ready to be sought
+                var events = ["durationchange", "loadeddata", "play", "loadedmetadata"];
+                var onMediaChange = function(e) {
+                    if (element.currentTime === 0 && element.duration >= seconds) {
+                        // seek only when video position is exactly zero,
+                        // as this is true only if video hasn't started yet or
+                        // user rewound to the very beginning
+                        // (but rewinding cannot happen as the first event with media of non-empty duration)
+                        console.log("seeking to " + seconds + " on " + e.type);
```
MrTimscampi, line 187:
```suggestion
                        console.log(`seeking to ${seconds} on ${e.type}`);
```

dmitrylyzo: How will old browsers react to this? File won't be transpiled in stable.

JustAMan: Fixed (and dropped logging level to `.debug()`).

JustAMan:
@Artiume 
> What else should we look into fixing for an android release?

IMHO this https://github.com/jellyfin/JavascriptSubtitlesOctopus/pull/14 plus new Octopus release plus bumped Octopus release in jf-web so Android gets client-side rendering of SSA back.

PR #1050, opened 2020-04-09
Fix parameters for POST to backend to set language in setup wizard

ZadenRB:
**Changes**
Modifies parameters sent to backend from wizard to be camel-case, necessary to properly set language(metadata & display) in startup wizard.

**Issues**
Fixes [jellyfin/jellyfin#2812](https://github.com/jellyfin/jellyfin/issues/2812)

### DIFF
--- a/src/controllers/wizard/settings.js
+++ b/src/controllers/wizard/settings.js
@@ -5,8 +5,8 @@ define(["loading", "emby-checkbox", "emby-button", "emby-select"], function (loa
         loading.show();
         var apiClient = ApiClient;
         apiClient.getJSON(apiClient.getUrl("Startup/Configuration")).then(function (config) {
-            config.PreferredMetadataLanguage = page.querySelector("#selectLanguage").value;
-            config.MetadataCountryCode = page.querySelector("#selectCountry").value;
+            config.preferredMetadataLanguage = page.querySelector("#selectLanguage").value;
+            config.metadataCountryCode = page.querySelector("#selectCountry").value;
             apiClient.ajax({
                 type: "POST",
                 data: config,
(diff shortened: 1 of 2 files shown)


dkanada:
@cvium did you intentionally make the new parameters camel case instead of title case? Just want to clarify the direction for the new endpoints.

ZadenRB:
@JustAMan I looked at fixing it on the backend (honestly I think it would be ideal), but the routes for the startup wizard are defined differently than all the other routes, which I assumed was for a reason. If those are fine to be rewritten I’m happy to push the fix there instead.

dkanada:
It's because the wizard is the first step in our goal to move all API endpoints to a new framework, but the casing change might have been accidental.

dkanada:
Alright, we decided to go with TitleCasing so if you want to open a PR on the server we can get this fix merged for the 10.5.4 release that will hopefully be done today.

ZadenRB:
Working on the backend fix now. It may take a bit, because the framework itself is converting the fields from title case to camel case, so I'll need to figure that out.

dkanada:
Hmmm that's not ideal. We could always merge this for now and then implement a proper fix later, there's no rush to get things done.

ZadenRB:
I think it may be that the route currently just returns a C# object, so I'll test with converting it to JSON first, but if that doesn't work, I'm happy to merge this and take my time figuring out the fix if need be.

ZadenRB:
@dkanada I would go ahead and merge this, but I should have a PR in the next couple days correctly setting the JSON options to not camel case everything.

dkanada:
We missed the merge window by a smidge, apologies.

ZadenRB:
No problem, I figured out the correct backend option anyways, so I'll close this and open a PR for the server-side fix.

PR #1065, opened 2020-04-11
Add unloading capability to the image loader

MrTimscampi:
**Changes**

Removes the scroll lazy loader (since IntersectionObserver is polyfilled), changes the lazy loader to pass the entry instead of the element and adds unloading to the image loader.

**Issues**

Should improve any low-memory device (Smart TV, mobile), contribute to improving #858 and improve responsiveness when the library limit is set to 0.


src/components/images/imageLoader.js
```diff
@@ -1,50 +1,90 @@
...
+            preloaderImg = null;
         });
     }
 
-    function fadeIn(elem) {
+    function emptyImageElement(elem) {
+        var url;
+
+        if (elem.tagName !== "IMG") {
+            url = elem.style.backgroundImage.slice(4, -1).replace(/"/g, "");
```
JustAMan, line 67: what this magic does?

src/components/images/style.css
```diff
@@ -1,26 +1,15 @@
...
-    -moz-animation-timing-function: cubic-bezier(0, 0, 0.5, 1);
-    -o-animation-timing-function: cubic-bezier(0, 0, 0.5, 1);
-    animation-timing-function: cubic-bezier(0, 0, 0.5, 1);
 }
 
-.lazy-image-fadein-fast {
-    animation: lazy-image-fadein 160ms ease-in normal both;
+.cardImageContainer.lazy.lazy-image-fadein {
+    opacity: 1;
+    transition: opacity 1s;
```
JustAMan, line 7: I personally would make "slow" one around 0.7s and fast one around 0.2, but that's my personal pref...

src/components/lazyloader/lazyloader-intersectionobserver.js
```diff
@@ -1,103 +1,67 @@
...
-        if (!observer) {
-            this.createObserver();
-            observer = this.observer;
+            Array.from(elements).forEach(element => {
+                observer.observe(element);
+            });
         }
 
-        this.elementCount = (this.elementCount || 0) + elements.length;
+        destroyObserver(elements) {
```
JustAMan, line 34: if it doesn't do anything with `elements` why does it accept those?

src/components/lazyloader/lazyloader-intersectionobserver.js
```diff
@@ -1,103 +1,67 @@
...
-    };
-
-    LazyLoader.prototype.destroyObserver = function (elements) {
-
-        var observer = this.observer;
 
-        if (observer) {
-            observer.disconnect();
-            this.observer = null;
+        destroy(elements) {
```
JustAMan, line 43: same

src/scripts/site.js
```diff
@@ -397,10 +392,6 @@ var AppInfo = {};
             define("registerElement", ["document-register-element"], returnFirstDependency);
         }
 
-        define("imageFetcher", [componentsPath + "/images/imageFetcher"], returnFirstDependency);
-
-        var preferNativeAlerts = browser.tv;
```
JustAMan, line 402: why is this gone?

src/components/images/imageLoader.js
MrTimscampi: It extracts the url from the value of background-image, which is like `url('something/somewhere.jpg')`.

src/scripts/site.js
MrTimscampi: imageFetcher was removed, since it was pretty much integrated into imageLoader.

preferNativeAlerts was a leftover from another PR that moved the castSender, dialog and confirm modules out of site.js.

src/components/images/imageLoader.js
dmitrylyzo: Regular expression for double-quotes. Maybe `/['"]/g`

JustAMan: @MrTimscampi Could you fix single-quoted URL support here? Also are you sure there's no native way aside of regex-ing it away?

MrTimscampi: There is no need to handle them since the JavaScript API will always return the browser-parsed version of the property, which will always use double quotes.

dkanada:
@MrTimscampi can this be added to the backports or did we never add the original improvements?

MrTimscampi:
> @MrTimscampi can this be added to the backports or did we never add the original improvements?

The original changes (adding animations) were in 10.5.0.

However, this feels a little large to be backported (It's completely changing a module)

MrTimscampi:
It also uses a lot of ES6, which 10.5.0 doesn't support.

src/scripts/settings/userSettings.js
```diff
@@ -265,3 +265,33 @@ import events from 'events';
     }
 
 /* eslint-enable indent */
+export default {
```
dmitrylyzo, line 268: I tried to turn off fast fadein, but `userSettings` is broken when imported at ES6 module.

MrTimscampi: That's weird, aside from Playback (which is fixed in another PR), user settings work perfectly for me.

Though I would like to set the default limit to 100 (because of the lack of infinite scrolling) and disable the fast fade-in by default.

dmitrylyzo: Maybe we are using two different instances of `userSerrings`: `userSerrings` (imported in non-ES6) and `userSerrings.default` (imported in ES6).
I am not sure, but we probably have something like this
```js
{
    setUserInfo: setUserInfo,
    getData: getData,
    importFrom: importFrom,
    set: set,
    get: get,
    ...
    default: {
        setUserInfo: setUserInfo,
        getData: getData,
        importFrom: importFrom,
        set: set,
        get: get,
        ...
    }
}
```

dmitrylyzo: here `this.currentUserId` is `undefined` when accessed from `imageLoader` and contains valid id when accessed from user settings.
https://github.com/jellyfin/jellyfin-web/blob/2cb5e8c5d07e4954f608ec1fb4bd3fbb62909a42/src/scripts/settings/userSettings.js#L67

MrTimscampi: Should be fixed with b8422a2

MrTimscampi:
> Now it works.
> 
> It seems that our `export default {}` doesn't work as we expected. At current moment, they may be broken for some modules (at least, `userSettings`).

Seems like we need to replace the imports with: `import * as whatever from 'whatever'` for it to work, at least for userSettings.

src/components/images/imageLoader.js
```diff
@@ -1,50 +1,83 @@
...
+            source = entry;
         }
 
-        if (!source) {
-            return;
+        if (entry.intersectionRatio > 0 && source) {
+            fillImageElement(entry.target, source);
+        } else if (!source) {
+            emptyImageElement(entry.target);
         }
```
dmitrylyzo, line 30: I played with it on TV layout.

- Open Music / Albums (may be any big library).
- Scroll down using keyboard.
- Then scroll up step by step.

Cards loose background. For some reason, they have `background` being out of view (looked in Inspector). Maybe loading is not fast enough for scrolling. And you make them empty because `source = entry.target.getAttribute('data-src')` is `undefined`.

dmitrylyzo: This will fix, but no fading effect: just show image if it is in view, regardless of value of `data-src` attribute.
```js
        if (entry.intersectionRatio > 0) {
            if (source) fillImageElement(entry.target, source);
        } else if (!source) {
            emptyImageElement(entry.target);
        }
```

MrTimscampi: I tried something, it seems to somewhat fix it for me. I have a few cases where images take a while to load, so the cards show up empty&& entry.intersectionRatio === 0 (But that happens with the current implementation as well).

src/components/images/imageLoader.js
```diff
@@ -1,50 +1,83 @@
...
+            throw new Error('entry cannot be null');
         }
 
-        if (!source) {
-            source = elem.getAttribute('data-src');
+        var source = undefined;
+        if (entry.target) {
+            source = entry.target.getAttribute('data-src');
+        } else {
+            source = entry;
```
dmitrylyzo, line 23: I think that if we get here, we will fail in `fillImageElement(entry.target, source);`

src/components/images/imageLoader.js
```diff
@@ -1,50 +1,83 @@
...
         }
 
-        if (!source) {
-            return;
+        if (entry.intersectionRatio > 0) {
+            fillImageElement(entry.target, source);
```
dmitrylyzo, line 27: You removed check for `source` - we will fail `if (url === undefined)`. That's why I left `if (source) ...`

dmitrylyzo:
```suggestion
            if (source) fillImageElement(entry.target, source);
```
You can fill image if you have `source`.

MrTimscampi: Added the suggestion, I can't seem to trigger the load error anymore :+1:

src/components/images/imageLoader.js
```diff
@@ -73,11 +69,8 @@ import 'css!./style';
 
         elem.setAttribute("data-src", url);
 
-        if (userSettings.enableFastFadein()) {
-            elem.classList.remove('lazy-image-fadein-fast');
-        } else {
-            elem.classList.remove('lazy-image-fadein');
-        }
+        elem.classList.remove('lazy-image-fadein-fast');
+        elem.classList.remove('lazy-image-fadein');
```
JustAMan, line 73: cannot we combine them in one line?

JustAMan:
@MrTimscampi you got merge conflicts here

PR #1067, opened 2020-04-12
Add "dummy" heartbeat on WebSocket

billyinferno:
**Changes**
Add dummy heartbeat on WebSocket to avoid cloudflare closing the web socket connection after 100s due no activity on the web socket.

### DIFF
--- a/src/libraries/apiclient/apiclientcore.js
+++ b/src/libraries/apiclient/apiclientcore.js
@@ -154,7 +154,9 @@ define(["events", "appStorage"], function(events, appStorage) {
 
     function onWebSocketOpen() {
         var instance = this;
-        console.debug("web socket connection opened"), events.trigger(instance, "websocketopen")
+        console.debug("web socket connection opened"), events.trigger(instance, "websocketopen");
+        // send socket heartbeat
+        socketHeartBeat(this._webSocket);
     }
 
     function onWebSocketError() {
@@ -173,6 +175,20 @@ define(["events", "appStorage"], function(events, appStorage) {
                 events.trigger(apiClient, "websocketclose")
             }, 0)
         }
+    }
+    
+    function socketHeartBeat(socket) {
+        // check if socket already define or not
+        if(!socket) return;
+        // ensure that socket already in the readyState
+        if(socket.readyState !== WebSocket.OPEN) return;
+        // sent heartbeat to websocket.
+        socket.send("heartbeat");
+        // set interval to 60s, because cloudflare will going to close the connection
+        // without any message being sent in 100s (for free-tier).
+        setTimeout(function() {
+            socketHeartBeat(socket);
+        }, 60000);
     }
 
     function normalizeReturnBitrate(instance, bitrate) {


MrTimscampi:
Ideally, the proper way to do this would be to send a "ping" message from the server, then handle that message on the client and send back a "pong".

This would ensure that, should the connection to the WebSocket be lost at some point, it can either be spun back up or the loss of connection can be handled properly, both on the client and the server (If the server sends a ping but doesn't receive a pong, the client is down. If the client doesn't receive a ping for a while, the server is down).

What does @jellyfin/backend think about it?

billyinferno:
> 
> 
> Ideally, the proper way to do this would be to send a "ping" message from the server, then handle that message on the client and send back a "pong".
> 
> This would ensure that, should the connection to the WebSocket be lost at some point, it can either be spun back up or the loss of connection can be handled properly, both on the client and the server (If the server sends a ping but doesn't receive a pong, the client is down. If the client doesn't receive a ping for a while, the server is down).
> 
> What does @jellyfin/backend think about it?

Agree on this.

I also tell the same thing on the reddit post regarding this. Because each browser have slightly different method of when the websocket is already connected, and each device will have different timing when it connected. So it will be hard to fine tune if this is from client.

By right, ping/pong frame should be handle by server.

src/libraries/apiclient/apiclientcore.js
```diff
@@ -174,6 +176,20 @@ define(["events", "appStorage"], function(events, appStorage) {
...
+        if(!socket) return;
+        // ensure that socket already in the readyState
+        if(socket.readyState !== WebSocket.OPEN) return;
+        // sent heartbeat to websocket.
+        socket.send("heartbeat");
+        // set interval to 60s, because cloudflare will going to close the connection
+        // without any message being sent in 100s (for free-tier).
+        setTimeout(function() {
+            socketHeartBeat(socket);
+        }, 60000);
```
JustAMan, line 191:
```suggestion
        }, 20000);
```
I think I read @mcarlton00 reporting 30 seconds being timed out somewhere else, and sending a heartbeat each 20 seconds should not consume much of resources anyway :)

mcarlton00: You might be thinking of something else.  I was having issues in clients where they haven't "checked in" in ~15 minutes, the connection drops.

billyinferno: @mcarlton00 
Means, no message being passed on the WS in 15 mins?
It probably happens also, because at the end it will depends on a lot of implementation and client (like firefox will be different with chrome), and also whether you behind reverse proxy or not (like your caddy or nginx configuration?)

@JustAMan 
actually we don't even need to sent strings, just like the actual ping/pong (which only change the control frame op code).
But it's just easier to check on the developer tools when I can see the "heartbeat" message.

mcarlton00: I think more context for mine would be helpful, sorry about that.

So the issue I've run into specifically is with the Mopidy extension.  If you just play shorter songs, it's all good.  But if you play something longer like an audiobook, it will report it's started playing to the server and then just continue going locally.  Currently I don't have it periodically posting updates, so the server receives a "client is playing" message and then doesn't hear anything else from the client.  After 15 minutes or so the server effectively says "It's probably not playing anymore" and stops tracking progress.  This is useful as a timeout measure if a client crashes, it doesn't mark a whole file as watched.  It just means that during playback clients need to post status updates regularly.

TLDR: the 60000 is likely fine.

MrTimscampi:
Do note that, as outlined in #1011, the server-side of that PR has a ping/pong feature that has a lot of overlap with this.

We probably want to use that for a keepalive, to avoid multiple implementations of essentially the same thing (However, I don't think that PR currently uses the ping/pong in a generic way throughout the client).

OancaAndrei:
The following would be my proposed implementation for this `KeepAlive` feature.

Briefly:
- once the client connects to the server (WebSocket connection), the server will send a `ForceKeepAlive` message containing the timeout after which the client is considered to be lost.
- the client will then have to send a `KeepAlive` message often enough to comply with the timeout.
- for every sent `KeepAlive` message the server will reply with a `KeepAlive` message as well.
- the server will also monitor all active connections and send a last `ForceKeepAlive` message to clients that haven't updated their last activity in a while.

The idea is to avoid contacting all clients at the same time but rather let them do the "hard work" as also distribute the messages across time.
Another addition would be to consider every transferred message a substitute for the `KeepAlive` message, thus saving some data, but I think this would render the client implementation to complex.

Handling of inactive clients isn't implemented yet. How should this part behave? Shall it close the WebSocket connection and notify the interested Session? Same question for the client side, shall it try to create a new connection, or just let the network do its work?

Server side [commit](https://github.com/jellyfin/jellyfin/pull/2733/commits/0b0a5c09ea66c81a1c735377fd39419942e4ea89).
Client side [commit](https://github.com/jellyfin/jellyfin-web/pull/1011/commits/0282cc7985ed84de4bef06fc24993da38cadfd00).

MrTimscampi:
@OancaAndrei Sounds good!

I can only speak for the client-side, but a nice way of doing things would be:

* Clients sends a `keepalive` but gets no response in X seconds
* The websocket connection is marked as "degraded" and a toast is shown to the user
* The websocket tries to re-connect
* If it fails, stop everything and move to a "Server is not available, trying to reconnect..." page
* Keep trying to contact the server
* Once a connection is made, go back to what the client was doing before (If playback was happening, perhaps bring them back to the item's details page instead)

MrTimscampi:
Closing as this was done as part of #1011

PR #1068, opened 2020-04-12
Pass title and filename to native shell

dkanada:


src/components/filedownloader.js
```diff
@@ -4,7 +4,7 @@ export function download(items) {
 
     if (window.NativeShell) {
         items.map(function (item) {
-            window.NativeShell.downloadFile(item.url);
+            window.NativeShell.downloadFile(item);
```
JustAMan, line 7: why?

PR #1069, opened 2020-04-12
Force the use of single quotes

MrTimscampi:
**Changes**

This performs an automated change of all double quotes to single quotes (through ESLint) and adds the linting rule for it.

**Issues**

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -40,6 +40,7 @@ module.exports = {
         'no-multiple-empty-lines': ["error", { "max": 1 }],
         'no-trailing-spaces': ["error"],
         'one-var': ["error", "never"],
+        'quotes': ["error", "single", { "avoidEscape": true, "allowTemplateLiterals": false }],
         'semi': ["error"],
         'space-before-blocks': ["error"]
     },
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -8,8 +8,8 @@ const htmlmin = require('gulp-htmlmin');
 const imagemin = require('gulp-imagemin');
 const sourcemaps = require('gulp-sourcemaps');
 const mode = require('gulp-mode')({
-    modes: ["development", "production"],
-    default: "development",
+    modes: ['development', 'production'],
+    default: 'development',
     verbose: false
 });
 const stream = require('webpack-stream');
@@ -55,7 +55,7 @@ const options = {
 function serve() {
     browserSync.init({
         server: {
-            baseDir: "./dist"
+            baseDir: './dist'
         },
         port: 8080
     });
--- a/src/bundle.js
+++ b/src/bundle.js
@@ -5,150 +5,150 @@
 var _define = window.define;
 
 // document-register-element
-var docRegister = require("document-register-element");
-_define("document-register-element", function() {
+var docRegister = require('document-register-element');
+_define('document-register-element', function() {
     return docRegister;
 });
 
 // fetch
-var fetch = require("whatwg-fetch");
-_define("fetch", function() {
+var fetch = require('whatwg-fetch');
+_define('fetch', function() {
     return fetch;
 });
 
 // query-string
-var query = require("query-string");
-_define("queryString", function() {
+var query = require('query-string');
+_define('queryString', function() {
     return query;
 });
 
 // flvjs
-var flvjs = require("flv.js/dist/flv").default;
-_define("flvjs", function() {
+var flvjs = require('flv.js/dist/flv').default;
+_define('flvjs', function() {
     return flvjs;
 });
 
 // jstree
-var jstree = require("jstree");
-require("jstree/dist/themes/default/style.css");
-_define("jstree", function() {
+var jstree = require('jstree');
+require('jstree/dist/themes/default/style.css');
+_define('jstree', function() {
     return jstree;
 });
 
 // jquery
-var jquery = require("jquery");
-_define("jQuery", function() {
+var jquery = require('jquery');
+_define('jQuery', function() {
     return jquery;
 });
 
 // hlsjs
-var hlsjs = require("hls.js");
-_define("hlsjs", function() {
+var hlsjs = require('hls.js');
+_define('hlsjs', function() {
     return hlsjs;
 });
 
 // howler
-var howler = require("howler");
-_define("howler", function() {
+var howler = require('howler');
+_define('howler', function() {
     return howler;
 });
 
 // resize-observer-polyfill
-var resize = require("resize-observer-polyfill").default;
-_define("resize-observer-polyfill", function() {
+var resize = require('resize-observer-polyfill').default;
+_define('resize-observer-polyfill', function() {
     return resize;
 });
 
 // shaka
-var shaka = require("shaka-player");
-_define("shaka", function() {
+var shaka = require('shaka-player');
+_define('shaka', function() {
     return shaka;
 });
 
 // swiper
-var swiper = require("swiper/js/swiper");
-require("swiper/css/swiper.min.css");
-_define("swiper", function() {
+var swiper = require('swiper/js/swiper');
+require('swiper/css/swiper.min.css');
+_define('swiper', function() {
     return swiper;
 });
 
 // sortable
-var sortable = require("sortablejs").default;
-_define("sortable", function() {
+var sortable = require('sortablejs').default;
+_define('sortable', function() {
     return sortable;
 });
 
 // webcomponents
-var webcomponents = require("webcomponents.js/webcomponents-lite");
-_define("webcomponents", function() {
+var webcomponents = require('webcomponents.js/webcomponents-lite');
+_define('webcomponents', function() {
     return webcomponents;
 });
 
 // libass-wasm
-var libassWasm = require("libass-wasm");
-_define("JavascriptSubtitlesOctopus", function() {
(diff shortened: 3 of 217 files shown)


ferferga:
@MrTimscampi One appreciation: I think that we should merge this once we close PRs that have been inactive more than 15 days ago. Iirc all of them have merge conflicts atm and merging this would make everything much worse and annoying for those contributors.

MrTimscampi:
> @MrTimscampi One appreciation: I think that we should merge this once we close PRs that have been inactive more than 15 days ago. Iirc all of them have merge conflicts atm and merging this would make everything much worse and annoying for those contributors.

This is pretty trivial to rebase for, though. Plus, there will always be older PRs that should be merged before this.

Waiting a long time will simply make reviewing this even more annoying, without any guarantee that we'll ever be in a state where we can merge this before anything else without impacting opened PRs.

.eslintrc.yml
```diff
@@ -75,6 +75,7 @@ rules:
   no-multiple-empty-lines: ["error", { "max": 1 }]
   no-trailing-spaces: ["error"]
   one-var: ["error", "never"]
+  quotes: ["error", "single", { "avoidEscape": true, "allowTemplateLiterals": false }]
```
thornbill, line 78: There are some advantages to using template literals for multiline or formatted strings. Could we leave it enabled? I can't say that I have seen it used inappropriately.

MrTimscampi: It doesn't forbid template literals in all code. It simply prevents simple strings from using the backticks instead of quotes.

So this is correct:
```javascript
const myLiteral = `Lorem ${ipsum} dolor sit amet.`
```

But this is not:
```javascript
const myString = `Lorem ipsum dolor sit amet.`
```

MrTimscampi:
For the record and to help this get reviewed quicker, here is some more info.

Here's the command used:
```bash
yarn lint --no-eslintrc --rule 'quotes: ["error", "single", { "avoidEscape": true, "allowTemplateLiterals": false }]' --fix
```

And here is the implementation of that automatic fix for sanity checking:
https://github.com/eslint/eslint/blob/master/lib/rules/quotes.js#L50

cromefire:
Is there a reason for single quotes? Those are _really_ annoying on German keyboards and many languages don't support single quote strings so it might be annoying to switch for some people

cromefire:
Examples are at least Java, kotlin, c & c++ and I'm not sure about C# atm

MrTimscampi:
It's what is used by most style guides, so what most people will be familiar with and would naturally use.

Standard: https://github.com/standard/eslint-config-standard/blob/master/eslintrc.json#L178  
Google: https://github.com/google/eslint-config-google/blob/master/index.js#L288  
Airbnb: https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/style.js#L451

cromefire:
Well those are usually companies from the us and also people working full time on js, but someone (at least from Germany I guess other countries also have this) the ' key is not accessible without pulling both hands to the complete sides of the keyboard and also not everyone here is a full time js dev as you can see above many languages don't allow this so it's not very natural when switching languages. (Also there's still the argument that English uses the ' in words often)

cromefire:
Tslint's default for example doesn't even make a choice about that.

MrTimscampi:
These three are, by far, the most widely used style guides in the JavaScript ecosystem. So I think it makes sense to take some inspiration from them.

cromefire:
Well I don't think putting popularity over dev exp is a good idea if you argue that way we also have to use react, but we chose not to because it's just not a good dev exp.
A lot of things are popular, but that's still no excuse for using it

cromefire:
And again as you said that's only the js ecosystem overall there'll probably be more double quotes.

.eslintrc.yml
thornbill: Right. My point was that this:
```js
const literal = `This
string
is

preformatted.`
```
Is easier to read than:
```js
const string = 'This\nstring\nis\n\nnot preformatted.'
```

MrTimscampi: That doesn't trigger the rule, though :)

cromefire: Eslint only enforces it when literally replacing the quotes works, so not on template strings and not on multiline

JustAMan:
@cromefire Let me put it blandly. The convenience should be reached for the major part of the contributing developers (see repo stats), and that majority isn't users with German keyboard. On most other keyboard layouts typing single quote is easier than double quote.

cromefire:
Well from what I see it's still easier to type on a US keyboard and that still has the problem that ' usually stands for char and might confuse people and interrupt their typing flow

JustAMan:
> still has the problem that ' usually stands for char

In a few languages. But JavaScript is not one of them. Let's not descend into a holywar here.

cromefire:
Well you also want to help people to work on this that do not come from javascript right?

cromefire:
Also from an linguistic point of view: ' is used for direct quotes and if you leave out chars.
There's just a lot of arguments for " and almost none (except it a slight bit easier on many keyboards) for ', that's the point I'm trying to make

MrTimscampi:
> Also from an linguistic point of view

This feels a bit like grasping at straws, honestly.

The main argument is that most JavaScript projects use single quotes. All the style guides that most people are accustomed to do, as well. As such, for most people, it will feel more natural to use single quotes in a JavaScript project.

We do encourage non-JavaScript people to try to contribute, but I doubt that quotes would be the hangup when there's going to be a fairly strict style guide when we're done.

Some people won't like that we use 4 spaces for indentation and would prefer 2 spaces or even tabs. Some people will want block spacing, some won't. We can't please everyone.

cromefire:
But what may be another major argument (the linguistic kind was just to show there are a lot of arguments of any kind) is that this prevents a consistent style of all jellyfin projects (because of the server & Android projects)

MrTimscampi:
> (because of the server & Android projects)

Which are in different languages, so not an issue here.

The Android app will be a native one down the line and will share most of its code with the Android TV client.

All the Javascript in the server (basically plugin pages) will be removed at some point to be replaced with a proper plugin configuration API.

cromefire:
I didn't even know there is js in the server I was talking about the other languages. When you're switching between projects a lot you're typing flow get disrupted. You don't really think about typing y but you type "automatically" disrupting this is the last thing you want to do, so a common style between all language is good (I recognize that this is only possible to so e extend)

ferferga:
@MrTimscampi I suppose you checked but are you sure this kind of lines for html code doesn't break, right?

So this:
``html += "<a style='blah'></a>"``

Is not converted into this:

``html += '<a style='blah'></a>'``

cromefire:
The rule handles that

MrTimscampi:
> @MrTimscampi I suppose you checked but are you sure this kind of lines for html code doesn't break, right?
> 
> So this:
> `html += "<a style='blah'></a>"`
> 
> Is not converted into this:
> 
> `html += '<a style='blah'></a>'`

The automatic fix used for this handles this case. Look here: https://github.com/eslint/eslint/blob/master/lib/rules/quotes.js#L50

src/components/fullscreenManager.js
```diff
@@ -0,0 +1,103 @@
+define(['events', 'dom', 'apphost', 'browser'], function (events, dom, appHost, browser) {
```
JustAMan, line 1: why is this file here?

dmitrylyzo: conflicts are not resolved

JustAMan:
For anyone interested, here's my script that operates on downloaded diff file:
<details><summary>Python 2.7 script</summary>

```python
from collections import defaultdict
import sys

with open('1069.diff') as inp:
    lines = inp.readlines()

minus = [(l[1:].rstrip(), idx) for idx,l in enumerate(lines) if l.startswith('-') and not l.startswith('---')]
plus = [(l[1:].rstrip(), idx) for idx,l in enumerate(lines) if l.startswith('+') and not l.startswith('+++')]

print len(minus), len(plus)

mm = defaultdict(list)
for m, idx in minus:
    mm[len(m)].append((m, m.replace('"', '').replace("'", ''), idx))

#print len(mm)

skipped = []
bad = []
checking = {}
for p, pidx in plus:
    p1 = p.replace('"', '').replace("'", '')
    matches = defaultdict(list)
    for m, m1, midx in mm[len(p)]:
        if m1 == p1:
            matches[p].append(m)
            break
    if not matches:
        skipped.append((p, pidx))
    elif len(matches) == 1 and len(matches.values()[0]) == 1:
        checking[pidx] = (p, matches.values()[0][0])
    else:
        bad.append(matches)

print 'skipped', len(skipped)
for p, pidx in skipped:
    print '%d:\t%s' % (pidx+1, p)
if bad:
    print 'bad items', len(bad)

for pidx, (p, m) in checking.iteritems():
    if p == m:
        continue
    diff = set()
    for pch, mch in zip(p, m):
        if pch != mch:
            diff.add((pch, mch))
    if diff != set([("'", '"')]):
        print 'BAD at %d' % (pidx+1)
        import pdb;pdb.set_trace()

print 'VERIFIED'
```
</details>

src/components/fullscreenManager.js
MrTimscampi: I need to rebase and re-run the fix for it to be current with master.

PR #1072, opened 2020-04-12
Move "hide watched media" checkbox

ferferga:
This checkbox is located in a way that makes you think that it affects only an specific library, when it affects all the libraries, so I moved it to a more sensible way, where it's function is clear.

**Before**
before

**After**
after

### DIFF
--- a/src/components/homescreensettings/homescreensettings.template.html
+++ b/src/components/homescreensettings/homescreensettings.template.html
@@ -10,6 +10,13 @@
             <div class="fieldDescription">${LabelPleaseRestart}</div>
         </div>
 
+        <div class="verticalSection verticalSection-extrabottompadding">
+            <label class="checkboxContainer">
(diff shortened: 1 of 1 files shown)


Artiume:
The Spanish threw me for a loop for a minute, was trying to figure out what button this was 😅

ferferga:
@Artiume You understand it now right? Don't hesitate to ask me. Next time I will try to put the texts in English before the screenshots hehe.

Artiume:
Yup, I like the PR. I would definitely fall into the category of assuming it only impacted specific stuff, not a global variable lol.

ferferga:
@MrTimscampi I don't understand what you mean. You say so because you think the option is dead and does nothing?

Try disabling it and watch something in the "Recent" section of your home screen. If you have it enabled, the movie will disappear from there. If you don't, it will be marked as viewed but it will be present.

PR #839, opened 2020-02-22
Added extra playback keybindings

mijofa:
**Changes**
Added playback keybindings for:
* ArrowUp & ArrowDown: Volume up & down
* Shift-P & Shift-N: Previous & next track
* 0-9: Jump to 0,10,20,30...90% of the current media
* Home, & End: Jump to 0% or 100% of the current media

**Issues**
N/A

**Other info**
I specifically wanted the vol up/down bindings, so I'm happy if only those get merged in. I did the rest because I felt some sort of defined standard should be followed, so I reffered to [YouTube's bindings](https://support.google.com/youtube/answer/7631406?hl=en) and implemented all of them.

This was also discussed in the Matrix/IRC channel:
> 2020-02-19T20:21:54+1100 <mijofa_> Is there any intention of (or interest in) adding extra keybindings to the web video player?        I've had a snippet of js that I wrote for Emby I've been dropping on the end of videoosd.js with each update, but I'm considering making it into an actual patch and pushing it upstream if it's actually got a chance of being merged in
> 2020-02-19T20:22:35+1100 <dkanada[m]> whats added?
> 2020-02-19T20:24:15+1100 <mijofa_> Main things I wanted/needed was play/pause on space, seek on left/right, and volume on up/down. But I added all the ones YouTube uses as documented at https://support.google.com/youtube/answer/7631406
> 2020-02-19T20:25:15+1100 <mijofa_> http://ix.io/2c7E/js
> 2020-02-19T20:25:57+1100 <telans[m]>  I can seem to seek & pause with keys at the moment
> 2020-02-19T20:26:09+1100 <mijofa_> (disclaimer, that code is horrible, it was written in the least "intrusive" way possible)
> 2020-02-19T20:26:42+1100 <mijofa_> If I recall correctly my main issue with the play/pause was that it wasn't entirely consistent, you needed to be focussed on the right element and the OSD needed to not be active. Or something like that
> 2020-02-19T20:27:38+1100 <mijofa_> Also I'm sure that some of this would've gotten better since I actually wrote it, but it's still not quite good enough for me
> 2020-02-19T20:44:05+1100 <JustAMan[m]> mijofa_: make a PR, but check existing ones first
> 2020-02-19T20:44:13+1100 <JustAMan[m]> I think I saw something pretty similar recently
> 2020-02-19T20:44:33+1100 <mijofa_> JustAMan[m]: Ok, will do, thanks. :)
[...]

PR #1074, opened 2020-04-13
Move delayed volume update to playbackManager

dmitrylyzo:
For #839
Idea is that `playbackManager` must control progress updating by itself.

**Changes**
Move delayed volume update to `playbackManager`.

Mute is also delayed now (send update, not actual mute).

### DIFF
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -3282,7 +3282,7 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
 
         function onPlaybackVolumeChange(e) {
             var player = this;
-            sendProgressUpdate(player, 'volumechange');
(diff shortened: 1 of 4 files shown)


dkanada:
@sparky8251 this solves your least favorite problem!

src/components/playback/playbackmanager.js
```diff
@@ -3377,7 +3377,14 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
 
         pluginManager.ofType('mediaplayer').map(initMediaPlayer);
 
+        /** Delay timer for sendProgressUpdate */
+        var sendProgressUpdateTimer;
+        /** Delay time in ms for sendProgressUpdate */
```
dkanada, line 3382:
```suggestion

        /** Delay time in ms for sendProgressUpdate */
```

PR #1083, opened 2020-04-13
Handle 401 & 403 Server Responses Correctly

mark-monteiro:
* Update the login page to handle a 403 response from the server. A new message/translation string has been added to to display to the user in this case.
* Update to expect a 403 response for parental control limitations

The associated server updates are in https://github.com/jellyfin/jellyfin/pull/2861

### DIFF
--- a/src/controllers/auth/login.js
+++ b/src/controllers/auth/login.js
@@ -26,7 +26,7 @@ define(["apphost", "appSettings", "dom", "connectionManager", "loading", "layout
 
             if (response.status === 401 || response.status === 403) {
                 require(["toast"], function (toast) {
-                    var messageKey = response.status === 401 ? "MessageInvalidUser" : "MessageUnauthorizedUser"
+                    var messageKey = response.status === 401 ? "MessageInvalidUser" : "MessageUnauthorizedUser";
                     toast(Globalize.translate(messageKey));
                 });
             } else {
--- a/src/controllers/auth/login.js
+++ b/src/controllers/auth/login.js
@@ -24,9 +24,10 @@ define(["apphost", "appSettings", "dom", "connectionManager", "loading", "layout
             page.querySelector("#txtManualPassword").value = "";
             loading.hide();
 
-            if (response.status === 401) {
+            if (response.status === 401 || response.status === 403) {
                 require(["toast"], function (toast) {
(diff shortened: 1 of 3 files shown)


src/controllers/auth/login.js
```diff
@@ -24,9 +24,10 @@ define(["apphost", "appSettings", "dom", "connectionManager", "loading", "layout
             page.querySelector("#txtManualPassword").value = "";
             loading.hide();
 
-            if (response.status === 401) {
+            if (response.status === 401 || response.status === 403) {
```
MrTimscampi, line 27:
```suggestion
            const UnauthorizedOrForbidden = [401, 403]
            if (UnauthorizedOrForbidden.inclues(response.status)) {
```

It doesn't make it shorter, but it's a bit more obvious and we can more the error codes to some utility module afterward.

For reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

src/controllers/auth/login.js
```diff
@@ -24,9 +24,10 @@ define(["apphost", "appSettings", "dom", "connectionManager", "loading", "layout
...
                 require(["toast"], function (toast) {
-                    toast(Globalize.translate("MessageInvalidUser"));
+                    var messageKey = response.status === 401 ? "MessageInvalidUser" : "MessageUnauthorizedUser";
```
MrTimscampi, line 29: Feel free to use `let` or `const` in modules, everything is passed through Babel (not only what is explicitly set to be an ES6 module).

src/controllers/auth/login.js
mark-monteiro: Done 👍

src/controllers/auth/login.js
mark-monteiro: Done 👍 So nice to be able to use modern features :D

MrTimscampi: We have polyfill linting too, now, so you can use new APIs and it'll warn you if stuff needs a polyfill :)

src/controllers/auth/login.js
dkanada: I think the other way was more readable for what it's worth.

MrTimscampi:
Requires jellyfin/jellyfin#2861 to be merged first.

PR #1084, opened 2020-04-13
Fix more Sonarqube issues

MrTimscampi:
## Sonarqube's Fixing Adventure - Part 2: Merging Tendencies

* Fixes all Vulnerabilities reported by Sonarqube
* Adds a utility CSS class to hide elements from the user but not from screen readers
* Adds screen reader hints for tables and icons
* Fixes a bunch of Major issues

### DIFF
--- a/src/components/guide/tvguide.template.html
+++ b/src/components/guide/tvguide.template.html
@@ -9,8 +9,8 @@
 
     <div class="guide-headerTimeslots">
         <div class="guide-channelTimeslotHeader">
-            <button is="paper-icon-button-light" type="button" class="btnGuideViewSettings">
-                <span class="material-icons btnGuideViewSettingsIcon more_horiz"></span>
+            <button is="paper-icon-button-light" type="button" class="btnGuideViewSettings" title="${ButtonMore}">
+                <span class="material-icons btnGuideViewSettingsIcon more_horiz" aria-hidden="true"></span>
             </button>
         </div>
         <div class="timeslotHeaders scrollX guideScroller"></div>
@@ -29,10 +29,10 @@
 </div>
 
 <div class="guideOptions hide">
-    <button is="paper-icon-button-light" type="button" class="btnPreviousPage">
-        <span class="material-icons arrow_back"></span>
+    <button is="paper-icon-button-light" type="button" class="btnPreviousPage" title="${LabelPrevious}">
+        <span class="material-icons arrow_back" aria-hidden="true"></span>
     </button>
-    <button is="paper-icon-button-light" type="button" class="btnNextPage">
-        <span class="material-icons arrow_forward"></span>
+    <button is="paper-icon-button-light" type="button" class="btnNextPage" title="${LabelNext}">
+        <span class="material-icons arrow_forward" aria-hidden="true"></span>
     </button>
 </div>
--- a/src/components/libraryoptionseditor/libraryoptionseditor.js
+++ b/src/components/libraryoptionseditor/libraryoptionseditor.js
@@ -120,7 +120,7 @@ define(['globalize', 'dom', 'emby-checkbox', 'emby-select', 'emby-input'], funct
             html += plugin.Name;
             html += '</h3>';
             html += '</div>';
-            i > 0 ? html += '<button type="button" is="paper-icon-button-light" title="' + globalize.translate('ButtonUp') + '" class="btnSortableMoveUp btnSortable" data-pluginindex="' + i + '"><span class="material-icons keyboard_arrow_up"></span></button>' : plugins.length > 1 && (html += '<button type="button" is="paper-icon-button-light" title="' + globalize.translate('ButtonDown') + '" class="btnSortableMoveDown btnSortable" data-pluginindex="' + i + '"><span class="material-icons keyboard_arrow_down"></span></button>'), html += '</div>';
+            index > 0 ? html += '<button type="button" is="paper-icon-button-light" title="' + globalize.translate('ButtonUp') + '" class="btnSortableMoveUp btnSortable" data-pluginindex="' + index + '"><span class="material-icons keyboard_arrow_up"></span></button>' : plugins.length > 1 && (html += '<button type="button" is="paper-icon-button-light" title="' + globalize.translate('ButtonDown') + '" class="btnSortableMoveDown btnSortable" data-pluginindex="' + index + '"><span class="material-icons keyboard_arrow_down"></span></button>'), html += '</div>';
         });
 
         html += '</div>';
@@ -297,8 +297,8 @@ define(['globalize', 'dom', 'emby-checkbox', 'emby-select', 'emby-input'], funct
                 typeOptions = {
                     Type: type
                 };
+                currentLibraryOptions.TypeOptions.push(typeOptions);
             }
-            currentLibraryOptions.TypeOptions.push(typeOptions);
             var availableOptions = getTypeOptions(currentAvailableOptions || {}, type);
             var imageOptionsEditor = new ImageOptionsEditor();
             imageOptionsEditor.show(type, typeOptions, availableOptions);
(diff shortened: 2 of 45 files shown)


src/strings/en-us.json
```diff
@@ -309,6 +309,7 @@
     "HeaderApiKey": "API Key",
     "HeaderApiKeys": "API Keys",
     "HeaderApiKeysHelp": "External applications are required to have an API key in order to communicate with Jellyfin Server. Keys are issued by logging in with a Jellyfin account, or by manually granting the application a key.",
+    "ApiKeysCaption": "List of the currently enable API keys",
```
dkanada, line 312:
```suggestion
    "ApiKeysCaption": "List of the currently enabled API keys",
```

src/controllers/auth/addserver.js
```diff
@@ -46,9 +46,7 @@ define(["appSettings", "loading", "browser", "emby-button"], function(appSetting
         view.querySelector(".addServerForm").addEventListener("submit", onServerSubmit);
         view.querySelector(".btnCancel").addEventListener("click", goBack);
 
-        require(["autoFocuser"], function (autoFocuser) {
-            autoFocuser.autoFocus(view);
-        });
```
dkanada, line 51: @dmitrylyzo thoughts?

MrTimscampi: The thought behind replacing this with a simpler focus is that AddServer really only has one thing you can focus on: the server address field.

src/controllers/auth/addserver.js
```diff
@@ -46,9 +46,7 @@ define(["appSettings", "loading", "browser", "emby-button"], function(appSetting
...
+        view.querySelector("#txtServerHost").focus();
```
dmitrylyzo, line 49: In the current case, there is no harm - scrolling is not required. But `scrollManager` expects `preventScroll: true`. So `focusManager.focus(elem)` is preferred.

src/controllers/auth/login.js
```diff
@@ -124,9 +126,13 @@ define(["apphost", "appSettings", "dom", "connectionManager", "loading", "layout
             html += "</div>";
             html += "</div>";
             html += "</button>";
-        }
+        });
 
-        context.querySelector("#divUsers").innerHTML = html;
+        const divUsers = context.querySelector("#divUsers");
```
dmitrylyzo, line 131: This file is not transpiled yet.

src/controllers/auth/login.js
```diff
@@ -124,9 +126,13 @@ define(["apphost", "appSettings", "dom", "connectionManager", "loading", "layout
...
+        divUsers.innerHTML = html;
+        if (!layoutManager.tv) {
+            divUsers.querySelector(".firstCard").focus();
```
dmitrylyzo, line 134: For TV, I set focus in `showVisualForm` but `autoFocuser` is TV only. In fact, it finds the first focusable element (first card).

src/controllers/auth/login.js
MrTimscampi: Everything is transpiled :)

The package,json list is for module conversion only (ie turning the import/export into define/return). But Babel runs on everything.

dmitrylyzo: OK then :p

src/controllers/auth/login.js
MrTimscampi: Hm, I see the following in showManualForm:

```javascript
if (focusPassword) {
    context.querySelector("#txtManualPassword").focus();
} else {
    context.querySelector("#txtManualName").focus();
}
```

Wouldn't that cause issues with the focusManager when in TV layout?

MrTimscampi: Also wouldn't setting focus in showVisualForm lead to the focus never being on the first user card (where it would make the most sense, imo)

dmitrylyzo: > Wouldn't that cause issues with the focusManager when in TV layout?

If you call `focusManager.focus(elem)` you will prevent scroll (required by `scrollManager`). But if element is within viewport, there is no difference. But I would prefer `focusManager.focus(elem)` - this is OK for all layouts.

> Also wouldn't setting focus in showVisualForm lead to the focus never being on the first user card (where it would make the most sense, imo)

I think I got what you mean.
Here `showVisualForm` with my focus will be called before user cards are generated.
https://github.com/jellyfin/jellyfin-web/blob/4b69f85bb9dd43e0aa3548d639ffe0a92f5a2f1a/src/controllers/auth/login.js#L210-L211
I think this only works because `require` is "deferred". tbh, I never have a problem with this, but, to be true, this is not ideal.

Originally, my idea was to place focus code into each page, but many of them will be equal. So I combined common logic into `autoFocuser`. Also, it has restoring of paging buttons. But `autoFocuser` is optional, so I think you are free to just focus on first card for all layouts.

MrTimscampi: @dmitrylyzo I removed the focus changes, they'll be better in their own PR down the line.

src/components/guide/tvguide.template.html
```diff
@@ -9,8 +9,8 @@
 
     <div class="guide-headerTimeslots">
         <div class="guide-channelTimeslotHeader">
-            <button is="paper-icon-button-light" type="button" class="btnGuideViewSettings">
-                <span class="material-icons btnGuideViewSettingsIcon more_horiz"></span>
+            <button is="paper-icon-button-light" type="button" class="btnGuideViewSettings" title="${ButtonMore}">
+                <span class="material-icons btnGuideViewSettingsIcon more_horiz" aria-hidden="true"></span>
```
dmitrylyzo, line 13: _Not a blocker_
This button opens dialog "Settings". Maybe we should name it as dialog that opens and change icon.

src/components/libraryoptionseditor/libraryoptionseditor.js
```diff
@@ -109,8 +109,8 @@ define(['globalize', 'dom', 'emby-checkbox', 'emby-select', 'emby-input'], funct
         html += '<div class="metadataFetcher" data-type="' + availableTypeOptions.Type + '">';
         html += '<h3 class="checkboxListLabel">' + globalize.translate('LabelTypeMetadataDownloaders', globalize.translate(availableTypeOptions.Type)) + '</h3>';
         html += '<div class="checkboxList paperList checkboxList-paperList">';
-        for (var i = 0; i < plugins.length; i++) {
-            var plugin = plugins[i];
+
+        plugins.array.forEach((plugin, index) => {
```
dmitrylyzo, line 113: I have an error `TypeError: Cannot read property 'forEach' of undefined`.
```suggestion
        plugins.forEach((plugin, index) => {
```

src/components/guide/tvguide.template.html
MrTimscampi: I don't have access to the TV part of the client, so I'm leaving this for someone else :)

dmitrylyzo: I added free M3U provider. No guide, but some LiveTV features can be tested.

src/components/playback/playbackmanager.js
```diff
@@ -309,13 +309,11 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
...
-                    if (condition.Condition === 'LessThanEqual' && condition.Property === 'AudioSampleRate') {
-                        maxAudioSampleRate = condition.Value;
-                    }
-                    if (condition.Condition === 'LessThanEqual' && condition.Property === 'AudioBitrate') {
-                        maxAudioBitrate = condition.Value;
+                        return maxAudioBitDepth = condition.Value;
+                    } else if (condition.Condition === 'LessThanEqual' && condition.Property === 'AudioSampleRate') {
+                        return maxAudioSampleRate = condition.Value;
+                    } else if (condition.Condition === 'LessThanEqual' && condition.Property === 'AudioBitrate') {
+                        return maxAudioBitrate = condition.Value;
```
dkanada, line 316: Is this equivalent in JavaScript?

MrTimscampi: So the if...if...if, in this case, translates well to an if...else if...else if.

Ideally, in the future, we'll use [pattern matching](https://github.com/tc39/proposal-pattern-matching).

As for the return, the map() expects one. Returning the assignment will execute it (thus assigning the value) and fix the warnings on a map() function without any return.

PR #1087, opened 2020-04-14
Improve font serving for bandwidth constrained and low storage devices

ferferga:
Follow-up of jellyfin/jellyfin#2871

This PR adds the CSS files of the **subsetted** version of the new jellyfin-noto repository, so all the clients will fallback automatically to fetching fonts from server.

I label this as ``discussion needed`` because these files will be overriden anyway while Jellyfin Server is being built. However I think it makes sense to leave them here anyway, so we don't need to rework the build scripts for all the apps that use the web client and **only clients that need/want the packaged version need to have changes in the build process**

## Testing

Follow the jellyfin/jellyfin#2871 instructions for building a server version for testing this changes.

## Merging order

In order to not lose fonts, this PR must be merged after (jellyfin server PR) is merged. Clients that want to use packaged fonts need to follow it. It would be something as follows:

Move my repo to Jellyfin Organization -> Merge jellyfin/jellyfin#2871 -> Merge this -> Merge jellyfin/jellyfin-android#312

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,6 @@
     "hls.js": "^0.13.1",
     "howler": "^2.1.3",
     "intersection-observer": "^0.7.0",
-    "jellyfin-noto": "https://github.com/jellyfin/jellyfin-noto",
     "jquery": "^3.4.1",
     "jstree": "^3.3.7",
     "libass-wasm": "https://github.com/jellyfin/JavascriptSubtitlesOctopus#4.0.0-jf-cordova",
--- a/src/assets/css/font-faces.css
+++ b/src/assets/css/font-faces.css
@@ -84,7 +84,7 @@
     unicode-range: U+0000-00FF, U+0131, U+0152-0153, U+02BB-02BC, U+02C6, U+02DA, U+02DC, U+2000-206F, U+2074, U+20AC, U+2122, U+2191, U+2193, U+2212, U+2215, U+FEFF, U+FFFD;
 }
 
-/* ==================== FONT-WEIGHT 700 STYLES =======================================*/
+/* ==================== FONT-WEIGHT 700 STYLES ======================================= */
 
 /* cyrillic-ext */
 
@@ -179,18 +179,23 @@
 /* ========================= NOTO SANS HK TYPEFACES================= */
 
 @import url(noto-css/HK.css);
+
 /* ========================= NOTO SANS JP TYPEFACES ======================= */
 
 @import url(noto-css/JP.css);
+
 /* ========================= NOTO SANS KR TYPEFACES ================= */
 
 @import url(noto-css/KR.css);
+
 /* ======================= NOTO SANS SC TYPEFACES ================= */
 
 @import url(noto-css/SC.css);
+
 /* ======================= NOTO SANS TC TYPEFACES ================= */
 
 @import url(noto-css/TC.css);
+
 /* ============================= ARABIC TYPEFACES ============================================= */
 
 @font-face {
--- a/src/assets/css/font-faces.css
+++ b/src/assets/css/font-faces.css
@@ -1,0 +1,271 @@
+@font-face {
+    font-family: 'Noto Sans';
+    font-style: normal;
+    font-weight: 500;
+    font-display: swap;
+    src: url(../fonts/notosans/woff2/cyrillic-ext.woff2) format('woff2'),
+        url(../fonts/notosans/woff/cyrillic-ext.woff) format('woff');
+    unicode-range: U+0460-052F, U+1C80-1C88, U+20B4, U+2DE0-2DFF, U+A640-A69F, U+FE2E-FE2F;
+}
+
+
+/* cyrillic */
+
+@font-face {
+    font-family: 'Noto Sans';
+    font-style: normal;
+    font-weight: 500;
+    font-display: swap;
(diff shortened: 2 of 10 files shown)


ferferga:
@MrTimscampi Check jellyfin/jellyfin-android#312.

All the clients will rely on the server by default. Standalone ones like Android can choose whether relying in **packaged** or **subsetted** Clients that opt for the **packaged** version need to replace the files at build time, in a similar way as the shown in that PR.

Clients like Tizen that are standalone but don't need **packaged** versions because low storage and no bandwidth restrictions will connect to the server once the connection to the server is established (the only place where system fonts will be displayed will be at first boot, where no server has been chosen)

MrTimscampi:
> All the clients will rely on the server by default.

But the plan is to be able to ship server without any web client and ship the web client separately.

I understand how this is supposed to work, but as is, the use case mentioned before will produce a client without the fonts.

Maybe @dkanada or @joshuaboniface can confirm what the plan is for shipping web and server separately and making web optional.

ferferga:
@MrTimscampi Oh, alright, now I understand.
That should be a matter of doing the following:
* Clone this repo
* Build it
* Clone jellyfin-noto and copy the **subsetted** version.
* Build the rpm, deb, or whatever with that.

MrTimscampi:
We can probably use this as part of the standalone gulp task to clone the repo: https://www.npmjs.com/package/gulp-git

Then it's a matter of having it copy things to the proper subdirectory.

ferferga:
@MrTimscampi But that leads us to the problem we're trying to solve, right? Fonts taking a lot of storage.

dkanada:
> If that happens, clients like Tizen will 404 ofc. But the instance that's running with the jellyfin-web package will use the fonts that were bundled locally.

I don't think this is a reasonable solution. The way that the web client will eventually get bundled is completely incompatible with this idea, because it will still be an entirely separate package served by nginx on the same machine.

@JustAMan we really need to decide whether all this trouble is worth it for an optional font. I added an `other` folder to the repository for NSSM and it would be trivial to put the fonts from @ferferga at that location and add a URL to the CSS. It would be a single line change if anyone ever wanted to remove them, and it wouldn't detract whatsoever from the user experience if they were blocked or missing.

ferferga:
@dkanada I'm sure I'm probably missing something, but after the separation, we couldn't simply leave server with `jellyfin-web/assets/css` like right now? Without the client.

With my current PR, if we separate the web, the build scripts would anyway copy the assets to that directory (leaving only the font assets, nothing else) and they should still be reachable if the server's code that references the `jellyfin-web` serve isn't removed (which I suppose it won't for normal installations, where jellyfin-server and jellyfin-web packages would be installed altogether).

For jellyfin-web we can use subsetted version and let it fetch from server like non-mobile clients are supposed to do. Or make two packages, one like I described before and a `jellyfin-web-complete` that has the packaged fonts so people who don't want to rely at all on server have also them (although this doesn't make so much sense for me, because you still need a server to connect to to have an useful client)

dkanada:
> if the server's code that references the jellyfin-web serve isn't removed

This is exactly what would happen actually. The server will have an API and nothing else, and if the Android client connects to the server, there will be absolutely no web assets on that machine to query for fonts. You actually bring up a good point for the clients that need a web source as well, but we can cross that bridge later.

ferferga:
@dkanada Android client should use the packaged assets for reduced bandwidth consumption. The problem is with clients like Tizen and WebOS where storage is a problem (so web shouldn't bundle packaged assets) and data caps are not a problem.

dkanada:
I disagree that the Android client should double or triple in size just to package the fonts, which only leaves a remote solution. Whether or not that's a valid opinion is up for discussion, but no one else seems interested in this feature besides the two of us. :smile:

MrTimscampi:
> but no one else seems interested in this feature besides the two of us. smile

I'm fully for a remote option on something like assets.jellyfin.org :D

I don't think fetching fonts is really a privacy issue... And on mobile, they'll be cached after one fetch anyway, so...

JustAMan:
Well, I don't understand what's blocking us from using subsetted fonts in a standalone client.
It's just fonts are part of a client, not a server, and, when built standalone, they're packaged with the client.

dkanada:
Because the fonts double or triple the app size which many people might be unhappy with.

MrTimscampi:
> Because the fonts double or triple the app size which many people might be unhappy with.

It also causes issues on Tizen, as the app there needs to be bundled in (As Tizen doesn't allow hosted apps for most people) and the large client size makes it impossible to install for some people.

dmitrylyzo:
> As Tizen doesn't allow hosted apps for most people

Most pre-installed apps on my TV are hosted (less than 1MB).

> It can be acceptable for some special cases only.
https://developer.samsung.com/smarttv/develop/support/documentation-qa/developing-applications-qa.html#hosted-applications

So maybe.

ferferga:
@dkanada That's why there is a packaged version which weights less than half.

JustAMan:
> Because the fonts double or triple the app size which many people might be unhappy with.

But standalone build of jf-web should not be used as part of app, you should use bundled mode :)

ferferga:
@jellyfin/web This problem still exists, although this exact solution might not be the best at this moment. What's our stake into this?

dkanada:
resolved with #2010

PR #1088, opened 2020-04-15
Update and cleanup MediaSessions

MrTimscampi:
**Changes**

* Updates the MediaSessions feature to the current spec, removing non-existing parameters.
* Send the full artwork image to NativeShell

**Issues**


src/components/playback/mediasession.js
```diff
@@ -1,45 +1,26 @@
...
-    function seriesImageUrl(item, options) {
+    let currentPlayer;
 
+    function seriesImageUrl(item, options = {}, type = options.type || 'Primary') {
         if (item.Type !== 'Episode') {
             return null;
-        }
-
-        options = options || {};
-        options.type = options.type || "Primary";
```
JustAMan, line 21: this setting is missed, what if server or api client uses that?..

src/components/playback/mediasession.js
```diff
@@ -49,121 +30,96 @@ define(['playbackManager', 'nowPlayingHelper', 'events', 'connectionManager'], f
         return null;
     }
 
-    function imageUrl(item, options) {
-
-        options = options || {};
-        options.type = options.type || "Primary";
```
JustAMan, line 55: same, what if `options.type` is unset? before it defaulted it to `Primary` but now it would remain unset

src/components/playback/mediasession.js
```diff
@@ -49,121 +30,96 @@ define(['playbackManager', 'nowPlayingHelper', 'events', 'connectionManager'], f
...
+
             return connectionManager.getApiClient(item.ServerId).getScaledImageUrl(item.AlbumId, options);
         }
 
         return null;
     }
 
-    function pushImageUrl(item, imageOptions, list) {
-        var url = seriesImageUrl(item, imageOptions) || imageUrl(item, imageOptions);
+    function pushImageUrl(item, imageOptions = {}) {
```
JustAMan, line 47: the name now looks wrong

src/components/playback/mediasession.js
```diff
@@ -49,121 +30,96 @@ define(['playbackManager', 'nowPlayingHelper', 'events', 'connectionManager'], f
...
 
-        var playState = state.PlayState || {};
-        var parts = nowPlayingHelper.getNowPlayingNames(item);
-        var artist = parts[parts.length - 1].text;
-        var title = parts.length === 1 ? '' : parts[0].text;
-        var albumArtist;
-
-        if (item.AlbumArtists && item.AlbumArtists[0]) {
-            albumArtist = item.AlbumArtists[0].Name;
-        }
```
JustAMan, line 134: so this is now lost?

src/components/playback/mediasession.js
```diff
@@ -49,121 +30,96 @@ define(['playbackManager', 'nowPlayingHelper', 'events', 'connectionManager'], f
...
                 title: title,
                 artist: artist,
                 album: album,
-                artwork: getImageUrls(item),
-                albumArtist: albumArtist,
-                currentTime: currentTime,
-                duration: duration,
-                paused: isPaused,
-                itemId: itemId,
-                mediaType: item.MediaType
```
JustAMan, line 157: why is this removed?

src/components/playback/mediasession.js
```diff
@@ -49,121 +30,96 @@ define(['playbackManager', 'nowPlayingHelper', 'events', 'connectionManager'], f
...
             });
         } else {
-            var imageUrl = [];
-            pushImageUrl(item, {maxHeight: 400}, imageUrl);
-
-            if (imageUrl.length) {
-                imageUrl = imageUrl[0].src;
-            } else {
-                imageUrl = null;
+            let itemImageUrl = seriesImageUrl(item) || imageUrl(item);
```
JustAMan, line 120: why `maxHeight` is gone?

src/components/playback/mediasession.js
```diff
@@ -49,121 +30,96 @@ define(['playbackManager', 'nowPlayingHelper', 'events', 'connectionManager'], f
...
+            if (!itemImageUrl) {
+                itemImageUrl = null;
```
JustAMan, line 122: cannot you do stuff like
```js
let itemImageUrl = seriesImageUrl(item) || imageUrl(item) || null;
```
and remove this `if`?

src/components/playback/mediasession.js
```diff
@@ -228,8 +172,6 @@ define(['playbackManager', 'nowPlayingHelper', 'events', 'connectionManager'], f
     }
 
     function hideMediaControls() {
-        lastUpdateTime = 0;
```
JustAMan, line 231: hmm, you sure?..

src/components/playback/mediasession.js
```diff
@@ -261,8 +202,8 @@ define(['playbackManager', 'nowPlayingHelper', 'events', 'connectionManager'], f
     function execute(name) {
         playbackManager[name](currentPlayer);
     }
-    if (navigator.mediaSession) {
 
+    if (navigator.mediaSession) {
```
JustAMan, line 206: why're you checking it differently? in code above you used `if 'mediaSession' in navigator`

src/components/playback/mediasession.js
```diff
@@ -1,45 +1,26 @@
...
-
-                return connectionManager.getApiClient(item.ServerId).getScaledImageUrl(item.SeriesId, options);
-            }
-        }
-
-        if (options.type === 'Thumb') {
+        } else if (type === 'Primary' && item.SeriesPrimaryImageTag) {
+            options.tag = item.SeriesPrimaryImageTag;
 
+            return connectionManager.getApiClient(item.ServerId).getScaledImageUrl(item.SeriesId, options);
```
ferferga, line 17: This line breaks

src/components/playback/mediasession.js
ferferga: @MrTimscampi I think this might be related to my review

src/components/playback/mediasession.js
```diff
@@ -1,45 +1,26 @@
...
     // Reports media playback to the device for lock screen control
 
-    var currentPlayer;
-    var lastUpdateTime = 0;
-
-    function seriesImageUrl(item, options) {
+    let currentPlayer;
 
+    function seriesImageUrl(item, options = {}, type = options.type || 'Primary') {
         if (item.Type !== 'Episode') {
```
ferferga, line 12:
```suggestion
        options = options || {};
        options.type = options.type || "Primary";
        if (item.Type !== 'Episode') {
```

Adding this back fixes what I commented in my review, as pointed out by @JustAMan.

MrTimscampi: The default for "options" is set in line 11.

The optional "type" argument, which is new, default to either the value of options.type or "Primary", as defaults for arguments have access to all of the previous argument's values.

MrTimscampi: Oh, I see what you mean.

src/components/playback/mediasession.js
MrTimscampi: It never worked, since it's only used in the MediaSessions object that is passed to the browser and, afaik, that has never supported album artists or some of the other properties set before...

src/components/playback/mediasession.js
MrTimscampi: Because the MediaMetadata object never supported them in the first place.

See here: https://developer.mozilla.org/en-US/docs/Web/API/MediaMetadata/MediaMetadata

src/components/playback/mediasession.js
MrTimscampi: This PR originated from @dkanada noticing that artwork on Android was VERY low resolution, most notably the album art on the lockscreen.

Since I couldn't find any recommendation from Google on which resolution to use, I removed the limit.

src/components/playback/mediasession.js
MrTimscampi: I'm not sure if it would work. And in any case, it'd probably make more sense to return `null` from seriesImageUrl and imageUrl.

src/components/playback/mediasession.js
MrTimscampi: It's unused, according to ESLint...

src/components/playback/mediasession.js
MrTimscampi: This was an original line that I moved down. I'll change the check too, for consistency.

src/components/playback/mediasession.js
ferferga: I think it makes sense to leave it at full res, so each device decides better how to fit it. It's only one image as well.

src/components/playback/mediasession.js
```diff
@@ -54,14 +58,16 @@ import connectionManager from 'connectionManager';
                 src: url,
                 sizes: height + 'x' + height
             };
+        } else {
+            return null;
```
JustAMan, line 62: won't this break later on when the result of `getImageUrls()` containing `null`s would be used by something else?

src/components/playback/mediasession.js
JustAMan: Maybe we should limit it to FullHD or smth? I'm imagining the 8K artwork being transferred to the phone over not-so-fast 3G...

src/components/playback/mediasession.js
MrTimscampi: It would already break since nothing is currently returned.

I added a explicit check to omit pushing the image to the array if it's null.

src/components/playback/mediasession.js
MrTimscampi: Apparently, the recommended size for Spotify is 3000x3000.

However, there is not resolution limit, the only real limit being a size under 4MB.

I think limiting the max height to 3000px should work well enough.

MrTimscampi:
@JustAMan Can't respond to your comment on line 120, but I looked a bit into it and the recommendation for Spotify seems to be 3000x3000.

We could set it to a maxHeight of 3000px.

JustAMan:
@MrTimscampi can you do so then?
@ferferga can you re-review?..

src/components/playback/mediasession.js
```diff
@@ -1,45 +1,28 @@
-define(['playbackManager', 'nowPlayingHelper', 'events', 'connectionManager'], function (playbackManager, nowPlayingHelper, events, connectionManager) {
-    "use strict";
-
-    // no support for mediaSession
-    if (!navigator.mediaSession && !window.NativeShell) {
```
ferferga, line 5: Why this is removed?

src/components/playback/mediasession.js
```diff
@@ -49,122 +32,101 @@ define(['playbackManager', 'nowPlayingHelper', 'events', 'connectionManager'], f
...
+        const duration = parseInt(item.RunTimeTicks ? (item.RunTimeTicks / 10000) : 0);
+        const currentTime = parseInt(playState.PositionTicks ? (playState.PositionTicks / 10000) : 0);
 
-        var isPaused = playState.IsPaused || false;
-        var canSeek = playState.CanSeek || false;
+        const isPaused = playState.IsPaused || false;
+        const canSeek = playState.CanSeek || false;
 
-        if (navigator.mediaSession) {
+        if ('mediaSession' in navigator) {
```
ferferga, line 121: I would remove all the checking here and just leave the one done at the beginning

src/components/playback/mediasession.js
```diff
@@ -228,17 +178,14 @@ define(['playbackManager', 'nowPlayingHelper', 'events', 'connectionManager'], f
     }
 
     function hideMediaControls() {
-        lastUpdateTime = 0;
-
-        if (navigator.mediaSession) {
+        if ('mediaSession' in navigator) {
```
ferferga, line 181: Same here

src/components/playback/mediasession.js
```diff
@@ -261,8 +208,8 @@ define(['playbackManager', 'nowPlayingHelper', 'events', 'connectionManager'], f
     function execute(name) {
         playbackManager[name](currentPlayer);
     }
-    if (navigator.mediaSession) {
 
+    if ('mediaSession' in navigator) {
```
ferferga, line 212: same

src/components/playback/mediasession.js
MrTimscampi: They have different purposes.

The one at the beginning is to check if either the Media Session API or Native Shell is available.  
This one is specifically to check if the Media Session API is available, since a client without Media Session but with Native Shell will load this file.

MrTimscampi:
@ferferga Since you can't return from an ES6 module, I moved the original check to the only loading point I could find.

As for the other check, as explained, they're there in case the client has NativeShell but doesn't support the Media Session API.

ferferga:
@MrTimscampi This is exactly what I was thinking!

Checking out this branch to (hopefully now) say that it's good to go! :)

PR #1091, opened 2020-04-15
Migrated HtmlVideoPlayer

cromefire:
This has to be reviewed closely

**Changes**
Migrated:
 - src/components/htmlvideoplay/plugin.js / src/plugins/htmlVideoPlayer/plugin.js

**Issues**
Relates to: #868


cromefire:
It's not tested yet

manzajaro:
I think this doesn't work in the old browser. For example in LG TV

src/components/htmlvideoplayer/plugin.js
```diff
@@ -147,99 +161,100 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
-        var lastCustomTrackMs = 0;
-        var currentClock;
-        var currentSubtitlesOctopus;
-        var currentAssRenderer;
-        var customTrackIndex = -1;
-
-        var showTrackOffset;
-        var currentTrackOffset;
+    /**
+     * Private:
```
MrTimscampi, line 181: Add a [private tag](https://jsdoc.app/tags-private.html) to JSDoc comments instead of doing this.

src/components/htmlvideoplayer/plugin.js
```diff
@@ -686,27 +711,30 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
-            var audioIndex = -1;
-            var i;
-            var length;
-            var stream;
+            let audioIndex = -1;
+            let i;
+            let length;
+            let stream;
 
             for (i = 0, length = streams.length; i < length; i++) {
```
MrTimscampi, line 739: Replace with `forEach()`

src/components/htmlvideoplayer/plugin.js
```diff
@@ -942,129 +1004,141 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
 
-        function renderSsaAss(videoElement, track, item) {
-            var attachments = self._currentPlayOptions.mediaSource.MediaAttachments || [];
-            var apiClient = connectionManager.getApiClient(item);
-            var options = {
+        /**
+         * @private
+         */
+        renderSsaAss(videoElement, track, item) {
+            const attachments = self._currentPlayOptions.mediaSource.MediaAttachments || [];
```
MrTimscampi, line 1126: Shouldn't `self` be `this` here? (To reference the class instance instead of `window.self`).

src/components/htmlvideoplayer/plugin.js
```diff
@@ -1231,15 +1323,15 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
+            const trackEvents = this.currentTrackEvents;
+            const subtitleTextElement = this.videoSubtitlesElem;
 
             if (trackEvents && subtitleTextElement) {
-                var ticks = timeMs * 10000;
-                var selectedTrackEvent;
-                for (var i = 0; i < trackEvents.length; i++) {
+                const ticks = timeMs * 10000;
+                let selectedTrackEvent;
+                for (let i = 0; i < trackEvents.length; i++) {
```
MrTimscampi, line 1332: Replace with `forEach()`.

MrTimscampi:
> I think this doesn't work in the old browser. For example in LG TV

We convert the recent syntax to something compatible with old LG TVs (and other clients) when building the web client.

it allows us to use very recent JavaScript syntax to make our lives easier while retaining compatibility with all the clients we support.

src/components/htmlvideoplayer/plugin.js
```diff
@@ -1,11 +1,24 @@
-define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackManager', 'appRouter', 'appSettings', 'connectionManager', 'htmlMediaHelper', 'itemHelper', 'fullscreenManager', 'globalize'], function (browser, require, events, appHost, loading, dom, playbackManager, appRouter, appSettings, connectionManager, htmlMediaHelper, itemHelper, fullscreenManager, globalize) {
-    "use strict";
+import browser from "browser";
```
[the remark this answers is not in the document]
cromefire: 😟

src/components/htmlvideoplayer/plugin.js
cromefire: That sady doesn't work because I could only declare the whole class private as fast as I figured, it'll be better with ts

src/components/htmlvideoplayer/plugin.js
cromefire: Will do, but couldn't that also just be for of? Makes it a bit more readable

src/components/htmlvideoplayer/plugin.js
cromefire: Totally missed that one

cromefire:
> This breaks the entire client with the following:

> ```
> TypeError: pluginFactory is not a constructor pluginManager.js:37:29
> ```

Yeah I think I have to do something like this:
```
export default () => new HtmlVideoPlayer();
```

cromefire:
I'll continue work on this today probably

src/components/htmlvideoplayer/plugin.js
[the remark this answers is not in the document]
cromefire: Also there should be a lint rule in place for that, it's totally mixed right now so I guess everyone just takes what's more reachable on their keyboard

src/components/htmlvideoplayer/plugin.js
cromefire: Sales as above

src/components/htmlvideoplayer/plugin.js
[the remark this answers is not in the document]
MrTimscampi: The lint rule is awaiting a merge in #1069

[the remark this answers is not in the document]
cromefire: Yeah it's just hard to rewrite it with out a rules or reformatting it

cromefire:
Is it possible that the amd plugin doesn't work correctly with default exports? because it exports
```typescript
{
    default: HtmlVideoPlayer()
}
```
instead of
```
HtmlVideoPlayer()
```

src/components/htmlvideoplayer/plugin.js
```diff
@@ -722,197 +746,230 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
-            if (!self._started) {
-                self._started = true;
-                this.removeAttribute('controls');
+        /**
+         * @private
+         */
+        onPlaying(e) {
+            if (!this._started) {
+                this._started = true;
+                this.onPlaying.removeAttribute('controls');
```
cromefire, line 896: How did this ever work?

src/components/htmlvideoplayer/plugin.js
```diff
@@ -722,197 +746,230 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
-                self.setAudioStreamIndex(audioTrackIndexToSetOnPlaying);
+            if (this.audioTrackIndexToSetOnPlaying != null && this.canSetAudioStreamIndex()) {
+                this.setAudioStreamIndex(this.audioTrackIndexToSetOnPlaying);
             }
         }
 
-        function onPlaying(e) {
-            if (!self._started) {
-                self._started = true;
-                this.removeAttribute('controls');
```
cromefire, line 856: How did this ever work?

src/components/htmlvideoplayer/plugin.js
MrTimscampi: I'm figuring out a way to use the Class Properties and Private Class Members proposals in code (both supported by Babel, but ESLint is tripping on them with our current config)

cromefire: Well yeah that why I left them out for now, once it works they can easily be declared as such

src/components/htmlvideoplayer/plugin.js
cromefire: Apparently by not binding stuff...

cromefire:
There are still errors in there I haven't found I hope somebody can find them

src/components/pluginManager.js
```diff
@@ -34,7 +34,8 @@ define(['events'], function (events) {
 
             require([url, 'globalize', 'appRouter'], function (pluginFactory, globalize, appRouter) {
 
-                var plugin = new pluginFactory();
+                // Because of es6 default exports
+                var plugin = pluginFactory.default ? new pluginFactory.default() : new pluginFactory();
```
dmitrylyzo, line 38: maybe a little trick?
```js
pluginFactory = pluginFactory.default || pluginFactory;
```

src/components/htmlvideoplayer/plugin.js
```diff
@@ -722,199 +746,234 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
+        }
 
-            destroyCustomTrack(this);
-            htmlMediaHelper.onEndedInternal(self, this, onError);
+        /**
+         * @private
+         */
+        onEnded() {
+            this.destroyCustomTrack(this);
+            htmlMediaHelper.onEndedInternal(this, this.onEnded.bind(this), this.onError.bind(this));
```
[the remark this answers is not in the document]
cromefire: Well can't really help it, this is how js works

src/components/htmlvideoplayer/plugin.js
```diff
@@ -722,199 +746,234 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
+
+                this.destroyCustomTrack(videoElement);
+                videoElement.removeEventListener('timeupdate', this.onTimeUpdate.bind(this));
+                videoElement.removeEventListener('ended', this.onEnded.bind(this));
+                videoElement.removeEventListener('volumechange', this.onVolumeChange.bind(this));
+                videoElement.removeEventListener('pause', this.onPause.bind(this));
+                videoElement.removeEventListener('playing', this.onPlaying.bind(this));
+                videoElement.removeEventListener('play', this.onPlay.bind(this));
+                videoElement.removeEventListener('click', this.onClick.bind(this));
+                videoElement.removeEventListener('dblclick', this.onDblClick.bind(this));
```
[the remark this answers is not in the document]
cromefire: Binding in the constructor isn't really considered a good thing to do in js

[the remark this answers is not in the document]
cromefire: Once it's ts something like bind-decorator can be used

src/components/htmlvideoplayer/plugin.js
```diff
@@ -722,199 +746,234 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
 
-        function onError() {
-            var errorCode = this.error ? (this.error.code || 0) : 0;
-            var errorMessage = this.error ? (this.error.message || '') : '';
+        /**
+         * @private
+         */
+        onError() {
+            const errorCode = this.onError.error ? (this.onError.error.code || 0) : 0;
+            const errorMessage = this.onError.error ? (this.onError.error.message || "") : "";
```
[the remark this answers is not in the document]
cromefire: This is so messy... Probably wasn't properly bound before...

src/components/pluginManager.js
cromefire: Well kinda does the same so I choose that because it looked more concise

src/components/htmlvideoplayer/plugin.js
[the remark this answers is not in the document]
dmitrylyzo: I just blind :smile: `onEnded` and `onError`.
But https://github.com/jellyfin/jellyfin-web/blob/24760a1158c9575a2a96a7e65aabac12d217dbb9/src/components/htmlMediaHelper.js#L406
so it should be
```js
            htmlMediaHelper.onEndedInternal(this, this._mediaElement, this.onError.bind(this));
```

src/components/htmlvideoplayer/plugin.js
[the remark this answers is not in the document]
dmitrylyzo: Are you sure that `this.onTimeUpdate.bind(this)` returns same object when `addEventListener` and here?
Otherwise, this won't work.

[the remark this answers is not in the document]
cromefire: Yes it returns the function, just that `this` in the function it the object it comes from (in this case `this`)

src/components/htmlvideoplayer/plugin.js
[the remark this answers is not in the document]
cromefire: Probably. I wish gh had a todo feature...

src/components/pluginManager.js
dmitrylyzo: In my case, you will get separate hack/fix/polyfill and leave other code untouched.

src/components/htmlvideoplayer/plugin.js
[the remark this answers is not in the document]
dmitrylyzo: Semantically the same, but with different reference.

src/components/htmlvideoplayer/plugin.js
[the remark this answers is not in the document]
dmitrylyzo: If I am right, in original `this` was an element. But it is really a mess.

https://github.com/jellyfin/jellyfin-web/blob/24760a1158c9575a2a96a7e65aabac12d217dbb9/src/components/htmlvideoplayer/plugin.js#L340
https://github.com/jellyfin/jellyfin-web/blob/24760a1158c9575a2a96a7e65aabac12d217dbb9/src/components/htmlMediaHelper.js#L317
https://github.com/jellyfin/jellyfin-web/blob/24760a1158c9575a2a96a7e65aabac12d217dbb9/src/components/htmlMediaHelper.js#L244
https://github.com/jellyfin/jellyfin-web/blob/24760a1158c9575a2a96a7e65aabac12d217dbb9/src/components/htmlMediaHelper.js#L228
You have changed semantic: `this` was an element, now it is a plugin.
`this._mediaElement.error` may help but I don't like how it look.

src/components/htmlvideoplayer/plugin.js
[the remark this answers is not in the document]
cromefire: Well I don't this performance is a real concern right now as it's still loading individual js files

src/components/htmlvideoplayer/plugin.js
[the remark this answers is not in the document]
cromefire: It's hard to do those things without ts right now I have to look into that further they extensive used the non binding way and I missed it

src/components/pluginManager.js
cromefire: Well after terser I'll probably the same anyway I just find mine more readable and that's exactly what we need right now

src/components/htmlvideoplayer/plugin.js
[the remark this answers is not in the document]
dmitrylyzo: This is not a question of performance. You add one object as listener, but remove another object, and first one will still be there.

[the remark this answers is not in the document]
cromefire: Hmm that an interesting question, it might be the same object but I'll have to test it

MrTimscampi:
Seems like none of the durations in the OSD work anymore and neither does the seekbar:

image

cromefire:
Yeah that's why I said there are error in there

dmitrylyzo:
I feel that event handling is broken: `this` was a video element, and now it is `HtmlVideoPlayer` instance. So all this._some\_property_ goes wrong way.

MrTimscampi:
> I feel that event handling is broken: `this` was a video element, and now it is `HtmlVideoPlayer` instance. So all this._some_property_ goes wrong way.

Agreed, this would explain why most of the events (time updates, seeing etc) are broken.

cromefire:
Yeah I'll look into that was a lot of magic happening in there

cromefire:
It's probably because they didn't properly bind it

.editorconfig
```diff
@@ -7,3 +7,6 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 end_of_line = lf
+
+[package.json]
```
dkanada, line 11: I'm fine requiring this for all `json` files.

cromefire: Yeah I has just preventing my ide from making a mess

MrTimscampi:
@cromefire Any plan to fix the remaining event issues in this?

cromefire:
Yeah but I hadn't time for it yet, I probably go the old approach, but documenting it

cromefire:
Blocked on class properties

src/components/htmlVideoPlayer/plugin.js
```diff
@@ -348,70 +356,69 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
             });
         }
 
-        function onShakaError(event) {
-
-            var error = event.detail;
+        /**
+         * @private
+         */
+        onShakaError = (event) => {
```
MrTimscampi, line 372:
```suggestion
        onShakaError(event) {
```

cromefire: This has to bet there

cromefire: I totally now this kills the build and eslint at the time, but without it that whole PR doesn't work, that's why I said I'd be good to have class properties supported

MrTimscampi: @cromefire Now that they are, any plans to come back to this PR and fix it up?

cromefire: Yeah probably ping me if I should forget it

cromefire:
See new (not broken) version: #1618

cromefire:
It's still failing on the class properties

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -728,208 +751,266 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
-            var errorMessage = this.error ? (this.error.message || '') : '';
+        /**
+         * @private
+         * @param e {Event} The event received from the `<video>` element
+         */
+        onError = (e) => {
+            /**
+             * @type {HTMLMediaElement}
+             */
+            const elem = e.target
```
Camc314, line 1008:
```suggestion
            const elem = e.target;
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -728,208 +751,266 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
+            const errorCode = elem.error ? (elem.error.code || 0) : 0;
+            const errorMessage = elem.error ? (elem.error.message || "") : "";
```
Camc314, line 1010:
```suggestion
            const errorMessage = elem.error ? (elem.error.message || '') : '';
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -959,130 +1040,140 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
 
-            incrementFetchQueue();
-            return new Promise(function (resolve, reject) {
-                var xhr = new XMLHttpRequest();
+            this.incrementFetchQueue();
+            return new Promise((resolve, reject) => {
+                const xhr = new XMLHttpRequest();
 
-                var url = getTextTrackUrl(track, item, '.js');
+                const url = getTextTrackUrl(track, item, ".js");
```
Camc314, line 1114:
```suggestion
                const url = getTextTrackUrl(track, item, '.js');
````

File: src/plugins/htmlVideoPlayer/plugin.js [hunk 17]
```diff
@@ -959,130 +1040,140 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
-                    htmlMediaHelper.onErrorInternal(self, 'mediadecodeerror');
+                onError() {
+                    htmlMediaHelper.onErrorInternal(htmlVideoPlayer, 'mediadecodeerror');
                 },
-                timeOffset: (self._currentPlayOptions.transcodingOffsetTicks || 0) / 10000000,
+                timeOffset: (this._currentPlayOptions.transcodingOffsetTicks || 0) / 10000000,
 
                 // new octopus options; override all, even defaults
-                renderMode: 'blend',
+                renderMode: "blend",
```
Camc314, line 1176:
```suggestion
                renderMode: 'blend',
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1116,42 +1209,51 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
-            fetchSubtitles(track, item).then(function (data) {
-                if (!videoSubtitlesElem) {
-                    var subtitlesContainer = document.createElement('div');
+        /**
+         * @private
+         */
+        renderSubtitlesWithCustomElement(videoElement, track, item) {
+            this.fetchSubtitles(track, item).then((data) => {
+                if (!this.videoSubtitlesElem) {
+                    const subtitlesContainer = document.createElement("div");
```
Camc314, line 1228:
```suggestion
                    const subtitlesContainer = document.createElement('div');
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1164,40 +1266,43 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
             return html;
         }
 
-        function setCueAppearance() {
+        /**
+         * @private
+         */
+        setCueAppearance() {
+            Promise.all([import('userSettings'), import('subtitleAppearanceHelper')]).then(([userSettings, subtitleAppearanceHelper]) => {
+                const elementId = this.id + "-cuestyle";
```
Camc314, line 1274:
```suggestion
                const elementId = this.id + '-cuestyle';
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1164,40 +1266,43 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
 
-        function renderTracksEvents(videoElement, track, item) {
-
+        /**
+         * @private
+         */
+        renderTracksEvents(videoElement, track, item) {
             if (!itemHelper.isLocalItem(item) || track.IsExternal) {
-                var format = (track.Codec || '').toLowerCase();
+                const format = (track.Codec || "").toLowerCase();
```
Camc314, line 1293:
```suggestion
                const format = (track.Codec || '').toLowerCase();
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1297,28 +1404,26 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
-        function createMediaElement(options) {
-
-            return new Promise(function (resolve, reject) {
-
-                var dlg = document.querySelector('.videoPlayerContainer');
+        /**
+         * @private
+         */
+        createMediaElement(options) {
+            const dlg = document.querySelector(".videoPlayerContainer");
```
Camc314, line 1411:
```suggestion
            const dlg = document.querySelector('.videoPlayerContainer');
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1297,28 +1404,26 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
 
                 if (!dlg) {
-
-                    require(['css!./style'], function () {
-
+                    return import('css!./style').then(() => {
                         loading.show();
 
-                        var dlg = document.createElement('div');
+                        const dlg = document.createElement("div");
```
Camc314, line 1417:
```suggestion
                        const dlg = document.createElement('div');
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1297,28 +1404,26 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
 
                         dlg.classList.add('videoPlayerContainer');
 
                         if (options.fullscreen) {
                             dlg.classList.add('videoPlayerContainer-onTop');
                         }
 
-                        var html = '';
-                        var cssClass = 'htmlvideoplayer';
+                        let html = "";
```
Camc314, line 1425:
```suggestion
                        let html = '';
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1297,28 +1404,26 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
+                        let cssClass = "htmlvideoplayer";
```
Camc314, line 1426:
```suggestion
                        let cssClass = 'htmlvideoplayer';
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1335,29 +1440,29 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
                         html += '</video>';
 
                         dlg.innerHTML = html;
-                        var videoElement = dlg.querySelector('video');
+                        const videoElement = dlg.querySelector("video");
```
Camc314, line 1443:
```suggestion
                        const videoElement = dlg.querySelector('video');
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1366,56 +1471,65 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
-        var list = [];
+    /**
+     * @private
+     */
+    static getSupportedFeatures() {
+        const list = [];
 
-        var video = document.createElement('video');
-        if (video.webkitSupportsPresentationMode && typeof video.webkitSetPresentationMode === 'function' || document.pictureInPictureEnabled) {
+        const video = document.createElement("video");
```
Camc314, line 1531:
```suggestion
        const video = document.createElement('video');
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1366,56 +1471,65 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
+        if (video.webkitSupportsPresentationMode && typeof video.webkitSetPresentationMode === "function" || document.pictureInPictureEnabled) {
```
Camc314, line 1532:
```suggestion
        if (video.webkitSupportsPresentationMode && typeof video.webkitSetPresentationMode === 'function' || document.pictureInPictureEnabled) {
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1432,178 +1546,171 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
         }
 
         list.push('SetBrightness');
-        list.push('SetAspectRatio');
+        list.push("SetAspectRatio");
```
Camc314, line 1549:
```suggestion
        list.push('SetAspectRatio');
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1432,178 +1546,171 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
             this.isPip = isEnabled;
             if (isEnabled) {
                 Windows.UI.ViewManagement.ApplicationView.getForCurrentView().tryEnterViewModeAsync(Windows.UI.ViewManagement.ApplicationViewMode.compactOverlay);
             } else {
                 Windows.UI.ViewManagement.ApplicationView.getForCurrentView().tryEnterViewModeAsync(Windows.UI.ViewManagement.ApplicationViewMode.default);
             }
         } else {
-            if (video && video.webkitSupportsPresentationMode && typeof video.webkitSetPresentationMode === 'function') {
-                video.webkitSetPresentationMode(isEnabled ? 'picture-in-picture' : 'inline');
+            if (video && video.webkitSupportsPresentationMode && typeof video.webkitSetPresentationMode === "function") {
```
Camc314, line 1630:
```suggestion
            if (video && video.webkitSupportsPresentationMode && typeof video.webkitSetPresentationMode === 'function') {
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1432,178 +1546,171 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
+                video.webkitSetPresentationMode(isEnabled ? "picture-in-picture" : "inline");
```
Camc314, line 1631:
```suggestion
                video.webkitSetPresentationMode(isEnabled ? 'picture-in-picture' : 'inline');
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1432,178 +1546,171 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
         if (document.pictureInPictureEnabled) {
             return document.pictureInPictureElement ? true : false;
         } else if (window.Windows) {
             return this.isPip || false;
         } else {
-            var video = this._mediaElement;
+            const video = this._mediaElement;
             if (video) {
-                return video.webkitPresentationMode === 'picture-in-picture';
+                return video.webkitPresentationMode === "picture-in-picture";
```
Camc314, line 1644:
```suggestion
                return video.webkitPresentationMode === 'picture-in-picture';
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1432,178 +1546,171 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
-        var video = this._mediaElement;
+    setAirPlayEnabled(isEnabled) {
+        const video = this._mediaElement;
 
         if (document.AirPlayEnabled) {
             if (video) {
                 if (isEnabled) {
                     video.requestAirPlay().catch(function(err) {
-                        console.error('Error requesting AirPlay', err);
+                        console.error("Error requesting AirPlay", err);
```
Camc314, line 1666:
```suggestion
                        console.error('Error requesting AirPlay', err);
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1432,178 +1546,171 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
                     });
                 } else {
                     document.exitAirPLay().catch(function(err) {
-                        console.error('Error exiting AirPlay', err);
+                        console.error("Error exiting AirPlay", err);
```
Camc314, line 1670:
```suggestion
                        console.error('Error exiting AirPlay', err);
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1432,178 +1546,171 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
         if (elem) {
             val = Math.max(0, val);
             val = Math.min(100, val);
 
-            var rawValue = val;
+            let rawValue = val;
             rawValue = Math.max(20, rawValue);
 
-            var cssValue = rawValue >= 100 ? 'none' : (rawValue / 100);
+            const cssValue = rawValue >= 100 ? "none" : (rawValue / 100);
```
Camc314, line 1689:
```suggestion
            const cssValue = rawValue >= 100 ? 'none' : (rawValue / 100);
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1617,167 +1724,165 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
+    }
 
-    HtmlVideoPlayer.prototype.setPlaybackRate = function (value) {
-        var mediaElement = this._mediaElement;
+    setPlaybackRate(value) {
+        const mediaElement = this._mediaElement;
         if (mediaElement) {
             mediaElement.playbackRate = value;
         }
     };
```
Camc314, line 1765:
```suggestion
    }
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1617,167 +1724,165 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
 
-    HtmlVideoPlayer.prototype.getPlaybackRate = function () {
-        var mediaElement = this._mediaElement;
+    getPlaybackRate() {
+        const mediaElement = this._mediaElement;
         if (mediaElement) {
             return mediaElement.playbackRate;
         }
         return null;
     };
```
Camc314, line 1773:
```suggestion
    }
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1617,167 +1724,165 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
+    }
 
-    HtmlVideoPlayer.prototype.setAspectRatio = function (val) {
-        var mediaElement = this._mediaElement;
+    setAspectRatio(val) {
+        const mediaElement = this._mediaElement;
         if (mediaElement) {
-            if ('auto' === val) {
-                mediaElement.style.removeProperty('object-fit');
+            if ("auto" === val) {
```
Camc314, line 1815:
```suggestion
            if ('auto' === val) {
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1617,167 +1724,165 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
+                mediaElement.style.removeProperty("object-fit");
```
Camc314, line 1816:
```suggestion
                mediaElement.style.removeProperty('object-fit');
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1617,167 +1724,165 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
             } else {
-                mediaElement.style['object-fit'] = val;
+                mediaElement.style["object-fit"] = val;
```
Camc314, line 1818:
```suggestion
                mediaElement.style['object-fit'] = val;
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1617,167 +1724,165 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
         }
         this._currentAspectRatio = val;
-    };
+    }
 
-    HtmlVideoPlayer.prototype.getAspectRatio = function () {
-        return this._currentAspectRatio || 'auto';
-    };
+    getAspectRatio() {
+        return this._currentAspectRatio || "auto";
```
Camc314, line 1825:
```suggestion
        return this._currentAspectRatio || 'auto';
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1617,167 +1724,165 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
+    }
 
-    HtmlVideoPlayer.prototype.getSupportedAspectRatios = function () {
+    getSupportedAspectRatios() {
         return [{
-            name: 'Auto',
-            id: 'auto'
+            name: "Auto",
```
Camc314, line 1830:
```suggestion
            name: 'Auto',
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1617,167 +1724,165 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
+            id: "auto"
```
Camc314, line 1831:
```suggestion
            id: 'auto'
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1617,167 +1724,165 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
         }, {
-            name: 'Cover',
-            id: 'cover'
+            name: "Cover",
```
Camc314, line 1833:
```suggestion
            name: 'Cover',
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1617,167 +1724,165 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
+            id: "cover"
         }, {
-            name: 'Fill',
-            id: 'fill'
+            name: "Fill",
```
Camc314, line 1836:
```suggestion
            name: 'Fill',
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1617,167 +1724,165 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
         return [{
-            name: 'Auto',
-            id: 'auto'
+            name: "Auto",
+            id: "auto"
         }, {
-            name: 'Cover',
-            id: 'cover'
+            name: "Cover",
+            id: "cover"
```
Camc314, line 1834:
```suggestion
            id: 'cover'
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1617,167 +1724,165 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
         }, {
-            name: 'Fill',
-            id: 'fill'
+            name: "Fill",
+            id: "fill"
```
Camc314, line 1837:
```suggestion
            id: 'fill'
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1617,167 +1724,165 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
             return Promise.resolve({
                 categories: categories
             });
         }
 
-        var mediaCategory = {
+        const mediaCategory = {
             stats: [],
-            type: 'media'
+            type: "media"
```
Camc314, line 1872:
```suggestion
            type: 'media'
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1617,167 +1724,165 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
         };
         categories.push(mediaCategory);
 
         if (playOptions.url) {
             //  create an anchor element (note: no need to append this element to the document)
-            var link = document.createElement('a');
+            let link = document.createElement("a");
```
Camc314, line 1878:
```suggestion
            let link = document.createElement('a');
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1617,167 +1724,165 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
             //  set href to any path
             link.setAttribute('href', playOptions.url);
-            var protocol = (link.protocol || '').replace(':', '');
+            const protocol = (link.protocol || "").replace(":", "");
```
Camc314, line 1881:
```suggestion
            const protocol = (link.protocol || '').replace(':', '');
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1617,167 +1724,165 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
 
             if (protocol) {
                 mediaCategory.stats.push({
-                    label: globalize.translate('LabelProtocol'),
+                    label: globalize.translate("LabelProtocol"),
```
Camc314, line 1885:
```suggestion
                    label: globalize.translate('LabelProtocol'),
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1787,30 +1892,30 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
 
         if (this._hlsPlayer || this._shakaPlayer) {
             mediaCategory.stats.push({
-                label: globalize.translate('LabelStreamType'),
+                label: globalize.translate("LabelStreamType"),
```
Camc314, line 1895:
```suggestion
                label: globalize.translate('LabelStreamType'),
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1787,30 +1892,30 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
                 value: 'HLS'
             });
         } else {
             mediaCategory.stats.push({
-                label: globalize.translate('LabelStreamType'),
+                label: globalize.translate("LabelStreamType"),
```
Camc314, line 1900:
```suggestion
                label: globalize.translate('LabelStreamType'),
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1787,30 +1892,30 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
                 value: 'Video'
             });
         }
 
-        var videoCategory = {
+        const videoCategory = {
             stats: [],
-            type: 'video'
+            type: "video"
```
Camc314, line 1907:
``` suggestion
            type: 'video'
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1787,30 +1892,30 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
-        var width = parseInt(rect.width);
+        const rect = mediaElement.getBoundingClientRect ? mediaElement.getBoundingClientRect() : {};
+        let height = parseInt(rect.height);
+        let width = parseInt(rect.width);
 
         // Don't show player dimensions on smart TVs because the app UI could be lower resolution than the video and this causes users to think there is a problem
         if (width && height && !browser.tv) {
             videoCategory.stats.push({
-                label: globalize.translate('LabelPlayerDimensions'),
+                label: globalize.translate("LabelPlayerDimensions"),
```
Camc314, line 1918:
```suggestion
                label: globalize.translate('LabelPlayerDimensions'),
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1820,34 +1925,34 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
 
         if (width && height) {
             videoCategory.stats.push({
-                label: globalize.translate('LabelVideoResolution'),
+                label: globalize.translate("LabelVideoResolution"),
```
Camc314, line 1928:
```suggestion
                label: globalize.translate('LabelVideoResolution'),
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1820,34 +1925,34 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
 
         if (mediaElement.getVideoPlaybackQuality) {
-            var playbackQuality = mediaElement.getVideoPlaybackQuality();
+            const playbackQuality = mediaElement.getVideoPlaybackQuality();
 
-            var droppedVideoFrames = playbackQuality.droppedVideoFrames || 0;
+            const droppedVideoFrames = playbackQuality.droppedVideoFrames || 0;
             videoCategory.stats.push({
-                label: globalize.translate('LabelDroppedFrames'),
+                label: globalize.translate("LabelDroppedFrames"),
```
Camc314, line 1938:
```suggestion
                label: globalize.translate('LabelDroppedFrames'),
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1820,34 +1925,34 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
                 value: droppedVideoFrames
             });
 
-            var corruptedVideoFrames = playbackQuality.corruptedVideoFrames || 0;
+            const corruptedVideoFrames = playbackQuality.corruptedVideoFrames || 0;
```
Camc314, line 1942:
```suggestion
                label: globalize.translate('LabelCorruptedFrames'),
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1820,34 +1925,34 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
+                label: globalize.translate("LabelCorruptedFrames"),
                 value: corruptedVideoFrames
             });
         }
 
-        var audioCategory = {
+        const audioCategory = {
             stats: [],
-            type: 'audio'
+            type: "audio"
```
Camc314, line 1951:
```suggestion
            type: 'audio'
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1,11 +1,24 @@
-define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackManager', 'appRouter', 'appSettings', 'connectionManager', 'htmlMediaHelper', 'itemHelper', 'screenfull', 'globalize'], function (browser, require, events, appHost, loading, dom, playbackManager, appRouter, appSettings, connectionManager, htmlMediaHelper, itemHelper, screenfull, globalize) {
-    'use strict';
+import browser from "browser";
```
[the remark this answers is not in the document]
cromefire: I'll fix that later now that eslint works again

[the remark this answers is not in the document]
Camc314: If you can fix lint with eslint, I can suggest the 2 changes to fix the rest

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -728,208 +751,266 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
                 loading.hide();
 
-                htmlMediaHelper.seekOnPlaybackStart(self, e.target, self._currentPlayOptions.playerStartPositionTicks, function () {
-                    if (currentSubtitlesOctopus) {
-                        currentSubtitlesOctopus.timeOffset = (self._currentPlayOptions.transcodingOffsetTicks || 0) / 10000000 + currentTrackOffset;
-                        currentSubtitlesOctopus.resize();
-                        currentSubtitlesOctopus.resetRenderAheadCache(false);
+                htmlMediaHelper.seekOnPlaybackStart(this, e.target, this._currentPlayOptions.playerStartPositionTicks, function () {
+                    if (this.currentSubtitlesOctopus) {
+                        this.currentSubtitlesOctopus.timeOffset = (this._currentPlayOptions.transcodingOffsetTicks || 0) / 10000000 + this.currentTrackOffset;
```
Camc314, line 929:
```suggestion
                        this.currentSubtitlesOctopus.timeOffset = (this._currentPlayOptions.transcodingOffsetTicks || 0) / 10000000 + this.currentTrackOffset();
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1093,13 +1184,15 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
                 resizeVariation: 0.2,
                 renderAhead: 90
             };
-            require(['JavascriptSubtitlesOctopus'], function(SubtitlesOctopus) {
-                currentSubtitlesOctopus = new SubtitlesOctopus(options);
+            import('JavascriptSubtitlesOctopus').then(SubtitlesOctopus => {
```
Camc314, line 1187:
```suggestion
            import('JavascriptSubtitlesOctopus').then(({default: SubtitlesOctopus}) => {
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -728,208 +751,266 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
 
                 loading.hide();
 
-                htmlMediaHelper.seekOnPlaybackStart(self, e.target, self._currentPlayOptions.playerStartPositionTicks, function () {
-                    if (currentSubtitlesOctopus) {
-                        currentSubtitlesOctopus.timeOffset = (self._currentPlayOptions.transcodingOffsetTicks || 0) / 10000000 + currentTrackOffset;
-                        currentSubtitlesOctopus.resize();
-                        currentSubtitlesOctopus.resetRenderAheadCache(false);
+                htmlMediaHelper.seekOnPlaybackStart(this, e.target, this._currentPlayOptions.playerStartPositionTicks, function () {
+                    if (this.currentSubtitlesOctopus) {
```
Camc314, line 928:
```suggestion
                    if (this.currentSubtitlesOctopus != null ) {
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -959,127 +1040,137 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
                     }
                 }
             }
 
-            customTrackIndex = -1;
-            currentClock = null;
-            self._currentAspectRatio = null;
+            this.customTrackIndex = -1;
+            this.currentClock = null;
+            this._currentAspectRatio = null;
```
Camc314, line 1076:
```suggestion
            this.currentSubtitlesOctopus = null;
            this._currentAspectRatio = null;
```

src/plugins/htmlVideoPlayer/plugin.js
cromefire: What's this all about?

src/plugins/htmlVideoPlayer/plugin.js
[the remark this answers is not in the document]
cromefire: @Camc314 All fixed now

src/plugins/htmlVideoPlayer/plugin.js
Camc314: yeah ignore that I think i copied the wrong line

Camc314: that can be resolved

src/plugins/htmlVideoPlayer/plugin.js
cromefire: This shouldn't make a difference, because `null` is falsy

Camc314: when i tested it, you get an error because it's undefined

cromefire: But `undefined` also evaluates to `false`, so it has to be something else

Camc314: I think it is because this is an object. I think you could do `this.hasOwnProperty('currentSubtitlesOctopus')` instead

cromefire: Can't reproduce it or did you explicitly play something with aas? (which I can't test)

cromefire: I'll check that, but I need to update my server first

Camc314: Feel free to check with/without it. When I tested it, it worked with the change and didn't without

src/plugins/htmlVideoPlayer/plugin.js
Camc314: this needs to be changed

cromefire: This isn't a method:
```javascript
resetSubtitleOffset() {
    this.currentTrackOffset = 0;
    this.showTrackOffset = false;
}
```

Camc314: Oh yeah, my mistake. I've been staring at github for too long :P

src/plugins/htmlVideoPlayer/plugin.js
Camc314: From what I can tell, and hopefully someone else can provide some input, it is throwing an error because `this` is undefined, therefore trying to access a property of something that is undefined is causing the error. If `this` is checked first and then the property it works 
`                    if (this && this.currentSubtitlesOctopus) {`

Camc314: That completely fixes the PR though so hopefully some other people can help check it to make sure there's not any other errors that I have missed

cromefire: That shouldn't be possible I think my server is updating right now, so we'll see

cromefire: Works for me, can't test with ass though

cromefire: Can you confirm it with the latest commits?

Camc314: I'll do it now

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -728,208 +751,266 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
+                elem.removeAttribute('controls');
 
                 loading.hide();
 
-                htmlMediaHelper.seekOnPlaybackStart(self, e.target, self._currentPlayOptions.playerStartPositionTicks, function () {
-                    if (currentSubtitlesOctopus) {
-                        currentSubtitlesOctopus.timeOffset = (self._currentPlayOptions.transcodingOffsetTicks || 0) / 10000000 + currentTrackOffset;
-                        currentSubtitlesOctopus.resize();
-                        currentSubtitlesOctopus.resetRenderAheadCache(false);
+                htmlMediaHelper.seekOnPlaybackStart(this, e.target, this._currentPlayOptions.playerStartPositionTicks, function () {
```
Camc314, line 927:
```suggestion
                htmlMediaHelper.seekOnPlaybackStart(this, e.target, this._currentPlayOptions.playerStartPositionTicks, () => {
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -155,99 +169,100 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
-        var lastCustomTrackMs = 0;
-        var currentClock;
-        var currentSubtitlesOctopus;
-        var currentAssRenderer;
-        var customTrackIndex = -1;
-
-        var showTrackOffset;
-        var currentTrackOffset;
+    /**
+     * Private:
```
MrTimscampi, line 189: No sure this is needed, since you can use `#privateProperties` now, plus the better way to do it would be using JSDoc's `@private` tag.

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -728,208 +751,266 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
 
             // https://msdn.microsoft.com/en-us/library/hh772507(v=vs.85).aspx
 
-            var elemAudioTracks = elem.audioTracks || [];
+            const elemAudioTracks = elem.audioTracks || [];
             console.debug('found ' + elemAudioTracks.length + ' audio tracks');
 
-            for (i = 0, length = elemAudioTracks.length; i < length; i++) {
-
+            elemAudioTracks.forEach((audioTrack, i) => {
```
MrTimscampi, line 764:
```suggestion
            for (const [index, audiotTrack] of elemAudioTracks.entries()) {
```
Should work here (untested)

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -959,127 +1040,138 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
+                const xhr = new XMLHttpRequest();
 
-                var url = getTextTrackUrl(track, item, '.js');
+                const url = getTextTrackUrl(track, item, '.js');
 
                 xhr.open('GET', url, true);
 
-                xhr.onload = function (e) {
+                xhr.onload = (e) => {
                     resolve(JSON.parse(this.response));
```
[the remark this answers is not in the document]
cromefire: Okay that should be fetch anyway... I'll fix that

src/plugins/htmlVideoPlayer/plugin.js
cromefire: Yeah JSDoc didn't work with without setting those to a value but now with private properties this might work

src/plugins/htmlVideoPlayer/plugin.js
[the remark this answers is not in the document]
MrTimscampi: Yep, converting it to the Fetch API would be the best fix, if you can :)

src/plugins/htmlVideoPlayer/plugin.js
cromefire: No there'll be one closing parenthesis to much, but anyway is there a specific reason to change that? Or is it just aesthetic?

Camc314: part of #1460

src/plugins/htmlVideoPlayer/plugin.js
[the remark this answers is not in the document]
cromefire: Can you see what content it fetches? Is it javascript as it seems like?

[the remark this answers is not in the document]
cromefire: It's fetch not but I still need confirmation on the request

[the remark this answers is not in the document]
MrTimscampi: It's working now :+1:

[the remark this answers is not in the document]
cromefire: Interesting, it's still just parsing the Json 😂

cromefire:
Cool, it only took 3 Months, but this monster is finally there.
(But pleeease don't request 73 Changes at once ever again (at least not if it's autofixable lint stuff))

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -155,154 +186,253 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
+            this.#currentTime = null;
 
-            return createMediaElement(options).then(function (elem) {
+            this.resetSubtitleOffset();
 
-                return updateVideoUrl(options).then(function () {
-                    return setCurrentSrc(elem, options);
+            return this.createMediaElement(options).then(elem => {
+                return this.updateVideoUrl(options).then(() => {
+                    return this.setCurrentSrc(elem, options);
```
dmitrylyzo, line 425: This will fix forced transcoding.
```suggestion
                    return this.setCurrentSrc(this.#mediaElement, options);
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1366,55 +1572,64 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
                         // don't animate on smart tv's, too slow
                         if (options.fullscreen && browser.supportsCssAnimation() && !browser.slow) {
                             zoomIn(dlg).then(function () {
```
dmitrylyzo, line 1574: This will fix forced transcoding.
```suggestion
                            return zoomIn(dlg).then(function () {
```

src/plugins/htmlVideoPlayer/plugin.js
dmitrylyzo: This is wrong. I've found the right one :crossed_fingers:

src/plugins/htmlVideoPlayer/plugin.js
cromefire: Okay cool, will try that, I probably wouldn't have found this

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -311,27 +424,29 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
             });
         }
 
-        function setSrcWithHlsJs(instance, elem, options, url) {
-
-            return new Promise(function (resolve, reject) {
+        /**
+         * @private
+         */
+        setSrcWithHlsJs(instance, elem, options, url) {
```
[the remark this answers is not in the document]
cromefire: As in: They get their own plugin?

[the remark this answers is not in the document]
dmitrylyzo: If I'm not mistaken, Chromecast already has.
Shaka will removed in #1599

cromefire:
Is there a reason that it always pops up twice?

MrTimscampi:
> Is there a reason that it always pops up twice?

The bot takes some time to delete the old post when posting a new one. If you wait a bit, there is pretty much always only one Sonarcloud post in the end.

cromefire:
@dmitrylyzo That didn't seem to work so well, it's just loading forever for me now and if it loaded it just plays 1s and then waits 10s just to play 1s again.
On release it just there instantly and plays just fine

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -706,124 +1070,125 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
-                    decrementFetchQueue();
-                };
-
-                xhr.onerror = function (e) {
-                    reject(e);
-                    decrementFetchQueue();
-                };
+            this.incrementFetchQueue();
+            try {
+                const response = await fetch(getTextTrackUrl(track, item, '.js'));
```
dmitrylyzo, line 1142: @thornbill said that errors from `fetch` should be handled https://github.com/jellyfin/jellyfin-web/pull/1503#discussion_r460453048

Maybe something like this?
```suggestion
                const response = await fetch(getTextTrackUrl(track, item, '.js'));

                if (!response.ok) {
                    throw new Error(response.status);
                }
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -706,124 +1070,125 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
+            this.resetSubtitleOffset();
+            const item = this._currentPlayOptions.item;
 
-            destroyCustomTrack(videoElement);
-            customTrackIndex = track.Index;
-            renderTracksEvents(videoElement, track, item);
+            this.destroyCustomTrack(videoElement);
+            this.#customTrackIndex = track.Index;
+            this.renderTracksEvents(videoElement, track, item);
+            this.#lastCustomTrackMs = 0;
```
dmitrylyzo, line 1170: #1599
```suggestion
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1487,7 +1880,7 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
             link = null;
         }
 
-        if (this._hlsPlayer) {
+        if (this._hlsPlayer || this.#shakaPlayer) {
```
dmitrylyzo, line 1883: #1599
```suggestion
        if (this._hlsPlayer) {
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -113,127 +145,239 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
+        #flvPlayer;
+        /**
+         * @private (used in other files)
+         * @type {any | undefined}
+         */
+        _hlsPlayer;
+        /**
+         * @type {any | undefined}
+         */
+        #shakaPlayer;
```
dmitrylyzo, line 265: #1599
```suggestion
```

cromefire:
> Only "resurrected" dead code remains :p

That's from the merge I didn't want to spend days on it

src/plugins/htmlVideoPlayer/plugin.js
cromefire: I'll throw the whole response
```suggestion
                const response = await fetch(getTextTrackUrl(track, item, '.js'));
                
                if (!response.ok) {
                    throw new Error(response);
                }
```

cromefire:
> Browser cache?

Seems lime it, we really need webpack

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -242,117 +374,156 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
+
+            const lrd = new cast.receiver.MediaManager.LoadRequestData();
+            lrd.currentTime = (options.playerStartPositionTicks || 0) / 10000000;
+            lrd.autoplay = true;
+            lrd.media = new cast.receiver.media.MediaInformation();
+
+            lrd.media.contentId = url;
+            lrd.media.contentType = options.mimeType;
+            lrd.media.streamType = cast.receiver.media.StreamType.OTHER;
+            lrd.media.customData = options;
```
dmitrylyzo, line 429: #1599
```suggestion
```
_I had to split suggestion - GH is seeing line deletion somewhere :man_shrugging: _

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -242,117 +374,156 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
+                // This is needed in setCurrentTrackElement
+                this.#currentSrc = url;
+
+                return Promise.resolve();
+            } catch (err) {
+                console.debug(`media manager error: ${err}`);
+                return Promise.reject();
+            }
+        }
+
```
dmitrylyzo, line 443: _continue of dead code_
```suggestion
```

cromefire: already removed locally

src/plugins/htmlVideoPlayer/plugin.js
cromefire: already deleted locally

cromefire:
I'm still having some playback problems and there's some unknown `uncaught promise rejection: undefined`

cromefire:
It even sometimes crashes my tab

cromefire:
Some times I just get `pluginFactory is undefined`

cromefire:
It's also getting stuck loading again with no error:
image

dmitrylyzo:
I have no such errors in Firefox nor Chrome. Everything is fine to me.

> there's some unknown `uncaught promise rejection: undefined`

This can be when you close context menu by clicking outside.

> Some times I just get `pluginFactory is undefined`

Any steps to reproduce?

cromefire:
> Any steps to reproduce?

Nope randomly on load

> I have no such errors in Firefox nor Chrome. Everything is fine to me.

I get them in both FF and FF Dev edition but not in chrome it seems like, it's just getting stuck.

cromefire:
(Never opened it in dev edition, so it's kinda a fresh start)

dmitrylyzo:
@cromefire I got some exception on server select page with `yarn serve` and clean app storage due to typo.
`Uncaught ReferenceError: exports is not defined`
Opened PR #1664

cromefire:
Yeah got that one too, but it's unrelated

cromefire:
I'll look at it again tomorrow, but I'm pretty sure it works in release so there has to be something that's different...

cromefire:
(I'll also check master to see, if it's definitely introduced by this PR or if it's something else

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -479,204 +633,264 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
 
             // https://msdn.microsoft.com/en-us/library/hh772507(v=vs.85).aspx
 
-            var elemAudioTracks = elem.audioTracks || [];
-            console.debug('found ' + elemAudioTracks.length + ' audio tracks');
+            const elemAudioTracks = elem.audioTracks || [];
+            console.debug(`found ${elemAudioTracks.length} audio tracks`);
 
-            for (i = 0, length = elemAudioTracks.length; i < length; i++) {
+            for (const [i, audioTrack] of elemAudioTracks.entries()) {
```
dmitrylyzo, line 646: On webOS 1.2, `elemAudioTracks` (`AudioTrackList`) doesn't have `entries`.
I cannot get to here in browsers, but probably the same for them.

cromefire: That's probably a missing polyfill there

cromefire: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/entries

dmitrylyzo: The problem is that `AudioTrackList` is not `Array`.
I had to revert the same for `HTMLOptionsCollection` #1279

cromefire: Oh okay I can work with that, `Array.from()` does the job

cromefire: There you go

cromefire:
> This can be when you close context menu by clicking outside.

Is this caused by this PR? And if so, where would I be looking for it?

cromefire:
Okay I got my issue reproduced on master so I'm going to try hunt it down and add an issue for it

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -640,10 +640,13 @@ function tryRemoveElement(elem) {
 
             // https://msdn.microsoft.com/en-us/library/hh772507(v=vs.85).aspx
 
+            /**
+             * @type {HTMLOptionsCollection|any[]}
```
dmitrylyzo, line 644:
```suggestion
             * @type {AudioTrackList|any[]}
```

cromefire: That type doesn't exist

cromefire: I fixed it at `ArrayLike<any>` for now

dmitrylyzo: At least it doesn't affect the execution.
Some info https://github.com/microsoft/TypeScript/issues/38544

cromefire: Yeah it probably needs some time to tickle down, now every IDE probably already has v3.9.3 yet for now I leave it like that once it's ts it'll be correctly inferred anyway

cromefire: Oh if I read it correctly it hasn't been in closed at all yet, that makes it more difficult anyway but that's something for when ts is there

Camc314:
I get `htmlMediaHelper.js:273 HLS Error: Type: mediaError Details: bufferStalledError Fatal: false` playing for longer than a few minutes or moving aroung in time throughout the video. Did you ever get the error?

MrTimscampi:
> I get `htmlMediaHelper.js:273 HLS Error: Type: mediaError Details: bufferStalledError Fatal: false` playing for longer than a few minutes or moving aroung in time throughout the video. Did you ever get the error?

I've been getting a few non-fatal buffer errors since before 10.6.

cromefire:
Yes I couldn't ever reproduce it though

Camc314:
Ok then, if it was there before, pre 10.6, and present on master, it is probable not that big of an issue

dmitrylyzo:
> I get `htmlMediaHelper.js:273 HLS Error: Type: mediaError Details: bufferStalledError Fatal: false` playing for longer than a few minutes or moving aroung in time throughout the video. Did you ever get the error?

I was getting such errors on seeking (mostly backward) of DirectStream/Transcode, iirc
At the moment I have managed to reproduce it once.

PR #1092, opened 2020-04-15
Use screenfull to handle fullscreen switching

MrTimscampi:
**Changes**

Removes fullscreenManager and uses [screenfull](https://github.com/sindresorhus/screenfull.js/) to handle fullscreen switching and status detection.

**Issues**


src/scripts/site.js
```diff
@@ -834,8 +835,6 @@ var AppInfo = {};
         define("searchFields", [componentsPath + "/search/searchfields"], returnFirstDependency);
         define("searchResults", [componentsPath + "/search/searchresults"], returnFirstDependency);
         define("upNextDialog", [componentsPath + "/upnextdialog/upnextdialog"], returnFirstDependency);
-        define("fullscreen-doubleclick", [componentsPath + "/fullscreen/fullscreen-dc"], returnFirstDependency);
```
JustAMan, line 837: so does `screenfull` also handle double clicks?

MrTimscampi: That module didn't exist in the source.

PR #1094, opened 2020-04-15
Removed avatar.png

ferferga:
We already have it in material icons, so no reason to keep it

**Before**
before2
before
**After**
after
after2


src/controllers/auth/login.js
```diff
@@ -56,29 +56,6 @@ define(["apphost", "appSettings", "dom", "connectionManager", "loading", "layout
...
     }
 
-    var metroColors = ["#6FBD45", "#4BB3DD", "#4164A5", "#E12026", "#800080", "#E1B222", "#008040", "#0094FF", "#FF00C7", "#FF870F", "#7F0037"];
-
-    function getRandomMetroColor() {
-        var index = Math.floor(Math.random() * (metroColors.length - 1));
-        return metroColors[index];
-    }
-
-    function getMetroColor(str) {
```
JustAMan, line 66: well, that function at least gave users a different (but stable) colors... maybe we should do some colored outlines, too, instead of completely dropping this and forcing users to either upload avatars or all look exactly the same?

MrTimscampi: We could use the color generator in cardBuilder to pick a color in the theme from one of the 5 defaultBackground classes, based on the name of the user.

MrTimscampi:
@ferferga What are the plans for this?

I'm for removing the PNG, but not for removing the background color.

ferferga:
@MrTimscampi This was intended as a quick cleanup PR, but it clearly needed a little bit more of love.

As I'm still away from the computer and I have no idea when I'll have time to work on this, I'll mark this as a draft.

What's on my mind for this PR:

* Consistent background color (like in login page) for users that don't have a photo in the following places:

    * Profile page
    * Users section in dashboard
    * Activity log

* Rounded cards in:

    * Profile page
    * Login page

Of course, all of them using the material icons glyph.

I also thought about applying the color to the avatar icon itself, instead of using the card design, but I'm not sure about the consistency of this. However, I think it fits for the activity log section, so we have **blue** for activities performed by accounts and the **avatar icon** to distinguish the user that performed the action.

What do you think about all of these proposals? @jellyfin/web

dkanada:
> Regarding the login page, we have an extremely clear design for what a card should look like, and these changes break that design. The cards themselves can potentially change in the future but we should not create a new element style just for the login page.

Same opinion as before, we shouldn't create circular cards anywhere. We don't even use circular cards for people on the item details page.

PR #1095, opened 2020-04-15
Fix .ass subtitles not starting on mobile

JustAMan:
**Changes**

On mobile browsers video element got created and started "playing" when it was empty if a video was transcoded (because mobile browsers, at least Chrome and Android WebView) support HLS natively. Thus Octopus render was confused thinking it was requested to render over 0x0 video element.

I've added a callback firing when _real_ video file is ready, and in that callback I'm requesting Octopus to re-assess video element size.

**Issues**

None that I know of

### DIFF
--- a/src/components/htmlMediaHelper.js
+++ b/src/components/htmlMediaHelper.js
@@ -162,7 +162,7 @@ define(['appSettings', 'browser', 'events'], function (appSettings, browser, eve
         }
     }
 
-    function seekOnPlaybackStart(instance, element, ticks) {
+    function seekOnPlaybackStart(instance, element, ticks, onMediaReady) {
 
         var seconds = (ticks || 0) / 10000000;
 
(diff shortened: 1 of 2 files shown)


thornbill:
LGTM but I don't have a good way of testing this.

JustAMan:
> LGTM but I don't have a good way of testing this.

Test steps I did:
1. Make a sample video (x265 + 5.1 AAC), embed some random `.ass` subtitle in it
2. Run 10.5.4 server
3. Open said video in mobile Chrome => see no subtitles
4. Apply the fix to jf-web, clear cache, open video in mobile Chrome => see subtitles

frankyifei:
works well for android Edge. I can not test with android Chrome as it always trigger transcode for the reason "SubtitleCodecNotSupported"

JustAMan:
@frankyifei what JF version you have installed? What subtitles are you trying to play?

frankyifei:
10.5.4 Jellyfin.  .ass subtitles. at least it works on android edge which is chrome based.

JustAMan:
> It is my client setting, I cleaned cache and reload. Now is working for Chrome and webview.

So this is a server issue which I have a PR to fix (incorrectly working cache). Isn't exactly relevant to subtitles :)

PR #1097, opened 2020-04-15
Limit rate of timeupdate playback report

dmitrylyzo:
PlaybackManager has two ways for `timeupdate` report:
* player event `timeupdate` (every 250ms)
* interval (every 10s)

**Changes**
Add delay for player and interval `timeupdate` to limit report rate to 1 per 3s.

Volume update rate will also be limited to 1 per 3s, but this applies only to server notifications, and actual volume will be set instantly.

This part may also be delayed, but Live (M3U) works OK so dunno.
https://github.com/jellyfin/jellyfin-web/blob/54a80f8ab668c7f93f67407ca9df678d88829b65/src/components/playback/playbackmanager.js#L3405-L3410

### DIFF
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -2959,7 +2959,7 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
 
         function onPlayerProgressInterval() {
             var player = this;
-            sendProgressUpdate(player, 'timeupdate');
+            sendProgressUpdateDelayed(player, 'timeupdate');
         }
 
         function startPlaybackProgressTimer(player) {
@@ -3267,7 +3267,7 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
 
         function onPlaybackTimeUpdate(e) {
             var player = this;
-            sendProgressUpdate(player, 'timeupdate');
+            sendProgressUpdateDelayed(player, 'timeupdate');
         }
 
         function onPlaybackPause(e) {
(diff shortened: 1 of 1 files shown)


dmitrylyzo:
I found something interest. It seems that `timeupdate` rate already limited.
https://github.com/jellyfin/jellyfin-web/blob/54a80f8ab668c7f93f67407ca9df678d88829b65/src/libraries/apiclient/apiclientcore.js#L1427-L1447

dmitrylyzo:
I think my approach a bit broken.
Suppose we are changing volume, and `timeupdate` is fired.
```
volumechange -> 3s delay
volumechange -> already delayed
volumechange -> already delayed
volumechange -> already delayed
// 3 seconds have not passed yet
timeupdate -> already delayed
// 3 seconds have passed
reportPlaybackProgress with options.EventName = volumechange (timeupdate is not mentioned)
```
Or
```
timeupdate -> 3s delay
volumechange -> already delayed
volumechange -> already delayed
volumechange -> already delayed
volumechange -> already delayed
// 3 seconds have passed
reportPlaybackProgress with options.EventName = timeupdate but 5s have not passed, so no report
```

dmitrylyzo:
Tried two approaches:

1. Remove all limits from `playbackManager` and limit rate in `ApiClient.reportPlaybackProgress` with this change:
`if (["timeupdate", "volumechange"].indexOf(options.EventName || "timeupdate") !== -1) {`
Rate of `timeupdate` and `volumechange` is limited by 1 per 10s (can be tuned).
I don't like the fact that report isn't scheduled and will be rejected if it is too early. But perhaps it doesn't really matter - in any case, it will be sent on stop/pause.

2. Remove limit from `ApiClient.reportPlaybackProgress` and limit rate in `playbackManager` (as in this PR).
Rate of `timeupdate` and `volumechange` is limited by 1 per 3s (can be tuned).
It is scheduled (will be sent) and data is taken at the moment of update - actual data is sent.
But `ApiClient.reportPlaybackProgress` has become "unprotected".

dkanada:
I think we should move it to the API client and then remove the excessive logging from here so we don't spam the logs anymore. If you can get that done today we can include it in the next Android release as well.

PR #1099, opened 2020-04-16
Fix dropdown not properly aligned for long subtitles

### LINKED ISSUE #1098: Dropdown list for subtitle selection not properly aligned if long title
**Describe the bug**
The dropdown list for selecting subtitles is not always displayed correctly, the selection arrow is not placed all the way to the right.

**To Reproduce**

1. Go to the page of a movie
2. Modify the subtitle in the dropdown list to use a looooooooooooooooooong title
3. See that the arrow is not correctly rendered

**Expected behavior**
The arrow for the dropdown list should be correctly aligned at the right.

**Logs**
Not relevant.

**Screenshots**
Screenshot_2020-04-16 Jellyfin

 - OS: Linux
 - Browser: Firefox 74
 - Jellyfin Version: 10.5.4

Delgan:
**Changes**
This is an easy fix, but I guess there might be a reason the class was not used in the first place?

**Issues**
Fixes #1098

### DIFF
--- a/src/itemdetails.html
+++ b/src/itemdetails.html
@@ -136,7 +136,7 @@
                             <div class="selectContainer selectAudioContainer hide trackSelectionFieldContainer flex-shrink-zero">
                                 <select is="emby-select" class="selectAudio detailTrackSelect" label=""></select>
                             </div>
-                            <div class="selectContainer selectSubtitlesContainer hide trackSelectionFieldContainer">
+                            <div class="selectContainer selectSubtitlesContainer hide trackSelectionFieldContainer flex-shrink-zero">
                                 <select is="emby-select" class="selectSubtitles detailTrackSelect" label=""></select>
                             </div>
                         </form>


YouKnowBlom:
I can't seem to repro this. Anyway, couldn't this also happen to the audio dropdown?

Delgan:
@YouKnowBlom Maybe this is fixed on `master`? I tested it using 10.5.3 / 10.5.4, but it seems the tracks selection has been modified in #949, which possibly also fixed that. :confused: 

Also, the parent `div` for audio already contains the `flex-shrink-zero` class so there is no problem.

It's related to the `max-width` of the `trackSelections` container.

Screenshot_2020-04-19 79692506-c7959c00-8265-11ea-9d5c-f31fbc63a6db png (PNG Image, 1920 × 1080 pixels)

Delgan:
I just tested the `master` branch, this seems fixed indeed. :+1: 

Screenshot_2020-04-19 Jellyfin

There's still an overlap problem, but that's a different subject I guess.

I would nevertheless suggest adding the `flex-shrink-zero` class to subtitles for consistency.

PR #1100, opened 2020-04-17
Add scheduled playback progress report

dmitrylyzo:
This is an alternative to #1097

**Changes**
I rewrote (maybe reinvented something :p) `ApiClient.reportPlaybackProgress` in a way that don't skip updates, but schedules them and "accumulates" new one until it (last) is sent.

Closes #1097

### DIFF
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -3280,7 +3280,7 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
 
         function onPlaybackVolumeChange(e) {
             var player = this;
-            sendProgressUpdateDelayed(player, 'volumechange');
+            sendProgressUpdate(player, 'volumechange');
         }
 
         function onRepeatModeChange(e) {
@@ -3375,16 +3375,7 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
 
         pluginManager.ofType('mediaplayer').map(initMediaPlayer);
 
-        /** Delay timer for sendProgressUpdate */
-        var sendProgressUpdateTimer;
-
-        /** Delay time in ms for sendProgressUpdate */
-        var sendProgressUpdateDelay = 700;
-
         function sendProgressUpdate(player, progressEventName, reportPlaylist) {
-            clearTimeout(sendProgressUpdateTimer);
-            sendProgressUpdateTimer = null;
-
             if (!player) {
                 throw new Error('player cannot be null');
             }
@@ -3406,14 +3397,6 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
                         getLiveStreamMediaInfo(player, streamInfo, self.currentMediaSource(player), streamInfo.liveStreamId, serverId);
                     }
                 }
-            }
-        }
-
-        function sendProgressUpdateDelayed(player, progressEventName, reportPlaylist) {
-            if (!sendProgressUpdateTimer) {
-                sendProgressUpdateTimer = setTimeout(function () {
-                    sendProgressUpdate(player, progressEventName, reportPlaylist);
-                }, sendProgressUpdateDelay);
             }
         }
 
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -38,6 +38,12 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
         events.trigger(playbackManagerInstance, 'playerchange', [newPlayer, newTarget, previousPlayer]);
     }
 
+    /** Last invoked method */
+    let reportPlaybackLastMethod;
+
+    /** Last invoke time of method */
+    let reportPlaybackLastTime;
+
     function reportPlayback(playbackManagerInstance, state, player, reportPlaylist, serverId, method, progressEventName) {
 
         if (!serverId) {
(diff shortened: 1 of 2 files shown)


src/components/playback/playbackmanager.js
```diff
@@ -57,7 +63,14 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
             addPlaylistToPlaybackReport(playbackManagerInstance, info, player, serverId);
         }
 
-        console.debug(method + '-' + JSON.stringify(info));
+        const now = (new Date).getTime();
+
+        if (method !== reportPlaybackLastMethod || now - (reportPlaybackLastTime || 0) >= 1e3) {
```
JustAMan, line 68: I would rather we extract this constant as a global variable somewhere on top, for easier tuning later on

src/components/playback/playbackmanager.js
```diff
@@ -57,7 +63,14 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
...
+            console.debug(method + '-' + JSON.stringify(info));
+            reportPlaybackLastMethod = method;
+            reportPlaybackLastTime = now;
+        }
```
JustAMan, line 72: don't we need an `else return` clause here? otherwise we're not spamming the console but _are_ reporting the playback

src/libraries/apiclient/apiclientcore.js
```diff
@@ -1426,25 +1438,74 @@ define(["events", "appStorage"], function(events, appStorage) {
...
+        const eventName = options.EventName || "timeupdate";
+        let reportRateLimitTime = reportRateLimits[eventName] || 0;
+
+        const now = (new Date).getTime();
+        const msSinceLastReport = now - (this.lastPlaybackProgressReport || 0);
+        const newPositionTicks = options.PositionTicks;
+
+        if (msSinceLastReport < reportRateLimitTime && eventName === "timeupdate" && newPositionTicks) {
+            const expectedReportTicks = 1e4 * msSinceLastReport + (this.lastPlaybackProgressReportTicks || 0);
+            if (Math.abs(newPositionTicks - expectedReportTicks) >= 5e7) reportRateLimitTime = 0;
```
JustAMan, line 1451: what does this line do?

src/libraries/apiclient/apiclientcore.js
```diff
@@ -1426,25 +1438,74 @@ define(["events", "appStorage"], function(events, appStorage) {
...
+
+        const eventName = options.EventName || "timeupdate";
+        let reportRateLimitTime = reportRateLimits[eventName] || 0;
+
+        const now = (new Date).getTime();
+        const msSinceLastReport = now - (this.lastPlaybackProgressReport || 0);
+        const newPositionTicks = options.PositionTicks;
+
+        if (msSinceLastReport < reportRateLimitTime && eventName === "timeupdate" && newPositionTicks) {
+            const expectedReportTicks = 1e4 * msSinceLastReport + (this.lastPlaybackProgressReportTicks || 0);
```
JustAMan, line 1450:
```suggestion
            const expectedReportTicks = reportRateLimitTime * msSinceLastReport + (this.lastPlaybackProgressReportTicks || 0);
```
you've extracted this `1e4` into global constant already ;)

src/libraries/apiclient/apiclientcore.js
```diff
@@ -1426,25 +1438,74 @@ define(["events", "appStorage"], function(events, appStorage) {
...
+                contentType: "application/json",
+                url: url
+            });
+        };
+
+        let delay = Math.max(0, reportRateLimitTime - msSinceLastReport);
+
+        promise = new Promise((resolve, reject) => setTimeout(resolve, delay)).then(() => {
+            if (cancelled) return Promise.resolve();
+            return sendReport(instance.lastPlaybackProgressOptions);
```
JustAMan, line 1496: can you explain what's going on here?

src/components/playback/playbackmanager.js
dmitrylyzo: All "magic" in `ApiClient.reportPlaybackProgress` - it remembers last update options to send most recent one.
`ApiClient.reportPlaybackStart` and `ApiClient.reportPlaybackStopped` should be invoked without limits.

src/libraries/apiclient/apiclientcore.js
dmitrylyzo: If we seek, difference will be greater 5 seconds (I suppose), and we temporary reset limit to send update right now.

src/libraries/apiclient/apiclientcore.js
dmitrylyzo: No, this is `ms` to `tick` conversion. Here approximately calculating how much ticks have been played for elapsed time.

src/libraries/apiclient/apiclientcore.js
dmitrylyzo: We delay report for time left to allowed update (last update + `reportRateLimitTime`). Then send report if promise was not canceled.
`lastPlaybackProgressOptions` will contain last report options.

src/components/playback/playbackmanager.js
```diff
@@ -1,6 +1,9 @@
 define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'playQueueManager', 'userSettings', 'globalize', 'connectionManager', 'loading', 'apphost', 'screenfull'], function (events, datetime, appSettings, itemHelper, pluginManager, PlayQueueManager, userSettings, globalize, connectionManager, loading, apphost, screenfull) {
     'use strict';
 
+    /** Delay time in ms for reportPlayback logging */
```
dkanada, line 4:
```suggestion
    /** Delay time in ms for reportPlayback logging */
```

PR #1109, opened 2020-04-18
Change Chromecast player name to Google Cast

YouKnowBlom:
**Changes**
Changes "Chromecast" players name to Google Cast


YouKnowBlom:
@JustAMan That's a good idea, I'm just not sure how to do it with the current implementation. Maybe an extra optional `tooltip` field for each player?

Issue #776, opened 2020-01-28
Gestures for default step on sliders / avoid sliders for values. 


k6Q28k5w5eWuRuuzTg9j7GrKCbjt7:
Hi. I'm having difficulties using the "sliders" to hit correct values on options such as: offset and seeking, especially on smaller devices (touch). 

There might be a better experience to allow input fields / plus-minus buttons, or similar. Another option is gestures to increase/decrease the slider by some default step.

YouKnowBlom:
Subtitle offset has an input field, click on the "0s" above the slider and it'll let you type it in. With that said, I do think it could be made more obvious

MrTimscampi:
Closing this, as we follow the pattern of most other media apps and there are no plans to change it.

PR #1114, opened 2020-04-22
Fix SubtitleSync-TextField behavior

redSpoutnik:
Fix some misbehaviors I noticed upon SubtitleSync-TextField.

**Changes**

* Use ```keydown``` instead of ```keypress``` to capture 'delete' key event.
* Prevent textfield to blur while it has focus (editing subtitle offset).
* Prevent textfield to reset while it has focus.

**Issues**

None, but I can modify the textfield to make it more obvious, as pointed out in #776 (open to suggestion...)


src/components/subtitlesync/subtitlesync.js
```diff
@@ -145,7 +156,9 @@ define(['playbackManager', 'layoutManager', 'text!./subtitlesync.template.html',
                         // if there is an external subtitle stream enabled
                         playbackManager.canHandleOffsetOnCurrentSubtitle(player)) {
                         // if no subtitle offset is defined
-                        if (!playbackManager.getPlayerSubtitleOffset(player)) {
+                        if (!(playbackManager.getPlayerSubtitleOffset(player) ||
+                        // or being defined (element has focus)
+                        subtitleSyncTextField.hasFocus)) {
```
dkanada, line 161: seems like this can be on a single line

redSpoutnik: I made a line break here to comment separted cases, like in:
https://github.com/jellyfin/jellyfin-web/blob/24760a1158c9575a2a96a7e65aabac12d217dbb9/src/components/subtitlesync/subtitlesync.js#L143-L146
but I can change it if you prefer.

Artiume: perhaps this?
 `// if showing subtitle sync is enabled && if there is an external subtitle stream enabled`

dkanada: Yeah, I'm not sure this makes to if-statement more readable so I'd rather have it on one line.

redSpoutnik: Ok, no problem.

PR #1119, opened 2020-04-23
Add check for funky Promise faced on WebOS 2

JustAMan:
I have faced a buggy `Promise()` on WebOS2 - for once, it doesn't understand `Promise.resolve()` throwing a `TypeError`, and `new Promise(function (resolve, reject))` passed `undefined` as `reject` and some funny object as`resolve`.

**Changes**

Force to use a Promise polyfill if missing or buggy `Promise` is detected.

**Issues**

### DIFF
--- a/src/scripts/apploader.js
+++ b/src/scripts/apploader.js
@@ -29,6 +29,16 @@
         );
     }
 
+    try {
+        Promise.resolve();
+    } catch (ex) {
+        // this checks for several cases actually, typical is
+        // Promise() being missing on some legacy browser, and a funky one
+        // is Promise() present but buggy on WebOS 2
+        window.Promise = undefined;
+        self.Promise = undefined;
+    }
+
     if (!self.Promise) {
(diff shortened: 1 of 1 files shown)


src/scripts/apploader.js
```diff
@@ -29,6 +29,16 @@
...
+        Promise.resolve();
+    } catch (ex) {
+        // this checks for several cases actually, typical is
+        // Promise() being missing on some legacy browser, and a funky one
+        // is Promise() present but buggy on WebOS 2
+        window.Promise = undefined;
+        self.Promise = undefined;
+    }
+
     if (!self.Promise) {
```
MrTimscampi, line 42: Can you also remove this condition while you're in there?

The promise polyfill already does that check, so no need to repeat it :)

dkanada: Too late, I'm already merging this to keep open PRs down :) just add it to an unrelated PR and no one will notice.

PR #1124, opened 2020-04-24
Use TV layout by default on PS4

### LINKED ISSUE #1054: Playstation 4 is not recogniced as TV
**Describe the bug**
The web browser of the PS4 is not recognized as a TV, therefore the desktop layout is used instead of the TV layout.
After switching to the TV layout, the focus is working fine, but the select does not work correct, it selects where the cursor is and not the current focused item.

Also in the OSD the AirPlay icon is visible which should not be the case

**To Reproduce**
Open JF with the PS4 web browser

**Expected behavior**
The layout should be the TV layout and the focused item should be selected and not the item behind the hidden cursor.

 - OS: Windows (Server)
 - Browser: PS4 Browser
 - Jellyfin Version: 10.5.2

MrTimscampi:
**Changes**

Sets the TV layout as the default layout or PS4.

**Issues**

Fixes #1054

### DIFF
--- a/src/components/apphost.js
+++ b/src/components/apphost.js
@@ -237,10 +237,6 @@ define(["appSettings", "browser", "events", "htmlMediaHelper", "webSettings"], f
             features.push("voiceinput");
         }
 
-        if (!browser.tv && !browser.xboxOne) {
-            browser.ps4;
-        }
-
         if (supportsHtmlMediaAutoplay()) {
             features.push("htmlaudioautoplay");
             features.push("htmlvideoautoplay");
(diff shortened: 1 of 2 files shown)


src/components/apphost.js
```diff
@@ -237,10 +237,6 @@ define(["appSettings", "browser", "events", "htmlMediaHelper", "webSettings"], f
             features.push("voiceinput");
         }
 
-        if (!browser.tv && !browser.xboxOne) {
-            browser.ps4;
-        }
```
dkanada, line 242: What was this even doing?

MrTimscampi: ¯\\\_(ツ)\_/¯

PR #1125, opened 2020-04-25
Fix event subscription on remotecontrol page

dmitrylyzo:
It seems that subtree is being recreated on `innerHTML  += ...` and all event handlers are lost.

**Changes**
Move HTML generating before event subscription.

**Issues**
`Repeat` button doesn't work.


dmitrylyzo:
Maybe it is better to move that block into `init`? `bindEvents` has a different purpose.
https://github.com/jellyfin/jellyfin-web/blob/e600943ab10569dc7c40d6f91d44c33b22b91333/src/components/remotecontrol/remotecontrol.js#L769-L771

PR #1126, opened 2020-04-25
Check appSettings for undefined

### LINKED ISSUE #1057: Unable to load Playback Settings page
**Describe the bug**
Screenshot from 2020-04-11 00-11-21

**Logs**
Screenshot from 2020-04-11 00-11-37

**System:**
 - OS: Arch Linux
 - Browser: Firefox, Chrome
 - Jellyfin Version: nightly

**Additional context**

This was recently built manually from the master and I cant remember the specific date.
The browser cache has been cleared.

MrTimscampi:
**Changes**

Prevents the infinite loading on the Playback settings by strictly checking for undefined or null instead of loosely checking for null.

**Issues**

Fixes #1057

### DIFF
--- a/src/scripts/settings/appSettings.js
+++ b/src/scripts/settings/appSettings.js
@@ -12,7 +12,7 @@ import events from 'events';
     }
 
     export function enableAutoLogin(val) {
-        if (val !== undefined || val !== null) {
+        if (val !== undefined) {
             this.set('enableAutoLogin', val.toString());
         }
 
@@ -20,7 +20,7 @@ import events from 'events';
     }
 
     export function enableSystemExternalPlayers(val) {
-        if (val !== undefined || val !== null) {
+        if (val !== undefined) {
             this.set('enableSystemExternalPlayers', val.toString());
         }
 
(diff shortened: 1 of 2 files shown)


src/scripts/settings/appSettings.js
```diff
@@ -12,15 +12,15 @@ import events from 'events';
     }
 
     export function enableAutoLogin(val) {
-        if (val != null) {
+        if (val !== undefined || null) {
```
dmitrylyzo, line 15: `val !== undefined || null` ~ `val !== undefined || false` ~ `val !== undefined`

maybe `val !== undefined && val !== null`?

MrTimscampi: Hm yeah, good point. I was trying to make it shorter but that's not ideal :p

YouKnowBlom:
I was a bit quick on the trigger. It seems `val !== null` is always true (indicated by sonarcloud). With that said, it fixed the issue :p

dmitrylyzo:
Originally, `val != null` worked for
- `enableAutoLogin()`
- `enableAutoLogin(undefined)`
- `enableAutoLogin(null)`

I believe we never do this, but what if someone calls `enableAutoLogin(null)`?

dkanada:
I think that would be overengineering in this case.

PR #1131, opened 2020-04-25
Icons for WebOS 2

dmitrylyzo:
WebOS 2 doesn't display most of Material Icons. Earlier, we moved icons with `_` (underscore) into `class` attribute. Now, the turn of the rest.

**Changes**
- All `material-icons` moved into `class` attribute.
- `i` tag changed to `span`
- CSS `i` changed to `.material-icons` to leave it compatible with plugins.

**Issues**
WebOS 2 doesn't display most of Material Icons.


src/components/guide/guide.js
```diff
@@ -577,7 +577,7 @@ define(['require', 'inputManager', 'browser', 'globalize', 'connectionManager',
                     html += '</div>';
 
                     if (program.IsHD && options.showHdIcon) {
-                        //html += '<i class="guideHdIcon material-icons programIcon">hd</i>';
+                        //html += '<span class="guideHdIcon material-icons programIcon hd"></span>';
```
ferferga, line 580: Not strictly in the scope of this PR, but any clue why we aren't using this icon? I think it would be really useful to check which channels are HD at a glance, no clue why it's not used honestly.

Not an user of Live TV, so you might know better.

src/controllers/itemdetailpage.js
```diff
@@ -2078,7 +2078,6 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "userSetti
         bindAll(view, ".btnCancelTimer", "click", onCancelTimerClick);
         bindAll(view, ".btnDeleteItem", "click", onDeleteClick);
         bindAll(view, ".btnDownload", "click", onDownloadClick);
-        view.querySelector(".btnMoreCommands i").innerHTML = "&#xE5D3;";
```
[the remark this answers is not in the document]
dmitrylyzo: In template
https://github.com/jellyfin/jellyfin-web/blob/d98645135b53612d852bf14991d247880f502265/src/itemdetails.html#L102-L107

src/videoosd.html
```diff
@@ -76,7 +76,7 @@ <h3 class="osdTitle"></h3>
 
                 <div class="volumeButtons hide-mouse-idle-tv">
                     <button is="paper-icon-button-light" class="buttonMute autoSize" title="${Mute} (m)">
-                        <i class="xlargePaperIconButton material-icons"></i>
+                        <span class="xlargePaperIconButton material-icons volume_up"></span>
```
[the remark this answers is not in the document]
dmitrylyzo: It was here. Then, we got double volume icon because of `iconElement.innerHTML = isMuted ? '&#xE04F;' : '&#xE050;';` and it was removed.
de28101f660a6f953b5971938b840791de24c5bc
Now this required to be here because of `classList.replace` - it requires "target".

ferferga:
Looks perfect to me!

The HD icon is non-blocked, although I would add it in this PR. Let's see what others think.

src/components/guide/guide.js
dmitrylyzo: First appearance de6ac33ec186d63eaf2e77800a61d2d743357edd
This is probably supposed to be an alternative to text markers. cc @JustAMan

JustAMan: Maybe we can try uncommenting it here

src/components/nowplayingbar/nowplayingbar.js
```diff
@@ -281,22 +279,13 @@ define(['require', 'datetime', 'itemHelper', 'events', 'browser', 'imageLoader',
...
         if (playPauseButtons) {
-            if (isPaused) {
-
-                for (i = 0, length = playPauseButtons.length; i < length; i++) {
-                    playPauseButtons[i].querySelector('i').innerHTML = '&#xE037;';
-                }
-
-            } else {
+            let icons = ['play_arrow', 'pause'];
+            if (isPaused) icons = icons.reverse();
```
JustAMan, line 284:
```suggestion
            let icons = isPaused ? ['pause', 'play_arrow'] : ['play_arrow', 'pause'];
```
maybe this is easier to understand?

src/components/nowplayingbar/nowplayingbar.js
```diff
@@ -397,11 +387,9 @@ define(['require', 'datetime', 'itemHelper', 'events', 'browser', 'imageLoader',
...
         }
 
-        if (isMuted) {
-            muteButton.querySelector('i').innerHTML = '&#xE04F;';
-        } else {
-            muteButton.querySelector('i').innerHTML = '&#xE050;';
-        }
+        let icons = ['volume_off', 'volume_up'];
+        if (isMuted) icons = icons.reverse();
+        muteButton.querySelector('.material-icons').classList.replace(icons[0], icons[1]);
```
JustAMan, line 392: this looks like a need for a helper method probably... also what happens if `icons[0]` is not present in the `.classList`?

src/controllers/playback/videoosd.js
```diff
@@ -746,12 +746,12 @@ define(["playbackManager", "dom", "inputManager", "datetime", "itemHelper", "med
         }
 
         function updatePlayPauseState(isPaused) {
-            var button = view.querySelector(".btnPause i");
+            var button = view.querySelector(".btnPause .material-icons");
             if (isPaused) {
-                button.innerHTML = "&#xE037;";
+                button.classList.replace("pause", "play_arrow");
                 button.setAttribute("title", globalize.translate("ButtonPlay") + " (k)");
```
JustAMan, line 752: I wonder what this `(k)` means...

src/elements/emby-ratingbutton/emby-ratingbutton.js
```diff
@@ -57,12 +57,12 @@ define(['connectionManager', 'serverNotifications', 'events', 'globalize', 'emby
...
     function setState(button, likes, isFavorite, updateAttribute) {
 
-        var icon = button.querySelector('i');
+        var icon = button.querySelector('.material-icons');
 
         if (isFavorite) {
 
             if (icon) {
-                icon.innerHTML = 'favorite';
+                icon.classList.add('favorite');
```
JustAMan, line 65: why you do `.classList.add()` here but `.replace()` in other places?

src/nowplaying.html
```diff
@@ -31,49 +31,49 @@ <h2 class="nowPlayingPageTitle"></h2>
...
                         </button>
                         
                         <button is="paper-icon-button-light" class="btnPreviousTrack btnPlayStateCommand autoSize" title="${ButtonPreviousTrack}">
-                            <i class="material-icons skip_previous"></i>
+                            <span class="material-icons skip_previous"></span>
                         </button>
 
                         <button is="paper-icon-button-light" class="btnPlayPause btnPlayStateCommand autoSize" title="${ButtonPause}">
-                            <i class="material-icons">pause</i>
+                            <span class="material-icons pause_circle_filled"></span>
```
JustAMan, line 42: why change icon?

src/components/nowplayingbar/nowplayingbar.js
dmitrylyzo: > also what happens if `icons[0]` is not present in the `.classList`?

It won't work - `replace` requires a "target".

Another approach is
- remove all possible classes
- add current state class

```suggestion
        const icon = muteButton.querySelector('.material-icons');
        icon.classList.remove('volume_off', 'volume_up');
        icon.classList.add(isMuted ? 'volume_off' : 'volume_up');
```

src/controllers/playback/videoosd.js
dmitrylyzo: This is hotkey.

src/elements/emby-ratingbutton/emby-ratingbutton.js
dmitrylyzo: Nothing to replace - 'favorite' is everywhere.

Ideally, `remove` should be at the beginning. Like here
https://github.com/jellyfin/jellyfin-web/blob/d98645135b53612d852bf14991d247880f502265/src/components/nowplayingbar/nowplayingbar.js#L331-L344

src/nowplaying.html
dmitrylyzo: https://github.com/jellyfin/jellyfin-web/blob/d98645135b53612d852bf14991d247880f502265/src/components/remotecontrol/remotecontrol.js#L385

Earlier, it was replaced by `innerHTML = ` - it doesn't matter what icon is here. Now there is `replace`.

If we will do remove+add then we can leave any icon here.

src/components/nowplayingbar/nowplayingbar.js
dmitrylyzo: At first I didn't use an array. Like here https://github.com/jellyfin/jellyfin-web/blob/d98645135b53612d852bf14991d247880f502265/src/components/remotecontrol/remotecontrol.js#L352-L358
Then I decided that icons are repeated a lot and packed them into an array.

More forgivable way is "remove+add": https://github.com/jellyfin/jellyfin-web/pull/1131#discussion_r418436749
In this case, you do not need to set initial icon (which will be the target for `replace`).

src/components/guide/guide.js
dmitrylyzo: hd_icon
left - icon, right - text

src/components/nowplayingbar/nowplayingbar.js
JustAMan: I like this more, let's use this pattern.

dmitrylyzo:
I didn't touch these
https://github.com/jellyfin/jellyfin-web/blob/03c0ae526fe3260e0eb5bdfdfcf619bc7ce4d1f8/src/components/slideshow/slideshow.js#L226-L241

And I wonder if `replace` has better rendering performance: there is no need to render if there was no replacement.

JustAMan:
> And I wonder if `replace` has better rendering performance: there is no need to render if there was no replacement.

I think most browsers would be smart enough to check pre- and post-states. Also this happens not very often, so I think impact would be minimal anyway.

PR #1134, opened 2020-04-26
Use posters for latest TV shows

### LINKED ISSUE #956: Prefer Primary image instead of Backdrop for thumbnail on home
On the home screen under the "latest media" section for TV Show media libraries the thumbnail provided is the woefully out of place backdrop and not the more relevant and less mismatched Primary image.

MrTimscampi:
**Changes**

Use the poster for Latest sections containing TV shows on the home page.

This brings it in line with other media types, which use their Primary images for the Latest sections.

image

It also switches Photo libraries to using square cards, to be more consistent with how these items appear in their libraries.

**Issues**

Fixes #956

### DIFF
--- a/src/components/homesections/homesections.js
+++ b/src/components/homesections/homesections.js
@@ -243,7 +243,7 @@ define(['connectionManager', 'cardBuilder', 'appSettings', 'dom', 'apphost', 'la
         return function (items) {
             var cardLayout = false;
             var shape;
-            if (itemType === 'Channel' || viewType === 'movies' || viewType === 'books') {
+            if (itemType === 'Channel' || viewType === 'movies' || viewType === 'books' || viewType === 'tvshows') {
                 shape = getPortraitShape();
             } else if (viewType === 'music') {
                 shape = getSquareShape();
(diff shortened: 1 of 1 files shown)


src/components/homesections/homesections.js
```diff
@@ -243,9 +243,11 @@ define(['connectionManager', 'cardBuilder', 'appSettings', 'dom', 'apphost', 'la
         return function (items) {
             var cardLayout = false;
             var shape;
-            if (itemType === 'Channel' || viewType === 'movies' || viewType === 'books') {
+            console.warn(viewType);
+            console.warn(itemType);
```
dmitrylyzo, line 247: :P

MrTimscampi: Dammit :p

MrTimscampi: I need to start using a debugger instead of always doing thing and forgetting to remove them x)

darkbee101:
Hi any idea how to implement this for Mixed contents?

MrTimscampi:
> Hi any idea how to implement this for Mixed contents?

It would need more changes, since these can contain every type of item. We do have a "mixed" card type that can accommodate it, but it hasn't been investigated yet.

PR #1140, opened 2020-04-26
Clean up network settings page

mark-monteiro:
This PR contains the client-side changes related to https://github.com/jellyfin/jellyfin/pull/2774. Changes are summarized as follows:

- remove 'Secure Connection Mode' dropdown and split it into two checkboxes: 'Enable Https' and 'Require Https' (IsBehindProxy is no longer a setting on the server so no need to add a checkbox for that)
- Reorganize settings on page
   - Create a fieldset to contain all the HTTPS-related settings
   - Create a fieldset for remote connection-related settings
   - Move baseUrl out of remote connection settings at closer to the top of the page


mark-monteiro:
I'm sure the UI folks will hate my use of `<fieldset>` here 😅 I think it's the correct semantic markup to use though, so mabye someone could provide some direction on adding some more appropriate styling to them?

dkanada:
Why not just have two options, HTTP and HTTPS respectively? That would be infinitely more clear in my opinion, especially if the port values are directly below.

mark-monteiro:
> 
> 
> Why not just have two options, HTTP and HTTPS respectively? That would be infinitely more clear in my opinion, especially if the port values are directly below.

mark-monteiro:
Sorry, pressed the wrong button 😅 

> Why not just have two options, HTTP and HTTPS respectively? That would be infinitely more clear in my opinion, especially if the port values are directly below.

Yeah, that does make more sense, I'll rework the layout. There isn't actually an option to disable HTTP at this point, it's always on. I think I'll leave adding that toggle out of these two PRs for now and open a second PR later to add it. This PR already has a kind of large scope and that change would involve adding a new config property, and a validation to make sure at least one option is enabled

src/strings/en-us.json
```diff
@@ -626,7 +627,7 @@
     "LabelEmbedAlbumArtDidl": "Embed album art in Didl",
     "LabelEmbedAlbumArtDidlHelp": "Some devices prefer this method for obtaining album art. Others may fail to play with this option enabled.",
     "LabelEnableAutomaticPortMap": "Enable automatic port mapping",
-    "LabelEnableAutomaticPortMapHelp": "Attempt to automatically map the public port to the local port via UPnP. This may not work with some router models. Changes will not apply until after a server restart.",
+    "LabelEnableAutomaticPortMapHelp": "Attempt to automatically map the public ports to local ports via UPnP. This may not work with some router models. Changes will not apply until after a server restart.",
```
Artiume, line 630: Is it the public port or the next nat layer port? Seems like a bit of a misnomer

mark-monteiro: Any suggestion on how to word it? I'm not too familiar with all the networking nomenclature

Artiume: Hm, not sure without confusing half the users. I'm not a fan of uPnP in general. I'd like to add a warning against using it as well. the IGD Protocol is how the port forwarding is mapped with upnp. 

http://www.upnp-hacks.org/upnp.html
https://en.wikipedia.org/wiki/Internet_Gateway_Device_Protocol

Artiume:
Could I get screenshots? I tried building it but the broken wizard is stopping me lol

MrTimscampi:
@Artiume 

image

It's very ugly, so we need some styling, but yes, it's the right markup :p

Artiume:
It's definitely an improvement. I like it a lot more than before. 

The 'allow remote connections' settings. We outta test that to see how well it actually works.

src/strings/en-us.json
PrplHaz4: How about `Automatically forward public ports on your router to local ports via UPnP. This may not work with some router models. Changes will not apply until after a server restart.`

Assumption being the primary use case here is a simple network topology (single router direct to server, no RP...etc).  This option is probably not useful or desired in a more complex setup...

Artiume: My thoughts is how to phrase it so people using docker + bridge mode will know that it won't work.

mark-monteiro: I added the suggestion from @PrplHaz4 with a bit of extra text explaining that this won't work for some network configurations. I'm thinking that if we want to go into additional detail on this it might make more sense to put that in the docs site (which we can link to from here if necessary), while the help text here is good enough for the most common use case. Open to other suggestions though

Artiume:
i built `artiume/jellyfin:network-2774` thats available for testing. It looks good for me, not sure what last minute things you need mark.

mark-monteiro:
> i built `artiume/jellyfin:network-2774` thats available for testing. It looks good for me, not sure what last minute things you need mark.

Functionally, this is good to go. I wasn't sure if we want to handle cleaning up the CSS styling before merging or in a later PR though

Artiume:
gotcha. It could also use a little ux/ui love, but I think that could be done later as well

mark-monteiro:
@jellyfin/web The associated server PR got merged already, so the networking page will be broken on nightly until this is also merged. Can we perhaps merge this now and clean up the UI/UX in a follow-up PR?

bengalih:
Is it possible these are the changes that broke my Chromecasting with NGINX RP in 10.6, detailed at:
https://github.com/jellyfin/jellyfin/issues/3712

My concern is that I am unable to choose the "Require HTTPS" option because to do so I need to install certs on JF itself.  Ofc I don't want to do that because my NGINX hosts all the certs and all communication should go through there.  However, since my ISP blocks 80, my JF server will never receive requests pointed at http://FQDN, which is what the logs appear to show in trying to make certain socket connections.

May be unrelated, but I know something with the ReverseProxy setting changes broke my 10.5.5 working config.

GlennGoddard:
With these change in settings now, how is selecting a cert from a reverse proxy selected now?  The guide still refers to the drop down menu that has been removed.

PR #1141, opened 2020-04-27
Register plugins with promises

cromefire:
**Changes**
Added the possibility to load plugins via Promises

### DIFF
--- a/src/components/pluginManager.js
+++ b/src/components/pluginManager.js
@@ -54,7 +54,7 @@ define(['events', 'globalize', 'appRouter'], function (events, globalize, appRou
         }
 
         if (typeof pluginSpec === "string") {
-            console.debug('Loading plugin: ' + pluginSpec);
+            console.debug('Loading plugin (via deprecated requirejs method): ' + pluginSpec);
 
             return new Promise(function (resolve, reject) {
                 require([pluginSpec], (pluginFactory) => {
@@ -97,6 +97,7 @@ define(['events', 'globalize', 'appRouter'], function (events, globalize, appRou
             return pluginSpec.then(pluginBuilder => {
                 return pluginBuilder();
             }).then(plugin => {
+                console.debug(`Plugin loaded: ${plugin.id}`);
                 return registerPlugin(plugin);
             });
         } else {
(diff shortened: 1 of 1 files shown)


src/components/pluginManager.js
```diff
@@ -25,68 +25,85 @@ define(['events'], function (events) {
...
+                            pluginSpec = appRouter.baseUrl() + '/' + pluginSpec;
+                        }
+                    }
 
-                if (plugin.getRoutes) {
-                    plugin.getRoutes().forEach(function (route) {
-                        definePluginRoute(instance, route, plugin);
-                    });
-                }
+                    var separatorIndex = Math.max(pluginSpec.lastIndexOf('/'), pluginSpec.lastIndexOf('\\'));
```
JustAMan, line 82: why `lastIndexOf('\\')`? I mean why care for both slashes?

src/components/pluginManager.js
```diff
@@ -25,68 +25,85 @@ define(['events'], function (events) {
...
+                });
+            }
+        }
 
-                if (existing) {
-                    resolve(url);
-                    return;
-                }
+        if (typeof pluginSpec === "string") {
+            console.debug('Loading plugin: ' + pluginSpec);
```
JustAMan, line 57: we should blurb that this is a deprecated method of registering a plugin here

cromefire: You mean like with a comment or so?

src/components/pluginManager.js
cromefire: Well I did not add this 🤷🏻‍♀️

src/components/pluginManager.js
JustAMan: no, I mean a message to browser console

src/components/pluginManager.js
JustAMan: but you're the one who's changing the code right now ;)

cromefire: Well didn't want to break something so I left everything I didn't understand why it was there

src/components/pluginManager.js
cromefire: Can do that

src/components/pluginManager.js
JustAMan: Well, okay, let's keep backwards compat here.

src/components/pluginManager.js
cromefire: Fixed

PR #1142, opened 2020-04-27
fix playlist translation for en language

### LINKED ISSUE #1066: Missing String on the Favourites Screen
On the Favourites Screen

image

sarab97:
Fixed the favorites header playlists translation for en languages.

**Changes**
Added HeaderFavoritePlaylists to src json file for en_us and en_bg language

**Issues**
Fixes #1066


sarab97:
Fixed the translation for Playlist and collections.

Changes
Added new entry for translatable in en language and ru for "new" and "filter"
in list return the dynamic library name after translation. 

Issues
Fixes #1120

src/controllers/list.js
```diff
@@ -610,7 +610,7 @@ define(["globalize", "listView", "layoutManager", "userSettings", "focusManager"
             }
 
             if (item) {
-                return item.Name;
+                return globalize.translate(item.Name);
```
dkanada, line 613: Are we sure this should be translated by the web client? It was unclear in chat exactly what's getting translated here.

sarab97: here item.Name is returning dynamic library name which needs to be translated. This can be removed if translation of dynamic library is not required.

MrTimscampi: I have a non-blocking issue with this, that's really more of a maintenance issue.

It potentially introduces strings in translations that aren't anywhere in the code. As such, there's a high chance of strings being removed from translations at some point (since we use code as a reference for cleaning strings).

And what if the name returned has a space in it? Is that a case we need to account for?

I think having the API for libraries being able to translate names itself would be better, but that requires way more engineering.

I'm okay with merging this as-is, but I think we need to rethink this once we rework the APIs.

src/strings/en-gb.json
```diff
@@ -527,6 +528,8 @@
     "Smart": "Smart",
     "SimultaneousConnectionLimitHelp": "The maximum number of allowed simultaneous streams. Enter 0 for no limit.",
     "Shuffle": "Shuffle",
+    "New": "New",
+    "Filter": "Filter",
```
JustAMan, line 532: we don't need to repeat what's present in "base" language (en_US in our case)

PR #1145, opened 2020-04-27
Add a CODEOWNERS file

MrTimscampi:
**Changes**

Adds a CODEOWNERS file to automatically tag people for reviews.

**Issues**


joshuaboniface:
I think we should, it forces a review request for the team. Now, the emails might get annoying so that's a valid concern. It's understood that #web = this repo, so it should be individuals under particular components probably.

MrTimscampi:
> so it should be individuals under particular components probably.

I think once the client is cleaned up further, we can start assigning real owners to various parts. Right now though, everybody works pretty much everywhere so there is no real "I'm somewhat responsible for what happens to this", aside from CI and Org stuff.

joshuaboniface:
Yea that's definitely a longer-term thing. I'd prefer people assign themselves to parts too, so if anyone knows any particular section very well they could add themselves to the codeowners file at a later date.

MrTimscampi:
Sounds good. I can probably assign myself to the Gulp, Webpack and linting files.

.github/CODEOWNERS
```diff
@@ -0,0 +1,3 @@
+*               @jellyfin/web
```
dkanada, line 1: So are we removing this for now or leaving it? I check all web PRs by default because I'm on the web team, so this seems redundant.

JustAMan: I'm for removing it, it creates extra spam for me. I'm getting notified by GH anyway, but direct mentions or tags go to e-mail while others stay in GH notification only (in my setup).

JustAMan, line 1:
```suggestion
```

dkanada: No one has commented again so I'll commit this and merge to clean out the active pull requests for now. We can always revisit this in chat later.

PR #1146, opened 2020-04-27
Fix some issues blocking WebOS 1.2 / 2.0

JustAMan:
**Changes**

* Fix silent app crash on WebOS 1.2 when app requested notification permission (this code is more correct anyway, and permission is granted on WebOS by default)
* Fix WebOS detection by user agent (typo - `big O` changed to `zero`).
* Create home tabs element properly using `document.createElement()` because instead the polyfill used would add custom class later (based on document mutation), and we need custom methods in this execution context. This is a more correct thing to do in any case, it's just in modern browsers element is _usually_ parsed into custom element immediately (but WWW standards do not require so).
* Require `fetch` explicitly and call Babel on some more modules within `node_modules` so WebOS doesn't choke on them.

**Issues**

No ready-made issues, but this being applied to latest master allows me to open my test JF instance in our jf-webos app in WebOS 1.2 and WebOS 2.0 (in emulators). Still not completely functional, but it's a start.

### DIFF
--- a/src/components/maintabsmanager.js
+++ b/src/components/maintabsmanager.js
@@ -139,8 +139,18 @@ define(['dom', 'browser', 'events', 'emby-tabs', 'emby-button'], function (dom, 
 
             var index = 0;
 
-            var indexAttribute = selectedIndex == null ? '' : (' data-index="' + selectedIndex + '"');
-            var tabsHtml = '<div is="emby-tabs"' + indexAttribute + ' class="tabs-viewmenubar"><div class="emby-tabs-slider" style="white-space:nowrap;">' + getTabsFn().map(function (t) {
+            var tabsElement;
+            try {
+                tabsElement = document.createElement('div', {is: 'emby-tabs'});
+            } catch (err) {
+                // older browser not supporting options for createElement
+                tabsElement = document.createElement('div', 'emby-tabs');
+            }
+            if (selectedIndex != null) {
+                tabsElement.setAttribute('data-index', selectedIndex);
+            }
+            tabsElement.className = 'tabs-viewmenubar';
+            tabsElement.innerHTML = '<div class="emby-tabs-slider" style="white-space:nowrap;">' + getTabsFn().map(function (t) {
 
                 var tabClass = 'emby-tab-button';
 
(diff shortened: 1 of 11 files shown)


src/scripts/site.js
```diff
@@ -709,6 +709,10 @@ var AppInfo = {};
             onError: onRequireJsError
         });
 
+        require(["fetch"], function (fetch) {
+            console.debug("fetch is here", fetch);
```
MrTimscampi, line 713: Why the debug log?

dmitrylyzo:
Chrome 79 does not work
https://github.com/jellyfin/jellyfin-web/blob/de27ce0432e599a1dc24584e9f8cb0a784e74918/src/components/maintabsmanager.js#L190
`tabsElem` is `null`

src/scripts/site.js
dmitrylyzo: This is from my patch. I tried to check module status:
* On PC and webOS 4, `fetch` is object
* On webOS 3, `fetch` is `undefined`, but it is works

dmitrylyzo:
For Chrome 79, this works
```js
tabsElement = document.createElement('div', 'emby-tabs');
```

JustAMan:
> For Chrome 79, this works
> 
> ```js
> tabsElement = document.createElement('div', 'emby-tabs');
> ```

But there is a `try..catch` block doing exactly this? Why doesn't it work? What does it say in the console?

Yeah, having a common place for both creating an element and for inserting it as a sole child (so as to replace `foo.innerHTML = bar`) would be good, I'd extract there if you can give me pointers where to put code.

dmitrylyzo:
> Why doesn't it work? What does it say in the console?

It just creates a simple `div`.

src/scripts/site.js
JustAMan: removed

JustAMan:
@dmitrylyzo I think I found a better way, so it _should_ work now on all browsers.

src/components/maintabsmanager.js
```diff
@@ -166,6 +166,7 @@ define(['dom', 'browser', 'events', 'emby-tabs', 'emby-button'], function (dom,
             }).join('') + '</div></div>';
 
             tabsContainerElem.innerHTML = tabsHtml;
+            CustomElements.upgradeSubtree(tabsContainerElem);
```
dmitrylyzo, line 169: to calm eslint
```suggestion
            window.CustomElements.upgradeSubtree(tabsContainerElem);
```

MrTimscampi: We can also add CustomElements to the globals list in .eslintrc.js, but using `window` is probably clearer.

PR #1150, opened 2020-04-28
Remove limit of 50 for tag filter

nielsvanvelzen:
**Changes**
- Removed the tags limit in the filterdialog
  - I did not notice any performance issues afterwards
  - I can now see all my tags (I have around 300 of them, don't ask)
  - For context: this component is used when filtering library views

**Issues**

- none


Artiume:
I wonder what caused the artificial limit to be set in the first place

PR #1155, opened 2020-04-29
Vertical menu

ferferga:
Not so sure about this change but I think it makes sense:

All the menu icons have been changed from the horizontal to vertical, to match the icons used in Android and many other applications (I've never seen one using the horizontal menu honestly, even in desktop)

**Before**
before

**After**
after


src/components/listview/listview.js
```diff
@@ -426,7 +426,7 @@ define(['itemHelper', 'mediaInfo', 'indicators', 'connectionManager', 'layoutMan
 
             html += '<div class="' + cssClass + '">';
 
-            var moreIcon = '&#xE5D3;';
+            var moreIcon = '&#xe5d4';
```
JustAMan, line 429: why do we use hex code here?

src/components/multiselect/multiselect.js
```diff
@@ -129,7 +129,7 @@ define(["browser", "appStorage", "apphost", "loading", "connectionManager", "glo
             html += '<button is="paper-icon-button-light" class="btnCloseSelectionPanel autoSize"><i class="material-icons">close</i></button>';
             html += '<h1 class="itemSelectionCount"></h1>';
 
-            var moreIcon = "&#xE5D3;";
+            var moreIcon = '&#xe5d4';
             html += '<button is="paper-icon-button-light" class="btnSelectionPanelOptions autoSize" style="margin-left:auto;"><i class="material-icons">' + moreIcon + '</i></button>';
```
JustAMan, line 133:
```suggestion
            html += '<button is="paper-icon-button-light" class="btnSelectionPanelOptions autoSize" style="margin-left:auto;"><span class="material-icons more_vert"></span></button>';
```

src/controllers/itemdetailpage.js
```diff
@@ -2078,7 +2078,7 @@ define(["loading", "appRouter", "layoutManager", "connectionManager", "userSetti
         bindAll(view, ".btnCancelTimer", "click", onCancelTimerClick);
         bindAll(view, ".btnDeleteItem", "click", onDeleteClick);
         bindAll(view, ".btnDownload", "click", onDownloadClick);
-        view.querySelector(".btnMoreCommands i").innerHTML = "&#xE5D3;";
+        view.querySelector(".btnMoreCommands i").innerHTML = "&#xe5d4;";
```
JustAMan, line 2081: we'd better use classes rather than inline stuff, inline stuff can break on WebOS and other obscure things

src/components/multiselect/multiselect.js
dmitrylyzo: `span` won't work right now - need my PR

src/controllers/itemdetailpage.js
dmitrylyzo: hex will work, but I removed this line :p - moved icon to HTML template

src/components/multiselect/multiselect.js
JustAMan: okay, ignore me here then :)

src/controllers/itemdetailpage.js
JustAMan: okay, ignore me here then :)

src/components/listview/listview.js
ferferga: No clue honestly. But I think that the main reason discussed in Matrix was for accessibility

JustAMan: we should be using named classes, but whatever

Artiume:
Now that I'm aware of this, I can't unsee it. Curse you @ferferga D:

MrTimscampi:
> Now that I'm aware of this, I can't unsee it. Curse you @ferferga D:

If that can help you, as I said on Matrix, design-wise, it being horizontal makes sense. It's continuing the direction of the other items.

Having it be vertical would be good for cases where it's the only one there, since then it'd be the same direction as the menu it opens.

Artiume:
I'm still torn lol

> image 
image

dkanada:
I think vertical looks better in every situation.

ferferga:
@MrTimscampi As I said in Matrix, I will made the changes once #1131 is merged. 

Also mentioned there, the only place where I'll leave it horizontal is on the item details page, where it makes sense with the labels imo.

dmitrylyzo:
buttons1

MrTimscampi:
> @MrTimscampi As I said in Matrix, I will made the changes once #1131 is merged.
> 
> Also mentioned there, the only place where I'll leave it horizontal is on the item details page, where it makes sense with the labels imo.

I don't have a preference, I was just trying to find an explanation for the original choice :p

Sound good though :+1:

dkanada:
Why can't we use one icon in all situations? Is there any service that actually uses a mix of both?

MrTimscampi:
> Why can't we use one icon in all situations? Is there any service that actually uses a mix of both?

It would look really terrible on the big ribbon on the details page, if we use a vertical one there. That's the main thing.

Artiume:
I say we merge it and see how the masses feel about it. It's an easy revert

ferferga:
@Artiume well, let me fix first merge conflicts :D

dkanada:
> It would look really terrible on the big ribbon on the details page, if we use a vertical one there. That's the main thing.

I really don't think it would, but my only request is that they all be kept the same direction. I'd prefer vertical, but I think it would just be bad design to include two different icons that do the same thing even if they're mostly the same.

MrTimscampi:
> > It would look really terrible on the big ribbon on the details page, if we use a vertical one there. That's the main thing.
> 
> I really don't think it would, but my only request is that they all be kept the same direction. I'd prefer vertical, but I think it would just be bad design to include two different icons that do the same thing even if they're mostly the same.

As we're removing the labels on the ribbon anyway, I think it'll look alright ;)

ferferga:
This is now ready to be merged @jellyfin/web @jellyfin/core

anthonylavado:
Whoops, clicked the wrong button at first. Merged.

PR #1163, opened 2020-04-30
Improve context menus on desktop

MrTimscampi:
**Changes**

- [x] Don't darken the background of context menus on desktop
- [x] Make the menu smaller
- [x] Make the menu stick to its origin element
- [ ] Handle cases where the menu would overflow from the current view

**Issues**

### DIFF
--- a/src/assets/css/librarybrowser.css
+++ b/src/assets/css/librarybrowser.css
@@ -158,7 +158,7 @@
     position: fixed;
     right: 0;
     left: 0;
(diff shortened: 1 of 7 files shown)


dkanada:
This was a good change though, why close it?

MrTimscampi:
It needs a complete rework and the way the menu works is kind of stupid (It makes a big div of the size of the screen, then positions the menu inside of it. That big div follows the screen).

PR #1165, opened 2020-05-01
Add dependabot config

MrTimscampi:
**Changes**

Adds a Dependabot config to allow it to keep dependencies updated in more cases than security updates.

**Issues**

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,0 +1,5 @@
+version: 1
+update_configs:
+  - package_manager: "javascript"
+    directory: "/"
+    update_schedule: "weekly"


dkanada:
Thought we were removing all YAML from the project?

MrTimscampi:
> Thought we were removing all YAML from the project?

Unfortunately, dependabot doesn't have any other option :/

Moving ESLint to JS was done primarily to make editing it easier (as most documentation and config online uses JSON or JS) and reduce the number of languages in the main tooling of the project.  
As this is pretty much a file for repository configuration, not project, it's fine as-is.

Issue #893, opened 2020-03-06
unneccesary transcoding of AV1 when target is compatible


PoepiePeppie:
Jellyfin transcodes AV1 content even though it is not necessary for the target to play AV1 content.
Disabling transcoding for the user makes the file play with no issues.

 - OS: Debian
 - Virtualization: None
 - Played on browser
 - Browser: Firefox 73.1
 - Jellyfin Version: 10.4.3
 - Playback: Direct Play/Transcoding/Remux
 - Installed Plugins: OpenSubtitles
 - Reverse Proxy: Apache
 - Base URL:  /jellyfin
 - Storage: Local

**Steps to reproduce behaviour**
1. Play an AV1 file on firefox 73.1 with transcoding permissions for the user.
2. Playback will crash and give a playback error

**Expected behavior**
AV1 content plays without transcoding if AV1 is supported.

**Logs**
[https://pastebin.com/fWA5xHHp](https://pastebin.com/fWA5xHHp)

**Additional context**
Playback of AV1 content is possible. Disable all transcoding permissions for the users makes the file playback without any issue.

JustAMan:
This is most likely a bug in `deviceprofile.js` or somewhere around, so moving it to web.

Maltahl:
Same problem here:
Log files: 
[transcode error AV1 + OPUS.txt](https://github.com/jellyfin/jellyfin-web/files/4394083/transcode.error.AV1.%2B.OPUS.txt)
[Remux log error AV1 + OPUS.txt](https://github.com/jellyfin/jellyfin-web/files/4394084/Remux.log.error.AV1.%2B.OPUS.txt)
[Jellyfin logs with av1 fail and success direct play.txt](https://github.com/jellyfin/jellyfin-web/files/4394085/Jellyfin.logs.with.av1.fail.and.success.direct.play.txt)

Video file used: 
https://tools.woolyss.com/html5-audio-video-tester/?u=woolyss.com/f/caminandes-1-llama-drama-av1-opus-480p.webm 
file from:
https://tools.woolyss.com/html5-audio-video-tester/

PoepiePeppie:
From looking in the file it seems it's not really a bug but rather that there's just nothing defined for AV1 content.

JustAMan:
@thornbill was it you who last time wanted to clean up `deviceprofile.js`?..

PoepiePeppie:
Should be resolved with #1167

PR #1167, opened 2020-05-01
Added AV1 Codec Profile to Avoid Transcode

Brissot:
**Changes**
New AV1 profile

**Issues**
Fixes: AV1 video can now direct play even if transcoding is allowed for the user so long as the browser supports it.

### DIFF
--- a/src/scripts/browserdeviceprofile.js
+++ b/src/scripts/browserdeviceprofile.js
@@ -334,6 +334,7 @@ define(['browser'], function (browser) {
 
         var canPlayVp8 = videoTestElement.canPlayType('video/webm; codecs="vp8"').replace(/no/, '');
         var canPlayVp9 = videoTestElement.canPlayType('video/webm; codecs="vp9"').replace(/no/, '');
+        var canPlayAv1 = videoTestElement.canPlayType('video/webm; codecs="av1"').replace(/no/, '');
         var webmAudioCodecs = ['vorbis'];
 
         var canPlayMkv = testCanPlayMkv(videoTestElement);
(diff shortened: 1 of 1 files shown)


PoepiePeppie:
we should link this issue #893

YouKnowBlom:
Any chance you could squash those commits into one? The rest looks good to me :p

Brissot:
> 
> 
> Any chance you could squash those commits into one? The rest looks good to me :p

Squashed

src/scripts/browserdeviceprofile.js
```diff
@@ -592,6 +593,15 @@ define(['browser'], function (browser) {
...
         }
 
+        if (canPlayAv1) {
+            profile.DirectPlayProfiles.push({
+                Container: 'webm',
+                Type: 'Video',
+                AudioCodec: webmAudioCodecs.join(','),
+                VideoCodec: 'AV1'
+            });
+        }
```
dmitrylyzo, line 603: I think we could merge these blocks by collecting supported video codecs in some array (`webmVideoCodecs`).

jaytohe:
Hi! Your solution works great for Firefox; I can stream AV1-encoded videos fine. However it doesn't seem to work with Chrome at all...I searched around online and Chrome seems to support AV1 codec. What gives?

Brissot:
> 
> 
> Hi! Your solution works great for Firefox; I can stream AV1-encoded videos fine. However it doesn't seem to work with Chrome at all...I searched around online and Chrome seems to support AV1 codec. What gives?

Interesting, on Google Chrome the logs still say that while Direct Play is supported for the file, there's no profile. I think it has something to do with Jellyfin in Google Chrome treating both mkv and webm as mkv's, while Firefox treats all mkv's and webm's as webm's which could require a bigger fix. I'll look into it when I have the time, but I think that's where the issue lies.

Issue #928, opened 2020-03-11
Detail page - Align title with other content during loading


Nickbert7:
During the loading of the detail page there is a gap on the left side of the title.
It should be aligned to the other content of the page.

![image](https://user-images.githubusercontent.com/14938497/76419639-fb48f080-63a0-11ea-9430-7e4bd10331da.png)
![image](https://user-images.githubusercontent.com/14938497/76419646-fe43e100-63a0-11ea-83a4-37142a50105a.png)

https://imgur.com/otXitb1

PR #1172, opened 2020-05-02
Fix details position while poster is loading

MrTimscampi:
**Changes**

Pushes the content of the details page to their proper position while the poster loads, to prevent having them slide over.

**Issues**

I'm pretty sure there was an issue for this, but I'm not able to find it.

### DIFF
--- a/src/assets/css/librarybrowser.css
+++ b/src/assets/css/librarybrowser.css
@@ -598,6 +598,7 @@
     position: relative;
     margin-top: -25vh;
     float: left;
+    min-height: 35vh;
     width: 25vw;
     z-index: 3;
 }


MrTimscampi:
Before

image

After

image

dmitrylyzo:
There will be a gap between card and progressbar if you have video with extracted image (for example, TV show episode).
progress-gap

I have some fix for placeholder cards (no poster) but stuck with WebOS 2: `float` (card) doesn't like `absolute` (progressbar) - page layout is messed up.

JustAMan:
@MrTimscampi why close?

Nickbert7:
> I'm pretty sure there was an issue for this, but I'm not able to find it.

Should be my issue https://github.com/jellyfin/jellyfin-web/issues/928

MrTimscampi:
> @MrTimscampi why close?

@JustAMan It's fixed in a better way as part of #1206 

@Nickbert7 Thanks :) I'll tag it in the other PR.

Issue #836, opened 2020-02-21
Changing stylesheets to SCSS


h1nk:
Seeing as this project is already using Webpack it'd be nice to make use of SCSS to allow for more DRY and organized stylesheets. I believe @MrTimscampi has already brought this idea up before. An initial transition to SCSS could be made without any stylesheet refactoring or changes since CSS3 is automatically valid SCSS.

---

**(Edit by MrTimscampi)**

Here's a list of the files to convert after merging #862, in order to keep track of progress.

When converting a file, don't hesitate to reorganize everything, try to find dead rules and, if necessary, create more files (`variables.scss` is one that comes to mind, for example)

- [ ] ./src/components/htmlvideoplayer/style.css
- [ ] ./src/components/filterdialog/style.css
- [ ] ./src/components/actionsheet/actionsheet.css
- [ ] ./src/components/youtubeplayer/style.css
- [ ] ./src/components/images/style.css
- [ ] ./src/components/dialogHelper/dialoghelper.css
- [ ] ./src/components/slideshow/style.css
- [ ] ./src/components/emby-scrollbuttons/emby-scrollbuttons.css
- [ ] ./src/components/alphapicker/style.css
- [ ] ./src/components/backdrop/backdrop.css
- [ ] ./src/components/headroom/headroom.css
- [ ] ./src/components/playback/iconosd.css
- [ ] ./src/components/subtitleeditor/subtitleeditor.css
- [ ] ./src/components/toast/toast.css
- [ ] ./src/components/viewManager/viewContainer.css
- [ ] ./src/components/logoscreensaver/style.css
- [ ] ./src/components/mediainfo/mediainfo.css
- [ ] ./src/components/imageuploader/style.css
- [ ] ./src/components/navdrawer/navdrawer.css
- [ ] ./src/components/lazyloader/lazyedgehack.css
- [ ] ./src/components/userdatabuttons/userdatabuttons.css
- [ ] ./src/components/cardbuilder/card.css
- [ ] ./src/components/appfooter/appfooter.css
- [ ] ./src/components/guide/guide.css
- [ ] ./src/components/guide/programs.css
- [ ] ./src/components/listview/listview.css
- [ ] ./src/components/homesections/homesections.css
- [ ] ./src/components/imageeditor/imageeditor.css
- [ ] ./src/components/upnextdialog/upnextdialog.css
- [ ] ./src/components/multiselect/multiselect.css
- [ ] ./src/components/search/searchfields.css
[...]

PR #1175, opened 2020-05-04
Add folder for common SCSS code

GuilhermeHideki:
**Changes**
This PR creates a file in a folder, named `/src/styles`, for reusable SCSS utilities, like:
- variables
- mixins

P.S.: The file is prefixed with `_` to be considered a partial file and to be not included in the `/dist` folder

Doesn't solve, but helps #836


src/stylesheets/mixins.scss
```diff
@@ -0,0 +1,13 @@
+@mixin background-cover($position) {
+    background-size: cover;
+    background-repeat: no-repeat;
+    background-position: $position;
+}
+
+@mixin position($position, $top: null, $right: null, $bottom: null, $left: null) {
```
GuilhermeHideki, line 7: This mixin does not save a lot of typing, but helps to track where absolute positioning is used for later refactoring

gulpfile.js
```diff
@@ -36,7 +36,7 @@ const options = {
         query: ['src/standalone.js', 'src/scripts/apploader.js']
     },
     css: {
-        query: ['src/**/*.css', 'src/**/*.scss']
+        query: ['src/**/*.css', 'src/**/*.scss', '!src/stylesheets/**/*']
```
GuilhermeHideki, line 39: Ignores the folder from outputting empty css files

Should have open an issue to decide the folder name, but since the change is small, it could be easily updated

GuilhermeHideki:
@dkanada I guess that I didn't write the "why change" in the description, sorry. Should I put this in the first post or create an issue to discuss this?

The main issues, to me, are:
- The assets folder, to me, should have only the files that client uses directly
  - We can just add a folder inside `assets/css` to don't generate files from that
- I would prefer to not to import the mixins from assets folder to use in the theme or components folder

I would like @MrTimscampi input on this, since he edited #836

MrTimscampi:
I think having a "styles" folder in `src` would be fine, for general utility SASS (Mixins, resets and such).

The CSS situation will be a mess all throughout the migrations, as ideally when we are using Vue, most CSS will live next to its component. Right now, `assets/css` has a bit of everything: app-wide fonts, some page-specific styles (the entire details page styles lives there, for example)

So I'm for a dedicated "utility" styles folder, though I'd name it "styles" instead of "stylesheets".

@dkanada Keep in mind that, when built, the client won't contain that folder, since everything in it will be imported by the final stylesheets.

dkanada:
Now that you two have explained it I agree :smiley: I haven't thought too much about the assets folder in some time. The JavaScript was in such a poor state it has become the main focus lately.

GuilhermeHideki:
@dkanada @MrTimscampi I applied the changes needed and rebased the PR.

gulpfile.js
```diff
@@ -36,7 +36,7 @@ const options = {
         query: ['src/standalone.js', 'src/scripts/apploader.js']
     },
     css: {
-        query: ['src/**/*.css', 'src/**/*.scss']
+        query: ['src/**/*.css', 'src/**/*.scss', '!src/styles/**/*']
```
GuilhermeHideki, line 39: Probably should filter the output, not input. When changing files in `src/styles`, they don't trigger the css watch 🤦

GuilhermeHideki: I guess that I should remove this change and just use sass partial convention (`_file.scss`)

PR #835, opened 2020-02-21
Add cinema mode

Ullmie02:
### Changes

This (re?)adds the cinema mode. It allows you to play a defined number of trailers before a movie. You can specify whether you only want to see trailers of movies not marked as watched.
The feature is discussed [here](https://features.jellyfin.org/posts/56/cinemamode-plugin) on Fider.

### Current Limitations

- ~~Currently the feature only works if you play the movie from the details page and not from any lists.~~ (fixed in 1eba6206be4434cbabf738da0172831424afd47d)

- ~~**The feature is currently not working correctly!** After the trailers finish playing, the backdropContainer is not correctly set as hidden, so you can't see the movie, only hear it.
Maybe someone could help me out here, as I'm very new to the code.~~ (fixed in 2ffa0a08ad6307a72221449e47387c2ce86c61c1)

PR #1178, opened 2020-05-04
Re-enable cinema mode

MrTimscampi:
**Changes**

I messed up #835, so here is the original contents of PR with my correction.

**Issues**


dmitrylyzo:
Good news: it works with patch below
Bad news: `IsUnPlayed` is "never played before", not unwatched/unmarked

<details>
<summary>Quick fix (just to make it work)</summary>

```diff
diff --git a/src/components/playback/playbackmanager.js b/src/components/playback/playbackmanager.js
index b03a59380..45e1c2d76 100644
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -334,13 +334,22 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
         return apiClient.getItems(apiClient.getCurrentUserId(), fetchOptions).then(function (unwatchedMovies) {
             var promises = [];
 
-            for (let unwatchedMovie in unwatchedMovies) {
+            for (let unwatchedMovie of unwatchedMovies.Items) {
                 const remoteTrailers = getRemoteTrailers(unwatchedMovie);
 
                 promises.push(getAllTrailers(unwatchedMovie, remoteTrailers));
             }
 
             return Promise.all(promises);
+        }).then(function (trailersOfUnwathedMovies) {
+            const introItems = [];
+
+            // Choose random trailer for each movie
+            for (let trailers of trailersOfUnwathedMovies) {
+                introItems.push(trailers[Math.floor(Math.random() * trailers.length)]);
+            }
+
+            return Promise.resolve({Items: introItems});
         }).catch(function (err) {
             console.error(`failed to get trailers: ${err}`);
 
diff --git a/src/scripts/site.js b/src/scripts/site.js
index ddb152eb4..6aa4213bc 100644
--- a/src/scripts/site.js
+++ b/src/scripts/site.js
@@ -845,7 +845,7 @@ var AppInfo = {};
         define('sortMenu', [componentsPath + '/sortmenu/sortmenu'], returnFirstDependency);
         define('sanitizefilename', [componentsPath + '/sanitizeFilename'], returnFirstDependency);
         define('toast', [componentsPath + '/toast/toast'], returnFirstDependency);
-        define('scrollHelper', [componentsPath + '/scrollHelper'], returnFirstDependency);
+        define('scrollHelper', [scriptsPath + '/scrollHelper'], returnFirstDependency);
         define('touchHelper', [scriptsPath + '/touchHelper'], returnFirstDependency);
         define('imageUploader', [componentsPath + '/imageUploader/imageUploader'], returnFirstDependency);
         define('htmlMediaHelper', [componentsPath + '/htmlMediaHelper'], returnFirstDependency);
```
</details>

MrTimscampi:
@jellyfin/web We should really get this merged, the original one is from late February.

src/scripts/settings/userSettings.js
```diff
@@ -89,9 +89,27 @@ import events from 'events';
         }
 
         val = this.get('enableCinemaMode', false);
+        if (val == null) return false;
         return val !== 'false';
```
dmitrylyzo, line 93: Not tested
```suggestion
        return val === 'true';
```

src/components/playback/playbackmanager.js
```diff
@@ -279,17 +279,79 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
...
+            } else {
+                return Promise.resolve(remoteTrailers);
+            }
+        }
 
-            return result;
+        return apiClient.getItems(apiClient.getCurrentUserId(), fetchOptions).then(function (unwatchedMovies) {
+            var promises = [];
+
+            for (let unwatchedMovie in unwatchedMovies.Items) {
```
dmitrylyzo, line 337:
```suggestion
            for (let unwatchedMovie of unwatchedMovies.Items) {
```
_My patch has this_

dmitrylyzo:
> The problem is that I see trailers of single movie.

`HasTrailer: true,` for some reason ignores remote trailers.
https://github.com/jellyfin/jellyfin/blob/d38adb95a727203b0d0dcee344f03b374b5d2b8f/MediaBrowser.Controller/Entities/UserViewBuilder.cs#L744-L750
_btw, we could use `hasTrailers.GetTrailerCount` - will save some nanoseconds :p_
https://github.com/jellyfin/jellyfin/blob/d38adb95a727203b0d0dcee344f03b374b5d2b8f/MediaBrowser.Controller/Entities/IHasTrailers.cs#L39-L40
This requires backend debugging.

> Hmm, after youtube trailer, videoosd doesn't want to be shown. There seems to be something wrong with switching players.

d7c2fd7337d35f6cf87d54b59769c663d3934f50
Everything works if we don't change players.
As an option, check for non-videoosd page.

src/components/playback/playbackmanager.js
```diff
@@ -279,17 +279,79 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
...
 
-        return apiClient.getIntros(firstItem.Id).then(function (result) {
+        var fetchOptions = {
+            Fields: 'RemoteTrailers,LocalTrailerCount,ServerID',
+            Filters: unwatchedOnly ? 'IsUnPlayed' : '',
+            HasTrailer: true,
+            IncludeItemTypes: 'Movie',
+            Limit: trailerCount,
+            Recursive: true,
+            SortBy: 'Random'
```
JustAMan, line 299: I don't think we should be shuffling stuff server-side, we better do it client-side for server performance

src/components/playback/playbackmanager.js
```diff
@@ -279,17 +279,79 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
...
+        function getRemoteTrailers(item) {
+            var remoteTrailers = item.RemoteTrailers || [];
+
+            if (!remoteTrailers.length) {
+                return [];
+            }
+
+            return remoteTrailers.map(function (t) {
+                return {
+                    Name: t.Name || (item.Name + ' Trailer'),
```
JustAMan, line 311: this `' Trailer'` string should be i18ed

src/components/playback/playbackmanager.js
```diff
@@ -279,17 +279,79 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
...
+                    Type: 'Trailer',
+                    ServerId: apiClient.serverId()
+                };
+            });
+        }
+
+        function getAllTrailers(item, remoteTrailers) {
+            if (item.LocalTrailerCount) {
+                return apiClient.getLocalTrailers(apiClient.getCurrentUserId(), item.Id).then(function (trailers) {
+                    trailers.push.apply(trailers, remoteTrailers);
```
JustAMan, line 323: any reason _not_ to call `trailers.push(remoteTrailers)`?

dmitrylyzo: We should push elements of `remoteTrailers`, not the array itself, i.e merge arrays

src/components/playback/playbackmanager.js
dmitrylyzo: ref https://github.com/jellyfin/jellyfin-web/pull/835#discussion_r399773115

PR #1198, opened 2020-05-07
Switch player backdrops to poster attribute of the video element

ferferga:
We've been using a custom implementation for displaying a video backdrop while the video that is going to be played was loading. However, this implementation was disabled in TVs and mobile devices because it looked weird and even in desktop it did weird things sometimes.

**Changes**
* Use the **poster** attribute of the video element for backdrops
* Show backdrops in all devices
* Backdrops now take the full size of the video element, instead of taking all the available space (so it should remove the seams that appeared once the video was loaded and it started playing)

**Before**
before

**After**
after

### DIFF
--- a/src/components/htmlvideoplayer/plugin.js
+++ b/src/components/htmlvideoplayer/plugin.js
@@ -836,7 +836,6 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
         function onNavigatedToOsd() {
             var dlg = videoDialog;
             if (dlg) {
-                dlg.classList.remove('videoPlayerContainer-withBackdrop');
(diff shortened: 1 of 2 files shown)


ferferga:
Although the poster attribute brought an standarised and uniform look for all devices, I wasn't so happy about the black borders. I decided that we can have both added together, as, with a blur effect, it doesn't matter that much that the background looks bad in portrait mode.

**Blur effect**
blur1
blur2
blur3

**Blur effect with the same darkening level applied to itemdetails' backdrops**
blur1-dark
blur2-dark
blur3-dark

ferferga:
@thornbill Yeah, it seems the best idea.

No worries about the back and forth, this is how reviewing goes :). I'll hopefully have some time to revert the commits this night and fix the CSS linting

PR #1208, opened 2020-05-09
Per default a new user has no access to any library

twinkybot:
**Changes**

The changes grants NO access per default to any library. The admin has to enable necessary libraries.
This is to prevent wrong access to media. e.g. Private media like Photos, ones own videos, Horror movies, Rated movies etc.

In contrast without the change a new user is automatically granted full access. Opt-in is in my humble opinon the way to go.


MrTimscampi:
> Forgot to mention due to the LDAP Bug. On the intial user creation page I unset the libraries I did not want to share but after creation neither password was set nor libary access was properly set.
> Thus leading to full access which I did not want.

This would probably warrant an issue on the server or LDAP plugin repository, if one doesn't exist;

PR #1209, opened 2020-05-09
Enable Shaka Player

MrTimscampi:
**Changes**

Enables Shaka Player as the default video player.

**Playback Tests**

- [x] Firefox 75
- [x] Chromium 81
- [x] WebOS 3.x
- [x] WebOS 4.x
- [x] WebOS 5.x

**Issues**


MrTimscampi:
@dmitrylyzo Could you check if Shaka Player is used on WebOS 1.2?

You should see a warning in the console that says "Using shaka-player".

If it works, we can probably remove the other players.

src/components/htmlvideoplayer/plugin.js
```diff
@@ -347,51 +292,46 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
             });
         }
 
-        function onShakaError(event) {
+        function onShakaErrorEvent(event) {
+            // Extract the shaka.util.Error object from the event.
+            onShakaError(event.detail);
+        }
```
dkanada, line 298: Why not just handle this in the method below with a variable or something?

src/scripts/browserdeviceprofile.js
```diff
@@ -609,7 +595,7 @@ define(['browser'], function (browser) {
         if (canPlayHls() && browser.enableHlsAudio !== false) {
             profile.TranscodingProfiles.push({
                 // hlsjs, edge, and android all seem to require ts container
-                Container: !canPlayNativeHls() || browser.edge || browser.android ? 'ts' : 'aac',
+                Container: !canPlayNativeHls() || browser.edge || browser.android || browser.firefox ? 'ts' : 'aac',
```
dkanada, line 598: Is this to fix the weird FLAC support on Firefox?

src/components/htmlMediaHelper.js
```diff
@@ -32,28 +32,7 @@ define(['appSettings', 'browser', 'events'], function (appSettings, browser, eve
...
-                }
-
-                //return false;
-            }
-
-            return true;
-        }
-
-        return false;
+        return true;
```
dkanada, line 35: We probably don't need to disable linting here anymore, but I also feel like this function could just be removed entirely.

src/scripts/browserdeviceprofile.js
Artiume: I think it's supposed to be fixed on FF in the next version.

MrTimscampi: I believe it was a test of sorts that I forgot to remove and forgot the purpose of. I've removed it for now.

src/components/htmlMediaHelper.js
```diff
@@ -30,30 +30,11 @@ define(['appSettings', 'browser', 'events'], function (appSettings, browser, eve
         return false;
     }
 
-    function enableHlsShakaPlayer(item, mediaSource, mediaType) {
+    function enableHlsShakaPlayer() {
         /* eslint-disable-next-line compat/compat */
         if (!!window.MediaSource && !!MediaSource.isTypeSupported) {
```
dmitrylyzo, line 35: `window.MediaSource` is `undefined` on WebOS 1.2 emulator. _No Shaka for you._

src/components/htmlMediaHelper.js
```diff
@@ -30,30 +30,11 @@ define(['appSettings', 'browser', 'events'], function (appSettings, browser, eve
...
-                    //}
-                }
-
-                //return false;
-            }
-
             return true;
         }
-
-        return false;
```
dmitrylyzo, line 56: I think you should leave this `return`.

src/components/htmlMediaHelper.js
MrTimscampi: It'll fall back down to the native player :)

HLS.js acts very much the same and doesn't work without MSE (Per the readme: `It relies on HTML5 video and MediaSource Extensions for playback.`)

src/components/htmlMediaHelper.js
MrTimscampi: Yep, good point.

ferferga:
It's late here and I'm going to bed soon, but I wanted to chime in and review this PR. However, I noticed the following and I wanted to comment that out, just to confirm that these situations have been tested:

* I can't resume anything. It seems that Shaka reliably gets to the point where I left off, but it doesn't go past the poster image. Tested against current master server and 10.5.4. The poster behaviour is specifically for 10.5.4, in master it doesn't start at all. Maybe @YouKnowBlom [has the solution?](https://matrix.to/#/!WXGcrduLZmrIdSyCrv:bonifacelabs.ca/$HShoDgd6JOQj00gQhJmYr_fZB4rVRfbQGOr0yF9gr8c?via=bonifacelabs.ca&via=matrix.org&via=matrix.possumlodge.me)

* Changing between subtitle tracks returned me an undefined in a promise. The stacktrace was very ambiguous and sadly I forgot to take a screenshot. It didn't happen to me all the times though.

PR #1222, opened 2020-05-11
Make dialogs more bearable

MrTimscampi:
**Changes**

Makes all dialogs smaller (They use the "small" size, other sizes have been removes) and redesigns the Metadata Editor a little.

Some fields have been made inline:
image

Before:

image

After:

image

**Issues**


hamburglar2160:
Why not go even further with the inline metadata fields?  Certainly more than three or four "at most 5 character fields" can fit on a row very comfortably.

MrTimscampi:
> Why not go even further with the inline metadata fields? Certainly more than three or four "at most 5 character fields" can fit on a row very comfortably.

It's difficult without a deeper redesign, which I'm not willing to do right now (I was only resizing the dialogs at first, but it somehow evolved into this).

I've grouped everything that makes sense and can be grouped right now.

MrTimscampi:
@ferferga They are aligned left on Firefox (see my screenshot). Seems like Chrome is being annoying again...

ferferga:
@MrTimscampi I'm afk right now but I remember fixing it by setting justify-content and align-items to flex-end in the formDialogButton class (iirc that's the name of the parent class)

ferferga:
@MrTimscampi Also, another thing I noticed in the short time I've been testing is that, in image editor, the scrollbar goes up to the titlebar.

MrTimscampi:
> @MrTimscampi Also, another thing I noticed in the short time I've been testing is that, in image editor, the scrollbar goes up to the titlebar.

That was already the case before ;)

ferferga:
@MrTimscampi Hope you don't mind my commit, I found what was causing the scrollers to go into the titles and, as you didn't modify any of the files that were needed to make this fix possible, I thoght that comitting directly was better and quicker than suggesting it.

ferferga:
I also found a tiny bug that (afaik) happens only with backdrop images and small window sizes. Input Manager can't recognise the action it needs to do when the delete button collapses:

MrTimscampi:
@ferferga That was a CSS issue, fixed :)

PR #1223, opened 2020-05-11
Fix wizard route and config load error

dkanada:


src/scripts/routes.js
```diff
@@ -154,13 +174,18 @@ define([
...
         path: '/home.html',
         autoFocus: false,
         controller: 'home',
         transition: 'fade',
         type: 'home'
     });
+    defineRoute({
+        path: '/search.html',
+        controller: 'searchpage'
+    });
```
dkanada, line 188: I'm slowly moving these into groups so they're easier to find.

src/scripts/settings/webSettings.js
```diff
@@ -2,7 +2,17 @@ let data;
...
+    return fetch('/config.json?nocache=' + new Date().getUTCMilliseconds()).then(response => {
+        data = response.json();
+        return data;
+    }).catch(error => {
+        return getDefaultConfig();
+    });
+}
+
+function getDefaultConfig() {
+    console.warn('web config file is missing so the template will be used');
```
dmitrylyzo, line 14: I would place `warn` in `catch` block - warning about problem with `getConfig`, not `getDefaultConfig`.

src/scripts/settings/webSettings.js
```diff
@@ -2,7 +2,17 @@ let data;
 
 function getConfig() {
     if (data) return Promise.resolve(data);
-    return fetch('/config.json?nocache=' + new Date().getUTCMilliseconds()).then(function (response) {
+    return fetch('/config.json?nocache=' + new Date().getUTCMilliseconds()).then(response => {
```
MrTimscampi, line 5: Watch out, as this still won't handle servers with a baseUrl set.

Not blocking though, it can be done later.

dkanada: That was intentional because I was running into some issues with that and wanted to look into it more. I haven't forgotten it though, I think I'll look into the plugin URLs at the same time.

PR #1224, opened 2020-05-11
Update page title for navigation and now playing

### LINKED ISSUE #1147: Update page title for navigaion and now playing music
When browsing Jellyfin web it seems to update the page title randomly. I can see it updates for entering a video library and playing a video. However it does not appear to update the title for any other pages or for now playing music. I would expect the title to always show the now playing music. When music or a video is not playing then every page should update the title instead of just defaulting to Jellyfin.

JasonGhent:
**Changes**
Update page title for navigation and now playing.

**Issues**
Fixes #1147

### DIFF
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -69,6 +69,8 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
         }
 
         const now = (new Date).getTime();
+        
+        (['reportPlaybackStopped'].includes(method) && window.LibraryMenu.setDefaultTitle() === undefined) || window.LibraryMenu.setTitle((info.IsPaused ? '⏸' : '▶') + ' ' + state.NowPlayingItem.AlbumArtist + ' - ' + state.NowPlayingItem.Name);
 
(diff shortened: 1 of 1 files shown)


dkanada:
Browsers will already show an indicator when audio is playing from a tab, and I think it's better to show the current page rather than the current song, since that is also handled by most browsers in a more native fashion.

Nickbert7:
> Browsers will already show an indicator when audio is playing from a tab, and I think it's better to show the current page rather than the current song, since that is also handled by most browsers in a more native fashion.

At least for music I'd like to see the name of the current song playing in the background in another tab.
Just the indicator doesn't tell me enough.
As a compromise, you could only show that on the now playing page.

MrTimscampi:
IMO, it should only be shown on the Now Playing page (and maybe the video playback page), as the other pages should keep their title while navigating the client.

I also think the code in this PR is needlessly minified, which makes it harder to read than it should be. The clarity of code is better than conciseness.

dkanada:
We decided to avoid changing the tab name with the now playing information for two reasons. Icons generally don't look good in tab names, and we already have extremely good media session support, which will display that information in most cases.

PR #1226, opened 2020-05-13
Convert dark theme to SCSS

cromefire:
**Changes**

Convert dark theme (default) to SCSS, for easier maintainability and remixability.


dkanada:
@cromefire do you want to fix this with the latest changes from master?

cromefire:
Hmm yeah I can try to re do it, maybe without the converter this time, because that was more of a mess than useful I think

MrTimscampi:
I think the proper way to do this would be to figure out what parts are shared between all the themes (even with different values) then make one baseTheme.scss file and use variables for variations, with some theme-specific overrides on some stuff if needed.

cromefire:
Yeah that would be the ideal case but I'm not sure how easy that is...

PR #1229, opened 2020-05-14
Keep scrollbar always present to avoid UI clipping

ferferga:
I found that we can fix the scrollbar clipping while switching between pages easily using ``overflow-y: scroller``. Pages that don't have scroller won't resize now.

**Bug**

* While the web client is being loaded, an ugly standard scroller is displayed (because our custom one is not loaded yet)


src/components/slideshow/slideshow.js
```diff
@@ -611,6 +614,9 @@ define(['dialogHelper', 'inputManager', 'connectionManager', 'layoutManager', 'f
         self.hide = function () {
             if (dialog) {
                 dialogHelper.close(dialog);
+                // Shows page scrollbar
+                document.body.classList.remove('hide-scroll');
+                document.body.classList.add('force-scroll');
```
dmitrylyzo, line 619: This should go to `onDialogClosed` because `hide` is for external closing.

ferferga: @dmitrylyzo Done in my last commit. When opening, my approach was correct, right?

dmitrylyzo: > When opening, my approach was correct, right?

It depends on when you want to hide scrollbar. As alternative, you could place `classList.add` in promise callback here https://github.com/jellyfin/jellyfin-web/blob/a5b8444e7e982183f94dac274eb5f7fe967cb2fc/src/components/slideshow/slideshow.js#L211-L213
But I didn't check what differs, and it is probably unnoticeable.

ferferga:
Maybe this can go into its own module so we reduce the duplication reported by Sonarqube? I think it's starting to make more sense, now that we're also going to make custom dialogs for ebooks and so on (each of them would need the same hide/show class switch for the scrollbar)

MrTimscampi:
> Maybe this can go into its own module so we reduce the duplication reported by Sonarqube? I think it's starting to make more sense, now that we're also going to make custom dialogs for ebooks and so on (each of them would need the same hide/show class switch for the scrollbar)

I think it'd be a good thing to have in a base class of sorts, but that's for later, I'd say :)

src/components/slideshow/slideshow.js
ferferga: @dmitrylyzo I just created a photos library and I like how it hides/shows, it's in the right time imho. I think I'll leave it as is.

I tried moving the add/remove to the promise, but it didn't do anything (didn't investigate why tho).

I'll re-request a review now that this is solved and I think nothing else is affected with the new behaviour.

ferferga:
@MrTimscampi The ugly scrollbar is fixed since I moved the class add/remove to skinManager here: https://github.com/jellyfin/jellyfin-web/pull/1229/commits/3250100fad772e87107f941800f4b3dfbb730371

Btw, can you bring back the underline to the selected tab in the superior bar? That was a nice addition and you rolled it back because of the clipping which this PR solves.

PR #1173, opened 2020-05-03
Migrate deleteHelper, actionSheet, indicators and playMenu to ES6

sarab97:
**Changes**
deletehelper file renamed to follow standard camelCase
deletehelper converted to ES6
actionSheet files and folder renamed to follow standard camelCase
actionSheet file cleaned code and migrate to ES6
playmenu file renamed to follow standard camelCase
playMenu migrate to ES6 

**Issues**
Issue #868

Issue #901, opened 2020-03-07
Library Page Resets Location Marker


Artiume:
While performing some actions (such as clicking Edit Metadata) will reset the location marker to the top of the pagnation instead of maintaining the current location.

Click Identify and then pressing Back will maintain the page location.

Click Edit metadata and then pressing Back will return you to the top of the page.

ferferga:
Fixed in #1442

As discussed in Matrix, the behaviour for edit metadata makes sense as we need to see the changes that have been made

PR #1232, opened 2020-05-14
Fix more sonarqube bugs

MrTimscampi:
## Sonarqube's Bizarre Adventures - Part 3: Stardust Programmers

**Changes**

Fixes all the remaining Sonarqube bugs but two:
* Lack of lang on the `html` tag
* Auto-executing anonymous function in site.js

Those require bigger fixes, so they'll be done at some point in the future.

It also removes all the references to MSIE, since we don't support it.

**Issues**


src/components/actionsheet/actionsheet.js
```diff
@@ -11,20 +11,10 @@ define(['dialogHelper', 'layoutManager', 'globalize', 'browser', 'dom', 'emby-bu
         }
 
         var box;
-        var elem;
+        for (let [index, elem] of elems) {
```
dmitrylyzo, line 14: #1173

src/components/appRouter.js
```diff
@@ -246,13 +246,11 @@ define(['loading', 'globalize', 'events', 'viewManager', 'layoutManager', 'skinM
...
         if (setQuality) {
-
-            var quality = 100;
-
+            var quality;
             var type = options.type || 'Primary';
 
             if (browser.tv || browser.slow) {
-
+                // TODO: wtf
```
dmitrylyzo, line 253: 230932b5a506f7a660e5279ea88b47bc628265e8
```js
var quality=100;
isBackdrop=options.type&&"backdrop"===options.type.toLowerCase();
quality=browser.tv||browser.slow?isBackdrop?60:50:isBackdrop?60:90;
options.quality=quality;
```
1c06eed0985da6bded90c3c26fb99a8ef61c3f86
```js
var quality=100;
type=options.type||"Primary";
quality=browser.tv||browser.slow?browser.chrome?"Primary"===type?40:50:"Backdrop"===type?60:50:"Backdrop"===type?70:90;
options.quality=quality;
```

src/scripts/playlists.js
```diff
@@ -69,10 +69,11 @@ define(['loading', 'listView', 'cardBuilder', 'libraryMenu', 'libraryBrowser', '
             showLoadingMessage();
             var query = getQuery(view);
             var promise1 = ApiClient.getItems(Dashboard.getCurrentUserId(), query);
+            // TODO: promise2 is unused, check if necessary.
             var promise2 = Dashboard.getCurrentUser();
             Promise.all([promise1, promise2]).then(function (responses) {
                 var result = responses[0];
-                responses[1];
+                // TODO: Is the scroll necessary?
                 window.scrollTo(0, 0);
```
dmitrylyzo, line 77: In case of paging, we should go to the top. But in the same time (on some pages), this is unacceptable when we change (or not) metadata - no individual updating. (#901)

dkanada:
> lack of lang on the html tag

I'm guessing we want to fix this with a default to English and then some client side JS to update the value with the browser's language? I haven't looked into this yet.

MrTimscampi:
> > lack of lang on the html tag
> 
> I'm guessing we want to fix this with a default to English and then some client side JS to update the value with the browser's language? I haven't looked into this yet.

Yep that's pretty much it.

PR #1237, opened 2020-05-16
Sort multi-version movie dropdown menu

DesertCookie:
**Changes**

Sorts the array of media sources received by the server in descending order for resolution names like `2160p` and in alphabetically ascending order for names like `Director's Cut`.

This is done by first separating the two name types into separate arrays, sorting them accordingly, then readding them to the array of media sources.  
Names ending with `p` are identified as the first, anything else as the second name type. Sorting the second array of names ignores capitalisation.

**Examples:**

* `1080p`, `2160p`, `360p`, `480p`, `720p` → `2160p`, `1080p`, `720p`, `480p`, `360p`
* `Extended Cut`, `Cinematic Cut`, `Director's Cut` → `Cinematic Cut`, `Director's Cut`, `Extended Cut`
* `Extended Cut - 1080p`, `Extended Cut - 2160p`, `Cinematic Cut - 1080p`, `Cinematic Cut - 2160p` → `Cinematic Cut - 2160p`, `Extended Cut - 2160p`, `Cinematic Cut - 1080p`, `Extended Cut - 1080p`

*Previously discussed on [Reddit](https://www.reddit.com/r/jellyfin/comments/gcn20v/how_are_movie_versions_sorted/).*


MrTimscampi:
Wouldn't this mess up cases like the following?

* 480p
* 720p
* 1080p
* 2160p

DesertCookie:
As of currently, it's already messed up. It would sort 2160p before 1080p in many cases for example.

DesertCookie:
> Wouldn't this mess up cases like the following?

I added some extra logic making sure resolution names like `1080p` are sorted from biggest to smallest.

src/controllers/itemdetailpage.js
```diff
@@ -1,5 +1,5 @@
...
-    'use strict';
+define(["loading", "appRouter", "layoutManager", "connectionManager", "userSettings", "cardBuilder", "datetime", "mediaInfo", "backdrop", "listView", "itemContextMenu", "itemHelper", "dom", "indicators", "imageLoader", "libraryMenu", "globalize", "browser", "events", "playbackManager", "scrollStyles", "emby-itemscontainer", "emby-checkbox", "emby-button", "emby-playstatebutton", "emby-ratingbutton", "emby-scroller", "emby-select"], function (loading, appRouter, layoutManager, connectionManager, userSettings, cardBuilder, datetime, mediaInfo, backdrop, listView, itemContextMenu, itemHelper, dom, indicators, imageLoader, libraryMenu, globalize, browser, events, playbackManager) {
+    "use strict";
```
JustAMan, line 2: You have replaced single quotes with double ones, this fails linting and our current coding style. Please fix this, otherwise it's a whopping of 1200 LoC while in reality it should be rather small.

Artiume:
So it might be beyond this PR but if you have a 2160 and 1080 quality, it'd be nice have a toggle to have 4k content to not be automatically first. I like being able to have 4k, but I can only support a single stream so I don't want to have 4k constantly being first pick.

DesertCookie:
I have this problem exactly, which is why I am naming my movie versions `FHD` and `UHD` which corrects the order for me.

src/controllers/itemdetailpage.js
DesertCookie: Strange. It uses double-quotes in the local file of my 10.5.5 installation. Just replacing them all with single-quotes will do?

Artiume: It's from this
https://github.com/jellyfin/jellyfin-web/pull/1069

Artiume:
UHD - 2160p would be higher than FHD - 1080p, right? And then if omitting the resolution would fix the order?

ferferga:
As I already posted on Matrix, I don't think this is the right approach: we should select by default the item that has whatever target resolution/audio codecs of the device, instead of doing it alphabetically.

DesertCookie:
> And then if omitting the resolution would fix the order?

Yes, because one orders by largest resolution first, the other by alphabet. The logic of looking for resolution strings like `1080p` is very basic (looking for a `p` at the end). To take care of the edge cases you'd need a lot of fancy logic (that I'm not capable of). This way you at least have some control over what is presented first.

> [...] we should select by default the item that has whatever target resolution/audio codecs of the device, instead of doing it alphabetically.

This is purely about ordering them to make sense. The selection process is a different topic. Right now, like before, it automatically selects the first entry of the list. Just now you have some control over what comes first, instead of it being randomly.  
If someone had the means of also implementing what you are asking for though, I'd be the last to not welcome it (even in this PR instead of a separate one).

src/controllers/itemdetailpage.js
ferferga: @DesertCookie it might be easier to use as "base" the version of the master branch, specially now that it diverged a lot from 10.5 as we're heading towards 10.6.

Not only because the double quote changes, but because some other stuff that has been changed since stable might be replaced as well.

ferferga:
@DesertCookie I would do it here, let's see what others think about it first.

We can help you with it, no worries :D

src/controllers/itemdetailpage.js
DesertCookie: I'll refresh my local copy of the repo - it's been a week or two and do the visual changes required.

Artiume:
What about 1080i?

DesertCookie:
> What about 1080i?

That'd be an easy enough fix.

> There's also a bug where if you click Download for a multi-version, it has the same behaviour, it'll download the first available version and not the version picked

I wonder if this is really a Jellyfin-Web problem. I'll scan through the JS.

MrTimscampi:
As outlined in my earlier comment and talked about more in-depth on Matrix, I don't think an alphabetical order fits here.

What we should be doing is sorting resolutions from biggest to lowest (biggest first), then pre-selecting one according to user playback preferences.

So what's needed for a correct fix, imo:

* Make the resolution selector go from largest to lowest
* As part of that, map common identifiers for resolutions that JF would support (HD => 1080, 4K/UHD => 2160)
* Add a setting to Playback in Video Quality, above "Internet quality" that selects a preferred quality for videos with multiple resolutions
* Have the resolution selector pre-select a quality according to that setting, if available. Otherwise, pre-select the highest one

Artiume:
> Add a setting to Playback in Video Quality, above "Internet quality" that selects a preferred quality for videos with multiple resolutions

This would be good to have, perhaps with the default being 1080p. 

Maybe not in this PR, but can we fix the logic requirement where you must use ` - ` to enable multi version mode? It'd be nice to allow familiarly named episodes to auto merge into multi version. Because of the poor UX logic, a lot of people think we don't even support multi version.

MrTimscampi:
> > Add a setting to Playback in Video Quality, above "Internet quality" that selects a preferred quality for videos with multiple resolutions
> 
> This would be good to have, perhaps with the default being 1080p.
> 
> Maybe not in this PR, but can we fix the logic requirement where you must use `-` to enable multi version mode? It'd be nice to allow familiarly named episodes to auto merge into multi version. Because of the poor UX logic, a lot of people think we don't even support multi version.

That'd be server-side, and it's something I want as well :p I don't want to rename my files, but I'd like files that resolve to the same item to be versions of that item instead of duplicates.

Different thing entirely, though :)

DesertCookie:
I thought about having qualifiers like `SD`, `HD`, `FHD`, `QHD`, `UHD` but haven't been able to come up with a nice system. Searching and sorting for these qualifiers would be quite more complex system than the current one. I guess it's actually not too hard to get the current screen resolution in JS and choose the closest available video resolution too.  
To be honest, the main reason I haven't (yet) gone down that road is that it would entail having fifty lines of very specific code that I'd hate to just have in a personal project, let alone push to a public one like this. Am I just to held back by my own insecurity here?

src/controllers/itemdetailpage.js
```diff
@@ -62,7 +62,7 @@ define(['loading', 'appRouter', 'layoutManager', 'connectionManager', 'userSetti
 
     function getProgramScheduleHtml(items) {
         var html = '';
-        html += '<div is="emby-itemscontainer" class="itemsContainer vertical-list" data-contextmenu="false">';
+        html += '<div is='emby-itemscontainer' class='itemsContainer vertical-list' data-contextmenu='false'>';
```
JustAMan, line 65: sorry, but this change is even worse than before %) it won't work!

DesertCookie: I probably should stop using Notepad++ to make my changes. It seems to mess up a lot of things.

DesertCookie: I started with a fresh copy of the file right out of the current master branch and lots of errors are coming up - all things I haven't touched. What's going on here?

src/controllers/itemdetailpage.js
```diff
@@ -146,6 +146,33 @@ define(['loading', 'appRouter', 'layoutManager', 'connectionManager', 'userSetti
             page.querySelector('.trackSelections').classList.remove('hide');
             select.setLabel(globalize.translate('LabelVersion'));
             var currentValue = select.value;
+
+            var resolutionNames = [];
+            var sourceNames = [];
+            mediaSources.forEach(function (v) {
+                if (v.Name.endsWith('p') || v.Name.endsWith('i')) {
```
Artiume, line 153: I think this would make Interlaced appear before Progressive, but meh, im not sure who would keep both I/P copies of the same quality lol

src/controllers/itemdetailpage.js
MrTimscampi: You likely need our ESLint config.

src/controllers/itemdetailpage.js
DesertCookie: Both very true points. I convert all my interlaced video into progressive and can only imagine many others do to for the improvement in visual quality.

src/controllers/itemdetailpage.js
DesertCookie: Clearing the IDE cache solved the "problems". Strange things happen in software world...

crobibero:
@DesertCookie You still have some merge conflicts

DesertCookie:
The long wait broke all kind of things as `itemDetails.js` got moved and renamed and most of the code rewritten. I'm trying to figure out how the new system works and adapt my old solution as best as possible.

thornbill (changes requested):
This was probably a result of a merge conflict, but it looks like this PR is adding an empty `src/controllers/itemDetails.js` file.

src/controllers/itemDetails/index.js
```diff
@@ -167,7 +167,30 @@ function renderTrackSelections(page, instance, item, forceReload) {
         return;
     }
 
-    const mediaSources = item.MediaSources;
+    var mediaSources = item.MediaSources;
+
+    var resolutionNames = [];
+    var sourceNames = [];
+    mediaSources.forEach(function (v) {
+        (v.Name.endsWith('p') || v.Name.endsWith('i')) ? resolutionNames.push(v) : sourceNames.push(v);
```
thornbill, line 175: There should probably be a check here to make sure that `parseInt(v.Name, 10)` returns a number and if not also push that to `sourceNames`. Something like:
```js
(v.Name.endsWith('p') || v.Name.endsWith('i')) && !isNaN(parseInt(v.Name, 10))
```

src/controllers/itemDetails/index.js
```diff
@@ -167,7 +167,30 @@ function renderTrackSelections(page, instance, item, forceReload) {
...
+    });
+
+    resolutionNames.sort((a, b) => parseInt(b.Name) - parseInt(a.Name));
```
thornbill, line 178: You should always specify a radix when calling `parseInt` as the behaviour can be unpredictable without it.

thornbill (changes requested):
Overall seems like a fine enhancement to the current behaviour! I made a couple minor notes and did some testing on [CodePen](https://codepen.io/billt2006/pen/vYGRNMJ?editors=0010)

DesertCookie (commented):
Thanks. I've applied your suggestions.

Edit: The build fails in ways I do not understand. Requesting help.

src/controllers/itemDetails/index.js
```diff
@@ -167,7 +167,30 @@ function renderTrackSelections(page, instance, item, forceReload) {
         return;
     }
 
-    const mediaSources = item.MediaSources;
+    var mediaSources = item.MediaSources;
+
+    var resolutionNames = [];
+    var sourceNames = [];
+    mediaSources.forEach(function (v) {
+        ((v.Name.endsWith('p') || v.Name.endsWith('i'))) && !Number.isNaN(parseInt(v.Name, 10) ? resolutionNames.push(v) : sourceNames.push(v);
```
thornbill, line 175:
```suggestion
        (v.Name.endsWith('p') || v.Name.endsWith('i')) && !Number.isNaN(parseInt(v.Name, 10)) ? resolutionNames.push(v) : sourceNames.push(v);
```
The lint task was failing because there was a missing parenthesis.

thornbill (changes requested):

Artiume:
It's angry again.
```
/home/vsts/work/1/s/src/controllers/itemDetails/index.js
  170:5  error  Unexpected var, use let or const instead  no-var
  172:5  error  Unexpected var, use let or const instead  no-var
  173:5  error  Unexpected var, use let or const instead  no-var
  180:9  error  Unexpected var, use let or const instead  no-var
  181:9  error  Unexpected var, use let or const instead  no-var
```

DesertCookie:
I caused one of these bugs as far as I can tell. Well...

thornbill (approved):

ferferga (approved):

PR #1238, opened 2020-05-16
Remove imageFetcher

ferferga:
Not needed anymore

cc @MrTimscampi


dkanada:
Weren't there also element classes that needed to be removed as well?

PR #1239, opened 2020-05-16
Revert 9a47428 changes

ferferga:
With the changes made to that field, it's impossible to give an IP address at the server page: You're forced to add the HTTP or HTTPS protocol.

firefox

### DIFF
--- a/src/addserver.html
+++ b/src/addserver.html
@@ -3,7 +3,7 @@
         <form class="addServerForm" style="margin: 0 auto;">
             <h1>${HeaderConnectToServer}</h1>
             <div class="inputContainer">
-                <input is="emby-input" type="url" id="txtServerHost" required="required" label="${LabelServerHost}"/>
+                <input is="emby-input" type="text" id="txtServerHost" required="required" label="${LabelServerHost}"/>
                 <div class="fieldDescription">${LabelServerHostHelp}</div>
             </div>
             <br />


src/addserver.html
```diff
@@ -3,7 +3,7 @@
         <form class="addServerForm" style="margin: 0 auto;">
             <h1>${HeaderConnectToServer}</h1>
             <div class="inputContainer">
-                <input is="emby-input" type="url" id="txtServerHost" required="required" label="${LabelServerHost}"/>
+                <input is="emby-input" type="text" id="txtServerHost" required="required" label="${LabelServerHost}"/>
```
dmitrylyzo, line 6: In this case, we should also revert `autocomplete="off" spellcheck="false" autocapitalize="none" autocorrect="off"`
cc @MrTimscampi

MrTimscampi:
It does allow IPs, but you need the protocol:

image

We should require the protocol anyway and keep this, as it does URL validation automatically, which saves us from having to manually check if the URL has a valid format further down the chain.

ferferga:
@MrTimscampi I knew it as specified in my first comment, but I don't think we should to be honest. If I have only an HTTPS server (or viceversa) that doesn't do a redirection, users would need to guess the protocol used, which is something easy for the admin who configured that, but not for family, for example. If users input something illegal there, they would already receive a "Couldn't establish a connection message" that should be enough most of the cases.

Also, addresses like jellyfin.org are illegal there, and all browsers are enforcing the deprecation of the whole protocol and www sequence in their address bars.

Imho, with that additional check we're making connections a bit more painful because you're enforcing the presence of the protocol.

dkanada:
> Also, addresses like jellyfin.org are illegal there, and all browsers are going towards enforcing the deprecation of the whole protocol and www sequence in their address bars.

As far as I know, only Chrome has done this so far, and I consider it one of the worst decisions they've made in a while.

MrTimscampi:
> **enforcing the deprecation of the whole protocol and www sequence in their address bars**

They haven't, though.

`www` not being there is a preference from companies, usually. It's not strictly necessary and many omit it (while having a CNAME for www still for their main domain).

As for the protocol, it's very much not being deprecated. It's for aesthetic reasons and they imply it with the lock icon that represents encryption. There were talks about deprecating straight HTTP, but that is still far away and not really a concern for now.

To get back to the PR itself, as talked about on Matrix, I'm against this change for the following reasons:

* This type of field lets accessibility tools know what the content of the field should be
* It provides built-in validation
* It provides a better UX on devices with an on-screen keyboard (shortcuts for protocols, TLDs and such)
* It remembers the previous URLs used (Making multi-server easier on desktop)
* It reduces the chances of encountering a current bug in the API client where Chrome-based browsers will fail to connect to a URL without protocol if the server uses a reverse proxy with a redirection to HTTPS

The only con to this is requiring that the users input the protocol, which we shouldn't be assuming in the first place anyway, imo.

ferferga:
@MrTimscampi We need a jellyfin-apiclient update for the latest changes to work, right?

MrTimscampi:
> @MrTimscampi We need a jellyfin-apiclient update for the latest changes to work, right?

In which way? The apiclient will already add a protocol if not present (Previously HTTP, now HTTPS).

It should eventually be changed to try HTTPS then try HTTP, but that's outside the scope of this PR.

PR #1240, opened 2020-05-16
Remove unused image types

MrTimscampi:
**Changes**

Removes unused image types from the web client (ie images we don't surface anywhere):

* Menu
* Box
* Box (rear)
* Disc

**Issues**

### DIFF
--- a/src/components/cardbuilder/cardBuilder.js
+++ b/src/components/cardbuilder/cardBuilder.js
@@ -522,30 +522,6 @@ import 'programStyles';
                     tag: item.ImageTags.Banner
                 });
 
-            } else if (options.preferDisc && item.ImageTags && item.ImageTags.Disc) {
-
-                imgUrl = apiClient.getScaledImageUrl(item.Id, {
-                    type: 'Disc',
-                    maxWidth: width,
-                    tag: item.ImageTags.Disc
-                });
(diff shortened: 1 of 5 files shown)


MrTimscampi:
Note: there is also "Art", which has no reference in web and is unused, but I'm not sure what it is, so leaving it for now.

Artiume:
I'm still on the fence about this cleanup. I dont have any disc images at the moment but I know those who collect them

MrTimscampi:
> I'm still on the fence about this cleanup. I don't have any disc images at the moment but I know those who collect them

They're unused and none of our clients surface them, except maybe Kodi. Leaving them enabled is kind of misleading, imo, and Kodi can probably use its own artwork providers to fetch these if wanted.

dkanada:
So what images are we actually keeping then?

mcarlton00:
That I know of, Kodi only utilizes Poster, Backdrop, and Banner depending on the skin

Artiume:
We might eventually get a metadata source that provides these images. Or there might be those that use it now in their own systems. They aren't hurting anyone and I think they should just stay until we have a reason to remove them

dkanada:
I think I'll vote for leaving them in right now.

PR #1242, opened 2020-05-16
Implement select menu for chromecast version

dkanada:


ferferga:
What's holding us back from using the nightly one as our stable?

dkanada:
A lack of testing mainly, which is what this pull request aims to solve.

dkanada:
In the future the nightly will also get built off every single merge to the chromecast repository as well so it will have value past the latest update.

PR #1251, opened 2020-05-17
Fix Quality Settings for Aspect Ratios less than 16:9 (1.77)

### LINKED ISSUE #1250: Transcoding 4:3 sources does not show correct available resolutions.
When transcoding sources less than 16:9, the correct height options are not available.
E.g 1440x1080p video does not show options for 1080p XMbps, only 720p and below
To Reproduce

Play 4:3 1080p video (or other aspect ratios less than 16:9)
1080p Options not available

Expected behavior
1080p should be available

System (please complete the following information):

    OS: Windows 10
    Browser: Firefox, Chrome
    Jellyfin Version: 10.5.5

Additional context

adavier:
**Changes**
qualityoptions.js now checks if the aspect ratio is less than 16:9 (1.77), and if so pads the width to 16:9, in order to show correct options for content narrower than 16:9 (1.77) such as 4:3 (1.33).

**Issues**

Fixes #1250

### DIFF
--- a/src/components/qualityOptions.js
+++ b/src/components/qualityOptions.js
@@ -9,8 +9,8 @@ define(['globalize'], function (globalize) {
 
         //If the aspect ratio is less than 16/9 (1.77), set the height as if it were pillerboxed.
         // 4:3 1440x1080 -> 1920x1080
-        if (videoWidth/videoHeight < 16/9) {
-            videoWidth=videoHeight*(16/9);
+        if (videoWidth / videoHeight < 16 / 9) {
+            videoWidth = videoHeight * (16 / 9);
         }
 
         var maxAllowedWidth = videoWidth || 4096;
--- a/src/components/qualityOptions.js
+++ b/src/components/qualityOptions.js
@@ -5,6 +5,13 @@ define(['globalize'], function (globalize) {
 
         var maxStreamingBitrate = options.currentMaxBitrate;
         var videoWidth = options.videoWidth;
+        var videoHeight = options.videoHeight;
+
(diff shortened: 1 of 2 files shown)


src/components/qualityOptions.js
```diff
@@ -5,6 +5,13 @@ define(['globalize'], function (globalize) {
 
         var maxStreamingBitrate = options.currentMaxBitrate;
         var videoWidth = options.videoWidth;
+        var videoHeight = options.videoHeight;
+
+        //If the aspect ratio is less than 16/9 (1.77), set the height as if it were pillerboxed.
+        // 4:3 1440x1080 -> 1920x1080
+        if (videoWidth/videoHeight < 16/9) {
+            videoWidth=videoHeight*(16/9);
+        }
```
MrTimscampi, line 14:
```suggestion
        if (videoWidth / videoHeight < 16 / 9) {
            videoWidth = videoHeight * (16 / 9);
        }
```

src/components/qualityOptions.js
```diff
@@ -5,6 +5,13 @@ define(['globalize'], function (globalize) {
 
         var maxStreamingBitrate = options.currentMaxBitrate;
         var videoWidth = options.videoWidth;
+        var videoHeight = options.videoHeight;
+
+        //If the aspect ratio is less than 16/9 (1.77), set the height as if it were pillerboxed.
```
MrTimscampi, line 10:
```suggestion
        //If the aspect ratio is less than 16/9 (1.77), set the width as if it were pillarboxed.
```

You're setting the with, not the height :)

PR #1255, opened 2020-05-17
Do not translate the "General" external id type

mark-monteiro:
**Changes**
This accompanies the associated server-side change in https://github.com/jellyfin/jellyfin/pull/3137 that adds a "General" external id type.

The new "General" value is returned instead of null. It does not need to be translated so the translation logic is changed here to ignore that value.


mark-monteiro:
I've made some changes to the server PR so that this is no longer necessary.

PR #1266, opened 2020-05-18
Move plugins to a dedicated directory

MrTimscampi:
**Changes**

This moves all the plugins to a directory specifically for them, away from components.

**Issues**


dkanada:
At least four of the plugins have other assets that should probably move to the new locations.

dkanada:
The `chromecastHelpers` file might also be better in the new location.

src/scripts/site.js
```diff
@@ -479,21 +479,21 @@ var AppInfo = {};
...
+            'plugins/youtubePlayer/plugin',
+            'plugins/backdropScreensaver/plugin',
+            'plugins/logoScreensaver/plugin'
         ];
 
         if (appHost.supports('remotecontrol')) {
-            list.push('components/sessionPlayer');
+            list.push('plugins/sessionPlayer/plugin');
 
             if (browser.chrome || browser.opera) {
```
ferferga, line 495: Completely unrelated, but we can also add the new Edge Chromium there I suppose. Do you know if it's under the "Chrome" spectrum or it has its own name?

MrTimscampi: I don't think we detect it specifically. I don't have it, so no idea what it'd show up as. Someone who has Windows available should check.

dkanada:
Oh, I thought this one had already been merged. Looks like there are conflicts again.

PR #1089, opened 2020-04-15
Migrating backdropscreensaver, filterdialog and fetchhelper to ES6

cromefire:
**Changes**
Migrating:
 - src/components/backdropScreensaver/plugin.js
 - src/components/filterdialog/filterdialog.js
 - src/components/fetchhelper.js

**Issues**
Relates to: #868

PR #1269, opened 2020-05-18
Convert photo player to class

dkanada:
Here's an extremely small player we can use as a base to convert the rest of them.


dkanada:
This should be merged in tandem with #1263 for the initialization change.

src/components/photoPlayer/plugin.js
```diff
@@ -1,19 +1,22 @@
...
+        this.name = 'Photo Player';
+        this.type = 'mediaplayer';
+        this.id = 'photoplayer';
+        this.priority = 1;
     }
 
-    PhotoPlayer.prototype.play = function (options) {
+    play(options) {
 
         return new Promise(function (resolve, reject) {
```
MrTimscampi, line 21: We can replace the require() on 23 with import(), the Babel transform will convert it. Then you can remove the "require" dependency.

src/components/photoPlayer/plugin.js
```diff
@@ -1,19 +1,22 @@
...
-
-        var self = this;
-
-        self.name = 'Photo Player';
-        self.type = 'mediaplayer';
-        self.id = 'photoplayer';
-
-        // Let any players created by plugins take priority
-        self.priority = 1;
+import browser from 'browser';
```
MrTimscampi, line 1: Unused dependency.

src/components/photoPlayer/plugin.js
```diff
@@ -1,19 +1,22 @@
...
+import require from 'require';
+import events from 'events';
```
MrTimscampi, line 3: Unused dependency.

src/components/photoPlayer/plugin.js
```diff
@@ -1,19 +1,22 @@
...
+import appHost from 'apphost';
```
MrTimscampi, line 4: Unused dependency.

src/components/photoPlayer/plugin.js
```diff
@@ -1,19 +1,22 @@
...
+import loading from 'loading';
```
MrTimscampi, line 5: Unused dependency.

src/components/photoPlayer/plugin.js
```diff
@@ -1,19 +1,22 @@
...
+import dom from 'dom';
```
MrTimscampi, line 6: Unused dependency.

src/components/photoPlayer/plugin.js
```diff
@@ -1,19 +1,22 @@
...
+import playbackManager from 'playbackManager';
```
MrTimscampi, line 7: Unused dependency.

src/components/photoPlayer/plugin.js
```diff
@@ -1,19 +1,22 @@
...
+import appRouter from 'appRouter';
```
MrTimscampi, line 8: Unused dependency.

MrTimscampi:
@dkanada This needs the plugin loader change (to handle Promises)

src/components/photoPlayer/plugin.js
```diff
@@ -1,23 +1,18 @@
...
+        this.priority = 1;
     }
 
-    PhotoPlayer.prototype.play = function (options) {
+    play(options) {
 
         return new Promise(function (resolve, reject) {
 
-            require(['slideshow'], function (slideshow) {
+            import('slideshow').then(slideshow => {
```
dmitrylyzo, line 15: _Same as in #1089_
`TypeError: slideshow is not a constructor`
`slideshow = {default: ƒ}`

```suggestion
            import('slideshow').then(({default: slideshow}) => {
```

src/components/photoPlayer/plugin.js
```diff
@@ -1,23 +1,18 @@
...
-
-        self.name = 'Photo Player';
-        self.type = 'mediaplayer';
-        self.id = 'photoplayer';
-
-        // Let any players created by plugins take priority
-        self.priority = 1;
+import connectionManager from 'connectionManager';
+
+export class PhotoPlayer {
```
thornbill, line 3: Should we specify this as the `default` export?

src/components/photoPlayer/plugin.js
dmitrylyzo: Or here is an alternative suggestion https://github.com/jellyfin/jellyfin-web/pull/1089#discussion_r432112537
_if possible_

src/components/photoPlayer/plugin.js
dkanada: I used the same format we seem to be using everywhere else, which is adding it to the bottom of the file.

thornbill: That seems odd. Why not just do `export default class PhotoPlayer`?

dkanada: @thornbill I can add that in a separate pull request if that's the only other reservation you had about the changes here.

dkanada: I just checked and this is actually the only instance, just removed it now.

dmitrylyzo: > I just checked and this is actually the only instance, just removed it now.

You are not the only one who has exported `default` at the end of file: #1089 (`backdropScreensaver`)

dkanada: Ah, well we haven't merged that yet so there's no problem quite yet.

PR #1271, opened 2020-05-19
Rename Mixed Content and TV Shows

Artiume:
**Changes**

**Issues**

### DIFF
--- a/src/strings/en-us.json
+++ b/src/strings/en-us.json
@@ -273,7 +273,7 @@
     "FolderTypeMusic": "Music",
     "FolderTypeMusicVideos": "Music Videos",
     "FolderTypeTvShows": "TV Shows",
-    "FolderTypeUnset": "Mixed Content",
+    "FolderTypeUnset": "Custom",
     "Folders": "Folders",
     "FormatValue": "Format: {0}",
     "Friday": "Friday",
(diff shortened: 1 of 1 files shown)


Artiume:
This should probably be done via Weblate ctually.

dkanada:
If we're making this change, I propose we also rename TV Shows to Shows for a slightly more generic term. If we don't make this change then disregard my comment.

Artiume:
Renamed TV Shows to Shows

thornbill:
“Custom” just seems so ambiguous. I believe that type is for unorganized combinations of tv shows and movies because it still scrapes metadata right?

I seem to recall you being vetoed on the “shows” change in the past @dkanada. Although I don’t personally have a preference on that one.

dkanada:
I was at the time because users were confused with the change, but if we're apparently renaming mixed content I figured I'd bring it up again.

Artiume:
Mixed Content's current job as mixing tv shows and movies doesnt work well and there's no real plans for us to fix the regex to make it perfect. I think Custom is a better approach because it's the actual intent of the Library, I do think we should change the metadata grabbers default to off so that you can enable them if you want, but they're OFF because they're not the real intent. It'll make more sense once we have additional metadata grabbers like YouTube and such.

PR #1275, opened 2020-05-19
Make DVR heading translatable.

### LINKED ISSUE #1270: DVR heading is not translateable
https://github.com/jellyfin/jellyfin-web/blob/e177198939d6a61a8598494d91f8a2103fcfb554/src/livetvsettings.html#L6

This heading is hard-coded and cannot be translated.

KucharczykL:
The heading of the DVR section was hard-coded and couldn't be translated.

Fixes #1270.

### DIFF
--- a/src/livetvsettings.html
+++ b/src/livetvsettings.html
@@ -3,7 +3,7 @@
         <div class="content-primary">
             <div class="verticalSection">
                 <div class="sectionTitleContainer flex align-items-center">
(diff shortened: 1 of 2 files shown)


ferferga:
Shouldn't the string be added to the string files? Impossible to translate atm

KucharczykL:
I can add the string manually (confirmed locally) but I'm not sure if that's the proper way of doing things. I hope someone else can chime in.

PR #1285, opened 2020-05-22
Remove restart and shutdown buttons from dashboard

dkanada:


PrplHaz4:
Why? Shutdown works fine on docker and is the only way to install plugins without diving into a terminal or different app.

Artiume:
The button doesn't work on Docker. It `restarts` on Docker. There's no consistent way to create expected behavior with the button especially between linux and windows.

MrTimscampi:
> Why? Shutdown works fine on docker and is the only way to install plugins without diving into a terminal or different app.

As Artiume implied, we've been thinking about this for a while due to differences between systems and the way that this needs to be done being less than ideal.

It's very hacky, as it relies on external scripts to control the services. These scripts are error-prone, difficult to maintain (there basically needs to be one per platform. On Linux, if you're not using the init we expect (systemd) it won't work. It doesn't work with Windows in Service mode, afaik, and who knows what else.

PrplHaz4:
The button does shutdown the service successfully. Then the docker engine starts it again. I'd expect any platform running as a service to behave the same way. 

I see the inconsistency due to semantics but this seems to be removing necessary (for plugins) functionality on 2 of 3 platforms.

PR #987, opened 2020-03-26
Blurhash support

GranPC:
**Changes**
Adds support for decoding blurhashes and using them as placeholders while the real images load.

**Issues**
Fixes jellyfin/jellyfin#2569

PR #1286, opened 2020-05-23
Implement blurhash (follow-up of #987 but from scratch)

ferferga:
Follow-up of #987 but done from scratch, as the changes to the lazy loader were breaking that PR.

**Differences & key changes**
* **Fully backwards compatible** (with minor changes, see below)
* Probably less code duplication in ``cardBuilder.js``
* Reimplementation to **work with the latest lazy loader changes**
* Blurhash is inside a canvas, a child of the image element. This allows for more customization, **as the canvas has its own class (so it can be easily customizable by themes)**
* Removal of the original lazy classes, as now we're going to use a custom one for blurhash, which allows for beautiful animations when the connection is really slow (Fast 3G in Chrome):

~~**The removal of the old lazy load classes has one tradeoff**: *images that doesn't have blurhash generated, as the movie "Bajo el mismo techo" in the above video, won't have animations*~~ Fixed in 1591aef

For reference, here is a video on fast networks:

**To do**:

* ~~Create the canvas asynchronously, so it doesn't freeze the UI. I suppose promises are the way to go for this, but I need some feedback, as I never have done asynchronous JS before. **I'm all ears for whatever advice you can give me in this matter :)**~~
* ~~Implement it in *listview*~~
* Add settings

**Needs discussion:**
There are many things that I want to discuss and have some feedback:

* ~~As you can see in my code's comments, I started blurhashing items way too far or completely outside the intersectionRatio <= 0.025). Most users won't need it as home networks are really quick to load the first rows of images quickly. However, I thought that blurhashing items already visible in the viewport should be made for mobile or slow networks, so we can get the loading effect. **Shall we keep this?**~~ All items will be blurhashed straightaway.

* ~~With the current pagination limits, the page takes around 50 MB with all the 100 items' blurhash canvas present. As you can see in the code's comments, I already did some tests with blurhash removal, so the memory usage doesn't go crazy. However, there are many ways to approach this, each one with their *pros and cons*.~~ **Blurhash memory footprint is really low, even for large libraries**

**Issues:**
Closes https://github.com/jellyfin/jellyfin/issues/2569


EraYaN:
Seems like removing the canvas when the image has been loaded (would a "regular" onload event work?) I probably the cleanest from a DOM perspective? But I'm not really sure if that would work since I didn't really look at the code too much.

ferferga:
@EraYaN That would return the black box that appears when scrolling out of the images and coming back. Although I haven't tried with disabled animations.

Maybe we will need to increase the range of the intersection observer, so it also watches for stuff that is a little outside the viewport.

src/components/images/imageLoader.js
```diff
@@ -11,6 +12,85 @@ import 'css!./style';
...
+            // if (document.getElementsByClassName('blurhashed').length <= 80) {
+
+            //} else {
+                // destroyBlurhash(target);
+            //}
+            let blurhashstr = target.getAttribute('data-blurhash');
+            if (blurhashstr && blurhash.isBlurhashValid(blurhashstr) && target.getElementsByClassName('blurhash-canvas').length === 0) {
+                console.log('Blurhashing item ' + target.parentElement.parentElement.parentElement.getAttribute('data-index') + ' with intersection ratio ' + entry.intersectionRatio);
+                let width = target.offsetWidth;
+                let height = target.offsetHeight;
```
MrTimscampi, line 47: Part of the performance issues come from this.

You are decoding it to a full size, which is much more expensive.

Per the README of Blurhash:
```
Similarly, when displaying the placeholders, very small images work very well when scaled up. We usually decode placeholders that are 32 or even 20 pixels wide, and then let the UI layer scale them up, which is indistinguishable from decoding them at full size.
```

So the ideal implementation here would be to compute the aspect ratio of the element, then set the width to 32 and use the ratio to get the height of the element.

This should significantly reduce the cost of blurhash.

ferferga: Missed those bits. Let me test that out. Thank you very much!

ferferga:
Seems that 18x18 is the sweetest spot between blazing fast speed and quality. I will keep doing tests with values and get workers up and running, although imo, not that necessary as before. I would say that this, in its current state, can be shipped for 10.6.

MrTimscampi:
>     * Removal of the original lazy classes, as now we're going to use a custom one for blurhash, which allows for beautiful animations when the connection is really slow (Fast 3G in Chrome):

As we're planning to make Blurhash optional, we should keep the current animations, imo.

src/components/cardbuilder/cardBuilder.js
```diff
@@ -505,6 +505,9 @@ import 'programStyles';
             let imgUrl = null;
             let coverImage = false;
             let uiAspect = null;
+            let blurhash;
```
MrTimscampi, line 508: Why do you initialize this here when it's first used at L730?

src/components/cardbuilder/cardBuilder.js
```diff
@@ -701,11 +723,15 @@ import 'programStyles';
...
                 });
+                imgtag = item.ParentBackdropImageTags[0];
 
             }
 
+            blurhash = imageLoader.getImageBlurhashStr(blurhashimg, imgtag);
+
             return {
                 imgUrl: imgUrl,
+                blurhash: blurhash,
```
MrTimscampi, line 734:
```suggestion
            return {
                imgUrl: imgUrl,
                blurhash: imageLoader.getImageBlurhashStr(blurhashimg, imgtag);,
```

You can do this directly, since the blurhash variable isn't used anywhere else.

src/components/images/imageLoader.js
```diff
@@ -45,13 +127,8 @@ import 'css!./style';
                 elem.setAttribute('src', url);
             }
 
-            if (userSettings.enableFastFadein()) {
-                elem.classList.add('lazy-image-fadein-fast');
-            } else {
-                elem.classList.add('lazy-image-fadein');
-            }
```
MrTimscampi, line 52: This breaks the animation if no blurhash exists or if the user disables them (which is planned).

ferferga:
> > ```
> > * Removal of the original lazy classes, as now we're going to use a custom one for blurhash, which allows for beautiful animations when the connection is really slow (Fast 3G in Chrome):
> > ```
> 
> As we're planning to make Blurhash optional, we should keep the current animations, imo.

@MrTimscampi Did we reach any agreement if we're going to make them default or opt-in?

src/components/cardbuilder/cardBuilder.js
ferferga: I guess it's cleaner? But good point.

JustAMan:
I would rather we keep the animations, and to make possible for a theme (by css) or a user (by settings) or device (by being slow) to turn blurhashes off.

JustAMan:
@ferferga I've adapted the code to my API changes in server side (plus I reduced the copy-pasteness a bit).

src/components/images/imageLoader.js
```diff
@@ -12,13 +12,6 @@ import 'css!./style';
         fillImageElement(elem, source);
     }
 
-    export function getImageBlurhashStr(hashes, tags) {
```
ferferga: @JustAMan You shouldn't have removed this one, because we still need to implement this check in listview

JustAMan: Wait, you cannot use same one-liner I used in other place?.. it literally checks if `hashes` map is not null and gets a value from it... `(hashes || {})[tags]` does _exactly the same_ and is shorter than this function signature 😛

ferferga:
@JustAMan All blurhash elements have their own class, so they can be customized by themes. That must be a setting, as with CSS we can't change the fact that blurhashed items are still being generated (users that probably don't want blurhash won't want them generated, not only hidden by CSS.

With the animations, not sure if you tried this branch, but they're fully animated. Not if the images aren't blurhashed. I'll bring the old ones back and keep them as a fallback for those cases and cases where users doesn't want blurhash.

src/components/images/imageLoader.js
```diff
@@ -11,6 +12,76 @@ import 'css!./style';
...
 
+    // function destroyBlurhash(target) {
+    //     let canvas = target.getElementsByClassName('blurhash-canvas')[0];
+    //     target.removeChild(canvas);
+    //     target.classList.remove('blurhashed');
+    // }
+
+    function itemBlurhashing(entry) {
+        // This intersection ratio ensures that items that are near the borders are also blurhashed, alongside items that are outside the viewport
+        // if (entry.intersectionRation <= 0.025)
```
JustAMan, line 23:
```suggestion
```

src/components/images/imageLoader.js
```diff
@@ -11,6 +12,76 @@ import 'css!./style';
...
+        if (entry.target) {
+            let target = entry.target;
+            // We only keep max 80 items blurhashed in screen to save memory
+            // if (document.getElementsByClassName('blurhashed').length <= 80) {
+
+            //} else {
+                // destroyBlurhash(target);
+            //}
```
JustAMan, line 31:
```suggestion
```

src/components/images/imageLoader.js
```diff
@@ -11,6 +12,76 @@ import 'css!./style';
...
+            let blurhashstr = target.getAttribute('data-blurhash');
+            if (blurhash.isBlurhashValid(blurhashstr) && target.getElementsByClassName('blurhash-canvas').length === 0) {
+                console.log('Blurhashing item ' + target.parentElement.parentElement.parentElement.getAttribute('data-index') + ' with intersection ratio ' + entry.intersectionRatio);
+                // let width = target.offsetWidth;
+                // let height = target.offsetHeight;
```
JustAMan, line 36:
```suggestion
                console.debug('Blurhashing item ' + target.parentElement.parentElement.parentElement.getAttribute('data-index') + ' with intersection ratio ' + entry.intersectionRatio);
```

src/components/images/imageLoader.js
```diff
@@ -11,6 +12,76 @@ import 'css!./style';
...
+                let width = 18;
+                let height = 18;
+                if (width && height) {
```
JustAMan, line 39: this `if` is not needed now because you define `width` and `height` yourself as constants... also width & height should probably follow image aspect ratio if that is known beforehand

src/components/images/imageLoader.js
```diff
@@ -11,6 +12,76 @@ import 'css!./style';
...
+                        child.classList.add('lazy-blurhash-fadein-fast');
+                    } else {
+                        child.classList.add('lazy-blurhash-fadein');
+                    }
+
+                    target.classList.add('blurhashed');
+                }
+            }
+        }
+        return;
```
JustAMan, line 70:
```suggestion
```

src/components/images/imageLoader.js
```diff
@@ -11,6 +12,76 @@ import 'css!./style';
...
+                // let width = target.offsetWidth;
+                // let height = target.offsetHeight;
+                let width = 18;
+                let height = 18;
+                if (width && height) {
+                    let pixels;
+                    try {
+                        pixels = blurhash.decode(blurhashstr, width, height);
+                    } catch (err) {
+                        console.log('Blurhash decode error: ' + err.toString());
```
JustAMan, line 44:
```suggestion
                        console.error('Blurhash decode error', err);
```

src/components/images/imageLoader.js
```diff
@@ -11,6 +12,76 @@ import 'css!./style';
...
+    }
+
+    function switchCanvas(elem) {
+        let child = elem.getElementsByClassName('blurhash-canvas')[0];
+        if (child) {
+            if (elem.getAttribute('data-src')) {
+                child.style.opacity = 1;
+            } else {
+                child.style.opacity = 0;
+            }
```
JustAMan, line 80:
```suggestion
            child.style.opacity = elem.getAttribute('data-src') ? 1 : 0;
```

src/components/images/imageLoader.js
```diff
@@ -11,6 +12,76 @@ import 'css!./style';
...
+        }
+        return;
```
JustAMan, line 82:
```suggestion
```

JustAMan:
> users that probably don't want blurhash won't want them generated, not only hidden by CSS

Not generating blurhashes if they're not enabled by the user is optimisation. What I'm trying to achieve here is user being able to not see those if they don't like blurred stuff (so it's logic, not optimisation). For a first pass it's perfectly fine for my taste to compute those but keep them hidden.

src/components/images/imageLoader.js
MrTimscampi: > also width & height should probably follow image aspect ratio if that is known beforehand

We should at least have the card shape at that point, so we have an estimation of it, at least.

KristupasSavickas:
>Not generating blurhashes if they're not enabled by the user is optimisation. What I'm trying to achieve here is user being able to not see those if they don't like blurred stuff (so it's logic, not optimisation). For a first pass it's perfectly fine for my taste to compute those but keep them hidden.

ferferga:
If you're coming across this PR, don't pay too much attentions at reviewing the stuff just yet, it's not complete. I prefer some testing for now to ensure that all blurhashes are good. i would also like to get some feedback in the canvas sizes values.

Once I remove all my comments and testing stuff and I think it's good to be reviewed, I'll request a review from the team.

From now on, [please test what I say here](https://github.com/jellyfin/jellyfin-web/pull/1286#issuecomment-633645764)

src/components/images/imageLoader.js
ferferga: The estimation is not needed for anything anyway, because the canvas will always fill the size using CSS. I think it's safe to remove and follow @JustAMan suggestions.

JustAMan: No, my comment was about making canvas have proper aspect ratio so it's scaled "more correctly". It's work regardless, but would look prettier if scaled uniformly.

ferferga: @JustAMan @MrTimscampi Everything I do here is based in https://blurhash-wasm.netlify.app/ implementation. They don't scale the canvas at all, they just use CSS for that. Blurhash implementation in their page is exactly the same.

Canvas and pixels must have the same width & height as well: otherwise the canvas won't be 100% covered (which is what we discussed on Matrix)

ferferga:
The client-side of blurhash is now ready to be merged imo. **There are issues with people's cards that still needs fixing server-side but I don't think they will require any API change in client**. And, if they do, I expect them to be really minor.

- **There is a little quirk when switching between blurhash and no blurhash**: Images that are already cached, will instantly appear in the viewport, without animations. That happens because the switching between the class that sets opacity to 0 and the one that sets opacity to 1 once the image is loaded happens before the event loop that refreshes the DOM has been fired. However, in slow connections, the images will still appear animated. Also when scrolling down and up to them again.

    It's a minor inconvenience but, it's non-blocking imo.

src/components/displaySettings/displaySettings.template.html
```diff
@@ -143,20 +143,28 @@ <h2 class="sectionTitle">
         <select is="emby-select" class="selectSoundEffects" label="${LabelSoundEffects}"></select>
     </div>
 
-    <div class="inputContainer inputContainer-withDescription fldFadein">
```
ferferga, line 146: Outside of the scope of this PR but it's a very minor thing to spend the time opening new PR and fixing conflicts between this one. Those classes are not used in JS, so they are useless.

src/strings/en-us.json
```diff
@@ -826,8 +826,8 @@
     "LabelSaveLocalMetadataHelp": "Saving artwork into media folders will put them in a place where they can be easily edited.",
     "LabelScheduledTaskLastRan": "Last ran {0}, taking {1}.",
     "LabelScreensaver": "Screensaver:",
-    "EnableFastImageFadeIn": "Fast Image Fade Animations",
-    "EnableFastImageFadeInHelp": "Show posters and other images with a quicker fade animation when they finish loading.",
+    "EnableFastImageFadeIn": "Faster Animations",
```
ferferga: Also, a little bit outisde the scope of this PR, but I decided to change those because we're no longer talking about images, but also the placeholder below, as this toggle controls the speed of blurhashes and images animations (whatever you have enabled). We've been also talking many times about the animation times of other UI components and the possibility of making them configurable, etc... in Matrix. Until we get to the point where we make that configurable, I think a single (and straight to the point) checkbox to rule them all is the best idea.

src/strings/en-us.json
```diff
@@ -826,8 +826,8 @@
     "LabelSaveLocalMetadataHelp": "Saving artwork into media folders will put them in a place where they can be easily edited.",
     "LabelScheduledTaskLastRan": "Last ran {0}, taking {1}.",
     "LabelScreensaver": "Screensaver:",
-    "EnableFastImageFadeIn": "Fast Image Fade Animations",
-    "EnableFastImageFadeInHelp": "Show posters and other images with a quicker fade animation when they finish loading.",
+    "EnableFastImageFadeIn": "Faster animations",
```
ferferga, line 829: Also, a little bit outside the scope of this PR, but I decided to change those because we're no longer talking about images, but also the placeholder below, as this toggle controls the speed of blurhashes and images animations (whatever you have enabled). We've been also talking many times about the animation times of other UI components and the possibility of making them configurable, etc... in Matrix. Until we get to the point where we make that configurable, I think a single (and straight to the point) checkbox to rule them all is the best idea.

src/components/images/imageLoader.js
```diff
@@ -11,22 +12,70 @@ import 'css!./style';
...
-            source = entry.target.getAttribute('data-src');
+
+        if (target) {
+            source = target.getAttribute('data-src');
         } else {
             source = entry;
         }
 
+        if (!target.classList.contains('blurhashed') && userSettings.enableBlurhash()) {
+            itemBlurhashing(target);
```
JustAMan, line 73:
```suggestion
            itemBlurhashing(target);
        }
```
you seem to be missing the closing bracket here

src/components/images/imageLoader.js
```diff
@@ -38,20 +87,27 @@ import 'css!./style';
         let preloaderImg = new Image();
         preloaderImg.src = url;
 
+        // This is necessary, so switching between blurhash enabled and disabled works
+        if (!userSettings.enableBlurhash()) elem.classList.add('lazy-hidden');
```
JustAMan, line 91:
```suggestion
        if (!userSettings.enableBlurhash()) {
            elem.classList.add('lazy-hidden');
        }
```

src/components/listview/listview.js
```diff
@@ -70,52 +70,49 @@ define(['itemHelper', 'mediaInfo', 'indicators', 'connectionManager', 'layoutMan
...
-
             options.tag = item.ParentPrimaryImageTag;
-            return apiClient.getScaledImageUrl(item.ParentPrimaryImageItemId, options);
+            itemId = item.ParentPrimaryImageItemId;
         }
+        let blurHashes = item.ImageBlurHashes || {};
+        let blurhashstr = (blurHashes[options.type] || {})[options.tag];
 
-        return null;
+        return { url: apiClient.getScaledImageUrl(itemId, options), blurhash: blurhashstr};
```
JustAMan, line 98: this doesn't handle the case when `itemId` is not filled, but it probably should.

src/components/listview/listview.js
```diff
@@ -70,52 +70,49 @@ define(['itemHelper', 'mediaInfo', 'indicators', 'connectionManager', 'layoutMan
...
         if (item.ChannelId && item.ChannelPrimaryImageTag) {
-
             options.tag = item.ChannelPrimaryImageTag;
-            return apiClient.getScaledImageUrl(item.ChannelId, options);
         }
+        let blurHashes = item.ImageBlurHashes || {};
+        let blurhashstr = (blurHashes[options.type])[options.tag];
 
-        return null;
+        return { url: apiClient.getScaledImageUrl(item.ChannelId, options), blurhash: blurhashstr};
```
JustAMan, line 115: same

src/components/listview/listview.js
ferferga: @JustAMan Not sure if I should (for style/consistency) do an inline check or how I've done it is good.

src/components/cardbuilder/cardBuilder.js
```diff
@@ -624,88 +568,56 @@ import 'programStyles';
...
+                imgTag = item.ParentThumbImageTag;
             } else if (item.ParentBackdropImageTags && item.ParentBackdropImageTags.length && options.inheritThumb !== false) {
+                imgType = 'Backdrop';
+                imgTag = item.ParentBackdropImageTags[0];
+            }
 
-                imgUrl = apiClient.getScaledImageUrl(item.ParentBackdropItemId, {
-                    type: 'Backdrop',
+            if (imgTag && imgType) {
+                imgUrl = apiClient.getScaledImageUrl(item.Id, {
```
dmitrylyzo, line 608: Doesn't it matter?
```
item.Id
item.ParentBackdropItemId
item.SeriesId
item.AlbumId
item.ParentThumbItemId
```

MrTimscampi: I think this is what is causing my display issue, as episodes in Next Up and Continue Watching don't load their proper image, with a 404 error ("<title of the episode> doesn't have an image of type Backdrop")

ferferga: @MrTimscampi I think I was able to reproduce what you say, but thought it was a server issue. Can you check with my latest commit, please?

src/components/cardbuilder/cardBuilder.js
```diff
@@ -624,88 +574,66 @@ import 'programStyles';
...
+                imgTag = item.ParentThumbImageTag;
+                itemId = item.ParentThumbItemId;
             } else if (item.ParentBackdropImageTags && item.ParentBackdropImageTags.length && options.inheritThumb !== false) {
+                imgType = 'Backdrop';
+                imgTag = item.ParentBackdropImageTags[0];
+                itemId = item.ParentBackdropItemId;
+            }
+
+            if (!itemId) {
+                itemId = item.Id;
```
MrTimscampi, line 620: Is there any reason for initializing it first to null then setting here instead of initializing it to item.Id directly?

src/components/images/imageLoader.js
```diff
@@ -11,22 +12,75 @@ import 'css!./style';
         fillImageElement(elem, source);
     }
 
+    async function itemBlurhashing(target, blurhashstr) {
+        if (blurhash.isBlurhashValid(blurhashstr)) {
+            // Although the default values provided by blurhash devs is 32x32, 18x18 seems to be the sweetest spot possible,
+            // cramping up a lot the performance and reducing the memory usage, while retaining almost full blur quality.
```
MrTimscampi, line 18:
```suggestion
            // Although the default values recommended by Blurhash is 32x32, a size of 18x18 seems to be the sweet spot for us,
            // improving the performance and reducing the memory usage, while retaining almost full blur quality.
```

src/components/images/imageLoader.js
```diff
@@ -11,22 +12,75 @@ import 'css!./style';
...
+    function switchCanvas(elem) {
+        let child = elem.getElementsByClassName('blurhash-canvas')[0];
+        if (child) {
+            child.style.opacity = elem.getAttribute('data-src') ? 1 : 0;
+        }
+    }
+
     export function fillImage(entry) {
         if (!entry) {
             throw new Error('entry cannot be null');
```
MrTimscampi, line 62:
```suggestion
            throw new TypeError('entry cannot be null');
```

While you're in there :)

src/strings/en-us.json
MrTimscampi: It'd be good to rename the key for translators as well :)

src/components/images/imageLoader.js
```diff
@@ -11,22 +12,75 @@ import 'css!./style';
...
+            var blurhashstr = target.getAttribute('data-blurhash');
         } else {
             source = entry;
         }
 
+        if (!target.classList.contains('blurhashed', 'non-blurhashable') && userSettings.enableBlurhash() && blurhashstr) {
+            itemBlurhashing(target, blurhashstr);
+        } else if (userSettings.enableBlurhash() && !blurhashstr && !target.classList.contains('blurhashed')) {
+            target.classList.add('non-blurhashable');
+        }
```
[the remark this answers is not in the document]
ferferga: @dmitrylyzo Now I don't remember why I put that there, but let me do my testing checklist and let's see if everything it's fine without it :P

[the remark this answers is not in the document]
ferferga: @dmitrylyzo After testing, now I remember why I did that! Items that don't have a blurhash string will get the non-blurhashable class when the blurhash option in usersettings is disabled.

Behavioral-wise, removing this doesn't make any difference. I just added it because, in my mind, it didn't make any sense to apply a class that was supposed to specify that we can't use the blurhash feature when the blurhash feature is disabled :).

Now I don't know what to do with this! :P

PR #1287, opened 2020-05-23
Support for faster playback rates.

jsharkey:
The HTML5 video element already has a well-supported "playbackRate" attribute
which can be used to increase playback rate.  This change wires up that control
to be displayed in the Jellyfish web player.

The playback rates offered are between 1x and 2x in 0.25x increments, which
matches the YouTube player.  This change also wires up the ">" and "<" key
events to increase and decrease the playback rate, which mirrors the keyboard
shortcuts supported by YouTube.

### DIFF
--- a/src/components/htmlVideoPlayer/plugin.js
+++ b/src/components/htmlVideoPlayer/plugin.js
@@ -996,6 +996,7 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
             customTrackIndex = -1;
             currentClock = null;
             self._currentAspectRatio = null;
+            self._currentPlaybackRate = null;
 
             var octopus = currentSubtitlesOctopus;
             if (octopus) {
@@ -1438,6 +1439,7 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
 
         list.push('SetBrightness');
         list.push('SetAspectRatio');
+        list.push('SetPlaybackRate');
 
         return list;
     }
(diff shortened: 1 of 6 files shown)


src/components/htmlVideoPlayer/plugin.js
```diff
@@ -1722,6 +1724,37 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
+        }, {
+            name: '1.5x',
+            id: 1.5
+        }, {
+            name: '1.75x',
+            id: 1.75
+        }, {
+            name: '2x',
+            id: 2.0
+        }];
```
Artiume, line 1755: Could we also do half speed? it could be useful

jsharkey: Sure, added 0.5x and 0.75x, since YouTube supports those too, which probably means they're useful.

Artiume:
Can you try basing this against the syncplay pr's to see how it plays? It's intended for 10.6 and I imagine could might complicate it lol

dkanada:
I could have sworn the SyncPlay one also implemented video speed.

Artiume:
Not for this PR, but I wonder if we should include a key binding guide at some point.  Like if the Playback Stats page is pulled up, nearby we can have a small guide for all key bindings.

dkanada:
@jsharkey it does seem like some level of playback rate control was added for SyncPlay in [this](https://github.com/jellyfin/jellyfin-web/pull/1011/files#diff-8cc2cabd6a4cde48b0ac68b500572723R3783) commit. Does this build on the changes there or were they independent implementations? It looks like yours adds a menu which is great, but the underlying core might be overlapping a bit, which would explain the merge conflicts. I'd love to get the additions in 10.6.0 if we can though.

dkanada:
This was rebased and submitted as a new pull request with your commits preserved, so I'll close this one. It should be included in the next feature release.

adyanth:
Was the playback speed support also introduced on the android app as well? I don't seem to see it present there.

Artiume:
It should be present on Android master. Android hasn't received an update yet because it has breaking changes with 10.6 so we're waiting to release the update once enough people migrate to 10.7

PR #1288, opened 2020-05-24
Add display preference for details banner and update some defaults

dkanada:


src/scripts/settings/userSettings.js
```diff
@@ -107,7 +107,7 @@ import events from 'events';
         }
 
         val = this.get('enableThemeSongs', false);
-        return val !== 'false';
+        return val === 'true';
```
thornbill, line 110: These statements are not equivalent. Consider the case when`val` Is undefined or null.

dkanada: It was intentional, I was changing the default settings to something I consider more sane. I don't think we should enable the backdrops by default, but that one's up for discussion. Theme songs and videos are also rarely used so it doesn't make sense to enable them by default.

thornbill: Ah I see! I think backdrops should b enabled my default, but agree on the others.

src/strings/en-us.json
```diff
@@ -822,8 +824,8 @@
     "LabelSaveLocalMetadataHelp": "Saving artwork into media folders will put them in a place where they can be easily edited.",
     "LabelScheduledTaskLastRan": "Last ran {0}, taking {1}.",
     "LabelScreensaver": "Screensaver:",
-    "EnableFastImageFadeIn": "Fast image fade-in",
-    "EnableFastImageFadeInHelp": "Enable faster fade-in animation for loaded images",
+    "EnableFastImageFadeIn": "Image Fade Animations",
+    "EnableFastImageFadeInHelp": "Show posters and other images with a fade animation once they load.",
```
thornbill, line 828: My understanding is this doesn’t disable the animation entirely, just makes it faster. Perhaps that should be clarified here?

dkanada:
Nope you're right, I've been trying to eradicate `enable` from all the checkbox preferences.

PR #1293, opened 2020-05-25
Support restarting the server on all platforms

mark-monteiro:
**Changes**
Remove logic to check if server restart is supported and always display the restart button in the dashboard.

This change goes along with the server PR https://github.com/jellyfin/jellyfin/pull/3171, that changes the restart functionality to be cross-platform and removes `CanSelfRestart` property


mark-monteiro:
FYI we might want to revert this. I don't think the server PR is going to get merged as-is, because it will break plugin updates

PR #1294, opened 2020-05-25
Remove legacy jquery function from source

dkanada:
I tested the checkbox behavior and couldn't find a difference. We'll still need to remove this from all plugins before it can be deleted entirely.


src/controllers/dashboard/dlna/settings.js
```diff
@@ -22,8 +22,8 @@ define(['jQuery', 'loading', 'libraryMenu', 'globalize', 'fnchecked'], function
             config.EnablePlayTo = form.querySelector('#chkEnablePlayTo').checked;
             config.EnableDebugLog = form.querySelector('#chkEnableDlnaDebugLogging').checked;
             config.ClientDiscoveryIntervalSeconds = $('#txtClientDiscoveryInterval', form).val();
-            config.EnableServer = $('#chkEnableServer', form).checked();
-            config.BlastAliveMessages = $('#chkBlastAliveMessages', form).checked();
+            config.EnableServer = $('#chkEnableServer', form).checked
```
dkanada, line 25:
```suggestion
            config.EnableServer = $('#chkEnableServer', form).checked;
```

PR #1295, opened 2020-05-25
Search by get request

apierre55000:
hello it's possible to make a research by get request via the url ? because it would be great to use jellyfin as a search engine into the browser like this : https://localhost:8096/search?q=%s


dkanada:
We do allow searches on the `Items` endpoint, is that sufficient?

apierre55000:
> We do allow searches on the `Items` endpoint, is that sufficient?

what is exactly the request because i think it require a token to do this no ? but it would be cool to have a research without token or a token with long life to make research because there is the possibility to add custom search engines into all web browsers

dkanada:
Oh, I see the reasoning now. That would require server changes and woud also definitely require a setting to enable it so the unauthenticated endpoint wouldn't be open by default.

apierre55000:
> Oh, I see the reasoning now. That would require server changes and woud also definitely require a setting to enable it so the unauthenticated endpoint wouldn't be open by default.

that's it ! it would be cool i supposed :)

dkanada:
I'll close this one for now since it would require server changes first, but feel free to open a request [here](https://features.jellyfin.org) so we can track progress!

PR #1315, opened 2020-05-27
Fix ability to change settings of another user

dmitrylyzo:
Ability to change settings of another user was broken again.

**Changes**
* Migrate to class.
* Add `currentSettings`.
* Add backward compatibility for non-ES6 modules.

**Issues**
`userSettings` module cannot be "constructed".


dmitrylyzo:
> Why don't we just make this constructable or add a singleton or something instead of adding a bunch of workarounds? I'm fine with this, just wondering if it's an ideal solution.

Constructable is for settings of another user.
Singleton is for session-wide settings of current user.
It was constructable and singleton in separate modules, but it seems you don't like how it looks.

All that wrappers are for is not to create conflicts with other modules. You can use `userSettings.currentSettings` directly - I suppose we will use this in the future.
I didn't try to make `currentSettings` `default`, so I don't know if it will make `UserSettings` inaccessible.

dmitrylyzo:
> I didn't try to make `currentSettings` `default`, so I don't know if it will make `UserSettings` inaccessible.

Just checked. You won't loose access to `UserSettings`, but you will need to explicitly specify `default` or `currentSettings` on `require`.
```js
{
    currentSettings: UserSettings,
    default: UserSettings,
    _callbacks: {…},
    __esModule: true,
    UserSettings: ƒ
}
```

dkanada:
Ah I see why you made the split now, it seems fine to me.

dmitrylyzo:
I have played with `default` a bit more.
We could add `export default currentSettings;` to simplify import in ES6 modules.

without `export default currentSettings;`
```js
import {currentSettings as userSettings} from 'userSettings';
import {UserSettings, currentSettings as userSettings} from 'userSettings';
```
with `export default currentSettings;`
```js
import userSettings from 'userSettings'; // default
import {currentSettings as userSettings} from 'userSettings';
import {UserSettings, currentSettings as userSettings} from 'userSettings';
import {UserSettings, default as userSettings} from 'userSettings'; // less obvious
```

dkanada:
What about a generic UserSettings file and then another file that handles the settings for the current user in the API client? We could also keep said file in the web client source. Not sure this is cleaner though, @MrTimscampi what do you think?

dkanada:
I'm fine with the existing code as well to be clear, but I'm sure we might run into this problem again, so might as well discuss it a bit more.

MrTimscampi:
@dmitrylyzo I seem to always be getting the default value or null, both when using this and when using master.

dmitrylyzo:
> I seem to always be getting the default value or null, both when using this and when using master.

Just checked - settings work as expected for both current user and another user.
Go to settings, check/uncheck option, back, go to settings again, checkbox is correctly restored;
Refresh, go to settings, checkbox is correctly restored.
I usually test on image loader fade animation.

_Browser cache?_

dkanada:
It will always be browser cache until we add hashes to the nightly assets.

PR #1316, opened 2020-05-27
Remove download images in advance from library options

ferferga:
Client change of jellyfin/jellyfin#3196


Artiume:
The chat moved a little fast when this came up. Is this just not needed anymore or was already on or what?

ferferga:
@Artiume Afaik, this check made that images for items were downloaded when the scanning was going on. If you leave that checkbox disabled, Jellyfin would query the TMDB (or whatever provider) for that image (a sort of reverse proxying) when a client queried for that image. I think it adds a failure chance (what if the image is removed from the provider and it gets queried by one client months later?

It also was one of the complications commented by @GranPC in the blurhash PR: he had troubles, because the encode blurhash function sometimes received URLs instead of local paths (the ones that were already cached/downloaded because the option was enabled).

I think this function just adds unnecessary complications and hassle (because you need to make the server code work with URLs as well as local files, which should be something handled only by the WebSockets and HTTP pieces of the code, not the image encoders. This web PR removes that checkbox, while the server one makes it behave as if it was enabled.

dkanada:
I don't think hiding this in the interface is a complete solution, we should either remove it from the backend at the same time or leave it alone until we decide to remove it.

I actually think we *shouldn't* remove it, and we should also remove the ability to download them only when requested as well, because it causes problems for other services like blurhash. If we also separated the metadata refresh from the library scans, I think that would be a good compromise, allowing users to scan without downloading anything from remote sources.

ferferga:
@dkanada Backend change is at jellyfin/jellyfin#3196

cvium (approved):

PR #1321, opened 2020-05-28
Lint and build improvements

MrTimscampi:
**Changes**

* Adds support for class properties (Babel + ESLint)
* Adds support for private class properties (Babel + ESLint)
* Removes unused dependencies
* Set the proper stats level for Webpack (Prevent the diarhea of info in the console when building the client)

**Issues**


package.json
```diff
@@ -6,13 +6,15 @@
...
+    "@babel/plugin-proposal-class-properties": "^7.10.1",
+    "@babel/plugin-proposal-private-methods": "^7.10.1",
     "@babel/plugin-transform-modules-amd": "^7.9.6",
     "@babel/polyfill": "^7.8.7",
     "@babel/preset-env": "^7.8.6",
     "autoprefixer": "^9.8.0",
+    "babel-eslint": "^11.0.0-beta.2",
     "babel-loader": "^8.0.6",
     "browser-sync": "^2.26.7",
-    "clean-webpack-plugin": "^3.0.0",
```
cromefire, line 15: Is there a reason for that? I'd consider it good to cleanup the results before the build (so there are no remaining files that don't belong there)

webpack.common.js
```diff
@@ -16,6 +16,7 @@ const Assets = [
 module.exports = {
     context: path.resolve(__dirname, 'src'),
     entry: './bundle.js',
+    stats: 'errors-only',
```
cromefire, line 19: What's up with this wouldn't it be better to set that by cli if needed

package.json
```diff
@@ -47,9 +49,6 @@
     "stylelint-no-browser-hacks": "^1.2.1",
     "stylelint-order": "^4.0.0",
     "webpack": "^4.41.5",
-    "webpack-cli": "^3.3.10",
```
cromefire, line 50: This should be needed right?

cromefire: Also it seems to be an old version

package.json
MrTimscampi: That's handled by Gulp.

package.json
cromefire: https://webpack.js.org/guides/getting-started/#basic-setup

webpack.common.js
MrTimscampi: It's meant to hide all the non essential output of webpack (stats about bundle size and so on) during builds

package.json
MrTimscampi: It's not used, webpack is called through Gulp.

PR #1342, opened 2020-06-01
Fix userSettings.chromecastVersion call

dmitrylyzo:
**Changes**
Make a call.

**Issues**
Logic is broken.


dmitrylyzo:
It seems that chromecast player is trying to access too early - settings are not configured for user yet.

hamburglar2160:
This works.

PR #1347, opened 2020-06-03
Fixing Chrome's Unnecessary AV1 Transcode and Combined WebM Profiles

Brissot:
Problems:
1) Chrome fails to DirectPlay AV1 Video
2) WebM profiles are coded redundantly

Solutions:
1) The canPlayType() function is currently inaccurate for AV1 on Chrome. Added temporary code to sidestep the issue. Also fixed AV1 with mkv.
Special thanks to @cconcolato and his [website](https://cconcolato.github.io/media-mime-support/) for stopping the pain and frustration of finding the issue.
2) Made a new array, webmVideoCodecs[] to handle webm video and combined the VP8, VP9 and AV1 device profiles together into one webm device profile.

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -176,6 +176,16 @@ define(['browser'], function (browser) {
         return false;
     }
 
+    function testCanPlayAv1(videoTestElement) {
+        if (browser.tizenVersion >= 5.5) {
+            return true;
+        } else if (browser.web0sVersion >= 5 && window.outerHeight >= 2160) {
+            return true;
+        }
+
+        return videoTestElement.canPlayType('video/webm; codecs="av01.0.15M.10"').replace(/no/, '');
+    }
+
     function testCanPlayTs() {
         return browser.tizen || browser.orsay || browser.web0s || browser.edgeUwp;
     }
@@ -317,7 +327,6 @@ define(['browser'], function (browser) {
 
         var canPlayVp8 = videoTestElement.canPlayType('video/webm; codecs="vp8"').replace(/no/, '');
         var canPlayVp9 = videoTestElement.canPlayType('video/webm; codecs="vp9"').replace(/no/, '');
-        var canPlayAv1 = videoTestElement.canPlayType('video/webm; codecs="av1"').replace(/no/, '');
         var webmAudioCodecs = ['vorbis'];
 
         var canPlayMkv = testCanPlayMkv(videoTestElement);
@@ -459,6 +468,7 @@ define(['browser'], function (browser) {
         });
 
         var mp4VideoCodecs = [];
+        var webmVideoCodecs = [];
         var hlsVideoCodecs = [];
 
         if (canPlayH264(videoTestElement)) {
@@ -490,14 +500,30 @@ define(['browser'], function (browser) {
 
         if (canPlayVp8) {
             mp4VideoCodecs.push('vp8');
+            webmVideoCodecs.push('vp8');
         }
 
         if (canPlayVp9) {
             mp4VideoCodecs.push('vp9');
+            webmVideoCodecs.push('vp9');
+        }
+
+        if (testCanPlayAv1(videoTestElement)) {
+            mp4VideoCodecs.push('av1');
+            webmVideoCodecs.push('av1');
         }
 
         if (canPlayVp8 || browser.tizen || browser.orsay) {
             videoAudioCodecs.push('vorbis');
+        }
+
+        if (webmVideoCodecs.length) {
+            profile.DirectPlayProfiles.push({
+                Container: 'webm',
+                Type: 'Video',
+                VideoCodec: webmVideoCodecs.join(','),
+                AudioCodec: webmAudioCodecs.join(',')
+            });
         }
 
         if (mp4VideoCodecs.length) {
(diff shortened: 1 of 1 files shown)


src/scripts/browserDeviceProfile.js
```diff
@@ -176,6 +176,13 @@ define(['browser'], function (browser) {
         return false;
     }
 
+    //temporary, until canPlayType(av1) is fixed upstream for chrome
+    function testCanPlayAv1() {
+        if (browser.chrome) {
+            return true;
+        }
```
MrTimscampi, line 183: @dmitrylyzo This will break WebOS and Tizen if I'm not mistaken, right?

Brissot: Yeah, you're right. I'll change it to have exceptions.

dmitrylyzo: If user agent has `chrome` (Tizen 3+ and WebOS 3+ have it), `browser.chrome` is `true`. So probably yes.
How about check for `browser.versionMajor`?
I have no idea if they support AV1, so maybe we should completely exclude them.


In addition, this function has no return at the end and can be replaced with just `return ...`.

MrTimscampi: > I have no idea if they support AV1, so maybe we should completely exclude them.

They don't, it's a fairly new format. WebOS 5 or the latest Tizen might but we'd need to check official docs.

dmitrylyzo: https://developer.samsung.com/smarttv/develop/specifications/media-specifications/2020-tv-video-specifications.html
https://ottball.com/tizen-tv-5-5-whats-new/
It looks like Tizen 5.5 (Chrome 69) supports AV1.

Maybe we should use Chrome version (`browser.chrome && browser.versionMajor >= 69`).
But according to https://caniuse.com/#feat=av1, Chrome 69 supports AV1 if it is enabled in config.

As for WebOS, this is not confirmed.
http://webostv.developer.lge.com/discover/specifications/supported-media-formats/
https://ottball.com/webos-tv-5-0/

MrTimscampi:
I am curious which version of Chrome you are using, because Chrome 83 reports that it properly supports every AV1 entry on the codec test page you linked.

Brissot:
> I am curious which version of Chrome you are using, because Chrome 83 reports that it properly supports every AV1 entry on the codec test page you linked.

I'm on Chrome 83 as well, Windows. Did you try the "Test your own codec" part? As far as I can see the table just references a premade table, while "Test your own codec" executes canPlayType() in your browser.

MrTimscampi:
The table is not pre-made, it's filled using a bunch of JS when the page loads (The source code is here: https://github.com/cconcolato/media-mime-support)

See here the first table on Firefox:
image

And on Chrome:
image

I suspect the detection for AV1 merged in #1167.

From testing in the web console, it seems that Firefox returns positive results for `video/webm; codecs="av1"` while Chrome doesn't. Per MDN, the proper canPlayType format is actually simiar to `av01.P.LLT.DD[.M[.CCC[.cp[.tc[.mc[.F]]]]]]`.

Hence, I think we should fix this by testing the profiles (Like the test page does) instead of returning a blanket `true`.

src/scripts/browserDeviceProfile.js
Brissot: I've added the exceptions for Orsay, Tizen and WebOS:
I can't find any information on Orsay, so I'll assume AV1 isn't supported. 
WebOS [support for AV1 was added in version 5](http://webostv.developer.lge.com/discover/specifications/supported-media-formats/webos-50/), but only for 4k and 8k models, so to avoid breaking 1080p TVs window detection has been added.
Tizen supports AV1 as of 5.5

MrTimscampi:
@Brissot I've been rewriting the browserDeviceProfile.js file.

From what I've gathered by reading specs and testing, the proper test for AV1 support would be something like this:

```javascript
videoTestElement.canPlayType('video/webm; codecs="av01.0.15M.10"').replace(/no/, '');
```

This works on both Chrome and Firefox. I'd be much more inclined on merging that as a fix for Chrome and Firefox than simply returning true :)

I think we can keep the specific detection for Tizen and WebOS (Especially since for WebOS it's resolution-dependent). We can drop the Orsay part since I'm taking that out completely in the rewrite (We don't have an Orsay client and the platform is old and unsupported anyway).

src/scripts/browserDeviceProfile.js
```diff
@@ -176,6 +176,21 @@ define(['browser'], function (browser) {
...
+        } else if (browser.tizenVersion >= 5.5) {
+            return true;
+        } else if (browser.web0sVersion >= 5 && window.outerHeight >= 2160) {
+            return true;
+        } else if (browser.chrome && browser.versionMajor >= 69) {
+            return true;
+        }
+
+        return false;
+    }
```
MrTimscampi, line 192:
```suggestion
    function testCanPlayAv1() {
        if (browser.tizenVersion >= 5.5) {
            return true;
        } else if (browser.web0sVersion >= 5 && window.outerHeight >= 2160) {
            return true;
        }
        
        return videoTestElement.canPlayType('video/webm; codecs="av01.0.15M.10"').replace(/no/, '');
    }
```

MrTimscampi: Also please remove the line with `var canPlayAv1 = videoTestElement.canPlayType('video/webm; codecs="av1"').replace(/no/, '');` since it's now unused.

src/scripts/browserDeviceProfile.js
```diff
@@ -490,16 +506,32 @@ define(['browser'], function (browser) {
...
             mp4VideoCodecs.push('vp8');
+            webmVideoCodecs.push('vp8');
         }
 
         if (canPlayVp9) {
             mp4VideoCodecs.push('vp9');
+            webmVideoCodecs.push('vp9');
+        }
+
+        if (canPlayAv1 || testCanPlayAv1) {
```
MrTimscampi, line 517:
```suggestion
        if (testCanPlayAv1(videoTestElement)) {
```

Brissot: Done and squashed in :)

PR #1350, opened 2020-06-03
Add horizontal home view for tv

grafixeyehero:


MrTimscampi:
@jellyfin/web Some sticking points for me that need to be discussed:

* It adds another component for tabbed pages (Adding to the existing ItemsTab and TabbedView)
* At adds a lot of new views very close to us enabling Vue, meaning we'll have more to port over
* It doesn't respect the user's settings for home views

Imo, it's better to focus on the TV UI once we're on Vue, since it will make everything a lot faster (We can load a different component for each route depending on the user's layout setting, for example).

What do you think?

grafixeyehero:
Vue is a new adventure for me. I am agree with you.
As you mentioned in the above for TV UI IMO until we achieved with Vue let use a polymer based that I have mentioned earlier about this repo https://github.com/grafixeyehero/Emby.Web.Theater this project was designed for tv  and electron projects??

Issue #1031, opened 2020-04-05
Lack of config.json by default produces a SyntaxError


MrTimscampi:
**Describe the bug**

Client settings are trying to load a "config.json" file which, by default, doesn't exist. This produces a SyntaxError in the console.

**To Reproduce**
1. Open web client
2. Open browser console

**Expected behavior**
No error should be produced.

**Logs**
```
SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data
```

**Screenshots**

**System (please complete the following information):**
 - OS: Arch Linux
 - Browser: All
 - Jellyfin Version: Master

**Additional context**
Renaming `config.default.json` to `config.json` fixes it, but we shouldn't ask the user to do that before using the client without producing an error.

@dkanada What do you think ? Should it fallback to using `config.default.json` if `config.json` doesn't exist ?

dkanada:
> Should it fallback to using config.default.json if config.json doesn't exist?

That seems like a good plan.

JustAMan:
I'm pretty sure that this `<` is a part of error message page. Why aren't we handling the error instead?

JustAMan:
Or, actually, we should be handling the errors in parsing the JSON and storing an empty object instead, maybe with a warning in the console.

Also I just noticed that current webConfig code doesn't handle baseURL being non-`/`, is this okay?

dkanada:
will be fixed with #1223

PR #1354, opened 2020-06-04
Fix baseUrl for web config file

### LINKED ISSUE #1254: Web config is loaded from the wrong URL when the web client is bundled with the server
**Describe the bug**
The web config file and the config template files are loaded from the wrong URL when the web client is being served by the server. For example, `/config.template.json` is used instead of `/web/config.template.json`

**To Reproduce**
1. Follow the instructions to serve the web client from the server: https://github.com/jellyfin/jellyfin#installing-the-web-client
2. Load the web client
3. See that the config file and template are requested from the wrong URL

**Expected behavior**
The config and config template should use the correct URL and load correctly.

**Logs**
From the server logs:
```log
[12:33:12] [ERR] [44] Emby.Server.Implementations.HttpServer.HttpListenerHost: Could not find handler for /config.template
[12:33:12] [ERR] [44] Emby.Server.Implementations.HttpServer.HttpListenerHost: Error processing request: Unable to find the specified file. URL: http://localhost:8096/config.template.json   
```
From the web client logs:
```log
cannot get web config: SyntaxError: "JSON.parse: unexpected character at line 1 column 1 of the JSON data" webSettings.js:25:16
```

MrTimscampi:
**Changes**

Untested due to the server building prrocess being broken.

**Issues**

Fixes #1254

### DIFF
--- a/src/scripts/settings/webSettings.js
+++ b/src/scripts/settings/webSettings.js
@@ -2,7 +2,7 @@ let data;
 
 function getConfig() {
     if (data) return Promise.resolve(data);
-    return fetch('/config.json?nocache=' + new Date().getUTCMilliseconds()).then(response => {
+    return fetch('config.json?nocache=' + new Date().getUTCMilliseconds()).then(response => {
         data = response.json();
         return data;
     }).catch(error => {
@@ -12,7 +12,7 @@ function getConfig() {
 }
 
 function getDefaultConfig() {
-    return fetch('/config.template.json').then(function (response) {
+    return fetch('config.template.json').then(function (response) {
         data = response.json();
         return data;
     });


src/scripts/settings/webSettings.js
```diff
@@ -1,8 +1,12 @@
 let data;
 
+function getBaseUrl() {
+    return location.pathname.substr(0, location.pathname.lastIndexOf('/'));
+}
+
 function getConfig() {
     if (data) return Promise.resolve(data);
-    return fetch('/config.json?nocache=' + new Date().getUTCMilliseconds()).then(response => {
+    return fetch(getBaseUrl() + '/config.json?nocache=' + new Date().getUTCMilliseconds()).then(response => {
```
dmitrylyzo, line 9: In fact, we can remove leading `/`
```js
fetch('config.json?nocache=' + new Date().getUTCMilliseconds())
```
browser will do the rest.

dkanada:
also fixes #1031

PR #1355, opened 2020-06-04
Make the default card colors look less like rainbow vomit

MrTimscampi:
**Changes**

Significantly improves the default card colors for most themes.

**Issues**

### DIFF
--- a/src/themes/blueradiance/theme.css
+++ b/src/themes/blueradiance/theme.css
@@ -131,23 +131,23 @@ html {
 }
 
 .defaultCardBackground1 {
-    background-color: #d2b019;
+    background-color: #213440;
 }
 
 .defaultCardBackground2 {
-    background-color: #338abb;
+    background-color: #8c6565;
 }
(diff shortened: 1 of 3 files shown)


Artiume:
Simple fix but can i get a before/after snapshot?

MrTimscampi:
**Dark**

Before
image

After
image

**Blue Radiance**

Before
image

After
image

**Purple Haze**
Before
image

After
image

I'm more of a fan of the Dark one, but I figured I should at least try one for the two most used non-default themes.

Artiume:
I'm not sure how I like the purple haze

MrTimscampi:
> We have 5 different colors for each theme but it seems that we only have two or three :P

There actually 5 in the new color scheme. The issue is that having them be random colors look bad, so I kept it to generally accepted color harmonies. It sort of limits the amount of variation, but we do still have 5 colors for each (though not all are visible in my screenshots)

PR #1361, opened 2020-06-05
Migration imageUploader, itemidentifier and itemMediaInfo to ES6 modules

grafixeyehero:
**Issues**
Part of #868


src/components/imageUploader/imageUploader.js
```diff
@@ -1,25 +1,41 @@
...
-    var currentFile;
-    var hasChanges = false;
-
-    function onFileReaderError(evt) {
+/* eslint-disable indent */
+
+/**
+ * Module for imageUploader.
+ * @module components/imageUploader/imageUploader
+ */
```
dkanada, line 6: These comments at the beginning of the files seem pretty useless in general since they don't actually include any extra information about the files themselves.

src/components/itemidentifier/itemidentifier.js
```diff
@@ -261,64 +280,64 @@ define(['dialogHelper', 'loading', 'connectionManager', 'require', 'globalize',
...
+        }, () => {
 
             loading.hide();
 
             dialogHelper.close(page);
         });
     }
 
-    function showIdentificationForm(page, item) {
+    function showIdentificationForm(page, {Id, Type}) {
```
dkanada, line 310: Same comment in most of these migrations about the `{}` expansion for parameters, but I'll wait to see what everyone else thinks.

src/components/imageUploader/imageUploader.js
MrTimscampi: The sentence should generally be a more useful description of what the file does.

Like:
```javascript
/*
 * Handles the metadata editing dialog and its related actions
 * @module whatever/the/real/path/is
 */
```

src/components/itemidentifier/itemidentifier.js
```diff
@@ -421,14 +440,15 @@ define(['dialogHelper', 'loading', 'connectionManager', 'require', 'globalize',
         }
     }
 
+    //TODO investigate where was used this?
```
dkanada, line 443:
```suggestion
    // TODO investigate where this was used
```

PR #1369, opened 2020-06-06
Small browser profile fixes

MrTimscampi:
**Changes**

* Removes all references to Chromecast and Orsay
* Should fix Chrome treating MKV like WebM
* Should fix MP3 audio transcoding when in video on some browsers

**Issues**

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -333,8 +333,9 @@ define(['browser'], function (browser) {
         var videoAudioCodecs = [];
         var hlsVideoAudioCodecs = [];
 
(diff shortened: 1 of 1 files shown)


src/scripts/browserDeviceProfile.js
```diff
@@ -794,20 +736,13 @@ define(['browser'], function (browser) {
...
-            //    Value: 'true',
-            //    IsRequired: false
-            //});
+        if (!browser.edgeUwp && !browser.tizen && !browser.web0s) {
+            profile.CodecProfiles[profile.CodecProfiles.length - 1].Conditions.push({
+                Condition: 'NotEquals',
+                Property: 'IsInterlaced',
+                Value: 'true',
+                IsRequired: false
+            });
```
MrTimscampi, line 745: This likely needs testing, mainly to ensure that it actually does something for interlaced video.

PR #1370, opened 2020-06-06
Migration listview to ES6 modules

grafixeyehero:
**Issues**
Part of #868

### DIFF
--- a/src/components/listview/listview.js
+++ b/src/components/listview/listview.js
@@ -1,5 +1,20 @@
-define(['itemHelper', 'mediaInfo', 'indicators', 'connectionManager', 'layoutManager', 'globalize', 'datetime', 'apphost', 'css!./listview', 'emby-ratingbutton', 'emby-playstatebutton'], function (itemHelper, mediaInfo, indicators, connectionManager, layoutManager, globalize, datetime, appHost) {
-    'use strict';
+/* eslint-disable indent */
+
+/**
+ * Module for display list view.
+ * @module components/listview/listview
+ */
+
+import itemHelper from 'itemHelper';
+import mediaInfo from 'mediaInfo';
+import indicators from 'indicators';
+import connectionManager from 'connectionManager';
+import layoutManager from 'layoutManager';
+import globalize from 'globalize';
+import datetime from 'datetime';
+import 'css!./listview';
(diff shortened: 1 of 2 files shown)


src/components/listview/listview.js
```diff
@@ -100,42 +115,42 @@ define(['itemHelper', 'mediaInfo', 'indicators', 'connectionManager', 'layoutMan
         }
     }
 
-    function getChannelImageUrl(item, width) {
+    function getChannelImageUrl({ServerId, ChannelId, ChannelPrimaryImageTag, ImageBlurHashes}, width) {
```
dkanada, line 118: I'm really not a fan of these over `item.ServerId` because the latter is much easier to search for in my opinion. @MrTimscampi what do you think?

grafixeyehero: @dkanada @MrTimscampi regarding to this do let me know to not repeat it in other file

MrTimscampi: On one hand, it makes the rest of the code cleaner (no `thing.stuff` everywhere...)

Destructuring is also commonly used, so I don't see much issue with it. It does make it harder to know what is what at a glance, which can be fixed by adding JSDoc for the function.

I'd say I'm for it, on the condition that function signatures that use it get some JSDoc to make them clearer (JSDoc reference: https://jsdoc.app/tags-param.html#parameters-with-properties)

dkanada: My issue here is that if I search for `ServerId` in a file, it could be about `user.ServerId` or `item.ServerId` but there's no way to distinguish them in searches now. This pattern also falls apart when you have to send two similar objects like `newPlugin` and `oldPlugin` to compare them, or even just `plugin` and `user` which will both have ID values. An even worse example is the ID field, which could refer to so many different things.

```
plugin.Id
genre.Id
artist.Id
server.Id
user.Id
item.Id
image.Id
```

grafixeyehero:
Oh my in this case I agree with @dkanada and i will reverse the Destructuring method

src/components/listview/listview.js
grafixeyehero: Oh my in this case I agree with @dkanada and i will reverse the Destructuring method

PR #1374, opened 2020-06-08
Adapt book plugin for mobile devices

itegulov:
**Changes**
Makes book plugin usable on mobile devices.

### DIFF
--- a/src/plugins/bookPlayer/plugin.js
+++ b/src/plugins/bookPlayer/plugin.js
@@ -19,6 +19,8 @@ export class BookPlayer {
 
         this.onDialogClosed = this.onDialogClosed.bind(this);
         this.openTableOfContents = this.openTableOfContents.bind(this);
+        this.prevChapter = this.prevChapter.bind(this);
+        this.nextChapter = this.nextChapter.bind(this);
         this.onWindowKeyUp = this.onWindowKeyUp.bind(this);
     }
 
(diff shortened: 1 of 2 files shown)


itegulov:
So far I am having some problems (at least on my android device with mobile chrome browser) with using true continous mode with epub.js. There is a bug which causes the screen to randomly flash and move your position to an arbitrary position in chapter. Table of Content navigation is also very wonky (randomly moves you to the middle of the chapter you clicked on).

Right now I am thinking of sticking with `scrolled-doc` mode and offset of 0 as was suggested [here](https://github.com/futurepress/epub.js/issues/936#issuecomment-483378964). This way we will have a continuous mode for the current chapter and some kind of button (or some other mechanism such as long swipe down) at the end of a chapter which will bring you to the next chapter.

dkanada:
Does the flickering exist in this pull request?

itegulov:
@dkanada I have not experienced flickering with the current configuration. However, there is no way to continue to the next chapter once you've finished reading the current one (aside from opening TOC and clicking on the next chapter).

dkanada:
Hmmm maybe just a next chapter button at the end of the EPUB element then? That seems good enough to me for now.

itegulov:
Yeah. I will try to find some time this week to add a button for next (and maybe previous?) chapter.

itegulov:
@dkanada sorry for the long wait, had a lot of going in the last three months. I managed to make it look and work decently but I had to remove touch controls you introduced in https://github.com/jellyfin/jellyfin-web/pull/1681 as they were interfering with continous scrolling.

The final approach I decided to go with is continous scrolling with chapter pagination, so you can read an entire chapter with scrolling and then have to press the "next chapter" button on the bottom to go to the next chapter. Could you please take a look and tell me what you think?

dkanada:
@itegulov perhaps a button to switch between the two manually instead? This looks good regardless, just thought giving a choice might be good.

dkanada (approved):

MrTimscampi (approved):

PR #1386, opened 2020-06-08
Fix episodes listing after blurhash

ferferga:
Images in episodes listing for TV Shows were repeated after adding blurhash.

Screenshot courtesy of @sparky8251 

Screenshot_20200608_155403

### DIFF
--- a/src/components/listview/listview.js
+++ b/src/components/listview/listview.js
@@ -93,7 +93,7 @@ define(['itemHelper', 'mediaInfo', 'indicators', 'connectionManager', 'layoutMan
                 itemId = item.ParentPrimaryImageItemId;
             }
         }
-        
+
         let blurHashes = item.ImageBlurHashes || {};
         let blurhashstr = (blurHashes[options.type] || {})[options.tag];
 
--- a/src/components/listview/listview.js
+++ b/src/components/listview/listview.js
@@ -80,18 +80,15 @@ define(['itemHelper', 'mediaInfo', 'indicators', 'connectionManager', 'layoutMan
         if (item.ImageTags && item.ImageTags.Primary) {
             options.tag = item.ImageTags.Primary;
             itemId = item.Id;
-        }
-        if (!itemId) {
-            if (item.AlbumId && item.AlbumPrimaryImageTag) {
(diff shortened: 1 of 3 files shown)


src/controllers/itemDetails.js
```diff
@@ -1840,7 +1840,8 @@ define(['loading', 'appRouter', 'layoutManager', 'connectionManager', 'userSetti
                 chaptercardbuilder.buildChapterCards(item, chapters, {
                     itemsContainer: scenesContent,
                     backdropShape: 'overflowBackdrop',
-                    squareShape: 'overflowSquare'
+                    squareShape: 'overflowSquare',
+                    imageBlurhashes: item.ImageBlurHashes
```
ferferga, line 1844: I added this here because people cards (which depends on cardBuilder) had this added in blurhash. Not sure when this is used though

cc @JustAMan

src/components/listview/listview.js
```diff
@@ -81,23 +81,26 @@ define(['itemHelper', 'mediaInfo', 'indicators', 'connectionManager', 'layoutMan
...
-            options.tag = item.AlbumPrimaryImageTag;
-            itemId = item.AlbumId;
-        } else if (item.SeriesId && item.SeriesPrimaryImageTag) {
-            options.tag = item.SeriesPrimaryImageTag;
-            itemId = item.SeriesId;
-        } else if (item.ParentPrimaryImageTag) {
-            options.tag = item.ParentPrimaryImageTag;
-            itemId = item.ParentPrimaryImageItemId;
+        if (!itemId) {
+            if (item.AlbumId && item.AlbumPrimaryImageTag) {
```
dmitrylyzo, line 85: How about just `else if (item.AlbumId && item.AlbumPrimaryImageTag) {`?

ferferga: @dmitrylyzo I don't understand how I didn't thought about this before 😅. Addressed.

ferferga:
@dmitrylyzo I was also aware of that but I thought the only way to fix it was to move them to the card element! That worked beautifully, thanks!

I'm going to push it in this PR as it's an small change.

PR #1389, opened 2020-06-09
Expand interface effects

ferferga:
This is a quick PR made after a Matrix discussion. This 
adds more effects when hovering images (+ proper customization), based on the blur effects we're adding with blurhash.

I also took the freedom to reorganize the display settings page, **moving all content related settings to a new "Library" section**:
image

## Effect showcase

**Darkening + blur + zoomIn** *(Default one)*
all

**Darkening + blur**
blur+dark

**Darkening + ZoomIn**
dark+zoom

**ZoomIn + Blur**
zoom+blur

**Blur**
blur

**ZoomIn**
zoom

**Darkening**
*This is what we have right now*

**Nothing**
nothing

## To do
**ZoomIn effect needs to be fixed for Purple Haze and themes that have rounded corners**

**This only affects card elements, other images are unaffected**

## Details

TV layout zoom affect is handled by mouseManager so, in TV layout, the zoom effect doesn't change anything.

### DIFF
--- a/src/components/cardbuilder/card.css
+++ b/src/components/cardbuilder/card.css
@@ -53,6 +53,7 @@ button::-moz-focus-inner {
 .cardScalable {
     position: relative;
     contain: layout style;
+    overflow: hidden;
 }
 
 .cardPadder-backdrop,
@@ -78,6 +79,7 @@ button::-moz-focus-inner {
 .overflowPortraitCard-textCardPadder {
     padding-bottom: 150%;
     contain: strict;
+    overflow: hidden;
 }
 
 .cardPadder-banner {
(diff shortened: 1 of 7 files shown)


hamburglar2160:
image

MrTimscampi:
> image

I don't think all of these should go into a "Libraries" section.

Page size obviously fits.

Faster animations is (should be) app-wide, not only in libraries.

Same for blurhash.

Zoom, blur and darken are for all cards, not only in libraries (home screen and details page as well).

Details banner is for the details page, not librairies.

Backdrops is app-wide.

Theme songs and videos are for the details page as well.

Imo we should give it more thought and move the redesign to another PR.

sparky8251:
Is there any way to feasibly make use of transparency and blur effects to get the blurhash image semi merged with the actual image when focused?

The blurring i see here doesnt feel like the kind id expect myself...

adir6:
Hi, I'm sorry as I'm not the coder but I've one nice idea about it.

Hover should be bright + zoomin
Non-Hover should be dark + zoomout/no-zoom

Thanks.

dkanada:
@ferferga should this be closed in favor of the vue client?

PR #1401, opened 2020-06-12
Remove standalone mode

### LINKED ISSUE #1433: 10.6.0 logout/login i have to select again my server
**Describe the bug**
i think end users have only one jellyfinserver, when i logout and login, i have to select my server, then connect, then login

**Expected behavior**
when you logout, the jellyserver should be automaticly selected, to show only user selection to login
and add a bouton, change server, if end users have two jellyfinserver or more

 - OS: docker
 - Browser: chrome
 - Jellyfin Version: 10.6.0

MrTimscampi:
**Changes**

**Untested** Injects the Javascript in all cases but keeps standalone mode only when running a standalone build.

The goal is to inject JS so the app is self-starting while still keeping the old behavior elsewhere in the client when bundled with the server.

**Issues**

Fixes #1433

### DIFF
--- a/.ci/azure-pipelines-build.yml
+++ b/.ci/azure-pipelines-build.yml
@@ -8,6 +8,8 @@ jobs:
         BuildConfiguration: development
       Production:
         BuildConfiguration: production
+      Standalone:
+        BuildConfiguration: standalone
 
   pool:
     vmImage: 'ubuntu-latest'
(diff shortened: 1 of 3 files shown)


crobibero:
If I had a bad server in my local storage the sever select page doesn't load.

I also still get the following when trying to start fresh.
```
ReferenceError: ApiClient is not defined start.js:37:28
    <anonymous> start.js:37
    onViewChange viewManager.js:62
    loadView viewManager.js:151
```

Thalhammer:
> If I had a bad server in my local storage the sever select page doesn't load.
> 
> I also still get the following when trying to start fresh.
> 
> ```
> ReferenceError: ApiClient is not defined start.js:37:28
>     <anonymous> start.js:37
>     onViewChange viewManager.js:62
>     loadView viewManager.js:151
> ```
This seems to be an issue in master not because of this commit. I just started and stumbled over the same problem. It's because of this line:
https://github.com/jellyfin/jellyfin-web/blob/1b6a600496bab236583b5f81a413bb0bf0dda1b7/src/scripts/site.js#L639
A browser is incorrectly detected as a nativeApp which causes ApiClient initialization to be skipped.

PR #1404, opened 2020-06-13
Migration collectionEditor and playlisteditor to ES6 modules

grafixeyehero:
**Issues**
Part of #868


dkanada:
Looks like rebasing causes issues since you originally branched off master instead.

grafixeyehero:
> 
> 
> Looks like rebasing causes issues since you originally branched off master instead.

oh i will reset to https://github.com/jellyfin/jellyfin-web/pull/1404/commits/cf931d429a1913e6272d7347a41817dc199e68ae commit and i will update shortcuts.js in another PR

PR #1406, opened 2020-06-14
Avoid async rendering of MediaSources tracks to prevent content jump

Delgan:
**Changes**

The `MediaSources` returned by `.getPlaybackMediaSources()` seems already available in the `item`, can we use it directly an avoid sending a request to the server?

Certainly `getPlaybackMediaSources()` is called for a good reason, but I tried to go down the rabbit hole on the server side and, well... I don't understand anything. That's quite frustrating.

`MediaSources` are retrieved here: [jellyfin/MediaInfoService.cs#L30](https://github.com/jellyfin/jellyfin/blob/403cd3205ffb970cfda88b6c49dc69127fada798/MediaBrowser.Api/Playback/MediaInfoService.cs#L30). There are many similarities between `item` and `playbackInfoResult`, I don't know why both are needed.

I'm opening this PR in the hope that there's someone who understands the topic. Certainly, it is possible to avoid making two calls to the api in one way or another, and the rendering would be greatly improved.

Before:
Peek 2020-06-14 21-30

After:
Peek 2020-06-14 21-28

**Issues**
Related to #1233

### DIFF
--- a/src/controllers/itemDetails.js
+++ b/src/controllers/itemDetails.js
@@ -141,29 +141,29 @@ define(['loading', 'appRouter', 'layoutManager', 'connectionManager', 'userSetti
             return;
         }
 
-        playbackManager.getPlaybackMediaSources(item).then(function (mediaSources) {
-            instance._currentPlaybackMediaSources = mediaSources;
-            page.querySelector('.trackSelections').classList.remove('hide');
-            select.setLabel(globalize.translate('LabelVersion'));
-            var currentValue = select.value;
-            var selectedId = mediaSources[0].Id;
-            select.innerHTML = mediaSources.map(function (v) {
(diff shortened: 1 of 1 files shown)


dkanada:
The answer is probably "legacy reasons" for most of this stuff. There are like 60,000 lines of code in this repository or something and we haven't exactly looked through all of them yet and fixed the issues! haha

dkanada:
If everything still loads properly after your testing and playback also works in every scenario you can check, this is probably fine. I'll run some tests as well, but this seems like a good speed increase to fit in the next update!

Delgan:
Thanks! I'm a little worried about possible side effects, but we will correct them in due course if there are any. :grin:

PR #1408, opened 2020-06-14
Fix subtitle line spacing and add position

### LINKED ISSUE #1380: Subtitle position
Hello,
is it possible to change the subtitles position? They are a bit higher compared to Emby or Plex and it would be great if I can move them down the screen.

If I am understanding this correctly it is harder to implement the option in the interface but isn't there anything else I could do maybe before creating the Jellyfin.wgt file?

When the subtitles are with two lines the spacing between them is big, so maybe just lower the upper subtitles with one row will be perfect, thanks!

Thank you for your hard work.

### LINKED ISSUE #1407: Subtitle row spacing
Hello,
I am having issue with the spacing between two subtitle rows.
The spacing is very big 2-3cm. Is there a way to decrease that spacing and from where I must edit? 
Also for the subtitles position. They should be on bottom but are at least 3-4cm high, I've tried several settings with no luck, maybe I am not editing the right file. 
Thank you for your hard work!

dmitrylyzo:
:warning: **I have no idea how this will work on Chromecast.**

`line-height` of child cannot be smaller than parent's value. WebKit text container has inline style with bigger font, and child text lines inherit its line height and cannot reduce it.

**Changes**
* Override `font-size` of container to make it similar to Firefox.
* Set small `line-height` for container so that its children can set smaller values.
* Add subtitle position

**Issues**
Fixes #1407 (only visible on `Small` and `Smaller`)
Fixes #1380

**Screenshots**
<details>
<summary>Smaller</summary>

Before (Chrome)
192 168 2 18_jellyfin_web_index html(FHD)
After (Chrome)
192 168 2 18_jellyfin_web_index html(FHD)
Firefox (unchanged)
Скриншот сделанный 2020-06-15 в 01 48 44
</details>

<details>
<summary>Normal</summary>

Before (Chrome)
192 168 2 18_jellyfin_web_index html(FHD) (1)
After (Chrome)
192 168 2 18_jellyfin_web_index html(FHD) (1)
Firefox (unchanged)
Скриншот сделанный 2020-06-15 в 01 56 28
</details>

<details>
<summary>Larger</summary>

Before (Chrome)
192 168 2 18_jellyfin_web_index html(FHD) (2)
After (Chrome)
192 168 2 18_jellyfin_web_index html(FHD) (2)
Firefox (unchanged)
Скриншот сделанный 2020-06-15 в 01 56 49
</details>

<details>
<summary>Extra Large</summary>

Before (Chrome)
192 168 2 18_jellyfin_web_index html(FHD)-old
After (Chrome)
192 168 2 18_jellyfin_web_index html(FHD)-new
Firefox (unchanged)
Скриншот сделанный 2020-06-15 в 09 29 16
</details>

<details>
<summary>Subtitle Position</summary>

pos
</details>


dkanada:
Chromecast is a completely separate client now, do you mean you have no idea how it will be implemented there?

dmitrylyzo:
> Chromecast is a completely separate client now, do you mean you have no idea how it will be implemented there?

This message
https://github.com/jellyfin/jellyfin-web/blob/5ea152f8c651328a9f175d60614db26dd4835c31/src/strings/en-us.json#L1408
In other words, I haven't tested Chromecast.

src/components/subtitlesettings/subtitlesettings.template.html
```diff
@@ -38,6 +38,45 @@ <h2 class="sectionTitle">
             ${HeaderSubtitleAppearance}
         </h2>
 
+        <style>
```
MrTimscampi, line 41: Could these use a `subtitlesettings.css` file instead?

iNfRaSoNiC13:
This one is a perfect addition, please approve it.

PR #1422, opened 2020-06-16
Migration backdrop to ES6 modules

grafixeyehero:
**Issues**
Part of #868

### DIFF
--- a/src/scripts/autoBackdrops.js
+++ b/src/scripts/autoBackdrops.js
@@ -50,7 +50,7 @@ define(['backdrop', 'userSettings', 'libraryMenu'], function (backdrop, userSett
                         return i;
                     }));
                 } else {
-                    backdrop.clear();
+                    backdrop.clearBackdrop();
                 }
             });
         }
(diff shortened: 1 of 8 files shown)


src/scripts/autoBackdrops.js
```diff
@@ -50,7 +50,7 @@ define(['backdrop', 'userSettings', 'libraryMenu'], function (backdrop, userSett
                         return i;
                     }));
                 } else {
-                    backdrop.clear();
+                    backdrop.clearBackdrop();
```
dkanada, line 53: I think `clear` was better than `clearBackdrop` since the filename already indicates the object.

MrTimscampi: Agreed, `backdrop.clear()` is nicer.

MrTimscampi: I think we can refactor that down the line, looks like there are  some other changes needed to have a logical and consistent naming scheme here.

PR #1424, opened 2020-06-16
Migration dialog and dialogHelper to ES6 modules

grafixeyehero:
**Issues**
Part of #868


src/components/dialogHelper/dialogHelper.js
```diff
@@ -25,28 +33,28 @@ define(['appRouter', 'focusManager', 'browser', 'layoutManager', 'inputManager',
...
-        var parentNode = elem.parentNode;
+        const parentNode = elem.parentNode;
         if (parentNode) {
 
             // Seeing crashes in edge webview
             try {
                 parentNode.removeChild(elem);
             } catch (err) {
-                console.error('error removing dialog element: ' + err);
+                console.error(`error removing dialog element: ${err}`);
```
dkanada, line 43: same thing with the error here, lets avoid the string literal

src/components/dialogHelper/dialogHelper.js
```diff
@@ -488,12 +496,13 @@ define(['appRouter', 'focusManager', 'browser', 'layoutManager', 'inputManager',
...
-});
+/* eslint-enable indent */
+
+export default {
+    open: open,
+    close: close,
+    createDialog: createDialog,
+    setOnOpen: function (val) {
+        globalOnOpenCallback = val;
+    }
```
dkanada, line 507: This looks interesting, maybe we should move this to an actual function instead?

PR #1056, opened 2020-04-09
Revamp player (nowplaying.html, remotecontrol.js)

samuel9554:
#923, #924 

Revamp of the page nowplaying.html for a better integration on mobile device.

Please note that I didn't optimize for landscape orientation.

Also I'm missing some experience with Jellyfin. I wasn't enough comfortable to finish the context menu. So if someone want to help I will appreciate.

Issue #1346, opened 2020-06-03
[10.6] When casting, minibar outputs 'undefined' values


hamburglar2160:
When casting to either old or new cast receiver clients on 10.6 master, the minibar (but not the fullscreen 'nowplaying'/'remote control' UIs) display 'undefined' values that were defined in 10.5.5
For music, Artist is 'undefined' (but maybe it should be album artist instead?)
For television, Series Name is 'undefined'
For movies, whatever value (year?) that was placed below movie name is 'undefined'
![image](https://user-images.githubusercontent.com/61879508/83659837-112a1400-a592-11ea-8bae-38f9dbd868e7.png)

Issue #1349, opened 2020-06-03
[10.6] nowplaying.html for TV shows and movies does not have hyperlinked item details


hamburglar2160:
On current 10.6 master with the revamped nowplaying UI, item details for currently playing music and movies (whether casted or not) are clickable and will lead to their respective item detail pages, as introduced in 10.5 for various items detail screens. This same behavior has not been brought to TV shows **edit: and movies.**

Preferred behavior: hyperlink "Episode Name" to the corresponding episode details screen and "Series name" to series details. Further, introduce the hyperlinked "Season # - Episode #" preceding "Episode" so that it reads:
"Season 4, Episode 2
Thirty-Eight Snub
Breaking Bad"

ferferga:
@hamburglar2160 Can we close this now that #1430 is merged?

hamburglar2160:
FWIW for remotecontrol (while casting) the album is clickable but not the artist for music. For movies and TV content, nothing is. Season # and Episode # also aren't displayed. Regardless--closed.

ferferga:
@hamburglar2160 That's one thing I will remember. There is an API inconsistency between Chromecast and remote players and the one that I'm using locally. I'll try to solve them in a Chromecast update.

hamburglar2160:
I'm also experiencing that, though I can click the album (as noted above), the screen goes lank and that album details screen never loads. I'm stuck with the spinner. While casting.

hamburglar2160:
Also on remotecontrol: fastforwarding 30 seconds *sometimes* goes forward 30 seconds but usually goes to next song instead. Works as intended on non-casting nowplaying.html as we know.

PR #1430, opened 2020-06-18
Player hotfixes and improvements

### LINKED ISSUE #923: Revamp the bottom half of the "Now Playing" screen
With the release of 10.5 comes a good opportunity to enhance the audio "Now Playing" ui both for cosmetics and accessibility.
image

1. "**Playlist**" is not the correct term to use here. This is really a "**Queue**" of audio items.  

2. Add a count of the number of items on queue preferably to the left of the "add to playlist" icon. For example: "(81 / 243)"

3. The top portion of the "Now Playing" ui includes the media controls. However when the user scrolls down through the queue (especially on mobile) the media controls disappear out of view. To remedy this, separate the Queue from the controls so that when one scrolls or swipes between the pictured green arrows, only the queue scrolls.

3a. Down the line the queue should reflect the Google Cast queue api implemented on both the v2 sender currently on jf-web and v3/CAF sender on cordova-chromecast-plugin.

### LINKED ISSUE #924: Revamp the top half of the "Now Playing" screen
This is a supplement to #923 

For users listening to audio on Jellyfin (especially on mobile) the "Now Playing" screen tends to get the most attention. Some cosmetic and accessibility changes would be welcome. to the top half of this screen.

1. Add a "shuffle queue" button to the media controls.

2. After the album art (as standard practice) show the album and song title on one line then the Album Artist directly below it.

3. Bring the design in line with 10.5's item details page?

image

### LINKED ISSUE #1195: Finish up the player revamp #1056
Need help to finish the player revamp #1056 

So I've been using the revamp for a few week now and it work great.
It will be nice if someone could finish the context menu. 

If music playing it would be nice to have those options.
- View Artist
- View Album
- Download or sync
- Media Info
- Instant Mix

If series casting to a device
- View Season
- View Serie
- Media Info

If movie casting to a device
- View Movie page
- Media Info

### LINKED ISSUE #1312: Opening the context menu of the "nowplaying.html" bar causes it to be opened two times
**Describe the bug**
Opening the context menu of the "nowplaying.html" bar causes it to be opened two times and throws an uncaught error upon closing.

**To Reproduce**

1. Start playing a song (might be true for other media types as well)
2. Click on the three dots in the corner
3. Click outside the context menu to close it

**Expected behavior**
The context menu dissapears

**Logs**
```
index.html#!/music.html?topParentId=7e64e319657a9516ec78490da03edccb:1 Uncaught (in promise) undefined
Promise.then (async)
show @ itemContextMenu.js?v=10.6.0:1
(anonymous) @ nowPlayingBar.js?v=10.6.0:1
2
```

**Screenshots**
recording-2020-05-27_11 25 07

 - OS: Ubuntu
 - Browser: Chromium
 - Jellyfin Version: `jellyfin-nightly 20200527`

### LINKED ISSUE #1348: [10.6] Desktop Music Player Queue defaults to hidden
The play queue for the revamped 10.6 Music Player on desktop is currently defaulted to 'hidden.'
To show it, one must press the erroneously named "playlist" button. When the next song begins, the queue returns to its default hidden state. 

Preferred behavior: Remove 'playlist' button and make the queue persistent, retaining its 10.5.5 behavior.
[split from #1343]

### LINKED ISSUE #1429: Songs Playlist not showing list of songs
When adding songs to playlist and old playlists which were showing lists of songs isn't working.
Was working in 10.5.5

The Playlist is having 39 minutes of playback. But it isn't showing any songs.
It was working in 10.5.5 this is not happening to all of the songs but some songs.
I am trying to check the pattern but am unable to pinpoint to the exact root cause of this issue.

image

Docker Debian
Nightly 10.6

i couldn't find the exact but try adding Ennio Morricone's soundtrack CD to check the playlist bug.

### LINKED ISSUE #1456: 10.6.0 Unable to open remote control dialog after selecting device
**Describe the bug**
After upgrading to 10.6.0 nightly, the casting button is non-functional after selecting a device to be remotely controlled. To disconnect from the device (or select a new one) a page refresh is required.

**To Reproduce**
1. Launch two instances of jellyfin web.
2. Click the cast icon from the first instance and select the second instance from the menu.
3. Click the cast icon again. No remote control dialog is displayed.

**Expected behavior**
The remote control dialog appears, with options for remote control, display mirroring and to disconnect.

**Logs**
```
RequireJS error: TypeError: playerInfo.supportedCommands is undefined. Failed modules: site.js:1:8034
    onRequireJsError https://example.com/web/scripts/site.js:1
    delayedError https://example.com/web/libraries/alameda.js:986
Uncaught (in promise) TypeError: playerInfo.supportedCommands is undefined
    showActivePlayerMenuInternal https://example.com/web/components/playback/playerSelectionMenu.js?v=20:1
    showActivePlayerMenu https://example.com/web/components/playback/playerSelectionMenu.js?v=20:1
    execCb https://example.com/web/libraries/alameda.js:1233
    defineModule https://example.com/web/libraries/alameda.js:493
    depFinished https://example.com/web/libraries/alameda.js:526
    waitForDep https://example.com/web/libraries/alameda.js:589
playerSelectionMenu.js:1:2909
```

**Screenshots**
n/a

 - OS: Debian 10 (buster)
 - Virtualization: ESXi
 - Browser: Firefox 78.0b9
 - Jellyfin version: Nightly 20200618 installed from the APT repository
 - Reverse proxy: nginx

**Additional context**
I upgraded jellyfin from 10.5.5.

### LINKED ISSUE #1491: "Previous" button does not restart song when listening to audio
**Describe the bug**
In pretty much every music player, if you press the "previous" button (image below) in the middle of a track then it restarts the song. Pressing the button in the first few seconds of a track will load the previous track. In Jellyfin, pressing the previous button at any point will stop the current track and load the one before it in the playlist. Not entirely unreasonable behavior, but unexpected behavior.
image 

**To Reproduce**

1. Play any track that is not the first one in an album
2. Either let it play for a bit or skip to about the halfway point
3. Press the previous button
4. Current track will stop playing and previous track will be loaded

**Expected behavior**

I expect that the song will restart when the button is pressed anytime after the first few seconds.

**Logs**

N/A

**Screenshots**

N/A

 - OS: Docker
 - Browser: Chrome
 - Jellyfin Version: [e.g. 10.5.5]

**Additional context**

ferferga:
**Fixes**

* Playlist not loading when item doesn't have an image (this also fixes library when using ``ListView`` mode). Fixes #1429 and #1431 
* Playlist doesn't disappear anymore when editing/removing items from the queue.
* Doubled context menu. Fixes #1312 
* Playback buttons are always accesible without scrolling on mobile layout. Fixes #1195 [(this specific comment)](https://github.com/jellyfin/jellyfin-web/issues/1195#issuecomment-629692880)
* Fix some overlap issues.
* Fix jumpy elements when volume slider was not visible.
* Show all artists for a song. Each one of them is clickable.
* Remove space in miniplayer when item doesn't have image. *(Screenshots from mobile, but this was also happening in Desktop and TV as well. All layouts fixed)*
**Before**
image
**After**
image

* Fix wrong contextmenu (from the miniplayer) appearing when clicking on an specific place in the right bottom corner
* Replaced context menu action ``ViewArtist`` for ``ViewAlbumArtist`` as specified in the code's comments:
```
// Show Album Artist by default, as a song can have multiple artists, which specific one would this option refer to?
// Although some albums can have multiple artists, it's not as common as songs.
```
* *Some other stuff I'm missing and I don't remember right now :P*

**Improvements**

* Better mobile layout.
* Add music-like shuffle mode. **Not tested in SyncPlay or remote sessions, but it will probably work out of the box**.
  * When enabling shuffle, old playlist will be saved in an array, so it will be used again when shuffle mode is disabled.
* Playlist is always shown in desktop layout. Fixes #1348
* Use our "native" context menu instead of a custom one. Fixes #1195 
* Revamped mobile miniplayer.
* Back button works like it does in most music players **(only audio affected, video is unaffected)**:
  * When more than 5 seconds of the song has been playing, player will start the current track from the beginning.
  * On double click after 5 seconds of the song, player will return to previous track.
  * If no previous item, no matter current song position, track will start from the beginning.

*I didn't build an apk with my branch, but this probably will need to be backported to Android and Expo, so OS media controls work. CC @jellyfin/client*

### Screenshots

**Revamped mobile miniplayer**
image

*Context menu was replaced by the next track button*

**Full screen mobile miniplayer**
image

*Removed rewinds and add repeat and shuffle commands*

**Desktop miniplayer**
image

*Playlist shown by default*

**Shuffle**
shuffle

*Old playlist is reused after disabling shuffle*

**Back button behaviour**
back-button

*Behaviour of the back button as explained above. Last press to back button is a double click*

**Related Issues and PRs**
Closes #923 
Closes #924
Completes and add fixes #1056
Closes #1491
Closes #1456


ferferga:
Although this PR seems to be more feature-wise than fixing-wise, most of the improvements here actually make the new layout usable, so I think this is fitting enough for 10.6. Feel free to disagree with me by commenting in this PR.

src/components/listview/listview.js
```diff
@@ -270,52 +270,54 @@ define(['itemHelper', 'mediaInfo', 'indicators', 'connectionManager', 'layoutMan
...
+                    });
 
-                if (progressHtml) {
-                    html += progressHtml;
+                    if (progressHtml) {
+                        html += progressHtml;
+                    }
+                    html += '</div>';
                 }
-                html += '</div>';
```
ferferga, line 318: With this, items that don't have images won't display anything. Example (from ``remotecontrol.js`` playlist):

image

Shall we use placeholder random backgrounds and logos? (As it's done in ``cardbuilder.js``)

src/components/playback/playbackmanager.js
```diff
@@ -3811,7 +3819,7 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
         });
     };
 
-    PlaybackManager.prototype.shuffle = function (shuffleItem, player, queryOptions) {
```
ferferga, line 3814: Did a bit of cleaning as well in ``playbackManager``. This was removed because it's unused in the function.

src/strings/es.json
```diff
@@ -1560,5 +1560,6 @@
     "EnableDetailsBannerHelp": "Mostrar imagen de banner en el tope de la página de detalles del elemento.",
     "EnableDetailsBanner": "Barra de Detalles",
     "ShowMore": "Mostrar más",
-    "ShowLess": "Mostrar menos"
+    "ShowLess": "Mostrar menos",
+    "ViewAlbumArtist": "Ver artista del álbum"
```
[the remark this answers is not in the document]
ferferga: @dmitrylyzo Good catch!

src/components/listview/listview.js
MrTimscampi: I'd remove the image for every item in that list, especially as there is already the album art near there.

But it's open for discussion, of course :)

ferferga:
My last commit improves the overall look of all the interface when resizing the window using desktop layout as miniplayer and mobile.

I want to keep improving the layout and make one that works well enough on tablets, but definitely not for 10.6

src/components/listview/listview.js
dkanada: I think we should always show the album art, because often queues will have lots of different material. It would also look weird to have some items with an image and some without. We should really just use the placeholder image for the lists.

I **am** in agreement that the image removal is fine on the playback bar though, since it only has a single image.

JustAMan: I support @dkanada here, hiding an image of _some entries_ in a list would make list look like a patchwork, let's use placeholders there. Hiding empty image in a single-item view is fine for me.

hamburglar2160:
I haven't been around a cast device to capture any logs yet but #1346 falls in the 'bug fixing' and #1349 in the 'finishing up revamp/improvements scope of this PR.

KristupasSavickas:
>Show all artists for a song. Each one of them is clickable.

This is huge, thanks!

dkanada:
@ferferga sometimes I add ten or fifteen lines to a pull request but this seems way too large to fit into a single pull request. There are also **tons** of style changes that could easily wait until the next minor release.

ferferga:
@dkanada Most of the style changes you see were already around our code since #1056 was merged. I just moved them outside the ``@media`` rule and applied them only with the ``mobilePlayer`` class (which is applied on ``remotecontrol.js`` when using mobile layout, making everything much more predictable). I only added the new shuffle button bits, the transparency class to solve [this](https://github.com/jellyfin/jellyfin-web/issues/1195#issuecomment-629692880) and fixed the margins in the miniplayer for songs without images and with images. Those changes are necessary so the text doesn't overflow the buttons with the newly shuffle one. 

I know shuffle could've gone into 10.7 instead, but adding the repeat button was something that was mentioned in the original PR as necessary to be accomplished to finish that up (It's coming too late, though). I did that, but it looked really awkward without no button on the right side, so a shuffle mode was necessary. Most of the things that needed fixes were either a few-line changes or doesn't make sense with each other apart.

I tested with all the possible layouts and the default viewports that Chrome & Firefox allows you to use, and I see no issues so far. I'm cool with this going for 10.6.1 but, **sincerely, I don't see the reason to release a half-baked player (+ the blog post announcement that will include it I guess) to deliver the full experience in a minor update later when we can do it directly for 10.6.0. We still have a week ahead for reviewing this if we're aiming at the 26th due date**

ferferga:
@dmitrylyzo Thank you very much for testing:

I'm going to do all your requested changes and test them on Monday, after finishing my exams.

Yes, that's the only caveat that Shuffle mode has at the moment. Before opening this PR, I did some tests with a draft implementation of shuffle mode and everything seemed that was going to be really complicated and it wouldn't take too much time and not be ready for 10.6. Joining all the pieces together turned out to be easier than expected though, so I moved everything to the remote branch and pushed it as is. I can start working on it on Monday. However, I don't think I should do that in this PR: **As pointed by @dkanada and @JustAMan, this PR is quite big already** and I rather have this reviewed and merged for 10.6 (or reviewers will hate me after adding more features here, instead of the requested changes 😁) and **leave that alone in another PR**. Although shuffle isn't 1-1 feature-wise compared with other apps, I think it's in an state good enough to be released in stable imho

About the last point, I'm not sure what you mean: In Spotify and other music apps, if I shuffle, my current queue shuffles completely. If I disable shuffle, the original queue goes back. Shuffling again generates a completely new queue. This is how it works currently. Can you give me examples with other apps?

dmitrylyzo:
> Can you give me examples with other apps?

_I can speak for PC apps_
I used Amarok for a long time until it was deprecated in Kubuntu (due to Qt4). It has random play ~ shuffle on each playlist repeat (as I suggested). Now I use Cantata+MPD with same logic.

In addition to my suggestion, you can shuffle the entire list, but move the current item to the beginning.

ferferga:
@dmitrylyzo I was talking about the last point. The first one it's perfectly clear :).

What do you mean with "It's applied once exactly?"

dmitrylyzo:
> What do you mean with "It's applied once exactly?"

You enable `Repeat`.
You enable `Shuffle`.
It shuffles playlist.
It plays entire list.
It starts from the beginning, but queue will be the same again and again. (_"Shuffle" is applied once_)

ferferga:
@dmitrylyzo Now I understand

* Your first point applies to all players I used as well, so no discussions here

* This is the first time I heard about this behaviour. Based on the players I tested:

    * *On Spotify, Google Play Music, Deezer, Tidal and Apple Music*: Shuffling does show you the resulting playlist, so you can see what's coming next. When it finishes, if no repeat, music will stop. If repeat, the playlist will start again from the beginning (the track you were listening when you hit shuffle) without further shuffling. Aside from *the track you were listening when you hit shuffle*, **this is the shuffling concept I developed for Jellyfin**. Something different than this is a bug.

    * I rechecked on **VLC** after a long time without using it (looking at you, streaming services): Shuffling doesn't result in a different playlist, visually. VLC simply takes a random song from the list. When all the songs of the playlist have been played, **even without repeat enabled** you will end up relistening to the whole playlist over and over, as the player doesn't take in account what we have listened: **it just picks items randomly**. Enabling *Repeat All* mode doesn't make sense in this context, just *Repeat One* changes how the player behaves (playing the current song in a loop).

* **Phonograph/Gelli**: Same as streaming services.

Although I'm used to the "new" shuffling methods of the streaming services, using VLC brought me good memories of the old times when I started listening to music: it was Windows Media Player with a (back then) really modern Windows XP. Doing these tests remembered me how it worked, in exactly the same way, and, to be honest, I miss it a lot. As a side note, I think Spotify (back when it was invite only) did it in the same way.

Given these results, I think you'll immediately agree with me that we should leave this behaviour intact for now, people will probably be used to what they have in other platforms. However, I'm 100% open (and plan to) add an option in settings, so users can switch between shuffling methods. But, as in my other message, that's for another PR :).

Is that OK in your opinion?

Btw, it's really cool to see how something as small as shuffle can be understood so differently between apps and people :P

dmitrylyzo:
> Given these results, I think you'll immediately agree with me that we should leave this behaviour intact for now, people will probably be used to what they have in other platforms. However, I'm 100% open (and plan to) add an option in settings, so users can switch between shuffling methods. But, as in my other message, that's for another PR :).
> 
> Is that OK in your opinion?

I agree with option.
I am just used to turning on music and forget about re-shuffling. :stuck_out_tongue:

src/components/nowPlayingBar/nowPlayingBar.js
```diff
@@ -545,21 +586,27 @@ define(['require', 'datetime', 'itemHelper', 'events', 'browser', 'imageLoader',
...
-                            itemContextMenu.show(Object.assign({
-                                item: item,
-                                user: user
-                            }, options ));
+                    if (!layoutManager.mobile) {
+                        let contextButton = document.querySelector('.nowPlayingBar').querySelector('.btnToggleContextMenu');
+                        // We remove the previous event listener by replacing the item in each update event
+                        let contextButtonClone = contextButton.cloneNode(true);
+                        contextButton.parentNode.replaceChild(contextButtonClone, contextButton);
+                        contextButton = document.querySelector('.nowPlayingBar').querySelector('.btnToggleContextMenu');
```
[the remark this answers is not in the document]
ferferga: @dmitrylyzo Here seems to work, but not in ``remotecontrol.js``. I guess it's because we reconstruct the whole DOM in the miniplayer. I'm going to leave them as I did for the sake of consistency.

I plan on revisit the players later, so this might be changed in the future.

ferferga:
## Changes in latest commits

* Styling cleanup
* Items in queue without images will display a placeholder.
* Cards placeholders now display **Audio** items with *audiotrack* icon
* Current playing items will be kept first when shuffled. Previous item in queue will also be shuffled (this might get improvements later)
* Fixed Chromecast and remote players reporting ``undefined``. Fixes #1346 
* Player can now sent shuffle actions to Chromecast and remote players now support shuffle mode.

*This avoids a crash in the player when casting. Chromecast still needs to have this implemented in their own code, something which I'm going to do soon. What I have done is allow playbackManager to **send** the "signal" to the remote players, but remote players also need to have this implemented*. **This means that shuffling when casting to Chromecast doesn't do anything (but doesn't crash as well)**

src/components/remotecontrol/remotecontrol.js
```diff
@@ -650,12 +708,40 @@ define(['browser', 'datetime', 'backdrop', 'libraryBrowser', 'listView', 'imageL
...
+                        } else {
+                            playbackManager.setQueueShuffleMode('Sorted', currentPlayer);
+                        }
+                    }
+                });
+            }
+
+            context.querySelector('.btnPreviousTrack').addEventListener('click', function (e) {
+                if (currentPlayer) {
+                    if (currentPlayer.id === 'htmlaudioplayer' && (currentPlayer._currentTime >= 5 || !playbackManager.previousTrack(currentPlayer))) {
```
MrTimscampi, line 725:
```suggestion
                    if (currentPlayer.id === 'htmlaudioplayer' && (currentPlayer._currentTime >== 5 || !playbackManager.previousTrack(currentPlayer))) {
```

src/components/itemContextMenu.js
```diff
@@ -288,10 +288,11 @@ define(['apphost', 'globalize', 'connectionManager', 'itemHelper', 'appRouter',
...
             });
         }
-
-        if (options.openArtist !== false && item.ArtistItems && item.ArtistItems.length) {
+        // Show Album Artist by default, as a song can have multiple artists, which specific one would this option refer to?
+        // Although some albums can have multiple artists, it's not as common as songs.
+        if (options.openArtist !== false && item.AlbumArtists && item.AlbumArtists.length) {
             commands.push({
-                name: globalize.translate('ViewArtist'),
+                name: globalize.translate('ViewAlbumArtist'),
```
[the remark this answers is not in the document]
MrTimscampi: I do think that it should send to the album artist, though.

src/components/playback/playbackmanager.js
```diff
@@ -4000,6 +4030,9 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
             case 'SetRepeatMode':
                 this.setRepeatMode(cmd.Arguments.RepeatMode, player);
                 break;
+            case 'SetQueueShuffleMode':
+                this.setQueueShuffleMode(cmd.Arguments.ShuffleMode, player);
+                break;
```
dkanada, line 4035: Why not match `SetRepeatMode` and label this `SetShuffleMode` to keep it more generic for now? This is an excellent change though, I'm always a fan of better command support.

ferferga: @dkanada That's how it was called initially. However, didn't want to make a confusion between the "normal" shuffle (the one you run in the library browser, item details, context menu, etc) and the queue one. @jelyfin/web thoughts?

src/components/itemContextMenu.js
```diff
@@ -458,7 +459,7 @@ define(['apphost', 'globalize', 'connectionManager', 'itemHelper', 'appRouter',
                     getResolveFunction(resolve, id)();
                     break;
                 case 'artist':
-                    appRouter.showItem(item.ArtistItems[0].Id, item.ServerId);
+                    appRouter.showItem(item.AlbumArtists[0].Id, item.ServerId);
```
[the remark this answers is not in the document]
ferferga: @MrTimscampi We already check that in the function above, where we push options to the array that builds the context menu. If no Album Artist, no option will be displayed, so no way this would be executed.

src/components/nowPlayingBar/nowPlayingBar.js
```diff
@@ -146,13 +157,38 @@ define(['require', 'datetime', 'itemHelper', 'events', 'browser', 'imageLoader',
...
-        elem.querySelector('.previousTrackButton').addEventListener('click', function () {
+        elem.querySelector('.previousTrackButton').addEventListener('click', function (e) {
+            if (currentPlayer) {
+                if (currentPlayer.id === 'htmlaudioplayer' && (currentPlayer._currentTime >= 5 || !playbackManager.previousTrack(currentPlayer))) {
+                    // Cancel this event if doubleclick is fired
+                    if (e.detail > 1 && playbackManager.previousTrack(currentPlayer)) {
+                        return;
+                    }
+                    playbackManager.seekPercent(0, currentPlayer);
+                    // This is done automatically by playbackManager, however, setting this here gives instant visual feedback
```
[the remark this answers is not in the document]
ferferga: @MrTimscampi The default for the dblclick JS event. We can set our own, but IMO it's fine and on par with what you would expect in all other music apps.

For context, the delay it has is the same as the one you have when double tapping on any mobile browser and it zooms.

Check https://developer.mozilla.org/en-US/docs/Web/API/Element/dblclick_event example also.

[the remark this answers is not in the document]
MrTimscampi: Hm, it's nearly instantaneous here. It should be mostly fine if we let playbackManager handle the position instead of forcing it.

src/components/playback/playbackmanager.js
MrTimscampi: I think all these should eventually be merged, so reusing the same name shouldn't be an issue.

src/components/nowPlayingBar/nowPlayingBar.js
```diff
@@ -146,13 +157,38 @@ define(['require', 'datetime', 'itemHelper', 'events', 'browser', 'imageLoader',
...
 
+        elem.querySelector('.previousTrackButton').addEventListener('dblclick', function () {
             if (currentPlayer) {
                 playbackManager.previousTrack(currentPlayer);
             }
         });
 
+        elem.querySelector('.btnShuffle').addEventListener('click', function () {
+            if (currentPlayer) {
+                if (playbackManager.getPlaylistShuffleMode(currentPlayer) === 'Sorted') {
```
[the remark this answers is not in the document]
ferferga: @MrTimscampi FYI this was my initial implementation, in order to keep things simple :). Changed it to using values to keep consistency (regarding Repeat modes) and also because I thought that maybe external third party players that use our playbackManager or some crazy functionality we add in the future could simply keep shuffling without switching modes.

Didn't thought about re-adding it as a helper function and I think it's excellent because its simple enough and gives full control.

Is there any problem if I add it and simply leave it unused? Or shall I change all my references in remotecontrol and nowplayingbar to it?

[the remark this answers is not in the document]
MrTimscampi: Ideally, nothing should be unused ;)

If it's not much work, feel free to add it and change references. Otherwise, simply adding a TODO will do for now :)

src/components/playback/playqueuemanager.js
```diff
@@ -56,6 +58,36 @@ define([], function () {
...
+            if (this._playlist[i] === currentPlaylistItem) {
+                this._playlist.splice(i, 1);
+                continue;
+            }
+            const j = Math.floor(Math.random() * i);
+            const temp = this._playlist[i];
+            this._playlist[i] = this._playlist[j];
+            this._playlist[j] = temp;
+        }
+        this._playlist.unshift(currentPlaylistItem);
```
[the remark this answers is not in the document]
ferferga: @dmitrylyzo Yes, in the splice. Afterwards, with the continue, it goes to the following item.

After adding this I did a search multiple times and I couldn't find duplicate items. You're saying this because you actually had this happening?

src/components/playback/playqueuemanager.js
```diff
@@ -176,21 +208,39 @@ define([], function () {
...
+        this._shuffleMode = 'Sorted';
     };
 
     PlayQueueManager.prototype.setRepeatMode = function (value) {
-
-        this._repeatMode = value;
+        if (value === 'RepeatOne' || value === 'RepeatAll' || value === 'RepeatNone') {
+            this._repeatMode = value;
+        } else {
+            throw new TypeError('invalid value provided for setRepeatMode');
```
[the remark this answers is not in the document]
ferferga: @MrTimscampi Yeah, with the current code is impossible as well. I thought about adding this here basically in case someone wants to touch something in the remote players and don't have a clue why shuffling doesn't work :)

src/components/itemContextMenu.js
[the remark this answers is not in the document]
ferferga: @MrTimscampi This is something we should change in Weblate instead, right? Or you mean replacing the key?

I just reused the ViewArtist one (renamed it so it gets retranslated). Or you mean creating an entirely new key? In that case, we should remove the old one.

src/components/playback/playqueuemanager.js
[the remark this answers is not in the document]
dmitrylyzo: > You're saying this because you actually had this happening?

Yes.
I placed breakpoints in "remove current" block and before `unshift`.
It didn't stop in "remove current" block.
Before `unhisht`, I still have current item.

Fun fact, when current item is shuffled at the end of queue, the latter is virtually looped - player switches to "first" current item instead of shuffled last.

src/components/nowPlayingBar/nowPlayingBar.js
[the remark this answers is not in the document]
ferferga: @MrTimscampi My bad, I thought you meant the back behaviour. Here is a video in Chrome (maybe in FF it's faster and you didn't notice it, I can say in Chrome it is). Maybe I'm too nitpicky, but I'm able to tell the difference easily

Both videos hit back at the 1:40 timestamp. Bear in mind with the low framerates of gifs, it's not the best way to show this here, to be honest.

**With "manual" change:**
after

**With ``playbackmanager``**:
before

src/components/playback/playqueuemanager.js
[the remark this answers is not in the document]
ferferga: @dmitrylyzo Check how I managed this in my next commits. I removed it before shuffling and then added it afterwards.

src/components/remotecontrol/remotecontrol.js
ferferga: @MrTimscampi This gives ``unexpected token error``

src/components/nowPlayingBar/nowPlayingBar.js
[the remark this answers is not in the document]
MrTimscampi: Hm, I think the difference is negligible and it's better to not have stuff interacting with something that's controlled in a central component without going through that component, imo.

It's not blocking, but I think it's better design-wise to let the playbackManager handle the seekbar :p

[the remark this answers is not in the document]
ferferga: @MrTimscampi I'll add a TODO so we can investigate later the delay that playbackManager has in the seek function. I'm full for this because makes the difference between a webpage and the "sensation" that we're in a native app, which I think Android users will appreciate.

src/components/remotecontrol/remotecontrol.css
```diff
@@ -157,42 +157,106 @@
...
+    margin: 0;
+    padding-right: 0;
+    border-radius: 0;
+}
+
+.layout-mobile .playlistSectionButton .volumecontrol {
+    margin: 0;
+    padding-right: 0;
+    border-radius: 0;
+}
```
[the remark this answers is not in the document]
ferferga: @dmitrylyzo This is the original styling that was applied in #1056, I just applied only to mobile layout. For the padding, done in my next commits.

src/components/cardbuilder/card.css
```diff
@@ -167,8 +167,9 @@ button::-moz-focus-inner {
     position: relative;
     background-clip: content-box !important;
     color: inherit;
+}
 
-    /* This is only needed for scalable cards */
+.cardImageContainer.cardScalable {
```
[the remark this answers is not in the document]
ferferga: @dmitrylyzo Good catch!

src/components/cardbuilder/cardBuilder.js
```diff
@@ -1534,8 +1534,11 @@ import 'programStyles';
                 case 'MusicAlbum':
                     return '<span class="cardImageIcon material-icons album"></span>';
                 case 'MusicArtist':
+                    return '<span class="cardImageIcon material-icons person"></span>';
```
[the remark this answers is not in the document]
ferferga: @dmitrylyzo Ooops! Didn't know this about the switch syntax.

src/components/remotecontrol/remotecontrol.js
```diff
@@ -316,25 +354,25 @@ define(['browser', 'datetime', 'backdrop', 'libraryBrowser', 'listView', 'imageL
...
+                if ('RepeatAll' == repeatMode) {
+                    toggleRepeatButton.innerHTML = "<span class='material-icons repeat'></span>";
+                    toggleRepeatButton.classList.add('repeatButton-active');
+                } else if ('RepeatOne' == repeatMode) {
+                    toggleRepeatButton.innerHTML = "<span class='material-icons repeat_one'></span>";
+                    toggleRepeatButton.classList.add('repeatButton-active');
+                } else {
+                    toggleRepeatButton.innerHTML = "<span class='material-icons repeat'></span>";
+                    toggleRepeatButton.classList.remove('repeatButton-active');
+                }
```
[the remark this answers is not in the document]
dmitrylyzo: _17 hours later_
Oh, wait. Won't work with `classList` because of add/remove.

src/components/playback/playqueuemanager.js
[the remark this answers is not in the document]
dmitrylyzo: It sounds promising.

src/components/nowPlayingBar/nowPlayingBar.js
```diff
@@ -501,21 +534,30 @@ define(['require', 'datetime', 'itemHelper', 'events', 'browser', 'imageLoader',
...
+                    text.innerHTML = textLines[1].text;
+                    secondaryText.appendChild(text);
+                }
             }
 
-            return `<div ${cssClass}>${nowPlayingText}</div>`;
-
-        }).join('');
+            if (textLines[0].text) {
+                if (textLines[0].text) {
```
dmitrylyzo, line 552: overdid :p
also you could leave creation order: `itemText`, `secondaryText`

src/components/remotecontrol/remotecontrol.js
```diff
@@ -118,19 +120,30 @@ define(['browser', 'datetime', 'backdrop', 'libraryBrowser', 'listView', 'imageL
...
+                            if (artist !== item.ArtistItems.slice(-1)[0]) {
+                                artistsSeries += ', ';
+                            }
+                        }
+                    } else if (item.Artists) {
+                        // For some reason, Chromecast Player doesn't return a item.ArtistItems object, so we need to fallback
+                        // to normal item.Artists item.
+                        // TODO: Normalise fields returned by all the players
+                        for (const artist of item.Artists) {
+                            artistsSeries += `<a>${artist}</a>`;
```
dmitrylyzo, line 139: not `emby-linkbutton`, just `a`?

src/components/remotecontrol/remotecontrol.js
```diff
@@ -316,25 +353,37 @@ define(['browser', 'datetime', 'backdrop', 'libraryBrowser', 'listView', 'imageL
...
-            } else if ('RepeatOne' == repeatMode) {
-                toggleRepeatButton.innerHTML = "<span class='material-icons repeat_one'></span>";
-                toggleRepeatButton.classList.add('repeatButton-active');
-            } else {
-                toggleRepeatButton.innerHTML = "<span class='material-icons repeat'></span>";
-                toggleRepeatButton.classList.remove('repeatButton-active');
+            let toggleRepeatButtons = context.querySelectorAll('.repeatToggleButton');
+            const cssClass = 'repeatButton-active';
+            let innHtml = '<span class="material-icons repeat"></span>';
+            let repeatOn = undefined;
```
dmitrylyzo, line 359: Why didn't you use boolean values?

src/components/remotecontrol/remotecontrol.js
ferferga: @dmitrylyzo Yes, because Chromecast doesn't return ``ArtistItems`` for some reason. No way I could generate a link without the id of the artist.

dmitrylyzo:
I got `throw new TypeError('invalid value for repeatMode');` on stop.

ferferga:
@dmitrylyzo Hmm true, because at stop it becomes ``undefined``. Shall we keep the switch or better revert back to the ``if`` structure? Not sure what's the best approach now.

src/components/nowPlayingBar/nowPlayingBar.js
ferferga: @dmitrylyzo Lol, not sure what I was thinking here :P

src/components/remotecontrol/remotecontrol.js
ferferga: @dmitrylyzo Same as above, not sure what I was thinking here :P

src/components/playback/playqueuemanager.js
```diff
@@ -127,6 +155,12 @@ define([], function () {
         var currentPlaylistItemId = this.getCurrentPlaylistItemId();
         var isCurrentIndex = playlistItemIds.indexOf(currentPlaylistItemId) !== -1;
 
+        if (this._sortedPlaylist.length <= playlistItemIds.length) {
+            this._sortedPlaylist = this._sortedPlaylist.splice(0).filter(function (item) {
+                return playlistItemIds.indexOf(item.PlaylistItemId) === -1;
+            });
+        }
```
dmitrylyzo, line 162: `playlistItemIds` is items to remove, so `this._sortedPlaylist.length <= playlistItemIds.length` doesn't work.
Also I see no reason to use `splice`.
```suggestion
        this._sortedPlaylist = this._sortedPlaylist.filter(function (item) {
            return !playlistItemIds.includes(item.PlaylistItemId);
        });
```

src/components/remotecontrol/remotecontrol.js
```diff
@@ -316,25 +353,39 @@ define(['browser', 'datetime', 'backdrop', 'libraryBrowser', 'listView', 'imageL
...
+                default:
+                    throw new TypeError('invalid value for repeatMode');
+            }
+
+            for (const toggleRepeatButton of toggleRepeatButtons) {
+                if (!repeatOn) {
+                    toggleRepeatButton.classList.remove(cssClass);
+                } else {
+                    toggleRepeatButton.classList.add(cssClass);
+                }
```
dmitrylyzo, line 381:
```suggestion
                toggleRepeatButton.classList.toggle(cssClass, repeatOn);
```
In fact, this can be applied in many cases that we have.

src/components/remotecontrol/remotecontrol.js
```diff
@@ -470,6 +523,28 @@ define(['browser', 'datetime', 'backdrop', 'libraryBrowser', 'listView', 'imageL
             updateRepeatModeDisplay(playbackManager.getRepeatMode(player));
         }
 
+        function onShuffleQueueModeChange() {
+            let shuffleMode = playbackManager.getQueueShuffleMode(this);
```
dmitrylyzo, line 527: Maybe `playbackManager.getQueueShuffleMode` shouldn't return an invalid value? And we can rely on it.

src/components/playback/playqueuemanager.js
ferferga: @dmitrylyzo The condition is wrong, correct. I added it after having it working and, although I retested everything before pushing, for some reason it seems that I missed this.

* ``splice`` is required, otherwise the resulting array will be empty. In this function, the playlist variable (returned from ``getCurrentPlaylist`` is already spliced, that's why not used later. Just tested and it doesn't work without splice.

* Also, we must check against -1 index, no clue why but your return doesn't work as well.

What I did here is replicate with the sorted playlist array what it's done in the current playlist. It's not the ideal solution but we're too late in the feature freeze and this is not yet migrated to ES6, so I think it makes sense to leave all the refactoring for later (if nobody comes first, I want to migrate the playback components to ES6, already got some appreciation to them :P) and stay in the safe side, even in the little details. Didn't inspect the objects deeply as well to improve this, will do on ES6 migration.

src/components/remotecontrol/remotecontrol.js
ferferga: @dmitrylyzo Didn't know about ``toggle``! This is a life saver! 😱 Thank you!

Will apply it only to the new toggles I added/edited in this PR. Will keep all the ones that were previously in this file intact, don't want to make the diffs any larger. Will take a note of this for the ES6 migration.

src/components/playback/playqueuemanager.js
dmitrylyzo: >     * `splice` is required, otherwise the resulting array will be empty.

Hmm, but it works without it. Moreover, `filter` returns a filtered copy.

ferferga: I see what is the problem now 👍

src/components/remotecontrol/remotecontrol.js
ferferga: @dmitrylyzo I think I know now what was causing us the issue in the first place. No need to use ``undefined`` in the switch anymore, repeat values from playbackmanager will always be ``RepeatNone``, ``RepeatAll`` and ``RepeatOne``. Same for shuffle. Let me test that and hopefully this get solved in next commit.

src/nowplaying.html
```diff
@@ -163,20 +177,17 @@ <h2 class="nowPlayingPageTitle"></h2>
         </div>
         <div class="playlistSection">
             <div class="playlistSectionButton flex align-items-center justify-content-center">
```
dmitrylyzo, line 179: For TV. Keep focus inside container with horizontal keyboard navigation.
```suggestion
            <div class="playlistSectionButton flex align-items-center justify-content-center focuscontainer-x">
```

ferferga:
@dmitrylyzo Hmm, TV layout doesn't have nowplayingbar, only remotecontrol component 😨. At least, never saw it. You switched while running?

No idea why that happens there

src/components/remotecontrol/remotecontrol.js
```diff
@@ -465,9 +511,29 @@ define(['browser', 'datetime', 'backdrop', 'libraryBrowser', 'listView', 'imageL
...
+            let context = dlg;
+            const cssClass = 'shuffleQueue-active';
+            let shuffleButtons = context.querySelectorAll('.btnShuffleQueue');
+
+            for (let shuffleButton of shuffleButtons) {
+                switch (shuffleMode) {
+                    case 'Sorted':
+                        shuffleButton.classList.toggle(cssClass, false);
```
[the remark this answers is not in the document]
ferferga: @dmitrylyzo It does the work right though

src/components/playback/playbackmanager.js
dkanada: Were you going to add this in a future commit?

ferferga: @dkanada Is already done in b7714ee

src/components/remotecontrol/remotecontrol.js
```diff
@@ -252,14 +252,16 @@ define(['browser', 'datetime', 'backdrop', 'libraryBrowser', 'listView', 'imageL
...
-                case 'RepeatNone':
-                    playbackManager.setRepeatMode('RepeatAll');
-                    break;
                 case 'RepeatAll':
                     playbackManager.setRepeatMode('RepeatOne');
                     break;
                 case 'RepeatOne':
                     playbackManager.setRepeatMode('RepeatNone');
+                    break;
+                default:
```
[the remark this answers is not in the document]
ferferga: @dmitrylyzo Given that we receive inconsistent responses from playbackManager, I thought it was a better idea to have all the cases handled, at least until we have some consistency in playbackManager. However, I've been playing with this a little and it seems fine, so I'm going to remove them as you suggest.

src/components/remotecontrol/remotecontrol.js
```diff
@@ -362,24 +404,28 @@ define(['browser', 'datetime', 'backdrop', 'libraryBrowser', 'listView', 'imageL
...
-                    nowPlayingVolumeSlider.value = volumeLevel || 0;
+                var nowPlayingVolumeSlider = context.querySelector('.nowPlayingVolumeSlider');
+                var nowPlayingVolumeSliderContainer = context.querySelector('.nowPlayingVolumeSliderContainer');
+
+                if (nowPlayingVolumeSlider) {
+                    if (showVolumeSlider) {
+                        nowPlayingVolumeSliderContainer.classList.remove('hide');
+                    } else {
+                        nowPlayingVolumeSliderContainer.classList.add('hide');
+                    }
```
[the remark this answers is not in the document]
ferferga: @dmitrylyzo Little change, so did it anyway. *Why not?* :P

src/components/remotecontrol/remotecontrol.js
```diff
@@ -465,9 +503,28 @@ define(['browser', 'datetime', 'backdrop', 'libraryBrowser', 'listView', 'imageL
...
+        function onShuffleQueueModeChange() {
+            let shuffleMode = playbackManager.getQueueShuffleMode(this);
+            let context = dlg;
+            const cssClass = 'shuffleQueue-active';
+            let shuffleButtons = context.querySelectorAll('.btnShuffleQueue');
+
+            for (let shuffleButton of shuffleButtons) {
+                switch (shuffleMode) {
+                    case 'Shuffle':
+                        shuffleButton.classList.toggle(cssClass, true);
```
dmitrylyzo, line 519:
```suggestion
                        shuffleButton.classList.add(cssClass);
```

src/components/remotecontrol/remotecontrol.js
```diff
@@ -465,9 +503,28 @@ define(['browser', 'datetime', 'backdrop', 'libraryBrowser', 'listView', 'imageL
...
+                        break;
+                    case 'Sorted':
+                    default:
+                        shuffleButton.classList.toggle(cssClass, false);
```
dmitrylyzo, line 523:
```suggestion
                        shuffleButton.classList.remove(cssClass);
```

ferferga:
@dmitrylyzo I think latest commits already make the playlist reload only when necessary. Still, this component needs lots of works, so the whole playlist is updated only when necessary, but not for this PR imo.

I also fixed the focus issue on TV (also applies to the rest of layouts, so in the future we can add some kind of indication for the active item on desktop layout)

src/components/nowPlayingBar/nowPlayingBar.js
```diff
@@ -159,7 +159,7 @@ define(['require', 'datetime', 'itemHelper', 'events', 'browser', 'imageLoader',
 
         elem.querySelector('.previousTrackButton').addEventListener('click', function (e) {
             if (currentPlayer) {
-                if (currentPlayer.id === 'htmlaudioplayer' && (currentPlayer._currentTime >= 5 || !playbackManager.previousTrack(currentPlayer))) {
+                if (lastPlayerState.NowPlayingItem.MediaType === 'Audio' && (currentPlayer._currentTime >= 5 || !playbackManager.previousTrack(currentPlayer))) {
```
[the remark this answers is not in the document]
ferferga: @dmitrylyzo I have a future plan of lending all the music and audio playback to **howler**, so we can switch between HTML Audio and WebAudio + a few other goodies like gapless playback (*The library to use ofc needs discussion in Matrix and so on of course*). With this check, we're **explicitly checking for HTML Audio**, so other audio plugins/"players" like WebAudio, howler, etc are excluded. However, if we explicitly check for the item type, we're 100% sure that what we're playing is Audio, no matter what we use to play it.

At first, I thought about making the change later, once we roll those changes to the music players. However, no harm in doing this now that we're polishing this PR :P

src/components/remotecontrol/remotecontrol.js
```diff
@@ -163,11 +176,29 @@ define(['browser', 'datetime', 'backdrop', 'libraryBrowser', 'listView', 'imageL
                 maxHeight: 300 * 2
             }) : null;
 
-            console.debug('updateNowPlayingInfo');
+            let contextButton = context.querySelector('.btnToggleContextMenu');
+            // We remove the previous event listener by replacing the item in each update event
+            let contextButtonClone = contextButton.cloneNode(true);
+            contextButton.parentNode.replaceChild(contextButtonClone, contextButton);
+            contextButton = context.querySelector('.btnToggleContextMenu');
```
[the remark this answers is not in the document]
ferferga: @dmitrylyzo As I said [here](https://github.com/jellyfin/jellyfin-web/pull/1430#discussion_r443620102), it doesn't seem to work without the replace child, so I'm going to commit your suggestion straightaway, without removing it.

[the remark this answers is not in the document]
dmitrylyzo: Looks like `contextButton` exists in the template.
As a possible solution, you could extract options that passed into `itemContextMenu.show` (`itemContextMenuOptions`) and update them here. `contextButton.addEventListener` goes  to `bindEvents`.
No rush - you can do it in the next PR.

ferferga:
@dmitrylyzo *For some reason I can't reply you to the review comment, so I do it here*:

Okay, I'll note this for a future PR.

Last 3 commits adds the following:

* Stop playback option in context menu (only mobile layout)
* Clear queue option in context menu
image

**Borrar la cola**: Clear queue
**Detener la reproducción**: Stop playback
* Option parity between miniplayer and remotecontrol in context menu: *Before, miniplayer showed more options, like download, see media info, etc. Now, those are also in remotecontrol, as you can see*

CC @dmitrylyzo @MrTimscampi and @dkanada

src/components/remotecontrol/remotecontrol.js
```diff
@@ -163,11 +176,38 @@ define(['browser', 'datetime', 'backdrop', 'libraryBrowser', 'listView', 'imageL
...
+            let contextButton = context.querySelector('.btnToggleContextMenu');
+            // We remove the previous event listener by replacing the item in each update event
+            const autoFocusContextButton = document.activeElement === contextButton;
+            let contextButtonClone = contextButton.cloneNode(true);
+            contextButton.parentNode.replaceChild(contextButtonClone, contextButton);
+            contextButton = context.querySelector('.btnToggleContextMenu');
+            if (autoFocusContextButton) {
+                contextButton.focus();
+            }
+            let stopPlayback = false || layoutManager.mobile;
```
dmitrylyzo, line 188: `false` seems unnecessary.
Do you want a guaranteed boolean in case of `undefined`?
```suggestion
            const stopPlayback = !!layoutManager.mobile;
```

src/components/playback/playbackmanager.js
```diff
@@ -2969,6 +2970,7 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
             } else {
                 self._playQueueManager.queue(items);
             }
+            events.trigger(player, 'playlistitemmove');
```
dmitrylyzo, line 2973: By meaning, `playlistitemadd` fits better. But it isn't handled.

ferferga: @dmitrylyzo Didn't know that we had an event for that and didn't want to mess so much with them. I checked and added it to ``remotecontrol.js```

Good catch!

PR #1444, opened 2020-06-19
Remove unused  loadingDialog & serverRestartDialog

grafixeyehero:
For reference **loadingDialog last used selectserver.js for wakeOnLan** that was Remove broken features from web interface #37 https://github.com/jellyfin/jellyfin-web/commit/2f3f17369ecbf691f2e060909338b6058b15bb28


dkanada:
I think we also have an unused restartServerDialog if you want to remove that as well.

grafixeyehero:
> 
> 
> I think we also have an unused restartServerDialog if you want to remove that as well.

did you mean serverRestartDialog?

dkanada:
yeah that's the one

PR #1448, opened 2020-06-20
Make the portable web archive and dir consistent

joshuaboniface:
*Changes**
Make the portable web archive and directory names consistent with the server archives. They use `jellyfin-server_version`, not `jellyfin-server-version`, so this should too.

**Issues**
N/A

### DIFF
--- a/deployment/build.portable
+++ b/deployment/build.portable
@@ -15,8 +15,8 @@ fi
 
 # build archives
 npx yarn install
-mv dist jellyfin-web-${version}
-tar -czf jellyfin-web-${version}-portable.tar.gz jellyfin-web-${version}
+mv dist jellyfin-web_${version}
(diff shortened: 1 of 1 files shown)


joshuaboniface:
Command from the main repo (server) for reference:

```
tar -czf jellyfin-server_${version}_portable.tar.gz -C dist jellyfin-server_${version}
```

dkanada:
I was just going to rename the archive files instead, because underscores are terrible. :laughing:

EraYaN:
Well they at least signify that they are different from the dash.

joshuaboniface:
Underscores aren't terrible, they're the standard separator for versions and names, especially since our name has a dash in it too, which makes filtering out the version with dashes much harder and more error prone.

PR #1458, opened 2020-06-21
Add page switching animations

MrTimscampi:
**Changes**

* Adds page fadein/fadeout.
* Adds fadein for sections.

~~**Known issues**~~
~~After playing a video, the page remains hidden.~~

**Preview**
output

*Combined with the details page fixes*
output

**Issues**

### DIFF
--- a/src/assets/css/librarybrowser.css
+++ b/src/assets/css/librarybrowser.css
@@ -953,7 +953,7 @@ div.itemDetailGalleryLink.defaultCardBackground {
 }
 
 .verticalSection {
-    animation: fadein 0.2s;
+    animation: fadein 0.7s;
 }
 
 .verticalSection-extrabottompadding {
--- a/src/assets/css/librarybrowser.css
+++ b/src/assets/css/librarybrowser.css
@@ -952,6 +952,10 @@ div.itemDetailGalleryLink.defaultCardBackground {
     flex-wrap: wrap;
 }
 
+.verticalSection {
+    animation: fadein 0.2s;
+}
+
 .verticalSection-extrabottompadding {
     margin-bottom: 2.7em;
 }
--- a/src/assets/css/site.css
+++ b/src/assets/css/site.css
@@ -51,7 +51,7 @@ body {
 }
 
 .mainAnimatedPage {
-    contain: style size !important;
+    animation: fadein 0.5s;
 }
 
 .pageContainer {
--- a/src/assets/css/site.css
+++ b/src/assets/css/site.css
@@ -50,8 +50,12 @@ body {
     -webkit-font-smoothing: antialiased;
 }
 
-.mainAnimatedPage {
-    animation: fadein 0.5s;
+.fadein {
+    animation: fadein 0.7s;
+}
+
+.fadein-fast {
+    animation: fadein 0.7s;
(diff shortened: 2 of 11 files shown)


ferferga:
@MrTimscampi This is really cool! You can add also another class with slower animations and make all the current ones 0.2 (like we used for images) to make our new and shiny "Enable faster animations" more useful :).

Also, not sure if you already did this, but did you take a look at #1229 to see how I switched the scrollbar on video? Maybe it can be useful for fixing the video player thing?

MrTimscampi:
> Also, not sure if you already did this, but did you take a look at #1229 to see how I switched the scrollbar on video? Maybe it can be useful for fixing the video player thing?

This is the next step :) I think I need to remove the fadeout class there to fix the issue.

Also good point about using the faster animation setting.

src/components/images/imageLoader.js
```diff
@@ -72,7 +72,9 @@ import 'css!./style';
         }
 
         if (userSettings.enableBlurhash()) {
-            if (!target.classList.contains('blurhashed', 'non-blurhashable') && blurhashstr) {
+            if (!target.classList.contains('blurhashed', 'non-blurhashable')
+                && blurhashstr
+                && target.querySelectorAll('canvas').length === 0) {
```
ferferga, line 77: I'm not sure exactly why you saw this, but this doesn't make sense, given that we remove the data-blurhash attribute of the elements (so in a further pass it would be impossible to draw the canvas again, as the attribute doesn't exist) and we also add a class that we use to check. This is completely redundant. See lines 27, 48 and 49.

MrTimscampi: To paraphrase my comment on Matrix: I had a weird issue where going to a library page, then to itemDetails then back to the library was causing items to have 3 canvases, one of which had an opacity of 0 and two with an opacity of 1.

Not finding what could cause this, I added that check in order to make sure it'll only create a canvas if there is none.

src/assets/css/librarybrowser.css
```diff
@@ -952,6 +952,10 @@ div.itemDetailGalleryLink.defaultCardBackground {
     flex-wrap: wrap;
 }
 
+.verticalSection {
+    animation: fadein 0.2s;
```
ferferga, line 956: I guess it already existed somewhere and you're just adding animations, however let me ask: where it's used?

Why not a *slow* 0.7 variation?

MrTimscampi: > Why not a slow 0.7 variation?

It's not supposed to be 0.2s, because that still looks like it pops in :smile: 

The goal was to make the sections fade-in as a way to add the egregious way that sections get added to pages.

src/assets/css/site.css
```diff
@@ -50,8 +50,12 @@ body {
...
 }
 
-.mainAnimatedPage {
-    contain: style size !important;
+.fadein {
+    animation: fadein 0.7s;
+}
+
+.fadein-fast {
+    animation: fadein 0.7s;
```
dmitrylyzo, line 58: `0.2s`?

dmitrylyzo:
There is no animation on `Back`. Or is it just me?

Issue #1455, opened 2020-05-16
Known Issue: v1.0.1 Media Notification Seekbar is broken


hamburglar2160:
The seekbar found in the expanded media notification for recent versions of Android is not hooked up to the app, so it doesn't do anything.

hamburglar2160:
Related reading on the Android side: https://developer.android.com/preview/features/media-controls

MrTimscampi:
If someone is interested in tackling this, here's some docs: https://developer.mozilla.org/en-US/docs/Web/API/MediaSession/setActionHandler

You need to hook up `seekto` in `mediasession.js` to seekPercent in playbackManager. Make sure to convert the time given by the Media Session API to a percentage, since that's what PlaybackManager expects.

KGT1:
@MrTimscampi would this be ok?
`        /* eslint-disable-next-line compat/compat */
        navigator.mediaSession.setActionHandler('seekto', function (object) {
            let item = playbackManager.getPlayerState(playbackManager.getCurrentPlayer()).NowPlayingItem;
            // Convert to ms
            let duration = parseInt(item.RunTimeTicks ? (item.RunTimeTicks / 10000) : 0);
            let wantedTime = object.seekTime*1000;
            playbackManager.seekPercent(wantedTime/duration*100,currentPlayer);
        });`

KGT1:
@MrTimscampi would something like this be ok?
```javascript
        /* eslint-disable-next-line compat/compat */
        navigator.mediaSession.setActionHandler('seekto', function (object) {
            let item = playbackManager.getPlayerState(playbackManager.getCurrentPlayer()).NowPlayingItem;
            // Convert to ms
            let duration = parseInt(item.RunTimeTicks ? (item.RunTimeTicks / 10000) : 0);
            let wantedTime = object.seekTime*1000;
            playbackManager.seekPercent(wantedTime/duration*100,currentPlayer);
        });
```

MrTimscampi:
> @MrTimscampi would something like this be ok?
> 
> ```js
>         /* eslint-disable-next-line compat/compat */
>         navigator.mediaSession.setActionHandler('seekto', function (object) {
>             let item = playbackManager.getPlayerState(playbackManager.getCurrentPlayer()).NowPlayingItem;
>             // Convert to ms
>             let duration = parseInt(item.RunTimeTicks ? (item.RunTimeTicks / 10000) : 0);
>             let wantedTime = object.seekTime*1000;
[...]

PR #1462, opened 2020-06-22
ActionHandler 'seekto' for Media Notification Seekbar

KGT1:
**Changes**
Adds 'seekto' ActionHandler mainly for new Android Devices, as they have a Media Notification Seekbar

**Issues**
#1455

### DIFF
--- a/src/components/playback/mediasession.js
+++ b/src/components/playback/mediasession.js
@@ -241,6 +241,15 @@ import connectionManager from 'connectionManager';
         navigator.mediaSession.setActionHandler('seekforward', function () {
             execute('fastForward');
         });
+
+        /* eslint-disable-next-line compat/compat */
+        navigator.mediaSession.setActionHandler('seekto', function (object) {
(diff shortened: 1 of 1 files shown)


src/components/playback/mediasession.js
```diff
@@ -241,6 +241,15 @@ import connectionManager from 'connectionManager';
         navigator.mediaSession.setActionHandler('seekforward', function () {
             execute('fastForward');
         });
+
+        /* eslint-disable-next-line compat/compat */
+        navigator.mediaSession.setActionHandler('seekto', function (object) {
+            let item = playbackManager.getPlayerState(playbackManager.getCurrentPlayer()).NowPlayingItem;
```
MrTimscampi, line 247: Why not use currentPlayer instead of playbackManager.getCurrentPlayer()?

KGT1: yes should be currentPlayer

PR #1464, opened 2020-06-22
Prevent removal of system plugins

neilsb:
**Changes**
Plugins now have a `CanUninstall` property (see jellyfin/jellyfin#3411).  When set to false the "Uninstall" option is not displayed on the Plugin Menu.

If the property is set to false, and the Plugin has no Configuration Page, then the Plugin Menu button is not displayed on the card.

Server PR:  jellyfin/jellyfin#3411

**Issues**
Refs jellyfin/jellyfin#2555

### DIFF
--- a/src/controllers/dashboard/plugins/installed.js
+++ b/src/controllers/dashboard/plugins/installed.js
@@ -120,7 +120,7 @@ define(['loading', 'libraryMenu', 'dom', 'globalize', 'cardStyle', 'emby-button'
             });
         }
 
-        if (removable == 'true') {
+        if (removable === 'true') {
             menuItems.push({
                 name: globalize.translate('ButtonUninstall'),
                 id: 'delete',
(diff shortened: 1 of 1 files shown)


src/controllers/dashboard/plugins/installed.js
```diff
@@ -37,7 +37,7 @@ define(['loading', 'libraryMenu', 'dom', 'globalize', 'cardStyle', 'emby-button'
         })[0];
         var configPageUrl = configPage ? Dashboard.getConfigurationPageUrl(configPage.Name) : null;
         var html = '';
-        html += "<div data-id='" + plugin.Id + "' data-name='" + plugin.Name + "' class='card backdropCard'>";
+        html += "<div data-id='" + plugin.Id + "' data-name='" + plugin.Name + "' data-removable='" + plugin.CanUninstall  + "' class='card backdropCard'>";
```
MrTimscampi, line 40:
```suggestion
        html += "<div data-id='" + plugin.Id + "' data-name='" + plugin.Name + "' data-removable='" + plugin.CanUninstall + "' class='card backdropCard'>";
```

src/controllers/dashboard/plugins/installed.js
```diff
@@ -115,11 +120,13 @@ define(['loading', 'libraryMenu', 'dom', 'globalize', 'cardStyle', 'emby-button'
             });
         }
 
-        menuItems.push({
-            name: globalize.translate('ButtonUninstall'),
-            id: 'delete',
-            icon: 'delete'
-        });
+        if (removable == 'true') {
```
JustAMan, line 123:
```suggestion
        if (removable === 'true') {
```
do we want to do so?

PR #1474, opened 2020-06-22
Fix number of items per row in home sections

MrTimscampi:
**Changes**

Previously, the limits for the number of items in home sections were kinda nonsensical and arbitrary.

This uses some cardBuilder utilities to get the number of cards that fit on one line of the screen and multiplies that by three, to get always a nice number of items in the home sections.

**Issues**

### DIFF
--- a/src/components/homesections/homesections.js
+++ b/src/components/homesections/homesections.js
@@ -22,7 +22,7 @@ define(['connectionManager', 'cardBuilder', 'appSettings', 'dom', 'apphost', 'la
         }
     }
 
-    function getSectionLimit(collectionType, shape) {
+    function getSectionLimit(collectionType, shape, multiplier = 3) {
         let screenWidth = dom.getWindowSize().innerWidth;
         const screenHeight = dom.getWindowSize().innerHeight;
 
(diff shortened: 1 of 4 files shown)


src/components/homesections/homesections.js
```diff
@@ -22,6 +22,28 @@ define(['connectionManager', 'cardBuilder', 'appSettings', 'dom', 'apphost', 'la
...
+        let screenWidth = dom.getWindowSize().innerWidth;
+        const screenHeight = dom.getWindowSize().innerHeight;
+
+        if (cardBuilder.isResizable(screenWidth)) {
+            const roundScreenTo = 100;
+            screenWidth = Math.floor(screenWidth / roundScreenTo) * roundScreenTo;
+        }
+
+        if (!shape) {
+            if (collectionType === 'movies' || collectionType === 'books' || collectionType === 'tvshows') {
```
hamburglar2160, line 35: collectionType or viewType === 'mixed' too as they're movies and tv shows!

MrTimscampi: Mixed Content do not have a collectionType, nor do they have a viewType.

Also, as far as I can tell, these libraries use the thumb shape in the latest section, not another shape.

In addition, Mixed Content libraries can contain any kind of content, not only TV or Movies.

PR #1475, opened 2020-06-22
remove sessionId parameter from api calls

crobibero:
https://github.com/jellyfin/jellyfin/pull/3378

I didn't go up the chain to remove unused parameters as I don't know JS very well.

### DIFF
--- a/src/components/syncPlay/groupSelectionMenu.js
+++ b/src/components/syncPlay/groupSelectionMenu.js
@@ -37,7 +37,7 @@ function showNewJoinGroupSelection (button, user, apiClient) {
         console.debug('No item is currently playing.');
     }
 
-    apiClient.sendSyncPlayCommand(sessionId, 'ListGroups').then(function (response) {
+    apiClient.sendSyncPlayCommand('ListGroups').then(function (response) {
         response.json().then(function (groups) {
             var menuItems = groups.map(function (group) {
                 return {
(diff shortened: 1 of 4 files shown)


src/components/syncPlay/groupSelectionMenu.js
```diff
@@ -37,7 +37,7 @@ function showNewJoinGroupSelection (button, user, apiClient) {
         console.debug('No item is currently playing.');
     }
 
-    apiClient.sendSyncPlayCommand(sessionId, 'ListGroups').then(function (response) {
+    apiClient.sendSyncPlayCommand('ListGroups').then(function (response) {
```
JustAMan, line 40: @MrTimscampi doesn't this need an update to apiclient?..

MrTimscampi: It's done already, the PR there (and this one) are waiting for the backend PR.

JustAMan: This PR should update the version of apiclient used.

MrTimscampi: We need to release that first, but that needs the backend PR to be merged first.

joshuaboniface: I believe this was addressed below.

PR #1484, opened 2020-06-24
Remove legacy loading

ferferga:
We really want to keep this?

Why these devices used that loader? I doubt something as simple as our loader should pose a performance degradation, so we should understand the reasoning behind *somehow*.


MrTimscampi:
We likely don't, but it needs testing on WebOS and Tizen before removal.

dkanada:
cc @dmitrylyzo

dmitrylyzo:
Tizen 4 (real): visually OK.
WebOS 1.2 (emulator): it works but has some artifacts (black holes over spinning lines) - maybe this is emulator's fault: it also deforms CD icon.

PR #1487, opened 2020-06-25
Tonemapping function dashboard options

nyanmisaka:
**Changes**
- Part1 -- https://github.com/jellyfin/jellyfin/pull/3442
- Part2 -- This

These PRs rely on the `tonemap_opencl` filter based on the hardware acceleration of the OpenCL device, can cooperate with the NVENC HEVC decoder to perform tone mapping from HDR to SDR while maintaining a decent transcoding speed. 

- Dashboard
dashboard


MrTimscampi:
@nyanmisaka Would it be feasible to (maybe in a further PR) add a VideoRange property to the client profile in order to only tonemap in cases where the client is SDR and the video HDR, but keep the HDR data if the client supports it?

nyanmisaka:
We want to avoid re-encoding the HEVC HDR10 video if client supports HDR. Our current target format is H264 8bit that does not support HDR info, but H264 10bit is a format that is not supported by hardware encoders.

Unless we re-encode the video to HEVC 10bit, it may be possible to adjust the parameters on a specific encoder to passthrough HDR info during the transcoding process.

Darkwolf1515:
Might be outside the scope of the request, but tonemapping could be a per client setting, so you would tell Jellyfin this specific client is HDR compatible, and it would avoid tonemapping then.

nyanmisaka:
> Might be outside the scope of the request, but tonemapping could be a per client setting, so you would tell Jellyfin this specific client is HDR compatible, and it would avoid tonemapping then.

The tonemap function in this PR is accompanied by transcoding. If it is a client that supports direct playback of HDR video, transcoding is not required.

Camc314 (approved):

blazarr:
I assume this is just waiting on another reviewer, the main server changes were already merged into master. Very excited for this to get merged in, can you tell? 😊

thornbill (approved):

nyanmisaka:
> I assume this is just waiting on another reviewer, the main server changes were already merged into master. Very excited for this to get merged in, can you tell? 😊

Just got merged. Keeping eyes on the upcoming unstable builds ;)

blazarr:
> > I assume this is just waiting on another reviewer, the main server changes were already merged into master. Very excited for this to get merged in, can you tell? 😊
> 
> Just got merged. Keeping eyes on the upcoming unstable builds ;)

Just tried it out, unfortunately the latest stable build seems to have broken transcoding detection. All of my clients keep reporting
`This client isn't compatible with the media and the server isn't sending a compatible media format.`

Womp womp, I will keep trying each new unstable build until I can test this out!

nyanmisaka:
> Just tried it out, unfortunately the latest unstable build seems to have broken transcoding detection. All of my clients keep reporting
> `This client isn't compatible with the media and the server isn't sending a compatible media format.`
> 
> Womp womp, I will keep trying each new unstable build until I can test this out!

It is caused by https://github.com/jellyfin/jellyfin/pull/4093 and https://github.com/jellyfin/jellyfin/pull/4069

blazarr:
> jellyfin/jellyfin#4093

Hah from first glance I would never have connected the issue to those PRs. Turned on debug logging and didn't really see anything useful. I'll patiently wait for these to get fixed. Thanks again!

PR #1488, opened 2020-06-26
Fix redirect issue on video player

dkanada:


src/controllers/playback/videoosd.js
```diff
@@ -1341,7 +1341,7 @@ define(['playbackManager', 'dom', 'inputManager', 'datetime', 'itemHelper', 'med
                 });
             } catch (e) {
                 require(['appRouter'], function(appRouter) {
-                    appRouter.showDirect('/');
+                    window.location.href = 'index.html';
```
MrTimscampi, line 1344: You can use `appRouter.goHome` to do this (It's defined in site.js), it's less hacky than overriding the href.

dkanada: I didn't want to mess with the horrendously long define statement at the top. We use this in a few other places, so I figured we could just replace them all at some point.

MrTimscampi: You're already requiring appRouter right above, though ;)

It's already in there at that point, you only have to call it.

dkanada: Well I feel stupid now :laughing: I was looking at the top for that dependency.

src/controllers/playback/videoosd.js
```diff
@@ -1341,7 +1341,7 @@ define(['playbackManager', 'dom', 'inputManager', 'datetime', 'itemHelper', 'med
                 });
             } catch (e) {
                 require(['appRouter'], function(appRouter) {
-                    appRouter.showDirect('/');
+                    appRouter.goHome()
```
MrTimscampi, line 1344:
```suggestion
                    appRouter.goHome();
```

PR #1489, opened 2020-06-26
Add a Netflix-like scroller section option

MrTimscampi:
**Changes**

Adds a Netflix-like scroller as a home section option which presents the latest 5 items server-wide.

Early implementation, design subject to change. I'm basically only opening this now so I don't forget I have it.

Screenshot_2020-06-26 Jellyfin
Screenshot_2020-06-26 Jellyfin(1)
Screenshot_2020-06-26 Jellyfin

**Issues**

### DIFF
--- a/src/assets/css/librarybrowser.css
+++ b/src/assets/css/librarybrowser.css
@@ -305,7 +305,7 @@
     }
 
     .pageWithAbsoluteTabs:not(.noSecondaryNavPage) {
-        padding-top: 6.7em !important;
+        padding-top: 4.4em !important;
     }
 
     .absolutePageTabContent {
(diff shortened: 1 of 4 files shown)


hamburglar2160:
Cool!
Consider doubling it to at least 10 items. 5 isn't all that much--HBO uses 12. Can we use the emby scroller through these (perhaps they appear only on mouse-over)? Include the most recent "Continue Watching" item in the mix?

On interfaces that have mouse inputs, perhaps do away with the buttons (and descriptions?) and have them appear on a nice blurhashed mouse-over :)

This would be a very good item to have customizable by the user. Being able to toggle on/off most of the descriptors to match what other streamers use would be very effective as for some (me) less might be more (HBO with that cue). For example, with the top screenshot, something contained within the top- or bottom-left corner of the banner:
"New <genre>
Tokyo Monogatari
<tagline>"

#1202
We also have a bunch of good UX references for this posted.

MrTimscampi:
> Consider doubling it to at least 10 items. 5 isn't all that much--HBO uses 12. Can we use the emby scroller through these (perhaps they appear only on mouse-over)? Include the most recent "Continue Watching" item in the mix?

The number will be configurable. As said, this is a very early POC.

It doesn't use the emby-scroller, mostly because that component relies on a severely out of date and unmaintained external dependency that I'd like to remove at some point (The last update for the library it uses is from 2015).  
It is using Swiper, the same component used for the photo player and the screensavers.

The slides change every 10 seconds for now (will be configurable as well) and can be manually changed using the bullets at the bottom. They don't slide but fade from one to the next, as a slide would look weird given the visual structure of these.

As for the content, it uses only one endpoint, to get the last few items added to the server. Imo, "continue watching" wouldn't make sense here, since the goal is to showcase items to users. Ideally, we'd hook it up to a suggestion endpoint, but we don't have a good one yet.

> On interfaces that have mouse inputs, perhaps do away with the buttons (and descriptions?) and have them appear on a nice blurhashed mouse-over :)

There is no blurhash on this, for now. The goal is to show info to "sell" the new content at glance to the user and act as a showcase for new items. Also we can't blurhash buttons or text, it's only for images, and blurhashing this wouldn't make sense.

> This would be a very good item to have customizable by the user. Being able to toggle on/off most of the descriptors to match what other streamers use would be very effective as for some (me) less might be more (HBO with that cue). For example, with the top screenshot, something contained within the top- or bottom-left corner of the banner:
> "New
> Tokyo Monogatari"

Eventually, my goal is to have every visual element of everything in Jellyfin be configurable in some way. However, this very much requires a complete redesign of the way we handle settings and present them to the user.  
Keep in mind that we do multiple passes when designing stuff like this. The details page is a good example of this. The version in 10.5 was early, has some visual inconsistencies and annoyances. Through using it more, we're bringing a few improvements to 10.6, as well as visual fixes and more options for how it behaves. I expect this way of doing things to continue in the future with all the redesigns and new features we introduce.

So see the result of this PR as a first pass, to be improved upon in the future.

> I'm guessing you're leaving music libraries out of this? I think that'd be wise.

It uses whatever the LatestItems endpoint returns, we don't really have any control on that aside from passing it library IDs, for now.

ri0thamus:
I love this idea.  The home page could use a visual change to separate it from the project that it originated from.

Darkwolf1515:
Defiantly a fantastic addition, I do have to ask about it using images for titles, did you manually add those? I don't think jellyfin downloads them as metadata, so im wondering if that will be added too.

MrTimscampi:
> Defiantly a fantastic addition, I do have to ask about it using images for titles, did you manually add those? I don't think jellyfin downloads them as metadata, so im wondering if that will be added too.

It will use the Logo instead of text for the title if there is one. They can be either added manually or automatically through the Fanart plugin.

Darkwolf1515:
> > Defiantly a fantastic addition, I do have to ask about it using images for titles, did you manually add those? I don't think jellyfin downloads them as metadata, so im wondering if that will be added too.
> 
> It will use the Logo instead of text for the title if there is one. They can be either added manually or automatically through the Fanart plugin.

Well that would explain where to get them from, thank you.

MominRaza:
This is a cool idea
Why this is still not merged or no progress?

PR #1492, opened 2020-06-28
Prevent viewcontainer from storing multiple pages

### LINKED ISSUE #1450: Navigation may result in duplicated moviesPage / indexPage in the DOM
**Describe the bug**
Navigating between pages seems to be causing unnecessary duplication in the DOM.

**To Reproduce**

1. Go to Home Page
2. Click on a library
3. Click on the Home button
4. Look at the DOM inside the `"mainAnimatedPages"`'s `div`: the `div` with id `indexPage` is duplicated

A similar problem occurs if you first load your server on the `movie.html` page, click on the Home button, click on a movie library: you will figure out duplicated `#moviesPage` in the DOM.

**Expected behavior**
No duplication of `div` with the same `id`.

**Screenshots**
Peek 2020-06-19 22-26

 - OS: Linux
 - Browser: Firefox 76
 - Jellyfin Version: `10.5.5` and `master`

**Additional context**
I think this is an easy fix for whoever understand the navigation process:
https://github.com/jellyfin/jellyfin-web/blob/bef977062e7cde96716efa23a98d06c4285e8f28/src/components/viewContainer.js#L71-L84
The `appendChild()` function is called thrice: on first load, on library click, and on go home. I think the "go home" button should maybe reset `allPages` array before creating the new child. :confused:

MrTimscampi:
**Changes**

For some reason, viewContainer is set to store 3 pages at a time and doesn't delete them when it doesn't need them anymore.
This sets the `pageContainerCount` to 1, in order to force it to remove old pages from DOM.

Obviously this is a temporary fix for a bigger issue (The issue being that viewContainer is stupid).

**Issues**

Fixes #1450

### DIFF
--- a/src/components/viewContainer.js
+++ b/src/components/viewContainer.js
@@ -247,7 +247,7 @@ define(['browser', 'dom', 'layoutManager', 'css!components/viewManager/viewConta
     var mainAnimatedPages = document.querySelector('.mainAnimatedPages');
     var allPages = [];
     var currentUrls = [];
-    var pageContainerCount = 3;
+    var pageContainerCount = 1;
     var selectedPageIndex = -1;
     reset();
     mainAnimatedPages.classList.remove('hide');


Delgan:
If old pages are entirely removed from the DOM, then the scroll position will be lost when clicking the "Go back" button. 

This can be frustrating for someone who is at the bottom of their movie list, clicks on a poster, then goes back and realizing he's returned to the top of the list.

dkanada:
Hmmm that's a really good reason to keep a few pages in the DOM during navigation.

MrTimscampi:
> Hmmm that's a really good reason to keep a few pages in the DOM during navigation.

Keeping only 1 also completely breaks saving display settings, for some reason.

The way it is implemented now, I think the duplicated index page is expected.

Basically, it has a "buffer" of 3 views. It will fill that buffer before replacing the oldest one with the new view.

I'll close this since it's "working" as expected and the proper fix would probably be to move to a real router instead doing this kind of shenanigans.

PR #1497, opened 2020-06-29
Add defered appLoader and splash screen

MrTimscampi:
**Changes**

Adds a `defer` tag to the injected appLoader script in order to load and parse `appLoader.js` before signaling `DOMContentLoaded`.

Adds an `async` tag to all the scripts inserted by appLoader. This allows the browser to fetch and parse the scripts at the same time, making them load fast.

Adds a nice splash screen with the Jellyfin logo during the bootstrap of the app.

*I get around 2s better loading with this PR than on current master (12s -> 10s before fully loaded)*

**Issues**

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -184,7 +184,12 @@ function copy(query) {
 function injectBundle() {
     return src(options.injectBundle.query, { base: './src/' })
         .pipe(inject(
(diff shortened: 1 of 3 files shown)


MrTimscampi:
> How did you get 12s?

My test server is remote and my internet may have been slow at the time. I went by Firefox's Network tab time.

PR #1498, opened 2020-06-29
Improve blurhash performance

MrTimscampi:
**Changes**

Consider this **very experimental**, but for me this makes blurhash load earlier and scrolling smoother.

**Issues**

### DIFF
--- a/src/components/images/imageLoader.js
+++ b/src/components/images/imageLoader.js
@@ -119,9 +119,9 @@ import 'css!./style';
     }
 
     export function lazyChildren(elem) {
-        for (const lazyElem of elem.getElementsByClassName('lazy')) {
-            if (userSettings.enableBlurhash()) {
-                var blurhashstr = lazyElem.getAttribute('data-blurhash');
+        if (userSettings.enableBlurhash()) {
+            for (const lazyElem of elem.getElementsByClassName('lazy')) {
+                const blurhashstr = lazyElem.getAttribute('data-blurhash');
                 if (!lazyElem.classList.contains('blurhashed', 'non-blurhashable') && blurhashstr) {
                     itemBlurhashing(lazyElem, blurhashstr);
                 } else if (!blurhashstr && !lazyElem.classList.contains('blurhashed')) {
--- a/src/components/images/imageLoader.js
+++ b/src/components/images/imageLoader.js
@@ -12,7 +12,7 @@ import 'css!./style';
         fillImageElement(elem, source);
     }
 
-    async function itemBlurhashing(target, blurhashstr) {
+    function itemBlurhashing(target, blurhashstr) {
         if (blurhash.isBlurhashValid(blurhashstr)) {
             // Although the default values recommended by Blurhash developers is 32x32, a size of 18x18 seems to be the sweet spot for us,
             // improving the performance and reducing the memory usage, while retaining almost full blur quality.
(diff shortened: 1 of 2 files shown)


src/components/images/imageLoader.js
```diff
@@ -132,15 +114,21 @@ import 'css!./style';
...
-        } else {
-            switchCanvas(elem);
-        }
+        elem.classList.remove('lazy-image-fadein-fast', 'lazy-image-fadein');
+        elem.classList.add('lazy-hidden');
     }
 
     export function lazyChildren(elem) {
+        for (const lazyElem of elem.getElementsByClassName('lazy')) {
+            if (userSettings.enableBlurhash()) {
```
ferferga, line 123: @MrTimscampi Why we don't do this check outside the ``for`` loop?

src/components/images/imageLoader.js
```diff
@@ -132,15 +114,21 @@ import 'css!./style';
...
+                var blurhashstr = lazyElem.getAttribute('data-blurhash');
```
ferferga, line 124: `
```suggestion
                let blurhashstr = lazyElem.getAttribute('data-blurhash');
```

src/components/images/imageLoader.js
MrTimscampi: Because I copypasted code around :smile:

src/components/images/imageLoader.js
```diff
@@ -132,15 +114,21 @@ import 'css!./style';
...
-            switchCanvas(elem);
-        }
+        elem.classList.remove('lazy-image-fadein-fast', 'lazy-image-fadein');
+        elem.classList.add('lazy-hidden');
     }
 
     export function lazyChildren(elem) {
+        if (userSettings.enableBlurhash()) {
+            for (const lazyElem of elem.getElementsByClassName('lazy')) {
+                const blurhashstr = lazyElem.getAttribute('data-blurhash');
```
dkanada, line 124: just `blurhash` seems fine

ferferga: @dkanada I think this might lead into confusion, as we also have the module Blurhash (that we use above for decoding)

dkanada: It should at least use camel casing then.

ferferga: @dkanada Fair enough :)

src/components/images/imageLoader.js
```diff
@@ -12,7 +12,7 @@ import 'css!./style';
         fillImageElement(elem, source);
     }
 
-    async function itemBlurhashing(target, blurhashstr) {
```
ferferga, line 15: Already asked you in Matrix, but I'll do it here again so it can be discussed publicly.

Why?

src/components/images/style.css
```diff
@@ -12,12 +12,22 @@
...
+}
+
 .lazy-blurhash-fadein-fast {
-    transition: opacity 0.2s;
+    animation: fadein 0.1s;
 }
 
 .lazy-blurhash-fadein {
-    transition: opacity 0.7s;
+    animation: fadein 0.4s;
```
ferferga, line 30: Why those values? Why not 0.2 and 0.7?

Imo, we should stick to an specific speed for each mode. In #1458 you're still using the old values of 0.7 and 0.2. Either those or 0.1 or 0.4, but not both combined crazily

src/components/images/imageLoader.js
MrTimscampi: It's useless in this case. Quoting MDN:

> The async and await keywords enable asynchronous, promise-based behavior to be written in a cleaner style, avoiding the need to explicitly configure promise chains.

This doesn't make use of Promises so it is not needed here (But I do want to clean up all the Promise.then() everywhere with async/await, at some point).

src/components/images/style.css
MrTimscampi: The normal ones feel too long, since the images on top also fades-in.

Animations in different places can have different length, we're not forced to stick to one all over the place. They ideally need to be adjusted individually for both settings, which is what was done here.

ferferga: @MrTimscampi Fair enough

PR #1499, opened 2020-06-29
Move tvlatest tab to suggestions section

grafixeyehero:
**Changes**
<details>
<summary>
After
</summary>

Annotation 2020-06-30 004916

Screenshot 2020-09-14 001159


grafixeyehero:
> 
> 
> @grafixeyehero I agree we should do this, but perhaps shows and movies should match the music ordering with suggestions in the first position. Either that or the music tabs should move suggestions to the second position, but I think the former is a better plan because the suggestions are kind of like a home view for each library.

@dkanada updated https://github.com/jellyfin/jellyfin-web/pull/1499/commits/71c44dccd6ab690bb720b47d01ec1c9b26bcf12d

dkanada (approved):

PR #1502, opened 2020-07-01
Pull themes and plugins from web config

dkanada:


src/scripts/settings/webSettings.js
```diff
@@ -18,11 +18,29 @@ function getDefaultConfig() {
     });
 }
 
-export function enableMultiServer() {
+export function getMultiServer() {
     return getConfig().then(config => {
         return config.multiserver;
     }).catch(error => {
-        console.log('cannot get web config:', error);
+        console.log('cannot get web config: ', error);
```
dkanada, line 25: Does JavaScript automatically insert the space here?

dmitrylyzo:
```
console.log('t:', window);
t: Window {parent: Window, opener: null, top: Window, length: 0, frames: Window, …}

console.log('t: ', window);
t:  Window {parent: Window, opener: null, top: Window, length: 0, frames: Window, …}
```

dkanada: Looks like it does, thanks.

src/scripts/autoThemes.js
```diff
@@ -0,0 +1,14 @@
+import * as userSettings from 'userSettings';
```
MrTimscampi, line 1: Not a fan of the name of this file. It could also probably be part of skinManager.

dkanada: I pulled this naming from `autoBackdrops` which I also hate. It was mainly to get a unirform name for those two weird files for now but we should probably remove them yeah.

PR #1503, opened 2020-07-01
Convert XMLHttpRequest to Fetch API

MrTimscampi:
**Changes**

This converts the instances of XMLHttpRequest in the code to tthe Fetch API.

The Fetch API is cleaner, faster and more readable than the old XMLHttpRequest.

Surface testing was done, but I might've missed stuff.

The various editors aren't covered in this since there are ES6 PRs for them currently opened.

**Issues**

### DIFF
--- a/src/components/packageManager.js
+++ b/src/components/packageManager.js
@@ -24,53 +24,43 @@ define(['appSettings', 'pluginManager'], function (appSettings, pluginManager) {
         appSettings.set(settingsKey, JSON.stringify(manifestUrls));
     }
 
-    function loadPackage(packageManager, url, throwError) {
+    async function loadPackage(packageManager, url, throwError) {
+        var originalUrl = url;
+        url += url.indexOf('?') === -1 ? '?' : '&';
+        url += 't=' + new Date().getTime();
 
-        return new Promise(function (resolve, reject) {
+        try {
+            const response = await fetch(url);
 
-            var xhr = new XMLHttpRequest();
-            var originalUrl = url;
-            url += url.indexOf('?') === -1 ? '?' : '&';
-            url += 't=' + new Date().getTime();
+            if (response.status < 400) {
+                var pkg = response.json();
+                pkg.url = originalUrl;
 
-            xhr.open('GET', url, true);
+                addPackage(packageManager, pkg);
 
-            var onError = function () {
-
-                if (throwError === true) {
-                    reject();
-                } else {
-                    removeUrl(originalUrl);
-                    resolve();
+                var plugins = pkg.plugins || [];
+                if (pkg.plugin) {
+                    plugins.push(pkg.plugin);
                 }
-            };
-
-            xhr.onload = function (e) {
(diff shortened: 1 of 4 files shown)


src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1017,31 +1017,22 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
...
-                    reject(e);
-                    decrementFetchQueue();
-                };
-
-                xhr.send();
-            });
+            try {
+                const response = await fetch(url);
+                decrementFetchQueue();
+                return await response.json;
```
grafixeyehero, line 1031: missing bracket

src/components/packageManager.js
```diff
@@ -24,53 +24,43 @@ define(['appSettings', 'pluginManager'], function (appSettings, pluginManager) {
...
-
-            xhr.onerror = onError;
-
-            xhr.send();
-        });
+                var promises = plugins.map(function (pluginUrl) {
+                    return pluginManager.loadPlugin(packageManager.mapPath(pkg, pluginUrl));
+                });
+                Promise.all(promises).then(() => {
+                    return;
```
thornbill, line 49: I don’t think these returns do what you think they do. 😉

src/components/require/requiretext.js
```diff
@@ -26,14 +26,10 @@ define(function () {
                 url += 'r=0';
             }
 
-            var xhr = new XMLHttpRequest();
-            xhr.open('GET', url, true);
+            const response = await fetch(url);
```
thornbill, line 29: Any call to fetch needs to check the response status.

src/scripts/globalize.js
```diff
@@ -164,22 +163,17 @@ define(['userSettings', 'events'], function (userSettings, events) {
...
-
-            xhr.onerror = function () {
-                resolve({});
-            };
-            xhr.send();
-        });
+            const response = await fetch(url);
+            const jsonConntent = await response.json();
+
+            if (response.status < 400) {
```
thornbill, line 169: Need to check this before parsing as json or it can throw an exception.

dmitrylyzo:
Hmm, then we have 2 more in that big ES6 PR.
https://github.com/jellyfin/jellyfin-web/blob/4709e76b5a642c60c92bdf1ea2235a2483b463ff/src/components/libraryoptionseditor/libraryoptionseditor.js#L367-L368
https://github.com/jellyfin/jellyfin-web/blob/4709e76b5a642c60c92bdf1ea2235a2483b463ff/src/components/imageOptionsEditor/imageOptionsEditor.js#L93-L94

MrTimscampi:
This seems to have been mostly taken care of during the move to webpack or unrelated changes, since all I can find anymore is a reference in the Packages and a hack around Fetch not supporting the `file:` protocol.

So I think this is safe to close without really losing anything.

PR #1504, opened 2020-07-01
Prevent text selection

MrTimscampi:
**Changes**

Prevent highlighting of text by the user in all the client (Except inputs, of course).

Makes the client feel a bit more "app-like".

**Issues**

### DIFF
--- a/src/assets/css/site.css
+++ b/src/assets/css/site.css
@@ -3,6 +3,7 @@ html {
     margin: 0;
     padding: 0;
     height: 100%;
+    user-select: none;
 }
 
 .clipForScreenReader {
@@ -18,7 +19,7 @@ html {
 
 .material-icons {
     /* Fix font ligatures on older WebOS versions */
-    -webkit-font-feature-settings: "liga";
+    font-feature-settings: "liga";
 }
 
 .backgroundContainer {
@@ -37,10 +38,6 @@ html {
 .layout-mobile,
 .layout-tv {
     -webkit-touch-callout: none;
-    -webkit-user-select: none;
-    -khtml-user-select: none;
-    -moz-user-select: none;
-    -ms-user-select: none;
     user-select: none;
 }
 


dmitrylyzo:
Maybe I hurried. Do we need the ability to select text in Dashboard?

DMouse10462:
There are many use cases for selecting text within the UI, especially for foreign titles. Here are some example scenarios:

- I want to search for "君の名は。" (Your Name, the 2016 animated film) in its original language. I'd much rather select and copy it from my library than having to search for "Your Name" (or "Kimi no na wa." if I happen to know the transliteration) then find a suitable result with the original script and copy it from there. This applies to albums, song titles, people names, and a variety of other things I have in my library not in a writing system included on my keyboard or otherwise inconvenient to retype.

- Diagnostic information found within the client, such as "Media Info" fields such as the path or other details I may want to copy somewhere such as a terminal. This is especially useful when Jellyfin misidentifies something and I want to quickly go to the path of the offending file(s). I can just open the Media Info dialog, select a portion of the path, and paste it into a `cd` command to take me to the files so that I can verify/rename/reorganize them as needed.

- As a response to @dmitrylyzo and Dashboard selection, the API keys are not inputs and definitely need to be selected for copying (unless you enjoy typing long hexadecimal strings by hand).

Some of this may seem obscure/low-use, but this seems like functionality being removed for the sake of "cleanliness" despite the workflow advantages the feature may provide. I'm surprised this didn't have more discussion prior to merging.

At the very least, can this be TV (and *maybe* mobile, if it really feels necessary) only? Most of these scenarios take place on desktop for me anyways, but I can see myself doing foreign-language searches on my phone just as easily.

telans:
I agree. Can reverting this be considered?

DMouse10462:
I don't want to do a revert PR just yet without hearing from @MrTimscampi first, partly to get some insight on this decision (I pinged him in our Matrix chat as well) and partly because there appears to be another change (removing a prefix from a font property) that I don't want to just blanket revert without consulting him or another core dev about first as I'm not too familiar with WebOS, the platform mentioned in the comment associated with the property modified.

MrTimscampi:
The rationale is to make it act less like a website and more like a proper client. Netflix does this and so does Spotify, so it's far from unheard of.

However, this does go a bit overboard and introduces issues when copying API keys, as pointed out.

> I want to search for "君の名は。" (Your Name, the 2016 animated film) in its original language. I'd much rather select and copy it from my library than having to search for "Your Name" (or "Kimi no na wa." if I happen to know the transliteration) then find a suitable result with the original script and copy it from there. This applies to albums, song titles, people names, and a variety of other things I have in my library not in a writing system included on my keyboard or otherwise inconvenient to retype.

I fail to see a use for this. It seems like a very specific thing that very few people will use. Most will simply use the links to various databases we have on the info page or will search for the title themselves.  

> Diagnostic information found within the client, such as "Media Info" fields such as the path or other details I may want to copy somewhere such as a terminal. This is especially useful when Jellyfin misidentifies something and I want to quickly go to the path of the offending file(s). I can just open the Media Info dialog, select a portion of the path, and paste it into a cd command to take me to the files so that I can verify/rename/reorganize them as needed.

This I can see a bit more, but I still think most people wouldn't need it.

> As a response to @dmitrylyzo and Dashboard selection, the API keys are not inputs and definitely need to be selected for copying (unless you enjoy typing long hexadecimal strings by hand).

This I fully agree with and didn't think to put in an override for API keys. Could be done in an upcoming PR.

Overall, I don't think this is "removing features", as you say. It may be a bit too eager on where it prevents selections, but that can be easily fixed.

Do keep in mind, as well, that you can still override this with custom CSS.

That said, @jellyfin/web, what's your opinion on this?

dkanada:
> This I fully agree with and didn't think to put in an override for API keys.

The best method would be a copy button next to the delete button or something, that's a very common thing to have.

MrTimscampi:
> > This I fully agree with and didn't think to put in an override for API keys.
> 
> The best method would be a copy button next to the delete button or something, that's a very common thing to have.

Yeah, we already have some code for that as well, for URLs.

DMouse10462:
> The rationale is to make it act less like a website and more like a proper client. Netflix does this and so does Spotify, so it's far from unheard of.

Just had a look at Netflix's interface and can indeed see that it's the case. Thanks for the examples. I can attest that many of the same use cases have been a hindrance in Spotify, especially when it comes to the transliteration example and anime OP/ED songs I discover through the service. How do I even begin to search for something I can't type or transliterate myself? Sure, that use case is niche (I'm not arguing that everyone does this all the time). But, and I'll touch on this later, I feel that this change removes an entire paradigm from the traditional desktop user interface guidelines outside of, as your changes in #1513 will create, approved zones that may not be an exhaustive list of the user's expectations. 

> However, this does go a bit overboard and introduces issues when copying API keys, as pointed out.
> 
> > I want to search for "君の名は。" .....(snip)
> 
> I fail to see a use for this. It seems like a very specific thing that very few people will use. Most will simply use the links to various databases we have on the info page or will search for the title themselves.

The keyboard-mouse interface of a desktop, as opposed to the touch and remote interfaces of mobile and TV, make text selection and manipulation a first-class paradigm on the platform. Such behavior has been standard for decades, and applications that don't support traditional platform UI guidelines tend to frustrate users and create barriers in the interface. 

That's exactly why desktop apps don't work well on mobile or TV without a lot of rework to support touch, gestures, display layouts, remote navigation, and other things that users come to expect on the platform. On mobile, disabling text selection is probably a wise decision after all, since it's generally more jarring and confusing to an end-user when they select something and a pop-up context bubble gets in the way of their task.

I completely understand the frustration of accidentally dragging through a list of cards and having my whole browser viewport go blue on me. I just feel that all of the text in front of me on a desktop (except for buttons and other form-style UI elements) should be capable of being selected so that I can do what I wish with it. 

(As a side note: If every browser has the F7 "Caret Browsing" mode that IE has, this wouldn't be an issue at all for me because that's an easy, browser-wide override. I never thought I'd vouch for an IE-only feature.)

> > Diagnostic information found within the client, such as "Media Info" ...(snip)
> 
> This I can see a bit more, but I still think most people wouldn't need it.

I could say that most average users (non-administrators) wouldn't need the media info at all, but the people who do certainly would benefit from selecting text. Thanks for the PR addressing this.

> > As a response to @dmitrylyzo and Dashboard selection, the API keys are not inputs and definitely need to be selected for copying (unless you enjoy typing long hexadecimal strings by hand).
> 
> This I fully agree with and didn't think to put in an override for API keys. Could be done in an upcoming PR.
> 
> Overall, I don't think this is "removing features", as you say. It may be a bit too eager on where it prevents selections, but that can be easily fixed.

"Removing features" may have been a bit outside of my intentions. Maybe "against desktop-class HIG" or "breaking workflows", especially in the context of administration (during which I spend more time in the end-user interfaces reviewing the libraries than in the Dashboard). The approach just feels too heavy-handed and I feel goes against the anti-walled-garden approach of open source software and approaches something that's designed for locked-down content (like the Netflix and Spotify examples, which definitely *are* examples of walled-off garden DRM aproaches). Disclaimer though, I super appreciate all the work you put towards this project and in no way do I feel that's what you're trying to accomplish here. I also absolutely love native-feeling mobile and SPA design and feel that you've played a large part in helping make that happen here.

> Do keep in mind, as well, that you can still override this with custom CSS.
> 
> That said, @jellyfin/web, what's your opinion on this?

DMouse10462:
Sorry for the double-post, but I wanted to make sure I actually proposed a course of action rather than just remarking on work already done:

I propose either:

- Reverting the change entirely for *desktop* (easiest)

- Ensuring that all header and body text that is not solely for navigation is selectable while keeping the property in place overall. (Harder, but solves the odd selection behaviors in the UI while allowing the freedom to select information. This is actually a very standard feeling in native applications on desktop.)

PR #1510, opened 2020-07-03
Fix back button not stopping video playback

MrTimscampi:
**Changes**

**Issues**

### DIFF
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -3710,7 +3710,9 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
         return textStreamUrl;
     };
 
-    PlaybackManager.prototype.stop = function (player = this._currentPlayer) {
+    PlaybackManager.prototype.stop = function (player) {
+        player = player || this._currentPlayer;
+
         if (player) {
 
             if (enableLocalPlaylistManagement(player)) {


src/components/playback/playbackmanager.js
```diff
@@ -3710,7 +3710,9 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
         return textStreamUrl;
     };
 
-    PlaybackManager.prototype.stop = function (player = this._currentPlayer) {
+    PlaybackManager.prototype.stop = function (player) {
```
ferferga, line 3713: Shouldn't we pass the player object from the video player when doing stop, so we override the one from the PlaybackManager instance? I can't check the code of the video player bc I'm in mobile but this doesn't seem the best solution.

MrTimscampi: As explained on Matrix, this aims to be a "safe" fix by reverting the code to its previous state.

Ideally, we'd refactor the mess around  this to work with your change, but I don't want to accidentally not cover a case this close to release. So, imo, reverting the change is the safest option here. We can revisit this when migrating PlaybackManager to ES6.

ferferga: @MrTimsampi Good point. Also, as @dmitrlyzo confirmed that stop works on mobile, this seems good to go.

PR #1511, opened 2020-07-03
Use builtins when injecting core-js

MrTimscampi:
**Changes**

This should slightly reduce the amount of core-js polyfills inserted into bundle.js, by only inserting what we actually need.

**Issues**

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,14 +85,14 @@
   },
   "babel": {
     "presets": [
-    [
-      "@babel/preset-env",
-      {
-        "useBuiltIns": "entry",
-        "corejs": "3"
-      }
-    ]
-  ],
+      [
+        "@babel/preset-env",
+        {
+            "useBuiltIns": "entry",
+            "corejs": "3"
+        }
+      ]
+    ],
     "overrides": [
       {
         "test": [
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "query-string": "^6.13.1",
     "resize-observer-polyfill": "^1.5.1",
     "screenfull": "^5.0.2",
-    "shaka-player": "^3.0.1",
+    "shaka-player": "^2.5.13",
     "sortablejs": "^1.10.2",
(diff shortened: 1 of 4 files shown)


dmitrylyzo:
Now new Shaka doesn't break webOS 1.2, but I got error:
`nowPlayingPositionSlider.enableKeyboardDragging` is `undefined`
https://github.com/jellyfin/jellyfin-web/blob/0849ea77d8f17a9e4db9449b1cf62d9bb3e90604/src/controllers/playback/videoosd.js#L1353
Does `CustomElements` work differently now?

dmitrylyzo:
I tried to move `babel/polyfill` to the very top on `master` and got the same error.
Looks like we need to load `webcomponents` before `core-js`.

MrTimscampi:
> Looks like we need to load `webcomponents` before `core-js`.

Afaik, core-js's injection always goes at the top of the file. However, the polyfill for Web Components v0 (the one we use) is `document-register-element`. `webcomponents.js` is for the v1 API, which we want to migrate to at some point, probably.

dmitrylyzo:
I have moved `shaka-player` in `bundle.js` after `babel/polyfill` in `master` branch - **webOS 1.2 is happy**.
Loading order:
```
document-register-element -> adds document.registerElement
webcomponents -> changes document.registerElement
babel/polyfill -> adds Symbol
shaka-player
```

**Some thoughts**
- We have several polyfills for one API.
- Not all polyfills are the same.
- Order in `bundle.js` matters because our code is written for a specific API/polyfill implementation.
- There is a chance that polyfills in our `bundle.js` may break external modules.

dkanada:
@dmitrylyzo is the current state of this PR causing issues with any clients?

MrTimscampi:
> @dmitrylyzo is the current state of this PR causing issues with any clients?

We're removing Shaka, so it should be good. But it'll need some rework once Shaka is removed, so better wait for the dead code removal first.

PR #1516, opened 2020-07-04
Update deps

MrTimscampi:
**Changes**

Manual round of dependency updates + revert the Shaka Player update that breaks Tizen.

**Issues**


src/scripts/site.js
```diff
@@ -672,7 +670,7 @@ var AppInfo = {};
             },
             bundles: {
                 bundle: [
-                    'document-register-element',
+                    //'document-register-element',
```
dmitrylyzo, line 673: leftover

dmitrylyzo:
Un-exclude `xmldom` from babel - it uses back-quotes.

MrTimscampi:
> webOS 1.2 - OK

Awesome! I'm worried that using the proper injection for core-js will break components.js in the future, but we'll get to it when we get to it. For now, this is good enough for 10.6.

> I found `const` in `bundle.js` - `libass-wasm`. But it doesn't break parsing confused

Hm, weird.

PR #1533, opened 2020-07-07
Smaller library tiles

Nickbert7:
If you have a lot of libraries you have to scroll through all cards via the small arrows to get to the one in the back.
All libraries should be accessible with one click

**Changes**
Switch the library tiles to small cards to see all of them on the homescreen:

Old:
image

New:
image


MrTimscampi:
That should definitely be put behind an option.

Also, please keep the "enableScrollX" thing, as we plan to make that a configuration option sometime soon (Likely as part of #1474)

Nickbert7:
Ah, I have missed this open PR.
Then for now I will close my PR and wait until your PR is done.

The enableScrollX thing is not such important for me.
I primarily want the little tiles as in general everything on the homescreen is to large for me, I like to see more at once...

PR #1542, opened 2020-07-09
Fix OSD lock

dmitrylyzo:
**Changes**
- Replace the lock counter with the flag.
- Keep OSD visible when dialog is open.

**Issues**
_from Matrix_ Sometimes OSD isn't hiding.
In Firefox, dragging a poster on `videoosd` page breaks the lock counter because `mouseup` event doesn't fire.
In Chrome, it is hard to reproduce - rapid clicking + drag&drop.

Known issue:
If you start dragging and cancel it with `Escape`, holding mouse button, OSD will be hidden anyway.
Event sequence: `mousedown -> dragstart -> dragend -> keyup -> [hold mouse button] -> mouseup`


dmitrylyzo:
Before
osd-menu-1
After
osd-menu-2

PR #1543, opened 2020-07-09
Blur backgrounds and add background to Now Playing page

MrTimscampi:
**Changes**

Blur. Blur everywhere.

Except on the details page.

image

**Issues**

### DIFF
--- a/src/components/remotecontrol/remotecontrol.js
+++ b/src/components/remotecontrol/remotecontrol.js
@@ -207,7 +207,6 @@ define(['dom', 'datetime', 'backdrop', 'libraryBrowser', 'listView', 'imageLoade
             });
             setImageUrl(context, state, url);
             if (item) {
-                console.warn(item);
                 if (item.Type === 'Audio') {
                     var imgUrl = apiClient.getScaledImageUrl(item.Id, {
                         type: 'Primary',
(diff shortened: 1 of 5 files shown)


src/components/remotecontrol/remotecontrol.js
```diff
@@ -207,7 +207,18 @@ define(['browser', 'datetime', 'backdrop', 'libraryBrowser', 'listView', 'imageL
             });
             setImageUrl(context, state, url);
             if (item) {
-                backdrop.setBackdrops([item]);
+                if (item.Type === 'Audio') {
+                    var imgUrl = apiClient.getScaledImageUrl(item.Id, {
+                        type: 'Primary',
+                        maxWidth: dom.getScreenWidth(),
+                        index: 0,
+                        tag: item.AlbumPrimaryImageTag[0]
```
dmitrylyzo, line 215: This doesn't work for me. No image: 404 error.
It generates URL for image of `item.Id`, but should be `item.AlbumId`, if one exists.
Looks like `AlbumPrimaryImageTag` might just be a string. And in this case, `index` shouldn't be specified.

MrTimscampi: > folder.jpg

If @dkanada can confirm, I don't think we get these right now? Or there were some weird stuff with them?

dmitrylyzo: > > folder.jpg
> 
> If @dkanada can confirm, I don't think we get these right now? Or there were some weird stuff with them?

I have them already in metadata.
Probably this https://github.com/jellyfin/jellyfin/blob/e3db0ac400e7a1a42e7a844a13ab052d26125612/MediaBrowser.LocalMetadata/Images/LocalImageProvider.cs#L29-L35

dkanada: LocalImageProvider is somehow different from the method a lot of the images use, it's quite annoying. Not sure how that would affect the web client though.

PR #1551, opened 2020-07-11
Fix some style issues on item details and now playing

dkanada:


src/nowplaying.html
```diff
@@ -15,35 +15,35 @@ <h2 class="nowPlayingPageTitle"></h2>
...
                             <span class="material-icons replay_10"></span>
                         </button>
-                        
+
                         <button is="paper-icon-button-light" class="btnPreviousTrack btnPlayStateCommand autoSize" title="${ButtonPreviousTrack}">
                             <span class="material-icons skip_previous"></span>
                         </button>
 
-                        <button is="paper-icon-button-light" class="btnPlayPause btnPlayStateCommand autoSize" title="${ButtonPause}">
+                        <button is="paper-icon-button-light" class="btnPlayPause btnPlayStateCommand autoSize cardOverlayButtonIcon" title="${ButtonPause}">
```
dmitrylyzo, line 46: Now `Play/Pause` is not lined up.
Before
play-old
After
play-new

src/themes/appletv/theme.css
```diff
@@ -449,12 +449,8 @@ html {
...
 }
 
-.repeatButton-active {
-    color: #4285f4;
-}
-
-.shuffleQueue-active {
-    color: #4285f4 !important;
+.buttonActive {
+    color: #00a4dc;
```
dmitrylyzo, line 453: `!important` is required to override hover state, iirc
But it might be better to fix hover state.

dmitrylyzo:
More to line up.
`Favorite` button at `nowPlayingBar` is 1px up: `listItemButton` or more precisely `display: inline-block` does it.

src/nowplaying.html
dkanada: It looks even to me when testing locally, do you have a fix I could apply for your screen?

dmitrylyzo: I have removed `cardOverlayButtonIcon` and got `Before` one.
Desktop layout also has uneven `Play/Pause` button. But it is bigger without `cardOverlayButtonIcon`.
Ah, this is Chrome vs Firefox issue. Chrome sucks here.

dkanada, line 46:
```suggestion
                        <button is="paper-icon-button-light" class="btnPlayPause btnPlayStateCommand autoSize" title="${ButtonPause}">
```

PR #1573, opened 2020-07-16
Minor fixes

### LINKED ISSUE #1565: Clicking on Song Card begins playback
When browsing the search results for an individual song, clicking anywhere on the displayed cards plays the song.

As there is no song "item details" page, I would expect that clicking on the card (i.e. anywhere that isn't the play button) would either 1. do nothing or 2. bring me to the album that contains the song because the card displays its cover art.

For context, clicking outside of the play button on any other type of card would bring me to its corresponding details page.

MrTimscampi:
**Changes**

* Remove play action on image in listView
* Remove play action on card from search results when it's a song
* Doubles the size of the main item image

**Issues**

Fixes #1565

### DIFF
--- a/src/components/appRouter.js
+++ b/src/components/appRouter.js
@@ -227,7 +227,7 @@ define(['loading', 'globalize', 'events', 'viewManager', 'skinManager', 'backdro
             setQuality = true;
         }
 
-        if (setQuality || !options.quality) {
+        if (setQuality && !options.quality) {
             options.quality = 90;
         }
     }
--- a/src/components/appRouter.js
+++ b/src/components/appRouter.js
@@ -222,46 +222,13 @@ define(['loading', 'globalize', 'events', 'viewManager', 'skinManager', 'backdro
     }
 
     function normalizeImageOptions(options) {
-        var scaleFactor = browser.tv ? 0.8 : 1;
-
         var setQuality;
-        if (options.maxWidth) {
-            options.maxWidth = Math.round(options.maxWidth * scaleFactor);
+        if (options.maxWidth || options.width || options.maxHeight || options.height) {
             setQuality = true;
         }
 
-        if (options.width) {
-            options.width = Math.round(options.width * scaleFactor);
-            setQuality = true;
-        }
-
-        if (options.maxHeight) {
-            options.maxHeight = Math.round(options.maxHeight * scaleFactor);
-            setQuality = true;
-        }
-
-        if (options.height) {
-            options.height = Math.round(options.height * scaleFactor);
-            setQuality = true;
-        }
-
-        if (setQuality) {
-            var quality;
-            var type = options.type || 'Primary';
-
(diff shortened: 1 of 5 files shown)


src/components/search/searchresults.js
```diff
@@ -464,7 +464,6 @@ define(['layoutManager', 'globalize', 'require', 'events', 'connectionManager',
             showTitle: true,
             overlayText: false,
             centerText: true,
-            action: 'play'
```
dmitrylyzo, line 467: Add `overlayPlayButton: true` to add play button for mobile.
TV will still go to itemDetails, but this is OK for now.

src/controllers/itemDetails/index.js
```diff
@@ -726,7 +726,7 @@ define(['loading', 'appRouter', 'layoutManager', 'connectionManager', 'userSetti
             disableIndicators: true,
             disableHoverMenu: true,
             overlayPlayButton: true,
-            width: dom.getWindowSize().innerWidth * 0.25
+            width: dom.getWindowSize().innerWidth * 0.5
```
dmitrylyzo, line 729: How about that `scaleFactor`?

dmitrylyzo:
I think we should change action to `link` here (for TV) https://github.com/jellyfin/jellyfin-web/blob/af11f8b3b7f81fb1f00961e9c6508af4d890f60f/src/controllers/itemDetails/index.js#L1364
That's was me who added 'resume' here to be able to play selected episode. But in this case, you can't reach item details of episode easily.

src/components/appRouter.js
```diff
@@ -222,46 +222,13 @@ define(['loading', 'globalize', 'events', 'viewManager', 'skinManager', 'backdro
...
-                    quality = type === 'Primary' ? 40 : 50;
-                } else {
-                    quality = type === 'Backdrop' ? 60 : 50;
-                }
-            } else {
-                quality = type === 'Backdrop' ? 70 : 90;
-            }
-
-            options.quality = quality;
+        if (setQuality || !options.quality) {
```
dmitrylyzo, line 230: I think that we shouldn't set `quality` if it is not required or already set.
```suggestion
        if (setQuality && !options.quality) {
```

src/controllers/itemDetails/index.js
```diff
@@ -1361,7 +1361,7 @@ define(['loading', 'appRouter', 'layoutManager', 'connectionManager', 'userSetti
                         imageSize: 'large',
                         enableSideMediaInfo: false,
                         highlight: false,
-                        action: layoutManager.tv ? 'resume' : 'none',
+                        action: layoutManager.tv ? 'link' : 'none',
```
dmitrylyzo, line 1364: Even better: you can go to item details on mobile by clicking on a row.
```suggestion
                        action: !layoutManager.desktop ? 'link' : 'none',
```

PR #1594, opened 2020-07-19
Redesign login and selectserver pages

MrTimscampi:
**Changes**

A small redesign to the selectserver and login pages.

Screenshot_2020-07-19 Jellyfin(2)

Screenshot_2020-07-19 Jellyfin(3)

Screenshot_2020-07-19 Jellyfin

Screenshot_2020-07-19 Jellyfin(1)

**Issues**

### DIFF
--- a/src/assets/css/site.css
+++ b/src/assets/css/site.css
@@ -129,3 +129,7 @@ div[data-role=page] {
 .hide-scroll {
     overflow-y: hidden;
 }
+
+.w-100 {
+    width: 100%;
+}
--- a/src/controllers/auth/login.js
+++ b/src/controllers/auth/login.js
@@ -1,4 +1,4 @@
-define(['apphost', 'appSettings', 'dom', 'connectionManager', 'loading', 'layoutManager', 'browser', 'globalize', 'cardStyle', 'emby-checkbox'], function (appHost, appSettings, dom, connectionManager, loading, layoutManager, browser, globalize) {
+define(['apphost', 'appSettings', 'dom', 'connectionManager', 'loading', 'layoutManager', 'libraryMenu', 'browser', 'globalize', 'cardStyle', 'emby-checkbox'], function (appHost, appSettings, dom, connectionManager, loading, layoutManager, libraryMenu, browser, globalize) {
     'use strict';
 
     var enableFocusTransform = !browser.slow && !browser.edge;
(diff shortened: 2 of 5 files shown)


dkanada:
Might be prudent to switch the strings to title casing while we're messing with the code here.

src/assets/css/site.css
```diff
@@ -129,3 +129,7 @@ div[data-role=page] {
 .hide-scroll {
     overflow-y: hidden;
 }
+
+.w-100 {
```
masterkoppa, line 133: Out of curiosity, would introducing something like tailwind help with creating these standard utility classes?

https://tailwindcss.com/

dkanada: We'll be switching to Vue at some point, but I suppose we haven't decided on the rest of the stack for components and CSS utilities. So in theory, yes that's an option, but our biggest priority is popularity within the industry so new contributors are more likely to know the tools.

MrTimscampi: To add a bit, yes, we should likely use a SASS utility package for this kind of stuff. We haven't really looked into it, but, as @dkanada mentioned, we have a few (flexible) conditions when bringing in new dependencies:

* The project is alive and supported (Ideally by more than one individual or company, though that's not a hard requirement)
* The project is popular (We're not looking for the most popular, but if something has only a couple hundred pulls on NPM, we're a bit wary. We did go with less popular alternatives in Vue and date-fns, over React and MomentJS respectively, though)
* The project is standards compliant and, if UI/UX-related, supports a11y, l10n/i18n and can be used with right-to-left script and non-western script (CJK, Hebrew, Arabic, Devanagari, etc)
* It's modern (This is very flexible, but ideally we don't want to base the project around "dying" technology)
* It fits our requirements

I'll need to have a look at what's available for minimal SASS frameworks, though. But I sort of feel like handling this after the ES6 migration is done will be easier than now.

sparky8251: tailwindcss is common enough even I see it and I really only do backend stuff. Seen it in several startup projects for fully Rust/wasm frontend applications.

I dont know much about its features, but if we are talking popularity, support, and broad use I'd say its there at least.

MrTimscampi: Tailwind seems good, indeed.

We'll likely take care of it once we're done with ES6, as that will make hoisting dependencies a lot easier...

PR #1595, opened 2020-07-19
Fix issue with sync menu and excessive logging

dkanada:
I'd like to get this in 10.6.0 so if someone sees this feel free to review. My chat may or may not be totally broken right now so I'll be gone from Matrix until Tuesday or Wednesday.

### DIFF
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -72,7 +72,6 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
         const now = (new Date).getTime();
 
         if (method !== reportPlaybackLastMethod || now - (reportPlaybackLastTime || 0) >= reportPlaybackLogDelay) {
-            console.debug(method + '-' + JSON.stringify(info));
             reportPlaybackLastMethod = method;
             reportPlaybackLastTime = now;
         }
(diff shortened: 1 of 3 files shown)


src/components/playback/playbackmanager.js
```diff
@@ -72,7 +72,6 @@ define(['events', 'datetime', 'appSettings', 'itemHelper', 'pluginManager', 'pla
         const now = (new Date).getTime();
 
         if (method !== reportPlaybackLastMethod || now - (reportPlaybackLastTime || 0) >= reportPlaybackLogDelay) {
-            console.debug(method + '-' + JSON.stringify(info));
```
dmitrylyzo, line 75: This block is for logging. It was more verbose :p
https://github.com/jellyfin/jellyfin-web/commit/bffafa69143b110f4a550c66642a4497366fad5b

dkanada: Ah, well I still think it's way too verbose for the debug logs :sweat_smile: might just want to remove it.

dmitrylyzo:
These also had to be removed:
https://github.com/jellyfin/jellyfin-web/blob/a8a85df303c88aa5d07ca06734ecf9d2404960f7/src/components/playback/playbackmanager.js#L46-L50
https://github.com/jellyfin/jellyfin-web/blob/a8a85df303c88aa5d07ca06734ecf9d2404960f7/src/components/playback/playbackmanager.js#L72-L77

PR #1599, opened 2020-07-19
Remove dead code

MrTimscampi:
**Changes**

Removes a ton of dead code and fixes all the `no-unused-vars` errors.

**Issues**


src/components/dialog/dialog.js
```diff
@@ -34,9 +34,9 @@ define(['dialogHelper', 'dom', 'layoutManager', 'scrollHelper', 'globalize', 're
             dlg.classList.add('dialog-fullscreen-lowres');
         }
 
-        //dlg.querySelector('.btnCancel').addEventListener('click', function (e) {
-        //    dialogHelper.close(dlg);
-        //});
+        dlg.querySelector('.btnCancel').addEventListener('click', function (e) {
```
ferferga, line 37: Shouldn't we remove ``e`` variables when they're not used everywhere as well?

MrTimscampi: Yes, but that's for another pass of ESLint :)

For this, I mainly used the 2 rules I added and found some other stuff along the way.

src/scripts/site.js
```diff
@@ -14,7 +14,7 @@ function getWindowLocationSearch(win) {
     return search || '';
 }
 
-function getParameterByName(name, url) {
+window.getParameterByName = function (name, url) {
```
dmitrylyzo, line 17: why add function explicitly?

src/plugins/logoScreensaver/plugin.js
```diff
@@ -59,23 +59,6 @@ define(['pluginManager'], function (pluginManager) {
...
-                { transform: 'translate3d(10px, 0, 0)', offset: 0.4 },
-                { transform: 'translate3d(-10px, 0, 0)', offset: 0.5 },
-                { transform: 'translate3d(10px, 0, 0)', offset: 0.6 },
-                { transform: 'translate3d(-10px, 0, 0)', offset: 0.7 },
-                { transform: 'translate3d(10px, 0, 0)', offset: 0.8 },
-                { transform: 'translate3d(-10px, 0, 0)', offset: 0.9 },
-                { transform: 'translate3d(0, 0, 0)', offset: 1 }];
-            var timing = { duration: 900, iterations: iterations };
-            return elem.animate(keyframes, timing);
-        }
```
dmitrylyzo, line 77: This animation just isn't added to the list.

src/components/maintabsmanager.js
```diff
@@ -22,7 +19,6 @@ define(['dom', 'browser', 'events', 'emby-tabs', 'emby-button'], function (dom,
                 footerTabsContainer.classList.add('footerTabs');
                 footerTabsContainer.classList.add('sectionTabs');
                 footerTabsContainer.classList.add('hide');
-                //appFooter.add(footerTabsContainer);
```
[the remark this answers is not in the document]
MrTimscampi: I think it's not really used anywhere. We might be able to remove this while section.

src/scripts/site.js
MrTimscampi: It's tripping up ESLint, as it thinks it's unused.

src/plugins/logoScreensaver/plugin.js
MrTimscampi: It isn't, but it's basically moving the image very fast  from side to side, which doesn't really look all that sexy. Hence why I removed it instead of re-enabling it.

src/components/maintabsmanager.js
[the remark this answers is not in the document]
dmitrylyzo: Yes, looks like it is disabled: https://github.com/jellyfin/jellyfin-web/blob/bc779ea2383c2ac39061145a1319453b9e12a67a/src/components/maintabsmanager.js#L10-L12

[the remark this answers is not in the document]
ferferga: Maybe we can bring it back so we can replace the negative margin stuff that we have right now for using two rows in tabs?

Unless we figure out first how to get rid of tabs completely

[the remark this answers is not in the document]
MrTimscampi: That would likely put the tabs where the mini player currently lives though, which is not ideal.

src/components/dialog/dialog.js
dmitrylyzo: There is no `btnCancel` on dialog `Delete media` (careful with that :wink: ).

MrTimscampi: I added some logic to the Metadata editor dialog when I reworked it. I figured it's better to handle that at the dialog level, since I'd like to go over the other dialogs to rework them a bit too.

dmitrylyzo: At current moment, dialog is closed this way:
https://github.com/jellyfin/jellyfin-web/blob/bc779ea2383c2ac39061145a1319453b9e12a67a/src/components/dialog/dialog.js#L97-L100
https://github.com/jellyfin/jellyfin-web/blob/bc779ea2383c2ac39061145a1319453b9e12a67a/src/components/dialog/dialog.js#L92-L95

src/elements/emby-slider/emby-slider.js
```diff
@@ -139,8 +139,9 @@ define(['browser', 'dom', 'layoutManager', 'keyboardnavigation', 'css!./emby-sli
             return;
         }
 
+        // TODO: Not sure what this is trying to accomplish.
         if (enableWidthWithTransform == null) {
-            //enableWidthWithTransform = browser.supportsCssAnimation();
+            enableWidthWithTransform = browser.supportsCssAnimation();
```
dmitrylyzo, line 144: With `transform`, I only see a little color transition.
slider-1
Without
slider-2

MrTimscampi: That doesn't look very good, I'll remove all of this as well, then.

src/controllers/livetv/livetvschedule.js
```diff
@@ -30,17 +30,14 @@ define(['layoutManager', 'cardBuilder', 'apphost', 'imageLoader', 'loading', 'sc
...
         recordingItems.innerHTML = cardBuilder.getCardsHtml(Object.assign({
             items: recordings,
             shape: enableScrollX() ? 'autooverflow' : 'auto',
             showTitle: true,
             showParentTitle: true,
             coverImage: true,
-            cardLayout: cardLayout,
-            centerText: !cardLayout,
+            cardLayout: appHost.preferVisualCards,
+            centerText: !appHost.preferVisualCards,
```
dmitrylyzo, line 40: Logic changed: `cardLayout = false;` overrides expression with `appHost.preferVisualCards`.

src/controllers/livetv/livetvschedule.js
```diff
@@ -30,17 +30,14 @@ define(['layoutManager', 'cardBuilder', 'apphost', 'imageLoader', 'loading', 'sc
...
         recordingItems.innerHTML = cardBuilder.getCardsHtml(Object.assign({
             items: recordings,
             shape: enableScrollX() ? 'autooverflow' : 'auto',
             showTitle: true,
             showParentTitle: true,
             coverImage: true,
-            cardLayout: cardLayout,
-            centerText: !cardLayout,
+            cardLayout: false,
+            centerText: false,
```
dmitrylyzo, line 40:
```suggestion
            centerText: true,
```

src/elements/emby-progressring/emby-progressring.js
```diff
@@ -11,23 +11,22 @@ define(['require', 'css!./emby-progressring', 'webcomponents'], function (requir
...
+                });
+
+                // configuration of the observer:
+                var config = { attributes: true, childList: false, characterData: false };
+
+                // pass in the target node, as well as the observer options
+                observer.observe(instance, config);
+
+                instance.observer = observer;
+            }
```
dmitrylyzo, line 29: Looks like we don't set `data-progress` anywhere.

src/libraries/scroller.js
```diff
@@ -872,12 +872,10 @@ define(['browser', 'layoutManager', 'dom', 'focusManager', 'ResizeObserver', 'sc
...
-
-            //if (!transform) {
-
-            //    item.scrollIntoView();
-            //    return;
-            //}
+            if (!transform) {
+                item.scrollIntoView();
+                return;
+            }
```
dmitrylyzo, line 878:
```
Uncaught ReferenceError: transform is not defined
    at scrollerFactory.to (scroller.js:875)
    at scrollerFactory.toCenter (scroller.js:921)
    at HTMLDivElement.<anonymous> (emby-scroller.js:14)
    at focus (focusManager.js:42)
    at Object.autoFocus (focusManager.js:22)
    at hometab.js:41
```
Because function is defined for prototype and var is out of scope.
Can be checked with TV layout.

src/elements/emby-progressring/emby-progressring.js
MrTimscampi: https://github.com/jellyfin/jellyfin-web/blob/master/src/controllers/dashboard/mediaLibrary.js#L282

https://github.com/jellyfin/jellyfin-web/blob/master/src/components/cardbuilder/cardBuilder.js#L1417

This would essentially allow us to dynamically update the progress on library refreshes, if we wish to.

dmitrylyzo: Currently, this blocks updating (by function?) and I see `0%`.
To reproduce, refresh one library.

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1860,9 +1708,5 @@ define(['browser', 'require', 'events', 'apphost', 'loading', 'dom', 'playbackMa
         });
     };
 
-    if (browser.chromecast) {
-        mediaManager = new cast.receiver.MediaManager(document.createElement('video'));
```
dmitrylyzo, line 1864: Is `PlayOn` feature implemented by dedicated player?
If so, all the code related to `mediaManager` can also be removed, right?

MrTimscampi: I don't think we ever even get to the Media Manager stuff, it's all hidden behind `if (mediaManager)` but that variable never gets assigned anything.

src/elements/emby-progressring/emby-progressring.js
MrTimscampi: I changed the way progress is updated in emby-itemrefreshindicator to use mutation instead of using setProgress directly in 44174a5, which fixes progress when refreshing libraries for me.

MrTimscampi: The main benefit of MutationObserver is that it's asynchronous. So we don't have to manually block execution each time something happens to update it. We simply set the value and let the JS engine determine when the best time to process the change is (It's a microtask, so it gets done once the queue is empty).

It doesn't use events and, managed properly, shouldn't really impact performance.

src/controllers/shows/tvlatest.js
```diff
@@ -23,8 +22,7 @@ import imageLoader from 'imageLoader';
 
     function loadLatest(context, params, promise) {
         promise.then(function (items) {
-            let html = '';
-            appHost.supports('imageanalysis');
+            var html = '';
```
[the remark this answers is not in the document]
MrTimscampi: I plan to make a PR to introduce the `no-var` rule and auto-fix these as soon as this is merged.

src/elements/emby-collapse/emby-collapse.js
```diff
@@ -11,11 +11,7 @@ import 'emby-button';
         elem.classList.remove('hide');
         elem.classList.add('expanded');
         elem.style.height = 'auto';
-        const height = elem.offsetHeight + 'px';
-        elem.style.height = '0';
-
-        // trigger reflow
-        const newHeight = elem.offsetHeight;
+        var height = elem.offsetHeight + 'px';
```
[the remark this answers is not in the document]
MrTimscampi: Erf, I was hoping we wouldn't need to force a reflow...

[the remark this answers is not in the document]
MrTimscampi: I added the forced reflow back in and wrapped it in a `requestAnimationFrame` so that it fires when it should.

src/controllers/shows/tvlatest.js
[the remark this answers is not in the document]
dmitrylyzo: You've reverted migrated code.

PR #1612, opened 2020-07-20
Add support for seeking with milliseconds

Maxr1998:
Needed for the media notification seekbar in jellyfin-android (available on Android 10+).
The relevant change needed in jellyfin-android is already merged.

See jellyfin/jellyfin-android#319.


dkanada:
I could have sworn we already did this.

MrTimscampi:
> I could have sworn we already did this.

We implemented something similar through the Media Session `seekto` action handler, thinking that it's what Android was using for the playback notifications, as that is what the mediasession.js file suggests (which is still useful, as other platforms use the Media Session API for notifications anyway).

Maxr1998:
Thanks!

PR #1620, opened 2020-07-21
Restore More button in metadata editor

MrTimscampi:
**Changes**

Fixes the removed More button in the metadata dialog.

**Issues**

### DIFF
--- a/src/components/metadataEditor/metadataEditor.js
+++ b/src/components/metadataEditor/metadataEditor.js
@@ -245,6 +245,43 @@ define(['itemHelper', 'dom', 'layoutManager', 'dialogHelper', 'datetime', 'loadi
         });
     }
 
+    function afterDeleted(context, item) {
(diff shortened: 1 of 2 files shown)


src/components/metadataEditor/metadataEditor.js
```diff
@@ -1088,6 +1128,7 @@ define(['itemHelper', 'dom', 'layoutManager', 'dialogHelper', 'datetime', 'loadi
                     elem.innerHTML = globalize.translateDocument(template, 'core');
 
                     elem.querySelector('.formDialogFooter').classList.remove('formDialogFooter');
+                    elem.querySelector('.btnClose').classList.add('hide');
```
dkanada, line 1131: Isn't this the same as `btnCancel`?

MrTimscampi: The purpose of that other selector is to have the cancel button outside of some cases but hide the top-right X button.

PR #1632, opened 2020-07-23
Remove debug leftovers

dmitrylyzo:
This is part of debug output (to limit verbosity) that should be removed in #1595
Origin: https://github.com/jellyfin/jellyfin-web/commit/bffafa69143b110f4a550c66642a4497366fad5b

**Changes**
Remove leftovers

Depends on #1595


dkanada:
thanks :sweat_smile:

Issue #934, opened 2020-03-12
All the Chinese plug-in subtitles in ASS format are displayed as squares


rushfly:
Here just one of them:
**Media Info of the file**

<details>
<pre>
<code>
General
Unique ID                   : 194207812477410874192100440539938439317 (0x921B106C07E042109BB612B44B544095)
Complete name               : (hide path)\Game.of.Thrones.S01E01.2011.BluRay.1080p.DTS.x264-CHD.mkv
Format                      : Matroska
Format version              : Version 4 / Version 2
File size                   : 5.99 GiB
Duration                    : 1 h 15 min
Overall bit rate            : 11.4 Mb/s
Encoded date                : UTC 2017-07-27 15:18:27
Writing application         : mkvmerge v14.0.0 ('Flow') 64bit
Writing library             : libebml v1.3.4 + libmatroska v1.4.5

Video
ID                          : 1
Format                      : AVC
Format/Info                 : Advanced Video Codec
Format profile              : High@L4.1
Format settings             : CABAC / 4 Ref Frames
Format settings, CABAC      : Yes
Format settings, ReFrames   : 4 frames
Codec ID                    : V_MPEG4/ISO/AVC
Duration                    : 1 h 1 min
Bit rate                    : 12.4 Mb/s
Width                       : 1 920 pixels
Height                      : 1 080 pixels
Display aspect ratio        : 16:9
Frame rate mode             : Constant
Frame rate                  : 23.976 FPS
Color space                 : YUV
Chroma subsampling          : 4:2:0
Bit depth                   : 8 bits
Scan type                   : Progressive
Bits/(Pixel*Frame)          : 0.250
Stream size                 : 5.34 GiB (89%)
Writing library             : x264 core 114 r1913 5fd3dce
[...]

PR #1636, opened 2020-07-24
Add options to use multiple custom fonts for subtitle rendering

nyanmisaka:
Part1: https://github.com/jellyfin/jellyfin/pull/3683
Part2: this

**Changes**
- Allow users to use multiple custom fallback font files as a solution for CJK fonts incorrectly rendered as boxes.
- In case too large fonts block the network, the maximum allowed total font size is 20 MB.
- Fonts in woff2 format are recommanded. It's lightweight and web-friendly.
- Allowed font format: `.woff .woff2 .ttf .otf`

**Issues**
https://github.com/jellyfin/jellyfin-web/issues/934
https://github.com/jellyfin/jellyfin-web/issues/913

CJK font samples:
https://github.com/CodePlayer/webfont-noto

fonts

font0


src/controllers/dashboard/encodingsettings.js
```diff
@@ -13,6 +13,8 @@ define(['jQuery', 'loading', 'globalize', 'dom', 'libraryMenu'], function ($, lo
         $('#txtDownMixAudioBoost', page).val(config.DownMixAudioBoost);
         page.querySelector('.txtEncoderPath').value = config.EncoderAppPathDisplay || '';
         $('#txtTranscodingTempPath', page).val(systemInfo.TranscodingTempPath || '');
+        $('#txtFallbackFontPath', page).val(config.FallbackFontPath || '');
```
MrTimscampi, line 16: Please use the non-jQuery version.

src/controllers/dashboard/encodingsettings.js
```diff
@@ -56,6 +58,8 @@ define(['jQuery', 'loading', 'globalize', 'dom', 'libraryMenu'], function ($, lo
             ApiClient.getNamedConfiguration('encoding').then(function (config) {
                 config.DownMixAudioBoost = $('#txtDownMixAudioBoost', form).val();
                 config.TranscodingTempPath = $('#txtTranscodingTempPath', form).val();
+                config.FallbackFontPath = $('#txtFallbackFontPath', form).val();
```
MrTimscampi, line 61: Same

src/controllers/dashboard/encodingsettings.js
```diff
@@ -181,6 +185,24 @@ define(['jQuery', 'loading', 'globalize', 'dom', 'libraryMenu'], function ($, lo
...
         });
+        $('#btnSelectFallbackFontPath', page).on('click.selectDirectory', function () {
+            require(['directorybrowser'], function (directoryBrowser) {
+                var picker = new directoryBrowser();
+                picker.show({
+                    includeFiles: true,
+                    includeDirectories: true,
+                    callback: function (path) {
+                        if (path) {
+                            $('#txtFallbackFontPath', page).val(path);
```
MrTimscampi, line 196: Same

src/controllers/dashboard/encodingsettings.js
nyanmisaka: Done.

frankyifei:
Can you make it support multiple woff2 fonts? I currently use three fonts each supports C, J and K. Because some subtitles are multilingual.

src/strings/en-us.json
```diff
@@ -1565,5 +1565,11 @@
     "ButtonCast": "Cast",
     "ButtonPlayer": "Player",
     "StopPlayback": "Stop playback",
-    "ClearQueue": "Clear queue"
+    "ClearQueue": "Clear queue",
+    "HeaderSelectFallbackFontPath" : "Select Fallback Font Folder Path",
+    "HeaderSelectFallbackFontPathHelp": "Browse or enter the path of the fallback font folder to use for rendering ASS/SSA subtitles.",
+    "LabelFallbackFontPath": "Fallback font folder path:",
+    "LabelFallbackFontPathHelp": "Specify a path containing fallback font for rendering ASS/SSA subtitles. The maximum allowed total font size is 20 Megabytes. It is recommended to use the lightweight and web-friendly woff2 format font. Leave it blank unless the font is rendered incorrectly.",
```
DMouse10462, line 1572: If fallback fonts are enabled with "EnableFallbackFont", there should be no reason to leave this blank (if you're leaving it blank, you should probably also disable the option entirely unless there are plans to have default _and configurable_ fallback fonts); maybe remove that guidance from the help text? See suggested change.

Restructured the help text for brevity and noun plurality. See suggested change.

jellyfin/jellyfin#3683 has the maximum individual size set to 10 MB and the server-side logging announces this as well. Please consider adding this to the help text alongside the max limit or remove the limit altogether (and let the server administrators decide what is best for their individual situation) per the suggestion from Bond-009: ["Why even have a max size?"](https://github.com/jellyfin/jellyfin/pull/3683#issuecomment-663647706) I'll leave this decision to others.

```suggestion
    "LabelFallbackFontPathHelp": "Specify a path containing fallback fonts for rendering ASS/SSA subtitles. The maximum allowed total font size is 20 MB. Lightweight and web-friendly font formats such as woff2 are recommended.",
```

src/strings/zh-cn.json
```diff
@@ -1578,5 +1578,11 @@
     "ClearQueue": "清空队列",
     "StopPlayback": "停止播放",
     "Writers": "作者",
-    "ViewAlbumArtist": "查看专辑艺术家"
+    "ViewAlbumArtist": "查看专辑艺术家",
+    "HeaderSelectFallbackFontPath" : "选择备用字体路径",
+    "HeaderSelectFallbackFontPathHelp": "浏览或输入一个包含备用字体文件的路径用于渲染 ASS/SSA 字幕。",
+    "LabelFallbackFontPath": "备用字体文件路径：",
+    "LabelFallbackFontPathHelp": "指定一个包含备用字体文件的路径用于渲染 ASS/SSA 字幕。允许的最大字体总容量为 20 MB。推荐使用轻量且适合网络传输的 woff2 格式字体。除非字体显示不正确，否则将其留空。",
```
DMouse10462, line 1585: Same here for the file size limit and suggestion to remove the guidance for leaving the field blank.

Also, is adding translations here okay for new source strings, or does Weblate not like that? I may be worrying too much here.

src/strings/en-us.json
DMouse10462: You may want to include all 4 supported formats here as you did in the body of the PR. That might be too verbose though; I'll leave that up to you and the other reviewers.

src/strings/zh-cn.json
nyanmisaka: > Also, is adding translations here okay for new source strings, or does Weblate not like that? I may be worrying too much here.

In addition to adding the basic English version when adding new strings, contributors can also add the language version they are familiar with. This is more direct and convenient than adding translation to Weblate after the PR is merged.

src/strings/en-us.json
nyanmisaka: > If fallback fonts are enabled with "EnableFallbackFont", there should be no reason to leave this blank (if you're leaving it blank, you should probably also disable the option entirely unless there are plans to have default and configurable fallback fonts)

There is currently no plan to preset fallback fonts on the server. Now only when the path is not empty can the fallback font be enabled. Otherwise the box will be automatically unchecked.

> jellyfin/jellyfin#3683 has the maximum individual size set to 10 MB and the server-side logging announces this as well. Please consider adding this to the help text alongside the max limit or remove the limit altogether

Now the total size of all fonts is 20 MB as the upper limit. This is sufficient for rendering subtitles using CJK subset fonts. Large fonts are very unfriendly when bandwidth is limited. But the end user may not know this.

PR #1647, opened 2020-07-26
Move wizard and user preference routes to nested folders

dkanada:


src/scripts/routes.js
```diff
@@ -61,40 +61,46 @@ define([
     });
 
     defineRoute({
-        path: '/mypreferencesmenu.html',
+        alias: '/mypreferencesmenu.html',
+        path: '/controllers/user/menu/index.html',
```
dkanada, line 65: I can also update the aliases here if we want to do that now.

dkanada:
I'll merge this one right after the video player one to avoid possible conflicts.

PR #1659, opened 2020-07-27
Fix incorrect h264 level support on iOS 12

thornbill:
**Changes**
Adds check to disallow h264 level 51 on iOS versions less than 13

**Issues**
https://github.com/jellyfin/jellyfin/issues/3263

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -5,6 +5,17 @@ define(['browser'], function (browser) {
         return !!(videoTestElement.canPlayType && videoTestElement.canPlayType('video/mp4; codecs="avc1.42E01E, mp4a.40.2"').replace(/no/, ''));
     }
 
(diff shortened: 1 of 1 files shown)


src/scripts/browserDeviceProfile.js
```diff
@@ -692,6 +692,8 @@ define(['browser'], function (browser) {
         var h264Profiles = 'high|main|baseline|constrained baseline';
 
         if (browser.tizen || browser.web0s ||
+            // iOS 12 incorrectly reports support
+            !(browser.iOS && browser.iOSVersion < 13) &&
```
MrTimscampi, line 696: Making a function like `canPlayH264Level51` or something would simplify the complexity of this a bit.

thornbill: :+1: updated.

thornbill:
Closing in favor of https://github.com/jellyfin/jellyfin-web/pull/1662

Issue #1631, opened 2020-07-23
Edgium can direct play HEVC on 10.5.0, but can't on 10.6.0


ThatNerdyPikachu:
**Describe the bug**
HEVC content is able to be direct played on Edgium in Jellyfin 10.5.0, but not in 10.6.0

**To Reproduce**
1. Play a HEVC video

**Expected behavior**
1. The video should be direct played

**System (please complete the following information):**
 - OS: Windows (unsure if macOS Edgium has HEVC support)
 - Browser: Edgium
 - Jellyfin Version: 10.6.0

ThibaultNocchi:
Does it work with JF 10.6.1?

nyanmisaka:
> Does it work with JF 10.6.1?

It should work. I have tried on my edge chromium.

PR #1660, opened 2020-07-27
Fix mkv directplay on Edge chromium

nyanmisaka:
**Changes**
- add initial support for Edge chromium.
- mark mkv as supported in Edge chromium. (It reports that it is unsupported but it actually does just like edgeUwp)
- H.265 directplay requires the extension from microsoft store and a capable GPU. (software playback is unsupported)

**Notes**
edg -> edge chromium on Windows or macOS
edga -> edge chromium on Android
edgios -> edge chromium on iOS

**Issues**
https://github.com/jellyfin/jellyfin-web/issues/1631

edgechromium


nyanmisaka:
Not sure if mkv is also supported on Edge chromium macOS.

ThibaultNocchi:
Should #1631 be closed then?

MrTimscampi:
> Should #1631 be closed then?

If someone can confirm that it works, then yes :)

This hit 10.6.1 so the fix is in the latest release (from last night)

ThibaultNocchi:
Ok then I'll test this evening if no one has done it

johnvick:
I still get Edge browser transcoding 10 bit HEVC on Jelly 10.6.1 whereas on 10.5.5 it played directly. I have the Microsoft HEVC addon installed.

josersanchez0117:
Here's the free HEVC codec. [https://www.microsoft.com/en-us/p/hevc-video-extensions-from-device-manufacturer/9n4wgh0z6vhq?activetab=pivot:overviewtab](https://www.microsoft.com/en-us/p/hevc-video-extensions-from-device-manufacturer/9n4wgh0z6vhq?activetab=pivot:overviewtab
)
Also, I tested it and it does work. HEVC Main 10 direct plays with no problems.

johnvick:
Tried this again on three computers - Surface Book 2 i5, AMD Ryzen + Nvidia 1660 and AMD Ryzen with APU. The conclusion is that some HEVC 10 bit play directly and some don't - the ones that don't (on my limited testing) have DTS soundtracks and the ones that do have AAC 2 channel.

nyanmisaka:
> Tried this again on three computers - Surface Book 2 i5, AMD Ryzen + Nvidia 1660 and AMD Ryzen with APU. The conclusion is that some HEVC 10 bit play directly and some don't - the ones that don't (on my limited testing) have DTS soundtracks and the ones that do have AAC 2 channel.

Edge chromium doesn't support DTS which should result in transcoding.

PR #1662, opened 2020-07-27
Add h264 codec profile for TS container

dmitrylyzo:
In this case we DirectPlay h264 level 5.1, but TS conainer will limited to 4.2.

**Changes**
Add special condition for TS container.

**Issues**
Fixes jellyfin/jellyfin#3263

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -707,33 +707,29 @@ define(['browser'], function (browser) {
             }
         }
 
-        profile.CodecProfiles.push({
-            Type: 'Video',
-            Codec: 'h264',
-            Conditions: [
-                {
-                    Condition: 'NotEquals',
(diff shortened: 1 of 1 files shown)


src/scripts/browserDeviceProfile.js
```diff
@@ -707,33 +707,29 @@ define(['browser'], function (browser) {
...
-                },
-                {
-                    Condition: 'LessThanEqual',
-                    Property: 'VideoLevel',
-                    Value: maxH264Level.toString(),
-                    IsRequired: false
-                }
-            ]
-        });
+        const h264CodecProfileConditions = [
```
thornbill, line 710: Is `const` safe to use in 10.6? @MrTimscampi

MrTimscampi: Yep, Babel converts everything back to ES5.

PR #1667, opened 2020-07-28
Get and tag with actual release version in CI

joshuaboniface:
**Changes**
Ensures that the intermediate Docker images have the proper version in their names for the collect script to handle.

**Issues**
N/A

### DIFF
--- a/.ci/azure-pipelines-package.yml
+++ b/.ci/azure-pipelines-package.yml
@@ -59,7 +59,15 @@ jobs:
   pool:
     vmImage: 'ubuntu-latest'
 
+  variables:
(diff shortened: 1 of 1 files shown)


joshuaboniface:
Actually, this may be broken.

```
joshua@base ~ $ awk -F "/" "{ print $NF }" <<<"refs/tags/v10.6.1"
refs/tags/v10.6.1
joshua@base ~ $ awk -F '/' '{ print $NF }' <<<"refs/tags/v10.6.1"
v10.6.1
```

I guess we'll see for `10.6.2`.

Issue #1417, opened 2020-06-16
Rework NativeShell/AppHost


MrTimscampi:
This issue is to track and discuss a rewrite of NativeShell/AppHost into a system that makes more sense and is well-documented, well-thought-out, and just generally better made.

Here is a preliminary idea of what it could look like, based on a 5min discussion between @dkanada and me.

```
Jellyfin
 -> Platform
        -> getPlugins
        -> getFeatures
        -> deviceName
        -> deviceId
 -> Device
        -> fullscreen
        -> openUrl
        -> rotate
 -> Storage
        -> download
        -> isCached
        -> localDirectory
```

dkanada:
We would have one major update with backwards compatibility to allow time for clients to update their global uses, and then drop the hacks. That would let us add a compatibility layer just before an update, and immediately revert it after a release since we use release branches.

Maxr1998:
I agree, this is a good idea. In the device section, there's missing a function that supplies the current media metadata though (e.g. for a notification).

We should also ask ourselves whether the new API methods should be synchronous or asynchronous. The new Android app uses [`WebView.addJavascriptInterface`](https://developer.android.com/reference/android/webkit/WebView#addJavascriptInterface(java.lang.Object,%20java.lang.String)) which registers itself as a window object with a collection of *synchronous* functions. Supporting async calls/callbacks there is possible, but kinda hacky - currently, I have to register the callbacks in a map on the Javascript side, which I can later call from Java with `WebView.loadUrl()`.

cromefire:
If it's synchronous it doesn't work with jf desktop at all, best just wrap any call with `Promise.resolve()` so it's irrelevant if it's sync or async

Maxr1998:
Yeah, that'll need JavaScript code though. Should that go into jf-web or jf-android?

cromefire:
So it's quite easy to write a wrapper that web (or Android) could used to call native shell methods an convert anything in to a promise (so catch all errors and pass them to `Promise.reject` and pass the result to `Promise.resolve`) it's like 10 lines or so

cromefire:
I'm also happy to help with this but only after web is ts because otherwise I'm sure there'll be quite a bunch of issues (and it's a pain to work with).

PR #1677, opened 2020-07-28
Improve NativeShell integration

Maxr1998:
**Changes**
- Move NativeShell calls into shell.js where possible to better group them
- Add "updateVolumeLevel" API (not used yet)

Regarding the `updateVolumeLevel()`: I've tried a few places where I could put it, but they either didn't work at all (cause they were probably the wrong place for remote/session playback) or the `player.getVolume()` wasn't updated yet at that point. If someone could point me to a callback where I could call that function, then I could integrate this as well. :smile:

**Issues**
- Related to #1417

### DIFF
--- a/src/scripts/shell.js
+++ b/src/scripts/shell.js
@@ -1,20 +1,43 @@
 // TODO: This seems like a good candidate for deprecation
 export default {
-    openUrl: function (url, target) {
+    enableFullscreen: function() {
+        window.NativeShell?.enableFullscreen();
+    },
+    disableFullscreen: function() {
+        window.NativeShell?.disableFullscreen();
+    },
+    openUrl: function(url, target) {
         if (window.NativeShell) {
             window.NativeShell.openUrl(url, target);
         } else {
             window.open(url, target || '_blank');
         }
     },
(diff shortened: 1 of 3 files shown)


EsaiasTilly (approved):

EsaiasTilly (approved):

src/scripts/shell.js
```diff
@@ -1,20 +1,53 @@
 // TODO: This seems like a good candidate for deprecation
 export default {
-    openUrl: function (url, target) {
+    enableFullscreen: function() {
+        if (window.NativeShell) {
+            window.NativeShell.enableFullscreen();
```
thornbill, line 5: This looks like a good place to use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) `?.`.

dkanada: Also, the diff would be much easier to read if you move the fullscreen methods back below the URL one, but I suppose it doesn't matter much.

dkanada (approved):

thornbill (approved):

PR #1685, opened 2020-07-29
Changed playback mouse timing.

Supergamer1337:
Co-authored-by: Esaias Tilly <esaias@tillys.se>
**Changes**
Changed inconsistent timing of mouse and OSD controls. Mouse now disappears and reappears at the same time as OSD.

**Issues**
None found...


dmitrylyzo:
Mouse timing is here https://github.com/jellyfin/jellyfin-web/blob/0595488a3fbc7bb93d3a6d7dcb7b7811bdda9064/src/scripts/mouseManager.js#L100-L113

PR #1699, opened 2020-07-30
Remove redundant variable assignments

MrTimscampi:
**Changes**

**Issues**


dkanada:
Is this an ESLint rule or something? Might want to enable it.

MrTimscampi:
> Is this an ESLint rule or something? Might want to enable it.

Not, a native one. It's a plugin with no updates for a year and less than 100 pulls per week, hence why it's not enabled :p

I figure these are easy enough to catch when they happen in the future.

PR #1718, opened 2020-08-01
Migrate routes.js and mouseManager to ES6

MrTimscampi:
**Changes**

Also adds a console group for messages output when registering core routes, because cleanliness.

**Issues**


src/components/viewContainer.js
```diff
@@ -23,35 +22,35 @@ define(['browser', 'dom', 'layoutManager', 'css!components/viewManager/viewConta
     }
 
     function getRequirePromise(deps) {
-        return new Promise(function (resolve, reject) {
+        return new Promise((resolve, reject) => {
             require(deps, resolve);
```
Camc314, line 26: remove require?

src/components/viewContainer.js
```diff
@@ -23,35 +22,35 @@ define(['browser', 'dom', 'layoutManager', 'css!components/viewManager/viewConta
...
+            const isPluginpage = options.url.toLowerCase().indexOf('/configurationpage') !== -1;
+            const newViewInfo = normalizeNewView(options, isPluginpage);
+            const newView = newViewInfo.elem;
+            const modulesToLoad = [];
 
-            return new Promise(function (resolve) {
-                require(modulesToLoad, function () {
-                    var currentPage = allPages[pageIndex];
+            return new Promise((resolve) => {
+                require(modulesToLoad, () => {
```
Camc314, line 46: as above

src/components/viewContainer.js
```diff
@@ -1,20 +1,19 @@
-define(['browser', 'dom', 'layoutManager', 'css!components/viewManager/viewContainer'], function (browser, dom, layoutManager) {
-    'use strict';
+/* eslint-disable indent */
```
dmitrylyzo, line 1: where is `css!components/viewManager/viewContainer`?

Camc314: [here?](https://github.com/jellyfin/jellyfin-web/blob/master/src/components/viewManager/viewContainer.css)

dmitrylyzo: I mean where is its `import`

Camc314: Oh right 🤦‍♀️

MrTimscampi: Indeed, I was a bit overzealous when removing unused imports :p

src/components/viewContainer.js
```diff
@@ -1,115 +1,106 @@
...
+        let controllerUrl = view.getAttribute('data-controller');
 
         if (controllerUrl) {
             if (controllerUrl.indexOf('__plugin/') === 0) {
                 controllerUrl = controllerUrl.substring('__plugin/'.length);
             }
 
             controllerUrl = Dashboard.getConfigurationResourceUrl(controllerUrl);
-            return getRequirePromise([controllerUrl]).then(function (ControllerFactory) {
+            return import(controllerUrl).then((ControllerFactory) => {
```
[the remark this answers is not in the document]
MrTimscampi: It's importing random stuff, so I'm not too sure.

It looks like it's used for frontend plugin controllers (Web plugins can add pages to the client), but we don't have any to test it, so...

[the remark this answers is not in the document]
Camc314: I don't know how relevant [this](https://github.com/jellyfin/jellyfin-web/blob/master/src/controllers/music/musicrecommended.js#L304) is, but for moviesreccomended/musicrecommended/tvrecommended, I have used `default` for the import. 

From what I have read, if a module is imported as above (`import(X).then((Y) => {`), when `Y` is called, it will need to be called as `Y.default(...`.
If it is imported as `import(X).then(({default: Y}) => {`, then it should be called as `Y(...`

This is also making the assumption that all `X` have a default export.

Hopefully that made some amount of sense.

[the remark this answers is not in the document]
MrTimscampi: Looking at the source, we never set the `data-controller` attribute, so this code is never reached, I think?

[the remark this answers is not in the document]
Camc314: I don't think so either. So I guess it's more unused code.

PR #1723, opened 2020-08-01
Update Swiper to 6.1.1

MrTimscampi:
**Changes**

Updates the dependency and changes the import paths (See: https://github.com/nolimits4web/swiper/blob/v6.0.0/CHANGELOG.md#swiper-600---released-on-july-3rd-2020)

Tested locally with a photo library.

**Issues**


dmitrylyzo:
> webOS 1.2 has problems - currently looking

In `dom7 3.0.0-alpha.7`, all `webkit*` properties are removed - that's why in old browser (webOS 1.2) slides don't slide.

MrTimscampi:
> > webOS 1.2 has problems - currently looking
> 
> In `dom7 3.0.0-alpha.7`, all `webkit*` properties are removed - that's why in old browser (webOS 1.2) slides don't slide.

You mean in CSS? If so, we can probably play with autoprefixer to fix this. If it's JS, however, I don't think there's a solution...

dmitrylyzo:
> You mean in CSS? If so, we can probably play with autoprefixer to fix this. If it's JS, however, I don't think there's a solution...

https://github.com/nolimits4web/dom7/blob/d9cc22fe95de082a35dc16c9733254421d614b49/src/methods.js#L180-L185
was
https://github.com/nolimits4web/dom7/blob/92eb034f3b11f641828b9294d3ad66da1e5ef0e7/src/methods.js#L185-L192

I wonder if we can somehow polyfill `transform` and `transition` properties in `CSSStyleDeclaration`?

dkanada:
@dmitrylyzo any updates on WebOS ideas here? This would be a good merge while we hold for webpack to finish.

dmitrylyzo:
I posted this on Matrix for MrTim:
```js
// Polyfill for vendor prefixed style properties

const vendorProperties = {
    'transform': ['webkitTransform'],
    'transition': ['webkitTransition']
};

const elem = document.createElement('div');

function polyfillProperty(name) {
    if (elem.style[name] === undefined) {
        for (const vendorName of vendorProperties[name] || []) {
            if (elem.style[vendorName] !== undefined) {
                console.debug('polyfill', name, vendorName);

                Object.defineProperty(CSSStyleDeclaration.prototype, name, {
                    get: function () { return this[vendorName]; },
                    set: function (val) { this[vendorName] = val; }
                });

                break;
            }
        }
    }
}

polyfillProperty('transform');
polyfillProperty('transition');
```

dkanada:
Seems fine to me, if you want to apply that I think this is mergeable for now. We'll want to remove that eventually but oh well.

PR #1727, opened 2020-08-01
Don't use nested Promises

MrTimscampi:
**Changes**

* Introduces the `promise/no-nesting` rule in our ESLint config (Commented, as this is a multiple PR effort)
* Converts a bunch of nested promises to chained `then`.

**Issues**

### DIFF
--- a/src/components/collectionEditor/collectionEditor.js
+++ b/src/components/collectionEditor/collectionEditor.js
@@ -67,23 +67,20 @@ import 'flexStyles';
 
     function addToCollection(apiClient, dlg, id) {
         const url = apiClient.getUrl(`Collections/${id}/Items`, {
-
             Ids: dlg.querySelector('.fldSelectedItemIds').value || ''
         });
 
-        apiClient.ajax({
-            type: 'POST',
-            url: url
-
-        }).then(() => {
+        apiClient.ajax({ type: 'POST', url: url })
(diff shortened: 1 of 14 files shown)


src/components/collectionEditor/collectionEditor.js
```diff
@@ -67,23 +67,20 @@ import 'flexStyles';
...
             Ids: dlg.querySelector('.fldSelectedItemIds').value || ''
         });
 
-        apiClient.ajax({
-            type: 'POST',
-            url: url
-
-        }).then(() => {
+        apiClient.ajax({ type: 'POST', url: url })
+        .then(() => {
```
dkanada, line 74:
```suggestion
        apiClient.ajax({ type: 'POST', url: url }).then(() => {
```

dkanada: You don't like this method?

MrTimscampi: I find putting them on the next line more readable. It's easier to see at a glance where the different blocks are.

It's also what Prettier uses, so if we ever decide to use that (there has been talks about it before), it's one less change (We're already not very far from the style it uses).

YouKnowBlom: I agree, newline looks much cleaner when chaining

Issue #940, opened 2020-03-13
Make the poster a play button


theg00s3:
The poster/episode screencap should play the movie/tv show when pressed, instead of editing the pictures. I've marked the area in case I'm not clear.
![Screenshot_2020-03-13 Jellyfin](https://user-images.githubusercontent.com/18522748/76604849-98ec1d80-6563-11ea-882c-cbc960d2c736.jpg)

Given that the "Continue Watching" row on the home page reacts similarly, it would feel more cohesive UI wise for the poster to do the same thing.

MrTimscampi:
Covered by an open PR, but please redirect feature requests and suggestions to https://features.jellyfin.org/

PR #1737, opened 2020-08-03
Added overlay and play functionality.

Supergamer1337:
Co-authored-by: Esaias Tilly <esaias@tillys.se>

**Changes**
Well, the poster on the detail page now brings up the overlay.

**Issues**
Adds the request feature of #940 

(Also, an oopsie was made on the last Pull Request... Which essentially "merged" the branch when trying to update.)

### DIFF
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -1,5 +1,6 @@
 import loading from 'loading';
 import appRouter from 'appRouter';
+import itemShortcuts from 'itemShortcuts';
 import layoutManager from 'layoutManager';
 import connectionManager from 'connectionManager';
 import * as userSettings from 'userSettings';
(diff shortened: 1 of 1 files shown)


Supergamer1337:
As an aside note, I think it would look better if we could remove the extra options beside the play button in the overlay on the details page.

src/controllers/itemDetails/index.js
```diff
@@ -1894,6 +1895,10 @@ import 'emby-select';
             playCurrentItem(this, this.getAttribute('data-mode'));
         }
 
+        function onPosterClick(e) {
+            itemShortcuts.onClick.call(view.querySelector('.detailImageContainer'), e);
```
dkanada, line 1899: Wouldn't you be able to use `this` instead of querying for the element again?

MrTimscampi: It's to be explicit as to what the container is, since `itemShortcuts.onClick` is usually used on viewContainers and seems to expect the root item to be a container.

dkanada:
Any chance you can remove the buttons in the corners for this card? Not sure we want to leave those in this situation.

PR #1742, opened 2020-08-03
Remove unused strings from source

dkanada:


scripts/duplicates.py
```diff
@@ -0,0 +1,33 @@
...
+    strings = json.load(en)
+    for key, value in strings.items():
+        if value not in reverse:
+            reverse[value] = [key]
+        else:
+            reverse[value].append(key)
+
+for key, value in reverse.items():
+    if len(value) > 1:
+        duplicates[key] = reverse[key]
```
dmitrylyzo, line 24: `value` is `reverse[key]`, no?

src/strings/en-us.json
```diff
@@ -93,7 +91,6 @@
     "ButtonHelp": "Help",
     "ButtonHome": "Home",
     "ButtonInfo": "Info",
-    "ButtonLearnMore": "Learn more",
```
dmitrylyzo, line 96: https://github.com/jellyfin/jellyfin-plugin-fanart/blob/8bb2c0ddd4e5caf9a3ac5c1c894e8158040300f0/Jellyfin.Plugin.Fanart/Configuration/configPage.html#L16

scripts/duplicates.py
dkanada, line 24:
```suggestion
        duplicates[key] = value
```

dkanada: good catch

src/strings/en-us.json
dkanada: The current translation method is a complete hack, so we shouldn't encourage plugins to be using it right now in my opinion.

PR #1755, opened 2020-08-04
Change Upcoming to Play next

KucharczykL:
**Changes**
I have changed Upcoming to Play next. I felt Upcoming is more for unaired TV shows (and this separate and and not being changed) than recommendations for what to play next from shows you've already started watching. This is also different than Continue watching that is meant for unfinished episodes.

Please let me know if I did something incorrectly. I am unsure if replacing strings with completely different strings is done like this.

**Issues**
None


dkanada:
After some discussion, we decided this tab isn't even in the scope of Jellyfin and will be removed for the next release.

PR #1776, opened 2020-08-06
Remove focus-prevent-scroll, migrate gamepadtokey and chromecastHelper to ES6

MrTimscampi:
**Changes**

Make sure to disable whitespace changes when reviewing.

**Issues**


src/plugins/chromecastPlayer/chromecastHelper.js
```diff
@@ -1,229 +1,229 @@
...
-            return Promise.resolve(serverAddress);
-        }
+    return false;
+}
 
-        var cachedValue = getCachedValue(serverAddress);
-        if (cachedValue) {
-            return Promise.resolve(cachedValue);
-        }
+function getServerAddress(apiClient) {
```
dmitrylyzo, line 173: Same as `default`.
```suggestion
export function getServerAddress(apiClient) {
```
tbh, I'm currently not sure if we should export each function/class.

src/components/scrollManager.js
```diff
@@ -532,9 +532,7 @@ import layoutManager from 'layoutManager';
 
     if (isEnabled()) {
         dom.addEventListener(window, 'focusin', function(e) {
-            setTimeout(function() {
```
MrTimscampi, line 535: This is untested.

src/plugins/chromecastPlayer/chromecastHelper.js
MrTimscampi: It's a decision we'll have to come back in once the Webpack PR is done, imo.

The "export default" route means we import everything, even stuff we don't need (And it ends up in the bundle). It also didn't work for me in some rare cases (Exporting a new anonymous instance of a class, for example).

I've changed a few files to use only "export whatever", especially those anonymous instances of a class.

src/components/scrollManager.js
dmitrylyzo: Doesn't work.
There's a timeout https://github.com/calvellido/focus-options-polyfill/blob/c841df179f81bb81ee51309d9c6b77066cfad155/index.js#L74-L80
It was added to support iOS (13?) but now polyfill has "blinking".

I tried to remove timeout from polyfill and returned timeout in `scrollManager` (it is required in this case) - `scrollManager` starts to work (obviously), but `emby-scroller` still doesn't work.

PR #1788, opened 2020-08-08
Rework CSS breakpoints and structure

MrTimscampi:
**Changes**

This starts implementing the improvements discussed in [this post](https://github.com/jellyfin/jellyfin-web/issues/836#issuecomment-654989680).

A few things are done:

* **Replace most existing breakpoints with the ones listed in the post linked above**
  The method used to select the proper replacement is to multiply the existing value by 16 (The number of pixels in 1em for media queries) then pick the closest one. In some cases, an arbitrary choice is made depending on the intended devices.
* **Remove obsolete queries**
  I spotted one or two queries that resolved to 800x600, which doesn't make sense in this day and age. They've been removed.
* **Merge duplicate queries (in progress)**
  In keeping with the idea of a "mobile-first" approach to design (which is common nowadays), this PR moves all the queries to the end of their files. Eventually, the goal is to rewrite the CSS to define the mobile layout first, then override it for different resolutions without relying too much on the `layout-mobile` class.
* **Minor changes to existing style**
  To start reaping the benefits of this change immediately, a few minor edits have been made to the existing style. It mostly amounts to changing some properties on some specific breakpoints or hiding elements at different resolutions.

**Issues**

Not really an issue, but worth a mention: the card sizes are a goddamn mess. They've been left largely untouched, due to needed some changes in JS code as well as CSS. They'll get a cleanup in a future PR.

### DIFF
--- a/src/assets/css/dashboard.css
+++ b/src/assets/css/dashboard.css
@@ -47,16 +47,6 @@ progress[aria-valuenow]::before {
     margin-bottom: 2.2em !important;
 }
 
-@media all and (min-width: 50em) {
-    .type-interior > div[data-role=content],
-    .type-interior > .ui-panel-content-wrap > div[data-role=content] {
-        padding-right: 0;
-        padding-left: 0;
-        padding-top: 0;
-        overflow: hidden;
-    }
-}
-
 .dashboardDocument .dashboardEntryHeaderButton,
 .dashboardDocument .lnkManageServer {
     display: none !important;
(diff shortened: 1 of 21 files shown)


dkanada:
I think we should try and be desktop first rather than mobile.

MrTimscampi:
> I think we should try and be desktop first rather than mobile.

The idea of "mobile first" isn't to make a mobile design then make it slightly bigger because of a bigger screen.

It generally leads to fewer lines of CSS and a more consistent experience.

It's not meant as "we do a mobile design and the rest is an after thought". It's simply that it's better to build on top of that simpler design than to take away elements of a complex one to fit on a mobile.

Currently, we rely on an artificial separation between the desktop and mobile layout, where we really should be fully responsive, which is a lot easier to do with a mobile-first approach.

Imo, "desktop first" only leads to messy CSS down the line (which we already have).

PR #1796, opened 2020-08-09
Fix some odd strings and remove duplicate values

dkanada:
```
ButtonAdd
LabelAll
TabArtists
ButtonRepeat
ButtonHelp
ButtonSave
ButtonToggleContextMenu
ButtonSearch
TabDisplay
HeaderDisplay
HeaderFilters
HeaderFeatures
HeaderTags
TabMetadata
TabTranscoding
TabPlayback
TabSuggestions
```


src/scripts/libraryMenu.js
```diff
@@ -387,7 +387,7 @@ import 'flexStyles';
             pageIds: ['mediaLibraryPage', 'librarySettingsPage', 'libraryDisplayPage', 'metadataImagesConfigurationPage', 'metadataNfoPage'],
             icon: 'folder'
         }, {
-            name: globalize.translate('TabPlayback'),
+            name: globalize.translate('TitlePlayback'),
```
MrTimscampi, line 390: Why not rename it simply "Playback"?

dkanada: There wasn't a duplicate string called `Playback`, only these two.

MrTimscampi: But shouldn't we rename it to `Playback` at some point, just to make it clearer? (It's not blocking though, to be clear :p )

dkanada: I definitely agree with that, I just didn't want to affect too much with this pull request.

PR #1799, opened 2020-08-11
Hitting space to check/uncheck checkboxes doesn't work on firefox

matjaz321:
**Changes**
It seems like checking/unchecking behaviour is already working correctly on the Firefox.
So when `this.checked = !this.checked` was called it just reset the checkbox to previous value.

**Issues**
Solves #1797

### DIFF
--- a/src/elements/emby-checkbox/emby-checkbox.js
+++ b/src/elements/emby-checkbox/emby-checkbox.js
@@ -10,8 +10,10 @@ import 'webcomponents';
     function onKeyDown(e) {
         // Don't submit form on enter
         // Real (non-emulator) Tizen does nothing on Space
-        if (e.keyCode === 13 || e.keyCode === 32) {
+        if ((e.keyCode === 13 || e.keyCode === 32) && !browser.firefox) {
             e.preventDefault();
+
+            this.checked = !this.checked;
 
             this.dispatchEvent(new CustomEvent('change', {
                 bubbles: true
--- a/src/elements/emby-checkbox/emby-checkbox.js
+++ b/src/elements/emby-checkbox/emby-checkbox.js
@@ -10,7 +10,7 @@ import 'webcomponents';
     function onKeyDown(e) {
         // Don't submit form on enter
         // Real (non-emulator) Tizen does nothing on Space
-        if ((e.keyCode === 13 || e.keyCode === 32) && !browser.firefox) {
+        if (e.keyCode === 13 || (e.keyCode === 32 && browser.tizen)) {
             e.preventDefault();
 
             this.checked = !this.checked;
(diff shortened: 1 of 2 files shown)


dkanada:
This doesn't look like it keeps support for other browsers, the top of the method has a comment about Tizen workarounds, and the method seems somewhat useless without this line.

matjaz321:
> This doesn't look like it keeps support for other browsers, the top of the method has a comment about Tizen workarounds, and the method seems somewhat useless without this line.

Hmm okei

MrTimscampi:
> > This doesn't look like it keeps support for other browsers, the top of the method has a comment about Tizen workarounds, and the method seems somewhat useless without this line.
> 
> Hmm okei, but I noticed that this is actually causing this bug in the firefox...

You can add a condition like `if (!browser.firefox) {}` if needed.

matjaz321:
> > > This doesn't look like it keeps support for other browsers, the top of the method has a comment about Tizen workarounds, and the method seems somewhat useless without this line.
> > 
> > 
> > Hmm okei, but I noticed that this is actually causing this bug in the firefox...
> 
> You can add a condition like `if (!browser.firefox) {}` if needed.

Yeah I thought about that

src/elements/emby-checkbox/emby-checkbox.js
```diff
@@ -10,7 +10,7 @@ import 'webcomponents';
     function onKeyDown(e) {
         // Don't submit form on enter
         // Real (non-emulator) Tizen does nothing on Space
-        if (e.keyCode === 13 || e.keyCode === 32) {
+        if ((e.keyCode === 13 || e.keyCode === 32) && !browser.firefox) {
```
dmitrylyzo, line 13: In this case, you will probably submit form on Firefox.
Looks like Firefox fires `click` even after `preventDefault`.
We can ignore `Space` for Firefox, but  we can also handle `Space` only for Tizen.
```suggestion
        if (e.keyCode === 13 || (e.keyCode === 32 && browser.tizen)) {
```

dmitrylyzo:
Same fix could be applied to `emby-radio`. It works just because it cannot be unchecked by click.
https://github.com/jellyfin/jellyfin-web/blob/ff5833a5f6950456b456845b68870f001b0c72fb/src/elements/emby-radio/emby-radio.js#L12

src/elements/emby-radio/emby-radio.js
```diff
@@ -9,7 +9,7 @@ import 'webcomponents';
     function onKeyDown(e) {
         // Don't submit form on enter
         // Real (non-emulator) Tizen does nothing on Space
-        if (e.keyCode === 13 || e.keyCode === 32) {
+        if (e.keyCode === 13 || (e.keyCode === 32 && browser.tizen)) {
```
dmitrylyzo, line 12: Add `browser` import

matjaz321: ah right I didn't see it :/

PR #1802, opened 2020-08-12
Reorder CSS loading

dmitrylyzo:
1. Base CSS, element CSS, etc. in head
2. Theme CSS in body
3. Branding CSS in body

It seems to work without placeholders (in `index.html`).
What is better?
- placeholders and remove creation
- no placeholder for branding CSS - will be appended
- no placeholders

**Changes**
Reorder CSS loading

**Issues**
White navdrawer.
_In fact, this can be fixed by removing `background-color` for `touch-menu-la`._

I am not sure, but this approach might help to reload theme without full reload.

### DIFF
--- a/src/scripts/site.js
+++ b/src/scripts/site.js
@@ -384,11 +384,15 @@ function initClient() {
                             return response.text();
                         })
                         .then(function(css) {
-                            // Inject the branding css as a dom element in body so it will take
-                            // precedence over other stylesheets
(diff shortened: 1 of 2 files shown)


dmitrylyzo:
If we have any problems with CSS order due to modules loading order, it will be easy to add placeholders (for now, I removed them).

EraYaN (approved):

PR #1806, opened 2020-08-13
Bug fix for: Cannot exit unavailable/private trailers

### LINKED ISSUE #1803: Cannot exit unavailable/private trailers
**Describe The Bug**
When attempting to view a trailer for certain shows, the trailer is unable to load because the trailer's YouTube video is marked private.
 A spinner is visible indefinitely over the YouTube player. Because the trailer has not finished loading, the controls are not visible preventing the user from exiting the YouTube player view. Back and Forward browser controls do nothing because the player is on top of all other content.
The user must refresh the browser to use the interface again

**Steps To Reproduce**
No steps to reproduce because most trailers are playable. I could not find any ways to override the trailer to insert a private video.

**Expected Behavior**
Controls should be visible allowing a user to exit along with Back and Forward buttons

**Logs**
None

**Screenshots**
This is a fullscreen capture of the browser window
https://i.imgur.com/6Vnovrg.png

 - Platform: Windows
 - Browser: Edge Chromium
 - Jellyfin Version: 10.6.2

**Additional Context**
While the issue stems from invalid 3rd party metadata, I think that Jellyfin should have some way to exit out of invalid trailers.

matjaz321:
**Changes**
I've added `onError` event listener to the iframe youtube player to destroy the youtube player whenever it receives an error
**Issues**
Fixes #1803
**Additional**
So I think the UX might not be ideal when youtube player has an error... So user clicks on trailer button, the player is opened in the full screen, loading spinner starts, then you can see youtube error `video unavailable` and then the player is closed and spinner is hidden


src/plugins/youtubePlayer/plugin.js
```diff
@@ -142,6 +142,10 @@ function setCurrentSrc(instance, elem, options) {
                             } else if (event.data === YT.PlayerState.PAUSED) {
                                 events.trigger(instance, 'pause');
                             }
+                        },
+                        'onError': () => {
+                            instance.destroy();
+                            loading.hide();
```
dmitrylyzo, line 148: `suicidal plugin` :)

@matjaz321 What if plugin will `reject` promise here?

As for `loading.hide`, I think it should be moved as `.finally(() => loading.hide());` to here https://github.com/jellyfin/jellyfin-web/blob/a50e4853e3206934c1fcaa9a950beb476e3a416f/src/plugins/youtubePlayer/plugin.js#L201

Then this can be removed:
https://github.com/jellyfin/jellyfin-web/blob/a50e4853e3206934c1fcaa9a950beb476e3a416f/src/plugins/youtubePlayer/plugin.js#L117-L119

matjaz321: w8 so `instance.destroy()` can reject a promise?

dmitrylyzo: > w8 so `instance.destroy()` can reject a promise?

I don't think so. Try replace it with `reject();`. If it won't fix target issue then we should find correct place to destroy player (somewhere in `playbackManager`), but at least Promise will be correctly finished.

matjaz321: got it thx

matjaz321: Okei I fixed that... Should we show any error message?
So as I mentioned if video is the current UX behavior is not that nice...

oddstr13 (approved):
Looks good.
This is a private video:
image

src/components/playback/playbackmanager.js
```diff
@@ -2240,13 +2240,14 @@ class PlaybackManager {
...
+                    return player.play(streamInfo).then(() => {
                         loading.hide();
                         onPlaybackStartedFn();
                         onPlaybackStarted(player, playOptions, streamInfo);
-                    }, function () {
-                        // TODO: show error message
+                    }).catch((errorCode) => {
                         self.stop(player);
+                        loading.hide();
+                        showPlaybackInfoErrorMessage(self, errorCode || '');
```
thornbill, line 2250: Should there be a default error message here?

dmitrylyzo: `"ErrorDefault": "There was an error processing the request. Please try again later."`?

thornbill (approved):

Camc314 (approved):

Issue #1790, opened 2020-08-08
Is currentTime in ticks or ms?


iwalton3:
Right now there is a bug in the web player, which causes cast clients such as MPV Shim and Chromecast to skip forward entire episodes when using the seek buttons:

When I hit the "seek back" button:
`2524100000000 / 10000000 / 3600` = absolute seek to 70 hours

When I hit the "seek forward" button:
`152050000000 / 10000000 / 3600` = absolute seek to 4.2 hours

The root cause of this problem is the currentTime value is treated as ticks in some sections of the code, and as milliseconds in other parts. This causes inconsistent behavior, and I've seen it worked around with checks like this:

```js
if (player.isLocalPlayer) {
    return player.seekRelative((ticks || 0) / 10000);
} else {
    return player.seekRelative(ticks);
}
```

I'm not sure what units that should be in. For instance, [here are the changes](https://github.com/iwalton3/jellyfin-web/compare/master...iwalton3:currenttime) you'd need to make all instances of currentTime be in milliseconds that I could find.

dkanada:
@Maxr1998 this might be related to the changes you made for the Android client.

Maxr1998:
> @Maxr1998 this might be related to the changes you made for the Android client.

I think my PR just added a new function to see with milliseconds, should affect the other seek methods..

iwalton3:
I was testing with this more. The only issue is if I hit the button multiple times, it will only seek backwards once unless I wait for the state update to be sent back (via the /Playing/Progress call back through via the websocket) between pressings of the button.

I'm not sure if this is a bug or if it requires a design change to stack the requests. I know that the controlled client gets an absolute seek value sent to it, so the controlling client would have to deal with it, probably by updating the local PositionTicks before the action has been confirmed.

PR #1810, opened 2020-08-15
Change all instances of currentTime to be in ms.

iwalton3:
This fixes seeking on remote players so that it will seek properly instead of skipping episodes. It also removes [sections like these](https://github.com/jellyfin/jellyfin-web/blob/master/src/components/playback/playbackmanager.js#L1573) which detect the type of player.

The following has been tested:
 - Seeking while casting to MPV Shim
 - Seeking while playing video locally
 - Seeking while playing audio locally
 - Seeking while using SyncPlay ~~(note caveat 1)~~

The following has not been tested:
 - Seeking while casting to Chromecast

**Changes**
Replace all known instances of player currentTime methods to be in milliseconds.

**Issues**
Fix https://github.com/jellyfin/jellyfin-web/issues/1790

**Caveats**
~~1. The SyncPlay menu is currently broken. [This change](https://github.com/jellyfin/jellyfin-web/compare/master...iwalton3:fix-syncplay-menu) is required for it to work.~~
2. It could be more ideal for currentTime to be in ticks instead. This PR is one possible solution to the problem.

**Alternative Versions**
https://github.com/jellyfin/jellyfin-web/pull/1854 is the alternative to this PR, which uses ticks instead of ms.


dkanada:
This looks like it will indeed fix the bug, but I hate that we have to track the logic everywhere we apply ticks now. It seems odd that the issue was undiscovered until the latest version, was this always a problem with the web client?

iwalton3:
It was being worked around with checks for if the player was local. But this is the second time I’ve had these kinds of issues with seeking, so just getting rid of the discrepancy seems like a good idea.

It is also possible to change the two web players to emit ticks from their respective currentTime functions, which is another approach.

iwalton3:
To clarify, I can make a version of this that makes everything ticks instead if that is preferred. Just let me know.

I don't care if it is in ticks or milliseconds as long as it is consistent.

dkanada:
@iwalton3 let's use ticks for now to keep everything consistent with the backend. We can probably merge this sometime this weekend to get unstable working at least.

src/components/playback/playbackmanager.js
```diff
@@ -1618,11 +1618,7 @@ class PlaybackManager {
 
             player = player || self._currentPlayer;
             if (player && !enableLocalPlaylistManagement(player)) {
-                if (player.isLocalPlayer) {
-                    return player.seek((ticks || 0) / 10000);
-                } else {
-                    return player.seek(ticks);
-                }
+                return player.seek(ticks);
```
dkanada, line 1621: Moving back to ticks in the web client would mean only these two sections need custom arithmetic instead of the other ten or twelve instances.

iwalton3: The problem is the inconsistency confuses developers and results in broken behavior every so often.

What I would do if I switched to ticks is go and fix the local player so that this code is still not needed. All instances of currentTime should use the same units.

iwalton3:
I will start work this evening on fixing everything to use ticks then.

cromefire:
The problem I see is the people who don't work a lot with date-time C# (which I guess is the majority) doesn't know what ticks are. Almost everybody know how to work with ms though so that seems like the natural choice to me.

iwalton3:
https://github.com/jellyfin/jellyfin-web/pull/1854 is a simpler change that is less error-prone.

src/components/playback/playbackmanager.js
MrTimscampi: To add to this, HTMLMediaElements handle duration and position in seconds, with a double-precision floating point value for milliseconds.

So, from a web client standpoint, it makes a lot more sense to use milliseconds, since that's what everything in the JS ecosystem uses and what JS devs will be familiar with.

Imo, only the values sent and returned from the server (for now) should be in ticks. Everything passed around the web client, as soon as it arrives, should be handled in milliseconds.

src/components/nowPlayingBar/nowPlayingBar.js
```diff
@@ -701,7 +701,7 @@ import 'emby-ratingbutton';
 
         const player = this;
         currentRuntimeTicks = playbackManager.duration(player);
-        updateTimeDisplay(playbackManager.currentTime(player), currentRuntimeTicks, playbackManager.getBufferedRanges(player));
+        updateTimeDisplay(playbackManager.currentTime(player) * 10000, currentRuntimeTicks, playbackManager.getBufferedRanges(player));
```
MrTimscampi, line 704: Why not adjust `currentTime` to be in ms directly?

src/components/playback/playbackmanager.js
```diff
@@ -3219,7 +3211,7 @@ class PlaybackManager {
             return player.currentTime();
         }
 
-        return this.getCurrentTicks(player);
+        return this.getCurrentTicks(player) / 10000;
```
MrTimscampi, line 3214: Reworking `getCurrentTicks` seems better than this (Converting this to `getCurrentTime` or removing it completely)

src/components/remotecontrol/remotecontrol.js
```diff
@@ -615,7 +615,7 @@ export default function () {
             lastUpdateTime = now;
             const player = this;
             currentRuntimeTicks = playbackManager.duration(player);
-            updateTimeDisplay(playbackManager.currentTime(player), currentRuntimeTicks);
+            updateTimeDisplay(playbackManager.currentTime(player) * 10000, currentRuntimeTicks);
```
MrTimscampi, line 618: We could avoid weird arithmetic by adjusting `playbackManager.duration(player)` to return a time in ms.

src/components/syncPlay/syncPlayManager.js
```diff
@@ -741,7 +741,7 @@ class SyncPlayManager {
 
         const playAtTime = this.lastCommand.When;
 
-        const currentPositionTicks = playbackManager.currentTime();
+        const currentPositionTicks = playbackManager.currentTime() * 10000;
```
MrTimscampi, line 744: Better fix would be on the next line, since what it does is essentially compute the difference between server position and current position, then convert to ms.

So converting the `serverPositionTicks` to ms would allow us to not have to convert this, then convert the result to ms afterwards.

src/components/upnextdialog/upnextdialog.js
```diff
@@ -256,7 +256,7 @@ import 'flexStyles';
             const runtimeTicks = playbackManager.duration(options.player);
 
             if (runtimeTicks) {
-                const timeRemainingTicks = runtimeTicks - playbackManager.currentTime(options.player);
+                const timeRemainingTicks = runtimeTicks - playbackManager.currentTime(options.player) * 10000;
```
MrTimscampi, line 259: Converting playbackManager to operate in ms (preferred solution, imo) would allow us to completely drop this function.

MrTimscampi (commented):
I vastly prefer this, compared to using ticks (Since the entire JS ecosystem works in seconds/milliseconds and that's what contributors would expect).

However, this feels like a half-step. I'd rather get the client to properly work in milliseconds internally, rather than doing a weird dance around various functions to convert things.

I'm not opposed to approving this if we determine that properly doing it would be too much of a pain, though.

iwalton3:
I would be willing to look at further converting more components to ms, however I would want to see a consensus that ms is desired, as it is a ton of work to convert the entire application instead of just currentTime as this PR does.

MrTimscampi:
> I would be willing to look at further converting more components to ms, however I would want to see a consensus that ms is desired, as it is a ton of work to convert the entire application instead of just currentTime as this PR does.

Sounds good. If we get the go from @jellyfin/web that this is what we prefer, I'll approve this to fix the issues and then we can start working on converting the client properly to use ms instead of ticks.

iwalton3:
Also note that the `duration` property has a similar issue of being in ticks and ms in different places. The "ticks only" PR also solves this, but this one does not.

MrTimscampi (approved):

dkanada (approved):

PR #1811, opened 2020-08-15
Fix syncplay menu.

iwalton3:
**Changes**
Allow the SyncPlay menu to be clicked. (Currently it raises an exception due to a name change.)

**Issues**
No known issues logged.

**Caveats**
The naming or import method may not be ideal.


dkanada:
Maybe change the import so it just pulls the file normally? I'm not a fan of importing singular methods except in certain situations.

dmitrylyzo:
Single-line fix: `import * as groupSelectionMenu from 'groupSelectionMenu';`
In fact, `groupSelectionMenu` only exports `show`.

Issue #1793, opened 2020-08-08
No Seek Buttons on Mobile


iwalton3:
**Describe The Bug**
Seek buttons do not appear on iOS, on Google Chrome when presenting the iOS user agent, or when the display mode is set to "Mobile" in settings.

**Steps To Reproduce**
1. Cast media to a player on mobile.
2. Seek buttons are not shown.

**Expected Behavior**
Seek buttons are shown instead of repeat and shuffle buttons, which should be in the lower menu. (At least for video playback, maybe not music?) See this screenshot, which is taken on the desktop view:
![image](https://user-images.githubusercontent.com/8078788/89721345-7fcc8900-d9aa-11ea-81df-3bf07539b6cb.png)

**Screenshots**
![image](https://user-images.githubusercontent.com/8078788/89721330-43992880-d9aa-11ea-9e52-ee057e64a912.png)

**System (please complete the following information):**
 - Platform: iOS (iPhone 11 Pro)
 - Browser: iOS Mobile App, Mobile Safari, and Chromium with Spoofed User Agent
     - UA Used: `Mozilla/5.0 (iPhone; CPU iPhone OS 12_2 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/15E148`
     - This also happens on desktop browsers when the display mode is set to Mobile in settings.
 - Jellyfin Version: 10.6.2

**Additional Context**
This appears to be caused by this commit: https://github.com/jellyfin/jellyfin-web/commit/74d32d3cbd#diff-41e838feabde38a10ddc99b4414d2379R296
Note that these seek buttons currently do not work, and need to be fixed per this issue: https://github.com/jellyfin/jellyfin-web/issues/1790

If you set the display mode in settings on the mobile app to "Desktop", it will look like this:
![image](https://user-images.githubusercontent.com/8078788/89721543-c02d0680-d9ac-11ea-8bd0-50613ca2c1f5.png)

If I open the web app in Mobile Safari and set the zoom to 75 percent, the icons will show up correctly in Desktop mode, but they will also be too small.

iwalton3:
I tested these changes, and they seemed to solve the issue: https://github.com/jellyfin/jellyfin-web/commit/2a2408cd093048852ad94bf3d3fdf57da37341ff

It does require stylesheet and template changes, for which I am not sure if there are any unintended consequences.

PR #1812, opened 2020-08-15
Show seek buttons when playing video on mobile.

iwalton3:
**Changes**
Display seek buttons on mobile when casting video:
playing-video
Display repeat/shuffle when playing audio:
playing-audio

**Issues**
Fix https://github.com/jellyfin/jellyfin-web/issues/1793

**Caveats**
Requires https://github.com/jellyfin/jellyfin-web/pull/1810 to be useful.


src/components/remotecontrol/remotecontrol.js
```diff
@@ -332,8 +332,14 @@ export default function () {
         buttonVisible(context.querySelector('.btnNextTrack'), item != null);
         buttonVisible(context.querySelector('.btnPreviousTrack'), item != null);
         if (layoutManager.mobile) {
-            buttonVisible(context.querySelector('.btnRewind'), false);
-            buttonVisible(context.querySelector('.btnFastForward'), false);
+            const playingVideo = playbackManager.isPlayingVideo() && item != null;
+            const playingAudio = !playbackManager.isPlayingVideo() && item != null;
```
matjaz321, line 336:
```suggestion
            const playingVideo = playbackManager.isPlayingVideo() && item !== null;
            const playingAudio = !playbackManager.isPlayingVideo() && item !== null;
```

dmitrylyzo:
I turned on device mode  in Chrome (iPhone 6/7/8 Plus), buttons `Rewind` and `FastForward` don't show up.
https://github.com/jellyfin/jellyfin-web/blob/bd9250f946890c7c4fcbc6669849a5d83a587051/src/assets/css/videoosd.css#L251-L252

iwalton3:
What ends up happening is that `videoosd.css` is only ever loaded if you play video on the page locally, but it sticks around and affects styling afterwards in unrelated components.

Setup 1:
 - Reload the page.
 - Click the "cast" button and select a player.
 - The bar on the bottom appears with controls. Click it.
 - The seek buttons are visible.

Setup 2:
 - Reload the page.
 - Play a video without casting, and quit from it.
 - Click the "cast" button and select a player.
 - The bar on the bottom appears with controls. Click it.
 - The seek buttons are not visible.

The commit I have added makes it so that the `videoosd.css` doesn't affect the remote control seek buttons.

dmitrylyzo:
More testing.
I had to add this to get buttons to hide/show properly. _ChromeCast is untested_
```diff
diff --git a/src/plugins/chromecastPlayer/plugin.js b/src/plugins/chromecastPlayer/plugin.js
index b7e6d0596..5ba859608 100644
--- a/src/plugins/chromecastPlayer/plugin.js
+++ b/src/plugins/chromecastPlayer/plugin.js
@@ -931,9 +931,9 @@ class ChromecastPlayer {
         return state.VolumeLevel == null ? 100 : state.VolumeLevel;
     }
 
-    isPlaying() {
+    isPlaying(mediaType) {
         const state = this.lastPlayerData || {};
-        return state.NowPlayingItem != null;
+        return state.NowPlayingItem != null && (state.NowPlayingItem.MediaType === mediaType || !mediaType);
     }
 
     isPlayingVideo() {
diff --git a/src/plugins/sessionPlayer/plugin.js b/src/plugins/sessionPlayer/plugin.js
index cbeb6f34b..7e5dc9f4f 100644
--- a/src/plugins/sessionPlayer/plugin.js
+++ b/src/plugins/sessionPlayer/plugin.js
@@ -466,9 +466,9 @@ class SessionPlayer {
         sendCommandByName(this, 'DisplayContent', options);
     }
 
-    isPlaying() {
+    isPlaying(mediaType) {
         const state = this.lastPlayerData || {};
-        return state.NowPlayingItem != null;
+        return state.NowPlayingItem != null && (state.NowPlayingItem.MediaType === mediaType || !mediaType);
     }
 
     isPlayingVideo() {
```
because `playbackManager` passes `mediaType` to check `isPlayingVideo`
https://github.com/jellyfin/jellyfin-web/blob/e9e722f15761b0c644ec5a3571302e63817b0309/src/components/playback/playbackmanager.js#L992-L996

Also, sometimes (after end of video or just if nothing is playing) I get runtime error here https://github.com/jellyfin/jellyfin-web/blob/e9e722f15761b0c644ec5a3571302e63817b0309/src/components/remotecontrol/remotecontrol.js#L138
`item` is `null` and `typeof item !== 'undefined'` is `true` (`typeof item` is `'object'`).
So maybe just change it to `if (item)`?

src/assets/css/videoosd.css
```diff
@@ -248,8 +248,8 @@
 }
 
 @media all and (max-width: 30em) {
-    .btnFastForward,
-    .btnRewind,
+    .osdControls .btnFastForward,
+    .osdControls .btnRewind,
```
dmitrylyzo, line 252: These lines are probably not required because of this https://github.com/jellyfin/jellyfin-web/blob/bd9250f946890c7c4fcbc6669849a5d83a587051/src/assets/css/videoosd.css#L272-L277
But no harm.

src/components/remotecontrol/remotecontrol.js
```diff
@@ -134,7 +134,7 @@ function imageUrl(item, options) {
 function updateNowPlayingInfo(context, state, serverId) {
     const item = state.NowPlayingItem;
     const displayName = item ? getNowPlayingNameHtml(item).replace('<br/>', ' - ') : '';
-    if (typeof item !== 'undefined') {
+    if (typeof item !== 'undefined' && item !== null) {
```
dmitrylyzo, line 137: Why so complicated?
```suggestion
    if (item) {
```

dmitrylyzo (approved):
`sessionPlayer` is still inoperable because it doesn't implement some getters, but this is for another PR.

src/components/remotecontrol/remotecontrol.js
iwalton3: I left a stricter check there because I don’t know why they checked for undefined only initially. It can be switched to the shorter version if there is no concern for unexpected behavior.

dmitrylyzo: Look at line above (this is legacy code): https://github.com/jellyfin/jellyfin-web/blob/7eff82d671d2df4c65988e229984af98a15189b8/src/components/remotecontrol/remotecontrol.js#L136
Stricter check was added #1056

iwalton3: The commit that added it: https://github.com/jellyfin/jellyfin-web/commit/3ab01d471fa62a645bfd241484de9ebcfbf654aa#diff-41e838feabde38a10ddc99b4414d2379

dmitrylyzo: This is another old code (now it is inside that condition) and it also uses simple `(item)` (in fact, every test in function `updateNowPlayingInfo` is `(item)` until that PR):
https://github.com/jellyfin/jellyfin-web/blob/7eff82d671d2df4c65988e229984af98a15189b8/src/components/remotecontrol/remotecontrol.js#L231-L242
btw, `else` should be related to that `if`.

I see no harm in using `(item)` - `item` is supposed to be an object.

iwalton3:
> `sessionPlayer` is still inoperable because it doesn't implement some getters, but this is for another PR.

Please elaborate on this. The last thing I want to do is break the session player.

neopc10:
A quick note regarding the seek buttons on Chromecast on 10.6.3: they function differently on the stable receiver client and on the unstable/nightly CAF. 

Previously on the Stable build, skipping around on Chromecasted content would start a second transcode and load to the desired time. On the CAF it'll stick with the first transcode, but now the remotecontrol seek back and forward buttons will return to the start and seek forward will go to the last second.

This is just something to keep in mind--when this gets tested for the Chromecast, do test with the unstable CAF and not the stable one from years ago. Thanks!

iwalton3:
I can confirm that the buttons appear and are fully functional in remote playback mode when this PR and #1810 are tested.

(Without #1810 both of the seek buttons will skip to the next episode instead of seeking due to an off-by-10000 error.)

neopc10:
I don't know if this is related specifically to sessionplayer but when casting to the unstable/CAF the remotecontrol isn't "hooked up" to what's casting (usually just with video) until the page is refreshed. So, it doesn't display how much time as elapsed and the seekbar doesn't do anything. I suspect that Jellyfin's "resume play" not working and starting from the beginning every time is related to that.

But again, I am not sure if this is entirely relevant.

dmitrylyzo:
> Please elaborate on this. The last thing I want to do is break the session player.

Don't worry - it is just unfinished.
https://github.com/jellyfin/jellyfin-web/blob/1a360a6f72b0383320756d672d2dee8e10529e1f/src/plugins/sessionPlayer/plugin.js#L453-L454
https://github.com/jellyfin/jellyfin-web/blob/1a360a6f72b0383320756d672d2dee8e10529e1f/src/plugins/sessionPlayer/plugin.js#L462-L463

Code smell is what I said: `else` should be moved out as continuous of fixed `if`: https://github.com/jellyfin/jellyfin-web/blob/e9e722f15761b0c644ec5a3571302e63817b0309/src/components/remotecontrol/remotecontrol.js#L137
_^ tested - OK but almost useless - cover is still there after stop music (clear queue)._

But what I think: if we want to fix that runtime error in 10.6, it should be done in separate PR (and that `if` too) to be able to cherry-pick.
Or just release 10.7 quicker :smile: 
cc @jellyfin/web

dmitrylyzo (approved):

dkanada (approved):

PR #1814, opened 2020-08-17
Fix download/edit metadata of specific version

danieladov:
Currently, download function only downloads the primary version, with this fix it will download the selected version. The same happens when trying to edit metadata. Even if version 2 is selected you can't edit it.


MrTimscampi (approved):

MrTimscampi:
Linting issue:

/home/vsts/work/1/s/src/controllers/itemDetails/index.js
  1940:59  error  A space is required after ','  comma-spacing

PR #1825, opened 2020-08-20
Redesign OSD and Up Next dialog

MrTimscampi:
**Changes**

Modernizes the video OSD and reduces the size of the Up Next dialog.

Also constrains the controls to a 16:9 ratio, in order to prevent ultrawide screens from being a pain to use.

Screenshot_2020-08-20 Jellyfin(1)

Screenshot_2020-08-20 Jellyfin

**Issues**


MrTimscampi:
Setting it back to draft since I need to mess with the mobile layout, still.

Camc314 (commented):
Overall looks really good 👍.

When on mobile, the OSD looks a bit weird because it is limited to 177vh. On mobile have the osd 100% of the width?

DMouse10462:
What does the OSD look like with a completely white video frame? The buttons over by the shirt on the right are a bit hard to "read," especially the fullscreen button. The transparent look is pretty cool, but some additional shadow (and maybe larger icons on desktop) could help the contrast a bit. Here's Netflix on a white frame:

Screenshot from 2020-08-24 18-39-57

They're certainly not going to be WCAG AA compliant unless you get super dark, and I know that tends to be a common problem with transparent controls (look at the episode name on that Netflix screenshot above - eek), but it would help to have some more contrast in these worst-case scenarios.

I'd personally love to see support for a high-contrast or fully WCAG compliant theme for these reasons. I'd be happy to work on some of that, but it's probably best saved for after a lot of the ES6/Vue stuff is complete.

Darkwolf1515:
I think all the changes are fantastic, bar the lack of poster, should be a toggle if possible.

MrTimscampi:
> What does the OSD look like with a completely white video frame? The buttons over by the shirt on the right are a bit hard to "read," especially the fullscreen button.

image

Seems like it's currently failing the WCAG contrast test with a 3.4:1, so I'll change it up a bit to get to a 4.5:1, in order to get at least a WCAG AA rating on normal text.

MrTimscampi:
> I think all the changes are fantastic, bar the lack of poster, should be a toggle if possible.

From a quick test, adding back the poster messes up a bunch of things (most importantly, it messes up with the width limit, which I value more than the poster).

I'm not into having the poster in the OSD, so I'll leave that for someone else who's motivated about it to bring it back as a toggle.

In the mean time, with the color fixed to get this a WCAG AA rating, I consider this ready for reviews.

src/assets/css/videoosd.css
```diff
@@ -6,29 +6,42 @@
...
-    -webkit-transition: opacity 0.3s ease-out;
-    -o-transition: opacity 0.3s ease-out;
     transition: opacity 0.3s ease-out;
     position: relative;
     z-index: 1;
-    background: rgba(0, 0, 0, 0.7) !important;
-    -webkit-backdrop-filter: none !important;
-    backdrop-filter: none !important;
-    color: #eee !important;
+    background: linear-gradient(180deg, rgba(16, 16, 16, 0.75) 0%, rgba(16, 16, 16, 0) 100%);
```
Camc314, line 31: The change to a gradient is making the following visible and causing the top bar on the OSD to be grey.

https://github.com/jellyfin/jellyfin-web/blob/507b4771173edaad6eb1faeedb3b167b84ca6463/src/themes/dark/theme.css#L23-L25

The style in theme.css should be removed?

MrTimscampi: Hm, I had fixed it by overriding it with a more precise selector. Might've accidentally stashed the changes, let me check.

MrTimscampi: There, it's fixed :)

Camc314 (approved):

dkanada (approved):

PR #1826, opened 2020-08-20
Removed the individual app info methods

cromefire:
**Changes**

Removed the individual app info methods in favor of the async init method

**Issues**


MrTimscampi (approved):

thornbill:
I’m not sure I follow why these need to be removed now when we plan to fully overhaul the NativeShell interface at some point.

cromefire:
Because I need a proper, clean promise based solution now, pretty much

dkanada:
This will break compatibility with older apps so we can't merge it without further changes. If both options were left in this would be fine, but since the NativeShell redesign is close we can just change everything at that point.

PR #1832, opened 2020-08-20
Fix cards without title

MrTimscampi:
**Changes**

Fixes the weird behavior of cards without titles having a buggy overlay.

**Issues**

Fixes #1829 629


src/components/cardbuilder/cardBuilder.js
```diff
@@ -986,6 +986,10 @@ import 'programStyles';
                 lines = [];
             }
 
+            if (overlayText && showTitle) {
```
thornbill, line 989: I don't really follow what is happening here, but it looks like this _could_ conflict with the logic on the lines above... should this be an `else if`?

MrTimscampi: I haven't found a place where a conflict happens. The issue here is that, when titles are disabled, cardBuilder tries to add a `<button>` into the overlay of the cards, which somehow messes up the DOM and makes the overlay move out to the root of the card, causing an issue where the overlays will take the entire size of the screen.

This basically tells it that if we want an overlay with the title, we explicitly want it to show the item's name in there, not a `<button>`.

I'm not sure what is supposed to trigger the condition above, though, as I haven't been able to find any example.

PR #1854, opened 2020-08-28
Change all instances of currentTime and duration to be in ticks.

iwalton3:
This fixes seeking on remote players so that it will seek properly instead of skipping episodes. It also removes [sections like these](https://github.com/jellyfin/jellyfin-web/blob/master/src/components/playback/playbackmanager.js#L1621) which detect the type of player.

The following has been tested:
 - Seeking while casting to MPV Shim
 - Seeking while playing video locally
 - Seeking while playing audio locally
 - Seeking while using SyncPlay (note caveat 1)

The following has not been tested:
 - Seeking while casting to Chromecast

**Changes**
Replace all known instances of player currentTime methods to be in ticks. Duration is also changed to ticks.

**Issues**
Fix https://github.com/jellyfin/jellyfin-web/issues/1790

**Caveats**
1. Creating a SyncPlay session from the web client is currently broken, but joining one works. This PR is not the cause.
2. It could be more ideal for currentTime to be in ms instead. This PR is one possible solution to the problem.

**Alternative Versions**
https://github.com/jellyfin/jellyfin-web/pull/1810 is the alternative to this PR, which uses ms instead of ticks.

### DIFF
--- a/src/plugins/htmlAudioPlayer/plugin.js
+++ b/src/plugins/htmlAudioPlayer/plugin.js
@@ -347,16 +347,16 @@ class HtmlAudioPlayer {
         const mediaElement = this._mediaElement;
         if (mediaElement) {
             if (val != null) {
-                mediaElement.currentTime = val / 1000;
+                mediaElement.currentTime = val / 10000000;
                 return;
             }
 
             const currentTime = this._currentTime;
             if (currentTime) {
-                return currentTime * 1000;
+                return currentTime * 10000000;
             }
 
-            return (mediaElement.currentTime || 0) * 1000;
+            return (mediaElement.currentTime || 0) * 10000000;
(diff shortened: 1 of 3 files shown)


src/plugins/htmlAudioPlayer/plugin.js
```diff
@@ -347,16 +347,16 @@ class HtmlAudioPlayer {
         const mediaElement = this._mediaElement;
         if (mediaElement) {
             if (val != null) {
-                mediaElement.currentTime = val / 1000;
+                mediaElement.currentTime = val / 10000000;
```
dkanada, line 350: I love that we have to do this since web doesn't use ticks.

dkanada (approved):

PR #1855, opened 2020-08-29
Fix bad global restriction

MrTimscampi:
**Changes**

Fix some `self` instances wrongly "fixed" to `window`.

**Issues**

### DIFF
--- a/src/scripts/apploader.js
+++ b/src/scripts/apploader.js
@@ -35,7 +35,8 @@
         // Promise() being missing on some legacy browser, and a funky one
         // is Promise() present but buggy on WebOS 2
         window.Promise = undefined;
-        window.Promise = undefined;
+        /* eslint-disable-next-line no-restricted-globals -- Explicit check on self needed */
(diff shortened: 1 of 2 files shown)


src/scripts/apploader.js
```diff
@@ -35,7 +35,8 @@
         // Promise() being missing on some legacy browser, and a funky one
         // is Promise() present but buggy on WebOS 2
         window.Promise = undefined;
-        window.Promise = undefined;
+        /* eslint-disable-next-line no-restricted-globals -- Explicit check on self needed */
+        self.Promise = undefined;
```
dmitrylyzo, line 39: I suppose `apploader.js` isn't used by WebWorkers. So `self` will always be `window` here.

MrTimscampi: It was mainly a security, as I'm not sure why @JustAMan did that initially.

dmitrylyzo (approved):

dkanada (approved):

Issue #1210, opened 2020-05-09
Playback requests are not configured to send credentials


fhriley:
When an identity provider is in front of jellyfin, a cookie needs to be sent with every request. The playback code is not configured to send credentials so the cookie is not sent. See https://github.com/jellyfin/jellyfin/issues/3083 for more detail.

PR #1856, opened 2020-08-29
Remove withCredentials from xhrSetup in hls.js

MrTimscampi:
**Changes**

This prevents playback from failing, by not sending credentials with the manifest request on playback.

The request contains an API key, so credentials are not needed.

It prevents the following error from happening:
```
The value of the 'Access-Control-Allow-Origin' header in the response must not be the wildcard '*' when the request's credentials mode is 'include'. The credentials mode of requests initiated by the XMLHttpRequest is controlled by the withCredentials attribute.
```

**Issues**

### DIFF
--- a/src/plugins/htmlAudioPlayer/plugin.js
+++ b/src/plugins/htmlAudioPlayer/plugin.js
@@ -132,10 +132,7 @@ class HtmlAudioPlayer {
                 return new Promise(function (resolve, reject) {
                     requireHlsPlayer(function () {
                         const hls = new Hls({
-                            manifestLoadingTimeOut: 20000,
-                            xhrSetup: function (xhr, url) {
-                                xhr.withCredentials = true;
-                            }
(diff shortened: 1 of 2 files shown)


dkanada (approved):

dmitrylyzo:
That was added in #1210
Has something changed in backend that it is no longer required?

MrTimscampi:
> That was added in #1210
> Has something changed in backend that it is no longer required?

The server properly supports CORS now. So that code is preventing playback on the current master when the `Access-Control-Allow-Origin` header is set to `*`.

MrTimscampi:
Imo, #1210 is a weird use-case. That person should really be using LDAP instead to handle authentication.

joshuaboniface:
This isn't needed in stable as we (a) have an older version of HLS, and (b) the CORS header was properly set to $host there.

PR #1807, opened 2020-08-14
Finish migration to ES6

MrTimscampi:
**Changes**

Very much not done, likely very broken.

Here be dragons.

**Issues**

Needs the remaining ES6 PRs merged to cleanup history a bit.

PR #1859, opened 2020-08-30
Use connectionManager global

MrTimscampi:
**Changes**

Extracting some changes from #1807 

This removes the weird `connectionManager` module defined in `site.js` and uses the already registered `window.connectionManager` global everywhere instead.

Also renames said global from `window.ConnectionManager` to `window.connectionManager`, since it's not a constructor and should follow proper capitalization practices to not be more confusing than it needs to be.

**Issues**


Camc314 (approved):

dkanada (approved):
that was some messy code

PR #1866, opened 2020-08-30
Migrate from ticks to milliseconds.

iwalton3:
Per the decision on https://github.com/jellyfin/jellyfin-web/issues/1790, this PR is the work in progress to use milliseconds instead of ticks natively in the web client, wherever possible.

**Changes**
 - Make time conversions explicit.

### DIFF
--- a/package.json
+++ b/package.json
@@ -336,6 +336,7 @@
           "src/scripts/fileDownloader.js",
           "src/scripts/filesystem.js",
           "src/scripts/globalize.js",
+          "src/scripts/timeConversions.js",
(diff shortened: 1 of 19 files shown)


iwalton3:
I think with the ES6 conversion out of the way this might be more practical to do now, but it likely won't be a focus with the MPV Shim work ahead.

neopc10:
Would this make the rewind/fast forward buttons in the remotecontrol work for Google Cast? Currently, the rewind/fast forward buttons will skip to the beginning or end of the currently playing track.

PR #1869, opened 2020-08-31
Handle no matches on iOS version regex

thornbill:
**Changes**
This is a quick fix to avoid a crash if this method is called on iPads running new versions of iOS that match on "MacIntel" but do not report the version in the expected way. This _probably_ won't happen currently since we don't detect them as running iOS, but it's probably best to just add a safety check here regardless.

**Issues**
N/A

### DIFF
--- a/src/scripts/browser.js
+++ b/src/scripts/browser.js
@@ -77,7 +77,9 @@ function iOSversion() {
     if (/iP(hone|od|ad)|MacIntel/.test(navigator.platform)) {
         // supports iOS 2.0 and later: <http://bit.ly/TJjs1V>
         const v = (navigator.appVersion).match(/OS (\d+)_(\d+)_?(\d+)?/);
-        return [parseInt(v[1], 10), parseInt(v[2], 10), parseInt(v[3] || 0, 10)];
+        if (v) {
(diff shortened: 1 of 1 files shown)


dkanada:
Should we backport this?

MrTimscampi (approved):

dmitrylyzo:
:innocent: https://github.com/jellyfin/jellyfin-web/pull/577#discussion_r346162865

MrTimscampi:
We could also simply steal this: https://github.com/google/shaka-player/blob/606457488f77e9cf5cac2c82e94bb4ce35d50e2a/lib/util/platform.js#L184

thornbill:
@MrTimscampi I was looking at doing something very similar, but their comments do answer one question I had lol

thornbill:
@dkanada probably doesn't deserve a backport since it is a condition that we shouldn't encounter as it is currently used... (we don't detect these new versions on iPads as iOS so we shouldn't call this function)

thornbill:
I'm going to close this in favor of having a method that can actually retrieve the iOS version on iPads with recent iOS versions.

thornbill:
New PR is here which fixes version detection and adds the workaround for detecting iPads https://github.com/jellyfin/jellyfin-web/pull/1880

PR #1893, opened 2020-09-01
Fix old iOS devices that do not support AC-3

thornbill:
**Changes**
Old iPhone and iPad devices do not support AC-3 and E-AC-3. See this [support matrix](https://developer.dolby.com/platforms/apple/ios/device-support/). Testing with `canPlayType` reports "probably"/"maybe" for support of these codecs. Which is not a good indicator of if the device actually supports playback since devices that do support it only report as "probably". Based on this [max iOS support matrix](https://iosref.com/ios), it looks like we can just check for iOS versions < 11.

**Issues**
N/A

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -70,12 +70,24 @@ define(['browser'], function (browser) {
             return true;
         }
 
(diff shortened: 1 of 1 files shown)


MrTimscampi (approved):
LGTM

PR #1894, opened 2020-09-01
Fix webOS 1.2

dmitrylyzo:
**Changes**
- Exclude `libarchive` from Babel exclude list.
- Add timeout for polyfilled CustomElements.

**Issues**
- `bundle.js` contains `class` (from `libarchive`).
- Runtime errors on video OSD and remotecontrol pages. `CustomElements` is async polyfilled (`MutationObserver`?).

Alternatively, we could call `window.CustomElements.upgradeSubtree(view)` on each page.

### DIFF
--- a/src/components/viewContainer.js
+++ b/src/components/viewContainer.js
@@ -36,54 +36,56 @@ import 'css!components/viewManager/viewContainer';
             const newViewInfo = normalizeNewView(options, isPluginpage);
             const newView = newViewInfo.elem;
 
-            return new Promise((resolve) => {
-                const currentPage = allPages[pageIndex];
+            const currentPage = allPages[pageIndex];
 
-                if (currentPage) {
-                    triggerDestroy(currentPage);
-                }
+            if (currentPage) {
+                triggerDestroy(currentPage);
+            }
 
-                let view = newView;
+            let view = newView;
 
-                if (typeof view == 'string') {
-                    view = document.createElement('div');
-                    view.innerHTML = newView;
-                }
+            if (typeof view == 'string') {
+                view = document.createElement('div');
+                view.innerHTML = newView;
+            }
 
-                view.classList.add('mainAnimatedPage');
+            view.classList.add('mainAnimatedPage');
 
-                if (currentPage) {
-                    if (newViewInfo.hasScript && window.$) {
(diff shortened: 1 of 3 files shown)


src/components/viewContainer.js
```diff
@@ -36,53 +36,56 @@ import 'css!components/viewManager/viewContainer';
...
-                setControllerClass(view, options).then(() => {
+            if (properties.length) {
+                view.setAttribute('data-properties', properties.join(','));
+            }
+
+            allPages[pageIndex] = view;
+
+            return setControllerClass(view, options)
+                // Timeout for polyfilled CustomElements (webOS 1.2)
+                .then(() => new Promise((resolve) => setTimeout(resolve, 0)))
```
thornbill, line 87: This seems pretty hacky. Is there not a better way to delay until the polyfill is ready?

dmitrylyzo: Need to wait for upgrade of elements (not the polyfill itself). Didn't find anything good.

Without timeout, it looks like:
```
set innerHTML/appendChild
call view constructor
    access to a not yet upgraded element
MutationObserver upgrades elements
```
With timeout, it is presumably:
```
set innerHTML/appendChild
schedule call of view constructor via zero timeout
MutationObserver upgrades elements
[timeout] call view constructor
    access to an upgraded element
```

As I said, we can use `window.CustomElements.upgradeSubtree(view, true)` (`true` to call `attachedCallback`; _tried - works for current issue_):
```js
return setControllerClass(view, options)
    .then(() => {
        // Polyfilled CustomElements (webOS 1.2, webOS 2, Firefox 59+) is implemented
        // via MutationObserver. Therefore, it takes time to upgrade the added elements.
        // But we can just force it to upgrade.
        window.CustomElements.upgradeSubtree(view, true);

        if (onBeforeChange) {
            onBeforeChange(view, false, options);
        }
```

**Another option**: add timeouts or `upgradeSubtree` in place.
https://github.com/jellyfin/jellyfin-web/blob/2e6211ce31b3966c19c461b9269d9478f87b08e4/src/components/subtitlesync/subtitlesync.js#L24-L27
_This is the only (known) place where Firefox fails because it also uses `webcomponents.js`_

thornbill (approved):

dkanada (approved):

PR #1903, opened 2020-09-02
Return permanent scrollbar

### LINKED ISSUE #1965: videoOsdBottom vanished after switching video quality or subtitles
**Describe The Bug**
videoOsdBottom from the [redesigned OSD](https://github.com/jellyfin/jellyfin-web/pull/1825/) vanished after switching video quality or subtitles.

**Steps To Reproduce**
1. Play a movie.
2. Change the video quality to make the play method changed.
     Or choose a PGS graphical sub to enable transcoding.
3. The bottom of OSD vanished once the video get reloaded.
4. Cannot control the playback by using keyboard.

**Expected Behavior**
OSD will not vanish.

**Logs**
Not observed any error in console.

**Screenshots**

osd-bottom

 - Platform: [Linux, Windows, iPhone]
 - Browser: [Firefox, Chrome, Safari]
 - Jellyfin Version: [10.7.0 unstable 20201002.5]

### LINKED ISSUE #2090: [Unstable] OSD becomes disabled when switching audio tracks
The OSD will disappear when selecting a different audio track. Also had an odd effect where it would momentarily show the first frame of the video while it loaded the new track. All Direct Play, no FFmpeg invoked.

dmitrylyzo:
Permanent scrollbar (#1229) was killed by refactoring #1502

**Changes**
- Return permanent scrollbar.
- Fix scrollbar (revived) for Youtube player.

Since `skinManager.js` has been deleted, instead of https://github.com/jellyfin/jellyfin-web/blob/f25728eca4f12596f3b989a1ec747ce6b52633ce/src/components/skinManager.js#L161-L163
~I set `overflow-y: scroll`~ I permanently add `force-scroll` to `body` on theme load in `autoThemes` and turn off it by existed `hide-scroll`

This https://github.com/jellyfin/jellyfin-web/blob/f25728eca4f12596f3b989a1ec747ce6b52633ce/src/components/skinManager.js#L140-L144
is replaced by class manupulation in `htmlVideoPlayer`.

:warning: Probably need to test book "player".

~:warning: Screensaver has scrollbar. Probably we should hide it for all fullscreen dialogs.~
:ballot_box_with_check: `dialogHelper` has the ability to hide the scrollbar (now fixed).
:ballot_box_with_check: ScreenSaverManager now hides scrollbar.

**Issues**
Fixes #1965
Fixes #2090


dkanada:
> Probably we should hide it for all fullscreen dialogs.

Most definitely yes, the EPUB player actually has this issue on desktop.

dmitrylyzo:
> Most definitely yes, the EPUB player actually has this issue on desktop.

The fun thing is that there is a disabling of scrollbar (`data-lockscroll`), but it is not applied due to the fact that browser supports `overscroll-behavior`.
https://github.com/jellyfin/jellyfin-web/blob/c5223cbc80af05fdbbb3fb1919917181e826dbff/src/components/dialogHelper/dialogHelper.js#L312-L330

dkanada:
I feel like if we disable scroll it should get disabled no matter what the browser supports.

dmitrylyzo:
> I feel like if we disable scroll it should get disabled no matter what the browser supports.

Agree. I moved that condition lower. It is probably a fix for some browser quirk.

dmitrylyzo:
I've removed hiding the pre-playback page (#1213) because scrollbar is now controlled by CSS.
- This fixes regression I introduced :sweat_smile: (in #1951):
[this](https://github.com/dmitrylyzo/jellyfin-web/blob/f9579e6f057702648c861d58c055bbb4110db8f4/src/plugins/htmlVideoPlayer/plugin.js#L1344-L1346) is for the case when playback starts from a page with an animated background.
But it also hides the video OSD when switching to the next item in the playlist.
- Also it fixes the restoration of scrollbar position on `back` when starting playback from the library (list), which was scrolled down.

I have to remove `z-index` from `alphaPicker` - it stays at the top until playback actually starts. At current moment, `alphaPicker` is placed in the padding zone of `itemContainer` and doesn't overlap anything. It is still clickable, so I don't see the need for `z-index`.

dmitrylyzo:
:disappointed: - _me looking at this PR_
Toggling scrollbar doesn't look good for screensaver and slideshow/dialogs (slow fade out): you can see the header and backdrop move.

It is better to have a scrollable container so that we could place fullscreen elements on top of it without juggling the scrollbar state. But this will complicate scrolling restoration when working with browser history.

NoyTheBiche (approved):

src/plugins/backdropScreensaver/plugin.js
```diff
@@ -42,6 +42,7 @@ class BackdropScreensaver {
                 this.currentSlideshow.hide();
                 this.currentSlideshow = null;
             }
+            return Promise.resolve();
```
ferferga, line 45: Why?

dmitrylyzo: I expect `Promise` here https://github.com/jellyfin/jellyfin-web/blob/517b69550e5c14931c816391b1fe8f7680cb6bf6/src/libraries/screensavermanager.js#L75-L77
but cannot do much in `BackdropScreensaver`.

At the same time, `BackdropScreensaver` also controls scrollbar state (`dialogHelper` to be precise). That's why I added an extra CSS class for screensaver manager. When screensaver manager removes its own class, class from `dialogHelper` remains and hides scrollbar until dialog (slideshow) disappears.

ferferga (approved):

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -103,11 +103,9 @@ function tryRemoveElement(elem) {
         });
     }
 
-    function hidePrePlaybackPage() {
-        const animatedPage = document.querySelector('.page:not(.hide)');
-        animatedPage.classList.add('hide');
+    function hideScroll() {
         // At this point, we must hide the scrollbar placeholder, so it's not being displayed while the item is being loaded
-        document.body.classList.remove('force-scroll');
+        document.body.classList.add('hide-scroll');
```
thornbill, line 108: It's probably better to inline this instead of having a one line function.

src/plugins/youtubePlayer/plugin.js
```diff
@@ -15,6 +15,10 @@ function zoomIn(elem, iterations) {
     return elem.animate(keyframes, timing);
 }
 
+function hideScroll() {
+    document.body.classList.add('hide-scroll');
```
thornbill, line 19: Same thing here. :smile:

src/plugins/htmlVideoPlayer/plugin.js
dmitrylyzo: It gets called twice for each file and tells you clearly what it is doing.

thornbill (approved):

Motophan:
@thornbill thanks fam, as a weeb im dying here because it auto pics subs and osd dissapears.

PR #1909, opened 2020-09-07
Add Auto Cast Feature 

iwalton3:
This adds an "Auto Cast" feature. If you would like to automatically cast to a player such as MPV Shim, you can set it as the default. This will cause the web client to automatically cast to that player on load. The setting is per-browser and is remembered using LocalStorage.

image

**Changes**
- Add auto cast feature.
- Ensure menu is rendered, even before user info is received.
- Ensure the correct cast info is always displayed.


package.json
```diff
@@ -177,6 +177,7 @@
           "src/components/remotecontrol/remotecontrol.js",
           "src/components/sanatizefilename.js",
           "src/components/scrollManager.js",
+          "src/components/autocast.js",
```
dkanada, line 180: I think this makes more sense in scripts for now. Components will hopefully be only UI components at some point.

dkanada:
The linting issues can be ignored and have been fixed in another pull request.

src/scripts/autocast.js
```diff
@@ -0,0 +1,49 @@
...
+    return typeof(Storage) !== 'undefined';
+}
+
+export function enable(enabled) {
+    if (!supported()) return;
+
+    if (enabled) {
+        const currentPlayerInfo = playbackManager.getPlayerInfo();
+
+        if (currentPlayerInfo && currentPlayerInfo.id && currentPlayerInfo.id) {
```
dkanada, line 14:
```suggestion
        if (currentPlayerInfo && currentPlayerInfo.id) {
```

dkanada (approved):
LGTM

src/scripts/libraryMenu.js
```diff
@@ -910,6 +911,12 @@ import 'flexStyles';
         }
     }
 
+    function ensureHeader() {
+        return new Promise(function (resolve) {
+            window.connectionManager.user(getCurrentApiClient()).then(updateUserInHeader).then(resolve);
+        });
+    }
+
```
dmitrylyzo, line 919: unused?

src/components/playback/playerSelectionMenu.js
```diff
@@ -219,6 +220,16 @@ function showActivePlayerMenuInternal(dialogHelper, playerInfo) {
...
+    html += '</div><div>';
+
+    if (autocast.supported()) {
+        html += '<label class="checkboxContainer">';
+        var checkedHtmlAC = autocast.isEnabled() ? ' checked' : '';
+        html += '<input type="checkbox" is="emby-checkbox" class="chkAutoCast"' + checkedHtmlAC + '/>';
+        html += '<span>' + globalize.translate('EnableAutoCast') + '</span>';
+        html += '</label>';
+    }
+
```
dmitrylyzo, line 232: _Picky mode activated :)_
If not supported there will be an empty `div`.

dmitrylyzo (approved):
LGTM

PR #1918, opened 2020-09-10
Fix startup wizard redirect and standalone mode

dkanada:


src/scripts/clientUtils.js
```diff
@@ -1,35 +1,41 @@
...
-    if (index != -1) {
-        return urlLower.substring(0, index);
-    }
+    if (current) return Promise.resolve(current);
 
-    const loc = window.location;
-    let address = loc.protocol + '//' + loc.hostname;
+    let urls = [];
+    urls.push(`${window.location.origin}/System/Info/Public`);
+    urls.push(`${window.location.protocol}//${window.location.hostname}:8096/System/Info/Public`);
```
dmitrylyzo, line 25: hardcoded port? :confused:

src/scripts/clientUtils.js
```diff
@@ -49,16 +55,9 @@ export function onServerChanged(userId, accessToken, apiClient) {
 
 export function logout() {
     window.connectionManager.logout().then(function () {
-        let loginPage;
-
-        if (AppInfo.isNativeApp) {
-            loginPage = 'selectserver.html';
-            window.ApiClient = null;
```
dmitrylyzo, line 56: Shouldn't `ApiClient` be nulled?

src/scripts/site.js
```diff
@@ -83,40 +83,38 @@ function initClient() {
...
-
-                        var apiClient = new apiClientFactory(Dashboard.serverAddress(), appHost.appName(), appHost.appVersion(), appHost.deviceName(), appHost.deviceId());
-
-                        apiClient.enableAutomaticNetworking = false;
-                        apiClient.manualAddressOnly = true;
-
-                        window.connectionManager.addApiClient(apiClient);
+                    console.debug('creating apiclient singleton');
+                    let parts = server.split('/');
+                    let url = parts[0] + '//' + parts[2];
```
dmitrylyzo, line 106: How about BaseUrl?

src/components/appRouter.js
```diff
@@ -510,42 +504,50 @@ class AppRouter {
...
 
-            if (firstResult.State !== 'SignedIn' && !route.anonymous) {
-                this.handleConnectionResult(firstResult);
-                return;
-            }
+        this.firstConnectionResult = null;
+        if (firstResult && firstResult.State === 'ServerSignIn' && !route.anonymous) {
+            let url = ApiClient.serverAddress() + '/System/Info/Public';
+            fetch(url).then(response => {
+                if (!response.ok) return;
```
dmitrylyzo, line 512: `return Promise.reject('fetch failed')`?

src/scripts/clientUtils.js
dkanada: Not hardcoded so much as a "sane default" to search for common server locations. The alternatives would be not searching at all, or searching every single port on the origin to check for servers.

src/scripts/site.js
dkanada: Hmmm good question, I think I'll need to figure something out for that. I think we can safely assume that someone would include the base URL in their original page load, which helps a lot.

src/scripts/clientUtils.js
dkanada: I don't believe so, it seems to work without this.

src/scripts/clientUtils.js
joshuaboniface: But can it be replaced easily? Sane default out of a config file would be nicer than in code IMO. Would then be scalable to automate the creation of multiple frontends if the remote port was altered.

Legogris: How about putting it in config or controllable via env var on server? Most common use-case scenario is probably for coupling with a self-managed Jellyfin server.

src/components/appRouter.js
```diff
@@ -510,42 +504,50 @@ class AppRouter {
 
     authenticate(ctx, route, callback) {
```
dmitrylyzo, line 505: I can't reach wizard.

This function is called 3 times.

1st, `ServerSignIn` but `data.StartupWizardCompleted` is `false` -> navigate `wizardstart.html`.
https://github.com/jellyfin/jellyfin-web/blob/982ae411c59cb95421ea870b0dd1726662fe15d5/src/components/appRouter.js#L516
But then: https://github.com/jellyfin/jellyfin-web/blob/982ae411c59cb95421ea870b0dd1726662fe15d5/src/components/appRouter.js#L533

2nd and 3rd, `this.firstConnectionResult` is `null`.

joshuaboniface (approved):

src/scripts/clientUtils.js
dkanada: You can specify any number of servers in the config file and they will all get queried as well. This is to accomodate the most standard configuration.

src/components/appRouter.js
dkanada: Hmmm it seems you're right, I must have made changes after I got it working. Do you have any suggestions?

dmitrylyzo: I haven't tried, but shouldn't we "return" from this block? 
https://github.com/jellyfin/jellyfin-web/blob/fa1483d3c0b9e860508e7e26f1fc33f2e95acffc/src/components/appRouter.js#L509-L523

src/scripts/clientUtils.js
```diff
@@ -1,35 +1,42 @@
...
-        return urlLower.substring(0, index);
-    }
+    if (current) return Promise.resolve(current);
 
-    const loc = window.location;
-    let address = loc.protocol + '//' + loc.hostname;
+    const urls = [];
+    urls.push(window.location.origin);
+    urls.push(`${window.location.protocol}//${window.location.hostname}:8096`);
+    urls.push(await webSettings.getServers());
```
dmitrylyzo, line 26: _Untested_
```suggestion
    urls.push.apply(urls, await webSettings.getServers());
```
Otherwise, server array will be pushed as an element.

dmitrylyzo, line 26: Or probably use spread operator. _Also untested_
```suggestion
    urls.push(...await webSettings.getServers());
```

src/scripts/settings/webSettings.js
```diff
@@ -85,6 +85,15 @@ export function getMultiServer() {
     });
 }
 
+export function getServers() {
+    return getConfig().then(config => {
+        return config.servers;
```
dmitrylyzo, line 90: In case of missing of `servers` field in config file.
```suggestion
        return config.servers || [];
```

src/scripts/clientUtils.js
```diff
@@ -1,35 +1,42 @@
...
-    if (index != -1) {
-        return urlLower.substring(0, index);
-    }
+    if (current) return Promise.resolve(current);
 
-    const loc = window.location;
-    let address = loc.protocol + '//' + loc.hostname;
+    const urls = [];
+    urls.push(window.location.origin);
+    urls.push(`${window.location.protocol}//${window.location.hostname}:8096`);
```
thornbill, line 25: Maybe this should be explicitly tested using http and add a test for https over the default https port (8920 I think)?

dkanada:
@dmitrylyzo I figured it out, this pull request still works. The catch is that you have to have no sites cached or you'll get sent to one of those sites instead. @thornbill what do you think the behavior should be there? I'm currently looking at what will happen if none of the cached sites are available, or if you manually access a site that hasn't gone through the wizard. Assuming both those scenarios have reasonable behavior, I think this sounds fine.

src/scripts/clientUtils.js
dkanada: adding in the next commit

src/components/appRouter.js
dkanada: I think this has been resolved, and I also fixed a few more edge cases.

src/scripts/clientUtils.js
```diff
@@ -1,35 +1,49 @@
...
+            return Promise.resolve();
+        });
+    });
 
-    return address;
+    return Promise.all(promises).then(responses => {
+        responses = responses.filter(response => response && response.ok);
+        return Promise.all(responses.map(response => response.json()));
+    }).then(configs => {
+        const selection = configs.find(config => !config.StartupWizardCompleted)
```
dkanada, line 40:
```suggestion
        let selection = configs.find(config => !config.StartupWizardCompleted);
```

MrTimscampi:
@dkanada What's the status on this?

src/scripts/site.js
```diff
@@ -82,18 +75,25 @@ function loadCoreDictionary() {
 }
 
 function init() {
-    ServerConnections.initApiClient();
+    serverAddress().then(server => {
+        if (!server) {
+            navigate('selectserver.html');
```
dmitrylyzo, line 80: For some reason, `window` is `undefined` in `page.js`. Probably not the best place.

src/components/appRouter.js
```diff
@@ -508,42 +500,52 @@ class AppRouter {
...
         const firstResult = this.firstConnectionResult;
-        if (firstResult) {
-            this.firstConnectionResult = null;
 
-            if (firstResult.State !== 'SignedIn' && !route.anonymous) {
-                this.handleConnectionResult(firstResult);
-                return;
-            }
+        this.firstConnectionResult = null;
+        if (firstResult && firstResult.State === 'ServerSignIn') {
```
dmitrylyzo, line 505: If user has a valid access token, we have `SignedIn` here and no wizard.
_I got this using "hackery" method to run wizard: change `system.xml/IsStartupWizardCompleted`. So probably safe under normal use._

hawken93:
I got a gray empty screen testing this... Uh..

hawken93:
`Uncaught Error: item or serverId cannot be null`

hawken93:
So what I think is happening is that the System/Info/Public endpoint is being relied upon for the server address if there is not one in the json config.
I think this is wrong. That endpoint exists, to my knowledge, because the chromecast uses 8.8.8.8 for resolver and this endpoint is to help it get around issues with regards to that.

Up until now things have worked decently well, because the client can assume that the server is at the parent of web/.
I really wish to keep this behavior.

src/scripts/clientUtils.js
```diff
@@ -1,43 +1,56 @@
...
+    urls.push(window.location.origin);
+    urls.push(`https://${window.location.hostname}:8920`);
+    urls.push(`http://${window.location.hostname}:8096`);
+    urls.push(...await webSettings.getServers());
 
-    if (loc.port) {
-        address += ':' + loc.port;
-    }
+    const promises = urls.map(url => {
+        return fetch(`${url}/System/Info/Public`).catch(error => {
```
hawken93, line 38: If `url` returns a valid response, then `url` is correct. However, `LocalAddress` is not trustworthy.

So it would be wise to return `url` if this fetch succeeds.

oddstr13 (approved):
`clientUtils.serverAddress` only ever returns `LocalAddress`, breaking docker installs.
Also Promises. How? I don't know.

LGTM 🤪

src/scripts/clientUtils.js
hawken93: This fixes my complaint

```patch
--- a/src/scripts/clientUtils.js
+++ b/src/scripts/clientUtils.js
@@ -35,18 +35,14 @@ export async function serverAddress() {
     urls.push(...await webSettings.getServers());

     const promises = urls.map(url => {
-        return fetch(`${url}/System/Info/Public`).catch(error => {
+        return fetch(`${url}/System/Info/Public`).then(resp => url).catch(error => {
             return Promise.resolve();
         });
     });

     return Promise.all(promises).then(responses => {
-        responses = responses.filter(response => response && response.ok);
-        return Promise.all(responses.map(response => response.json()));
-    }).then(configs => {
-        let selection = configs.find(config => !config.StartupWizardCompleted);
-        if (!selection) selection = configs[0];
-        return Promise.resolve(selection.LocalAddress);
+        responses = responses.filter(response => response);
+        return responses[0];
     }).catch(error => {
         console.log(error);
         return Promise.resolve();
```

src/scripts/clientUtils.js
```diff
@@ -1,43 +1,52 @@
...
+    const promises = urls.map(url => {
+        return fetch(`${url}/System/Info/Public`).then(resp => url).catch(error => {
+            return Promise.resolve();
+        });
+    });
 
-    return address;
+    return Promise.all(promises).then(responses => {
+        responses = responses.filter(response => response);
+        return responses[0];
```
hawken93, line 45: "previously, I ran through the configs and checked the wizard specifically so any server found with an uncompleted wizard was given precedence" - dkanada

commented for follow-up to make this happen later

PR #1952, opened 2020-09-26
Disable 'Remember Me' by default

dmitrylyzo:
https://github.com/jellyfin/jellyfin-tizen/issues/51 in short (and interpreted):

User has 2 accounts on the server:
- admin/parent
- child

Client is on TV (or any shared device).

When you are logging in via manual form, `Remember Me` is checked by default or restored from `localStorage` and you have to disable it for admin/parent account each time other user enabled it.

Unset `enableAutoLogin` is treated as `true`, so if you have valid `AccessToken`, app logs in.

When you are logging in via visual form, you cannot turn on/off auto login (that's probably why unset `enableAutoLogin` is treated as `true`). So if no one set `Remember Me` (via manual form), you will be logged in.

**Changes**
- `Remember Me` is unchecked and not restored - you are presumably starting a new session. And if you want to be remembered you check `Remember Me` once (in opposite to unchecking it each time).
- Unset `enableAutoLogin` is treated as `false`.
- Visual form turns off `enableAutoLogin`, which disables auto login via visual form. If you have no password, you can easily click your card. With password, you must use manual form (switched automatically).

**Issues**
Fixes https://github.com/jellyfin/jellyfin-tizen/issues/51

@varlesh also said that `Remember Me` isn't restored from previous state. So there might be a `localStorage` issue, but it works on Tizen 4 and 5 (emulator).

### DIFF
--- a/src/controllers/session/login/index.html
+++ b/src/controllers/session/login/index.html
@@ -13,7 +13,7 @@
             </div>
 
             <label class="checkboxContainer">
-                <input is="emby-checkbox" type="checkbox" class="chkRememberLogin" checked />
+                <input is="emby-checkbox" type="checkbox" class="chkRememberLogin" />
                 <span>${RememberMe}</span>
             </label>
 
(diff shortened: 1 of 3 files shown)


oddstr13 (approved):

joshuaboniface (approved):

dmitrylyzo:
Closed in favor of #2481

PR #1955, opened 2020-09-28
Fix ES6 import - browser

dmitrylyzo:
In Tizen 3 (emulator), app crashes on injecting of page (`selectServer`, but this doesn't matter).
At first, I found out that with non-ES6 `browser` everything is fine.
Next, I found out that crash happens if `fonts.css` is used. With `fonts.sized.css`, everything is fine.
And finally, I found out that `browser` isn't imported correctly.

**Changes**
Add `default` target when import `browser`.

**Issues**
Tizen 3 app crashes.

### DIFF
--- a/src/scripts/site.js
+++ b/src/scripts/site.js
@@ -203,7 +203,7 @@ function initClient() {
             createConnectionManager().then(function () {
                 console.debug('initAfterDependencies promises resolved');
 
-                require(['globalize', 'browser'], function (globalize, browser) {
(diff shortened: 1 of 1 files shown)


dmitrylyzo:
App crashes because of this line of `fonts.css` https://github.com/jellyfin/jellyfin-web/blob/d5dad66eecad2331763393f26b436c3c55b63852/src/assets/css/fonts.css#L5

dkanada (approved):

Camc314 (approved):

PR #1975, opened 2020-10-07
Manual changes for no-var eslint rule

dkanada:
closes #1683

### DIFF
--- a/src/libraries/navdrawer/navdrawer.js
+++ b/src/libraries/navdrawer/navdrawer.js
@@ -2,6 +2,7 @@
  * and will be replaced soon by a Vue component.
  */
 
+/* eslint-disable no-var */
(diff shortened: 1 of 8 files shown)


Camc314 (approved):

src/libraries/navdrawer/navdrawer.js
```diff
@@ -2,6 +2,7 @@
  * and will be replaced soon by a Vue component.
  */
 
+/* eslint-disable no-var */
```
thornbill, line 5: Why disable the rule for this file?

dkanada: see the lines directly above, we should remove this file eventually

thornbill (approved):

PR #1990, opened 2020-10-14
WebRTC time syncing in SyncPlay

OancaAndrei:
**Changes**
This PR adds support for WebRTC, making it possible to time sync browsers on the same network.
The experienced playback delay between devices on the same network will be less noticeable (this is the expected result at least).

This PR is related to the [server side one](https://github.com/jellyfin/jellyfin/pull/4329).
Depends on #2204.


joshuaboniface:
@OancaAndrei Same as other PR, is this one good?

OancaAndrei:
> @OancaAndrei Same as other PR, is this one good?

Should be good for testing but needs server PR as well, which hasn't been reviewed.

MrTimscampi:
I have some concerns with this.

You mention that the playback delay between browsers will be smaller on local networks. What kind of difference are we talking, here? It'd be beneficial to have benchmarks or comparisons here, to gauge how truly useful this is.

I'm saying this because 3k lines of code to maintain is a big ask for only a few milliseconds of synchronization. I'm not against WebRTC itself (though I do have concerns about privacy, especially in regards to IP leaks on large servers where you might not know everyone) but it adds a ton of complexity, so imo it should make a large difference or it's just not worth it :/

OancaAndrei:
I made this PR with two scenarios in mind: 
- server is far away but majority of clients are on LAN;
- some clients can't properly sync (because of playback limitations of the system). 

For the first scenario, syncing music is the major use-case that is gaining from WebRTC, as audio-to-audio delays are more noticeable than audio-to-video. Peer-to-peer syncing is useful because it completely removes delays deriving from time syncing with a remote server.

The second scenario covers my Chromecast which can't speed up playback nor skip to the right point fast enough to not get lost again. The solution in this case is to let other devices deviate from the server's proposed time and adapt to the Chromecast's one. 

In terms of numbers, *SpeedToSync* already solves almost all the delays caused by network and/or player limitations and can cut them under 10 ms. WebRTC syncing would also target those devices that cannot use *SpeedToSync*. For instance, my Chromecast quite always starts playback with one or two seconds of delay, even with the server being on the same network, and this delay can't really be corrected on the device itself.

I think I'll record some audio for comparison, it will sound great. :wink:

---
The leak of local IP addresses might be a drawback for this feature. Right now the signaling messages (that contain the IP addresses) are shared with all users of the same group. A simple solution would be to restrict those massages only to the sessions of the same user. Considering that those messages could/should travel over HTTPS, the leak becomes less problematic, imo.

Also, WebRTC can be disabled entirely client side. If needed, it could be disabled server wise as well (not implemented).

---
Regarding the number of lines of code, in these past days I've been splitting this big PR into multiple ones that implement one single feature at a time. This one can be further split into "WebRTC time syncing" and "WebRTC peer-to-peer media sharing". Also, I'll have to rebase onto the right branch, as it is now targeting "Group visibility and Playback/Playlist Access Control" which isn't needed for this PR, thus further reducing the code base.

joshuaboniface:
Would IP multicast not be the ideal solution for these sort of problems?

joshuaboniface:
With you're explanation here (I didn't see this before responding in that other PR), I'm more OK with this solution than I was, but still think that working around a tech that exists (IP multicast) in favour of a lot of synchronization workaround seems like the wrong move long-term.

OancaAndrei:
> Would IP multicast not be the ideal solution for these sort of problems?

I've referenced this in the other PR but for clarity I'll add two lines here as well. The goal is to sync the time/playback between local devices. How the media is delivered to the devices won't matter as the problem of syncing has to be corrected on the player itself.

Just noticed that I've forgot to update the initial description of this PR: no P2P media sharing here.

OancaAndrei:
Reverting this PR to draft so that I can focus on polishing existing features and fix current issues.

PR #1993, opened 2020-10-18
Removed excess quality options to make menu more manageable

sparky8251:
**Changes**
Just removed all kinds of excess quality options to make the quality selector menu not require scrolling on a 1080 screen.

Removed some duplicate bitrates as well, favoring the lower bitrate being in the lower resolution (as high res + low bitrate typically looks worse in my experience).

Also, moved Auto to the top of the list (vs the bottom) and dropped 240 and 144 resolutions outright. Smallest is now 360 and I dropped the bitrate on that slightly to make it stand out among the low end of 480 bitrates.

Previous look on desktop:

Screenshot_20201017_191358

Current look on desktop:
Screenshot_20201017_200530

### DIFF
--- a/src/components/qualityOptions.js
+++ b/src/components/qualityOptions.js
@@ -15,56 +15,6 @@ export function getVideoQualityOptions(options) {
 
     const qualityOptions = [];
 
-    if (maxAllowedWidth >= 3800) {
-        qualityOptions.push({ name: '4K - 120 Mbps', maxHeight: 2160, bitrate: 120000000 });
-        qualityOptions.push({ name: '4K - 100 Mbps', maxHeight: 2160, bitrate: 100000000 });
-        qualityOptions.push({ name: '4K - 80 Mbps', maxHeight: 2160, bitrate: 80000000 });
-    }
-
-    // Some 1080- videos are reported as 1912?
-    if (maxAllowedWidth >= 1900) {
-        qualityOptions.push({ name: '1080p - 60 Mbps', maxHeight: 1080, bitrate: 60000000 });
-        qualityOptions.push({ name: '1080p - 50 Mbps', maxHeight: 1080, bitrate: 50000000 });
-        qualityOptions.push({ name: '1080p - 40 Mbps', maxHeight: 1080, bitrate: 40000000 });
-        qualityOptions.push({ name: '1080p - 30 Mbps', maxHeight: 1080, bitrate: 30000000 });
-        qualityOptions.push({ name: '1080p - 25 Mbps', maxHeight: 1080, bitrate: 25000000 });
-        qualityOptions.push({ name: '1080p - 20 Mbps', maxHeight: 1080, bitrate: 20000000 });
-        qualityOptions.push({ name: '1080p - 15 Mbps', maxHeight: 1080, bitrate: 15000000 });
-        qualityOptions.push({ name: '1080p - 10 Mbps', maxHeight: 1080, bitrate: 10000001 });
-        qualityOptions.push({ name: '1080p - 8 Mbps', maxHeight: 1080, bitrate: 8000001 });
-        qualityOptions.push({ name: '1080p - 6 Mbps', maxHeight: 1080, bitrate: 6000001 });
-        qualityOptions.push({ name: '1080p - 5 Mbps', maxHeight: 1080, bitrate: 5000001 });
-        qualityOptions.push({ name: '1080p - 4 Mbps', maxHeight: 1080, bitrate: 4000002 });
-    } else if (maxAllowedWidth >= 1260) {
-        qualityOptions.push({ name: '720p - 10 Mbps', maxHeight: 720, bitrate: 10000000 });
-        qualityOptions.push({ name: '720p - 8 Mbps', maxHeight: 720, bitrate: 8000000 });
-        qualityOptions.push({ name: '720p - 6 Mbps', maxHeight: 720, bitrate: 6000000 });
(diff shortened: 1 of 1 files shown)


oddstr13 (approved):

neopc10:
Good start -- perhaps it's worth expanding this to audio as well, since it could use some love.
Following the examples of Tidal and Qobuz, maximum streaming qualities (not in mbps):
Hi-Res 24-Bit / up to 192 kHz
Hi-Res 24-Bit / up to 96 kHz
CD 16-Bit / 44.1 kHz (lossless)
MP3/AAC/whatever 320 kbps
MP3/AAC/whatever 192 kbps

src/components/qualityOptions.js
```diff
@@ -15,66 +15,42 @@ export function getVideoQualityOptions(options) {
...
-    const autoQualityOption = {
-        name: globalize.translate('Auto'),
-        bitrate: 0,
-        selected: options.isAutomaticBitrateEnabled
-    };
-
-    if (options.enableAuto) {
-        qualityOptions.push(autoQualityOption);
+    if (maxAllowedWidth >= 620) {
+        qualityOptions.push({ name: '480p - 1.5 Mbps', maxHeight: 480, bitrate: 1500000 });
```
dkanada, line 47: I think 480 should get 1Mbps and 3Mbps instead.

src/components/qualityOptions.js
```diff
@@ -15,66 +15,42 @@ export function getVideoQualityOptions(options) {
...
-        qualityOptions.push({ name: '480p - 2 Mbps', maxHeight: 480, bitrate: 2000001 });
-        qualityOptions.push({ name: '480p - 1.5 Mbps', maxHeight: 480, bitrate: 1500001 });
+        qualityOptions.push({ name: '1080p - 10 Mbps', maxHeight: 1080, bitrate: 10000000 });
     }
-
     if (maxAllowedWidth >= 1260) {
-        qualityOptions.push({ name: '720p - 4 Mbps', maxHeight: 720, bitrate: 4000000 });
+        qualityOptions.push({ name: '720p - 8 Mbps', maxHeight: 720, bitrate: 8000000 });
+        qualityOptions.push({ name: '720p - 6 Mbps', maxHeight: 720, bitrate: 6000000 });
         qualityOptions.push({ name: '720p - 3 Mbps', maxHeight: 720, bitrate: 3000000 });
```
dkanada, line 44: This is usually too low for 720 content so I prefer to set it at 4Mbps at most.

ferferga:
This is a great start to clean up the clutter. However, I want to give the concept I had in mind for making the quality switcher great and this PR seems the best place to do so:

* First of all, quality options shouldn't be handled by clients. According to the file's bitrate, server should generate a combination of possible options and return them to client. This is good for **consistency**, as users will have same quality options in web and in Android.

With the current system, a *1080p -30 Mbps* option appears when I'm watching a 15 Mbps one, which doesn't make any sense.

* We should only have two "default" options:

      * **Maximum**: Stream original file when device is capable of direct play or a the maximum quality transcoded version that the device can handle (in case it's not compatible)

      * **Auto**: Stream's quality changes in real time based on the connectivity of the client.

* **Keep only resolution options and remove any mention to bitrates**. They're confusing for most people. We can add in the nerd stats of each player an option for requesting specific bitrates in case we *really* want to keep this. In that case, client should know from server (as mentioned above) the bitrate of the original file, so user can't request a file with a bitrate higher than the original one.

* Add an option in each's client settings (Data saver mode for instance) where the bitrates (except in Maximum) of each of the resolutions are cut in half or by a quarter. So, for instance, if for a movie en 4K@50Mbps server determines that best bitrate for 720p is 4 Mbps, client forces the server to use 2 Mbps instead.

TLDR, the idea is to make the quality switcher simpler and common, like the one found in YouTube, Vimeo, etc... While still being useful for power users and the use case of a media server.

Maybe I should make this in a separate discussion thread so we don't forget and provide feedback about this? I guess it will need to wait for 11.0 once we have an stable API.

Would be great to have all your input in this matter @jellyfin/web @jellyfin/client @jellyfin/server

src/components/qualityOptions.js
```diff
@@ -15,65 +15,41 @@ export function getVideoQualityOptions(options) {
...
         qualityOptions.push({ name: '720p - 8 Mbps', maxHeight: 720, bitrate: 8000000 });
         qualityOptions.push({ name: '720p - 6 Mbps', maxHeight: 720, bitrate: 6000000 });
-        qualityOptions.push({ name: '720p - 5 Mbps', maxHeight: 720, bitrate: 5000000 });
-    } else if (maxAllowedWidth >= 620) {
-        qualityOptions.push({ name: '480p - 4 Mbps', maxHeight: 480, bitrate: 4000001 });
-        qualityOptions.push({ name: '480p - 3 Mbps', maxHeight: 480, bitrate: 3000001 });
-        qualityOptions.push({ name: '480p - 2.5 Mbps', maxHeight: 480, bitrate: 2500000 });
-        qualityOptions.push({ name: '480p - 2 Mbps', maxHeight: 480, bitrate: 2000001 });
-        qualityOptions.push({ name: '480p - 1.5 Mbps', maxHeight: 480, bitrate: 1500001 });
+        qualityOptions.push({ name: '720p - 4 Mbps', maxHeight: 720, bitrate: 3000000 });
```
nyanmisaka, line 44:
```suggestion
        qualityOptions.push({ name: '720p - 4 Mbps', maxHeight: 720, bitrate: 4000000 });
```

ferferga (approved):

thornbill (approved):

Darkwolf1515:
image

I mean, at least its honest.

PR #2007, opened 2020-10-24
Prevent doubleclick on buttons from bubbling to video for fullscreen

### LINKED ISSUE #2006: Rapid clicks on playback buttons cause inadvertent fullscreen switch
**Describe The Bug**
Clicking multiple times on a button should not cause fullscreening/unfullscreening since it's interpreted as a double-click.

**Steps To Reproduce**
1. Play a video
2. Click "Fast-Forward" button twice quickly
3. Jellyfin goes full/unfull screen

**Expected Behavior**
I would expect rapid repeated activation of the button that was clicked.

bugfixin:
**Changes**
Fairly naive way of preventing the double click propagation - let me know if this is workable

**Issues**
Fixes #2006

### DIFF
--- a/src/controllers/playback/video/index.js
+++ b/src/controllers/playback/video/index.js
@@ -1528,23 +1528,44 @@ import 'css!assets/css/videoosd';
         view.querySelector('.btnPreviousTrack').addEventListener('click', function () {
             playbackManager.previousTrack(currentPlayer);
         });
+        view.querySelector('.btnPreviousTrack').addEventListener('dblclick', function (event) {
+            event.stopPropagation();
+        });
         view.querySelector('.btnPause').addEventListener('click', function () {
             // Ignore 'click' if another element was originally clicked (Firefox/Edge issue)
             if (this.contains(clickedElement)) {
                 playbackManager.playPause(currentPlayer);
             }
         });
+        view.querySelector('.btnPause').addEventListener('dblclick', function (event) {
+            event.stopPropagation();
+        });
         view.querySelector('.btnNextTrack').addEventListener('click', function () {
             playbackManager.nextTrack(currentPlayer);
+        });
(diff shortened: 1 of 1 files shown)


dmitrylyzo:
> let me know if this is workable

Yes, it is, but this approach requires overriding `dblclick` event in too many places: buttons, sliders.

I've tried this (see below), but it narrows the double click area due to the gradients of the top and bottom "panels".
```js
dom.addEventListener(view, 'dblclick', function (e) {
    if (e.target !== view) return; // <-- skip bubbled event
    playbackManager.toggleFullscreen(currentPlayer);
}, options);
```

bugfixin:
That's where my concern would be too - my previous approach requires setting a dblclick handler on each item that shouldn't cause fullscreening. 

I think a best of both worlds would be adding the dblclick handler to stop propagation to the .videoOsdBottom element. I'll update my PR with that approach when I get back to my workstation, it should be maintainable and only prevent fullscreening from happening when clicked within the osd area.

dmitrylyzo:
It turns out that header doesn't respond to double-clicking - only bottom panel belongs to videoOsd (`view`). The latter includes:
- `videoOsdBottom`
- `upNextContainer`

I think both should be excluded from double click. This is easier to do in the existing event handler, where I have shown.

By the way, `click` is ignored like this: https://github.com/jellyfin/jellyfin-web/blob/fd17088014fadc6183cf8535d7d8a07924abe847/src/controllers/playback/video/index.js#L1420-L1422
So I'm fine with the `return` in `view` handler, since `click` has the same narrowed area.

dkanada:
@bugfixin any thoughts? We have a release coming up very soon if you want to get this included.

dkanada:
@dmitrylyzo opened an alternate PR that seemed to work so it was merged in preparation for the upcoming release. Feel free to open this one again if it still has any issues though.

PR #2010, opened 2020-10-25
pull fonts from official repository

dkanada:


nielsvanvelzen:
If this is merged we **must** update the documentation and add a note to the 10.7 release notes to warn people that enabled CSP to update their configuration.

All documentation for reverse proxies (like [nginx](https://jellyfin.org/docs/general/networking/nginx.html)) don't include the Jellyfin repository at this moment in the CSP config.

MrTimscampi:
It's important to not that currently, the CSS files on our repository will load content from Google servers.

Do we want that or not?

dkanada:
Someone mentioned it on Matrix, and I'll be fixing it soon.

MrTimscampi (approved):

ferferga (approved):
All good now!

PR #2027, opened 2020-10-31
Remove custom hover style

### LINKED ISSUE #1833: Repeat/Shuffle Button Stays in Hover State

dmitrylyzo:
#1867 disables `hover` style on mobile. But there is one more place that is causing #1833. This is only relevant for narrow screens (`43em`).
https://github.com/jellyfin/jellyfin-web/blob/6a82956cf46135e44a2a9ad620bcac6118b110f4/src/components/remotecontrol/remotecontrol.css#L375-L378
~These lines were probably added to fix the same `hover` style issue, and became unnecessary because #1867 disabled unwanted `hover` styles.~

**Desktop**
Before
before_desktop

After
test2

**Mobile**
Before
before_mobile

After
output

**Changes**
Remove custom hover style.

**Issues**
Fixes #1833

### DIFF
--- a/src/components/remotecontrol/remotecontrol.css
+++ b/src/components/remotecontrol/remotecontrol.css
@@ -372,8 +372,7 @@
         font-size: smaller;
     }
 
-    .paper-icon-button-light:hover {
-        color: #fff !important;
+    .paper-icon-button-light {
         background-color: transparent !important;
     }
 
@@ -381,10 +380,6 @@
         padding: 0;
         margin: 0;
         font-size: 1.7em;
-    }
-
-    .btnPlayPause:hover {
-        background-color: transparent !important;
     }
 
     .nowPlayingPageImage {


dkanada:
I think it was originally added because the play button hover looks bad. In reality, the white circle should probably turn blue instead of a separate hover circle.

dmitrylyzo:
> I think it was originally added because the play button hover looks bad.

Or because of this (hover on fast-forward)
before-focus
This is due to the alignment of the button using `padding-right`.

By the way,  in normal (non-compact) version of UI, play button has a hover indicator.
_`master` branch_
master1

> In reality, the white circle should probably turn blue instead of a separate hover circle.

Modified version of initial approach (no `hover` and `color`).
```css
.paper-icon-button-light {
    background-color: transparent !important;
}
```
test2

What about blue `color`? On mobile, button flashes with this color on touch.
Personally, I think it should be controlled by theme.

MrTimscampi:
> Personally, I think it should be controlled by theme.

I would agree with this. Flashing a color on touch is fine (it helps give feedback, even), but the color should be theme-controlled.

dmitrylyzo:
I updated the first post with a new before/after.

joshuaboniface (approved):

dkanada (approved):

PR #2028, opened 2020-11-02
Hide progress bar when playing theme media

danieladov:
Currently, jellyfin plays the theme songs correctly but the progress bar is always present during it.  As here
themeSongsBar

With these changes the progress bar is hidden.

### DIFF
--- a/src/components/themeMediaPlayer.js
+++ b/src/components/themeMediaPlayer.js
@@ -26,8 +26,7 @@ function playThemeMedia(items, ownerId) {
             enableRemotePlayers: false
         }).then(function () {
             currentOwnerId = ownerId;
-        })
-       
+        });
     } else {
         stopIfPlaying();
     }
--- a/src/components/themeMediaPlayer.js
+++ b/src/components/themeMediaPlayer.js
@@ -26,8 +26,8 @@ function playThemeMedia(items, ownerId) {
             enableRemotePlayers: false
         }).then(function () {
             currentOwnerId = ownerId;
-        });
-        document.getElementsByClassName("nowPlayingBarTop")[0].style.display = "none";
+        })
+       
     } else {
         stopIfPlaying();
     }
--- a/src/components/themeMediaPlayer.js
+++ b/src/components/themeMediaPlayer.js
@@ -27,6 +27,7 @@ function playThemeMedia(items, ownerId) {
         }).then(function () {
             currentOwnerId = ownerId;
         });
+        document.getElementsByClassName("nowPlayingBarTop")[0].style.display = "none";
     } else {
         stopIfPlaying();
     }
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -41,6 +41,7 @@
  - [Cromefire_](https://github.com/cromefire)
  - [Orry Verducci](https://github.com/orryverducci)
  - [Camc314](https://github.com/camc314)
+ - [danieladov](https://github.com/danieladov)
 
 # Emby Contributors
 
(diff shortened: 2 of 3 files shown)


src/components/themeMediaPlayer.js
```diff
@@ -27,6 +27,7 @@ function playThemeMedia(items, ownerId) {
         }).then(function () {
             currentOwnerId = ownerId;
         });
+        document.getElementsByClassName("nowPlayingBarTop")[0].style.display = "none";
```
dmitrylyzo, line 30: tbh, looks a bit hacky.
1. Doesn't this break `nowPlayingBar` for normal music playback after touching theme song?
  - Go to item page with theme song.
  - After theme song started, go to the music library and start normal music playback.
2. TVs (the real ones - with `browser.tv === true`) don't have `nowPlayingBar`, iirc. So this will probably fail to access to `undefined` element.

I think we need to tell `nowPlayingBar` somehow that we don't want it to be displayed. Then hide it like this:
https://github.com/jellyfin/jellyfin-web/blob/4502c408f1656eb1948adc88b8b1eb96d207c336/src/components/nowPlayingBar/nowPlayingBar.js#L667-L670

P.S.: we use single quotes.

danieladov: Yes, it's a bit hacky. Any idea of how to do it properly?

dmitrylyzo: I changed mentioned block like this
```js
        if (!state.NowPlayingItem || layoutManager.tv || !state.IsFullscreen) {
            hideNowPlayingBar();
            return;
        }
```
`state.IsFullscreen` property replicates `fullscreen` in options for `playbackManager.play`. By default it is `true` and only in `themeMediaPlayer` it is set to `false`. Video player uses `fullscreen` to choose between normal mode and backdrop mode (animated background). So it can probably be used to hide `nowPlayingBar` (similar to "backdrop mode").
I am still not totally sure, but it works and everything is done in `nowPlayingBar`.

danieladov: That's a way better solution, thank you so much.

dmitrylyzo:
I just thought. If we hide `nowPlayingBar`, we cannot change volume (other controls are less usable for theme song).
Keyboard media buttons (volume-related) can be handled in `scripts/keyboardNavigation.js`.
TVs (real ones) have a physical volume control - no problem.
But what about mouse users?

As an additional idea, we could add volume option next to the theme-related checkboxes in `Display` user settings - just like `Music volume` in games. Then use it to suppress theme songs volume.

src/components/themeMediaPlayer.js
```diff
@@ -26,7 +26,8 @@ function playThemeMedia(items, ownerId) {
             enableRemotePlayers: false
         }).then(function () {
             currentOwnerId = ownerId;
-        });
+        })
+
```
dmitrylyzo, line 30:
```suggestion
        });
```

danieladov:
> I just thought. If we hide `nowPlayingBar`, we cannot change volume (other controls are less usable for theme song).
> Keyboard media buttons (volume-related) can be handled in `scripts/keyboardNavigation.js`.
> TVs (real ones) have a physical volume control - no problem.
> But what about mouse users?
> 
> As an additional idea, we could add volume option next to the theme-related checkboxes in `Display` user settings - just like `Music volume` in games. Then use it to suppress theme songs volume.
Yes, this is a problem, I guess we can set a default volume configurable through settings as you say

MrTimscampi (approved):

thornbill (approved):

PR #2039, opened 2020-11-04
Photos fullscreen button + hiding exit and arrows buttons on autoplay

ThibaultNocchi:
**Changes**
* Added fullscreen button on the photos viewer
* Automatically hides exit, left and right buttons on the photos viewer
* Added method to calculate keyframes in `slideshow.js` based on where is the hidden element (up, down, left, right) and if it is going to be hidden or shown

### DIFF
--- a/src/components/slideshow/slideshow.js
+++ b/src/components/slideshow/slideshow.js
@@ -16,6 +16,7 @@ import ServerConnections from '../ServerConnections';
 // eslint-disable-next-line import/named, import/namespace
 import { Swiper } from 'swiper/swiper-bundle.esm';
 import 'swiper/swiper-bundle.css';
+import screenfull from 'screenfull';
 
 /**
  * Name of transition event.
@@ -177,6 +178,10 @@ export default function (options) {
                 if (appHost.supports('sharing')) {
                     html += getIcon('share', 'btnShare slideshowButton', true);
                 }
+                if (screenfull.isEnabled) {
+                    html += getIcon('fullscreen', 'btnFullscreen', true);
+                    html += getIcon('fullscreen_exit', 'btnFullscreenExit hide', true);
+                }
             }
             html += getIcon('close', 'slideshowButton btnSlideshowExit hide-mouse-idle-tv', false);
             html += '</div>';
@@ -190,6 +195,10 @@ export default function (options) {
                 }
                 if (appHost.supports('sharing')) {
                     html += getIcon('share', 'btnShare slideshowButton', true);
+                }
+                if (screenfull.isEnabled) {
+                    html += getIcon('fullscreen', 'btnFullscreen', true);
+                    html += getIcon('fullscreen_exit', 'btnFullscreenExit hide', true);
                 }
 
                 html += '</div>';
@@ -218,6 +227,22 @@ export default function (options) {
             const btnShare = dialog.querySelector('.btnShare');
             if (btnShare) {
                 btnShare.addEventListener('click', share);
+            }
+
+            const btnFullscreen = dialog.querySelector('.btnFullscreen');
+            if (btnFullscreen) {
+                btnFullscreen.addEventListener('click', fullscreen);
+            }
+
+            const btnFullscreenExit = dialog.querySelector('.btnFullscreenExit');
+            if (btnFullscreenExit) {
+                btnFullscreenExit.addEventListener('click', fullscreenExit);
+            }
+
+            if (screenfull.isEnabled) {
+                screenfull.on('change', function () {
+                    toggleFullscreenButtons(screenfull.isFullscreen);
+                });
             }
         }
 
@@ -450,6 +475,35 @@ export default function (options) {
     }
 
     /**
+     * Goes to fullscreen using screenfull plugin
+     */
+    function fullscreen() {
+        if (!screenfull.isFullscreen) screenfull.request();
+        toggleFullscreenButtons(true);
+    }
+
+    /**
+     * Exits fullscreen using screenfull plugin
+     */
+    function fullscreenExit() {
+        if (screenfull.isFullscreen) screenfull.exit();
+        toggleFullscreenButtons(false);
+    }
+
+    /**
+     * Updates the display of fullscreen buttons
+     * @param {boolean} isFullscreen - Whether the wanted state of buttons is fullscreen or not
+     */
+    function toggleFullscreenButtons(isFullscreen) {
+        const btnFullscreen = dialog.querySelector('.btnFullscreen');
+        const btnFullscreenExit = dialog.querySelector('.btnFullscreenExit');
+        if (btnFullscreen)
+            btnFullscreen.classList.toggle('hide', isFullscreen);
+        if (btnFullscreenExit)
+            btnFullscreenExit.classList.toggle('hide', !isFullscreen);
+    }
+
+    /**
      * Starts the autoplay feature of the Swiper instance.
      */
     function play() {
@@ -483,6 +537,9 @@ export default function (options) {
      * Closes the dialog and destroys the Swiper instance.
      */
     function onDialogClosed() {
+        // Exits fullscreen
+        fullscreenExit();
+
         const swiper = swiperInstance;
         if (swiper) {
             swiper.destroy(true, true);
--- a/src/components/slideshow/slideshow.js
+++ b/src/components/slideshow/slideshow.js
@@ -269,6 +269,12 @@ export default function (options) {
         if (btnSlideshowPause) {
             btnSlideshowPause.classList.replace('play_arrow', 'pause');
         }
+
+        const left = dialog.querySelector('.btnSlideshowPrevious');
+        if (left) slideToHide(left, 'left');
+
+        const right = dialog.querySelector('.btnSlideshowNext');
+        if (right) slideToHide(right, 'right');
     }
 
     /**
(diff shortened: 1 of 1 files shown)


MrTimscampi (approved):

src/components/slideshow/slideshow.js
```diff
@@ -215,6 +226,22 @@ export default function (options) {
             if (btnShare) {
                 btnShare.addEventListener('click', share);
             }
+
+            const btnFullscreen = dialog.querySelector('.btnFullscreen', true);
```
dmitrylyzo, line 230: What is `true` for?

ThibaultNocchi: Wow nice catch, I don't know. Fixed it!

src/components/slideshow/slideshow.js
```diff
@@ -501,9 +572,22 @@ export default function (options) {
...
+        if (exit) slideToShow(exit, 'up');
+
+        // When autoplaying, we don't put back the arrows
+        if (!autoPlayActive) {
+            const left = dialog.querySelector('.btnSlideshowPrevious');
+            if (left) slideToShow(left, 'left');
+
+            const right = dialog.querySelector('.btnSlideshowNext');
+            if (right) slideToShow(right, 'right');
+        }
```
dmitrylyzo, line 588: The nice thing about these buttons is that you can manually flip slides and stop autoplay at the same time. I think this is very helpful.
The same thing happens when you swipe slides on mobile.

Maybe left them in place even with autoplay?

dmitrylyzo (commented):
An error occurs: `element.querySelector('.btnSlideshowPause')` returns `null`, because now `element` may differ from the bottom OSD.
https://github.com/jellyfin/jellyfin-web/blob/bb56f8168a038d85b46ed68ccb6787aea0bec940/src/components/slideshow/slideshow.js#L663

src/components/slideshow/slideshow.js
```diff
@@ -447,6 +488,33 @@ export default function (options) {
...
+
+    /**
+     * Updates the display of fullscreen buttons
+     * @param {boolean} isFullscreen - Whether the wanted state of buttons is fullscreen or not
+     */
+    function toggleFullscreenButtons(isFullscreen) {
+        const btnFullscreen = dialog.querySelector('.btnFullscreen');
+        const btnFullscreenExit = dialog.querySelector('.btnFullscreenExit');
+        isFullscreen ? btnFullscreen.classList.add('hide') : btnFullscreen.classList.remove('hide');
+        isFullscreen ? btnFullscreenExit.classList.remove('hide') : btnFullscreenExit.classList.add('hide');
```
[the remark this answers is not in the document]
ThibaultNocchi: Nice one, implemented thanks!

src/components/slideshow/slideshow.js
ThibaultNocchi: The main reason I put this in is because someone asked if it was possible to go "full" fullscreen for a photo frame. If I remember correctly, it was @MrTimscampi who told me to hide them in autoplay, but maybe we could put them back?

ThibaultNocchi:
> An error occurs: `element.querySelector('.btnSlideshowPause')` returns `null`, because now `element` may differ from the bottom OSD.
> 
> https://github.com/jellyfin/jellyfin-web/blob/bb56f8168a038d85b46ed68ccb6787aea0bec940/src/components/slideshow/slideshow.js#L663

Fixed in latest forced push

src/components/slideshow/slideshow.js
dmitrylyzo: I think it's enough that they hide on user idle.

ThibaultNocchi: Fair to me, fixed!

src/components/slideshow/slideshow.js
```diff
@@ -481,6 +545,9 @@ export default function (options) {
      * Closes the dialog and destroys the Swiper instance.
      */
     function onDialogClosed() {
+        // Exits fullscreen
+        fullscreenExit();
```
dmitrylyzo, line 549: In TV layout, this gives an error here https://github.com/jellyfin/jellyfin-web/blob/f7f96bee73127c5e0699ea03e549fa9374426c87/src/components/slideshow/slideshow.js#L510
because buttons were not created.

ThibaultNocchi: Thanks again, forgot the `if(btnFullscreen)` before it, did the same for the exit fullscreen button. :)

dmitrylyzo (commented):
It's workable right now.
As for suggestions, no need to force-push.

One more thing. When you start photo viewer, some buttons (close, left, right) stay visible until you move mouse cursor. I tried to call `startHideTimer` after `loadSwiper`, but that breaks Mobile: close button stays hidden forever - no way to close viewer. This can be solved later because need to think about UI for Mobile layout.

src/components/slideshow/slideshow.js
```diff
@@ -431,7 +468,7 @@ export default function (options) {
     function download() {
         const imageInfo = getCurrentImageInfo();
 
-        import('fileDownloader').then(({default: fileDownloader}) => {
+        import('fileDownloader').then(({ default: fileDownloader }) => {
```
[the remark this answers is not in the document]
ThibaultNocchi: Why is that? :) Cause changing the parameter for this formatting edits a few other lines I haven't written in the file, and I'm not sure if I should also commit them.

ThibaultNocchi:
I've fixed the suggestions you've done. Also, the left, right and exit icons shouldn't be displayed on picture opening now (I just apply the "hide" class to them when swiper is created). Just a question, why shouldn't I use force push? I like to use it to keep the commit tree clean, especially when almost all of them are just fixes to respect the repo guidelines or because I forgot an edge case. Thanks! :)

src/components/slideshow/slideshow.js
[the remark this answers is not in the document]
dmitrylyzo: This just reverts unwanted formatting changes - in ES6 PR these lines will be changed as well.

src/components/slideshow/slideshow.js
```diff
@@ -230,6 +255,13 @@ export default function (options) {
         dialog.addEventListener('close', onDialogClosed);
 
         loadSwiper(dialog, options);
+
+        const btnSlideshowExit = dialog.querySelector('.btnSlideshowExit');
+        if (btnSlideshowExit) btnSlideshowExit.classList.add('hide');
```
dmitrylyzo, line 260: Same problem as when calling `startHideTimer` - you cannot close slideshow on mobile. You can activate "device mode" in Chrome or Firefox using some phone/tablet profile to test it.
Let's do it for Desktop layout only.
Moreover, I think we can hide/show a top button group instead of a single `Exit` button.

```suggestion
        if (layoutManager.desktop) {
            const topActionButtons = dialog.querySelector('.topActionButtons');
            if (topActionButtons) topActionButtons.classList.add('hide');
        }

```

src/components/slideshow/slideshow.js
```diff
@@ -501,9 +565,19 @@ export default function (options) {
     function showOsd() {
         const bottom = dialog.querySelector('.slideshowBottomBar');
         if (bottom) {
-            slideUpToShow(bottom);
-            startHideTimer();
+            slideToShow(bottom, 'down');
         }
+
+        const exit = dialog.querySelector('.btnSlideshowExit');
+        if (exit) slideToShow(exit, 'up');
```
dmitrylyzo, line 572:
```suggestion
        const topActionButtons = dialog.querySelector('.topActionButtons');
        if (topActionButtons) slideToShow(topActionButtons, 'up');
```

src/components/slideshow/slideshow.js
```diff
@@ -512,8 +586,17 @@ export default function (options) {
     function hideOsd() {
         const bottom = dialog.querySelector('.slideshowBottomBar');
         if (bottom) {
-            slideDownToHide(bottom);
+            slideToHide(bottom, 'down');
         }
+
+        const exit = dialog.querySelector('.btnSlideshowExit');
+        if (exit) slideToHide(exit, 'up');
```
dmitrylyzo, line 593:
```suggestion
        const topActionButtons = dialog.querySelector('.topActionButtons');
        if (topActionButtons) slideToHide(topActionButtons, 'up');
```

dmitrylyzo (commented):
> I've fixed the suggestions you've done. Also, the left, right and exit icons shouldn't be displayed on picture opening now (I just apply the "hide" class to them when swiper is created).

It would be better to leave it for another PR, but since you started :smile: 
I think we are free to hide arrow buttons for any layout (as you did it):
- On Desktop, they will be shown on mouse move.
- On Mobile, they are not needed.
- On TV, ... there is no buttons.

But there is a problem with `Exit`. I think there is no need to solve it right now - just leave it operable. I left comment below.

> Just a question, why shouldn't I use force push? I like to use it to keep the commit tree clean, especially when almost all of them are just fixes to respect the repo guidelines or because I forgot an edge case.

I lose what I have reviewed and what I have not. :upside_down_face:
I also use force-push sometimes, but I try to do it before the review.

dkanada:
@ThibaultNocchi we are freezing all changes in a few days for the next release, did you want to try and get this in before then?

ThibaultNocchi:
@dkanada Yeah I'll try, I'm on it this afternoon. :)

ThibaultNocchi:
@dmitrylyzo I've applied your comments, and also I've taken note on why to not force push. Finally, I've rebased against master to avoid conflicts and tried it.

dmitrylyzo (approved):

PR #2044, opened 2020-11-07
Fix Schedules Direct Listings

neilsb:
Schedules Direct not able to be added as a new Listing Provider

**Changes**
 * Changed Password Encoding to be SHA-1 as required by SchedulesDirect rather than SHA256
 * Removed check for self.TextEncoder as this is never set
 * Removed Plain Text password from API call, as not required

Relies on server PR jellyfin/jellyfin#4440

### DIFF
--- a/src/components/tvproviders/schedulesdirect.js
+++ b/src/components/tvproviders/schedulesdirect.js
@@ -84,9 +84,6 @@ export default function (page, providerId, options) {
     }
 
     function sha256(str) {
-        if (!self.TextEncoder) {
-            return Promise.resolve('');
-        }
 
         const buffer = new TextEncoder('utf-8').encode(str);
         return crypto.subtle.digest('SHA-256', buffer).then(function (hash) {
--- a/src/components/tvproviders/schedulesdirect.js
+++ b/src/components/tvproviders/schedulesdirect.js
@@ -83,20 +83,6 @@ export default function (page, providerId, options) {
         loading.hide();
     }
 
-    function hex(buffer) {
-        const hexCodes = [];
-        const view = new DataView(buffer);
-
-        for (let i = 0; i < view.byteLength; i += 4) {
-            const value = view.getUint32(i);
-            const stringValue = value.toString(16);
-            const paddedValue = ('00000000' + stringValue).slice(-'00000000'.length);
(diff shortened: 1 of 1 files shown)


src/components/tvproviders/schedulesdirect.js
```diff
@@ -83,13 +83,9 @@ export default function (page, providerId, options) {
         loading.hide();
     }
 
-    function sha256(str) {
-        if (!self.TextEncoder) {
```
MrTimscampi, line 87: This should probably be
```
if (!window.TextEncoder)
```

The goal being to test if TextEncoder is supported or not.

MrTimscampi: Although, theoretically, this would only ever be executed on desktop or mobile, since TVs don't have access to the dashboard. And these always have TextEncoder available, so not sure if it's needed or not...

src/components/tvproviders/schedulesdirect.js
```diff
@@ -110,13 +106,12 @@ export default function (page, providerId, options) {
 
     function submitLoginForm() {
         loading.show();
-        sha256(page.querySelector('.txtPass').value).then(function (passwordHash) {
+        sha1(page.querySelector('.txtPass').value).then(function (passwordHash) {
```
MrTimscampi, line 109: I still don't get why we do that in the client instead of on the server, but eh.

It feels like we should always send passwords in clear and expect HTTPS.

src/components/tvproviders/schedulesdirect.js
neilsb: Just safe to remove the hashing client side and always do the it on the server then?

neilsb:
Updated to send plaintext password.   Now relies on jellyfin/jellyfin#4440

src/components/tvproviders/schedulesdirect.js
MrTimscampi: Yeah, removing the hashing here and doing it on the server, you can remove all of this and just send the unhashed password :)

MrTimscampi (approved):

Camc314 (approved):
LGTM

PR #2064, opened 2020-11-12
Add initial profile for HEVC over FMP4-HLS

nyanmisaka:
**Server changes**
https://github.com/jellyfin/jellyfin/pull/4432

**Changes**
- Add an initial profile for streaming fMP4-HLS.
- Add an option to enable fMP4-HLS container.
- Add an option to disable HEVC encoding.
- Add an option to select the maximum audio channels. (for direct streaming and transcoding)
- Fix SSA subtitle render not working on iOS client.

**Notes**
Currently streaming HEVC over fMP4-HLS is only supported on Safari, Tizen and WebOS per their docs. `hls.js` doesn't support HEVC in fMP4-HLS even in Edge Chromium with hw codecs installed. The same is true for Chrome and Firefox.

FA908CF2-3C18-42B3-9866-BEC35BA8B708


src/strings/en-us.json
```diff
@@ -1428,5 +1429,12 @@
     "SubtitleVerticalPositionHelp": "Line number where text appears. Positive numbers indicate top down. Negative numbers indicate bottom up.",
     "Preview": "Preview",
     "LabelMaxMuxingQueueSize": "Max muxing queue size:",
-    "LabelMaxMuxingQueueSizeHelp": "Maximum number of packets that can be buffered while waiting for all streams to initialize. Try to increase it if you still encounter \"Too many packets buffered for output stream\" error in ffmpeg logs. The recommended value is 2048."
+    "LabelMaxMuxingQueueSizeHelp": "Maximum number of packets that can be buffered while waiting for all streams to initialize. Try to increase it if you still encounter \"Too many packets buffered for output stream\" error in ffmpeg logs. The recommended value is 2048.",
+    "PreferFmp4HlsContainer": "Prefer fMP4-HLS Media Container",
+    "PreferFmp4HlsContainerHelp": "Prefer to use fMP4 as the default container for HLS, making it possible to direct streaming HEVC content on supported devices.",
```
MrTimscampi, line 1434:
```suggestion
    "PreferFmp4HlsContainerHelp": "Prefer to use fMP4 as the default container for HLS, making it possible to direct stream HEVC content on supported devices.",
```

MrTimscampi (approved):

Camc314 (approved):

src/scripts/browserDeviceProfile.js
```diff
@@ -413,38 +436,54 @@ import browser from './browser';
...
             return (options.disableHlsVideoAudioCodecs || []).indexOf(c) === -1;
         });
 
         const mp4VideoCodecs = [];
         const webmVideoCodecs = [];
-        const hlsVideoCodecs = [];
+        const hlsInTsVideoCodecs = [];
+        const hlsInFmp4VideoCodecs = [];
+
+        if (browser.safari || browser.tizen || browser.web0s && canPlayHevc(videoTestElement, options)) {
```
dmitrylyzo, line 470: Shouldn't it be like this?
```suggestion
        if ((browser.safari || browser.tizen || browser.web0s) && canPlayHevc(videoTestElement, options)) {
```
Otherwise, `canPlayHevc` is called only for webOS.

nyanmisaka: oops!

PR #2065, opened 2020-11-13
Make mediaInfo report viewing progress

danielhlewis:
On the details page for video files, if the user has partially viewed
the video, the number of minutes remaining will be displayed along
with the total number of minutes in the file. Also, the 'Ends at'
time will be calculated from the resume point, rather than the
beginning of the video.

**Changes**

**Issues**

### DIFF
--- a/src/components/mediainfo/mediainfo.js
+++ b/src/components/mediainfo/mediainfo.js
@@ -257,16 +257,18 @@ import 'emby-button';
             if (item.Type === 'Audio') {
                 miscInfo.push(datetime.getDisplayRunningTime(item.RunTimeTicks));
             } else {
-                minutes = item.RunTimeTicks / 600000000;
+                // 10,000 ticks per millisecond * 60,000 milliseconds per minute = 600,000,000 ticks per minute
+                const ticksPerMinute = 10000 * 60000;
+                minutes = item.RunTimeTicks / ticksPerMinute;
 
                 minutes = minutes || 1;
 
                 if (item.UserData?.PlaybackPositionTicks) {
-                    let remainingMinutes = (item.RunTimeTicks - item.UserData.PlaybackPositionTicks) / 600000000;
+                    let remainingMinutes = (item.RunTimeTicks - item.UserData.PlaybackPositionTicks) / ticksPerMinute;
                     remainingMinutes = remainingMinutes || 1;
-                    miscInfo.push(`${Math.round(minutes)} mins (${Math.round(remainingMinutes)} remaining)`);
+                    miscInfo.push(`${globalize.translate('ValueMinutes', Math.round(minutes))} (${globalize.translate("NumberRemaining", Math.round(remainingMinutes))})`);
                 } else {
-                    miscInfo.push(`${Math.round(minutes)} mins`);
+                    miscInfo.push(`${globalize.translate('ValueMinutes', Math.round(minutes))}`);
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


dkanada:
@iwalton3 this relates to your tick to millisecond changes.

iwalton3:
:+1: Good to know. I haven't worked much on that recently, though.

src/components/mediainfo/mediainfo.js
```diff
@@ -261,7 +261,13 @@ import 'emby-button';
 
                 minutes = minutes || 1;
 
-                miscInfo.push(`${Math.round(minutes)} mins`);
+                if (item.UserData?.PlaybackPositionTicks) {
+                    let remainingMinutes = (item.RunTimeTicks - item.UserData.PlaybackPositionTicks) / 600000000;
+                    remainingMinutes = remainingMinutes || 1;
+                    miscInfo.push(`${Math.round(minutes)} mins (${Math.round(remainingMinutes)} remaining)`);
```
MrTimscampi, line 267: This should be translated.

src/components/mediainfo/mediainfo.js
```diff
@@ -261,7 +261,13 @@ import 'emby-button';
...
+                } else {
+                    miscInfo.push(`${Math.round(minutes)} mins`);
```
MrTimscampi, line 269: This should be translated as well.

src/components/mediainfo/mediainfo.js
```diff
@@ -261,7 +261,13 @@ import 'emby-button';
 
                 minutes = minutes || 1;
 
-                miscInfo.push(`${Math.round(minutes)} mins`);
+                if (item.UserData?.PlaybackPositionTicks) {
+                    let remainingMinutes = (item.RunTimeTicks - item.UserData.PlaybackPositionTicks) / 600000000;
```
MrTimscampi, line 265: It's not immediately obvious to new contributors why we're dividing by 600000000 so, for maintainability, it'd be good to decompose this operation a bit and add a comment about why and how.

MrTimscampi (changes requested):

src/components/mediainfo/mediainfo.js
danielhlewis: As a new contributor, I agree that it's not immediately obvious, so I just copied the magic number from a couple lines above.  I did a global search for RunTimeTicks and didn't really find it explained anywhere else, though it looks like it's 10,000 ticks per ms * 60,0000 ms per minute.  Do you think changing it to something like this would be helpful?

    // 10,000 ticks per millisecond * 60,000 milliseconds per minute = 600,000,000 ticks per minute
    const ticksPerMinute = 10000 * 60000;
    let remainingMinutes = (item.RunTimeTicks - item.UserData.PlaybackPositionTicks) / ticksPerMinute;

src/components/mediainfo/mediainfo.js
danielhlewis: Added translation in the new commit.  Wasn't sure what the process was for adding new strings to be translated, so I just tacked it on at the end of the en-us.json file.  Let me know if there's something else I should have done.

danielhlewis:
I think I've addressed the requested changes.  Let me know if there's anything else that needs attention.

thornbill:
Shoo bot.

This needs updated, but seems like a good addition now that we’re out of code freeze.

Issue #879, opened 2020-03-04
Video playback offset after skipping in Safari on macOS


lldsolitude:
**Update 4/13**
I have discovered that it's not the subtitle that is out of sync, but rather the video (and audio) gets randomly offset after skipping forward or back during playback and does not match the timestamp displayed in the progress bar anymore. The subtitles are correctly timed.

**Describe the bug**

Subtitle goes out of sync whenever I skip forward during playback or when playback starts at a time other than the beginning. Also the subtitle can't be disabled once it's enabled; the subtitle still displays even if I switch the setting to "Off". 
**To Reproduce**

Out of sync:
1. Start playback.
2. Click fast-forward button or progress bar to skip forward.
or
1. Resume incomplete playback.

Subtitle can't be disabled:
1. Start playback.
2. Choose any subtitle track in subtitle selection menu.
3. Choose "Off" in subtitle selection menu. The subtitle still displays.

**Expected behavior**

1. Subtitles stay in sync at all times.
2. Subtitles can be turned off.

**System (please complete the following information):**
 - OS: macOS 10.15.3
 - Browser: Safari 13.0.5
 - Jellyfin Version: 10.4.3

JustAMan:
Wait until we release 10.5.0 (expected really soon) then update and re-test.
I doubt we would be fixing issues on 10.4.x branch with 10.5.0 being almost ready.

lldsolitude:
I have tested 10.5.0. Subs are now in sync all the time, but still can't be turned off. I can switch between different languages but can't turn it off.

JustAMan:
cc @redSpoutnik as our main contributor to subtitles sync :)

redSpoutnik:
Unfortunately I don't have access to a macOS device, and it seems that safari is unavailable for linux.

It will be difficult to me to reproduce the error.

JustAMan:
@lldsolitude what subtitles are you using? can you try with Chrome/Firefox?

lldsolitude:
I'm using English srt subtitles. Chrome/Firefox doesn't have sync problems, only Safari does. Unfortunately, only Safari supports subtitle display in picture-in-picture mode (in Chrome/Firefox the subtitles stay in the main window and not in the PiP window), so I'd like subtitle sync to work on Safari if possible.

JustAMan:
[...]

PR #2069, opened 2020-11-16
Show Remux as a Playback Method

Artiume:
**Changes**

**Issues**

### DIFF
--- a/src/components/playback/playmethodhelper.js
+++ b/src/components/playback/playmethodhelper.js
@@ -3,7 +3,9 @@ export function getDisplayPlayMethod(session) {
         return null;
     }
 
-    if (session.TranscodingInfo && session.TranscodingInfo.IsVideoDirect) {
+    if (session.TranscodingInfo && session.TranscodingInfo.IsVideoDirect && session.TranscodingInfo.IsAudioDirect) {
+        return 'Remux';
+    } else if (session.TranscodingInfo && session.TranscodingInfo.IsVideoDirect) {
         return 'DirectStream';
     } else if (session.PlayState.PlayMethod === 'Transcode') {
         return 'Transcode';


MrTimscampi (approved):

Camc314 (approved):

src/components/playback/playmethodhelper.js
```diff
@@ -3,7 +3,9 @@ export function getDisplayPlayMethod(session) {
         return null;
     }
 
-    if (session.TranscodingInfo && session.TranscodingInfo.IsVideoDirect) {
+    if (session.TranscodingInfo && session.TranscodingInfo.IsVideoDirect && session.TranscodingInfo.IsAudioDirect) {
+        return 'Remux';
+    } else if (session.TranscodingInfo && session.TranscodingInfo.IsVideoDirect) {
         return 'DirectStream';
```
dkanada, line 9: @MrTimscampi don't we want to remove this term entirely? Might as well do it now if that's the case.

MrTimscampi: It depends if we want to separate a transcode with the video from one without.

Imo, having more terms make no sense, so we may as well remove it, yeah. It's either a transcode or a remux, there's nothing in-between (and we have the playback data for more info as to what is happening)

nyanmisaka: Transcoding audio on the fly isn't a thing comparing to the video re-encoding. Thus separating (vcopy+a copy) and (vcopy+acodec) will just confuse users.

Artiume: I want to keep it because of ffmpeg behavior. take for example our keyframe bugs.

jellyfin/jellyfin#2547 / jellyfin/jellyfin-web#879 / https://github.com/jellyfin/jellyfin-android/issues/301 / jellyfin/jellyfin-expo#54 Audio/Video playback position wrong after changing playback position due to non-keyframe HLS use

this bug impacts Remux and DirectStream Only.

Artiume: To update all:

Direct Play - Lossless Process
Remux - Lossless process, invokes FFMpeg to HLS stream. Suffers from keyframe issues
DirectStream - Converts Audio, Can be process intensive for some low powered machines when converting something such as 7.1 DTS audio. Suffers from keyframe issues.
Transcode - Video Transcoding. Audio doesnt matter.

I'd like to keep all 4 terms. Each scenario has its own issues and uniqueness. I'm down for renaming DirectStream but I can't think of a good term for it.

ferferga: Been a while without looking into current's web code but shouldn't this be translated?

oddstr13: How about renaming DirectStream to Audio Transcode, as that actually describes what happens?

Artiume: That's a fair compromise. Off the top of my head,  the ffmpeg logs on backend would need to renamed. I'm sure there is other places that'd need it too

PR #2070, opened 2020-11-16
Update config.json

### LINKED ISSUE #2048: Replace remote fonts with local hosted fonts
**Describe The Bug**
Right now fonts are loaded from repo.jellyfin.org which causes CORS error,
Please replace it with Locally hosted files

**Steps To Reproduce**

**Expected Behavior**

**Logs**

**Screenshots**

Screenshot_20201108_132505

 - Platform: Linux
 - Browser: Firefox, Chrome, Brave
 - Jellyfin Version: 10.7.0
**Additional Context**

joshuaboniface:
**Changes**
Disable default downloading of jellyfin-noto fonts. Pending a better solution if one is possible.

**Issues**
Resolves #2048

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -23,10 +23,11 @@
     }
   ],
   "fonts": [
-    "https://repo.jellyfin.org/releases/other/jellyfin-noto/font-faces.css",
-    "https://repo.jellyfin.org/releases/other/jellyfin-noto/css/KR.css",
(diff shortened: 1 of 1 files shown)


joshuaboniface:
Right,

>JSON does not support comments. It was also never intended to be used for configuration files where comments would be needed. Hjson is a configuration file format for humans. Relaxed syntax, fewer mistakes, more comments.

:-/

PR #2078, opened 2020-11-17
Tweak OSD duration display for narrow screen

nyanmisaka:
**Changes**
- Tweak the duration display position in OSD.
- Fix a bug in slider that makes buffered range cannot be displayed.
- Add translations for video aspect ratio.

osd1
DD20971B-4B2C-4361-9D90-3F53D1189C4A


dmitrylyzo:
The slider will move to the right as soon as the current position becomes greater than the remained duration.
Time placeholders should probably have a fixed width and should be moved above the slider for very narrow screens.
Fixed width could be calculated from total duration.

nyanmisaka:
> The slider will move to the right as soon as the current position becomes greater than the remained duration.
> Time placeholders should probably have a fixed width and should be moved above the slider for very narrow screens.
> Fixed width could be calculated from total duration.

Actually I am not so familiar with these web stuffs.
In addition, without deliberate observation, it is hard to aware of this slight change.
Can we leave it as is during the freeze and improve it in future?

src/strings/en-us.json
```diff
@@ -1428,5 +1428,7 @@
     "SubtitleVerticalPositionHelp": "Line number where text appears. Positive numbers indicate top down. Negative numbers indicate bottom up.",
     "Preview": "Preview",
     "LabelMaxMuxingQueueSize": "Max muxing queue size:",
-    "LabelMaxMuxingQueueSizeHelp": "Maximum number of packets that can be buffered while waiting for all streams to initialize. Try to increase it if you still encounter \"Too many packets buffered for output stream\" error in ffmpeg logs. The recommended value is 2048."
+    "LabelMaxMuxingQueueSizeHelp": "Maximum number of packets that can be buffered while waiting for all streams to initialize. Try to increase it if you still encounter \"Too many packets buffered for output stream\" error in ffmpeg logs. The recommended value is 2048.",
+    "AspectRatioCover": "Cover",
+    "AspectRatioFill": "Fill"
```
dkanada, line 1433: Did these not exist yet? That sounds unlikely to me.

YouKnowBlom (approved):

ferferga (approved):

PR #2080, opened 2020-11-18
Add prepare script to allow CI to skip build

thornbill:
**Changes**
* Adds a prepare script that allows the build on install to be skipped by setting a `SKIP_PREPARE` environment variable
* Removes condition that skips install if cache is hit

**Issues**
:crossed_fingers: hopefully the recent CI build failures


thornbill:
*Current approach:* Skips `yarn install` if `node_modules` exists. This seems to be having issues where `node_modules` exists but not all dependencies are available for some reason which causes CI failures.

*This approach:* Always runs `yarn install` which should be very fast ~1s when all packages are current. _But_ we have a `prepare` script defined to run a production build on install this is required to support installing an npm package from a git repository. In order to avoid calling the build during the prepare step, I added a small bash script that checks if a `SKIP_PREPARE` environment variable has been set.

*Alternatives:*
* Include the `--ignore-scripts` parameter when calling `yarn install`
  * Cons: May cause random breakages when a dependency needs to run a script during build (https://github.com/yarnpkg/yarn/issues/7338)
* Remove the prepare script entirely
  * Cons: Breaks the ability to install jellyfin-web via the git repository as a dependency in a separate project. Maybe this doesn't matter though since we are moving away from bundled clients?

EraYaN (approved):

dkanada (approved):

scripts/prepare.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+if [ -z "${SKIP_PREPARE}" ]; then
+    npx gulp --production
```
dkanada, line 4:
```suggestion
    webpack --config webpack.prod.js
```

PR #1994, opened 2020-10-18
Migration to ES6

vitorsemeano:
This is a continuation of the work done in #1807 .

**Changes**
Almost all files of the project, mostly implorts and otimization of lazy imports.

Some modules were also converted into classes for better management.
 
Probably there will be issues left, but minor for the majority of the tests i made.

**Issues**

This PR converts the project into a es6 project.

PR #2085, opened 2020-11-19
Just return original server address in chromecastHelper

### LINKED ISSUE #1971: Chromecast fails if the LocalAddress is inaccessible
**Describe The Bug**
The `getServerAddress` function In `chromecastHelper.js` doesn't verify if the `LocalAddress` in the response of the public server info request works. This can make the chromecast receiver fail.

https://github.com/jellyfin/jellyfin-web/blob/master/src/plugins/chromecastPlayer/chromecastHelper.js#L173-L201

**Steps To Reproduce**
1. Have the server report a faulty value for the local address
2. Start chromecast
3. See error

**Expected Behavior**

Ideally the server should respond with a correct address but it won't hurt to validate it in the webui or chromecast receiver.

**Logs**

None

**Screenshots**

None

 - Platform: Cast
 - Browser: All
 - Jellyfin Version: 10.6.x

**Additional Context**

If possible, we should send both the local address and the current to the cast receiver so we can try connecting to both from the receiver, since it's a different device and might contain different network restrictions.

hawken93:
Remove chromecast serveraddress logic

**Changes**
The file I have changed serves the purpose of asking the server if the chromecast seems to be on the same LAN, and if so it switches over to the address that jellyfin thinks is best. However, this only really works if the LAN is fully routed and this direct line of communication works. Docker and other solutions will use NAT within the LAN. Some servers fail because they are multihomed.

**Issues**
Fixes #1971 
Fixes (probably) https://github.com/jellyfin/jellyfin/issues/4365
Fixes https://github.com/jellyfin/jellyfin/issues/3852
Fixes https://github.com/jellyfin/jellyfin/issues/3712
Fixes https://github.com/jellyfin/jellyfin/issues/2624
Fixes https://github.com/jellyfin/jellyfin/issues/1396


hawken93:
It's only used from plugin.js in the same folder. But I'm rusty in JS and have no idea how to refactor that instance. Help appreciated! Let's get rid of the whole file!

ferferga:
Blocked until #1994 is merged

MrTimscampi (approved):
LGTM, though it'd probably make sense to completely remove the file and just replace the calls to this file with API client calls directly.

Untested, as I don't use Chromecast.

cvium:
If we're going to rip out everything but 1 line, I second MrTim. Just remove the file and fix the calls which most likely will be `getServerAddress(apiClient)` -> `apiClient.serverAddress()`.

hawken93:
I've made an attempt to refactor the call but I'm not sure if I did it correctly :)

hawken93:
I'll be testing the code later on; please don't merge it untested

YouKnowBlom:
Edited description to unlink jellyfin/jellyfin#4365 until we know for sure

YouKnowBlom (commented):
Seems to work well for me. Solves all the CORS issues as well which is a plus, nice work! :)

Holding off on my approval just a bit for some more testing to be done.

hawken93:
I made a 10.6.4 backport and tested it out.
```
                        Stable          Nightly
10.6.4                  Wrongip         Wrongip
10.6.4+workaround       OK              Plays
10.6.4-fix              Plays           Plays
```

10.6.4+workaround is when I return https://jellyfin.example.com instead of the usual localaddress in the System/Info/Public endpoint.
10.6.4-fix is the backported fix.

I should mention that I had lots of trouble with Nightly switching between versions, it would connect but seem to instantly give up playback. Mostly on 10.6.4+workaround. But after clearing everything I had it working.

I don't know how to get debug logs off my chromecast, it seems like a completely opaque playform.

hawken93:
Just clarifying for all the interested parties: This fix affects the code running on the device that you use to cast to your chromecast device. It will send the chromecast the correct endpoint for connecting to your server.

One of the easier ways to get this working is with an old-fashioned install. It has roughly the following steps:
1: check out the code from my branch on here
2: Get `yarn` installed and working, sometimes `npm install -g yarn` is the right command for that.
3: Run `yarn install` in the directory with the code
4: Find jellyfin-web in your jellyfin installation, rename that to something else (jellyfin-web.old)
5: Move and rename the `dist` folder into the same place and name as jellyfin-web in your server installation
6: Start the jellyfin server and clear your browser cache (Especially cookes, cached files and images, site data for jellyfin).

You'll be able to revert back by deleting the jellyfin-web folder after and renaming the old jellyfin-web folder back.

You'll find assistance on the jellyfin community on matrix :)

hawken93:
Just one thing worthy of notice.

If this applies to you:
* You're using LAN only DNS
* You're not overriding the 8.8.8.8 IP and sending it to your LAN DNS
* The LocalAddress field contains an endpoint that will work on your LAN

Then I think this will break chromecast for you. Workarounds are:
* Override 8.8.8.8
* Connect to jellyfin by IP or by any name that 8.8.8.8 will resolve.

YouKnowBlom:
> I don't know how to get debug logs off my chromecast, it seems like a completely opaque playform.

It's a bit of work to get it. You have to register on the cast dashboard, pay $5 and then register your device. Lastly you have to host the receiver files through a web server and then register a new application pointing to those files.

This seems to work better than before despite the issues you listed above at least, so I'm ready to approve

dkanada (approved):

briandipalma:
I pulled the nightly linuxserver Jellyfin image and now the android app doesn't even list my Chromecast. The VLC/Netflix apps both do so it doesn't seem like a network issue.

hawken93:
I don't think this code is related to that problem, jellyfin-web would be what's running in your app..

PR #2095, opened 2020-11-21
Fix epub player issues

thornbill:
**Changes**
Fixes various issues with the epub player

**Issues**
N/A


joshuaboniface (approved):

thornbill:
The "bugs" from SonarCloud are due to using `<i>` elements for the icons, but that seems to be the standard approach we use throughout web. :smile:

dmitrylyzo:
> The "bugs" from SonarCloud are due to using `<i>` elements for the icons, but that seems to be the standard approach we use throughout web. smile

We use `span`.

thornbill:
:+1: Updated. There are a few other places where `<i>` is still used. It seems SonarCloud doesn't find them when they are used in a js file. Moving this from js to an html template caused it to be detected as an issue.

dkanada (approved):

PR #2101, opened 2020-11-22
Take baseurl into account, use original url, not LocalAddress

oddstr13:
Related to #1918

### DIFF
--- a/src/scripts/clientUtils.js
+++ b/src/scripts/clientUtils.js
@@ -67,7 +67,7 @@ export async function serverAddress() {
             };
         }));
     }).then(configs => {
-        let selection = configs.find(obj => !obj.config.StartupWizardCompleted) || configs[0];
+        const selection = configs.find(obj => !obj.config.StartupWizardCompleted) || configs[0];
         return Promise.resolve(selection.url);
     }).catch(error => {
         console.log(error);
--- a/src/scripts/clientUtils.js
+++ b/src/scripts/clientUtils.js
@@ -32,7 +32,7 @@ export async function serverAddress() {
     if (window.location.href.indexOf('/web/') !== -1) { // Do we polyfill String.prototype.includes?
         const split = window.location.href.split('/web/');
 
-        for (let i=split.length - 1; i ; i--) {
+        for (let i = split.length - 1; i ; i--) {
             urls.push(split.slice(0, i).join('/web/'));
         }
     }
@@ -45,13 +45,13 @@ export async function serverAddress() {
 
     urls.push(...await webSettings.getServers());
 
-    console.log("URL candidates:", urls);
+    console.log('URL candidates:', urls);
 
     const promises = urls.map(url => {
         return fetch(`${url}/System/Info/Public`).then(resp => {
             return {
                 url: url,
-                response: resp,
+                response: resp
             };
         }).catch(error => {
             return Promise.resolve();
@@ -63,8 +63,8 @@ export async function serverAddress() {
         return Promise.all(responses.map(obj => {
             return {
                 url: obj.url,
-                config: obj.response.json(),
-            }
+                config: obj.response.json()
+            };
         }));
     }).then(configs => {
         let selection = configs.find(obj => !obj.config.StartupWizardCompleted);
--- a/src/scripts/clientUtils.js
+++ b/src/scripts/clientUtils.js
@@ -29,23 +29,47 @@ export async function serverAddress() {
     // if (current) return Promise.resolve(current);
 
     const urls = [];
+    if (window.location.href.indexOf('/web/') !== -1) { // Do we polyfill String.prototype.includes?
+        const split = window.location.href.split('/web/');
+
+        for (let i=split.length - 1; i ; i--) {
+            urls.push(split.slice(0, i).join('/web/'));
+        }
+    }
     urls.push(window.location.origin);
     urls.push(`https://${window.location.hostname}:8920`);
-    urls.push(...await webSettings.getServers());
 
     if (window.location.protocol === 'http') {
         urls.push(`http://${window.location.hostname}:8096`);
     }
 
+    urls.push(...await webSettings.getServers());
+
+    console.log("URL candidates:", urls);
+
     const promises = urls.map(url => {
-        return fetch(`${url}/System/Info/Public`).then(resp => url).catch(error => {
+        return fetch(`${url}/System/Info/Public`).then(resp => {
+            return {
+                url: url,
+                response: resp,
+            };
+        }).catch(error => {
             return Promise.resolve();
         });
     });
 
     return Promise.all(promises).then(responses => {
-        responses = responses.filter(response => response);
-        return responses[0];
+        responses = responses.filter(obj => obj && obj.response.ok);
+        return Promise.all(responses.map(obj => {
+            return {
+                url: obj.url,
+                config: obj.response.json(),
+            }
+        }));
+    }).then(configs => {
+        let selection = configs.find(obj => !obj.config.StartupWizardCompleted);
+        if (!selection) selection = configs[0];
+        return Promise.resolve(selection.url);
     }).catch(error => {
         console.log(error);
         return Promise.resolve();
--- a/src/scripts/clientUtils.js
+++ b/src/scripts/clientUtils.js
@@ -29,10 +29,10 @@ export async function serverAddress() {
     // if (current) return Promise.resolve(current);
 
     const urls = [];
-    if (window.location.href.indexOf('/web/') !== -1) { // Do we polyfill String.prototype.includes?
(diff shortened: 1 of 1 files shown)


joshuaboniface (approved):

hawken93:
Hi @oddstr13

I like your one better because it restores what I removed from @dkanada's code.
However, I feel like my version is more readable for the /web index splitting. If you agree then feel free to take that code.

Also, in my PR I remove some items from the url list. I tried to justify in my PR.

This PR can supersede mine. :)

src/scripts/clientUtils.js
```diff
@@ -29,20 +29,43 @@ export async function serverAddress() {
...
+    if (window.location.href.indexOf('/web/') !== -1) { // Do we polyfill String.prototype.includes?
+        const split = window.location.href.split('/web/');
+
+        for (let i = split.length - 1; i ; i--) {
+            urls.push(split.slice(0, i).join('/web/'));
+        }
+    }
     urls.push(window.location.origin);
     urls.push(`https://${window.location.hostname}:8920`);
     urls.push(`http://${window.location.hostname}:8096`);
```
hawken93, line 41: In my version I have removed these because I just think they are incorrect. The two last ones cause CORS issues. That's just my opinion though.

src/scripts/clientUtils.js
```diff
@@ -29,20 +29,43 @@ export async function serverAddress() {
     // if (current) return Promise.resolve(current);
 
     const urls = [];
+    if (window.location.href.indexOf('/web/') !== -1) { // Do we polyfill String.prototype.includes?
+        const split = window.location.href.split('/web/');
+
+        for (let i = split.length - 1; i ; i--) {
+            urls.push(split.slice(0, i).join('/web/'));
+        }
+    }
```
hawken93, line 38: I feel like this section is more complicated than it needs to be..

```js
const urlLower = window.location.href.toLowerCase();
const index = urlLower.lastIndexOf('/web');
if (index != -1) {
    urls.push(urlLower.substring(0, index));
}
```

I don't know why we are looping. In my opinion the last /web should suffice.

hawken93 (changes requested):

oddstr13:
I'm actually looping over the split, using this… weird baseurl setup as a test case:
`http://media/foo/web/1070l-jellyfin/web/index.html`

image

hawken93:
I understand.
It's my opinion that only entry 0 is the correct one. I'm assuming that the entire server is shifted in relation to base-url. If this is not the case, I think we can require the user to specify in config.json

oddstr13:
For comparison, this is how the list looks like on the default setup of just going to host:port;
image

The code will pick the first one in the list that hasn't ran the setup wizard, otherwise it'll pick the first one. That'll in any case be `[0]` 99% of the time.
While the CORS error isn't ideal, I don't see how it hurts, other than a kinda scary message in the console.

My intent with this PR isn't to change the base functionality of this function, other than making sure it works with baseurl.

src/scripts/clientUtils.js
oddstr13: looping in case someone had the silly idea of including `/web/` in their baseurl

src/scripts/clientUtils.js
oddstr13: This doesn't really matter to me either way, but I'm leaning more towards leaving as-is, and letting someone else pull them out if it is deemed necessary at a later point in time.

More opinions on this matter is welcome however.

hawken93:
Before the wizard fix PR, think it just assumed 0. It's never been an issue...

I think it's suboptimal for two reasons:
* the CORS issues
* The function being way harder to understand than it could be.

I just feel like it was unjustified to make it do so many weird guesses. I don't know in which case they are actually needed.

src/scripts/clientUtils.js
```diff
@@ -29,20 +29,43 @@ export async function serverAddress() {
...
     urls.push(window.location.origin);
     urls.push(`https://${window.location.hostname}:8920`);
     urls.push(`http://${window.location.hostname}:8096`);
     urls.push(...await webSettings.getServers());
 
+    console.log('URL candidates:', urls);
```
nielsvanvelzen, line 44:
```suggestion
    console.debug('URL candidates:', urls);
```

src/scripts/clientUtils.js
```diff
@@ -29,20 +29,43 @@ export async function serverAddress() {
     // if (current) return Promise.resolve(current);
 
     const urls = [];
+    if (window.location.href.indexOf('/web/') !== -1) { // Do we polyfill String.prototype.includes?
+        const split = window.location.href.split('/web/');
+
+        for (let i = split.length - 1; i ; i--) {
```
nielsvanvelzen, line 35:
```suggestion
        for (let i = split.length - 1; i >= 0; i--) {
```
(maybe `> 0` instead of `>= 0`)

src/scripts/clientUtils.js
```diff
@@ -29,20 +29,43 @@ export async function serverAddress() {
     // if (current) return Promise.resolve(current);
 
     const urls = [];
+    if (window.location.href.indexOf('/web/') !== -1) { // Do we polyfill String.prototype.includes?
```
nielsvanvelzen, line 32:
```suggestion
    if (window.location.href.indexOf('/web/') !== -1) {
```
I don't see babel-polyfill added in the webpack config so probably not

src/scripts/clientUtils.js
```diff
@@ -29,20 +29,43 @@ export async function serverAddress() {
...
+        return Promise.all(responses.map(obj => {
+            return {
+                url: obj.url,
+                config: obj.response.json()
+            };
+        }));
+    }).then(configs => {
+        let selection = configs.find(obj => !obj.config.StartupWizardCompleted);
+        if (!selection) selection = configs[0];
+        return Promise.resolve(selection.url);
```
nielsvanvelzen, line 68:
```suggestion
        return Promise.resolve(configs.find(obj => !obj.config.StartupWizardCompleted) || configs[0]);
```

nielsvanvelzen (changes requested):

src/scripts/clientUtils.js
thornbill: It should be covered by the corejs import... babel-polyfill is deprecated now

src/scripts/clientUtils.js
thornbill: I'm in favor of getting this fix in and then addressing the concerns here in a follow up PR.

src/scripts/clientUtils.js
oddstr13: Is it poor practice to use 0 as boolean false like this in JavaScript?
(should be `> 0` if so)

src/scripts/clientUtils.js
oddstr13: Why? This changes behavior, and will most likely break everything.

hawken93: I think you forgot to return `.url` there in the suggestion, otherwise it's certainly more compact.

src/scripts/clientUtils.js
nielsvanvelzen: I think it makes the code harder to read (but I mostly use type-safe languages so it's probably because of that)

src/scripts/clientUtils.js
nielsvanvelzen: `return Promise.resolve((configs.find(obj => !obj.config.StartupWizardCompleted) || configs[0]).url);`

hawken93:
```js
let selection = configs.find(obj => !obj.config.StartupWizardCompleted) || configs[0];
return Promise.resolve(selection.url);
```

hawken93: I mean, it's a 2-liner but I think it's worth it for the readability.

src/scripts/clientUtils.js
```diff
@@ -29,20 +29,42 @@ export async function serverAddress() {
...
-        return responses[0];
+        responses = responses.filter(obj => obj && obj.response.ok);
+        return Promise.all(responses.map(obj => {
+            return {
+                url: obj.url,
+                config: obj.response.json()
+            };
+        }));
+    }).then(configs => {
+        let selection = configs.find(obj => !obj.config.StartupWizardCompleted) || configs[0];
```
nielsvanvelzen, line 66: Could be `const` now

nielsvanvelzen (approved):

nielsvanvelzen:
Can you rebase/merge master into this branch?

hawken93 (approved):
I'll suggest my changes in a new PR, the code seems to work as intended and that's good enough for me.

nielsvanvelzen (approved):

thornbill (approved):

PR #2116, opened 2020-11-24
Disable chromecast in unsupported browsers

thornbill:
**Changes**
This fixes a regression from https://github.com/jellyfin/jellyfin-web/pull/1502 which caused the chromecast player to no longer be disabled on unsupported browsers. The regression also allowed for chromecast to remain enabled when "remotecontrol" was not supported if in a browser that supports chromecast.

**Issues**
N/A

### DIFF
--- a/src/scripts/site.js
+++ b/src/scripts/site.js
@@ -125,13 +125,13 @@ function loadPlugins() {
     console.groupCollapsed('loading installed plugins');
     console.dir(pluginManager);
     return getPlugins().then(function (list) {
-        // these two plugins are dependent on features
         if (!appHost.supports('remotecontrol')) {
+            // Disable remote player plugins if not supported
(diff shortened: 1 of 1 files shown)


oddstr13 (commented):
The bracket location changes between these two (and thus where the argument `1` goes – is this intentional?
```javascript
list.splice(list.indexOf('sessionPlayer'), 1);
list.splice(list.indexOf('chromecastPlayer', 1));
```

thornbill:
No it wasn't intentional. I copy-pasted the existing line and it had an error. I did confirm it disabled chromecast, but as it was it was probably removing several other plugins also. :smile: 

Should be good now.

oddstr13 (approved):

MrTimscampi (approved):

PR #2122, opened 2020-11-25
Use static imports for html templates

thornbill:
**Changes**
Use static import for html files when possible. This reduces the number of tiny html bundles that are requested at runtime. The html files for controllers are still dynamically imported since that would be a more significant change to the routing logic.

**Issues**
N/A

### DIFF
--- a/src/components/accessSchedule/accessSchedule.js
+++ b/src/components/accessSchedule/accessSchedule.js
@@ -12,6 +12,7 @@ import globalize from '../../scripts/globalize';
 import '../../elements/emby-select/emby-select';
 import '../../elements/emby-button/paper-icon-button-light';
 import '../formdialog.css';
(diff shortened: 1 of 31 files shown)


MrTimscampi (approved):

MrTimscampi (approved):
Re-reviewing just the linter fixes

dmitrylyzo:
`template` code smells relate to this PR and can be easily fixed.

dmitrylyzo (commented):
Should we remove `async` if no `await`?
https://github.com/jellyfin/jellyfin-web/blob/7c46427761e82bd63b0a7a7b7aa7e6a77bca7acd/src/components/dialog/dialog.js#L121

dmitrylyzo (approved):

PR #2125, opened 2020-11-25
Images on Plugins

BaronGreenback:
Uses Plugins/{id}/Image api.

Depends on: https://github.com/jellyfin/jellyfin/pull/4583


Artiume:
Could the plugins provide their own images? Doesn't look like third party repo's could supply their own images in this way.

dkanada:
The images are basically proxied through the server, they aren't actually stored there.

h1dden-da3m0n:
aren't the changes in this PR already superseded by or more accurately included in jellyfin-web#2225?

BaronGreenback:
> aren't the changes in this PR already superseded by or more accurately included in #2225?

Yes it is - sorry - will close this one.

PR #2127, opened 2020-11-26
Remove standalone.js and broken navigation

thornbill:
**Changes**
* Removes standalone.js which no longer serves any purpose
* Removes the call to navigate prior to the router being initialized. This redirect seems to be handled during the router initialization, but I'm not sure if there are some cases that this could still fail?

**Issues**
N/A

### DIFF
--- a/src/scripts/site.js
+++ b/src/scripts/site.js
@@ -26,7 +26,7 @@ import './libraryMenu';
 import './routes';
 import '../components/themeMediaPlayer';
 import './autoBackdrops';
-import { navigate, pageClassOn, serverAddress } from './clientUtils';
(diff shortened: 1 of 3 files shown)


dkanada (approved):

dmitrylyzo:
It was https://github.com/jellyfin/jellyfin-web/blob/d4c8d1a334bfaae627b04abd5e3c8c4f7d203aff/src/scripts/site.js#L99-L102
Maybe I missed something during merge?

Regarding PR: 1 code smell

Camc314 (approved):

MrTimscampi (approved):

PR #2137, opened 2020-11-29
Pass plugin name to pluginManager.loadPlugin

nielsvanvelzen:
**Changes**

The pluginManager will use the import when necessary. By passing just the name we can use other ways to load plugins in the nativeshell.

This is required for jellyfin-android and possibly the MPV shim.

Relevant code:
see https://github.com/jellyfin/jellyfin-web/blob/fb4ac021e40d52dc25a25dd03ade18c72de6ed6c/src/components/pluginManager.js#L71-L101

**Issues**

Linked PR (not dependent): #2139

### DIFF
--- a/src/scripts/site.js
+++ b/src/scripts/site.js
@@ -137,7 +137,7 @@ function loadPlugins() {
         }
 
         Promise.all(list.map((plugin) => {
-            return pluginManager.loadPlugin(import(/* webpackChunkName: "[request]" */ `../plugins/${plugin}`));
+            return pluginManager.loadPlugin(plugin);
         }))
             .then(function () {
                 console.debug('finished loading plugins');


MrTimscampi (commented):
Wouldn't this break the lpugins bundled with web? The idea with how it was done was to have a dynamic expression for the list, since Webpack needs a directory to infer what we're trying to do here (See https://webpack.js.org/api/module-methods/#dynamic-expressions-in-import)

nielsvanvelzen:
The same import statement is used in the plugin manager so I assume it just works, did not test it though.

thornbill (approved):
Confirmed local plugins still load as expected with this change.

MrTimscampi (approved):

Issue #1931, opened 2020-09-16
Unusable player controls for mobile app on small screens


hewham:
iOS app
iPhone SE 2016

The timebar for music tracks on small phone screens (iPhone SE 2016) is completely blocked by the larger buttons on top

Scrubbing through songs (or long podcasts) is impossible, and there is no landscape orientation to get around it

Screenshot tells the story:

![AE15F95C-F95C-4F0D-95E9-DC92C5ED25DA](https://user-images.githubusercontent.com/28872198/93295526-377b6480-f7bb-11ea-8689-1d02834c1cdc.png)

SightSpirit:
I am also having this problem on my Pixel 2 XL (which has a large screen) with album art that is tall enough. (Android 11; app version 2.0.1) There should probably be a maximum height for art in the player view.

PR #2138, opened 2020-11-29
Fix the overlap in iOS music view and the hidden nowPlayingBar

### LINKED ISSUE #1602: iOS music view overflows with long album titles.
64312B6B-5B79-49EB-B064-1FFA2516C246
Longer album titles cause overflow on the new mobile music view in 10.6, at least on iOS, iPhone 8

nyanmisaka:
**Changes**
- Fix the overlap in iOS music view.
- Fix the issue that nowPlayingBar is still hidden after exiting the music control pannel.

Before: 

After: 100%  125%: 

**Issues**
Fixes #1602 #1931

### DIFF
--- a/src/components/nowPlayingBar/nowPlayingBar.js
+++ b/src/components/nowPlayingBar/nowPlayingBar.js
@@ -660,7 +660,7 @@ import { appRouter } from '../appRouter';
         console.debug('nowplaying event: ' + event.type);
         const player = this;
 
-        if (!state.NowPlayingItem || layoutManager.tv || !state.IsFullscreen) {
+        if (!state.NowPlayingItem || layoutManager.tv) {
             hideNowPlayingBar();
             return;
         }
--- a/src/components/nowPlayingBar/nowPlayingBar.js
+++ b/src/components/nowPlayingBar/nowPlayingBar.js
@@ -660,7 +660,7 @@ import { appRouter } from '../appRouter';
         console.debug('nowplaying event: ' + event.type);
         const player = this;
 
-        if (!state.NowPlayingItem || layoutManager.tv) {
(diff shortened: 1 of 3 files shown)


MrTimscampi (approved):

Camc314 (approved):

src/components/nowPlayingBar/nowPlayingBar.js
```diff
@@ -660,7 +660,7 @@ import { appRouter } from '../appRouter';
         console.debug('nowplaying event: ' + event.type);
         const player = this;
 
-        if (!state.NowPlayingItem || layoutManager.tv || !state.IsFullscreen) {
+        if (!state.NowPlayingItem || layoutManager.tv) {
```
thornbill, line 663: This change reverts https://github.com/jellyfin/jellyfin-web/pull/2028. Should we handle that differently somehow?

MrTimscampi: Probably, but the main issue is that it's using the "music" player to play theme songs, which is of course attached to all the audio playing stuff in Jellyfin (playlist, controls, etc).

So short of reworking theme songs, I don't see a good way to handle this :/

dmitrylyzo: `state.IsFullscreen` is only set when playback starts. Probably should be updated in this function https://github.com/jellyfin/jellyfin-web/blob/e56414ebe8876f6ba23a667ddbdc6b0dc8264678/src/components/playback/playbackmanager.js#L1932

Quick fix.
```suggestion
        if (!state.NowPlayingItem || layoutManager.tv || state.IsFullscreen === false) {
```

nyanmisaka: Nice! It works.

thornbill (approved):

PR #1211, opened 2020-05-09
Fix #1210: Add credentials to XHR requests

fhriley:
Please note: https://github.com/jellyfin/jellyfin/pull/3084 must be merged before this PR will work.

PR #2151, opened 2020-11-30
Remove use-credentials flag that breaks cross-origin playback

thornbill:
**Changes**
Fixes a playback issue where a CORS error would be triggered when playing media in standalone web

**Relevant changes**
Partial revert https://github.com/jellyfin/jellyfin-web/pull/1856
Initial change https://github.com/jellyfin/jellyfin-web/pull/1211

**Issues**
N/A


thornbill:
Closing in favor of adding an option in the web config to control this.

PR #2171, opened 2020-12-04
Fix the issue where the bitrate option is always Auto

nyanmisaka:
**Changes**
- Fix the issue where the bitrate option is always Auto.

**Issues**
Introduced in #1993

### DIFF
--- a/src/components/qualityOptions.js
+++ b/src/components/qualityOptions.js
@@ -52,7 +52,7 @@ export function getVideoQualityOptions(options) {
     qualityOptions.push({ name: '360p - 420 kbps', maxHeight: 360, bitrate: 420000 });
 
     if (maxStreamingBitrate) {
-        let selectedIndex = 0;
+        let selectedIndex = qualityOptions.length - 1;
         for (let i = 0, length = qualityOptions.length; i < length; i++) {
             const option = qualityOptions[i];
 
@@ -100,7 +100,7 @@ export function getAudioQualityOptions(options) {
     qualityOptions.push({ name: '64 kbps', bitrate: 64000 });
 
     if (maxStreamingBitrate) {
-        let selectedIndex = 0;
+        let selectedIndex = qualityOptions.length - 1;
         for (let i = 0, length = qualityOptions.length; i < length; i++) {
             const option = qualityOptions[i];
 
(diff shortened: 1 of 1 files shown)


thornbill (approved):

src/components/qualityOptions.js
```diff
@@ -53,7 +53,7 @@ export function getVideoQualityOptions(options) {
 
     if (maxStreamingBitrate) {
         let selectedIndex = -1;
-        for (let i = 0, length = qualityOptions.length; i < length; i++) {
+        for (let i = 1, length = qualityOptions.length; i < length; i++) {
             const option = qualityOptions[i];
 
             if (selectedIndex === -1 && option.bitrate <= maxStreamingBitrate) {
```
dmitrylyzo, line 59: Given that the bitrates are sorted in descending order, we can stop the loop as soon as `option.bitrate <= maxStreamingBitrate` and remove `selectedIndex === -1`. The same for audio.

src/components/qualityOptions.js
```diff
@@ -53,7 +53,7 @@ export function getVideoQualityOptions(options) {
 
     if (maxStreamingBitrate) {
         let selectedIndex = -1;
-        for (let i = 0, length = qualityOptions.length; i < length; i++) {
+        for (let i = 1, length = qualityOptions.length; i < length; i++) {
```
dmitrylyzo, line 56: `auto` is optional, so sometimes `i` must start from `0`.
https://github.com/jellyfin/jellyfin-web/blob/aadd99e617a48ac11e128e8ac0f63d48d42f43cd/src/components/qualityOptions.js#L24-L26

src/components/qualityOptions.js
```diff
@@ -52,19 +52,16 @@ export function getVideoQualityOptions(options) {
     qualityOptions.push({ name: '360p - 420 kbps', maxHeight: 360, bitrate: 420000 });
 
     if (maxStreamingBitrate) {
-        let selectedIndex = -1;
+        let selectedIndex = 0;
```
dmitrylyzo, line 55: Since you removed the fallback below, we should probably init as `qualityOptions.length - 1`. Therefore, if there is no options less than `maxStreamingBitrate`, `selectedIndex` will stay unchanged.

nyanmisaka: Done.

dmitrylyzo (approved):

PR #2174, opened 2020-12-04
Fix comics player

thornbill:
**Changes**
* Fixes invalid import for Swiper
* Fixes potential memory link by not calling `URL.revokeObjectURL()` on generated URLs (refs: https://developer.mozilla.org/en-US/docs/Web/API/URL/revokeObjectURL)
* Fixes double loading of files via `loadPromise` and `load()`
* Removes unused properties in `ArchiveSource`

**Issues**
N/A

### DIFF
--- a/src/plugins/comicsPlayer/plugin.js
+++ b/src/plugins/comicsPlayer/plugin.js
@@ -113,7 +113,7 @@ export class ComicsPlayer {
             loading.hide();
             this.swiperInstance = new Swiper(elem.querySelector('.slideshowSwiperContainer'), {
                 direction: 'horizontal',
-                // loop is disabled due to the lack ofSwiper support in virtual slides
+                // loop is disabled due to the lack of Swiper support in virtual slides
                 loop: false,
                 zoom: {
                     minRatio: 1,
--- a/src/plugins/comicsPlayer/plugin.js
+++ b/src/plugins/comicsPlayer/plugin.js
@@ -198,7 +198,7 @@ class ArchiveSource {
     release() {
         this.files = [];
         /* eslint-disable-next-line compat/compat */
-        this.urls.map(URL.revokeObjectURL);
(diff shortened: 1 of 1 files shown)


src/plugins/comicsPlayer/plugin.js
```diff
@@ -5,6 +5,9 @@ import dialogHelper from '../../components/dialogHelper/dialogHelper';
 import keyboardnavigation from '../../scripts/keyboardNavigation';
 import { appRouter } from '../../components/appRouter';
 import ServerConnections from '../../components/ServerConnections';
+// eslint-disable-next-line import/named, import/namespace
```
dmitrylyzo, line 8: Is this for `No Babel config ...` from ESLint? If so, it will be fixed in my ES6 PR.

src/plugins/comicsPlayer/plugin.js
```diff
@@ -101,41 +106,36 @@ export class ComicsPlayer {
...
-                    });
-                });
+        const downloadUrl = apiClient.getItemDownloadUrl(item.Id);
+        this.archiveSource = new ArchiveSource(downloadUrl);
+
+        return this.archiveSource.load().then(() => {
+            loading.hide();
+            this.swiperInstance = new Swiper(elem.querySelector('.slideshowSwiperContainer'), {
+                direction: 'horizontal',
+                // loop is disabled due to the lack ofSwiper support in virtual slides
```
dmitrylyzo, line 116:
```suggestion
                // loop is disabled due to the lack of Swiper support in virtual slides
```

dmitrylyzo (approved):

src/plugins/comicsPlayer/plugin.js
thornbill: Probably. I just copied it from the image slideshow import.

joshuaboniface (approved):

Issue #2157, opened 2020-12-02
PluginManager fails to load classes


nielsvanvelzen:
**Describe The Bug**

At this line:
https://github.com/jellyfin/jellyfin-web/blob/db6391c6c130839f96463459b5eece236ee77339/src/components/pluginManager.js#L79

We should also check if the property is a class, if it is a class it should **not** await.

It would also be great if the catch statement on this line:

https://github.com/jellyfin/jellyfin-web/blob/75b7b4d7a70c28fc519b56e1a95d4e87537733f0/src/scripts/site.js#L145

Would log the actual error, right now debugging plugins is a PITA because all errors are hidden...

thornbill:
@nielsvanvelzen you are talking about the `await` on this line not the earlier one right?

https://github.com/jellyfin/jellyfin-web/blob/db6391c6c130839f96463459b5eece236ee77339/src/components/pluginManager.js#L80

nielsvanvelzen:
Yup that one, if a function is returned it _might_ be async or return a promise so we want to always `await` it. (if it's not a promise/async it will resolve just fine). But in the case that a class is returned we do not want to call it as a function because that is illegal in JavaScript (need to use the `new` keyword for classes).

thornbill:
I think we can just always use the `new` keyword here since it isn't trivial to determine if a variable is a Class vs a Function https://stackoverflow.com/questions/30758961/how-to-check-if-a-variable-is-an-es6-class-declaration

nielsvanvelzen:
![image](https://user-images.githubusercontent.com/2305178/100925811-ce1a1080-34e2-11eb-8dc9-cecd4fb02169.png)

Adding the `new` keyword should indeed work.

nielsvanvelzen:
Since I'm using functions now for jellyfin-android we could just not support exposing the class directly but I think it's nice to have as an option.

PR #2186, opened 2020-12-06
Fix plugin loader for function definitions in window

Maxr1998:
This reverts commit 2dede415df2558c9be4baa48260a69cc10dea8e5 and fixes supporting functions as well as classes as plugin definitions in window.

The previous change breaks plugin loading on Android, as plugin definitions with a function return a Promise to the class (async function), which cannot be instantiated. The Promise needs to be awaited on first; and the instantiation requires the default dependencies as constructor arguments.

### DIFF
--- a/src/components/pluginManager.js
+++ b/src/components/pluginManager.js
@@ -77,7 +77,7 @@ import { playbackManager } from './playback/playbackmanager';
                     let pluginInstance = await window[pluginSpec];
 
                     if (typeof pluginInstance === 'function') {
-                        pluginInstance = await new pluginInstance();
+                        pluginInstance = await pluginInstance();
                     }
 
                     // init plugin and pass basic dependencies
--- a/src/components/pluginManager.js
+++ b/src/components/pluginManager.js
@@ -74,14 +74,23 @@ import { playbackManager } from './playback/playbackmanager';
             if (typeof pluginSpec === 'string') {
                 if (pluginSpec in window) {
                     console.log(`Loading plugin (via window): ${pluginSpec}`);
-                    let pluginInstance = await window[pluginSpec];
 
-                    if (typeof pluginInstance === 'function') {
-                        pluginInstance = await pluginInstance();
+                    const pluginDefinition = await window[pluginSpec];
+                    if (typeof pluginDefinition !== 'function') {
+                        const err = new TypeError('Plugin definitions in window have to be an (async) function returning the plugin class');
+                        console.error(err);
+                        throw err;
+                    }
+
+                    const pluginClass = await pluginDefinition();
+                    if (typeof pluginClass !== 'function') {
+                        const err = new TypeError(`Plugin definition doesn't return a class for '${pluginSpec}'`);
+                        console.error(err);
+                        throw err;
                     }
 
                     // init plugin and pass basic dependencies
-                    plugin = new pluginInstance({
+                    plugin = new pluginClass({
                         events: Events,
                         loading,
                         appSettings,
--- a/src/components/pluginManager.js
+++ b/src/components/pluginManager.js
@@ -77,16 +77,12 @@ import { playbackManager } from './playback/playbackmanager';
 
                     const pluginDefinition = await window[pluginSpec];
                     if (typeof pluginDefinition !== 'function') {
-                        const err = new TypeError('Plugin definitions in window have to be an (async) function returning the plugin class');
-                        console.error(err);
-                        throw err;
+                        throw new TypeError('Plugin definitions in window have to be an (async) function returning the plugin class');
                     }
 
                     const pluginClass = await pluginDefinition();
                     if (typeof pluginClass !== 'function') {
-                        const err = new TypeError(`Plugin definition doesn't return a class for '${pluginSpec}'`);
-                        console.error(err);
-                        throw err;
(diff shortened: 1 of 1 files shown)


nielsvanvelzen:
This will fail if `pluginInstance` is a class because constructing a class needs the `new` keyword.

See #2157 and #2159

Maxr1998:
But if it is a class, it'll still break because it's missing the arguments/it's calling new twice. We have to fix this differently then..

nielsvanvelzen:
Something like this _might_ work, did not test it:
` if (typeof pluginInstance === 'function' && !pluginInstance.constructor) {`

Maxr1998:
Yes, that could work. I'll try it later and if it does, I'll add it to this PR.

Maxr1998:
Looks like is doesn't, since functions also have the constructor parameter.

Maxr1998:
I just pushed a "fix", the code looks a bit Spaghetti, but *should* be working. Gotta do some additional testing though.

Maxr1998:
Tested on production with the Android app, and it works *just fine* :tada:

oddstr13 (approved):

thornbill:
I'm not a fan of the spaghetti this adds. Maybe we should just straight revert and not support classes directly? Is it bad if we just say for classes you must return them from a function? `window.MyPlugin = () => MyPluginClass`

It would be nice if we had a well defined spec for these plugins and what is supported.

Maxr1998:
As long as there's a documented spec, I'm happy with both options. We just need to decide and document it somewhere.

nielsvanvelzen:
> Maybe we should just straight revert and not support classes directly?


jellyfin-android does use functions because classes didn't work when I implemented it so I guess we can remove it yes.

Maxr1998:
So, only keep the first commit then?

Maxr1998:
I reverted my spagehtti-ish workaround, as we discussed internally to not support the classes directly in window. Instead, that should be documented and this PR could be merged now. However, I still want to use some changes from my earlier commit (4b7ccc40c) to refactor the current code a bit and make it more readable.

src/components/pluginManager.js
```diff
@@ -74,14 +74,23 @@ import { playbackManager } from './playback/playbackmanager';
             if (typeof pluginSpec === 'string') {
                 if (pluginSpec in window) {
                     console.log(`Loading plugin (via window): ${pluginSpec}`);
-                    let pluginInstance = await window[pluginSpec];
 
-                    if (typeof pluginInstance === 'function') {
-                        pluginInstance = await new pluginInstance();
+                    const pluginDefinition = await window[pluginSpec];
```
MrTimscampi, line 78: Not sure how that works on the Android side, maybe you already do this, but wouldn't it be a good idea to make the plugins a property of a "plugins" property on Window? (Maybe even `window.jellyfin.plugins` or something)

Just to make window a bit less dirty.

MrTimscampi (approved):

src/components/pluginManager.js
Maxr1998: I'd be in favor, but I don't know how other clients use the plugins in window.

thornbill: This would probably be doable (I think Android and old desktop are the only clients using these), but it shouldn't be done in this PR since it will be backported as a bugfix.

Maxr1998: Yes, that should be improved in a future version.

src/components/pluginManager.js
```diff
@@ -74,14 +74,23 @@ import { playbackManager } from './playback/playbackmanager';
...
+                    if (typeof pluginDefinition !== 'function') {
+                        const err = new TypeError('Plugin definitions in window have to be an (async) function returning the plugin class');
```
thornbill, line 80: Doesn't need to be async. A normal function should also work fine.

src/components/pluginManager.js
```diff
@@ -74,14 +74,23 @@ import { playbackManager } from './playback/playbackmanager';
...
+                        console.error(err);
```
thornbill, line 81: Do we need to log here when we log errors when an error is thrown by this function?

src/components/pluginManager.js
```diff
@@ -74,14 +74,23 @@ import { playbackManager } from './playback/playbackmanager';
...
+                        throw err;
+                    }
+
+                    const pluginClass = await pluginDefinition();
+                    if (typeof pluginClass !== 'function') {
+                        const err = new TypeError(`Plugin definition doesn't return a class for '${pluginSpec}'`);
+                        console.error(err);
```
thornbill, line 88: Same question here regarding logging.

thornbill (changes requested):

src/components/pluginManager.js
Maxr1998: That's why it's in parentheses, both are supported.

src/components/pluginManager.js
Maxr1998: I did it that way because further down the same function the existing code did the same. Don't know if it's necessary.

Maxr1998: It's in fact not necessary, pushed a commit removing the extra logging calls.

src/components/pluginManager.js
Maxr1998: See above.

thornbill (approved):

PR #2195, opened 2020-12-06
Fix SyncPlay switching to next item in queue

### LINKED ISSUE #2193: SyncPlay breaks when switching to next item in queue
**Describe The Bug**
SyncPlay does not correctly detect the player change when reaching the end of the currently playing item. The player will then play the next item and won't respond to user input. Using `SyncPlay icon` > `Stop local playback`/`Resume local playback` is the only way to correct this.

**Steps To Reproduce**

1. Enable SyncPlay and start playing a TV series;
2. Skip to end of episode and wait for next item;
3. Next item will start playing but player and SyncPlay won't be responsive.

**Expected Behavior**
SyncPlay should keep working as if the `Next track` button has been clicked manually.

**Logs**
After 30 seconds the following will show up:
```
QueueCore.js:186 Error while waiting for `playbackstart` event! setCurrentPlaylistItem Timed out.
```
**System:**
 - Jellyfin Version: 10.7.0-rc1

OancaAndrei:
Quick one liner.

**Changes**
Don't drop a still valid player.

**Issues**
Fixes #2193

### DIFF
--- a/src/components/syncPlay/core/PlaybackCore.js
+++ b/src/components/syncPlay/core/PlaybackCore.js
@@ -70,7 +70,6 @@ class PlaybackCore {
     onPlaybackStop(stopInfo) {
         this.lastCommand = null;
         Events.trigger(this.manager, 'playbackstop', [stopInfo]);
-        this.manager.releaseCurrentPlayer();
     }
 
     /**


OancaAndrei:
Or should I include this in the [next "big" PR](https://github.com/jellyfin/jellyfin-web/pull/1990)?

thornbill:
Smaller PRs are much easier to review. 😉
This also looks like it should make it in 10.7 where the larger change won’t at this point.

thornbill (approved):

MrTimscampi (approved):

PR #2204, opened 2020-12-08
Add basic SyncPlay settings

OancaAndrei:
**Changes**
Add dialog for editing local playback settings related to SyncPlay.
Accessible from `SyncPlay icon > Settings` after joining a group.

Available options:
- Enable/disable active syncing (time correction while playback is unpaused);
- Enable/disable *SpeedToSync*;
- Enable/disable *SkipToSync*;
- Tweak playback sync parameters for *SpeedToSync* and *SkipToSync* (advanced tab);
- Tweak time sync with server (advanced tab).

<details>
<summary>Screenshots</summary>

localhost_8096_web_index html (2)

localhost_8096_web_index html

localhost_8096_web_index html (1)
</details>


OancaAndrei:
Getting this into `10.7` would be a nice touch for UX.

thornbill:
10.7 is limited to bug fixes only at this point so I don’t think this will go in.

OancaAndrei:
I'm not sure why the build is failing. I can merge this branch into `master` with no conflicts on my machine.

dmitrylyzo:
_Answering the question about TV layout (https://github.com/jellyfin/jellyfin-tizen/issues/71#issuecomment-810047846)_

On TV, `Save` button cannot be accessed directly - you have to navigate with keyboard first.
It works, but I think there is no need to display it right away.
SAMSUNG SmartTV Tizen 4 0
It would be better to "unstick" it and place it in the same container with forms. Probably only for TV layout, because for the rest of the layouts the current state looks fine.

thornbill:
@OancaAndrei will you have time to address the comments here sometime?

src/components/syncPlay/core/timeSync/TimeSyncCore.js
```diff
@@ -40,6 +48,32 @@ class TimeSyncCore {
         this.timeSyncServer.forceUpdate();
     }
 
+    /**
+     * Gets the list of available devices for time sync.
+     * @returns {Array} The list of devices.
+     */
+    getDevices() {
+        const devices = [{
```
thornbill, line 56: We should just remove the device setting since the server is the only device.

src/components/syncPlay/core/Settings.js
```diff
@@ -0,0 +1,84 @@
...
+ * Module that manages SyncPlay settings.
+ * @module components/syncPlay/core/Settings
+ */
+
+import { Events, AppStorage } from 'jellyfin-apiclient';
+
+/**
+ * Class that manages SyncPlay settings.
+ */
+class SyncPlaySettings {
```
thornbill, line 11: This abstraction seems unnecessary.

src/components/syncPlay/ui/groupSelectionMenu.js
```diff
@@ -144,6 +153,10 @@ class GroupSelectionMenu {
                 SyncPlay.Manager.haltGroupPlayback(apiClient);
             } else if (id == 'leave-group') {
                 apiClient.leaveSyncPlayGroup();
+            } else if (id == 'settings') {
+                new SyncPlaySettingsEditor(apiClient, SyncPlay.Manager.getTimeSyncCore(), {
```
thornbill, line 157: SonarCloud is failing here because this object instance is never used.

thornbill:
I removed the tabbed interface in the settings dialog. I think it is a better experience to have them in a single place.

Screenshot 2021-09-03 at 01-23-36 Jellyfin

joshuaboniface (approved):

dmitrylyzo:
We have a problem with dialogs with `resolveOnClick` enabled (SyncPlay and PlayOn in the main menu). More precisely, with dialog chains.
The order of `popstate` event and Promise resolving differs time to time (_second opening of SyncPlay settings_). The previous dialog doesn't yet pop state, but the new one pushes its state. So, the previous dialog pops wrong state.

Disabling `resolveOnClick` makes it stable, but the dialog hides before the new one appears.

dmitrylyzo (approved):
With disabled `resolveOnClick` (#2917), it works fine.

PR #2207, opened 2020-12-09
Offline Playback in SyncPlay

OancaAndrei:
**Changes**
Enable SyncPlay to use local files for playback. A dialog is shown to allow the user to open the required files.

Depends on #2204.


thornbill:
At first glance this seems like it maybe outside the scope of functionality we should support in Jellyfin. I think further discussion will be needed on this.

MrTimscampi:
This seems... suboptimal.

It looks to essentially be a local media player which, imo, is completely out of scope. We're not VLC or MPV, we don't play local files. Plus this brings a ton of issues (What if the two files are different? Or the same movie in different versions, for example with different start logos?)

Overall, this seems to be out of scope to me. A proper implementation of "offline playback" for SyncPlay would be to implement offline sync first (ie the ability to sync files from the server to the client). Then the playback manager as a whole (not syncplay itself) would essentially check if it's synced locally, then play that instead to reduce bandwidth.

nvllsvm:
Out of scope for a centralized media management server.

OancaAndrei:
I had to try.

Because of the missing offline database I thought that this would be the simplest way to cover some use-cases where people would like so sync something but the server isn't be able to handle all the streams. 

Maybe I'll split this into a separate web player app.

PR #2209, opened 2020-12-09
P2P media sharing for Hls.js in SyncPlay group

OancaAndrei:
**Changes**
Implement P2P media sharing between members of SyncPlay group. This is done by using [p2p-media-loader](https://github.com/Novage/p2p-media-loader/tree/master/p2p-media-loader-hlsjs#p2p-media-loader---hlsjs-integration) which, given a BitTorrent tracker URL, handles the discovery between peers and shares the media chunks using WebRTC.

- Offloads the server when multiple users are watching the same content.
- Opt-in feature.
- Useful when multiple devices on the same network are streaming from the same remote server.

Depends on #2204.


MrTimscampi:
> given a BitTorrent tracker URL

I hadn't seen this, but imo that makes it a no-go, even as opt-in. I'd like to get @jellyfin/core and especially @joshuaboniface 's opinion on this.

OancaAndrei:
I can see how this feature again falls outside the scope of Jellyfin. I wish there was some form of client-side plugin system I could leverage. :thinking:

thornbill:
We’ve talked about having client side plugins a few times but don’t have a good proposal on how to implement that yet.

dkanada:
@OancaAndrei technically the NativeShell object is supposed to be a tool for plugins to utilize, it just hasn't seen very much work in our history. There's an issue somewhere outlining the plan we had for it.

joshuaboniface:
I agree that all these sort of implementations are out of scope.

Jellyfin is fundamentally a media streaming server. It does not, and should not, have any concept of "client-side" files. Be that local files, P2P files, or anything like that. The videos are on the server and are served to the clients.

What is the underlying issue you're trying to solve @OancaAndrei with these?

joshuaboniface:
I should clarify there - something like an offline cache in an app is OK of course, but I mean using a server to interface with local files that the server doesn't know about.

thornbill:
Closing based on the comments above.

OancaAndrei:
I'm back :)

> What is the underlying issue you're trying to solve @OancaAndrei with these?

Simply put, I'd like to save internet bandwidth and data usage.

After re-reading your replies I have the feeling that I've might presented this feature in the wrong way. So, fresh start.

First of all, this feature is completely unrelated to the other *WebRTC* one (that focuses on time/playback syncing), it just happens that they share the same technology to exchange data. Regarding the *IP multicast* proposal, I don't think it fits on that one; maybe it could on this one but I'd need to check (in a moment you'll read why).\
And it's not related to the offline-playback feature either.

### Now, onto a practical example.
- **Scenario**: remote server accessed through a single internet connection. Multiple devices, connected to the same local network, are attempting to stream the same media from the remote server.
- **Problem**: after a certain number of streams the internet connection bandwidth will become saturated thus limiting the number of devices that can use SyncPlay simultaneously.
- **Solution**: let one client download the chunks directly from the server and then share those chunks with the other local devices.

### Following, a *technical* view of the solution.
Before jumping into the code, I've looked around for existing implementations and stumbled upon [Peer5](https://www.peer5.com/) and [cdnbye](https://p2p.cdnbye.com/en/), both being paid services that offer something too big for what's needed. Then I luckily found [p2p-media-loader](https://github.com/Novage/p2p-media-loader) that basically offers the core feature of the two services mentioned previously.

**p2p-media-loader** sits between the server and the *Hls.js* player and fetches the required chunks (the `.ts` files) either from peers or the server, preferring the former when possible.\
Chunks are shared between peers using WebRTC connections. To establish a WebRTC connection some form of signaling is required and **p2p-media-loader** opted to leverage BitTorrent trackers (kinda smart IMO as they're popular and it avoids reinventing the wheel).

### Finally, onto your concerns.
> Jellyfin is fundamentally a media streaming server. It does not, and should not, have any concept of "client-side" files. Be that local files, P2P files, or anything like that. The videos are on the server and are served to the clients. Thus any and all portions of syncplay must be synchronized on the server and then sent to the various configured clients.

I totally agree. And that's why there's no server-side PR of this feature: it's all client side. As I wrote in some other comment, SyncPlay's only focus is playback syncing and does not care about fetching the media / how the media is fetched. This P2P media sharing feature fits perfectly with SyncPlay because the latter creates the opportunity for multiple devices to be watching the same content.

> given a BitTorrent tracker URL

Also, I now see why this line was concerning to you all: Jellyfin is definitely NOT becoming a torrent streaming service (not because of this PR at least).

The "peers" I've been talking about earlier are not "torrent peers" but merely "local network peers" that are streaming the same media, either directly from the server over HTTP or from a WebRTC peer.\
The tracker URL is needed by the clients to discover each other (WebRTC signaling); data transfers happen over WebRTC therefore the clients won't be participating in any "real torrent traffic". The tracker server can be hosted anywhere, probably on the same network of the peers if used for this specific scenario.

> Ideally, this would be implemented with some sort of IP multicast

This would surely help when the clients can be reached this way and should be preferred over P2P (or maybe even using them both could be something).

> But there are so many problems and hacks that will be needed to get the server in sync with random client files (whatever their source) that I'm extremely hesitant about this.

And this won't be a problem with this proposed feature, as the shared file would still be generated by the server itself.

### Advantages, as I see them
- Reduces data/bandwidth usage:
  - server can handle more clients;
  - clients consume less of their internet;

The only hassle is having to provide this additional tracker server.

A form of client plugin would have been cool because this feature could be used even outside SyncPlay and not only for devices that are on the same network. At the moment this would only work on web-clients as **p2p-media-loader** is for browsers.

I really hope to have shed a more positive light on what I was attempting with this PR.\
Curious to hear your thoughts. :grin:

PR #2210, opened 2020-12-09
Set the step of subtitle offset slider to 0.1

nyanmisaka:
**Changes**
- Set the step of subtitle offset slider to 0.1

Browsers such as Safari that often have problems with subtitle synchronization need this to make more precise adjustments.

### DIFF
--- a/src/components/subtitlesync/subtitlesync.js
+++ b/src/components/subtitlesync/subtitlesync.js
@@ -45,11 +45,11 @@ function init(instance) {
             let inputOffset = /[-+]?\d+\.?\d*/g.exec(this.textContent);
             if (inputOffset) {
                 inputOffset = inputOffset[0];
+                inputOffset = parseFloat(inputOffset);
+                inputOffset = Math.min(30, Math.max(-30, inputOffset));
 
                 // replace current text by considered offset
                 this.textContent = inputOffset + 's';
-
-                inputOffset = parseFloat(inputOffset);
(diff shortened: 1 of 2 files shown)


src/components/subtitlesync/subtitlesync.js
```diff
@@ -45,11 +45,11 @@ function init(instance) {
             let inputOffset = /[-+]?\d+\.?\d*/g.exec(this.textContent);
             if (inputOffset) {
                 inputOffset = inputOffset[0];
+                inputOffset = parseFloat(inputOffset);
+                inputOffset = Math.min(30, Math.max(-30, inputOffset));
```
thornbill, line 49: Why is this bounded between -30 and 30?

nyanmisaka: https://github.com/jellyfin/jellyfin-web/blob/4243fd7f47c00b15de91b78a72a01382418a7de2/src/components/subtitlesync/subtitlesync.js#L113-L114

https://github.com/jellyfin/jellyfin-web/blob/4243fd7f47c00b15de91b78a72a01382418a7de2/src/components/subtitlesync/subtitlesync.js#L119-L120

Only ±30s are valid values.
Without this boundary, even if the user inputs 1000s, the actual offset is still 30s.

thornbill (approved):

dkanada (approved):

Issue #2088, opened 2020-11-20
[Unstable] Control Icons on top of each other


Artiume:
![test](https://user-images.githubusercontent.com/20715731/99778982-67f4bd00-2ae2-11eb-9347-efb9158d6996.png)

PR #2218, opened 2020-12-10
Fix style issues on dashboard page

thornbill:
**Changes**
* Adjusts the column layout on the dashboard page to avoid tiny columns on small screens
* Adds a min-width to the session cards to avoid tiny cards on small screens
* Adds top/bottom margins for stacked emby-buttons

**Before:**
Screenshot_2020-12-10 Jellyfin

**After:**
Screenshot_2020-12-10 Jellyfin(1)

**Issues**
Fixes: #2088

### DIFF
--- a/src/assets/css/dashboard.css
+++ b/src/assets/css/dashboard.css
@@ -287,7 +287,7 @@ div[data-role=controlgroup] a.ui-btn-active {
 }
 
 .activeSession {
-    min-width: 15rem;
+    min-width: 20rem;
     width: 100% !important;
 }
 
--- a/src/assets/css/dashboard.css
+++ b/src/assets/css/dashboard.css
@@ -127,8 +127,8 @@ div[data-role=controlgroup] a.ui-btn-active {
 }
 
 .sessionAppInfo img {
-    max-width: 2.25rem;
-    max-height: 2.25rem;
+    max-width: 40px;
+    max-height: 40px;
     margin-right: 8px;
 }
 
@@ -202,6 +202,10 @@ div[data-role=controlgroup] a.ui-btn-active {
     -webkit-box-flex: 1;
     -webkit-flex-grow: 1;
     flex-grow: 1;
+}
+
+.dashboardActionsContainer {
+    margin: 1em -0.3em 0;
 }
 
 .sessionNowPlayingContent {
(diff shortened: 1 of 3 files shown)


nyanmisaka:
Tried on 3840x2160 225% screen.

dkanada (dismissed):

thornbill:
I doubt I will be able to reproduce the issue above because it’s probably specific to the resolution and pixel ratio. Increasing the min-width would probably fix it. I experimented with values from 10-15 rem. Maybe 12 would work?

nyanmisaka:
Change to 14rem seems to be fine on edge. But session cards on iPad are too big than before.
Can we just reduce the font and icon size in session cards to avoid this issue?

Before this PR:
IMG_0457
After:
IMG_0456

thornbill:
This still needs some tweaking as gap has poor browser support.

thornbill:
This should be good for review now.

nyanmisaka:
The card will still be enlarged when the sessions is more than 2.
Can we set a maximum width for the card? 
Everything else is fine.

thornbill:
@nyanmisaka I'm not seeing that now, could you grab a screenshot of that? Thanks!

nyanmisaka:
> @nyanmisaka I'm not seeing that now, could you grab a screenshot of that? Thanks!

https://github.com/jellyfin/jellyfin-web/pull/2218#issuecomment-743362824
The latter is very empty when it is not playing. And it occupies almost the entire page. It would be fine if it were the former.

ferferga (approved):

thornbill:
@nyanmisaka would you care to check how this looks now? I don't think they seem _too_ large now, but I can still tweak it if needed!

nyanmisaka (approved):
It looks good for me.

src/assets/css/dashboard.css
```diff
@@ -291,6 +287,7 @@ div[data-role=controlgroup] a.ui-btn-active {
 }
 
 .activeSession {
+    min-width: 15rem;
```
nyanmisaka, line 290:
```suggestion
    min-width: 20rem;
```

nyanmisaka: 15rem is still a bit narrow on iPad safari.

15rem
20rem

nyanmisaka (changes requested):

joshuaboniface (approved):

PR #2234, opened 2020-12-14
Use Noto Sans from Fontsource

thornbill:
**Changes**
* Removes jellyfin-noto
* Uses noto-sans font packages from [fontsource](https://github.com/fontsource/fontsource)
* Uses custom stylesheets to only include woff2 files for fonts

**Issues**
Fixes the vertical alignment issues of jellyfin-noto (particularly noticeable on buttons and badges)


src/assets/css/fonts.scss
```diff
@@ -1,5 +1,7 @@
+@import "../../styles/noto-sans/index.scss";
+
 @mixin font($weight: null, $size: null) {
-    font-family: "Noto Sans", sans-serif;
+    font-family: "Noto Sans", "Noto Sans HK", "Noto Sans JP", "Noto Sans KR", "Noto Sans SC", sans-serif;
```
MrTimscampi, line 4: Warning that some of these have conflicting characters (JP and SC, from memory).

Some of the kanji are different between the two scripts, but share the same Unicode number.

MrTimscampi (approved):

src/assets/css/fonts.scss
thornbill: Yeah I found this issue https://github.com/jellyfin/jellyfin-web/issues/913 and read through the notes there. I guess we should add a setting to set a hierarchy or preferred character set somewhere.

EraYaN: KR also has all the Hanja (full characters) in it, so this needs to be set on a per language basis.

thornbill: I was thinking we probably shouldn't just rely on the display language since some people may have a larger amount of content in a specific language that is different than their display language. Then they would probably have a preference for which language should be preferred.

This _could_ be done with custom css until a proper preference is added.

nyanmisaka:
There are still some vertical alignment issues in TextButton. Especially in mobile safari.
IMG_9036

thornbill:
@nyanmisaka that seems to not be caused by the font itself, but the styles that are being applied.

ferferga (approved):

PR #2266, opened 2020-12-26
remove touch controls from video player

dkanada:


src/controllers/playback/video/index.js
```diff
@@ -1466,15 +1417,9 @@ import { appRouter } from '../../../components/appRouter';
             passive: true
         });
 
-        if (browser.touch) {
-            dom.addEventListener(view, 'dblclick', onDoubleClick, {});
-        } else {
-            const options = { passive: true };
-            dom.addEventListener(view, 'dblclick', (e) => {
-                if (e.target !== view) return;
```
dmitrylyzo, line 1474: Removing this line breaks #2006
As I said on Matrix, need to add `pointer-events: none;` to `syncPlayContainer`, because the latter overlaps the view and grabs `dblclick`.

src/controllers/playback/video/index.js
```diff
@@ -1466,15 +1417,9 @@ import { appRouter } from '../../../components/appRouter';
...
-                playbackManager.toggleFullscreen(currentPlayer);
-            }, options);
-        }
+        dom.addEventListener(view, 'dblclick', function () {
+            playbackManager.toggleFullscreen(currentPlayer);
```
dkanada, line 1421:
```suggestion
            if (e.target !== view) return;
            playbackManager.toggleFullscreen(currentPlayer);
```

src/controllers/playback/video/index.js
```diff
@@ -1466,15 +1417,10 @@ import { appRouter } from '../../../components/appRouter';
...
-        if (browser.touch) {
-            dom.addEventListener(view, 'dblclick', onDoubleClick, {});
-        } else {
-            const options = { passive: true };
-            dom.addEventListener(view, 'dblclick', (e) => {
-                if (e.target !== view) return;
-                playbackManager.toggleFullscreen(currentPlayer);
-            }, options);
-        }
+        dom.addEventListener(view, 'dblclick', function () {
```
dkanada, line 1420:
```suggestion
        dom.addEventListener(view, 'dblclick', (e) => {
```

dmitrylyzo (approved):

thornbill:
Why are these controls being removed?

dkanada:
Touch controls on an HTML video player are already unreliable in my experience, but these are especially troublesome. The Android app is also migrating over to ExoPlayer (and native touch controls) for a much better experience, so I consider it an improvement to remove these controls on the web player.

thornbill:
Based on our discussion on Matrix, I don't object to removing this, but we need an issue to track the following:

- [ ] Known issues with the current implementation that justify its removal
- [ ] Functional requirements for a new implementation

dkanada:
#2377

thornbill (approved):

Issue #2267, opened 2020-12-27
TA (Tamil) Language missing in 10.7.0 rc1 (Jellyfin & Jellyfin-web)


Oatavandi:
**Describe The Bug**
Tamil Language (TA) is not showing up in the 10.7.0 RC1.
I tried to select the language from user Profile screen and it is not showing up.

I am the one who translated this for Jellyfin and jellyfin-web, and it is upto date.
It is available in the jellyfin and jellyfin-web package, but not showing up in user setting language list.

**Steps To Reproduce**
1. Login into Jellyfin web
2. User Settings -> Display - Display Language - "Tamil" language is not showing up.
4. Admin -> General - Display Language - "Tamil" language is not showing up.

**Expected Behavior**
Tamil language should be shown in the list to be selected.

**Screenshots**

**System (please complete the following information):**
 - Platform: Docker
 - Browser: Safari
 - Jellyfin Version: 10.7.0 RC1

Oatavandi:
Seems the template html needs to be updated with all the available languages

/jellyfin-web-10.7.0-rc1/src/components/displaySettings/displaySettings.template.html

Oatavandi:
This is closed with the fix for **"Add missing Languages to Web #2280"**

PR #2280, opened 2020-12-28
Add missing Languages to Web

Artiume:
**Changes**

**Issues**

https://github.com/jellyfin/jellyfin-web/issues/2267


src/components/displaySettings/displaySettings.template.html
```diff
@@ -77,23 +93,29 @@ <h2 class="sectionTitle">
...
+            <option value="fil">Filipino</option>
             <option value="fi">Finnish</option>
             <option value="fr">French</option>
             <option value="fr-CA">French (Canada)</option>
+            <option value="gl">Galician</option>
             <option value="de">German</option>
             <option value="el">Greek</option>
             <option value="he">Hebrew</option>
-            <option value="hi-IN">Hindi (India)</option>
+            <option value="hi-IN">Hindi</option>
```
dkanada, line 107: I assume the suffix is applied to any code with a dash, so this makes sense to me. Why remove it?

Artiume: To match what is on Weblate. We could split it and add normal Hindi, not sure if Weblate would then re-add the parenthesis

www.lingoes.net/en/translator/langcode.htm

Artiume: There isn't any other language codes for Hindi so i'm not sure if we should even add it as I think the India got appended because someone selected hi-IN instead of hi

dkanada: Matching weblate is a good reason, sounds good.

thornbill (changes requested):
New languages also need added here: https://github.com/jellyfin/jellyfin-web/blob/master/src/scripts/site.js#L71

New date locales need to be mapped appropriately here: https://github.com/jellyfin/jellyfin-web/blob/master/src/scripts/dfnshelper.js#L6

Artiume:
for locales, would i use 'bn_bd': bn, or use a dash?

Artiume:
esperato has no locale, ignore it or give it something else?

thornbill:
It looks like bn and eo should work https://github.com/date-fns/date-fns/tree/master/src/locale

src/components/displaySettings/displaySettings.template.html
```diff
@@ -63,50 +78,65 @@ <h2 class="sectionTitle">
     <div class="selectContainer fldDateTimeLocale hide">
         <select is="emby-select" class="selectDateTimeLocale" label="${LabelDateTimeLocale}">
             <option value="">${Auto}</option>
+            <option value="af">Afrikaans</option>
+            <option value="sq">Albanian</option>
             <option value="ar">Arabic</option>
-            <option value="be-BY">Belarusian (Belarus)</option>
-            <option value="bg-BG">Bulgarian (Bulgaria)</option>
+            <option value="be-BY">Belarusian</option>
+            <option value="bn_BD">Bengali (Bangladesh)</option>
```
thornbill, line 85: Some of these don't exist as date-fns locales. Should they all be?

Also some of them don't match the name of the locales in dfnshelper. There seems like there may be another place where they get normalized as some of the existing locales also don't match, but I'm not familiar with where or how that happens.

Artiume: I was hoping you knew lol. I noticed it was being parsed differently so I wasn't sure how to verify my work.

joshuaboniface (approved):

src/components/displaySettings/displaySettings.template.html
ferferga: Our locales are here: https://github.com/jellyfin/jellyfin-web/blob/master/src/scripts/site.js#L71

The strings in dfnshelper matches those locales using our code as a key with the date fns locale object as value

Artiume: mr wasnt a locale listed so i removed it, it should be good after that.

thornbill (approved):

PR #2287, opened 2020-12-30
Upgrade Custom Elements API to v1 specification

grafixeyehero:
Note: This need to be test on old browser


grafixeyehero:
closing it has some issue on older browser

PR #2225, opened 2020-12-13
Plugin manager changes

BaronGreenback:
Works alongside See https://github.com/jellyfin/jellyfin/pull/4709.

This PR basically adds the enable/disable/rollback options for the user - as well as providing the visual status of each plugin.

_Plugin status images are demo's, are totally optional, and can be changed._

image

PR #2288, opened 2020-12-31
Bump API client to 1.6.0

joshuaboniface:
**Changes**
Bump API client to new version. Needed for #2225.

**Issues**
N/A


thornbill (changes requested):
The lockfile also has to be updated for this. The CI tasks are failing because the lockfile is out of sync with package.json.

thornbill (approved):

barronpm (approved):

PR #2289, opened 2021-01-01
Fix mkv direct play in chromium based browsers

nyanmisaka:
**Changes**
- Fix mkv direct play in chromium based browsers

**Issues**
`canPlayType('video/x-matroska')` and `canPlayType('video/mkv')` cannot check MKV support properly.

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -159,7 +159,10 @@ import browser from './browser';
             return true;
         }
 
-        if (browser.edgeChromium && browser.windows) {
+        // There's no proper method to check MKV in web browsers.
+        // Chromium based browsers support MKV container.
+        // Chrome on iOS/iPad OS is regarded as safari.
+        if (browser.chrome || browser.edg || browser.edga) {
             return true;
         }
 


Artiume:
I thought we had disabled this because some videos were causing the playback to freeze. I think it was videos with chapters?

nyanmisaka:
> I thought we had disabled this because some videos were causing the playback to freeze. I think it was videos with chapters?

Where's that PR?

Artiume:
https://github.com/jellyfin/jellyfin-web/pull/1369

Here's the PR. I can't locate the issue ticket.

nyanmisaka:
There is already a fix for the missing mkv support in Edge chromium. I have direct played many videos with chapter info on it.

It is not clear why chrome was removed in that PR. AFAIK both plex and emby can direct play the video of the MKV container in chrome.

Artiume:
let's try it then. if it ends up having issues, it's an easy fix.

dkanada:
@nyanmisaka backport?

nyanmisaka:
> @nyanmisaka backport?

Sure. I don't have right to add the tag here.

MrTimscampi:
Chrome supports WebM but not MKV proper. The two standards are different and, while it may work for some files, it won't for others.

Imo, we shouldn't treat MKV as a WebM, as it is bound to cause issues in the future for some users.

dkanada (approved):

thornbill:
Closing this due to the points mentioned above. If anyone objects, we can revisit.

PR #2290, opened 2021-01-01
Add Chromecast error messages to the locales

MrTimscampi:
**Changes**

**Issues**


nielsvanvelzen (approved):

Camc314 (approved):

jobhh:
Is it by design that MessagePlaybackError does not specify chromecast like the other entry?

MrTimscampi:
It's what the code currently uses, but that message is only emitted when casting.

PR #2309, opened 2021-01-05
fix(card): white flashing images

ferferga:
~~Not sure why the card's image containers have been changed to buttons, but this solves the white flashing that happens on cards when images are placed into the background.~~

~~Not sure if other components with images that doesn't contain the ``cardImageContainer`` class were changed to buttons, those are not fixed here then.~~

EDIT: Turns out we started setting background colors in the lazy image loader. Now we change canvas opacity instead of messing with background colors. **Untested though, as I don't have any image in my library that has transparency**.

### DIFF
--- a/src/components/images/imageLoader.js
+++ b/src/components/images/imageLoader.js
@@ -87,9 +87,6 @@ import './style.css';
             requestAnimationFrame(() => {
                 if (elem.tagName !== 'IMG') {
                     elem.style.backgroundImage = "url('" + url + "')";
-                    if (elem.classList.contains('blurhashed')) {
-                        elem.style.backgroundColor = '#fff';
-                    }
                 } else {
                     elem.setAttribute('src', url);
                 }
@@ -101,6 +98,12 @@ import './style.css';
                 } else {
                     elem.classList.add('lazy-image-fadein');
                 }
+
+                const canvas = elem.previousSibling;
+                if (elem.classList.contains('blurhashed') && canvas && canvas.tagName === 'CANVAS') {
+                    canvas.classList.remove('lazy-image-fadein-fast', 'lazy-image-fadein');
+                    canvas.classList.add('lazy-hidden');
+                }
             });
         });
     }
@@ -111,7 +114,6 @@ import './style.css';
         if (elem.tagName !== 'IMG') {
             url = elem.style.backgroundImage.slice(4, -1).replace(/"/g, '');
             elem.style.backgroundImage = 'none';
-            elem.style.backgroundColor = null;
         } else {
             url = elem.getAttribute('src');
             elem.setAttribute('src', '');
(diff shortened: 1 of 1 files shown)


src/components/images/imageLoader.js
```diff
@@ -87,9 +87,6 @@ import './style.css';
             requestAnimationFrame(() => {
                 if (elem.tagName !== 'IMG') {
                     elem.style.backgroundImage = "url('" + url + "')";
-                    if (elem.classList.contains('blurhashed')) {
-                        elem.style.backgroundColor = '#fff';
```
nielsvanvelzen, line 91: I think this was done to fix transparency issues

ferferga: @nielsvanvelzen Yeah, I noticed now after tracking where the change came from. Pushed a commit that changes the canvas opacity, that should work. Can't test it as I don't have any image with transparency

src/components/cardbuilder/cardBuilder.js
```diff
@@ -1336,9 +1336,9 @@ import ServerConnections from '../ServerConnections';
                 cardImageContainerClose = '</div>';
             } else {
                 // Don't use the IMG tag with safari because it puts a white border around it
-                cardImageContainerOpen = imgUrl ? ('<button data-action="' + action + '" class="' + cardImageContainerClass + ' ' + cardContentClass + ' itemAction lazy" data-src="' + imgUrl + '" ' + blurhashAttrib + '>') : ('<button data-action="' + action + '" class="' + cardImageContainerClass + ' ' + cardContentClass + ' itemAction">');
+                cardImageContainerOpen = imgUrl ? ('<div data-action="' + action + '" class="' + cardImageContainerClass + ' ' + cardContentClass + ' itemAction lazy" data-src="' + imgUrl + '" ' + blurhashAttrib + '>') : ('<div data-action="' + action + '" class="' + cardImageContainerClass + ' ' + cardContentClass + ' itemAction">');
```
thornbill, line 1339: Why change this? A button feels more semantically correct for an element with an action attached to it on click.

src/components/images/imageLoader.js
thornbill: I can test. The change was made specifically for channel images for live tv.

src/components/cardbuilder/cardBuilder.js
ferferga: @thornbill At first, I thought it was caused because it was a button element, later I realized your changes in image loader. I left this as the div per-se it's not the card element or the play button, it's the image container, so i think it makes sense that it stays as a div instead. The ``cardOverlayContainer`` element will already handle the click action.

With this, this whole if/else can be refactored to a single line now that we don't distinguish between elements, but I didn't do that because I don't think we should change that in a backport, just in case bad stuff happens.

src/components/images/imageLoader.js
thornbill: No idea when it changed, but it seems like blurhashes are no longer being generated for the channel logos. :tada:

src/components/cardbuilder/cardBuilder.js
thornbill: My main concern would be if this has unintended side effects for keyboard navigation or screen readers. The card itself often has an action defined for it in addition to the buttons that are overlayed.

ferferga: @thornbill Good point, and, as I mentioned before, better to not touch unrelated things because we can know what can happens, so better not to backport this stuff :)

Force pushed without those modifications.

ferferga:
@thornbill Didn't notice anything different (aside from the animation of the canvas hiding as well) but now that I look closer, it's true!

thornbill (approved):

nielsvanvelzen (approved):

PR #2312, opened 2021-01-05
Use local version of Noto Sans if available

jarnedemeulemeester:
**Changes**
Use the locally installed version of Noto Sans if available instead of downloading it.

**Issues**

### DIFF
--- a/src/styles/noto-sans/index.scss
+++ b/src/styles/noto-sans/index.scss
@@ -1,3 +1,8 @@
+@use "sass:map";
+
+// Map font weight numbers to names
+$weight-names: (100: "Thin", 200: "ExtraLight", 300: "Light", 400: "Regular", 500: "Medium", 600: "SemiBold", 700: "Bold", 800: "ExtraBold", 900: "Black");
+
 // Mixin to help create the Noto Sans font-faces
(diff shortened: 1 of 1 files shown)


thornbill:
This is great!

I needed to add the weight at the end of the font name to get it load the local version though. So instead of "Noto Sans" it needed to be "Noto Sans Regular". We may need to support multiple local names depending on where and how it is installed.

jarnedemeulemeester:
> I needed to add the weight at the end of the font name to get it load the local version though. So instead of "Noto Sans" it needed to be "Noto Sans Regular". We may need to support multiple local names depending on where and how it is installed.

Good point, I'll look into it tomorrow!

jarnedemeulemeester:
Tested and working on:

Arch Linux (5.10.4) (Noto Sans installed using noto-fonts package):
- Chromium 87.0.4280.88
- Firefox 84.0.1
- Gnome Web 3.38.2

Windows 10 (2004) (Noto Sans installed from Google Fonts):
- Google Chrome 87.0.4280.88
- Microsoft Edge 87.0.664.66
- Firefox 84.0.2

thornbill (approved):

nielsvanvelzen (approved):

PR #2323, opened 2021-01-11
Fix replay icon not getting replaced with play_arrow icon

### LINKED ISSUE #2279: Play button fails to replace resume button
10.6.4 and Firefox

If you go to a media's page that is partially watched and has a resume button, marking it as Watched will cause the resume button to disappear but the play button never populates so you have to refresh the page inorder to watch it.

jarnedemeulemeester:
**Changes**
Change the replay icon to a play_arrow icon when a user marks a media as watched.

**Issues**
Fixes #2279


thornbill (approved):

nielsvanvelzen (approved):

ferferga (approved):
LGTM

PR #2327, opened 2021-01-12
Edit admin dashboard menu for plugins

MrChip53:
**Changes**
I think it would look a little more tidy in the menu if un sorted plugin menu items were under their own section. Also, would like to be able to add plugin items under Live TV(Maybe should allow this for all sections?).

Uhh sorry if its trash code and if you all don't want it thats A-OK 😅

**Issues**

### DIFF
--- a/src/scripts/libraryMenu.js
+++ b/src/scripts/libraryMenu.js
@@ -464,6 +464,7 @@ import Headroom from 'headroom.js';
             pageIds: ['liveTvSettingsPage'],
             icon: 'dvr'
         });
+        addPluginPagesToMainMenu(links, pluginItems, 'livetv');
         links.push({
             divider: true,
             name: globalize.translate('TabAdvanced')
@@ -503,6 +504,10 @@ import Headroom from 'headroom.js';
             href: '#!/scheduledtasks.html',
             pageIds: ['scheduledTasksPage', 'scheduledTaskPage'],
             icon: 'schedule'
+        });
+        links.push({
+            divider: true,
+            name: globalize.translate('TabPlugins')
         });
         addPluginPagesToMainMenu(links, pluginItems);
         return links;
--- a/src/scripts/libraryMenu.js
+++ b/src/scripts/libraryMenu.js
@@ -512,7 +512,6 @@ import Headroom from 'headroom.js';
             });
             addPluginPagesToMainMenu(links, pluginItems);
         }
-        
         return links;
     }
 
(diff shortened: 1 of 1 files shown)


MrChip53:
No Plugins:

Screenshot (2)

With unsorted and live tv plugins:

Screenshot (3)

src/scripts/libraryMenu.js
```diff
@@ -504,10 +505,27 @@ import Headroom from 'headroom.js';
...
+                divider: true,
+                name: globalize.translate('TabPlugins')
+            });
+            addPluginPagesToMainMenu(links, pluginItems);
+        }
         return links;
     }
 
+    function hasUnsortedPlugins(pluginItems) {
+        for (let i = 0, length = pluginItems.length; i < length; i++) {
```
nielsvanvelzen, line 519: I think you can do `for (const pluginItem of pluginItems) {` here

nielsvanvelzen (changes requested):

MrTimscampi:
Some thoughts about UX, which may or may not be valid:

Currently all the config pages are in one place only. It's easy to find them since they all end up in the same place.

Allowing plugins to add themselves to the menu leads to worse UX, imo, since now plugins can end up wherever (I know the feature is already there, but nothing really used it until now).

Furthermore, with the expected plans for the v2 API, plugin pages will be auto-generated from JSON, meaning they'll essentially only be a simple form for configuration. This reduces the utility of this quite a bit, imo.

MrChip53:
> Furthermore, with the expected plans for the v2 API, plugin pages will be auto-generated from JSON, meaning they'll essentially only be a simple form for configuration. This reduces the utility of this quite a bit, imo.

This seems like a major regression and severely limits the capabilities of plugins. Can this be an optional feature just used to maybe lower the bar of entry for creating a plugin?

MrTimscampi:
> Can this be an optional feature just used to maybe lower the bar of entry for creating a plugin?

The current way plugin pages work is pretty dirty and unsafe (We're taking HTML wholesale, parsing it all and inserting it into the DOM), and relies pretty much entirely on jellyfin-web polluting `window` to have access to anything.

The idea is to remove the dependency on web. Ultimately, we want the web client to be "just another client". Currently, the server is heavily tied to it and that's an issue.

nielsvanvelzen (approved):

thornbill (approved):

thornbill:
I don't see any reason to block this right now. It's a marginal improvement to how plugins currently work. When we start planning for a new plugin configuration schema we can decide what features should (or should not) be included.

PR #2343, opened 2021-01-24
fix image alignment on plugin cards

dkanada:


src/components/cardbuilder/card.css
```diff
@@ -174,6 +174,7 @@ button::-moz-focus-inner {
 
 .cardScalable .cardImageContainer {
     height: 100%;
+    width: 100%;
```
thornbill, line 177: This doesn't seem to affect anything is it needed?

dkanada: It definitely fixed some issues with the layout when I was testing, are you on Firefox or another browser?

thornbill: Firefox

dkanada: I've confirmed this fixes an issue on my browser.

src/controllers/dashboard/plugins/available/index.js
```diff
@@ -102,10 +103,18 @@ function getPluginHtml(plugin, options, installedPlugins) {
...
     html += '<div class="cardPadder cardPadder-backdrop"></div>';
-    html += '<a class="cardContent cardImageContainer" is="emby-linkbutton" href="' + href + '"' + target + '>';
-    html += '<span class="cardImageIcon material-icons folder"></span>';
+    html += `<div class="cardContent ${cardBuilder.getDefaultBackgroundClass()}">`;
+    html += `<a class="cardImageContainer" is="emby-linkbutton" style="margin:0;padding:0" href="${href}" ${target}>`;
+
+    if (plugin.imageUrl) {
+        html += `<img src="${plugin.imageUrl}" style="width:100%" />`;
+    } else {
+        html += `<span class="cardImageIcon material-icons extension"></span>`;
```
dkanada, line 112:
```suggestion
        html += '<span class="cardImageIcon material-icons extension"></span>';
```

src/controllers/dashboard/plugins/installed/index.js
```diff
@@ -53,22 +54,30 @@ function getPluginCardHtml(plugin, pluginConfigurationPages) {
...
+    } else {
+        html += '<div class="cardImageContainer noConfigPluginCard noHoverEffect emby-button">';
+    }
+
     if (plugin.HasImage) {
-        html += `"><img src="/Plugins/${plugin.Id}/${plugin.Version}/Image" style="width:100%;height:auto"/>`;
+        html += `<img src="/Plugins/${plugin.Id}/${plugin.Version}/Image" style="width:100%" />`;
     } else {
-        html += ' material-icons folder">';
+        html += `<span class="cardImageIcon material-icons extension"></span>`;
```
dkanada, line 75:
```suggestion
        html += '<span class="cardImageIcon material-icons extension"></span>';
```

src/controllers/dashboard/plugins/installed/index.js
```diff
@@ -53,22 +54,30 @@ function getPluginCardHtml(plugin, pluginConfigurationPages) {
...
+
     if (plugin.HasImage) {
-        html += `"><img src="/Plugins/${plugin.Id}/${plugin.Version}/Image" style="width:100%;height:auto"/>`;
+        html += `<img src="/Plugins/${plugin.Id}/${plugin.Version}/Image" style="width:100%" />`;
     } else {
-        html += ' material-icons folder">';
+        html += '<span class="cardImageIcon material-icons extension"></span>';
     }
-    html += '</span> ';
+
```
dkanada, line 77:
```suggestion
```

thornbill (commented):
I am not a fan of the rainbow vomit introduced by adding the background color to all these cards in the dashboard.

dkanada:
We can always tweak the default color generation method, but the cards themselves need some kind of placeholder to differentiate the image section from the footers.

thornbill:
The addition of the background colors looks pretty terrible imo. The color is shown behind transparent user images. A large number of cards on a page becomes very overwhelming. The colors change if you leave a page and return or reload the page. And it seems they aren't applied to all cards (see the tuner cards).

Screenshot_2021-02-14 Jellyfin

Screenshot_2021-02-14 Jellyfin(1)

Screenshot_2021-02-14 Jellyfin(2)

Screenshot_2021-02-14 Jellyfin(3)

dkanada:
The intent was definitely to style them all the same, tuner cards were only unaffected because I didn't know they existed. I can add them as well if you'd like.

I personally love all the other images you perceived as negative. How would it sound to reduce the number of colors in the pallette or make them more similar? I will admit the brighest shade of blue is a bit too much for my taste. Regarding the transparent images, I can remove the background when an image is present to keep the old behavior.

If we're putting this much effort into the style changes I'd still like to backport them for 10.7.1 as a UX improvement, which I consider valid for backports.

Camc314:
Don't we already use these sort of range of colors for item cards, Personally, I don't mind what we decided to do, but it wouldnt it make sense to use the same sort of colors everywhere?

dkanada:
We did indeed, and it would indeed, which was the idea behind this pull request. However, I'm not against tweaking the pallette used at the same time. Like I said, the brighter shades are definitely a bit in your face.

thornbill:
We do use these colors for other cards also but only ever as placeholders when an image is missing. I don’t like the choice of colors for those cases either but it’s much more aggressive having a full page of these all over the dashboard.

dkanada:
We shouldn't assume that every user has metadata good enough to scrape images for a majority of their media, or even that every user wants to enable images globally. For example, I use images for posters and thumbnails but prefer to leave genres and people untouched. I don't even enable metadata providers for most of my libraries. We should just design the placeholders with the assumption that they should look good even when no images are present.

Like I said, feel free to make suggestions about what colors (or color) we should use, we can always change the default theme to make them less aggresive when it's a full page of placeholders. Also, I'd like to point out that even though it's a stretch to call the device images placeholders, both the active devices and plugin cards are *definitely* placeholders in the proper sense, so they should be styled as such. Devices and tuners are the only two I can think of that don't really have actual images to prioritize, but I styled them to match literally every other card that exists in the client.

src/components/cardbuilder/card.css
```diff
@@ -364,7 +365,7 @@ button::-moz-focus-inner {
 }
 
 .cardImageIcon {
-    font-size: 5em;
+    font-size: 5em !important;
```
thornbill, line 368: Could we increase the specificity of the selector here instead of using `!important`?

src/controllers/session/login/index.js
```diff
@@ -185,11 +163,12 @@ import baseAlert from '../../../components/alert';
                     tag: user.PrimaryImageTag,
                     type: 'Primary'
                 });
+
                 html += '<div class="cardImageContainer coveredImage" style="background-image:url(\'' + imgUrl + "');\"></div>";
             } else {
-                const background = getMetroColor(user.Id);
-                imgUrl = 'assets/img/avatar.png';
-                html += '<div class="cardImageContainer coveredImage" style="background-image:url(\'' + imgUrl + "');background-color:" + background + ';"></div>';
+                html += `<div class="cardImage flex align-items-center justify-content-center ${imgUrl ? '' : cardBuilder.getDefaultBackgroundClass()}">`;
```
thornbill, line 169: `imgUrl` would always be undefined here right?

thornbill (changes requested):

thornbill (approved):
:+1: lgtm

MrTimscampi (approved):

PR #2355, opened 2021-01-27
Fix warnings

thornbill:
**Changes**
Removes eslint rules that were downgraded to warnings and fixes the errors.

**Issues**
N/A

### DIFF
--- a/src/scripts/clientUtils.js
+++ b/src/scripts/clientUtils.js
@@ -19,12 +19,6 @@ export async function serverAddress() {
         return Promise.resolve(apiClient.serverAddress());
     }
 
-    const current = await ServerConnections.getAvailableServers().then(servers => {
-        if (servers.length !== 0) {
-            return Promise.resolve(servers[0].ManualAddress);
-        }
-    });
-
     // TODO this makes things faster but it also blocks the wizard in some scenarios
     // if (current) return Promise.resolve(current);
 
@@ -50,7 +44,7 @@ export async function serverAddress() {
                 url: url,
                 response: resp
             };
-        }).catch(error => {
+        }).catch(() => {
             return Promise.resolve();
         });
     });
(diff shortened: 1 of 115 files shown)


thornbill:
I need to finish cleaning this up before it is ready for review. SonarQube took issue with the removed function parameters and some additional linting issues have surfaced while cleaning that up.

thornbill:
Not entirely sure why the Lint and CodeQL checks did not run, but everything should be passing now.

src/scripts/clientUtils.js
```diff
@@ -19,12 +19,6 @@ export async function serverAddress() {
...
     }
 
-    const current = await ServerConnections.getAvailableServers().then(servers => {
-        if (servers.length !== 0) {
-            return Promise.resolve(servers[0].ManualAddress);
-        }
-    });
-
     // TODO this makes things faster but it also blocks the wizard in some scenarios
     // if (current) return Promise.resolve(current);
```
dmitrylyzo, line 23: seems unneeded

dmitrylyzo (approved):

src/scripts/clientUtils.js
thornbill: You're probably right, but I'd like to keep the scope of this PR as limited as possible. :smile:

src/scripts/clientUtils.js
```diff
@@ -19,12 +19,6 @@ export async function serverAddress() {
...
```
dmitrylyzo, line 24: I still think comment should be removed because the associated code is removed in this PR.
```suggestion
```
_Hope this removes 3 lines - GH shows only 2_

thornbill: Opened a separate PR since it is not _strictly_ related to fixing the warnings here. :wink:

https://github.com/jellyfin/jellyfin-web/pull/2559

dmitrylyzo (approved):

PR #2356, opened 2021-01-27
fix: notched devices area not covered

ferferga:
### ~~Untested yet~~

~~Should fix jellyfin/jellyfin-android#289~~

Read https://github.com/jellyfin/jellyfin-web/pull/2356#issuecomment-768377889, https://github.com/jellyfin/jellyfin-web/pull/2356#issuecomment-768400058 and https://github.com/jellyfin/jellyfin-web/pull/2356#issuecomment-772606784

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -2,7 +2,7 @@
 <html class="preload">
 <head>
     <meta charset="utf-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1, minimum-scale=1, maximum-scale=1, user-scalable=no">
+    <meta name="viewport" content="width=device-width, initial-scale=1, minimum-scale=1, maximum-scale=1, user-scalable=no, viewport-fit=cover">
     <link rel="manifest" href="manifest.json">
     <meta name="format-detection" content="telephone=no">
     <meta name="msapplication-tap-highlight" content="no">


MrTimscampi:
Wouldn't this lead to cut videos? (As in the notch area will be used like normal screen real estate, leading to the video appearing behind the notch)

ferferga:
@MrTimscampi This will work only when aspect ratio in the playback settings is set to "Cover". By default it's contain, which crops borders at the right and at the left. That *should* be unaffected.

nielsvanvelzen:
This value will be ignored by the custom webview implementation in the Android app.

ferferga:
I will test if it makes a difference for mobile browsers and, if it doesn't, I'll close this.

Maxr1998:
See my comment in jellyfin/jellyfin-android#289, the webview padding is determined by the app, the wrong color was a bug in the app and is already fixed.

ferferga:
@Maxr1998 I saw it already :)

The thing is that, if I'm using Edge or Chrome in my phone (not the app) the notch area is still not 100% covered and this might fix that.

That's what I want to verify 👍🏻

ferferga:
@nielsvanvelzen @Maxr1998 @MrTimscampi This does indeed allows the video to take notch area in the web client in an Huawei P30 with Firefox, so I think this is good to have 👍.

nielsvanvelzen (approved):

dkanada (approved):

PR #2357, opened 2021-01-28
Fix epub player height

thornbill:
**Changes**
Fixes an issue where the epub player height would be wrong due to the dialog open animation

**Issues**
N/A


ferferga (approved):
LGTM

dkanada (approved):

PR #2378, opened 2021-02-06
update style for active sessions

dkanada:
This basically just tweaks the device cards on the dashboard to match our style more closely.

- Replaces the border line with a proper placeholder image to serparate the footer
- Adds a background behind the progress bar to make the vertical spacing uniform across all cards
- Moves the transcoding description to the info button to make all cards have the same height


src/controllers/dashboard/dashboard.js
```diff
@@ -251,12 +254,13 @@ import confirm from '../../components/confirm/confirm';
...
                 const className = 'scalableCard card activeSession backdropCard backdropCard-scalable';
+                const imgUrl = DashboardPage.getNowPlayingImageUrl(nowPlayingItem);
+
                 html += '<div class="' + className + '" id="' + rowId + '">';
                 html += '<div class="cardBox visualCardBox">';
                 html += '<div class="cardScalable visualCardBox-cardScalable">';
                 html += '<div class="cardPadder cardPadder-backdrop"></div>';
-                html += '<div class="cardContent">';
-                const imgUrl = DashboardPage.getNowPlayingImageUrl(nowPlayingItem);
+                html += `<div class="cardContent ${imgUrl ? '' : cardBuilder.getDefaultBackgroundClass()}">`;
```
dkanada, line 263:
```suggestion
                html += `<div class="cardContent ${cardBuilder.getDefaultBackgroundClass()}">`;
```

MrTimscampi (approved):
Not sure if we should backport this at this point (It's not big, but it's not really a bug fix either), but the code itself looks okay.

thornbill:
Can you add screenshots and/or a description of what should have changed here?

thornbill:
Just adding screenshots of the before/after here for reference:

**Before:**
Screenshot_2021-02-16 Jellyfin

**After:**
Screenshot_2021-02-16 Jellyfin(1)

src/assets/css/dashboard.css
```diff
@@ -409,9 +401,14 @@ div[data-role=controlgroup] a.ui-btn-active {
 }
 
 .transcodingProgress > div {
+    z-index: 900;
```
thornbill, line 404: This seems arbitrarily high.

src/controllers/dashboard/dashboard.js
```diff
@@ -287,29 +291,19 @@ import confirm from '../../components/confirm/confirm';
...
-                    html += indicators.getProgressHtml(percent, {
-                        containerClass: 'playbackProgress'
-                    });
-                } else {
-                    // need to leave the element in just in case the device starts playback
-                    html += indicators.getProgressHtml(0, {
-                        containerClass: 'playbackProgress hide'
-                    });
-                }
+                let percent = 100 * session?.PlayState?.PositionTicks / nowPlayingItem?.RunTimeTicks;
```
thornbill, line 294: May need to handle this better since 100 * null / null = NaN

src/controllers/dashboard/dashboard.js
```diff
@@ -287,29 +291,19 @@ import confirm from '../../components/confirm/confirm';
...
+                html += indicators.getProgressHtml(percent ?? 0, {
```
thornbill, line 295: Looks like `??` is new for ES2020. Does our babel config transpile this correctly?

thornbill (changes requested):
Overall I think these changes look great!

I'm not sure about the playback info positioning though. It seems a bit off since all the other items in that row are actions to send to the remote session. The alert dialog is a bit weird ux wise since it only displays 1-2 words. I'm not sure how that could be handled better though.

There also seems to be a bug where the playback info isn't always being displayed? I saw it once, but on later sessions it never showed up even after trying to refresh, stop, resume playback etc. There are no console errors or anything that would indicate what the issue is.

src/controllers/dashboard/dashboard.js
dkanada: This is why I added the next line, is that sufficient?

src/controllers/dashboard/dashboard.js
dkanada: I just checked and it looks like this is properly transpiled.

src/assets/css/dashboard.css
dkanada: I can set it to an arbitrarily lower number if you want.

src/controllers/dashboard/dashboard.js
thornbill: In my testing the `percent ?? 0` does not get changed to 0 if percent is `NaN`.

src/assets/css/dashboard.css
thornbill: The lower the better. By setting the z-index to a really high value then things like modals could be messed up if their z-index doesn't exceed the arbitrarily high value.

thornbill:
I did some additional testing and it seems like for me the info button is not available if playback starts after the page is loaded. Also the "Direct Play" modal feels lacking. Maybe adding a title could help?

Screenshot_2021-02-24 Jellyfin

Screenshot_2021-02-24 Jellyfin(1)

thornbill (approved):
:purple_heart: Looks great now!

PR #2379, opened 2021-02-06
fix some unhandled promises and refactor some code

dkanada:


MrTimscampi (approved):
LGTM.

src/components/dialogHelper/dialogHelper.js
```diff
@@ -106,15 +106,6 @@ import '../../assets/css/scrollstyles.css';
                     tryRemoveElement(dlg);
                 }
             }
-
-            //resolve();
-            // if we just called history.back(), then use a timeout to allow the history events to fire first
-            setTimeout(() => {
```
thornbill, line 112: Why was this removed?

src/components/alert.js
```diff
@@ -33,14 +33,7 @@ import globalize from '../scripts/globalize';
...
             options.buttons = items;
-
-            return dialog.show(options).then(function (result) {
-                if (result === 'ok') {
-                    return Promise.resolve();
-                }
-
-                return Promise.reject();
-            });
+            return dialog.show(options);
```
thornbill, line 36: Doing a quick search, it looks like there are other places that would be handling this promise rejection. How did you confirm this is not needed?

thornbill (changes requested):

src/components/dialogHelper/dialogHelper.js
dkanada: I was seeing tons of unhandled exceptions for dialogs and traced it back to this weird section of code, and removing it had no ill effects. History is already broken on dialogs in stable and I didn't want to look into it too much, so I figured it was fine.

src/components/alert.js
dkanada: I actually found the opposite, can you point to an example? It looked to me like there were two or three methods with the same name, and this one wasn't using the promise rejection anywhere. It was actually another source of unhandled exceptions whenever the dialog was closed without explicitly clicking the OK button. The only use I saw was the one to supress an unhandled exception.

thornbill (approved):

PR #2403, opened 2021-02-11
 Add TailwindCSS and start cleaning up styles and HTML

MrTimscampi:
**Changes**

This is more a proof of concept for discussion than a real PR, but it should be able to be merged directly if the consensus goes towards making this a thing.

This adds TailwindCSS to the PostCSS plugins.

As a start, this PR also does the following:

* Remove flexstyles.scss
  All of the classes have been renamed to their Tailwind equivalent.
* Migrate most of `dashboard.html` to use Tailwind
  This can still use more cleanup, but we first need to process other pages and files, as well as use tailwind's breakpoints (and unify our own breakpoints) to clean up most of the rest.
* Replace our own `hide` with Tailwind's `hidden`. This doesn't touch a bunch of things where `display: none` is set through the style attribute, since that should likely be done in a separate cleanup PR.

**Issues**


tailwind.config.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+    purge: [],
```
Ben-Wallner, line 2: I don't if it was intentional but you forgot to specify the purge path

src/controllers/dashboard/dashboard.html
```diff
@@ -112,10 +112,9 @@ <h3>${HeaderPaths}</h3>
             </div>
         </div>
 
-        <div class="dashboardFooter">
-            <div style="height:1px;" class="ui-bar-inherit"></div>
```
thornbill, line 116: Is this ui bar element not needed?

src/controllers/dashboard/dashboard.html
```diff
@@ -1,11 +1,11 @@
 <div id="dashboardPage" data-role="page" class="page type-interior dashboardHomePage fullWidthContent">
     <div class="content-primary">
-        <div class="dashboardSections" style="padding-top:.5em;">
-            <div class="dashboardColumn dashboardColumn-2-60 dashboardColumn-3-46">
```
thornbill, line 4: I seem to remember these `dashboardColumn-3-x` classes were unused, but just want to confirm. :smile:

src/controllers/dashboard/dashboard.html
```diff
@@ -19,91 +19,91 @@ <h3>${TabServer}</h3>
...
-            <div class="dashboardColumn dashboardColumn-2-40 dashboardColumn-3-27">
-                <div class="dashboardSection">
-                    <a is="emby-linkbutton" href="#!/serveractivity.html?useractivity=true" class="button-flat sectionTitleTextButton">
-                        <h3>${HeaderActivity}</h3>
-                        <span class="material-icons chevron_right"></span>
+            <div class="flex flex-col flex-shrink-0 flex-grow dashboardColumn-2-40">
+                <div class="flex-shrink-0 mb-8 dashboardSection">
+                    <a is="emby-linkbutton" href="#!/serveractivity.html?useractivity=true" class="m-0 button-flat sectionTitleTextButton">
+                        <h3 class="my-2">${HeaderActivity}</h3>
+                        <span class="my-2 pt-0  material-icons chevron_right"></span>
```
thornbill, line 51: Would it be better to use `@apply` and define these classes in the sass file to avoid adding the same list of classes to each and every section here? This approach seems error prone imho.

https://tailwindcss.com/docs/using-with-preprocessors

thornbill (changes requested):

PR #2404, opened 2021-02-12
Add TypeScript support and migrate itemContextMenu

MrTimscampi:
**Changes**

**Hide the spaces when reviewing. I removed the disable for the indentation rule.**

This adds TypeScript support and the types from the Axios API client, to get a head start on typing things from the API.

It also migrates `itemContextMenu` to TypeScript.

It's not set in strict mode, otherwise I'd pull my hair out when trying to migrate a single file. But the final goal is to have everything in strict.

image


thornbill:
imho this should be blocked until sources for the axios client are available (https://github.com/jellyfin/jellyfin-meta/issues/10).

MrTimscampi:
Closing because #2722 does the same thing, but better.

PR #2428, opened 2021-02-18
Fix: prevent some backdrops from showing on homepage

acvigue:
**Changes**
prevents backdrops from items with an OfficialRating > TV-MA from showing when no ParentId given

**Issues**
None

### DIFF
--- a/src/scripts/autoBackdrops.js
+++ b/src/scripts/autoBackdrops.js
@@ -27,7 +27,7 @@ function getBackdropItemIds(apiClient, userId, types, parentId) {
         ImageTypes: 'Backdrop',
         ParentId: parentId,
         EnableTotalRecordCount: false,
-        MaxOfficialRating: parentId ? '' : 'TV-MA'
+        MaxOfficialRating: parentId ? '' : 'PG-13'
     };
     return apiClient.getItems(apiClient.getCurrentUserId(), options).then(function (result) {
         const images = result.Items.map(function (i) {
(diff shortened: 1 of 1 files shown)


dkanada:
Does this affect backdrops when nothing in the library has ratings?

acvigue:
@dkanada No, it does not. If an item has no rating, it will still be included.

MrTimscampi:
I'd even put a lower rating than TV-MA.

Also, some questions:

* How does that work for non-US ratings? How does it handle stuff like France's "12+" or other countries' rating systems?
* How does that work for non-TV US ratings? Afaik, TV-MA is specific to the US and to TV shows. How does this handle rated-R movies?

acvigue:
I would assume it follows the order shown in the chart when editing metadata within Jellyfin.
image

TV-MA is technically _above_ the rating of rated-R movies, so those backdrops are still shown.

thornbill:
I wonder if this should use something lower like "TV-14"?

MrTimscampi:
> I wonder if this should use something lower like "TV-14"?

I would probably go PG-13 for safety? TV-14 could still have blood and such, which some parents wouldn't like.

PG-13, on the other hand, is usually more about language, so that would be safer.

acvigue:
Thanks for the input, i switched it to be PG-13

thornbill (approved):
:+1: Works as expected in my testing.

PR #2442, opened 2021-02-23
minor improvements to plugin pages

dkanada:
- always selects the newest plugin during installation
- fix margin and padding for new plugins
- reduce the number of lines for installed plugin cards
- fix missing icons for enabling or disabling plugins


src/controllers/dashboard/plugins/add/index.js
```diff
@@ -33,7 +33,7 @@ function populateVersions(packageInfo, page, installedPlugin) {
         $('#pCurrentVersion', page).hide().html('');
     }
 
-    const packageVersion = packageInfo.versions[0];
+    const packageVersion = packageInfo.versions.sort().pop();
```
thornbill, line 36: This seems to be selecting the _oldest_ version in my testing?

thornbill (changes requested):

thornbill (approved):
This looks fine now. I haven’t tested it to verify it works as expected though.

src/controllers/dashboard/plugins/add/index.js
```diff
@@ -22,6 +22,10 @@ function populateHistory(packageInfo, page) {
 function populateVersions(packageInfo, page, installedPlugin) {
     let html = '';
 
+    packageInfo.versions.sort((a, b) => {
+        return b.timestamp < a.timestamp ? 1 : -1;
+    });
+
```
dmitrylyzo, line 28: I see versions like this:
```
|Reports |v|
9.0.0.0
10.0.0.0
11.0.0.0 <-- selected
```
Everything is fine. But wouldn't it be better if versions go in reverse order?
```
|Reports |v|
11.0.0.0 <-- selected
10.0.0.0
9.0.0.0
```

src/controllers/dashboard/plugins/installed/index.js
```diff
@@ -87,12 +87,10 @@ function getPluginCardHtml(plugin, pluginConfigurationPages) {
...
 
-    html += "<div class='cardText'>";
-    html += configPage && configPage.DisplayName ? configPage.DisplayName : plugin.Name;
-    html += `<br/>${globalize.translate('LabelStatus')} ${plugin.Status}</div>`;
-    html += "<div class='cardText cardText-secondary'>";
-    html += plugin.Version;
+    html += '<div class="cardText">';
+    html += `${plugin.Name}<span class='cardText cardText-secondary'>${plugin.Version}</span>`;
     html += '</div>';
+    html += `<div class="cardText">${globalize.translate('LabelStatus')} ${plugin.Status}</div>`;
```
dmitrylyzo, line 93: Is `${plugin.Status}` supposed to be translated by server?

dkanada: Not at the moment.

src/controllers/dashboard/plugins/add/index.js
dkanada: I suppose that might be useful, do you have an idea about the code changes? I can't test at the moment so it might not make 10.7.0 if we wait for me to find a desktop.

src/controllers/dashboard/plugins/add/index.js
```diff
@@ -22,6 +22,10 @@ function populateHistory(packageInfo, page) {
 function populateVersions(packageInfo, page, installedPlugin) {
     let html = '';
 
+    packageInfo.versions.sort((a, b) => {
+        return b.timestamp < a.timestamp ? 1 : -1;
```
dmitrylyzo, line 26: just reverse here + `[0]` instead of `.pop()` (see below)
```suggestion
        return b.timestamp < a.timestamp ? -1 : 1;
```

src/controllers/dashboard/plugins/add/index.js
```diff
@@ -33,7 +37,7 @@ function populateVersions(packageInfo, page, installedPlugin) {
         $('#pCurrentVersion', page).hide().html('');
     }
 
-    const packageVersion = packageInfo.versions[0];
+    const packageVersion = packageInfo.versions.pop();
```
dmitrylyzo, line 40:
```suggestion
    const packageVersion = packageInfo.versions[0];
```

src/controllers/dashboard/plugins/add/index.js
dmitrylyzo: > I suppose that might be useful, do you have an idea about the code changes?

See below.

We also need to check `targetAbi`. Someday :wink:

dmitrylyzo (approved):

PR #2464, opened 2021-02-28
Tune some styles on now playing page

dmitrylyzo:
`x-large` and `xx-large` are absolute sizes. We have some font size adjustments for the root element that don't apply to buttons with mentioned font sizes.

**Desktop**
_We use 93% for `font-size` of root element. Now play button is 2x from 93%. Spacing changed._
Before:
old-desktop
After:
new-desktop

**iPhone5 (small screen)**
_On mobile, we use 90% for `font-size` of root element. But for iOS we use 82%. Now playback buttons are 1.5x from 90% (82% on iOS)._
Before:
old-iphone5
After:
new-iphone5

**TV**
_We use 27px for `font-size` of root element (for FHD). Now play button is 2x from 27px. Spacing changed._
Before:
old-tv
After:
new-tv

**Changes**
- Make font size relative
- Make spacing between buttons more even

**Issues**
:eyes:

### DIFF
--- a/src/components/remotecontrol/remotecontrol.scss
+++ b/src/components/remotecontrol/remotecontrol.scss
@@ -120,7 +120,6 @@
 .btnPlayPause {
     font-size: 2em;
     padding: 0;
-    margin: 0;
 }
 
 .nowPlayingInfoControls {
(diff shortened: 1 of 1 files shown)


thornbill (approved):

dmitrylyzo:
I think the play button got small for the iPhone X:


`2em` still looks fine on iPhone 5.

dkanada (approved):

PR #2467, opened 2021-02-28
Add padding to prevent music controls from clipping

albeeeert:
**Changes**
Add bottom padding for artist view and e.g. photos view to prevent music controls from clipping with information. Removes a needless inline style from list.html.

**Issues**
Fixes issue #2314.

### DIFF
--- a/src/assets/css/site.scss
+++ b/src/assets/css/site.scss
@@ -81,7 +81,8 @@ div[data-role="page"] {
 .content-primary,
 .padded-bottom-page,
 .page,
-.pageWithAbsoluteTabs .pageTabContent {
+.pageWithAbsoluteTabs .pageTabContent,
+.itemsContainer {
     /* provides room for the music controls */
     padding-bottom: 5em !important;
 }
--- a/src/assets/css/site.scss
+++ b/src/assets/css/site.scss
@@ -81,8 +81,7 @@ div[data-role="page"] {
 .content-primary,
 .padded-bottom-page,
 .page,
-.pageWithAbsoluteTabs .pageTabContent,
-.itemsContainer {
+.pageWithAbsoluteTabs .pageTabContent {
     /* provides room for the music controls */
     padding-bottom: 5em !important;
 }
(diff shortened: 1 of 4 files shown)


src/assets/css/site.scss
```diff
@@ -81,7 +81,8 @@ div[data-role="page"] {
 .content-primary,
 .padded-bottom-page,
 .page,
-.pageWithAbsoluteTabs .pageTabContent {
+.pageWithAbsoluteTabs .pageTabContent,
+.itemsContainer {
```
dmitrylyzo, line 85: `itemContainer` is too generic - all containers obtain bottom padding. Instead, you can add `padded-bottom-page` to `detailPageWrapperContainer` on music item details page.

src/controllers/list.html
```diff
@@ -48,7 +48,7 @@
                 <span class="material-icons more_vert"></span>
             </button>
         </div>
-        <div is="emby-itemscontainer" class="vertical-wrap itemsContainer centered" style="padding: 0 0 10vh 0;">
+        <div is="emby-itemscontainer" class="vertical-wrap itemsContainer centered">
```
dmitrylyzo, line 51: Agree with this. Also I would add mentioned `padded-bottom-page` to `pageContainerTopPadding` element as above for the same purpose.

src/assets/css/librarybrowser.css
```diff
@@ -1052,7 +1052,8 @@ div.itemDetailGalleryLink.defaultCardBackground {
 }
 
 .itemsContainer {
-    margin: 0 auto;
+    margin-left: auto;
+    margin-right: auto;
```
dmitrylyzo, line 1056: With `padded-bottom-page` class, this may be left as it was.

dkanada:
@albeeeert did you have any comments about the suggestions?

src/assets/css/site.scss
albeeeert: Yeah, I agree. I amended as you said, and it seems to work.

albeeeert:
Made changes as @dmitrylyzo stated in the latest commit. Seems to work and I agree that the itemsContainer was way too generic.

dmitrylyzo (approved):

dkanada (approved):

PR #2470, opened 2021-03-01
browser.js: Avoid misdetecting Chrome OS as OS X

### LINKED ISSUE #2477: Chrome OS is mistaken for a "Chrome iPad", leading to video playback issues.
Many videos in my collection fail to play on Intel based Chrome OS devices, only showing an error that the format is unsupported by the client (I think "This client isn't compatible with the media and the server isn't sending a compatible media format", but my install already contains the fix so I can't reproduce)

**Steps To Reproduce**
1. On an Intel Chrome OS pick a video with some h.264 "high" profile
2. Try to play it
3. Observe error
4. Go to devices list in the dashboard and see that the device is considered a "Chrome iPad"

**Expected Behavior**
The video plays and the device is not mistaken for an iPad.

**System:**
 - Platform: Chrome OS (Intel)
 - Browser: Chrome
 - Jellyfin Version: unstable / 10.8.0

pgeorgi:
Matching for lower-case "os x" not only finds "Mac OS X" but also
"CrOS x86_64". Mismatching Chrome OS as Apple device leads to
playback issues due to differences in supported client formats.

I checked User Agent lists on the Internet (e.g. https://developers.whatismybrowser.com/useragents/explore/) and all Apple browsers seem to report "Mac OS X $version" (mac OS) or something similar to "iPhone OS $version like Mac OS X" (iOS device family), so that extended regex should still work.

Fixes #2477

### DIFF
--- a/src/scripts/browser.js
+++ b/src/scripts/browser.js
@@ -210,7 +210,7 @@ if (!browser.chrome && !browser.edgeChromium && !browser.edge && !browser.opera 
     browser.safari = true;
 }
 
-browser.osx = userAgent.toLowerCase().indexOf('os x') !== -1;
+browser.osx = userAgent.toLowerCase().indexOf('mac os x') !== -1;
 
 // This is a workaround to detect iPads on iOS 13+ that report as desktop Safari
 // This may break in the future if Apple releases a touchscreen Mac


thornbill (approved):
👍 I checked a couple open source user agent libraries and this seems to match how they check for macOS.

dkanada (approved):

PR #2476, opened 2021-03-01
Switch to npm

ferferga:
Follow up of https://github.com/jellyfin/jellyfin-vue/pull/815

*This needs further discussion across all the client devs, as this has been discussed more thoroughly between the Vue contributors, but not as much with the rest of the client developers and. Most client developers showed interest in switching to npm, however, no formal discussion has taken place and, ideally, we should use the same package manager everywhere*

Major reasons for switching:

* All the yarn's advantages over npm are non-existent today
* Yarn introduced some breaking changes in 2.x that caused controversies and avoid us to upgrade directly, which means we are stuck in a version in maintenance mode. So far, all the npm major updates could be updated like a charm, and it's likely they will commit to it. The same can't be said about yarn.
* Npm comes with the node installation, so it's frictionless for the user
* Npm is more mainstream and all the package maintainers add instructions with how to use it to install their packages. The same doesn't apply to yarn. Although adapting the commands to yarn is trivial, for new contributors it might cause some friction.

## Extra

* Added dependabot for dependency updates in GitHub Actions


dkanada (commented):
I have no opinion on this but we can't very well switch when CI is failing.

ferferga:
@dkanada Yeah, of course.

The purpose of this was mainly to formalise the already internal discussion across the Web and client team for this change, so I just made a quick PR that, of course, will probably need some changes.

Something like this needs the unanimous approval of everyone involved. Once that consensus exist across all the projects (as, if we're doing this, it's a full project move), I can rebase this and work deeply on it :).

I wouldn't also merge this until 10.7 is released, as it might cause conflicts with backports. I thought I labeled this as ``blocked`` but I didn't, so I'm doing that now :)

thornbill:
do it

README.md
```diff
@@ -45,7 +45,7 @@ Jellyfin Web is the frontend used for most of the clients available for end user
 ### Dependencies
 
 - [Node.js](https://nodejs.org/en/download)
-- [Yarn 1.22.5](https://classic.yarnpkg.com/en/docs/install)
+- npm (included in Node.js)
```
nielsvanvelzen, line 48: Will any NPM version or is there a minimum requires version? (Like 7>=)

ThibaultNocchi: We've had problems with "old" versions of npm on Vue, right @ferferga ?

ferferga: @nielsvanvielzen Good thing you're asking this and thank you very much for @ThibaultNocchi four pointing it out.

The main feature we use yarn for is for deterministic packages. That has been the case for ``npm`` for quite a few years. That's the *minimum* we need to keep our workflow as equal to yarn as possible.

NodeJS 14 and 12 includes ``npm 6``, while package-lock.json was introduced in ``npm 5``. While npm 7 is supposedly more performant than the 6 and 5 counterparts, they work, so it's frictionless for everybody, while still keeping the freedom for those that want the greatest and latest and upgrade to npm 7.

Now, this leads us to the Vue problem: In my system I don't have the LTS version included with Node, as I updated it for other packages (I don't remember which version I had, but I was already in the 7.x branch)

I discovered that running ``npm ci`` (equivalent to ``--frozen-lockfile`` argument) command missed the dependencies of some dependencies. That made CI fail. Upgrading to 7.6.0 fixed the problem, so I took the easy way of amending my PR by updating always to latest npm version, and that fixed all the problems. 

So, **TLDR, npm install has always been reliable, the problem has been with npm ci**. So "users" or us were never affected *directly*

I need to investigate more which versions are exactly affected with this bug if we want to stick with the bundled npm versions during our CIs, instead of following my approach of always using the latest available.

ferferga: I confirm that 6.14 (the included version in Node 14) works perfectly both in ``install`` and ``ci``, so no need for the ugly hack I used in Vue.

ferferga:
10.7 is released now and we're starting 10.8 cycle, so unlabeling this as blocked

thornbill (approved):
lgtm could probably use a review for @joshuaboniface for the build stuff though

joshuaboniface (approved):
Looks fine by me as long as everything builds as expected.

joshuaboniface (approved):

PR #2509, opened 2021-03-14
Disable first episodes in Next Up home section

### LINKED ISSUE #2430: Next up inconsistent behavor without partially watched series
**Describe The Bug**
Inconsistent behavior when there is no Next Up content based on started series available.

**Steps To Reproduce**
1. Create a new user, so it doesn't have any watched items
2. Now Next Up displays all existing shows
3. Mark any episode as watched
4. Next Up now only shows that series
5. Mark entire series as watched
6. Next Up is gone

**Expected Behavior**
It seems inconsistent to me that the behavior is different in the following situations:
- No episode at all is watched
- All episodes in all started series are watched

It seem to me that either Next Up should disappear in both cases or show (seemingly) random shows in both cases.

**Logs**
–

**Screenshots**
–

 - Platform: Windows
 - Browser: Firefox
 - Jellyfin Version: 10.6.4

**Additional Context**
–

MrTimscampi:
**Changes**
This makes use of the new API parameter to remove the first episodes which show up by default when the user hasn't watched anything yet.

Pretty much all the streaming services behave this way, and so do other media servers.

**Issues**

Fix #2430

### DIFF
--- a/src/components/homesections/homesections.js
+++ b/src/components/homesections/homesections.js
@@ -665,7 +665,8 @@ import ServerConnections from '../ServerConnections';
                 UserId: apiClient.getCurrentUserId(),
                 ImageTypeLimit: 1,
                 EnableImageTypes: 'Primary,Backdrop,Banner,Thumb',
-                EnableTotalRecordCount: false
+                EnableTotalRecordCount: false,
+                DisableFirstEpisode: true
             });
         };
     }


MrTimscampi:
Marked it as a stable backport since it's a one-liner and I tested it against a 10.7.0 server. Feel free to remove if it's not something we'd want to backport.

cvium:
If it's not configurable, I don't see why it should be backported tbh

MrTimscampi:
> If it's not configurable, I don't see why it should be backported tbh

Because it's technically a bug. It was fixed during 10.7 but had to be reversed (and this parameter added) because Android TV relied on the weird implementation of that endpoint for some things.

We've had a few issues opened for it before, so imo it would warrant a fix.

Essentially, we're the only weird ones that do "Next up" like this. It's not a suggestions section, it's supposed to list things you're currently watching and may want another episode of. The behavior is also likely inconsistent (as mentioned in #2430) and I don't think holding a 1-line fix for 6+ months is the right call.

thornbill:
Related server PRs for context:
https://github.com/jellyfin/jellyfin/pull/4595
https://github.com/jellyfin/jellyfin/pull/5027

thornbill:
This does constitute a behavioral change which strictly speaking should have been in before the final 10.7 release especially given the extended release candidate cycle this time around. I'm also not convinced that this is really a bug. It was probably intended to provide additional content on the home screen for the initial user experience.

Code wise it is a very minor change and there is at least the one issue you linked for it, so maybe it is fine to backport without causing backlash. :man_shrugging:

dkanada (approved):

thornbill (changes requested):
Any relevant issues should be linked in the description so they get tracked properly.

joshuaboniface:
I'm not opposed to backporting it, but I'd be concerned if people *liked* this setup. But that said, I think it's obscure enough, and if there's been several issues mentioning it like a bug, I'd say let's just go for it.

thornbill (approved):

TimGels:
I had a discussion with some people over at Discord about this and it seems to be happening again with 10.8.8. I created a new user to try and reproduce this (this user hasn't watched anything yet). Result, it seems to be happening.
afbeelding

PR #2511, opened 2021-03-14
Default to MP3 for transcoding

knackebrot:
ffmpeg's AAC encoder has some audible artifacts (at least to my ears) and is considered worse than LAME on both [ffmpeg wiki](https://trac.ffmpeg.org/wiki/Encode/HighQualityAudio#AudioencodersFFmpegcanuse) and HydrogenAudio.
[FDK-AAC is just too slow](https://github.com/jellyfin/jellyfin-ffmpeg/pull/61#issuecomment-774654736) to the point where it slows down the encoding process when using HW acceleration, maybe due to its fixed-point nature.
Therefore I propose defaulting to MP3 at least until there's a better AAC encoder available.

**Changes**
Invert the order of preferred audio codecs


EraYaN:
The mp3 encoder needs a higher bandwidth to reach the same quality so if anything this should be configurable. And besides the differences are only really noticeable on lower bit rates than that jellyfin should use for music all those tests at HydrogenAudio were done with 96kbps and 128kbps. I don't know I'm not really a fan.

And FDK-AAC is unshippable so that one is essentially out of the question if we don't make it configurable by the user. The better alternative is probably just using FAAC 1.30 but I'm not sure if we (and ffmpeg) are shipping that yet. Other options is exhale but client devices will need to support that.

In the end for anything above 250 it really doesn't matter all that much, the power of the noise introduces is very low at that point and the choice of codec is almost moot.

knackebrot:
> The mp3 encoder needs a higher bandwidth to reach the same quality so if anything this should be configurable. 
This is simply not true. LC-AAC codec has the potential for better compression, yes, but the encoder needs to be good as well, which ffmpeg 'aac' is not, I hear artifacts at any bitrate. Googling it gave me this twitter: https://twitter.com/marcan42/status/1255196511001956352
FDK-AAC is shippable under certain conditions, as Fedora has proven, but I'm not a fan of that, as stated earlier.  FAAC is not a good encoder either. Jellyfin defaults to 192kbps for stereo audio I think, which is high enough for MP3. Yeah, we can make it more configurable, but I consider the current default to be suboptimal.

EraYaN:
You can distribute fdk-aac separately yes (but still controversial), but not as part of ffmpeg due to it's license incompatibility with the GPL.

FAAC 1.30 is pretty different from 1.29 as shown by the [HydrogenAudio thread](https://hydrogenaud.io/index.php?topic=119861.0) linked in that twitter thread. It should be started that the differences (as always) are vastly overstated in that twitter thread, and I always wonder if those guys have ever actually heard a bad audio codec or not. And they are always combative in on the issue tracker without ever providing patches or fixes so, you can safely take the guys' words with a grain of salt. The hydrogen audio thread though is an interesting view of the lower bitrate performance of the different codecs.

knackebrot:
[FSF's issue with FDK was that it does not grant patent licenses](https://listening-test.coresv.net/img2/mos_by_corpus.png). [Red Hat & Fedora have repeatedly stated](https://lists.fedoraproject.org/archives/list/legal@lists.fedoraproject.org/thread/OVW25JRWOKOLVMW3XGUX7E4OXFUR2RCG/) that this is a non-issue for their fdk-aac-free fork that only enables MPEG-2 AAC, because its patents have already expired. See my pull request for jellyfin-ffmpeg.

Based on that HA thread FAAC is similar or a bit better than ffmpeg AAC at those bitrates, yeah, but I'd bet it's still worse than LAME. Plus, ffmpeg removed support for it years ago. Here's another nice graph from HA:
image
Again, my argument against ffmpeg aac is not its compression ratio, it's that, to my ears, it does not sound "right" at any bitrate.

dmitrylyzo (commented):
Wouldn't it be easier to fill in the supported codecs first and then reorder them?
Codec order in user preferences :thinking:

src/scripts/browserDeviceProfile.js
```diff
@@ -328,7 +362,7 @@ import browser from './browser';
                                     || videoTestElement.canPlayType('video/mp4; codecs="avc1.640029, mp3"').replace(/no/, '');
 
         // Not sure how to test for this
-        const supportsMp2VideoAudio = browser.edgeUwp || browser.tizen || browser.web0s;
+        const supportsMp2VideoAudio = browser.edgeUwp || browser.tizen || browser.web0s || browser.chrome || browser.edgeChromium;
```
MrTimscampi, line 365: It'd be good to document why you're adding this and how to test it.

src/scripts/browserDeviceProfile.js
```diff
@@ -294,12 +294,46 @@ import browser from './browser';
...
+
+            return audioCtx.destination.maxChannelCount;
+        }
+
+        return -1;
+    }
+
+    function getPhysicalAudioChannels(options) {
+        const numChannels = getSpeakerCount();
+        const isSurroundSoundSupportedBrowser = browser.safari || browser.chrome || browser.edgeChromium || browser.firefox || browser.tv || browser.ps4 || browser.xboxOne || false;
```
MrTimscampi, line 311: I know this is copy-pasted from below, but I believe this covers pretty much all the browsers we target, so not sure how useful this is in practice.

src/scripts/browserDeviceProfile.js
```diff
@@ -346,26 +380,19 @@ import browser from './browser';
...
         if (supportsMp3VideoAudio) {
             videoAudioCodecs.push('mp3');
 
-            // PS4 fails to load HLS with mp3 audio
-            if (!browser.ps4) {
+            // PS4 fails to load HLS with MP3 audio
+            // Only prefer MP3 when the client wants stereo or mono to prevent downmixing
+            if (!browser.ps4 && physicalAudioChannels <= 2) {
                 hlsInTsVideoAudioCodecs.push('mp3');
+                hlsInFmp4VideoAudioCodecs.push('mp3');
```
MrTimscampi, line 392: Do we know if the PS4 supports fMP4 and if MP3 works in fMP4 on a PS4 or has the same issue as HLS?

src/scripts/browserDeviceProfile.js
```diff
@@ -346,26 +380,19 @@ import browser from './browser';
...
             }
-
-            hlsInFmp4VideoAudioCodecs.push('mp3');
         }
 
-        // For AC3/EAC3 remuxing.
-        // Do not use AC3 for audio transcoding unless AAC and MP3 are not supported.
```
MrTimscampi, line 368: It'd be useful to know why you're removing comments.

MrTimscampi (commented):
Imo, there should be comparison samples attached to this PR is you want to push this.

Just basing the entire reasoning on HydrogenAudio and the FFMPEG Wiki is a bit thin, since the former is well known for exaggerating things and the latter just has a basic list without any real comments on it.

*How much better* is Lame compared to the AAC encoder? What is the bandwidth difference?

You say:

> Jellyfin defaults to 192kbps for stereo audio I think, which is high enough for MP3.

But then complain about audio quality issues? 192kbps is "alright", but it's hardly high quality.

Though, again, if we're pushing the "audio quality" argument for this, there needs to be samples provided (Ideally a "master" file in FLAC, and Jellyfin-equivalent MP3 and AAC files)

src/scripts/browserDeviceProfile.js
knackebrot: My solution determines the number of speakers the browser has available, so that users with stereo speakers don't get a 6 channel track by default when transcoding sources with surround sound. Previous logic was: if userSettings.allowedAudioChannels() or options.audioChannels are not set, use 6 channels unless isSurroundSoundSupportedBrowser == false

src/scripts/browserDeviceProfile.js
knackebrot: Don't have a PS4, the documentation (https://www.sie.com/content/dam/corporate/jp/guideline/PS4_Web_Content-Guidelines_e.pdf) states it only supports this specification: https://tools.ietf.org/html/draft-pantos-http-live-streaming-05 which didn't specify fmp4 yet. If it does it wouldn't be the first time their documentation was lacking, though.

src/scripts/browserDeviceProfile.js
knackebrot: It's not only used for remuxing and AC-3 is preferred to AAC.

src/scripts/browserDeviceProfile.js
knackebrot: Can be useful for TV or DVD rips, both use MPEG 1 Layer 2 heavily.
`ffmpeg -i source.mp4 -c:a mp2 output.mp4`

src/scripts/browserDeviceProfile.js
```diff
@@ -383,6 +410,17 @@ import browser from './browser';
...
+        if (canPlayAacVideoAudio) {
+            videoAudioCodecs.push('aac');
+            hlsInTsVideoAudioCodecs.push('aac');
+            hlsInFmp4VideoAudioCodecs.push('aac');
+        }
+
+        if (!browser.ps4 && physicalAudioChannels > 2) {
+            hlsInTsVideoAudioCodecs.push('mp3');
+            hlsInFmp4VideoAudioCodecs.push('mp3');
+        }
```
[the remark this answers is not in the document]
knackebrot: Good catch, fixed

knackebrot:
> Imo, there should be comparison samples attached to this PR is you want to push this.

Attaching samples. Finding the real killers would take a while, and every encoder has some. I downmixed Dolby TrueHD 7.1 to 16-bit stereo. The difference is subtle, but for me, MP3 and AC-3 have better stereo separation and provide a bit "richer" sound.
[test3.zip](https://github.com/jellyfin/jellyfin-web/files/6177838/test3.zip)
```
ffmpeg -i "test3.flac" -c:a aac -b:a 192k -f mp4 -movflags +faststart "test3.aac.m4a"
ffmpeg -i "test3.flac" -c:a ac3 -b:a 192k -f mp4 -movflags +faststart "test3.ac3.m4a"
ffmpeg -i "test3.flac" -c:a libmp3lame -b:a 192k -f mp4 -movflags +faststart "test3.mp3.m4a
```

> Just basing the entire reasoning on HydrogenAudio and the FFMPEG Wiki is a bit thin, since the former is well known for exaggerating things and the latter just has a basic list without any real comments on it.

FFMPEG Wiki states:
image

MrTimscampi:
> The difference is subtle, but for me, MP3 and AC-3 have better stereo separation and provide a bit "richer" sound.

I don't really hear any noticeable difference, to be honest. Certainly nothing in the realm of richer sound or better separation. 

These all sound mostly the same to me on a pair of Steelseries Arctis 7.

> FFMPEG Wiki states

The ffmpeg wiki is editable publicly. It's as trusty as a non-moderated wikipedia article, really.

In fact, just look at the history of the section you use as reference:
image
image

The references HydrogenAudio thread ([here](https://hydrogenaud.io/index.php?topic=119861.0)) only compares AAC-LC and xHE-AAC. It doesn't pit them against any other format or encoder, and it uses pitiful bitrates, which aren't representative of actual usage in Jellyfin:

> The targeted bitrate was 96kbps and 128kbps.

The author of the wiki edit, megapro17, doesn't seem to be an ffmpeg contributor in any way, except for this wiki edit and a few issues over the years.

I maintain that this is all pretty shaky claims. Both Plex and Emby use AAC as the default as well, afaik, and the Plex documentation even states:

> MP3: Only carries 2 channels and is relatively low bitrate. Not that common in more recent files.
> AAC: The closest thing we can get to a universal standard these days. The drawback with AAC is that while it can carry 6-channel sound, not many receivers handle it natively and it’s not designed for it.
> Dolby Digital / AC3: Probably the most common multi-channel format. It handles at most 6 channels (5.1) and has very wide native support in televisions and receivers. However, it’s relatively low bitrate, so it might not give the best fidelity. Most TV shows have AC3 tracks.

Source: https://support.plex.tv/articles/212072167-audio-configuration-guide-plex-media-player/

This aligns with what we do/should do. Essentially:

* AC3 for everything that supports it of there's > 2 channels
* AAC for the rest
* MP3 on the off-chance there is no AAC support (Which is pretty much never, with our supported targets)

I would be more inclined to trust Plex than a random person on an open wiki.

knackebrot:
> > MP3: Only carries 2 channels and is relatively low bitrate. Not that common in more recent files.
> > AAC: The closest thing we can get to a universal standard these days. The drawback with AAC is that while it can carry 6-channel sound, not many receivers handle it natively and it’s not designed for it.
> > Dolby Digital / AC3: Probably the most common multi-channel format. It handles at most 6 channels (5.1) and has very wide native support in televisions and receivers. However, it’s relatively low bitrate, so it might not give the best fidelity. Most TV shows have AC3 tracks.

That support page essentially presents 3 arguments:
1. MP3 and AC-3 are "low bitrate": Completely irrelevant for us as well as for Plex.
2.  Compatibility: We already handle that.
3.  Popularity: here AAC wins. Pretty much every HLS stream on the Internet has an AAC track.

So, for stereo, where LAME wins?
- it's faster (6x on my laptop)
- it has a mature VBR mode
- it sounds better to me and to a bunch of people on the Internet

And where FFmpeg AAC wins?
- Plex and Emby use it
- seemingly in the heads of every reviewer here :D

Even without changing the default encoder to mp3, I'm still interested in merging the other part of this PR, client speaker count detection. I really dislike that since 10.7.0 I'm getting a 640 kbps 6-channel AAC track by default just because my browser can handle it.

dkanada:
I have no preference between the two but I believe there's already a user setting to control the maximum number of channels for any given client.

knackebrot:
@dkanada 
I know there's a user setting, and I handled that. What I'm trying to make better are the defaults - currently, physicalAudioChannels is set to 6 in every supported browser, with no regard to the actual number of audio channels the user has. My solution retrieves that number.

src/scripts/browserDeviceProfile.js
dmitrylyzo: What if we:
- remove `isSurroundSoundSupportedBrowser`
- use `allowedAudioChannels` if set
- use `numChannels` if `> 0`
- return 2 as default

So if user wants 7.1 on stereo speakers, let it be :upside_down_face:

dmitrylyzo (commented):
I am for splitting PR:
- numChannels
- MP3 (if it is required).
This can also be solved by adding an option to the user settings. But more options means more questions to users during troubleshooting.

PR #2514, opened 2021-03-17
Add support for fillWidth and fillHeight

### LINKED ISSUE #2191: 10.7.2 uses a very low resolution for images in "next up" home section.
**Describe The Bug**
image
Crispy high resolution image. All other sections work fine.

The HTTP request to the image uses this url:
`https://jellyfin/Items/895ed51c4c595d404855abcc06d81d11/Images/Backdrop?maxWidth=355&tag=61a4233cbf0e9cde73d1b8933c4fcb19&quality=80`

The returned image is 355x75. Which is totally wrong.
image

oddstr13:
I have probably missed a few places it should be used, but a quick test did show that the ones I've noticed being a problem where fixed.

Related PRs;
- jellyfin/jellyfin-apiclient-javascript#167
- jellyfin/jellyfin#5556

**Changes**

Use `fillWidth` and `fillHeight` instead of `max*` in some places

**Issues**

Fixes #2191

### DIFF
--- a/src/components/listview/listview.js
+++ b/src/components/listview/listview.js
@@ -76,12 +76,13 @@ import ServerConnections from '../ServerConnections';
         return '';
     }
 
-    function getImageUrl(item, width) {
+    function getImageUrl(item, size) {
         const apiClient = ServerConnections.getApiClient(item.ServerId);
         let itemId;
 
         const options = {
-            maxWidth: width,
+            fillWidth: size,
+            fillHeight: size,
             type: 'Primary'
         };
 
@@ -105,10 +106,11 @@ import ServerConnections from '../ServerConnections';
         return null;
     }
 
-    function getChannelImageUrl(item, width) {
+    function getChannelImageUrl(item, size) {
         const apiClient = ServerConnections.getApiClient(item.ServerId);
         const options = {
-            maxWidth: width,
+            fillWidth: size,
+            fillHeight: size,
             type: 'Primary'
         };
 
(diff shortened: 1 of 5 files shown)


src/components/listview/listview.js
```diff
@@ -76,12 +76,13 @@ import ServerConnections from '../ServerConnections';
...
 
-    function getImageUrl(item, width) {
+    function getImageUrl(item, size) {
         const apiClient = ServerConnections.getApiClient(item.ServerId);
         let itemId;
 
         const options = {
-            maxWidth: width,
+            fillWidth: size,
+            fillHeight: size,
```
thornbill, line 85: Couldn't this result in images being scaled too small when the width is less than the height assuming the width is still what is being passed here?

oddstr13: listview is the small square one, what this is saying is «Scale this image down, but not under this width and height»
image
_(screenshot from current release, not this PR)_

If I ask for fillWidth=1, fillHeight=500, the height will be 500 (assuming the source image is larger), and width will be whatever is needed to keep the ratio.

The browser does the cropping (as opposed to scale up then crop which is what is done now if the ratio is weird – like a banner style wide image here for example)

thornbill: Right, but here and below you are using the same value for width and height. So in your example of width=1 you are requesting fillWidth=1 and fillHeight=1.

If these are only used for square images that makes sense, just wanted to check because it's not immediately clear.

oddstr13: as far as I can tell, the images requested in listview.js are always displayed in a square, yea

thornbill (approved):

PR #2516, opened 2021-03-19
Show current group members in syncplay group dialog

tonyfinn:
**Changes**

Shows the members of the current group in the group dialog when joined, similar to what is done for groups in the group listings.

image

**Issues**

### DIFF
--- a/src/components/syncPlay/core/Manager.js
+++ b/src/components/syncPlay/core/Manager.js
@@ -193,10 +193,19 @@ class Manager {
                 this.queueCore.updatePlayQueue(apiClient, cmd.Data);
                 break;
             case 'UserJoined':
+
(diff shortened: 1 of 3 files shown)


src/components/syncPlay/ui/groupSelectionMenu.css
```diff
@@ -0,0 +1,4 @@
+.syncPlayGroupMenu .actionSheetText {
```
thornbill, line 1: I don't think this was the case when you made this change, but we have moved to using sass everywhere (except for themes) so this should have a `.scss` extension.

dkanada:
@tonyfinn if you can fix the conflicts and address the change request we can probably fit this in the next release.

tonyfinn:
Sure, I'll rebase it tomorrow after work

thornbill (approved):

PR #2519, opened 2021-03-20
Add HEVC in TS for Tizen and webOS

dmitrylyzo:
In #2064, `hevc` in `ts` for Tizen and webOS was removed.

**Changes**
Add HEVC in TS for Tizen and webOS

**Issues**
https://github.com/jellyfin/jellyfin-tizen/issues/69

But there is an issue with multiple video codecs in the transcoding profile: https://github.com/jellyfin/jellyfin/issues/5576

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -463,6 +463,10 @@ import browser from './browser';
             if (!browser.safari) {
                 mp4VideoCodecs.push('hevc');
             }
+
+            if (browser.tizen || browser.web0s) {
+                hlsInTsVideoCodecs.push('hevc');
+            }
         }
 
         if (supportsMpeg2Video()) {


thornbill (approved):

dmitrylyzo:
I have tested this on Tizen 3 and Tizen 4.
HEVC SDR media has `Direct stream` :ballot_box_with_check:
HEVC HDR media is transcoded because of [this](https://github.com/jellyfin/jellyfin/blob/8410a9a26601d754e3c8bed36641c990f1d0e22b/Jellyfin.Api/Helpers/DynamicHlsHelper.cs#L217-L241).
There is some dependence on the order of the codecs: if `hevc` goes first, HEVC HDR has `Direct stream`.

dmitrylyzo:
HDR transcoding is probably a device issue (https://github.com/jellyfin/jellyfin/issues/5576#issuecomment-808703679).
Anyway, DirectPlay shouldn't depend on codec order, but it requires additional changes. So for now, at least we will get a direct HEVC SDR.

PR #2532, opened 2021-03-28
Add controls page to the user settings

### LINKED ISSUE #2531: Honeycomb Bravo USB Controller Seeks Back Playback
Having the Honeycomb Bravo attached which is a USB joystick it causes the Jellyfin web UI to seek back to the start of the video currently playing. 
image

dmitrylyzo:
This PR adds a new page (`Controls`) to the user settings.
It is hidden for `Mobile` layout (we could hide for `Desktop` as well, but we could add keybindings there later).

menu

The page contains only one `Enable Gamepad` option.
- It is app wide and requires a page refresh.
- Disabled by default.

controls

**Changes**
Add gamepad enabling/disabling

**Issues**
Fixes #2531

### DIFF
--- a/src/controllers/user/controls/index.js
+++ b/src/controllers/user/controls/index.js
@@ -1,0 +1,28 @@
+import { Events } from 'jellyfin-apiclient';
+import toast from '../../../components/toast/toast';
+import globalize from '../../../scripts/globalize';
+import appSettings from '../../../scripts/settings/appSettings';
+
+export default function (view, params) {
+    function submit(e) {
+        appSettings.enableGamepad(view.querySelector('.chkEnableGamepad').checked);
+
+        toast(globalize.translate('SettingsSaved'));
+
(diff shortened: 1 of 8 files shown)


src/controllers/user/controls/index.js
```diff
@@ -0,0 +1,28 @@
...
+
+export default function (view, params) {
+    function submit(e) {
+        appSettings.enableGamepad(view.querySelector('.chkEnableGamepad').checked);
+
+        toast(globalize.translate('SettingsSaved'));
+
+        Events.trigger(view, 'saved');
+
+        if (e) e.preventDefault();
```
thornbill, line 14:
```suggestion
        e?.preventDefault();
```

thornbill (changes requested):
I don't have a controller setup to properly test this, but looks good! Just one minor recommendation.

thornbill (approved):

PR #2534, opened 2021-03-28
Improve behaviour of space bar pausing/playing video

### LINKED ISSUE #2376: Spacebar Play/Pause behaviour
Implementation of this behaviour from #805 is really buggy. Very often space key just stops working all of a sudden.

**Steps To Reproduce**
1. Play a video
2. Click on screen to pause it
3. press space to play it
4. Nothing happens

Now wait for a few seconds, and press space again and it works then, it dosn't matter whether OSD is on or not
 
**Expected Behavior**
should play/pause on pressing space, whenver jellyfin is playing media and tan is foccussed. 

**Logs**
[log](https://gist.github.com/kushraj/4416584a6a43c83ab1e4b0415570f7d8)

 - Platform: Arch Linuxj
 - Browser: Firefox
 - Jellyfin Version:  10.6.4

jackrvan:
**Changes**
The issue seemed to stem from us not pausing if the playbackManager is set. This would work if the focused element was the pause button since that is handled differently. I modified the code to get the current focused element. If the element is the pause button then we do nothing since the space bar will behave the same as if we hit the button with the mouse. If that is not the case then we call playPause() to manually play/pause the video.

**Issues**
Fixes #2376

### DIFF
--- a/src/controllers/playback/video/index.js
+++ b/src/controllers/playback/video/index.js
@@ -992,9 +992,13 @@ import { appRouter } from '../../../components/appRouter';
 
             const key = keyboardnavigation.getKeyName(e);
             const isKeyModified = e.ctrlKey || e.altKey || e.metaKey;
+            const currentElement = document.activeElement;
 
-            if (!currentVisibleMenu && e.keyCode === 32) {
-                playbackManager.playPause(currentPlayer);
+            if (e.keyCode === 32) {
+                if (!currentElement.className.split(' ').includes('btnPause')) {
+                    // If the focused button is the pause button it will already play/pause it
+                    playbackManager.playPause(currentPlayer);
+                }
                 showOsd();
                 return;
             }
--- a/src/controllers/playback/video/index.js
+++ b/src/controllers/playback/video/index.js
@@ -1,5 +1,6 @@
 import { playbackManager } from '../../../components/playback/playbackmanager';
 import SyncPlay from '../../../components/syncPlay/core';
+import browser from '../../../scripts/browser';
 import dom from '../../../scripts/dom';
 import inputManager from '../../../scripts/inputManager';
 import mouseManager from '../../../scripts/mouseManager';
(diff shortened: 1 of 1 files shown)


src/controllers/playback/video/index.js
```diff
@@ -992,9 +992,13 @@ import { appRouter } from '../../../components/appRouter';
 
             const key = keyboardnavigation.getKeyName(e);
             const isKeyModified = e.ctrlKey || e.altKey || e.metaKey;
+            const currentElement = document.activeElement;
```
dmitrylyzo, line 995: You can use `clickedElement`.
```suggestion
```

src/controllers/playback/video/index.js
```diff
@@ -992,9 +992,13 @@ import { appRouter } from '../../../components/appRouter';
...
 
-            if (!currentVisibleMenu && e.keyCode === 32) {
-                playbackManager.playPause(currentPlayer);
+            if (e.keyCode === 32) {
+                if (!currentElement.className.split(' ').includes('btnPause')) {
+                    // If the focused button is the pause button it will already play/pause it
```
dmitrylyzo, line 999: Skip any `BUTTON` element.
```suggestion
                // Let the button perform its action
                if (clickedElement.tagName !== 'BUTTON') {
```

dmitrylyzo (commented):
In TV layout, if you stand on `Rewind`, both actions will be performed: pause and rewind.

In addition to suggestions, you can also move `if` body into `switch` above `case 'k'` and use `case 'Space'`.

dmitrylyzo:
My suggestions breaks fix in `Desktop` layout. :thinking:

dmitrylyzo:
I found solution.  It touches unchanged lines. Do you mind if I push it in your PR?

jackrvan:
> 
> 
> I found solution. It touches unchanged lines. Do you mind if I push it in your PR?

Nope please go ahead I was planning on taking a look tonight but go ahead. Id love to take a look at the solution since Im new to this could be nice to see how you went about it

dmitrylyzo:
I have pushed changes:
- We do `pause` if not a button was clicked or not in TV layout.
- I extended workaround for Firefox bug to all buttons: it emits `click` even if `preventDefault` was used on `keydown`.
I don't remember if Edge still needs it, but workaround is probably harmless.

thornbill:
I am still seeing buggy behavior in Firefox with this change. If I click to pause, pressing spacebar does not start playback until the OSD hides.

dmitrylyzo:
> I am still seeing buggy behavior in Firefox with this change. If I click to pause, pressing spacebar does not start playback until the OSD hides.

I can't reproduce this in Firefox 89.0 (Desktop and TV layouts).
1. Start the playback
2. Click the `Pause` button / click the player -> _playback paused_
3. Press the spacebar (OSD visible) -> _playback starts_

thornbill:
It must have just been a caching issue. It seems to be working as expected now! :rocket:

thornbill (approved):

PR #2535, opened 2021-03-29
Ask for playback permission in SyncPlay

OancaAndrei:
**Changes**

- Show dialog to user when playback permission needs to be granted.
- Avoid spoilers caused by incorrectly jumping to a position when playback is paused.

**Issues**

Should partially fix [jellyfin/#4680](https://github.com/jellyfin/jellyfin/issues/4680) (clients won't be stuck anymore).


src/components/syncPlay/core/Manager.js
```diff
@@ -308,6 +308,7 @@ class Manager {
      * @param {Object} apiClient The ApiClient.
      */
     resumeGroupPlayback(apiClient) {
+        // TODO: rename this method, it's not clear what it does.
         this.followGroupPlayback(apiClient).then(() => {
```
thornbill, line 312: Should these methods be renamed in this PR?

OancaAndrei: No, not really. I've left those there as a personal reminder. I'm not sure whether I was going to fix them in another open PR.
Is it ok to leave TODOs around like this? I can remove them otherwise.

thornbill (changes requested):
This appears to be broken on iOS. It always asks for permission to play, but then doesn't play anything even though I click allow. It even asks when the iOS device is the one initiating playback.

OancaAndrei:
> This appears to be broken on iOS. It always asks for permission to play, but then doesn't play anything even though I click allow. It even asks when the iOS device is the one initiating playback.

Thanks for testing this on iOS. I'll have to spin up a simulator to see what's going on :thinking:.

PR #2538, opened 2021-03-30
Fix ScrollManager for dialogs

dmitrylyzo:
**Changes**
- Add `smoothScrollX` and `smoothScrollY` as scrollers.
- Treat dialogs as a scrolling scope.

**Issues**
- ScrollManager doesn't work on `smoothScroll`* containers.
It gets the document as a scroller, so navigating within dialog (for example, sort dialog) causes the document to scroll.
- Dialogs are `fixed`. So ScrollManager doesn't scroll element to the center.

### DIFF
--- a/src/components/scrollManager.js
+++ b/src/components/scrollManager.js
@@ -201,6 +201,19 @@ import layoutManager from './layoutManager';
      */
     const documentScroller = new DocumentScroller();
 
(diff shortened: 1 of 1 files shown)


dmitrylyzo:
Close in favor of #2613

PR #2539, opened 2021-03-31
Changed calculation for media info run time to render hours

### LINKED ISSUE #2551: Playlist duration doesn't convert to hours
**Describe the bug**

For very long playlists (such as with movies) duration is still listed in minutes. Example 568 minutes

 - OS: Debian
 - Virtualization: Docker
 - Clients: Browser
 - Browser: Chrome 89
 - Jellyfin Version: 10.7.0

**To Reproduce**

Select a couple movies, add to playlist, view playlist details

**Expected behavior**

Playlist duration should display in hours/minutes when over an hour.

**Screenshots**

image

BrianCArnold:
Altered Media Info runtime calculation to show hours when appropriate.

Previously showed "523 mins", now shows "8 hrs, 43 mins"

**Changes**
Total Minutes are now calculated as (ticks / 600000000).
Rendered minutes are now TotalMinutes % 60.
Rendered hours are now Math.floor(TotalMinutes / 60).

**Issues**
Fixes #2551


Artiume:
Another method to pull total minutes is below. I used it in https://github.com/jellyfin/jellyfin/pull/4891

TimeSpan.FromTicks(runtimeTicks).TotalMinutes;

BrianCArnold:
I'm afraid I don't understand how that would be relevant, since we are currently getting the total time without any issues, the issue is rendering the amount of time using Hours and Minutes when the total playtime exceeds 60 minutes.

BrianCArnold:
I believe everything in your pull request is in C#, this is all in JavaScript, which doesn't have a "TimeSpan" class as far as I'm aware?

Artiume:
I was just offering a method of obtaining minutes without having to math
the ticks. But you might be right, it might be c# only and not js

BrianCArnold:
Ahhh, gotchya. I realized what you meant mere seconds after I left my original comment. I got confused for a bit, since I read it as C# instantly, and I know this is all JS, so even though the actual logic makes perfect sense here, in my head `C# != JS` is so strong that I couldn't even get past the language difference to see why the code was related.

Also, I need sleep really bad.

src/components/mediainfo/mediainfo.js
```diff
@@ -257,11 +257,18 @@ import '../../elements/emby-button/emby-button';
             if (item.Type === 'Audio') {
                 miscInfo.push(datetime.getDisplayRunningTime(item.RunTimeTicks));
             } else {
-                minutes = item.RunTimeTicks / 600000000;
```
thornbill, line 260: Could we just use `getDisplayRunningTime()` here also? It would display in the `00:00:00` format.

thornbill (dismissed):
There is also a lint error that can be fixed. :slightly_smiling_face:

src/components/mediainfo/mediainfo.js
```diff
@@ -257,11 +256,17 @@ import '../../elements/emby-button/emby-button';
...
-
-                miscInfo.push(`${Math.round(minutes)} mins`);
+                const totalMinutes = Math.round(item.RunTimeTicks / 600000000) || 1;
+                const totalHours = Math.floor(totalMinutes / 60);
+                const remainderMinutes = totalMinutes % 60;
+                const result = [];
+                if (totalHours > 0) {
+                    result.push(`${totalHours}h`);
+                }
+                if (totalMinutes > 0) {
```
dmitrylyzo, line 266: This is always `true` because `totalMinutes` is `1` if the rounding result is zero.

thornbill: I debated removing this already, so I'll just remove it lol

dmitrylyzo (approved):

PR #2557, opened 2021-04-07
NativeShell enhancements for Jellyfin Media Player

iwalton3:
This allows the stock jellyfin-web to work as a web client in Jellyfin Media Player.

**Changes**
 - Add player property `useServerPlaybackInfoForAudio` that forces server-side media decisions for music.
 - Add player property `useFullSubtitleUrls` that resolves relative external subtitles URLs into full URLs for MPV.
 - Add player property `syncPlayWrapAs` which tells SyncPlay how to wrap plugins in NativeShell.
 - Add support to SyncPlay for `currentTimeAsync` method required to get accurate time from MPV.
 - Update the mouse movement checkers to also use `clientX`/`clientY` as QtWebEngine returns zeros otherwise.
 - Make PlaybackPermissionManager honor the autoplay capabilities reported by NativeShell.
 - Allow partial implementation of NativeShell interface.


src/components/apphost.js
```diff
@@ -275,7 +275,7 @@ const supportedFeatures = function () {
      */
 function doExit() {
     try {
-        if (window.NativeShell) {
+        if (window.NativeShell && window.NativeShell.AppHost.exit) {
```
thornbill, line 278: I would suggest using optional chaining for this type of checks.

```suggestion
        if (window.NativeShell?.AppHost?.exit) {
```

src/components/syncPlay/ui/playbackPermissionManager.js
```diff
@@ -32,8 +34,12 @@ class PlaybackPermissionManager {
      * Tests playback permission. Grabs the permission when called inside a click event (or any other valid user interaction).
      * @returns {Promise} Promise that resolves succesfully if playback permission is allowed.
      */
-    check () {
-        return new Promise((resolve, reject) => {
+    async check () {
```
thornbill, line 37: async/await doesn't seem necessary here. I think you could just return `Promise.resolve(true)` below and remove the await when returning the promise.

src/components/upnextdialog/upnextdialog.js
```diff
@@ -159,17 +159,21 @@ import '../../assets/css/flexstyles.scss';
...
-        void elem.offsetWidth;
-
-        elem.classList.add('upNextDialog-hidden');
-
         const fn = onHideAnimationComplete.bind(instance);
         instance._onHideAnimationComplete = fn;
 
-        dom.addEventListener(elem, transitionEndEventName, fn, {
-            once: true
+        const e = await new Promise((resolve) => {
```
thornbill, line 165: A more descriptive variable name might be good here instead of "e".

thornbill (changes requested):
Looks good! Just a couple minor notes.

thornbill (approved):
:+1: lgtm. Have you tested this with any of the other wrapper apps?

iwalton3:
I have tested with JMP and the web app only.

PR #2558, opened 2021-04-07
Add server selection response handling

dmitrylyzo:
If server is down or server hash has been changed, connection state is `ServerSelection`.
Redirect user to the server selection page.
This is not the best solution, but at least user won't get hung up with "Circle Of Death".

**Changes**
Add server selection response handling

**Issues**
This issue occurs on standalone and bundled web.
https://github.com/jellyfin/jellyfin-tizen/issues/75

_Review with `Hide whitespace changes`_

### DIFF
--- a/src/components/appRouter.js
+++ b/src/components/appRouter.js
@@ -503,23 +503,29 @@ class AppRouter {
         const firstResult = this.firstConnectionResult;
 
         this.firstConnectionResult = null;
-        if (firstResult && firstResult.State === 'ServerSignIn') {
-            const url = firstResult.ApiClient.serverAddress() + '/System/Info/Public';
-            fetch(url).then(response => {
-                if (!response.ok) return Promise.reject('fetch failed');
-                return response.json();
-            }).then(data => {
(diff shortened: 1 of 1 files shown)


dmitrylyzo:
Close in favor of https://github.com/jellyfin/jellyfin-apiclient-javascript/pull/187

It will be handled correctly by this https://github.com/jellyfin/jellyfin-web/blob/98814a5b66813c4c23fcd9f4254239b89b939506/src/components/appRouter.js#L532-L536

dmitrylyzo:
Web gets stuck if the server was selected and turned off (for bundled/standalone web) or changed its hash, but you were not logged in:
`!shouldExitApp -> true`
`!apiClient -> false`
`!apiClient.isLoggedIn() -> true`
`!route.anonymous -> false` - login page

**Steps To Reproduce**
1. Connect to the server but do not log in.
2. Change server hash in `localStorage`.
3. Refresh page (login page).
4. Get "Circle of Death" issue.

To test a server down situation, you need a standalone or bundled web (app).

PR #2572, opened 2021-04-09
Fix music videos on artist and album page

Ullmie02:
**Changes**
Album detail page:
- Use correct query selector for music video content

Artist detail page:
- Use ArtistIds instead of AlbumArtistIds as described in issue
- Use correct card shape for music videos

**Issues**
Fixes jellyfin/jellyfin#5156


Ullmie02:
Maybe we can backport this to 10.7?

MrTimscampi (approved):

PR #2594, opened 2021-04-14
chore(ci): replace with actions-label-merge-conflict

ferferga:
I tested the query this action performs in GraphQL explorer and checked the source code and it should fit all our needs. It will label PRs with conflicts only, **without paying attention to the rest of the conditions (reviews, rebase, etc)**.

It's also maintained by its owner, while the original action had a PR from me (to add the exact behaviour we want) sitting for days so, in the long-term, this action should be a better solution.

Needs the ``GH_TOKEN`` added by @anthonylavado so the label is applied by jellyfin-bot instead of github-actions

### DIFF
--- a/.github/workflows/automation.yml
+++ b/.github/workflows/automation.yml
@@ -10,6 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: eps1lon/actions-label-merge-conflict@v2.0.1
+        if: ${{ github.repository == 'jellyfin/jellyfin-web' }}
         with:
           dirtyLabel: 'merge conflict'
           repoToken: ${{ secrets.JF_BOT_TOKEN }}
--- a/.github/workflows/automation.yml
+++ b/.github/workflows/automation.yml
@@ -1,17 +1,15 @@
-name: 'Merge Conflicts'
+name: 'Automation'
 
 on:
   push:
-    branches:
-      - master
   pull_request_target:
-    types:
-      - synchronize
+
 jobs:
   triage:
+    name: 'Merge conflict labeling'
     runs-on: ubuntu-latest
     steps:
       - uses: eps1lon/actions-label-merge-conflict@v2.0.1
         with:
           dirtyLabel: 'merge conflict'
-          repoToken: ${{ secrets.GH_TOKEN }}
+          repoToken: ${{ secrets.JF_BOT_TOKEN }}
--- a/.github/workflows/merge-conflicts.yml
+++ b/.github/workflows/merge-conflicts.yml
@@ -1,18 +1,17 @@
-name: "Merge Conflicts"
+name: 'Merge Conflicts'
 
 on:
   push:
     branches:
       - master
-  pull_request:
+  pull_request_target:
     types:
       - synchronize
 jobs:
   triage:
     runs-on: ubuntu-latest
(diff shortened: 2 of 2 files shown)


nielsvanvelzen (approved):

.github/workflows/merge-conflicts.yml
```diff
@@ -1,18 +1,17 @@
...
     branches:
       - master
-  pull_request:
+  pull_request_target:
     types:
       - synchronize
 jobs:
   triage:
     runs-on: ubuntu-latest
-    if: github.repository == 'jellyfin/jellyfin-web'
```
thornbill, line 13: Is this not needed to prevent running on forks?

.github/workflows/merge-conflicts.yml
```diff
@@ -1,18 +1,17 @@
-name: "Merge Conflicts"
+name: 'Merge Conflicts'
 
 on:
   push:
     branches:
       - master
```
nielsvanvelzen, line 6: Allowing it on all pushes helps with multiple PR's for feature branches.

.github/workflows/merge-conflicts.yml
nielsvanvelzen: Yes it is. I 'd say it should be kept.

nielsvanvelzen (changes requested):

.github/workflows/merge-conflicts.yml
ferferga: @thornbill @nielsvanvelzen Not needed with ``pull_request_target`` trigger

ferferga:
I formatted the workflow as we're doing in some of our other repos by renaming the workflow to "Automation".

.github/workflows/automation.yml
```diff
@@ -0,0 +1,15 @@
+name: 'Automation'
+
+on:
+  push:
+  pull_request_target:
+
+jobs:
+  triage:
+    name: 'Merge conflict labeling'
+    runs-on: ubuntu-latest
```
nielsvanvelzen, line 10:
```suggestion
    runs-on: ubuntu-latest
    if: github.repository == 'jellyfin/jellyfin-web'
```

Forks do not need this workflow, workflows should be disabled by default for forks unless they are useful for contributor forks. Every time I update my web or server fork I instantly get emails from failing builds because of these scripts that I don't need on my repo.

Also same job is disabled in at least the Android TV, Android Mobile and Kotlin SDK repositories.

nielsvanvelzen (changes requested):

.github/workflows/automation.yml
ferferga: I checked those repos and they indeed have their condition from when we used the old action and ``pull_request`` event.

With ``pull_request_target`` is [not necessary at all](https://docs.github.com/actions/reference/events-that-trigger-workflows#pull_request_target) as the action runs in the context of this repo even in forks. No issues in Vue without the condition.

I'm adding it regardless to stay in the safe side 👍

nielsvanvelzen: > I'm adding it regardless to stay in the safe side 👍

I think that's the best way to do it with GH Workflows, especially when using custom secrets.

nielsvanvelzen (approved):

thornbill (approved):

PR #2599, opened 2021-04-19
Use episode images in the "Next Up" section

### LINKED ISSUE #329: (optional) Use Episode image for NEXT UP / Continue Watching
I would like the Episode image to be used instead of show thumbnail, on the home screen.
if you click on NEXT UP > it does show the episode images.

Screenshot at 2019-05-19 20:05:04
Screenshot at 2019-05-19 20:05:24

omeryagmurlu:
"Next Up" Section currently shows a parent image if it can't find a suitable thumbnail for the episode (which is often the case). It then falls back to parent thumbnails, which is unintuitive for the "Next Up" section.

**Changes**
Added `inheritThumb: false` option to `getNextUpItemsHtml` in order to disallow it from using parent images if possible. This way, it falls back to extracted images before parent images.

**Issues**
Fixes #329, https://features.jellyfin.org/posts/152/optional-use-episode-image-for-next-up-continue-watching

### DIFF
--- a/src/components/homesections/homesections.js
+++ b/src/components/homesections/homesections.js
@@ -449,7 +449,7 @@ import ServerConnections from '../ServerConnections';
                 showYear: true,
                 lines: 2
             });
-        }
+        };
     }
 
     function getOnNowFetchFn(serverId) {
--- a/src/components/homesections/homesections.js
+++ b/src/components/homesections/homesections.js
@@ -144,17 +144,17 @@ import ServerConnections from '../ServerConnections';
         } else if (section === 'librarybuttons') {
             loadlibraryButtons(elem, apiClient, user, userSettings, userViews);
         } else if (section === 'resume') {
-            return loadResume(elem, apiClient, 'HeaderContinueWatching', 'Video');
+            return loadResume(elem, apiClient, 'HeaderContinueWatching', 'Video', userSettings);
         } else if (section === 'resumeaudio') {
-            return loadResume(elem, apiClient, 'HeaderContinueListening', 'Audio');
+            return loadResume(elem, apiClient, 'HeaderContinueListening', 'Audio', userSettings);
         } else if (section === 'activerecordings') {
             loadLatestLiveTvRecordings(elem, true, apiClient);
         } else if (section === 'nextup') {
-            loadNextUp(elem, apiClient);
+            loadNextUp(elem, apiClient, userSettings);
         } else if (section === 'onnow' || section === 'livetv') {
             return loadOnNow(elem, apiClient, user);
         } else if (section === 'resumebook') {
-            return loadResume(elem, apiClient, 'HeaderContinueReading', 'Book');
+            return loadResume(elem, apiClient, 'HeaderContinueReading', 'Book', userSettings);
         } else {
             elem.innerHTML = '';
             return Promise.resolve();
@@ -374,7 +374,7 @@ import ServerConnections from '../ServerConnections';
         'Video': 'videoplayback,markplayed'
     };
 
-    function loadResume(elem, apiClient, headerText, mediaType) {
+    function loadResume(elem, apiClient, headerText, mediaType, userSettings) {
         let html = '';
 
         const dataMonitor = dataMonitorHints[mediaType] || 'markplayed';
@@ -397,7 +397,7 @@ import ServerConnections from '../ServerConnections';
 
         const itemsContainer = elem.querySelector('.itemsContainer');
         itemsContainer.fetchData = getItemsToResumeFn(mediaType, apiClient.serverId());
-        itemsContainer.getItemsHtml = getItemsToResumeHtml;
+        itemsContainer.getItemsHtml = getItemsToResumeHtmlFn(userSettings.useEpisodeImagesInNextUpAndResume());
         itemsContainer.parentContainer = elem;
     }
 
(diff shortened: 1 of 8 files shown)


Artiume:
Could this be an option instead of changing the behavior? I like to show the shows image whereas some may want the episode image. I think this is similar to the 10.7 home image changes, we have people requesting the old images functionality restored

omeryagmurlu:
Yeah, it makes sense to put this as an option, probably under User -> Home. I'm not yet that familiar with the codebase, gonna look around and see how it's done

Artiume:
here's a PR showing how an option was added to the user section

https://github.com/jellyfin/jellyfin-web/pull/2064

let us know if you have any questions :)

omeryagmurlu:
Added the following checkbox:
image

It controls the images shown in Next Up and Continue Watching section in Home, List and TV Suggestions pages.

Artiume:
Awesome! Appreciate the effort. 

instead of saying parent image, should we say prinary Show poster or something to that effect? end users won't necessarily know what it means by the parent image

omeryagmurlu:
Changed the help text to: "Next Up and Continue Watching sections will use episode images as thumbnails instead of the primary thumbnail of the show."

I'm not sure if I should fix the failing SonarCloud warning, all other settings are done in exactly the same way, and I don't want to change unrelated stuff with this pull request.

src/scripts/settings/userSettings.js
```diff
@@ -253,6 +253,20 @@ export class UserSettings {
         return val !== 'false';
     }
 
+    /**
+     * Get or set 'Use Episode Images in Next Up' state.
+     * @param {boolean|undefined} val - Flag to enable 'Use Episode Images in Next Up' or undefined.
```
thornbill, line 258: Using `string|boolean|undefined` here _might_ make the SonarQube check pass.

src/components/homesections/homesections.js
```diff
@@ -392,28 +392,31 @@ import ServerConnections from '../ServerConnections';
...
+                allowBottomPadding: false,
+                cardLayout: cardLayout,
+                showYear: true,
+                lines: 2
+            });
+        };
     }
 
-    function loadResumeVideo(elem, apiClient) {
+    function loadResumeVideo(elem, apiClient, user, userSettings) {
```
thornbill, line 419: `user` seems to be unused?

src/components/homesections/homesections.js
```diff
@@ -669,25 +672,28 @@ import ServerConnections from '../ServerConnections';
...
+                context: 'home',
+                centerText: !cardLayout,
+                allowBottomPadding: !enableScrollX(),
+                cardLayout: cardLayout
+            });
+        };
     }
 
-    function loadNextUp(elem, apiClient) {
+    function loadNextUp(elem, apiClient, user, userSettings) {
```
thornbill, line 696: `user` seems to be unused?

thornbill (changes requested):

src/components/homesections/homesections.js
omeryagmurlu: It's unused, but all the other `load*` functions had similar signatures (see line 134), that's why I added that in. Can easily remove it if that's not a concern.

omeryagmurlu (commented):
made the requested changes

src/scripts/settings/userSettings.js
dmitrylyzo: I prefer this:
```js
return this.get('useEpisodeImagesInNextUpAndResume', false) === 'true';
```

thornbill:
It would probably make sense to include this on the playback settings screen with the other "next up" options as seen below.

Screenshot_2021-05-04 Jellyfin

omeryagmurlu:
This is not related to playback, so I don't think that it would make sense to include it there. This option only changes which image you see on the 'Continue Watching' and 'Next Up' sections and changes some other default image preference logic to make sure episode images are used whenever possible instead of parent images. It doesn't have to anything with playback, placing it under Display is more appropiate in my opinion

omeryagmurlu:
Hi, is there anything missing in this PR blocking a review/merge? Did I miss something?

src/scripts/settings/userSettings.js
```diff
@@ -253,6 +253,20 @@ export class UserSettings {
...
+     * @return {boolean} 'Use Episode Images in Next Up' state.
+     */
+    useEpisodeImagesInNextUpAndResume(val) {
+        if (val !== undefined) {
+            return this.set('useEpisodeImagesInNextUpAndResume', val.toString(), false);
+        }
+
+        val = this.get('useEpisodeImagesInNextUpAndResume', false);
+        return val === 'true';
+    }
```
dmitrylyzo, line 268: Send settings to the server. This way, the settings will be in-sync on different clients for the same user.
```suggestion
    useEpisodeImagesInNextUpAndResume(val) {
        if (val !== undefined) {
            return this.set('useEpisodeImagesInNextUpAndResume', val.toString(), true);
        }

        val = this.get('useEpisodeImagesInNextUpAndResume', true);
        return val === 'true';
    }
```

dmitrylyzo (approved):

omeryagmurlu:
@thornbill I've made the requested changes, can you review this PR again?

thornbill (approved):
Initially I thought this was changing the behavior of the next up popup at the end of an episode which is why I thought it would be better in the Playback settings.

This seems to be working as expected though and can be merged as soon as the conflicts are resolved! :rocket:

src/components/homesections/homesections.js
```diff
@@ -428,25 +428,28 @@ import ServerConnections from '../ServerConnections';
...
+                showDetailsMenu: true,
+                overlayPlayButton: true,
+                context: 'home',
+                centerText: !cardLayout,
+                allowBottomPadding: false,
+                cardLayout: cardLayout,
+                showYear: true,
+                lines: 2
+            });
+        }
```
thornbill, line 452: Looks like there is a lint error here now due to a missing semicolon.

omeryagmurlu: Fixed. Apparently I've overlooked the mail from Github Actions about that problem. Everything should be fine right now.

thornbill: That is easy to do. :sweat_smile:

Issue #2609, opened 2021-04-21
While in TV Mode the scroll bar becomes un-themed.


n8detar:
**Describe The Bug**
When the web client is in TV Mode (or Mobile Mode) the scroll bar becomes un-themed.

**Steps To Reproduce**

1. Enable TV Mode  (User Settings -> Display -> Display mode -> TV (or Mobile) -> Save -> Refresh Page)
2. Scroll bar is now un-themed.

**Expected Behavior**
With TV Mode enabled you would expect the scroll bar to be themed the same way as in Desktop Mode.

**Logs**

**Screenshots**

Desktop Mode:
![Desktop](https://user-images.githubusercontent.com/24966292/115631241-4464b700-a2ba-11eb-8391-f56e35c5b850.png)
TV Mode:
![TV](https://user-images.githubusercontent.com/24966292/115631248-4a5a9800-a2ba-11eb-8aee-eed52ee0ffd8.png)

**System (please complete the following information):**
 - Platform: Linux Docker
 - Browser: Chrome
 - Jellyfin Version: 10.7.2

**Additional Context**

dmitrylyzo:
This is because width/height style isn't applied for TV layout: https://github.com/jellyfin/jellyfin-web/blob/1794ab53febc28b249651d7f886c217bfdfab76c/src/themes/dark/theme.css#L453-L456

We have 2 options:
- Hide scrollbars on TV (https://github.com/dmitrylyzo/jellyfin-web/tree/fix-tv-scrollbar-style)
- Make scrollbars styled as on Desktop (https://github.com/dmitrylyzo/jellyfin-web/tree/fix-tv-scrollbar-style-2)

n8detar:
I don't use mobile, I thought I would just point it out. As for TV Mode, I don't know which would be better, it would be more consistent to make them themed the same way as on desktop, but since it is in TV Mode a scroll bar isn't really necessary.

dmitrylyzo:
Fixed by #2617

PR #2617, opened 2021-04-24
Set scrollbar width in TV mode.

iwalton3:
**Changes**
 - Set the scrollbar width in TV mode. It seems this is required for the other custom scrollbar styling to take effect, at least in JMP.

Before:
image

After:
image

*This has been tested and can be backported if desired.*


dmitrylyzo:
Fixes #2609
:smile: https://github.com/jellyfin/jellyfin-web/issues/2609#issuecomment-824828179

dmitrylyzo (approved):

nielsvanvelzen (approved):

PR #2618, opened 2021-04-24
Allow client custom css and disabling global server custom css.

iwalton3:
**Changes**
 - Allow users to disable custom css set by the server owner.
      - This is important as custom css can break things, such as TV mode.
      - Making users able to disable it without asking a server owner is valuable for troubleshooting. 
 - Allow users to set their own custom css locally.

image

*This has been tested and can be backported if desired.*


iwalton3:
The SonarCloud code analysis complaint for this is incorrect, as if that were true the feature would not work.
image

src/scripts/settings/userSettings.js
```diff
@@ -239,6 +255,33 @@ export class UserSettings {
...
+     * @param {boolean|undefined} val - Flag to enable 'disableCustomCss' or undefined.
+     * @return {boolean} 'disableCustomCss' state.
+     */
+    disableCustomCss(val) {
+        if (val !== undefined) {
+            return this.set('disableCustomCss', val.toString(), false);
+        }
+
+        val = this.get('disableCustomCss', false);
+        return val === 'true';
```
dmitrylyzo, line 269: I think this will calm SonarCloud
```suggestion
        return this.get('disableCustomCss', false) === 'true';
```

thornbill (approved):

PR #2621, opened 2021-04-24
Add subtitle/audio auto-set feature.

iwalton3:
This adds an option under Playback Settings:
image

If you turn this on, it'll automatically set your subtitle and audio tracks to match the ones in the previous video track. For instance, if you are watching Anime and all of the videos default to English Dubbed, if you select the Japanese and Full Subtitles tracks, this will ensure you can keep watching episodes uninterrupted with the same desired subtitle setting.

This is no simple problem to solve, as metadata and track names vary. The algorithm used is as follows:

If the previous track is a subtitle and none was used, do not use one for the next track.
Compute a score for all possible tracks, comparing the values:
|Attribute|Points|
|:-:|:-:|
|Codec|1|
|Track Id|1|
|Title|2|
|Language|2|

Use the track with the highest score. If no track has at least 3 points, then use the default.


ffiarpg:
How does this work if you go from watching one episode of japanese audio english subtitle anime to an english audio film with multiple languages of subtitles? Would it turn on english subtitles with english audio because it would matches language and title/track id to exceed 3 points?

iwalton3:
Yes that would happen if the track name or id number happened to be the same.

Pyxyll:
Would love to see this implimented.

seelenamt:
I'd also like to see this implemented.

lxqueen:
Is it also worth giving the track's disposition a score as well? I don't believe I've come across many videos that have this in the metadata, but it may help with a little additional boost to accuracy. _(Hell, I didn't even realise this was an option until recently.)_

**Pulled from [https://ffmpeg.org/ffmpeg.html](https://ffmpeg.org/ffmpeg.html):**

`-disposition[:stream_specifier] value (output,per-stream)`

Sets the disposition for a stream.

This option overrides the disposition copied from the input stream. It is also possible to delete the disposition by setting it to 0.

The following dispositions are recognized:

- default
- dub
- original
- comment
- lyrics
- karaoke
- forced
- hearing_impaired
- visual_impaired
- clean_effects
- attached_pic
- captions
- descriptions
- dependent
- metadata

iwalton3:
Yeah I’ve never seen those fields actually set but it would be so nice if they were used. It take a lot of the guess work out of this.

whiteworthy:
If there are both PGS and SRT subtitles which one would it select? SRT has .default on filename if that matters.

iwalton3:
> SRT has .default on filename if that matters.

This only affects playback of several videos. So if the last video was an SRT with similar metadata, it will use it for subsequent playback. For the first video, it’ll use the default or the user selection.

thornbill (approved):

cactushydrocodone:
Ive tested this with the newest jellyfin unstable image and it does not work at all.
The following i did try:
1. Select english for show
2. Start show
3. Jump to 75% of the show
4. Wait 1 minute
5. Skip to next episode of show
6. next episode is not in english, but in french (its a dual language file).

Ive looked at the log, but didnt find anything particular poking out.

mo-st:
Hi, thanks for your contribution! Any chance we could get this setting on the androidTV app as well?

PR #2627, opened 2021-04-26
Added ability to burn in all subtitles, including SRT.

BrianCArnold:
Current version of Jellyfin-Web doesn't allow users to select "All" for burning in subtitles. Some users appear to prefer always burning in subtitles.

**Changes**
Added dropdown option for "All" to subtitle burn in.

**Issues**
N/A

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -944,21 +944,23 @@ import browser from './browser';
         // Subtitle profiles
         // External vtt or burn in
         profile.SubtitleProfiles = [];
-        if (supportsTextTracks()) {
-            profile.SubtitleProfiles.push({
-                Format: 'vtt',
-                Method: 'External'
-            });
-        }
-        if (options.enableSsaRender !== false && (!options.isRetry && appSettings.get('subtitleburnin') !== 'allcomplexformats')) {
-            profile.SubtitleProfiles.push({
-                Format: 'ass',
-                Method: 'External'
-            });
-            profile.SubtitleProfiles.push({
-                Format: 'ssa',
-                Method: 'External'
-            });
+        if (!options.isRetry && appSettings.get('subtitleburnin') !== 'all') {
+            if (supportsTextTracks()) {
+                profile.SubtitleProfiles.push({
+                    Format: 'vtt',
+                    Method: 'External'
+                });
+            }
+            if (options.enableSsaRender !== false && appSettings.get('subtitleburnin') !== 'allcomplexformats') {
+                profile.SubtitleProfiles.push({
+                    Format: 'ass',
+                    Method: 'External'
+                });
+                profile.SubtitleProfiles.push({
+                    Format: 'ssa',
+                    Method: 'External'
+                });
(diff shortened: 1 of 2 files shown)


src/scripts/browserDeviceProfile.js
```diff
@@ -944,21 +944,23 @@ import browser from './browser';
...
-        if (options.enableSsaRender !== false && (!options.isRetry && appSettings.get('subtitleburnin') !== 'allcomplexformats')) {
-            profile.SubtitleProfiles.push({
-                Format: 'ass',
-                Method: 'External'
-            });
-            profile.SubtitleProfiles.push({
-                Format: 'ssa',
-                Method: 'External'
-            });
+        if (!options.isRetry && appSettings.get('subtitleburnin') !== 'all') {
```
dmitrylyzo, line 947: Get `subtitleburnin` once.

dmitrylyzo:
Works. Tested in Chrome on PC (internal/external ASS/SRT). I am sure that it will work on TVs.

On Android, you cannot set the option because the app doesn't specify `subtitleburnsettings` feature.

This probably won't work on iOS because it generates profile by itself (adds `vtt`).

@BrianCArnold Could you rebase onto the most recent master?
_I tested it after the rebasing._

BrianCArnold:
I'll rebase here shortly.

src/scripts/browserDeviceProfile.js
```diff
@@ -944,21 +944,24 @@ import browser from './browser';
...
+            if (options.enableSsaRender !== false && subtitleBurninSetting !== 'allcomplexformats') {
+                profile.SubtitleProfiles.push({
+                    Format: 'ass',
+                    Method: 'External'
+                });
+                profile.SubtitleProfiles.push({
+                    Format: 'ssa',
+                    Method: 'External'
+                });
+            }
```
dmitrylyzo, line 964: _from Matrix_
To respect the old logic:
- burn in text subtitles only by user's decision.
- burn in SSA/ASS if playback error occurs or by user's decision.
`SubtitlesOctopus` can throw `mediadecodeerror`. _It might be better to have a different error._
https://github.com/jellyfin/jellyfin-web/blob/d96b6ac4ee6899e3d389250eee10de2634785975/src/plugins/htmlVideoPlayer/plugin.js#L1067-L1069

```suggestion
        const subtitleBurninSetting = appSettings.get('subtitleburnin');
        if (subtitleBurninSetting !== 'all') {
            if (supportsTextTracks()) {
                profile.SubtitleProfiles.push({
                    Format: 'vtt',
                    Method: 'External'
                });
            }
            if (options.enableSsaRender !== false && !options.isRetry && subtitleBurninSetting !== 'allcomplexformats') {
                profile.SubtitleProfiles.push({
                    Format: 'ass',
                    Method: 'External'
                });
                profile.SubtitleProfiles.push({
                    Format: 'ssa',
                    Method: 'External'
                });
            }
```

dmitrylyzo (approved):
SonarCloud gone crazy. None related to this PR.

PR #2646, opened 2021-05-03
Prevent Buttons from Sitting Over Text

shortspider:
**Changes**
Tweaked the CSS to prevent buttons from sitting over text on small screens like this:
image

Please note that I'm pretty hopeless at CSS, so there may be a better way to do this but I can't figure it out.


dkanada:
@shortspider can you rebase this against the master branch?

shortspider:
@dkanada done.

dkanada (approved):

PR #2672, opened 2021-05-18
Remove OPUS from supported HLS audio formats

nyanmisaka:
**Changes**
- Remove OPUS from supported HLS audio formats

OPUS is not a standard HLS audio format, it's not mentioned in Apple's HLS document.
Remuxing/Direct Streaming films with OPUS audio to Android Chrome will result in no sound.
Thus leave it only for DirectPlay and Tizen.

**Issues**
Fixes https://github.com/jellyfin/jellyfin/issues/5116

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -414,8 +414,10 @@ import browser from './browser';
 
         if (canPlayAudioFormat('opus')) {
             videoAudioCodecs.push('opus');
-            hlsInTsVideoAudioCodecs.push('opus');
             webmAudioCodecs.push('opus');
+            if (browser.tizen) {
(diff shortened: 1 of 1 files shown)


cvium (approved):

dmitrylyzo:
Couldn't it be that Samsung and LG support it despite the standard (as it is with HEVC in TS)?

nyanmisaka:
> Couldn't it be that Samsung and LG support it despite the standard (as it is with HEVC in TS)?

OPUS has been disabled on WebOS in the code.
I don't have a Samsung TV to test but OPUS is supported in TS as per their docs. I will leave it only for Tizen.

dmitrylyzo (approved):

thornbill (approved):

PR #2682, opened 2021-05-27
Album Art to not force media buttons behind playlist buttons for 

phcreery:
**Changes**

Changes Mobile Now Playing (Remote Control) Album Art .scss to prevent it from pushing the Media Navigation buttons behind the Playlist buttons (rendering the former inoperable).

Affected by mobile devices with short screens. (Just taller than square. ex. iPhone SE web client)

Before/After
image

### DIFF
--- a/src/components/remotecontrol/remotecontrol.scss
+++ b/src/components/remotecontrol/remotecontrol.scss
@@ -328,6 +328,7 @@
     .nowPlayingPageImageContainer {
         width: 100%;
         margin: auto auto 0.5em;
+        height: calc(100% - 20vh - 4.2em);
     }
 
     .nowPlayingPageImageContainerNoAlbum .cardImageContainer .cardImageIcon {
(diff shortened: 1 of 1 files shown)


thornbill (approved):
Appears to be working as expected!

It does look like this will need rebased for the expected CI tasks to run.

src/components/remotecontrol/remotecontrol.scss
```diff
@@ -399,7 +400,12 @@
     }
 
     .nowPlayingPageImage.nowPlayingPageImageAudio {
-        width: 100%;
+        width: auto;
+        max-width: 100%;
+        height: auto;
+        max-height: 100%;
+        filter: drop-shadow(0px 0px 1px grey);
```
thornbill, line 407: The units need dropped from the `0` values here to make the linter happy. They can just be `0` instead of `0px`.

thornbill (changes requested):

thornbill:
Should be fixed by https://github.com/jellyfin/jellyfin-web/pull/3018

Issue #825, opened 2020-02-18
[Firefox] Contextual menu truncated


simonbcn:
- Arch Linux
- Firefox 73.0
- Jellyfin 10.5.0 (_installed from jellyfin-git AUR package_)

When I click the  `...` on an item in my library that is near the end of the page the context menu exits the page and there are options that remain hidden and inaccessible.

![Selección_00115](https://user-images.githubusercontent.com/428231/74731546-7a3a8580-5248-11ea-8abf-bd85a53ce31d.jpg)

dkanada:
This menu should be much smaller for mobile layouts, which I think is the real bug.

JustAMan:
Why do you think it's mobile layout?..

dkanada:
Ah, it definitely isn't a mobile device.

MrTimscampi:
This still happens and, after investigating the way context menus work in #1163, it's due to the terrible way position is calculated for these.

I've tried handling it better as part of that PR, but I it's not done yet.

tylerszabo:
It looks like there's a race condition in the `getPosition` helper for `actionSheet`. If you set a breakpoint to effect a delay right before `dlg.offsetHeight` is read the menu will display properly. If you set the breakpoint immediately after reading `dlg.offsetHeight` but before `dlg.offsetWidth` then it will only be cut off vertically (since the with had time to populate and the height did not).

https://github.com/jellyfin/jellyfin-web/blob/5c09077a2f12ae44e7b7d356ea8d9f35862b0b60/src/components/actionSheet/actionSheet.js#L45-L46

At the time the height and width are read but they're smaller; I'm wondering if this might be due to effects of animations?

tylerszabo:
This seems to have resolved itself (at least on 96.0.3 (64-bit) on Windows 10); perhaps Firefox fixed a bug.

theevilapplepie:
@simonbcn - Can you confirm if this is resolved for you with newer firefox versions?

PR #2685, opened 2021-05-29
Compute menu positions using nominal size #825

tylerszabo:
**Changes**

Animation of the menu that occurred during position calculation for initial placement caused size of the menu and thus the position to be incorrectly calculated on Firefox.

By stashing the nominal size of the dialog before starting the animation the correct size can be used to compute the location during positioning.

**Issues**

#825

### DIFF
--- a/src/components/actionSheet/actionSheet.js
+++ b/src/components/actionSheet/actionSheet.js
@@ -42,8 +42,8 @@ function getPosition(options, dlg) {
 
     pos.left += (pos.width || 0) / 2;
 
-    const height = dlg.offsetHeight || 300;
-    const width = dlg.offsetWidth || 160;
+    const height = dlg.nominalOffsetHeight || 300;
+    const width = dlg.nominalOffsetWidth || 160;
 
     // Account for popup size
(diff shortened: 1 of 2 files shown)


tylerszabo:
It seems the checks have stalled; I'm not sure how to rerun them.

src/components/actionSheet/actionSheet.js
```diff
@@ -42,8 +42,8 @@ function getPosition(options, dlg) {
 
     pos.left += (pos.width || 0) / 2;
 
-    const height = dlg.offsetHeight || 300;
-    const width = dlg.offsetWidth || 160;
+    const height = dlg.nominalOffsetHeight || 300;
```
thornbill, line 45: We may want to include fallbacks here for the normal `offsetHeight` and `offsetWidth`.

```js
dlg.nominalOffsetHeight || dlg.offsetHeight || 300;
```

thornbill (approved):

thornbill (changes requested):
Also, it looks like this will need rebased for the expected CI tasks to run.

PR #2697, opened 2021-06-02
Fix #2294 [bug] more than 301 photos in directory => wrong fullscreen…

### LINKED ISSUE #2294: [bug] more than 301 photos in directory => wrong fullscreen photo
**Describe the bug**
Whenever a directory contains more than 301 photos I have this error: the webpage with these photos (thumbnails) shows all photos as expected but if I click on a photo to get a fullscreen picture this only works for the first 301 photos. All other photos will show the oldest photo of this folder in fullscreen and not the actual one.

 - OS: OpenSuse Tumbleweed
 - Virtualization:  none
 - Clients: browser
 - Browser: any
 - Jellyfin Version: 10.7.0 rc1 / 10.6.4
 - Playback:  -
 - Installed Plugins: none
 - Reverse Proxy: apache
 - Base URL: none
 - Networking:  -
 - Storage: local

**To Reproduce**
1. rebuild photo library in directory with more than 301 photos.
2. click on thumbnail to view photo

**Expected behavior**
fullscreen photo is same as thumbnail photo

**Additional context**
If I delete the photo library and add it again to completely rebuild all informations the same problem shows up again.

WindLi001:
… photo

Allow browsing the 301st and after items.

**Changes**

**Issues**


MrTimscampi:
Do note that this will have side effects on all other media types.

Most notably, people trying to shuffle large music libraries will have completely broken (or very slow) playback with this limit removed.

This was touched upon in the Vue client recently:
https://github.com/jellyfin/jellyfin-vue/pull/1008
https://github.com/jellyfin/jellyfin-vue/discussions/1023

So simply removing said limit is not a good way to handle it, I think.

Do also note that the issue will present itself for large photo albums as well. If you have a few thousand photos in an album, you will have to wait a while for playback to start, which isn't a good experience either.

WindLi001:
Thanks for your comments, MrTimscampi. My English is not so well, but I will try my best to explain my opinion. I still think deleting the limit in code is the best choice.

I found this issue when browsing photo but not playing music. I already test the fix on my own server with folder including about 2000 photos, and the photo appears quickly just like with the limit. My server is not powerful. It's a Windows server  with J1900 CPU and 8GB RAM, and the photos are stored on a HDD.

Let's discuss all the possibilities.

There are 2 choices:
1.  Limit the number of medias in one folder to, for example, 300. Just as usual.
2.  No limit in code. As the fix.

There are 3 kinds of folders:
1.  Including a few of medias under the limit of code, for example, 200.
2.  Including a little more medias over the limit of code, but under the performance of server, for example, 2000.
3.  Including a lot of medias over the limit of code and the performace of server, for example, 10000.

The result of all the situations will be:
--------------------------------------------------------------------------------------------------------------------------
                    |      Folder1      |       Folder2                                          |      Folder3    
--------------------------------------------------------------------------------------------------------------------------
Choice1       |       OK            |    Only 300 medias can be accessed    |  Only 300 medias can be accessed
--------------------------------------------------------------------------------------------------------------------------
Choice2       |       OK            |    OK                                                     |  All the medias can be accessed but slowly
--------------------------------------------------------------------------------------------------------------------------

What is need to be discussed is Folder3. Which choice is better under this situation? I think Choice2 is better. Firstly, slowly is better than none. Secondly, more time for more workload, which is easy to be understanded by the user. Thirdly, both choice needs the user spliting the folder to subfolders with less medias, but the number of subfolders in Choice2 is less than Choice1, which means less work too. 

So as above, I think deleting the limit in code is the best choice.

WindLi001:
> Do note that this will have side effects on all other media types.
> 
> Most notably, people trying to shuffle large music libraries will have completely broken (or very slow) playback with this limit removed.
> 
> This was touched upon in the Vue client recently:
> [jellyfin/jellyfin-vue#1008](https://github.com/jellyfin/jellyfin-vue/pull/1008)
> [jellyfin/jellyfin-vue#1023](https://github.com/jellyfin/jellyfin-vue/discussions/1023)
> 
> So simply removing said limit is not a good way to handle it, I think.
> 
> Do also note that the issue will present itself for large photo albums as well. If you have a few thousand photos in an album, you will have to wait a while for playback to start, which isn't a good experience either.



> Do note that this will have side effects on all other media types.
> 
> Most notably, people trying to shuffle large music libraries will have completely broken (or very slow) playback with this limit removed.
> 
> This was touched upon in the Vue client recently:
> [jellyfin/jellyfin-vue#1008](https://github.com/jellyfin/jellyfin-vue/pull/1008)
> [jellyfin/jellyfin-vue#1023](https://github.com/jellyfin/jellyfin-vue/discussions/1023)
> 
> So simply removing said limit is not a good way to handle it, I think.
> 
> Do also note that the issue will present itself for large photo albums as well. If you have a few thousand photos in an album, you will have to wait a while for playback to start, which isn't a good experience either.

Thanks for your comments, MrTimscampi. My English is not so well, but I will try my best to explain my opinion. I still think deleting the limit in code is the best choice.

I found this issue when browsing photo but not playing music. I already test the fix on my own server with folder including about 2000 photos, and the photo appears quickly just like with the limit. My server is not powerful. It's a Windows server  with J1900 CPU and 8GB RAM, and the photos are stored on a HDD.

Let's discuss all the possibilities.

There are 2 choices:
1.  Limit the number of medias in one folder to, for example, 300. Just as usual.
2.  No limit in code. As the fix.

There are 3 kinds of folders:
1.  Including a few of medias under the limit of code, for example, 200.
2.  Including a little more medias over the limit of code, but under the performance of server, for example, 2000.
3.  Including a lot of medias over the limit of code and the performace of server, for example, 10000.

The result of all the situations will be:
--------------------------------------------------------------------------------------------------------------------------
                    |      Folder1      |       Folder2                                          |      Folder3    
--------------------------------------------------------------------------------------------------------------------------
Choice1       |       OK            |    Only 300 medias can be accessed    |  Only 300 medias can be accessed
--------------------------------------------------------------------------------------------------------------------------
Choice2       |       OK            |    OK                                                     |  All the medias can be accessed but slowly
--------------------------------------------------------------------------------------------------------------------------

What is need to be discussed is Folder3. Which choice is better under this situation? I think Choice2 is better. Firstly, slowly is better than none. Secondly, more time for more workload, which is easy to be understanded by the user. Thirdly, both choice needs the user spliting the folder to subfolders with less medias, but the number of subfolders in Choice2 is less than Choice1, which means less work too. 

So as above, I think deleting the limit in code is the best choice.

WindLi001:
Sorry I don't know how to keep the table in my comment :-(

MrTimscampi:
The main issue is that it takes an exponentially longer time the more items you have in a folder or in a request.

Here are some numbers I benchmarked a while ago when working on this for the Vue client:

```
Translating items for playback (1 request): 2749ms
Translating items for playback (1 request, limit 300): 428ms
```

This is for a music library with 319 albums, so not too big, but not too small either.

What this number means is that Jellyfin takes about 2.7 seconds to get all the tracks for playback (In the database, music tracks are represented in the exact same way and in the same tables as photos, hence why this comparison holds).

So, with everything else, that's almost 3 seconds to start playback, which is an eternity in terms of user interactions. And the numbers only get worse, the more items you are trying to play at one time.

Hence my concern: while this does indeed fix the issue of not allowing you to see items past 300 (most visible in photo libraries), it leads to a noticeably worse experience for most users in some (pretty common) cases (Shuffling to your music library being the most common example).

WindLi001:
I think we both agree that the diffrent results of the 2 choices when too many items under a folder (or request) are:

Choice 1: with limit as usual,
Result 1: only 300 items can be accessed.

Choice 2: without limit as fix,
Result 2: waiting a few seconds ( according to the server's perfomance and the number of items ), and then all the items can be accessed.

If we can find a choice 3 or more, which have the result _all the items can be accessed without wating_ , that is perfect. But now I cannot implement it. I have to admit that I am a newbie of not only Jellyfin code but also Node.JS. Maybe I have some thought, but I cannot transfer it to codes.

There are some choices I think maybe works:
Choice 3: Spliting a folder into several pages, and each page including fixed number of, for example, 100 items. When openning a folder, only 100 items of the first page are showen, so there is no waiting. Then the user can navigate to the other page to show 100 items on it, and there is no waiting too.

Choice 4: Spliting a folder into pages and loading the pages in background. When openning a folder, only 100 items of the first page are showen immediatly, so there is no waiting. And then the other pages of items are loaded automatically in backgroud, so the waiting won't be awared by the user.

I think choice 3 is enough for the case of photo for me, but the choice 4 is necessary for the case of shuffering music you mentioned. And both ones are beyond my coding skill of Node.JS now.

And I have to say, if there is only choice 1 and choice 2, the choice 2 is better. And it is simple to be coded by me. :-)

DinoHS:
@MrTimscampi - my understanding of the implementation details of JF is very naive, but just thinking about you said here it seems strange to lump photos and music together.  Photo browsing doesn't have the same kind of technical/performance considerations as loading audio files, and so some of the benchmarks you talk about don't necessarily have to apply.  This is an apples vs. oranges kind of thing.

Might it be time to separate photos from music in the data model?  (again, naive user suggestion)

MrTimscampi:
> Photo browsing doesn't have the same kind of technical/performance considerations as loading audio files

The bottleneck is not on the file loading part, it's on the database query part.

The more things you request from the database, currently, the slower it is. 300 items is around the 400ms mark, a few thousands easily gets you to multiple seconds for a query.

You are right in that Photos work differently in jf-web than, say, videos and music, as far as browsing the libraries go. They don't respect the pagination settings and basically simply list everything all the time.

However, that's not really the issue here.

The way Jellyfin Web does playback is as follows:

* Click on "Play" somewhere
* The item you are trying to play is sent to the playback manager
* It goes through "Translation", ie making a non-playable item actually playable (Through this function: https://github.com/jellyfin/jellyfin-web/blob/master/src/components/playback/playbackmanager.js#L1753-L1877)
* The resulting items are then added to the queue

Every media type follow the same process, hence me equating Photos to Music. Yes, loading a photo is different from loading an audio file, but at the moment, in practice, Jellyfin handles everything the same.

The new database schema should make DB queries a LOT faster than the current garbage (Even 400ms for 300 items is incredibly long) and once that is done, work on the new API will start, which will then likely handle all this item translation and playback stuff on the server, as part of the client session directly.

PR #2702, opened 2021-06-08
Update Name of search section video to video/Live Tv Channels

candry7731:
Updated Videos section to say, Videos/Live TV Channels

Videos section in the search results is pulling in live tv channels. The name of the section needs to be updated to reflect the content in the section. 

**Changes**
Update Name of search section video to video/Live Tv Channels

**Issues**
No issues, just renaming the section

### DIFF
--- a/src/components/search/searchresults.template.html
+++ b/src/components/search/searchresults.template.html
@@ -65,7 +65,7 @@
 </div>
 
 <div class="hide verticalSection videoResults">
-    <h2 class="sectionTitle sectionTitle-cards focuscontainer-x padded-left padded-right">${Videos}</h2>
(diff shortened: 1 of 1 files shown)


thornbill:
Thanks for the PR! I do have some feedback though.

- Currently I have a PR open that rewrites the search page in React, so this change would conflict with those changes (https://github.com/jellyfin/jellyfin-web/pull/2683).
- The value that you changed there is actually a translation key and not the text that gets displayed to users. Although it does look like it's not using the correct key currently. https://github.com/jellyfin/jellyfin-web/blob/master/src/strings/en-us.json#L471
- I think the best solution for this would be to have a separate row dedicated to channels as I think this row is more intended for home videos or other videos that aren't explicitly movies or TV shows. The fact that channels are showing up in this row seems like an unintended side effect since the row is hidden when searching in a Live TV library.

PR #2722, opened 2021-06-11
Add TypeScript support for React components

MrTimscampi:
**Changes**

Adds TypeScript support and converts the few React components to TSX.

Compared to #2404, it doesn't add the typed API client, but defines `window.ApiClient` as any.

**Issues**
Supercedes #2404


thornbill (changes requested):
I'm new to TypeScript, but it looks like the type definitions you added for props should completely replace the `PropTypes` used previously.

MrTimscampi:
> I'm new to TypeScript, but it looks like the type definitions you added for props should completely replace the `PropTypes` used previously.

From what I looked at, PropTypes is at runtime, while TypeScript is at build time. They serve different purposes.

Of course, we could remove PropTypes, but then, we'd lose runtime validation for prop types. Do we want that?

thornbill:
In the docs it sounds like it's not needed if you use Typescript or Flow. https://reactjs.org/docs/typechecking-with-proptypes.html

MrTimscampi:
Fair enough :)

I removed the `prop-types` package and adjusted things a bit to please the linter in regards to prop validation.

src/components/pages/SearchPage.tsx
```diff
@@ -1,31 +1,36 @@
...
+import React, { FunctionComponent, useState } from 'react';
 
 import SearchFields from '../search/SearchFields';
 import SearchResults from '../search/SearchResults';
 import SearchSuggestions from '../search/SearchSuggestions';
 import LiveTVSearchResults from '../search/LiveTVSearchResults';
 
-const SearchPage = ({ serverId, parentId, collectionType }) => {
+type SearchProps = {
+    serverId: string,
```
thornbill, line 9: Should we use `?:` here since these are optional?

thornbill (approved):

PR #2740, opened 2021-06-14
Replace newlines in plugin changelogs with the proper HTML tag

MrTimscampi:
**Changes**

**Issues**


nielsvanvelzen (approved):
Although this fixes new lines, this change still allows for XSS injections in plugin repositories (like third party ones).

MrTimscampi:
> Although this fixes new lines, this change still allows for XSS injections in plugin repositories (like third party ones).

Yep, unfortunately :/

The proper way to fix that would be to move all the code to use proper element creation instead of innerHTML. Since we're starting to move to React on this client, though, it may be a lot of line changes to fix all the XSS issues, which will all get killed by the React move afterwards anyway.

PR #2746, opened 2021-06-18
Remove jQuery from controllers/users

grafixeyehero:


grafixeyehero:
Closing because of #2954

PR #2753, opened 2021-06-23
Add setting for max days for a show to appear in next up list

jackrvan:
**Changes**
This pull request in the base jellyfin repo added the ability to set a limit for the amount of time a show can appear in the next up list https://github.com/jellyfin/jellyfin/pull/5828. This pull request is an attempt to add a setting for this.


thornbill (changes requested):
This does not seem to work in my testing. I'm not sure the server change is working. I am fairly certain that the change here is not sending the date in a valid format, but even playing with a few different formats for the date, I do not see any difference in the response from the server.

jackrvan:
It seems to be working for me. Here are some screenshots. I have watched Malcolm in the Middle today and havent watched Futurama for 3 days.
With the setting at 4 days:
image
With the setting at 2 days:
image

As you can see from the screenshot the first one contains both shows because both have been watched within the last 4 days. On the second screenshot only Malcolm in the  middle is visible because Futurama has not been watched in the last 2 days.  I am running the jellyfin server on the latest master with "dotnet run --project Jellyfin.Server --nowebclient" and im running jellyfin web on my version of master with "npm start"

Im not sure why it would not be working for you? Are you sure you were running the server on a version that has #5828 in it? If there is any more information I can provide please let me know but Im not sure how to proceed right now since it appears to work for me

src/scripts/settings/userSettings.js
```diff
@@ -429,6 +429,25 @@ export class UserSettings {
...
+        if (val !== undefined) {
+            return this.set('maxDaysForNextUp', parseInt(val, 10), false);
+        }
+
+        const maxDaysForNextUp = parseInt(this.get('maxDaysForNextUp', false), 10);
+        if (maxDaysForNextUp === 0) {
+            // Explicitly return 0 to avoid returning 100 because 0 is falsy.
+            return 0;
+        } else {
+            return maxDaysForNextUp || 100;
```
thornbill, line 447: I think 100 may be too low of a default for a field that was unlimited previously. Especially if you consider that the fallback behavior is to display the first episode of every show. This was the issue I was having in my testing. Perhaps showing one year by default would be better. :man_shrugging:

thornbill (changes requested):

src/scripts/settings/userSettings.js
jackrvan: Ya I think a year is better so changed.

src/components/homesections/homesections.js
```diff
@@ -595,17 +595,20 @@ import ServerConnections from '../ServerConnections';
...
             return apiClient.getNextUpEpisodes({
                 Limit: enableScrollX() ? 24 : 15,
                 Fields: 'PrimaryImageAspectRatio,DateCreated,BasicSyncInfo,Path',
                 UserId: apiClient.getCurrentUserId(),
                 ImageTypeLimit: 1,
                 EnableImageTypes: 'Primary,Backdrop,Banner,Thumb',
                 EnableTotalRecordCount: false,
-                DisableFirstEpisode: true
+                DisableFirstEpisode: false,
+                NextUpDateCutoff: oldestDateForNextUp.toUTCString()
```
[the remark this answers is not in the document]
jackrvan: Thanks. Im new to javascript so recommendations like this are very helpful

thornbill (approved):

PR #2754, opened 2021-06-24
Removed all legacy Emby CSS

Ben-Wallner:
**Changes**

I removed all legacy Emby CSS naming to clean up the code

**Issues**

none


cvium:
What is the benefit to this exactly? It's a humongous PR.

oddstr13 (changes requested):
This will cause a massive amount of work in *all* plugins that utilizes the web pages.

There are about 300 lines with `emby-` in them in the currently active official plugins alone, and this is very likely to also affect *all* unofficial plugins.

MrTimscampi:
Marking for discussion, as this is a potentially contentious change.

Not only does it break all the plugins (First and third party) but it can also very easily lead to issues in the client, requiring ideally an entire cycle to test and validate.

thornbill:
Closing based on points above.

PR #2776, opened 2021-07-03
Add hardware encoding status to playback data

gnuyent:
Hi, this is my first contribution to jellyfin. Let me know if I need to change/fix anything!

**Changes**
Inside Playback Info, add hardware encoding information.
image
image

**Issues**

Resolves https://github.com/jellyfin/jellyfin/issues/6087

Connected PR: https://github.com/jellyfin/jellyfin/pull/6258

### DIFF
--- a/src/components/playerstats/playerstats.js
+++ b/src/components/playerstats/playerstats.js
@@ -173,6 +173,10 @@ import ServerConnections from '../ServerConnections';
                     value: session.TranscodingInfo.TranscodeReasons.map(translateReason).join('<br/>')
                 });
             }
+            sessionStats.push({
+                label: globalize.translate('LabelHardwareEncoding'),
+                value: session.TranscodingInfo.IsHardwareEncode
+            });
         }
 
         return sessionStats;
--- a/src/components/playerstats/playerstats.js
+++ b/src/components/playerstats/playerstats.js
@@ -175,7 +175,7 @@ import ServerConnections from '../ServerConnections';
             }
             sessionStats.push({
                 label: globalize.translate('LabelHardwareEncoding'),
-                value: session.TranscodingInfo.IsHardwareEncode
+                value: session.TranscodingInfo.HardwareAccelerationType
             });
         }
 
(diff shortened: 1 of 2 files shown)


YouKnowBlom:
Hello, thank you for contributing!:) From my testing it seems like `HardwareAccelerationType` is not always present in `TranscodingInfo`. Your first screenshot has the value "none" but I'm not quite sure where that is set?

src/components/playerstats/playerstats.js
```diff
@@ -173,6 +173,10 @@ import ServerConnections from '../ServerConnections';
                     value: session.TranscodingInfo.TranscodeReasons.map(translateReason).join('<br/>')
                 });
             }
+            sessionStats.push({
```
YouKnowBlom, line 176: I think it would be a good idea to wrap this in an if like the stats above, just in case.

YouKnowBlom: bit of a nitpick but visually I think it will nicer if it's above the transcoding reasons.

gnuyent: I'm not sure if that would work out since this variable will always be set to something and always display. Not too sure if this is desired behavior though, will wait on more feedback!

YouKnowBlom: Fair point! It should be rare enough with mismatching web and server versions for this to be an issue.

cvium: Everything else is wrapped in `if`. We should follow the "standard".

src/components/playerstats/playerstats.js
```diff
@@ -173,6 +173,10 @@ import ServerConnections from '../ServerConnections';
...
+                label: globalize.translate('LabelHardwareEncoding'),
+                value: session.TranscodingInfo.HardwareAccelerationType
+            });
```
cvium, line 179:
```suggestion
            if (session.TranscodingInfo.HardwareAccelerationType) {
                sessionStats.push({
                    label: globalize.translate('LabelHardwareEncoding'),
                    value: session.TranscodingInfo.HardwareAccelerationType
                });
            }
```

src/components/playerstats/playerstats.js
gnuyent: Sorry for the delay, just accepted the change!

cvium (approved):

Issue #2756, opened 2021-06-28
LG webOS TV browser: top bar doesn't hide completely


n1trux:
since webOS 3.23.05 the top bar of the LG browser doesn't hide the bar completely, it just slides away and there's still a 20px or so wide stripe on the top of the screen (see attached screenshot).

Maybe if Jellyfin went into fullscreen mode when playing media this issue could go away. There's no way to go into full screen mode within the browser itself, as far as I can see.

**Steps To Reproduce**
1. open LG browser
2. go to Jellyfin
3. start playing any media
4. bar slides up but is still visible

**Expected Behavior**
The bar should disappear because Jellyfin goes into full screen mode when it detects a TV browser.

**Screenshots**
Bar shown:
![bar_shown](https://user-images.githubusercontent.com/7255542/123626160-ccae8d80-d810-11eb-9b84-714c6bcd281b.jpg)
Bar "hidden" (more like: slid up):
![bar_hidden](https://user-images.githubusercontent.com/7255542/123626157-cc15f700-d810-11eb-96d6-924cc66edf94.jpg)

**System (please complete the following information):**
 - Platform: LG 55NANO913NA TV
 - Browser: LG browser (version not visible anywhere, OS version 3.23.05)
 - Jellyfin Version: 10.7.6

hundehausen:
It goes away after some time on my LG OLED.

FluffyDiscord:
Once it goes away, in about 5 mins, subtitles go away too.

hundehausen:
And at some point the screen saver kicks in.

imchasingshadows:
The new WebOS browser appears to have dropped the `Web0S` part from the useragent causing jellyfin to not correctly detect the new browser. In addition to this jellyfin hides the fullscreen toggle when the browser is detected to be a tv

the useragent im seeing in my access logs;
`Mozilla/5.0 (Linux; NetCast; U) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/79.0.3945.79 Safari/537.36 SmartTV/10.0 Colt/2.0`

I was able to resolve the issue locally by adding a new variable to the browser object
```js
browser.js:237

browser.web0s = userAgent.toLowerCase().indexOf('Web0S'.toLowerCase()) !== -1;
// im not sure which part of the useragent is unique but this seemed to work
browser.web0sChromium = userAgent.toLowerCase().indexOf('SmartTV'.toLowerCase()) !== -1;
```

[...]

PR #2781, opened 2021-07-03
bugfix/web0s-fullscreen

imchasingshadows:
Fix user agent detection for new webos browser
Fix greybar on top of the screen in new webos browser by entering fullscreen

Tested on my lg E9 with the recently released browser and on Firefox.

It appears that lg changed the behavior of the back (461) button as it no longer fires a keydown event. I would've preferred to call `approuter.back()` during this event, but as a workaround i call it during the fullscreenchanged event.

Also first pr :)

**Changes**
1. Added netcast as a way to identify the new webos browser
2. Automatically enter fullscreen when layout is tv
3. Stop playback when video exits fullscreen (tv layout)

**Issues**
#2756


whiteworthy:
compiled this and it works perfectly if anyone is wondering

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1369,12 +1369,17 @@ function tryRemoveElement(elem) {
...
 
-                        // don't animate on smart tv's, too slow
-                        if (options.fullscreen && browser.supportsCssAnimation() && !browser.slow) {
-                            return zoomIn(dlg).then(function () {
-                                return videoElement;
-                            });
-                        } else {
+                        if (options.fullscreen) {
+                            if (browser.tv && Screenfull.isEnabled) {
+                                Screenfull.request();
```
thornbill, line 1374: Should there be a return here or maybe the next condition should be an `else if`? Or is it intended for TVs to continue through the rest of the following logic?

src/scripts/browser.js
```diff
@@ -14,6 +14,23 @@ function isTv() {
...
 
+    if (userAgent.indexOf('netcast') !== -1) {
+        return true;
+    }
+
+    if (userAgent.indexOf('web0s') !== -1) {
+        return true;
+    }
+
+    return false;
```
thornbill, line 25: The lines above here could just be replaced by `return isWebOs();` probably.

thornbill (changes requested):
:+1: I just had a couple questions and this needs to be rebased.

src/plugins/htmlVideoPlayer/plugin.js
imchasingshadows: Yes, if im correct tv's have `browser.slow` set to true anyway but its more clear to just immediately return

src/controllers/playback/video/index.js
```diff
@@ -402,6 +402,14 @@ import { appRouter } from '../../../components/appRouter';
             }
         }
 
+        function onFullscreenChanged() {
+            if (!layoutManager.tv) {
+                updateFullscreenIcon();
+            } else if (!playbackManager.isFullscreen(currentPlayer)) {
+                appRouter.back();
+            }
```
dmitrylyzo, line 410: This breaks playback on webOS 1.2 (emulator) because there is no `webkitIsFullScreen` property: https://github.com/jellyfin/jellyfin-web/blob/f5ea7bc83081d7f3b5a38b00cbbe3b19a101316f/src/components/playback/playbackmanager.js#L1452

dmitrylyzo (commented):
> It appears that lg changed the behavior of the back (461) button as it no longer fires a keydown event.

On Tizen, `Back` button doesn't fire `keydown` because it is handled by browser. Probably the same with webOS.

dmitrylyzo (commented):
I tried this to prevent a return if not forced fullscreen mode:
```js
Screenfull.request().then(() => {
    this.isFullscreenHack = true;
})
```
_reset it somewhere?_

then (in `video/index.js`)
```js
if (currentPlayer.isFullscreenHack && !playbackManager.isFullscreen(currentPlayer)) {
    appRouter.back();
    return;
}
```
_hack over hack_ :smile:

dkanada:
@imchasingshadows fair warning, we're moving towards a larger release just in case you want to get this in before that happens.

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1366,12 +1366,18 @@ function tryRemoveElement(elem) {
                             document.body.classList.add('hide-scroll');
                         }
 
-                        // don't animate on smart tv's, too slow
-                        if (options.fullscreen && browser.supportsCssAnimation() && !browser.slow) {
-                            return zoomIn(dlg).then(function () {
+                        if (options.fullscreen) {
+                            if (browser.tv && Screenfull.isEnabled) {
```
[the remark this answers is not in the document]
imchasingshadows: I'm a little confused does this invalidate the need for the `isFullcreenHack` you've proposed earlier?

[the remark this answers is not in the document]
dmitrylyzo: No, it's just to avoid running a hack where it's definitely not needed (Tizen and apps).
`isFullcreenHack` is still required.

dmitrylyzo:
@imchasingshadows Try this patch (_it is hard to do it in suggestions_)
```diff
diff --git a/src/controllers/playback/video/index.js b/src/controllers/playback/video/index.js
index c82ee4fb1..fa8603f30 100644
--- a/src/controllers/playback/video/index.js
+++ b/src/controllers/playback/video/index.js
@@ -403,7 +403,7 @@ import { appRouter } from '../../../components/appRouter';
         }
 
         function onFullscreenChanged() {
-            if (!playbackManager.isFullscreen(currentPlayer)) {
+            if (currentPlayer.forcedFullscreen && !playbackManager.isFullscreen(currentPlayer)) {
                 appRouter.back();
                 return;
             }
diff --git a/src/plugins/htmlVideoPlayer/plugin.js b/src/plugins/htmlVideoPlayer/plugin.js
index 0f1e48a8b..9161ffb56 100644
--- a/src/plugins/htmlVideoPlayer/plugin.js
+++ b/src/plugins/htmlVideoPlayer/plugin.js
@@ -1361,23 +1361,26 @@ function tryRemoveElement(elem) {
                         this.#videoDialog = dlg;
                         this.#mediaElement = videoElement;
 
+                        delete this.forcedFullscreen;
+
                         if (options.fullscreen) {
                             // At this point, we must hide the scrollbar placeholder, so it's not being displayed while the item is being loaded
                             document.body.classList.add('hide-scroll');
-                        }
 
-                        if (options.fullscreen) {
+                            // Enter fullscreen in the webOS browser to hide the top bar
                             if (!window.NativeShell && browser.web0s && Screenfull.isEnabled) {
-                                Screenfull.request();
+                                Screenfull.request().then(() => {
+                                    this.forcedFullscreen = true;
+                                });
                                 return videoElement;
                             }
+
                             // don't animate on smart tv's, too slow
                             if (!browser.slow && browser.supportsCssAnimation()) {
                                 return zoomIn(dlg).then(function () {
                                     return videoElement;
                                 });
                             }
-
                         }
 
                         return videoElement;
```

imchasingshadows:
@dmitrylyzo i've applied the suggested patch you made. Would you be able to test it?

dmitrylyzo:
> @dmitrylyzo i've applied the suggested patch you made.

I suppose you did it manually.
[This block](https://github.com/jellyfin/jellyfin-web/blob/3e749d20a868adb3754ebdd3e27034fdbea2665a/src/plugins/htmlVideoPlayer/plugin.js#L1370-L1384) should be in [if-block](https://github.com/jellyfin/jellyfin-web/blob/3e749d20a868adb3754ebdd3e27034fdbea2665a/src/plugins/htmlVideoPlayer/plugin.js#L1366-L1369).
_You can amend the previous commit and force-push._

>  Would you be able to test it?

_Speaking of the original patch._
I tested it in Tizen/webOS emulators (browser, app) and in Chrome on PC - it doesn't go back unless forced fullscreen has been activated (_as intended_).
In webOS 1 and 3, browsers don't go fullscreen. I don't know if this is some kind of policy, but that's why there is a promise handler - to capture the real entering fullscreen.
It would be better to test this PR on real webOS. I believe it should work if it worked before, because we haven't changed the main logic :crossed_fingers:

dmitrylyzo (changes requested):
@imchasingshadows You applied the patch incorrectly. Could you please fix it?
[This block](https://github.com/jellyfin/jellyfin-web/blob/3e749d20a868adb3754ebdd3e27034fdbea2665a/src/plugins/htmlVideoPlayer/plugin.js#L1370-L1384) should be in [if-block](https://github.com/jellyfin/jellyfin-web/blob/3e749d20a868adb3754ebdd3e27034fdbea2665a/src/plugins/htmlVideoPlayer/plugin.js#L1366-L1369).
_You can amend the previous commit and force-push._

imchasingshadows:
@dmitrylyzo Im sorry im not that familiar with patches. I've amended the commit i hope its resolved now :) let me know if i've made it worse haha

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1361,19 +1361,30 @@ function tryRemoveElement(elem) {
...
+                            // don't animate on smart tv's, too slow
+                            if (!browser.slow && browser.supportsCssAnimation()) {
+                                return zoomIn(dlg).then(function () {
+                                    return videoElement;
+                                });
+                            }
                         }
+
+
+                        return videoElement;
```
dmitrylyzo, line 1387: extra empty line
```suggestion
                        }

                        return videoElement;
```

src/scripts/browser.js
```diff
@@ -14,6 +14,19 @@ function isTv() {
...
+    if (isWeb0s()) {
+        return true;
+    }
+
+    return false;
+}
+
+function isWeb0s() {
+    const userAgent = navigator.userAgent.toLowerCase();
+    if (userAgent.indexOf('netcast') !== -1) {
```
dmitrylyzo, line 26: just for a better look
```suggestion
    const userAgent = navigator.userAgent.toLowerCase();

    if (userAgent.indexOf('netcast') !== -1) {
```

dmitrylyzo:
> @dmitrylyzo Im sorry im not that familiar with patches. I've amended the commit i hope its resolved now :) let me know if i've made it worse haha

:+1: Almost fine.
I put 2 cosmetic suggestions. Use `Add suggestion to batch` to make a single commit.

dmitrylyzo:
Not sure if we need to do it in this PR. If the theme background is enabled, we also need to enter fullscreen.

Replace
https://github.com/jellyfin/jellyfin-web/blob/2e66ce13e6c0bcb20de5c1568f857d7bc8677715/src/plugins/htmlVideoPlayer/plugin.js#L1389-L1392
with
```js
                    delete this.forcedFullscreen;

                    if (options.fullscreen) {
                        // we need to hide scrollbar when starting playback from page with animated background
                        document.body.classList.add('hide-scroll');

                        // Enter fullscreen in the webOS browser to hide the top bar
                        if (!window.NativeShell && browser.web0s && Screenfull.isEnabled) {
                            Screenfull.request().then(() => {
                                this.forcedFullscreen = true;
                            });
                        }
                    }
```

_probably for the next PR_
I think we need to split `createMediaElement` promise to remove duplicates:
- create element
- enter fullscreen and hide scrollbar

dmitrylyzo:
> Not sure if we need to do it in this PR. If the theme background is enabled, we also need to enter fullscreen.

My suggestion might break the hack for sequential playback (playlist) because fullscreen is an API restricted to user gestures.
So maybe left for later.

dmitrylyzo:
I fixed suggestion in https://github.com/jellyfin/jellyfin-web/pull/2781#issuecomment-920979045
With that, it won't repeat the hack if it has already been.

dmitrylyzo:
@imchasingshadows not that. Below is a block that is used to re-use the video player.

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1361,24 +1361,41 @@ function tryRemoveElement(elem) {
...
                     // we need to hide scrollbar when starting playback from page with animated background
                     if (options.fullscreen) {
                         document.body.classList.add('hide-scroll');
+
+                        // Enter fullscreen in the webOS browser to hide the top bar
+                        if (!this.forcedFullscreen && !window.NativeShell && browser.web0s && Screenfull.isEnabled) {
+                            Screenfull.request().then(() => {
+                                this.forcedFullscreen = true;
+                            });
+                        }
```
dmitrylyzo, line 1398: move the comment to the right place
```suggestion
                    if (options.fullscreen) {
                        // we need to hide scrollbar when starting playback from page with animated background
                        document.body.classList.add('hide-scroll');

                        // Enter fullscreen in the webOS browser to hide the top bar
                        if (!this.forcedFullscreen && !window.NativeShell && browser.web0s && Screenfull.isEnabled) {
                            Screenfull.request().then(() => {
                                this.forcedFullscreen = true;
                            });
                        }
```

dmitrylyzo (approved):
It still needs to be tested on real webOS 6 to make sure that the `Screenfull.request` promise is resolving.

thornbill (approved):

PR #2786, opened 2021-07-10
Improve people pages and metadata editor

MrTimscampi:
**Changes**

Adds ages, fixes date format to be localized (according to browser locale, I didn't want to wrestle with jf-web's wonky locale system), fixes an untranslated string in the metadata editor and adapt the field names of the editor for people.

If the person is still alive, we show the age next to the birth date:
image
Otherwise next to their death date:
image

People now have "names" and not "titles". Original name has also been unhidden, because people might like to know the name of an actor in the original script:
image

**Issues**


src/components/metadataEditor/metadataEditor.js
```diff
@@ -638,6 +638,9 @@ import template from './metadataEditor.template.html';
 
         if (item.Type === 'Person') {
             //todo
```
thornbill, line 640: Todo what? :upside_down_face:

MrTimscampi: Not sure what Luke wanted to do :yum: 
We can probably remove it though, yeah.

thornbill (approved):

PR #2798, opened 2021-07-22
Add upload image via URL

MrTimscampi:
**Changes**

Adds a "Download from URL" button to the image uploader.

This will download the image client-side, then build a file blob out of it and set it as the selected file on the file input. The rest of the process is unchanged.

image

An important note is that some websites might return a CORS error (ThePosterDB comes to mind). This is handled using a try..catch + a check on `response.ok`.

**Issues**


YouKnowBlom (approved):

src/components/imageUploader/imageUploader.js
```diff
@@ -124,6 +125,40 @@ import template from './imageUploader.template.html';
...
+
+            try {
+                const response = await fetch(url);
+
+                if (response.ok) {
+                    const fileBlob = await response.blob();
+
+                    const extension = mime[response.headers.get('Content-Type') || 'image/jpeg'].extensions[0];
+
+                    const urlFileBlob = new File([fileBlob], `image.${extension}`, {
```
thornbill, line 150: Any idea if the File API will be supported by webOS/Tizen? It looks like older versions of Chrome may not https://caniuse.com/fileapi

thornbill (changes requested):

src/components/imageUploader/imageUploader.js
MrTimscampi: Good point.

Should we allow editing on TV at all, though? Seems like it'd be nicer to simply handle navigation/playback on TV and leave all the admin stuff out.

I haven't really tried to use the edit dialog in TV mode, but given how it looks, I doubt it's really usable :yum:

thornbill: I agree that a lot of stuff should probably just be disabled in TV mode, but if it's not this has the potential to crash the client.

thornbill:
I think it is a good feature, but this should support should be added server side for handling the image downloading.

kees-:
Is there any possibility of this feature being reintroduced?

Almost all images I manually upload are saved from the web and it adds a silly amount of friction to constantly save and reupload.

PR #2821, opened 2021-08-03
Update merge conflict action trigger

thornbill:
**Changes**
This is an attempt to fix the action failing on PRs. This config is used in jellyfin-androidtv and I have not seen these failures there.

**Issues**
N/A

### DIFF
--- a/.github/workflows/automation.yml
+++ b/.github/workflows/automation.yml
@@ -2,15 +2,19 @@ name: 'Automation'
 
 on:
   push:
+    branches:
+      - master
   pull_request_target:
+    types:
+      - synchronize
 
 jobs:
   triage:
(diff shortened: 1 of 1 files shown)


h1dden-da3m0n:
okay I think this wont have your desired effect, I much rather suspect the misplaced `if: ${{ github.repository == 'jellyfin/jellyfin-web' }}` to be the culprit of the errors.
Cause if you are in a fork and this CI triggers it would try to run the first job and fail caus there are no valid steps => fail.
moving the if to the job level rather than the step lvl should fix taht

h1dden-da3m0n (approved):

anthonylavado (approved):

Issue #2401, opened 2021-02-08
Volume slider scale not linear


axelsimon:
**Describe The Bug**
When dragging the volume slider control, the sound isn't changed linearly.
For instance, dropping down to a volume of 10 out of 100 is still quite loud, and even 1 is still far louder than "1% of the initial volume".
You can see this also when comparing the Jellyfin-web's slider with the PulseAudio volume in Pulse Audio Volume Control: when you reach 1 in Jellyfin-web, you are still at about 20% volume in PulseAudio.

**Steps To Reproduce**
1. Play something, a song for instance
2. Bring volume down
3. Volume is reduced, but not as much as expected

**Expected Behaviour**
Normal sound control, similar to other apps and PulseAudio in particular.

**Logs**
not relevant

**Screenshots**
Will try to share a screen capture.

**System (please complete the following information):**
 - Platform: Fedora 32
 - Browser: Firefox 85.0
 - Jellyfin Version: 10.6.4

**Additional Context**
None that I can think of.

axelsimon:
AFAICT, this issue is still valid.

MrTimscampi:
This is fixed in unstable (future 10.8)

PR #2837, opened 2021-08-07
Made volume slider exponential rather than linear

keegandahm:
This makes the web volume slider exponential rather than linear, making it easier to pick a good volume as we perceive loudness on a logarithmic scale. Please see [this UX stackoverflow answer](https://ux.stackexchange.com/a/116300), and the webpage it links to, for why this is desirable.

**Changes**
Updates both the htmlAudioPlayer and htmlVideoPlayer to set the mediaElement volume on an exponential scale.

**Issues**
This has an open [Issue #2401](https://github.com/jellyfin/jellyfin-web/issues/2401).

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -46,6 +46,7 @@
  - [danieladov](https://github.com/danieladov)
  - [Stephane Senart](https://github.com/ssenart)
  - [Ömer Erdinç Yağmurlu](https://github.com/omeryagmurlu)
+ - [Keegan Dahm](https://github.com/keegandahm)
 
 # Emby Contributors
 
--- a/src/plugins/htmlAudioPlayer/plugin.js
+++ b/src/plugins/htmlAudioPlayer/plugin.js
@@ -457,14 +457,14 @@ class HtmlAudioPlayer {
     setVolume(val) {
         const mediaElement = this._mediaElement;
         if (mediaElement) {
-            mediaElement.volume = Math.pow(val / 100, 2);
+            mediaElement.volume = Math.pow(val / 100, 3);
         }
     }
 
     getVolume() {
         const mediaElement = this._mediaElement;
         if (mediaElement) {
-            return Math.min(Math.round(Math.pow(mediaElement.volume, 0.5) * 100), 100);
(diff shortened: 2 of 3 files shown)


MrTimscampi:
The math doesn't seem quite right, here (from some quick graphing)

There's been some research on this we could take from, including formulas for an "ideal" curve: https://www.dr-lex.be/info-stuff/volumecontrols.html

keegandahm:
For a curve of x², the math is correct; It squares to get the multiplier the media player uses, and takes the square root (or, 0.5th power) to convert back to the UI volume. Like the stackoverflow answer mentions, having any exponential curve is a huge improvement over a linear one, and it's definitely worth trying the branch yourself to see the difference.

Would you prefer a different curve? Other media players seem to prefer x³. For reference, [VLC uses x³](https://github.com/videolan/vlc/blob/master/modules/audio_output/waveout.c#L907), and [mpv also uses x³](https://github.com/mpv-player/mpv/blob/master/player/audio.c#L169), with some extra tweaking after.

MrTimscampi:
> Would you prefer a different curve? Other media players seem to prefer x³. For reference, [VLC uses x³](https://github.com/videolan/vlc/blob/master/modules/audio_output/waveout.c#L907), and [mpv also uses x³](https://github.com/mpv-player/mpv/blob/master/player/audio.c#L169), with some extra tweaking after.

Going for x³ is probably a better UX choice then, since most users would be used to it.

If you can make that change, that's good for me :+1:

keegandahm:
Sounds good, I just changed the curve to x³

MrTimscampi (approved):

keegandahm:
Thank you for authorizing my PR! It looks like I still need a maintainer to approve running 2 more workflows, since this is my first contribution. Can you approve the remaining workflows?

keegandahm:
One last block, I'm not authorized to merge the PR, the base branch is restricted. I think you'll have to do that as well, sorry!

MrTimscampi:
> One last block, I'm not authorized to merge the PR, the base branch is restricted. I think you'll have to do that as well, sorry!

Someone in the core team needs to come around and merge it. They'll likely do that in the coming days.

PR #2857, opened 2021-08-10
Allow web player to buffer ahead a reasonable amount of data

### LINKED ISSUE #2856: Frequent playback stalls, apparently due to insufficient buffering
**Describe The Bug**

Previously, playing video files in jellyfin-web was consistently smooth. Now, I regularly see playback freeze for ~1-2 seconds and then resume. It happens perhaps every few minutes.

When it happens, the browser console shows:

```
HLS Error: Type: mediaError Details: bufferStalledError Fatal: false
```

Watching in the networking tab, it seems to involve fetch requests that stall for a second or two before succeeding. I assume my wifi is a bit flaky or something. Nonetheless, it should be possible to get smooth playback even if the underlying network occasionally freezes for a second or two.

 - Platform: Linux (both server and client)
 - Browser: Firefox
 - Jellyfin Version: 10.7.6

**Additional Context**

It looks like https://github.com/jellyfin/jellyfin-web/pull/2224 recently reduced the buffering on Firefox down to 6 seconds, rather than the 30 seconds it used to be set to. This might be the issue?

Watching the devtools developer tab, I do see jellyfin-web making HLS requests every ~6 seconds. They're quite tiny -- less than 10 megabytes apiece.

Looking at the history on #2224, it seems that the issue is that some browsers get cranky if you try to buffer too many megabytes of video data at once. So I'd expect that to be handled by controlling how much data gets buffered. Instead, though, that PR and #2148 set a limit on how many seconds of video can be buffered, which is a bit awkward and indirect.

Looking at [the hls.js docs](https://github.com/video-dev/hls.js/blob/master/docs/API.md#maxmaxbufferlength), it looks like their strategy is:

- by default, buffer `maxBufferSize` bytes of data (default 60 MB)
- but if you have a really high-bitrate video, where `maxBufferSize` bytes gives you less than `maxBufferLength` seconds, then bump the buffer up to handle `maxBufferLength` seconds (default 30 seconds)
- and if you have a really low-bitrate video, where `maxBufferSize` bytes would be like an hour or something silly like that, then reduce the buffer down to `maxMaxBufferLength` seconds of data

This makes a lot of sense. It seems like the problem #2224 and #2148 are fixing is that jellyfin used to set a high `maxBufferLength`, so for high-bitrate videos we were forcing hls.js to ignore `maxBufferSize` and buffer hundreds of megabytes, which caused problems.

However, those PRs didn't just change `maxBufferLength` -- they also changed `maxMaxBufferLength`, which means that they're forcing the playback to be chopped up into little 6 second segments at largest, even if the network and browser could happily handle much larger segments. This doesn't make sense to me. And it would explain why jellyfin-web seems to have become super-sensitive to minor networking glitches.

So I suspect the answer is:

- keep `maxBufferLength` small, so high bitrate videos don't annoy browsers with tons of data all at once
- but increase `maxMaxBufferLength` to something much more reasonable (the default is 600 seconds, which seems fine), so that for lower bitrate videos we mostly use `maxBufferSize` to control buffering

Unfortunately, I couldn't easily figure out how to try this to confirm it solves the problem. It feels like I ought to be something I could try temporarily by pasting a line or two of janky js into the browser console, but idk how to find any of the relevant objects.

njsmith:
The hls.js maxBufferLength parameter sets the minimum amount of video
that hls.js will try to buffer. We used to have this set to 30
seconds, but this caused problems with high-bitrate videos trying to
buffer too much data and making Chrome/Firefox cranky, so
maxBufferLength was reduced in #2148 and #2224.

However, those PRs *also* reduced maxMaxBufferLength to 6 seconds,
which doesn't make sense -- basically this is telling hls.js that it
must *never* buffer more than 6 seconds at a time, even if there's
plenty of memory and it wouldn't make the browser cranky at all. And
obviously such a tiny buffer makes it very easy for small network
glitches to cause user-visible playback problems.

So this PR removes the maxMaxBufferLength configuration entirely,
accepting hls.js's default (which is currently 600 seconds).

Fixes #2856

### DIFF
--- a/src/plugins/htmlVideoPlayer/plugin.js
+++ b/src/plugins/htmlVideoPlayer/plugin.js
@@ -385,7 +385,6 @@ function tryRemoveElement(elem) {
             return new Promise((resolve, reject) => {
                 requireHlsPlayer(async () => {
                     let maxBufferLength = 30;
-                    let maxMaxBufferLength = 600;
 
                     // Some browsers cannot handle huge fragments in high bitrate.
                     // This issue usually happens when using HWA encoders with a high bitrate setting.
@@ -393,7 +392,6 @@ function tryRemoveElement(elem) {
                     // https://github.com/video-dev/hls.js/issues/876
                     if ((browser.chrome || browser.edgeChromium || browser.firefox) && playbackManager.getMaxStreamingBitrate(this) >= 25000000) {
                         maxBufferLength = 6;
-                        maxMaxBufferLength = 6;
                     }
 
                     const includeCorsCredentials = await getIncludeCorsCredentials();
@@ -401,7 +399,6 @@ function tryRemoveElement(elem) {
                     const hls = new Hls({
                         manifestLoadingTimeOut: 20000,
                         maxBufferLength: maxBufferLength,
-                        maxMaxBufferLength: maxMaxBufferLength,
                         xhrSetup(xhr) {
                             xhr.withCredentials = includeCorsCredentials;
                         }


MrTimscampi (approved):

thornbill (approved):

gaalos:
Hi,
You merge it on jellyfin:master 
If if check the master the merge is not on.
I think i got issues because the maxBufferLength  is set to 30 inside master code

PR #2872, opened 2021-08-20
Send vp8 and vp9 over vpx as transcode profiles

GodTamIt:
**Changes**
This change sends `vp8` and `vp9` along with `vpx` as a transition to being more specific about the codecs.

**Issues**
Progress on jellyfin/jellyfin#6455

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -668,11 +668,37 @@ import browser from './browser';
         }
 
         if (canPlayVp8) {
+            // TODO: Remove vpx entry once servers are migrated
             profile.TranscodingProfiles.push({
                 Container: 'webm',
                 Type: 'Video',
-                AudioCodec: 'vorbis',
+                AudioCodec: webmAudioCodecs.includes('opus') ? 'opus' : 'vorbis',
                 VideoCodec: 'vpx',
+                Context: 'Streaming',
+                Protocol: 'http',
+                // If audio transcoding is needed, limit channels to number of physical audio channels
+                // Trying to transcode to 5 channels when there are only 2 speakers generally does not sound good
+                MaxAudioChannels: physicalAudioChannels.toString()
+            });
+            profile.TranscodingProfiles.push({
+                Container: 'webm',
+                Type: 'Video',
+                AudioCodec: webmAudioCodecs.includes('opus') ? 'opus' : 'vorbis',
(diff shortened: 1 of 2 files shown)


src/scripts/browserDeviceProfile.js
```diff
@@ -668,17 +668,43 @@ import browser from './browser';
         }
 
         if (canPlayVp8) {
+            // TODO: Remove vpx entry once servers are migrated
             profile.TranscodingProfiles.push({
                 Container: 'webm',
                 Type: 'Video',
-                AudioCodec: 'vorbis',
+                AudioCodec: webmAudioCodecs.includes('opus') ? 'opus' : 'vorbis',
                 VideoCodec: 'vpx',
```
thornbill, line 676: Maybe it would make sense to use `webmVideoCodecs` here and just append `vpx`. I'm not sure if av1 is currently supported when transcoding to webm though. It looks like it can be included in the mp4 transcoding profile now.

GodTamIt: - Just for clarity, we're trying to deprecate `vpx` here. Adding it to `webmVideoCodecs` is fine until jellyfin/jellyfin#6456 lands and is sufficiently migrated to.
- The `VideoCodec` field of a `TranscodingProfile` is expected to only have one codec as of now. So regardless, we'd have to add them all separately.
- av1 is transcoding is not fully supported by the server yet, though I might create a pull request for it. I don't imagine it will be heavily used since most servers won't be powerful enough to transcode in realtime.

thornbill: > The `VideoCodec` field of a `TranscodingProfile` is expected to only have one codec as of now. So regardless, we'd have to add them all separately.

Is this just a limitation of the code for the webm container, because every other video container contains multiple comma separated values for the `VideoCodec` field?

GodTamIt: You're right, I misremembered and had to double check on the server side. I've made the change now in a new commit.

dkanada:
@GodTamIt just for reference, should these be merged in a specific order?

GodTamIt:
> @GodTamIt just for reference, should these be merged in a specific order?

Technically, they can land in any order, since we do send `vpx` for backwards compatibility. However, I'll probably just wait till the server change lands first if there's no hurry.

dkanada (approved):

PR #2889, opened 2021-08-24
Fix dialog history

### LINKED ISSUE #2888: Can't change settings
Hello,
can you please check why I can't change the settings using my Samsung TV.
For example if I want to change the subtitles size or font I choose the option i like but when I hit "OK" on my remote the option is not saved. 
I've just installed the latest releases on my Samsung 65Q70R

Thank you.

dmitrylyzo:
It seems that #2660 is not enough.

<details>
<summary>There was a special handling in the customized `page.js`</summary>

https://github.com/jellyfin/jellyfin-web/blob/c25db80cbdf036807b898b6ce19d3b7455d77157/src/libraries/pagejs/page.js#L615-L627
</details>

<details>
<summary>and the router part (called by `dialogHelper`)</summary>

https://github.com/jellyfin/jellyfin-web/blob/c25db80cbdf036807b898b6ce19d3b7455d77157/src/components/appRouter.js#L667-L670
</details>

So `page.js` didn't reload the view in this case.

**Changes**
- Add a dummy route for dialogs.
- Don't reload the view when go to or return from a dummy route.
- Re-enable history for some dialogs.

**Issues**
Fixes #2888

### DIFF
--- a/src/components/appRouter.js
+++ b/src/components/appRouter.js
@@ -25,6 +25,7 @@ class AppRouter {
     msgTimeout;
     popstateOccurred = false;
     resolveOnNextShow;
+    previousRoute = {};
     /**
      * Pages of "no return" (when "Go back" should behave differently, probably quitting the application).
      */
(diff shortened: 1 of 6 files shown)


dkanada (approved):
Seems like an improvement over the previous behavior so I'm fine with this change.

PR #2934, opened 2021-09-06
Update bump_version to new NPM method

joshuaboniface:
**Changes**
Update the `bump_version` script to use the new NPM method with `package.json` and fix related issues.

**Issues**
N/A

### DIFF
--- a/bump_version
+++ b/bump_version
@@ -4,7 +4,6 @@
 
 set -o errexit
 set -o pipefail
-set -o xtrace
 
 usage() {
     echo -e "bump_version - increase the shared version and generate changelogs"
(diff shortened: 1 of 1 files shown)


bump_version
```diff
@@ -18,22 +17,17 @@ if [[ -z $1 ]]; then
...
-new_version_sed="$( cut -f1 -d'-' <<<"${new_version}" )"
-sed -i "s/${old_version_sed}/${new_version_sed}/g" ${shared_version_file}
-
+# Parse the version from build.yaml
 old_version="$( grep "version:" ${build_file} | sed -E 's/version: "([0-9\.]+[-a-z0-9]*)"/\1/' )"
-echo "Old version in ${build_file}: ${old_version}"
+echo "Old version: ${old_version}"
+
+# Bump the NPM version
+npm --no-git-tag-version version v${new_version}
```
dmitrylyzo, line 30: Don't we need to remove `-*` suffix from the version passed to NPM?

joshuaboniface: There isn't that suffix in the `build.yaml` that this reads the version from, so no need.

dmitrylyzo: `new_version` is the input param and https://github.com/jellyfin/jellyfin-web/blob/36f195e643534ef46115c64b7796427cebbd0b09/bump_version#L36-L38

bump_version
```diff
@@ -70,6 +65,7 @@ pushd ${fedora_spec_temp_dir}
 # Split out the stuff before and after changelog
 csplit jellyfin-web.spec  "/^%changelog/" # produces xx00 xx01
 # Update the version in xx00
+new_version_sed="$( sed 's/\./\\./g' <<<"${new_version}" )" # Escape the '.' chars
```
dmitrylyzo, line 68: I am testing with `10.8.1-qqq`. RPM version cannot contain `-`. For DEB it will be `10.8.1~qqq`.
Since the escaping is unnecessary (not a regex), we have `new_version_sed` above (with `-.*` removed), and RPM wants a "clean" version, we need to remove this line.
```suggestion
```

PR #2939, opened 2021-09-08
Log warning instead of error when dictionary has not loaded

thornbill:
**Changes**
This prevents errors for missing translations from being logged on page load when the dictionary has not been loaded yet.

**Issues**
N/A

### DIFF
--- a/src/scripts/globalize.js
+++ b/src/scripts/globalize.js
@@ -187,18 +187,22 @@ import * as userSettings from './settings/userSettings';
 
     function translateKeyFromModule(key, module) {
         let dictionary = getDictionary(module, getCurrentLocale());
-        if (!dictionary || !dictionary[key]) {
-            dictionary = getDictionary(module, fallbackCulture);
+        if (dictionary && dictionary[key]) {
+            return dictionary[key];
         }
-        if (!dictionary || !dictionary[key]) {
-            if (dictionary && isEmpty(dictionary)) {
-                console.warn('Translation dictionary is empty.');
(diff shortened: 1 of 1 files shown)


src/scripts/globalize.js
```diff
@@ -189,7 +191,11 @@ import { Events } from 'jellyfin-apiclient';
...
         if (!dictionary || !dictionary[key]) {
-            console.error(`Translation key is missing from dictionary: ${key}`);
+            if (dictionary && isEmpty(dictionary)) {
+                console.warn('Translation dictionary is empty.');
+            } else {
+                console.error(`Translation key is missing from dictionary: ${key}`);
+            }
             return key;
         }
         return dictionary[key];
```
dmitrylyzo, line 201: I tried to do fewer checks :sweat_smile: 
```js
        let dictionary = getDictionary(module, getCurrentLocale());

        if (dictionary) {
            const value = dictionary[key];
            if (value) return value;
        }

        dictionary = getDictionary(module, fallbackCulture);

        if (dictionary) {
            const value = dictionary[key];
            if (value) return value;
        }

        if (dictionary && isEmpty(dictionary)) {
            console.warn('Translation dictionary is empty.');
        } else {
            console.error(`Translation key is missing from dictionary: ${key}`);
        }

        return key;
```

thornbill: Good call that method was in need of a good cleanup!

dmitrylyzo (approved):

PR #2942, opened 2021-09-09
transcode dts audio on webos devices

dkanada:
New WebOS devices cannot direct play DTS audio but we can transcode it for now until someone implements a check. There's also a chance the codec test works fine on the older devices, but I have no ability to test this.

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -387,7 +387,7 @@ import browser from './browser';
             videoAudioCodecs.push('mp2');
         }
 
-        let supportsDts = browser.tizen || browser.web0s || options.supportsDts || videoTestElement.canPlayType('video/mp4; codecs="dts-"').replace(/no/, '') || videoTestElement.canPlayType('video/mp4; codecs="dts+"').replace(/no/, '');
+        let supportsDts = browser.tizen || videoTestElement.canPlayType('video/mp4; codecs="dts-"').replace(/no/, '') || videoTestElement.canPlayType('video/mp4; codecs="dts+"').replace(/no/, '');
 
         // DTS audio not supported in 2018 models (Tizen 4.0)
         if (browser.tizenVersion >= 4) {
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -387,7 +387,7 @@ import browser from './browser';
             videoAudioCodecs.push('mp2');
         }
 
-        let supportsDts = browser.tizen || videoTestElement.canPlayType('video/mp4; codecs="dts-"').replace(/no/, '') || videoTestElement.canPlayType('video/mp4; codecs="dts+"').replace(/no/, '');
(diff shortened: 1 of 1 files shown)


src/scripts/browserDeviceProfile.js
```diff
@@ -387,7 +387,7 @@ import browser from './browser';
             videoAudioCodecs.push('mp2');
         }
 
-        let supportsDts = browser.tizen || browser.web0s || options.supportsDts || videoTestElement.canPlayType('video/mp4; codecs="dts-"').replace(/no/, '') || videoTestElement.canPlayType('video/mp4; codecs="dts+"').replace(/no/, '');
+        let supportsDts = browser.tizen || videoTestElement.canPlayType('video/mp4; codecs="dts-"').replace(/no/, '') || videoTestElement.canPlayType('video/mp4; codecs="dts+"').replace(/no/, '');
```
dmitrylyzo, line 390: It is not a "perfect" test, but:
webOS 1.2 emulator says `probably`.
webOS 3 emulator says \``.
webOS 4 emulator says \``.

Why did you remove `options.supportsDts`?

dkanada: I ran a search for `supportsDts` with no results, and we should remove dead code when it exists. Based on your test it seems like this will work fine for now right?

dmitrylyzo: > I ran a search for `supportsDts` with no results, and we should remove dead code when it exists.

In fact, we could use it from the apps: https://github.com/jellyfin/jellyfin-tizen/blob/bc869bd5432add0fcedfd3267f86f76b5e9a6e47/tizen.js#L91
`profileBuilder({ enableMkvProgressive: false, enableSsaRender: true })`
Can't say if it's in use now. In addition, some apps construct the device profile themselves (without `profileBuilder`).

> Based on your test it seems like this will work fine for now right?

_probably_ :p
tbh, probably not. LG declares DTS for webOS 4.0: https://webostv.developer.lge.com/discover/specifications/supported-media-formats/webos-40/
But it also declares DTS for webOS 6.0: https://webostv.developer.lge.com/discover/specifications/supported-media-formats/webos-60/
:man_shrugging: 

On my Tizen 3 TV, `canPlayType` returns '', but DTS track plays directly.
On my Tizen 4 TV, `canPlayType` returns '', this is true because DTS is not supported.
So we shouldn't trust this function.

dkanada, line 390:
```suggestion
        let supportsDts = browser.tizen || options.supportsDts || videoTestElement.canPlayType('video/mp4; codecs="dts-"').replace(/no/, '') || videoTestElement.canPlayType('video/mp4; codecs="dts+"').replace(/no/, '');
```

dkanada: I added back the unrelated option at your suggestion. Regarding WebOS support, it sounds like this should be mostly fine in the short term. WebOS 1.2 is properly reporting support and WebOS 5+ no longer support DTS at all. The versions in between are vague, but without any devices to test I would rather just transcode the audio than hope we can catch every edge case between OS version and install method.

dmitrylyzo:
Superseded by #2971

PR #2944, opened 2021-09-09
Show playlist item count

SegiH:
**Changes**
Add condition so when viewing a playlist, it shows the # of items (usually audio tracks) in the playlist

**Issues**
None known


dmitrylyzo:
I added `SongCount` and `ChildCount` fetching to display the track count in list mode.
Screenshot_20210911_134840

As for the initial change, I guess we never get `item.MediaType === 'Playlist` (do we?). But there is no harm to leave it.

thornbill (approved):

PR #2945, opened 2021-09-10
Fix media indicator color

thornbill:
**Changes**
The media (count) indicators were not using the same background color as the other card indicators. This fixes that.

**Issues**
N/A

### DIFF
--- a/src/components/cardbuilder/card.scss
+++ b/src/components/cardbuilder/card.scss
@@ -150,11 +150,14 @@ button::-moz-focus-inner {
     left: 0.3em;
     text-align: center;
     vertical-align: middle;
-    width: 1.6em;
-    height: 1.6em;
+    font-size: 88%;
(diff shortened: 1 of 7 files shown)


dmitrylyzo:
Shouldn't it be the same size as the item counter?
_Just a mockup. The presence of two indicators is confusing._
Screenshot_20210911_142812

Screenshot_20210911_142613

This requires:
https://github.com/jellyfin/jellyfin-web/blob/9778f2db657a3be6ec44173e618768873d186ff6/src/components/cardbuilder/card.scss#L153-L154 to be `2em`
and
https://github.com/jellyfin/jellyfin-web/blob/9778f2db657a3be6ec44173e618768873d186ff6/src/components/indicators/indicators.scss#L53

dmitrylyzo (approved):

PR #2971, opened 2021-09-14
Fix webOS DTS audio support

dmitrylyzo:
_previously: #2942_

_After discussion on Matrix, we decided to make a new PR._

As far as I understand, LG has discontinued supporting DTS on 2020 TV models.
I can't find an official statement, and [the documentation](https://webostv.developer.lge.com/discover/specifications/supported-media-formats/webos-50/) is inconsistent (DTS is in `mkv`, but at least it's not in `avi`).

**Changes**
- Add webOS version detection
- Disable DTS on webOS 5+

**Issues**
https://github.com/jellyfin/jellyfin-webos/issues/56

To support this in webOS browser, we need to detect `NetCast` as webOS (#2781).

### DIFF
--- a/src/scripts/browser.js
+++ b/src/scripts/browser.js
@@ -95,6 +95,39 @@ function iOSversion() {
     return [];
 }
 
+function web0sVersion(browser) {
+    // Detect webOS version by web engine version
+
+    if (browser.chrome) {
+        if (browser.versionMajor >= 79) {
+            return 6;
+        } else if (browser.versionMajor >= 68) {
+            return 5;
+        } else if (browser.versionMajor >= 53) {
+            return 4;
+        } else if (browser.versionMajor >= 38) {
+            return 3;
+        } else if (browser.versionMajor >= 34) {
+            // webOS 2 browser
+            return 2;
+        } else if (browser.versionMajor >= 26) {
+            // webOS 1 browser
+            return 1;
+        }
+    } else if (browser.versionMajor >= 538) {
+        // webOS 2 app
+        return 2;
+    } else if (browser.versionMajor >= 537) {
+        // webOS 1 app
+        return 1;
+    }
+
+    console.error('Unable to detect webOS version - assume 1');
+
+    // Let's assume that we have 1.
+    return 1;
+}
+
 let _supportsCssAnimation;
 let _supportsCssAnimationWithPrefix;
 function supportsCssAnimation(allowPrefix) {
@@ -237,14 +270,16 @@ browser.tizen = userAgent.toLowerCase().indexOf('tizen') !== -1 || window.tizen 
 browser.web0s = userAgent.toLowerCase().indexOf('Web0S'.toLowerCase()) !== -1;
 browser.edgeUwp = browser.edge && (userAgent.toLowerCase().indexOf('msapphost') !== -1 || userAgent.toLowerCase().indexOf('webview') !== -1);
 
-if (!browser.tizen) {
-    browser.orsay = userAgent.toLowerCase().indexOf('smarthub') !== -1;
-} else {
+if (browser.web0s) {
+    browser.web0sVersion = web0sVersion(browser);
+} else if (browser.tizen) {
     // UserAgent string contains 'Safari' and 'safari' is set by matched browser, but we only want 'tizen' to be true
     delete browser.safari;
 
     const v = (navigator.appVersion).match(/Tizen (\d+).(\d+)/);
     browser.tizenVersion = parseInt(v[1]);
+} else {
+    browser.orsay = userAgent.toLowerCase().indexOf('smarthub') !== -1;
 }
 
 if (browser.edgeUwp) {
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -389,8 +389,8 @@ import browser from './browser';
 
         let supportsDts = browser.tizen || browser.web0s || options.supportsDts || videoTestElement.canPlayType('video/mp4; codecs="dts-"').replace(/no/, '') || videoTestElement.canPlayType('video/mp4; codecs="dts+"').replace(/no/, '');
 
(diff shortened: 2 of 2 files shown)


EraYaN (approved):
LGTM, seems like they stopped paying for the license.

EraYaN:
So it seems those versions don't seem to hold up in the webOS browser, since at least on my Chromium 53 equipped WebApp Engine based TV the browser is actually Chromium 79

dmitrylyzo:
tbh, NetCast version detection looks unreliable.
We need a user agent string for webOS 5 (6 optional).

whiteworthy:
Can confirm detection doesn't work on the browser on webOS 5, although it does work on the webOS app, playback appears to completely break after 1 to 3 minutes (on the same exact timestamp every time I test it, so it isn't random), which is weird since it doesn't happen with other transcoded codecs like TrueHD. Can't see anything special on logs or ffmpeg logs so if anyone has tested this yet please let me know if you encounter a similar issue. I found a workaround that requires you to select a direct play audio track such as an AC3 commentary and start playback with it, then switching back to the main DTS track and it won't freeze, however, it is unreliable since not every movie has a commentary track on it... I have no idea what's going on.

dmitrylyzo:
@whiteworthy
After you switch back to DTS, the video plays as `Direct stream`, transcoding the audio, right?
Could you post mediainfo of that broken playback?

Also, could you open https://shaka-player-demo.appspot.com/support.html in the webOS browser and post version of `Chrome` and `SmartTV` from the user agent string at the top?

Mchar7:
> _previously: #2942_
> 
> I can't find an official statement, and [the documentation](https://webostv.developer.lge.com/discover/specifications/supported-media-formats/webos-50/) is inconsistent (DTS is in `mkv`, but at least it's not in `avi`).
> 

Pretty sure the DTS support in an MKV container on thar chart is an error. I cannot get any DTS audio to play, including with MKVs. I'd assume there's no DTS across the board.

whiteworthy:
Can confirm the above from Mchar7.

Yes it's playing as "Direct stream"

SmartTV is 10.0 and Chrome is 79.0.3945.79 (even though it's a 2020 model), I guess that confirms EraYaN suspicions that the LG documentation you linked is wrong.

I've been able to play some content without freezing, so I will post both mediainfo from content I've been able to play and content that freezes at a specific timestamp.

One that freezes at 1:14: https://gist.github.com/whiteworthy/8e349f2fd508312d408b41fbe3b4071e
One that doesn't freeze: https://gist.github.com/whiteworthy/e64a1c3ab61b4a25636b8571c8275e3a

dmitrylyzo:
So what do we have for the browser:
webOS 4: `Chrome/79 SmartTV/10`
webOS 5: `Chrome/79 SmartTV/10`
:man_shrugging: 

We could try this (still not perfect because it can be false negative):
- don't detect version for `NetCast` _Anyway, we cannot do it reliably_
- test codec support via `canPlayType` (as in #2942) if `web0sVersion` is not detected.

I hope we can forget about the browser when we release the app :sweat_smile: 

About video freezing.
I have no idea, except perhaps that selecting the `DirectPlay` combination allows to bypass [this](https://github.com/jellyfin/jellyfin/issues/5576#issuecomment-806178759) block (_looking at HEVC HDR_).
@whiteworthy btw, for that freezing video, does it play as `Transcoding` (before it freezes) when you select DTS on the item detail page?

whiteworthy:
If you mean stats for nerds it shows as Direct streaming "HEVC direct copy" and below it shows DTS to AAC, audio plays fine before the freeze, same info on jellyfin dashboard. Maybe it's a webOS app specific bug? I will test on the browser once you implement the fixes. I haven't found a 1080p video that freezes yet (I will test further soon), so it could be due to HDR, but it's confusing since I can play 4K HEVC HDR just fine on other movies. I will attach logs of when I try to play the freezing video, not sure if there's anything useful though. 

Jellyfin (I stop playback right after video freezes): https://gist.github.com/whiteworthy/4bb3a0187861a36baa6c319051715b45
ffmpeg: https://gist.github.com/whiteworthy/b0479d253e79f554034728626353b024

I'm pretty sure it's due to a webOS update since I compiled a Default.cs transcoding profile that transcodes DTS and it worked just fine on the browser without freezing, and the fact my browser Chrome version was 68 before I updated further confirms it, couldn't find a changelog and there's no way to downgrade, unfortunately. I will make sure to never update webOS again.

dmitrylyzo:
I added what I said. The webOS browser should now transcode DTS if it isn't supported according to `canPlayType`.
I think that's all we can do. :sweat_smile:

viennaa:
I could test it on a WebOS 5 LG TV. Is there a way to get the latest 10.8.0 release version into jellyfin-webos for deploying it on the TV?

viennaa:
@dmitrylyzo @thornbill I've figured it out (`podman build -t $USER/jellyfin .` in the main repo) and have built a new image with latest master of jellyfin and jellyfin-web. The fix present works and finally there is sound for DTS files again, tested on a WebOS 5 LG CX tv.

log:
```
[18:04:27] [WRN] [10] Jellyfin.Server.Middleware.ResponseTimeMiddleware: Slow HTTP Response from http://10.0.0.1:8096/videos/***/hls1/main/58.ts?DeviceId=***&MediaSourceId=***&VideoCodec=h264,hevc&AudioCodec=aac,mp3,ac3,eac3&AudioStreamIndex=1&VideoBitrate=119616000&AudioBitrate=384000&PlaySessionId=***&api_key=***&TranscodingMaxAudioChannels=2&RequireAvc=false&Tag=***&SegmentContainer=ts&MinSegments=1&BreakOnNonKeyFrames=False&h264-profile=high,main,baseline,constrainedbaseline&h264-level=52&hevc-profile=main,main10&hevc-level=183&TranscodeReasons=AudioCodecNotSupported to 10.0.1.2 in 0:00:00.9310308 with Status Code 206
```

Really amazing work guys, thanks a lot for this fix.

PR #2972, opened 2021-09-15
Improve audio player UI for small mobile screen

tomferreira:
Shrink the cover image so elements don't overlap when using a small mobile screen (360 x 548 or less).

**Changes**
Use the flex-shrink and max-width / max-height attribute to allow the cover image to fit correctly the screen size.

| Before  | After |
| ------------- | ------------- |
| smallscreen-360x548-before | smallscreen-360x548-after
  |

### DIFF
--- a/src/components/remotecontrol/remotecontrol.scss
+++ b/src/components/remotecontrol/remotecontrol.scss
@@ -403,6 +403,7 @@
     }
 
     .nowPlayingPageImage.nowPlayingPageImageAudio {
+        width: auto;
         max-width: 100%;
         max-height: 100%;
     }
--- a/src/components/remotecontrol/remotecontrol.scss
+++ b/src/components/remotecontrol/remotecontrol.scss
@@ -326,8 +326,12 @@
     }
 
     .nowPlayingPageImageContainer {
-        width: 100%;
+        width: auto;
+        max-width: 100%;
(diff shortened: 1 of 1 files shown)


dmitrylyzo:
Similar to #2682

thornbill:
I think this change works slightly better and should have a bit better support for older browsers than the other PR.

src/components/remotecontrol/remotecontrol.scss
```diff
@@ -399,7 +403,8 @@
     }
 
     .nowPlayingPageImage.nowPlayingPageImageAudio {
-        width: 100%;
+        max-width: 100%;
```
dmitrylyzo, line 406: Without `width: auto;`, posters don't keep the aspect ratio.
```suggestion
        width: auto;
        max-width: 100%;
```

dmitrylyzo (commented):
> I think this change works slightly better and should have a bit better support for older browsers than the other PR.

In other PR, with `calc`, the poster pushes the buttons down.
In this PR they are placed as before - in the middle of the remaining space.

thornbill:
I noticed that but it should be fixable with flex grow/shrink.

thornbill:
This should be fixed by https://github.com/jellyfin/jellyfin-web/pull/3018

PR #3001, opened 2021-09-24
Update apiclient

### LINKED ISSUE #2909: Update QuickConnect
Update QuickConnect implementation to be in line with newest server changes

Server changes: https://github.com/jellyfin/jellyfin/pull/6201

thornbill:
**Changes**
Bumps the apiclient to the latest release

**Issues**
Fixes #2909

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8457,9 +8457,9 @@
       "dev": true
     },
     "jellyfin-apiclient": {
-      "version": "1.8.0",
-      "resolved": "https://registry.npmjs.org/jellyfin-apiclient/-/jellyfin-apiclient-1.8.0.tgz",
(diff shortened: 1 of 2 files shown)


dmitrylyzo (approved):

dmitrylyzo (changes requested):
Now I'm thinking, _What was I testing then?_ :sweat_smile: 
Because of [this](https://github.com/jellyfin/jellyfin-apiclient-javascript/blob/18cc70b52fca88ddea2d915d445a7bd5bc1a4167/babel.config.js#L5) babel doesn't use specified targets in the browserlist - there is `let` in the `dist`.

thornbill:
We should fix the apiclient and publish a new version imho. I opened a PR to fix it here: https://github.com/jellyfin/jellyfin-apiclient-javascript/pull/304.

dmitrylyzo (approved):

PR #3018, opened 2021-09-30
Fix audio player overlapping on small screens

thornbill:
**Changes**
Fixes the audio player ui on small screens.

This is an alternative to https://github.com/jellyfin/jellyfin-web/pull/2972 and https://github.com/jellyfin/jellyfin-web/pull/2682 that uses flexbox to size items correctly.

**Issues**
N/A

### DIFF
--- a/src/components/remotecontrol/remotecontrol.scss
+++ b/src/components/remotecontrol/remotecontrol.scss
@@ -297,7 +297,6 @@
     }
 
     .nowPlayingPageTitle {
-        /* text-align: center; */
         margin: 0;
     }
 
@@ -318,7 +317,6 @@
     }
 
     .nowPlayingInfoButtons {
-        /* margin: 1.5em 0 0 0; */
         -webkit-box-pack: center;
         -webkit-justify-content: center;
         justify-content: center;
@@ -400,7 +398,6 @@
     }
 
     .nowPlayingPageImage {
-        /* width: inherit; */
         overflow-y: hidden;
         overflow: hidden;
         margin: 0 auto;
--- a/src/components/remotecontrol/remotecontrol.scss
+++ b/src/components/remotecontrol/remotecontrol.scss
@@ -398,17 +398,11 @@
         font-size: 2em;
     }
 
-    .nowPlayingPageImage {
-        overflow-y: hidden;
-        overflow: hidden;
-        margin: 0 auto;
-
-        &.nowPlayingPageImageAudio,
-        &.nowPlayingPageImagePoster {
-            width: auto;
-            max-width: 100%;
-            max-height: 100%;
-        }
+    .nowPlayingPageImage.nowPlayingPageImageAudio,
+    .nowPlayingPageImage.nowPlayingPageImagePoster {
+        width: auto;
+        max-width: 100%;
+        max-height: 100%;
     }
 
     .playlistSectionButton .volumecontrol {
--- a/src/components/remotecontrol/remotecontrol.scss
+++ b/src/components/remotecontrol/remotecontrol.scss
@@ -325,13 +325,15 @@
     }
 
     .nowPlayingPageImageContainer {
+        display: flex;
+        flex-direction: column;
         flex-grow: 1;
         flex-shrink: 1;
+        align-items: center;
+        justify-content: center;
         width: 100%;
         min-height: 0;
         margin: 0 auto 0.5em;
-        display: flex;
-        align-items: center;
     }
 
     .nowPlayingPageImageContainerNoAlbum .cardImageContainer .cardImageIcon {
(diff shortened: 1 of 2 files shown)


thornbill:
This has an issue where the player and playlist controls overlap on small screens on iOS 10. I believe this is an issue with `flex-shrink` not working correctly in this version of Safari such that the album art is not properly scaled. imho this should probably not be considered a blocker since it seems to work correctly on any newer browser and iOS 10 usage should be pretty low at this point. We are also going to be dropping support for iOS 10 and 11 in the next app update (due to Expo no longer supporting those versions).

Screenshot 2022-02-23 at 00-08-40 iPhone 7 v10 3
(Screenshot is from an iPhone 7 running iOS 10 via BrowserStack)

src/components/remotecontrol/remotecontrol.scss
```diff
@@ -327,8 +325,13 @@
...
     .nowPlayingPageImageContainer {
+        flex-grow: 1;
+        flex-shrink: 1;
         width: 100%;
-        margin: auto auto 0.5em;
+        min-height: 0;
+        margin: 0 auto 0.5em;
+        display: flex;
+        align-items: center;
     }
```
dmitrylyzo, line 335: For Chrome 79 - it stretches the image when height is small.
Also Safari 10 starts shrinking the image, but doesn't respect the aspect ratio.
```suggestion
        flex-direction: column;
        justify-content: center;
        align-items: center;
    }
```

dmitrylyzo (commented):
Safari 10 seems to handle flexbox in a non-standard way (#3456).

src/components/remotecontrol/remotecontrol.scss
thornbill: I'm not sure `justify-content` is actually doing anything here, but I added it in case it has some effect on one of the problematic browsers. :smile:

src/components/remotecontrol/remotecontrol.scss
```diff
@@ -393,19 +400,16 @@
...
         overflow: hidden;
         margin: 0 auto;
     }
 
-    .nowPlayingPageImage.nowPlayingPageImageAudio {
-        width: 100%;
-    }
-
+    .nowPlayingPageImage.nowPlayingPageImageAudio,
     .nowPlayingPageImageContainer.nowPlayingPageImagePoster {
```
dmitrylyzo, line 409: I believe this (#1056):
https://github.com/jellyfin/jellyfin-web/blob/68f7002ca82b7064c857c4631261d06b79551535/src/components/remotecontrol/remotecontrol.js#L251-L257
should be:
```js
        context.querySelector('.nowPlayingPageImage').classList.toggle('nowPlayingPageImageAudio', item.Type === 'Audio');
        context.querySelector('.nowPlayingPageImage').classList.toggle('nowPlayingPageImagePoster', item.Type !== 'Audio');
```

and 
```suggestion
    .nowPlayingPageImage.nowPlayingPageImagePoster {
```

and remove
https://github.com/jellyfin/jellyfin-web/blob/68f7002ca82b7064c857c4631261d06b79551535/src/components/remotecontrol/remotecontrol.scss#L415-L418

dmitrylyzo (commented):
During video playback, the volume slider is overlapped by the buttons:


As a quick fix, remove this line:
https://github.com/jellyfin/jellyfin-web/blob/68f7002ca82b7064c857c4631261d06b79551535/src/components/remotecontrol/remotecontrol.scss#L324

src/components/remotecontrol/remotecontrol.scss
dmitrylyzo: Since we are changing the main axis in this way, we should use `justify-content` for vertical alignment.

There seems to be a downside: the poster is horizontally out of its container because it has a border (overall size > 100%), and `align-items` doesn't help.
What about `box-sizing: border-box;` on the image?

thornbill: It was hard for me to notice the misalignment in my testing. I did remove the border because it seemed out of place and added the same border radius we use on item cards.

src/components/remotecontrol/remotecontrol.scss
thornbill: I believe these should be good now. :crossed_fingers:

src/components/remotecontrol/remotecontrol.scss
```diff
@@ -393,24 +399,16 @@
     }
 
     .nowPlayingPageImage {
-        /* width: inherit; */
         overflow-y: hidden;
         overflow: hidden;
         margin: 0 auto;
```
dmitrylyzo, line 404: In my tests, this is unnecessary.

dmitrylyzo (commented):
There is a poster for cover-less audio.
https://github.com/jellyfin/jellyfin-web/blob/7ad5614795edae27f939dac33850b80def148b4c/src/components/remotecontrol/remotecontrol.scss#L105

I tried `width: 80%` for small media.

dmitrylyzo (approved):

PR #3019, opened 2021-09-30
Enable multiserver in development environments

thornbill:
**Changes**
Enables multi-server mode in development environments. This seems useful when running locally for testing, but I'm open to other opinions.

**Issues**
N/A

### DIFF
--- a/src/scripts/settings/webSettings.js
+++ b/src/scripts/settings/webSettings.js
@@ -72,7 +72,9 @@ export function getIncludeCorsCredentials() {
 
 export function getMultiServer() {
     return getConfig().then(config => {
-        return !!config.multiserver;
+        // Enable multi-server support in development environments
+        return process.env.NODE_ENV === 'development' // eslint-disable-line no-undef
+            || !!config.multiserver;
     }).catch(error => {
         console.log('cannot get web config:', error);
         return false;


src/scripts/settings/webSettings.js
```diff
@@ -72,7 +72,9 @@ export function getIncludeCorsCredentials() {
 
 export function getMultiServer() {
     return getConfig().then(config => {
-        return !!config.multiserver;
+        // Enable multi-server support in development environments
+        return process.env.NODE_ENV === 'development' // eslint-disable-line no-undef
```
dmitrylyzo, line 76: Don't we need to check that `process` is defined?
like this _(not tested)_
```suggestion
        return process?.env?.NODE_ENV === 'development' // eslint-disable-line no-undef
```

thornbill: Webpack replaces it with true/false at build time. I'm actually not sure if it would work with optional chaining lol. We use it in the serviceworker file already. https://github.com/jellyfin/jellyfin-web/blob/master/src/serviceworker.js#L49

dmitrylyzo: Wow, so it's like a preprocessor. But it also replaces `!!config.multiserver` to `0` :confused: 

I use dev's dist in place of installed to test webOS, but [this](https://github.com/jellyfin/jellyfin-web/blob/a70c83de3cffe65f99824e32b5d85b417ecade48/webpack.dev.js#L6) breaks targets.
So I am going to split dev config into `dev` (unminified, targeted on legacy) and `srv` (as it is now).

Is there a way to enable multi-server only for server mode? Somehow detect `serve`

thornbill: I can probably separate the conditions out so it's not replacing the config property also.

thornbill: I have updated this to use the define plugin to add a custom `WEBPACK_SERVE` variable replacement that we can use instead of the more generic `NODE_ENV`.

dmitrylyzo: I don't get how to change its value for #3027.
If var is not defined, webpack leaves it as is in the code (this can make a runtime error).
I tried to copy `DefinePlugin` to the common config with `false` value, but that from the dev config doesn't work anymore.

thornbill: :man_facepalming: yeah I should not have did it this way... it will also break in production builds since I did not add it there... I think using a single global variable would be better.

src/scripts/settings/webSettings.js
```diff
@@ -72,6 +72,10 @@ export function getIncludeCorsCredentials() {
 
 export function getMultiServer() {
     return getConfig().then(config => {
+        // Enable multi-server support when served by webpack
+        if (__WEBPACK_SERVE__) { // eslint-disable-line no-undef
+            return true;
+        }
         return !!config.multiserver;
```
dmitrylyzo, line 79: in case we have a bad config
```suggestion
    // Enable multi-server support when served by webpack
    if (__WEBPACK_SERVE__) { // eslint-disable-line no-undef
        return Promise.resolve(true);
    }

    return getConfig().then(config => {
        return !!config.multiserver;
```

PR #3020, opened 2021-09-30
Fix remote fullscreen event handler

### LINKED ISSUE #2949: Toggle fullscreen doesn't work when casting to jellyfin-mpv-shim
iPhone XR running iOS 14.7.1 with Jellyfin Mobile version 1.4.1

I'm using app to remote control `jellyfin-mpv-shim` running on my HTPC.
Some command like volume have immediate effect on my player.
However for others, like toggle fullscreen and sometimes pause/play it takes 5 or more taps to trigger it.

Thanks for great app!

thornbill:
**Changes**
Fixes a bug where clicking on some parts of the fullscreen button on remote player controls did not work because the event name was missing.

**Issues**
Fixes #2949

### DIFF
--- a/src/components/remotecontrol/remotecontrol.js
+++ b/src/components/remotecontrol/remotecontrol.js
@@ -718,11 +718,9 @@ export default function () {
             btnCommand[i].addEventListener('click', onBtnCommandClick);
         }
 
-        context.querySelector('.btnToggleFullscreen').addEventListener('click', function (e) {
+        context.querySelector('.btnToggleFullscreen').addEventListener('click', function () {
             if (currentPlayer) {
-                playbackManager.sendCommand({
-                    Name: e.target.getAttribute('data-command')
-                }, currentPlayer);
+                playbackManager.toggleFullscreen(currentPlayer);
             }
         });
         context.querySelector('.btnAudioTracks').addEventListener('click', function (e) {


dmitrylyzo:
This can probably be fixed by replacing `e.target` (button or button child) -> `this` (button), as here https://github.com/jellyfin/jellyfin-web/blob/30d33de1452247409ad8a99572a0f15eb67bfa02/src/components/remotecontrol/remotecontrol.js#L683

thornbill:
I believe that would also work, but I don't see any reason not to use the specific method for toggling fullscreen here similar to all the playback control handlers.

https://github.com/jellyfin/jellyfin-web/blob/30d33de1452247409ad8a99572a0f15eb67bfa02/src/components/remotecontrol/remotecontrol.js#L736-L740

dmitrylyzo:
> I believe that would also work, but I don't see any reason not to use the specific method for toggling fullscreen here similar to all the playback control handlers.

`btnStop` has no `data-command` because it must be supported.

`btnToggleFullscreen` handler can be replaced with the common one by adding the `btnCommand` class to the button:
https://github.com/jellyfin/jellyfin-web/blob/30d33de1452247409ad8a99572a0f15eb67bfa02/src/components/remotecontrol/remotecontrol.js#L717-L719

On the other hand, a direct call is faster.

dmitrylyzo (approved):

PR #3023, opened 2021-10-02
Show chapters markers to the seek bar

chp-1960:
Adding chapters markers to the seek bar of the web client

**Changes**
It will be useful to add the feature of showing the "chapters markers" to the seek bar of the web client regardless of extracting chapters pictures. 

Requested new features :
1) If the video / movie already include normalized chapters markers : show them in the seek bar.
2) if the video / movie doesn't include chapter markers. don't show them
3) The video / movie has generated chapter markers pictures, then show the associated picture and the chapter marker in the seek bar.

Optional features :
3) ability to show or to hide these markers within the client preferences.
4) Ability to extract pictures chapter marker per video library (actually, it's for whole libraries which consume HDD space and time while scanning the libraries)  

**Issues**
No issues yet


Shadowghost:
A PR is the wrong place for and the wrong way of submitting a feature request.

If we'd allow feature requests on Github it would be as an issue but since we do not, please (re)submit your request on our fider instance at https://features.jellyfin.org
Check if a similar request already exists before submitting.

PR #3044, opened 2021-10-05
Fix repository url link not opening as external

thornbill:
**Changes**
Fixes repository url links not being treated as external links. This would break web wrapper apps since the link would be opened in the webview instead of calling the NativeShell handler.

**Issues**
N/A


dmitrylyzo (approved):

blob03:
I suggest we also add a "cursor:default" style to the enclosing div. This will get rid of the pointer cursor inherited from class 'listItem'
which gives a false sense that everything is clickable.
line 66: `html += '<div class="listItem listItem-border" style="cursor:default">';`

PR #3070, opened 2021-10-15
Convert edit user page to react

grafixeyehero:
Depend on  convert NewUserPage to react #3026


grafixeyehero:
> 
> 
> It looks like this will need rebased since your other PR has been merged.

This is my first rebase

src/components/dashboard/users/LnkEditUserPreferences.tsx
```diff
@@ -0,0 +1,30 @@
...
+    __html: `<a
+        is="emby-linkbutton"
+        class="${className}"
+        href='#'
+        >
+        ${title}
+    </a>`
+});
+
+const LnkEditUserPreferences: FunctionComponent<IProps> = ({ className, title }: IProps) => {
```
thornbill, line 19: It looks like this should be "Link". The file name has the same typo.

src/components/dashboard/users/SelectElement.tsx
```diff
@@ -0,0 +1,38 @@
...
+        >
+        ${option}
+    </select>`
+});
+
+type IProps = {
+    className?: string;
+    label?: string;
+    currentProviderId: string;
+    providers: any
```
thornbill, line 18: The type should probably be `any[]`

thornbill (changes requested):
I have not finished reviewing this, but here are a couple items I noticed after a first pass.

src/components/dashboard/users/LnkEditUserPreferences.tsx
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/3070/commits/1a66afa6104ce9c68bac300626a35f5d1426a15f

src/components/dashboard/users/SelectElement.tsx
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/3070/commits/1a66afa6104ce9c68bac300626a35f5d1426a15f

src/components/dashboard/users/TabLinkElement.tsx
```diff
@@ -0,0 +1,36 @@
+// eslint-disable-next-line eslint-comments/disable-enable-pair
```
thornbill, line 1: This should be removed and the rule enabled at the end of the file.

thornbill (changes requested):
In addition to the one comment here, I noticed a few issues when testing.
 
- [ ] Tabs are not styled correctly when on the "Profile" tab
- [ ] Spacing seems a bit tighter than it was previously particularly in the checkbox lists
- [x] The section for allowing media deletion has some issues
  - All libraries checkbox should be included in list of libraries
  - `DeleteFoldersAccess` and `DeleteFoldersAccessHelp` sections should be removed

thornbill (approved):

PR #3071, opened 2021-10-15
Convert userLibraryAccessPage to react

grafixeyehero:
Depend on   Convert edit user page to react #3070


src/components/pages/UserLibraryAccessPage.tsx
```diff
@@ -0,0 +1,334 @@
...
+        loadChannels(user, channels);
+        loadMediaFolders(user, mediaFolders);
+        loadDevices(user, devices);
+        loading.hide();
+    }, [loadChannels, loadDevices, loadMediaFolders]);
+
+    const loadData = useCallback(() => {
+        loading.show();
+        const userId = appRouter.param('userId');
+        // eslint-disable-next-line compat/compat
```
thornbill, line 123: Is this required to use `Promise`?

src/components/pages/UserLibraryAccessPage.tsx
```diff
@@ -0,0 +1,334 @@
...
+            loading.hide();
+            toast(globalize.translate('SettingsSaved'));
+        };
+
+        element?.current?.querySelector('.chkEnableAllDevices').addEventListener('change', function (this: HTMLInputElement) {
+            if (this.checked) {
+                element?.current?.querySelector('.deviceAccessListContainer').classList.add('hide');
+            } else {
+                element?.current?.querySelector('.deviceAccessListContainer').classList.remove('hide');
+            }
```
thornbill, line 186: It would probably be better to use `classList.toggle()` for these.

thornbill (changes requested):
Looks great overall! I just had a couple minor questions/comments.

Should the console.log statements be removed?

src/components/pages/UserLibraryAccessPage.tsx
grafixeyehero: cf39bc0

src/components/pages/UserLibraryAccessPage.tsx
grafixeyehero: previously was used  jQuery $.deferred
https://github.com/jellyfin/jellyfin-web/blob/7ea8ce7debae59cb91c8f1f3a647931f49217c93/src/controllers/dashboard/users/userlibraryaccess.js#L165-L173

thornbill: Ah I see this is an issue with our eslint configuration. The polyfills are not being included for jsx/tsx files. That probably deserves a separate PR. :smile:

thornbill (approved):

PR #3076, opened 2021-10-17
Refactoring duplicate code

grafixeyehero:
Depend on Convert userPasswordPage & UserImagePage to react #3073

### DIFF
--- a/src/components/dashboard/users/AccessContainer.tsx
+++ b/src/components/dashboard/users/AccessContainer.tsx
@@ -3,26 +3,26 @@ import globalize from '../../../scripts/globalize';
 import CheckBoxElement from './CheckBoxElement';
 
 type IProps = {
-    ContainerClassName?: string;
-    HeaderTitle?: string;
-    CheckBoxClassName?: string;
-    CheckBoxTitle?: string;
-    ListContainerClassName?: string;
-    AccessClassName?: string;
-    ListTitle?: string;
-    Description?: string;
+    containerClassName?: string;
+    headerTitle?: string;
+    checkBoxClassName?: string;
+    checkBoxTitle?: string;
+    listContainerClassName?: string;
+    accessClassName?: string;
+    listTitle?: string;
+    description?: string;
     children?: React.ReactNode
 }
 
-const AccessContainer: FunctionComponent<IProps> = ({ContainerClassName, HeaderTitle, CheckBoxClassName, CheckBoxTitle, ListContainerClassName, AccessClassName, ListTitle, Description, children }: IProps) => {
+const AccessContainer: FunctionComponent<IProps> = ({containerClassName, headerTitle, checkBoxClassName, checkBoxTitle, listContainerClassName, accessClassName, listTitle, description, children }: IProps) => {
     return (
-        <div className={ContainerClassName}>
-            <h2>{globalize.translate(HeaderTitle)}</h2>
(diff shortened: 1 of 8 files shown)


thornbill:
@grafixeyehero this one would still be good to get in right? Would you want to rebase it to clear up the conflicts?

src/components/dashboard/users/AccessContainer.tsx
```diff
@@ -0,0 +1,41 @@
+import React, { FunctionComponent } from 'react';
+import globalize from '../../../scripts/globalize';
+import CheckBoxElement from './CheckBoxElement';
+
+type IProps = {
+    ContainerClassName?: string;
```
thornbill, line 6: I think we should probably standardize on camelCase for prop names.

src/components/dashboard/users/AccessContainer.tsx
```diff
@@ -0,0 +1,41 @@
...
+    AccessClassName?: string;
+    ListTitle?: string;
+    Description?: string;
+    children?: React.ReactNode
+}
+
+const AccessContainer: FunctionComponent<IProps> = ({ContainerClassName, HeaderTitle, CheckBoxClassName, CheckBoxTitle, ListContainerClassName, AccessClassName, ListTitle, Description, children }: IProps) => {
+    return (
+        <div className={ContainerClassName}>
+            <h2>{globalize.translate(HeaderTitle)}</h2>
```
thornbill, line 20: Should we pass in the translated strings instead like we do for `SectionTitleContainer`?

src/components/dashboard/users/SectionTitleContainer.tsx
```diff
@@ -0,0 +1,35 @@
...
+import SectionTitleButtonElement from './SectionTitleButtonElement';
+import SectionTitleLinkElement from './SectionTitleLinkElement';
+
+type IProps = {
+    title: string;
+    isBtnVisible?: boolean;
+    titleLink?: string;
+}
+
+const SectionTitleContainer: FunctionComponent<IProps> = ({title, isBtnVisible, titleLink}: IProps) => {
```
thornbill, line 11: It looks like we could default `isBtnVisible` to false since that seems to be the case for most instances.

thornbill (changes requested):
Looks good overall! Just had a couple questions with regards to some consistency things we should consider.

src/components/dashboard/users/SectionTitleContainer.tsx
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/3076/commits/9ca943fcca9c573867cf8b4ee0a9444fa790f093

src/components/dashboard/users/AccessContainer.tsx
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/3076/commits/0a881d6f6f1c87efb1a5330499b676d9991461c2

src/components/pages/UserProfilesPage.tsx
```diff
@@ -133,30 +132,16 @@ const UserProfilesPage: FunctionComponent = () => {
     return (
         <div ref={element}>
             <div className='content-primary'>
-                <div className='verticalSection verticalSection-extrabottompadding'>
```
thornbill, line 136: It looks like we are losing the extra padding here. Does that matter?

grafixeyehero: No need.

thornbill (approved):

PR #3144, opened 2021-11-05
Decode blurhash in a WebWorker

ferferga:
"Backported" some good stuff from Vue that I attempted here but was not possible before ES6 :)

* Decode blurhash in a web worker. 
* Increased the resolution of the sample by 2x2, which is even a sweeter spot than 18x18 (as we have "extra" capabilities now without blocking the main thread).
* Wait for transitions to end before switching image and blurhash: https://github.com/jellyfin/jellyfin-web/pull/1823 was a necessary fix, but it added some flashing while switching. Was partly fixed by https://github.com/jellyfin/jellyfin-web/pull/2309 but still some flashing appearing while scrolling fast. I was not happy with the result either, so I took the chance of this PR to do another revisit, so now everything is correctly timed and looks like the original implementation :)

### DIFF
--- a/src/components/images/imageLoader.js
+++ b/src/components/images/imageLoader.js
@@ -48,7 +48,7 @@ worker.addEventListener(
 
     function itemBlurhashing(target, hash) {
         try {
-            // Although the default values recommended by Blurhash developers is 32x32, a size of 18x18 seems to be the sweet spot for us,
+            // Although the default values recommended by Blurhash developers is 32x32, a size of 20x20 seems to be the sweet spot for us,
             // improving the performance and reducing the memory usage, while retaining almost full blur quality.
             // Lower values had more visible pixelation
             const width = 20;
--- a/src/components/images/imageLoader.js
+++ b/src/components/images/imageLoader.js
@@ -23,8 +23,8 @@ import './style.scss';
             // Although the default values recommended by Blurhash developers is 32x32, a size of 18x18 seems to be the sweet spot for us,
             // improving the performance and reducing the memory usage, while retaining almost full blur quality.
             // Lower values had more visible pixelation
-            const width = 32;
-            const height = 32;
+            const width = 20;
+            const height = 20;
             const pixels = await getPixels({
                 hash,
                 width,
--- a/src/components/images/imageLoader.js
+++ b/src/components/images/imageLoader.js
@@ -54,7 +54,8 @@ worker.addEventListener(
             // Lower values had more visible pixelation
             const width = 20;
             const height = 20;
-            targetDic[hash] = target;
+            targetDic[hash] = (targetDic[hash] || []).filter(item => item !== target);
+            targetDic[hash].push(target);
 
             worker.postMessage({
                 hash,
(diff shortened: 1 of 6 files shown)


dmitrylyzo:
This doesn't work on webOS 1.2:
- `comlink` needs to be transpiled (`const`).
- ReferenceError: Can't find variable: Proxy - main
- ReferenceError: Can't find variable: Promise - worker

MrTimscampi:
> ReferenceError: Can't find variable: Proxy - main

Comlink uses [ES6 Proxy](https://caniuse.com/proxy), which can't really be polyfilled properly.

Polyfill is here: https://github.com/tvcutsem/harmony-reflect

But it does mention significant differences to the ES2015 spec, as well as needing to enable a feature flag on Chromium >= v19 && <= v37 (Which would encompass WebOS 1.x to 3.x). Afaik, it's not possible to enable feature flags on WebOS.

It's part of why Vue 3 only supports evergreen browsers now, since they reworked their reactivity layer to use Proxy.

Even with transpiling `comlink`, it still won't work on these browsers.

> ReferenceError: Can't find variable: Promise - worker (webOS 1.2)

The Promise polyfill only works in the main context, as it is loaded currently. You'd need to load it in the worker too for it to be available there.

ferferga:
@dmitrylyzo Can you test now please?

I removed the hack you added in #3147 by removing the ``requestAnimationFrame`` callback (which makes all that section of the logic fully synchronous, so the attributes of the element can't change between the time elapsed on the callback request and  it being fired. This was also needed to fix an issue in this new canvas/image switching logic where changing scroll directions too fast sometimes wouldn't show the canvas fast enough (Chromium).

Is the current fix for that issue OK for you?

dmitrylyzo (commented):
> Is the current fix for that issue OK for you?

It seems to work fine.

But I still have `ReferenceError: Can't find variable: Promise` inside the worker on webOS 1.2 - as a result, there is no blurhash.
This comes from Webpack itself:
```js
/******/ 	/* webpack/runtime/ensure chunk */
/******/ 	!function() {
/******/ 		__webpack_require__.f = {};
/******/ 		// This file contains only the entry chunk.
/******/ 		// The chunk loading function for additional chunks
/******/ 		__webpack_require__.e = function(chunkId) {
/******/ 			return Promise.all(Object.keys(__webpack_require__.f).reduce(function(promises, key) {
/******/ 				__webpack_require__.f[key](chunkId, promises);
/******/ 				return promises;
/******/ 			}, []));
/******/ 		};
/******/ 	}();
```

src/components/images/imageLoader.js
```diff
@@ -12,42 +26,45 @@ import './style.scss';
...
-                target.removeAttribute('data-blurhash');
+            target.parentNode.insertBefore(canvas, target);
+            target.classList.add('blurhashed');
+            target.removeAttribute('data-blurhash');
+        });
+    }
+
+    function itemBlurhashing(target, hash) {
+        try {
+            // Although the default values recommended by Blurhash developers is 32x32, a size of 18x18 seems to be the sweet spot for us,
```
thornbill, line 51: We seem to be using 20x20 now instead of 18x18

ferferga: Oops!

This lead me to another question: I used 20x20 because it's right a bit better than 18x18 and we take better advantage of our new "super power" of unloading the task from the main thread, but not *too much* either.

It's that fine for you or better use 32x32 as Blurhash people prefer?

thornbill: 20x20 seems fine to me 🤷‍♂️

PR #3159, opened 2021-11-10
Add SKIP_PREPARE to install instructions to skip building

thornbill:
**Changes**
Adds `SKIP_PREPARE` to the install instructions to avoid running a production build

**Issues**
N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ Jellyfin Web is the frontend used for most of the clients available for end user
 2. Install build dependencies in the project directory.
 
    ```sh
-   npm install
+   SKIP_PREPARE="true" npm install
    ```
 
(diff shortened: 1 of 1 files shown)


thornbill:
This will not work on Windows, we should include Windows instructions also.

dmitrylyzo:
I use `SKIP_PREPARE=1 npm ci --no-audit`

> This will not work on Windows, we should include Windows instructions also.

For Tizen, I had to make a specialized guide (in Wiki): https://github.com/jellyfin/jellyfin-tizen/wiki/Build-on-Windows

thornbill:
As we move towards removing the prepare script, this should become irrelevant.

PR #3161, opened 2021-11-10
Rollback hls.js to 0.14

### LINKED ISSUE #3158: Cannot seek/rewind videos that require transcoding (`bufferSeekOverHole` error from hls.js)
**Describe The Bug**

`bufferSeekOverHole` error when seeking videos that require transcoding. Basically the exact same issue described [here](https://github.com/video-dev/hls.js/issues/3313).

To recreate the issue in the Chrome browser web client:

[Google drive link to video that will need transcoding to play in Chrome.](https://drive.google.com/file/d/1W0x7JPxWtq2KOv9X3N5MlJqsmVBn6Res/view?usp=sharing)

1. Open a video that needs transcoding and start playing from the beginning.
2. Skip to a random timestamp `t` using the seek bar.
3. Wait until the video starts playing at timestamp `t`.
4. Seek to any timestamp between `0` and `t` using the seek bar.
5. The playing position will get pushed back to `t`.

A `bufferSeekOverHole` error is shown in the console.

The expected behavior is to have the video start buffering and playing at the desired timestamp. 

**Screenshots**

alt text

 - Server Platform: Ubuntu 21.10 Server
 - Browser: Chrome 95.0.4638.69 on Ubuntu 21.10 Desktop
 - Jellyfin Version: 10.7.7.0 running natively as a service. 

**Additional Context**

I have tried adding/changing the hls.js config parameters at this [line where an hls.js instance is created](https://github.com/jellyfin/jellyfin-web/blob/0831207057c26ba5d760b6e9617d05260a142f13/src/plugins/htmlVideoPlayer/plugin.js#L400) (e.g. `maxBufferHole` and `highBufferWatchdogPeriod`), but the issue still persists. I'm also using intel QuickSync as the hardware transcoder, but I don't think that affects the web client frontend.

thornbill:
**Changes**
This rollsback hls.js to 0.14 and prevents dependabot from opening PRs for major version upgrades due to this upstream issue: https://github.com/video-dev/hls.js/issues/4367

**Issues**
Fixes #3158

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -5,6 +5,9 @@ updates:
     schedule:
       interval: weekly
     open-pull-requests-limit: 10
+    ignore:
+      - dependency-name: hls.js
+        update-types: [ version-update:semver-major ]
 
   - package-ecosystem: github-actions
     directory: /
--- a/package-lock.json
+++ b/package-lock.json
@@ -6073,8 +6073,7 @@
     "eventemitter3": {
       "version": "4.0.7",
       "resolved": "https://registry.npmjs.org/eventemitter3/-/eventemitter3-4.0.7.tgz",
-      "integrity": "sha512-8guHBZCwKnFhYdHr2ysuRWErTwhoN2X8XELRlrRwpmfeY2jjuUN4taQMsULKUVo1K4DvZl+0pgfyoysHxvmvEw==",
-      "dev": true
+      "integrity": "sha512-8guHBZCwKnFhYdHr2ysuRWErTwhoN2X8XELRlrRwpmfeY2jjuUN4taQMsULKUVo1K4DvZl+0pgfyoysHxvmvEw=="
     },
     "events": {
       "version": "3.3.0",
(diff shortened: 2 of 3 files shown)


ThibaultNocchi (approved):

ccheng32:
Hi @thornbill thanks for the fix. Have you verified that hls.js v0.14.7 fixes #3158?

AFAIK I encountered the problem on an unmodified jellyfin server running v10.7.7, which seems to serve hls.js v0.14.17 for the web frontend. ([link](https://github.com/jellyfin/jellyfin-web/blob/be195b0e240b4a4b871eb6a7fe8e4830bd9ab12c/package.json#L65))

ccheng32:
Some more updates from my testing. I acutally tried out some more browsers and found out that Chrome and Firefox have this issue, while Safari on MacOS does not have this issue at all. I suspect this line may be the culprit:

https://github.com/jellyfin/jellyfin-web/blob/0831207057c26ba5d760b6e9617d05260a142f13/src/plugins/htmlVideoPlayer/plugin.js#L394

ThibaultNocchi:
Safari does handle HLS natively, and I think the client takes profit of it. @thornbill could confirm, but I guess it's why there's no problem with Safari.

crobibero (approved):

thornbill:
I did not see any errors in Firefox, but I didn't test it too thoroughly either. 😅

PR #3185, opened 2021-11-22
Disable any type warnings

thornbill:
**Changes**
Disables warnings for using the `any` type in places where types are currently unavailable (api objects).

**Issues**
N/A

### DIFF
--- a/src/components/dashboard/users/UserCardBox.tsx
+++ b/src/components/dashboard/users/UserCardBox.tsx
@@ -25,7 +25,7 @@ const createButtonElement = () => ({
 });
 
 type IProps = {
-    user?: Record<string, any>;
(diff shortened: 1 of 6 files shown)


ferferga:
You can have proper typing support by importing the types we have in our own axios apiclient.

Previously, as it was generated using CommonJS instead of ES6, just installing the package meant that all the components of it would be added to the bundle. However, since one of the 10.7.X minor updates (after https://github.com/OpenAPITools/openapi-generator/pull/10308 got merged) everything is using ES6.

In ES6 only modules that are used are imported into the output bundle. As types are an abstraction of Typescript (and not used at runtime), it doesn't matter that you add the axios client package to Web, only types will be used from the package. And as types are not necessary, no further increase in output bundle size will be seen.

thornbill:
#3433 is a better solution for this.

PR #3206, opened 2021-12-01
Update HWA settings to adapt the server code changes

nyanmisaka:
**Changes**
- Show tone-mapping options to QSV on Windows.
- Allow QSV users to use native DXVA or VAAPI decoder.
- Add Intel Low-Power encoding options.
- Add AV1 hardware decoding option.
- Add threading options up to 16.
- Add related translations.
- Fix the issue that the name of V4L2 was hardcoded to `h264_v4l2m2m`

**Issues**
Fixes https://github.com/jellyfin/jellyfin/issues/6931


neo-neo1:
> Fix the issue that the name of V4L2 was hardcoded to `h264_v4l2m2m`

Thanks for the quick work!

src/strings/en-us.json
```diff
@@ -1459,7 +1459,7 @@
     "TitleHardwareAcceleration": "Hardware Acceleration",
     "TitleHostingSettings": "Hosting Settings",
     "TitlePlayback": "Playback",
-    "TonemappingAlgorithmHelp": "Tone mapping can be fine-tuned. If you are not familiar with these options, just keep the default. The recommended value is Hable.",
+    "TonemappingAlgorithmHelp": "Tone mapping can be fine-tuned. If you are not familiar with these options, just keep the default. The recommended value is BT.2390.",
```
crobibero, line 1462: Since the recommended value changed, should the default value also be changed?

nyanmisaka: There should be no default value for tone-mapping algorithm. In theory, the SDR version of each HDR movie has a different algorithm. We recommend BT.2390 to follow the default behavior of mpv.

crobibero: I get what you're saying, but when I opened the webpage on a brand new Jellyfin build, I had the tone-mapping set to "Hable", which will be saved to the server and then used.. I think the default in the drop down should match what the strings say

nyanmisaka: Have you upgraded the server to master branch?
The default value is set on the server side.

https://github.com/jellyfin/jellyfin/blob/78bb581f0c7d27f7f9a7a6d3ac81a2a6c6fbe781/MediaBrowser.Model/Configuration/EncodingOptions.cs#L21

crobibero: you're 100% correct, I needed to re-re-clean my config. Works great when starting from scratch

crobibero (approved):

joshuaboniface (approved):

PR #3231, opened 2021-12-19
fix: Oculus browser is no longer marked as TV

emilrowland:
**Changes**

Changed the browser feature detection to correctly identify the oculus browser. Now all features are working and the oculus browsers built in VR support is also working for 3D videos.

### DIFF
--- a/src/scripts/browser.js
+++ b/src/scripts/browser.js
@@ -2,6 +2,11 @@ function isTv() {
     // This is going to be really difficult to get right
     const userAgent = navigator.userAgent.toLowerCase();
 
+    // The OculusBrowsers userAgent also has the samsungbrowser defined but is not a tv.
+    if (userAgent.indexOf('oculusbrowser') !== -1) {
+        return false;
+    }
+
     if (userAgent.indexOf('tv') !== -1) {
         return true;
     }


dmitrylyzo:
So you need Jellyfin to be like Desktop (mouse, resize), right?
Or are there other features?

emilrowland:
Yes VR version of web pages are usually the Desktop version if not a specific VR version of the page.
For Jellyfin the Desktop page works great on my VR setup. You have access to everything that you would need in VR like Download and full screen support of the Video player, that wasn't available when it thought my Oculus Setup was an TV.

But there is no need for anymore features.

src/scripts/browser.js
```diff
@@ -2,6 +2,11 @@ function isTv() {
     // This is going to be really difficult to get right
     const userAgent = navigator.userAgent.toLowerCase();
 
+    //The OculusBrowsers userAgent also has the samsungbrowser defined but is not a tv.
+    if(userAgent.indexOf('oculusbrowser') !== -1) {
```
dmitrylyzo, line 6: Please amend the commit with these changes and I think it's fine.
```suggestion
    // The OculusBrowsers userAgent also has the samsungbrowser defined but is not a tv.
    if (userAgent.indexOf('oculusbrowser') !== -1) {
```

emilrowland: Sure, I have updated the commit.

dmitrylyzo:
`dist.tar`?

emilrowland:
Sorry, noticed my build file there. Will remove now.

dmitrylyzo (approved):
LGTM

PR #3237, opened 2021-12-22
Added hovereffect on listItems for all themes

Piccar:
Every menu has a hovereffect but the menuitems in the usermenu lacks that feature.

### DIFF
--- a/src/themes/blueradiance/theme.css
+++ b/src/themes/blueradiance/theme.css
@@ -261,6 +261,10 @@ html {
     background: rgba(0, 0, 0, 0.3);
 }
 
+.listItem:hover{
+    background: #10223e;
+}
+
 .progressring-spiner {
     border-color: #00a4dc;
 }
--- a/src/themes/dark/theme.css
+++ b/src/themes/dark/theme.css
@@ -238,6 +238,10 @@ html {
     background: #333;
 }
 
+.listItem:hover{
+    background: #252528;
+}
+
 .progressring-spiner {
     border-color: #00a4dc;
 }
--- a/src/themes/wmc/theme.css
+++ b/src/themes/wmc/theme.css
@@ -244,6 +244,10 @@ html {
     background: #333;
 }
 
+.listItem:hover {
+    background: #0c2a4e;
+}
+
 .progressring-spiner {
     border-color: #00a4dc;
 }
--- a/src/themes/light/theme.css
+++ b/src/themes/light/theme.css
@@ -269,6 +269,10 @@ html {
     background: #ddd;
 }
 
+.listItem:hover{
+    background: #f2f2f2;
+}
+
 .progressring-spiner {
     border-color: #00a4dc;
 }
--- a/src/themes/purplehaze/theme.css
+++ b/src/themes/purplehaze/theme.css
@@ -347,6 +347,10 @@ a[data-role=button] {
     background: rgba(0, 0, 0, 0.3);
 }
 
+.listItem:hover{
+    background: #12122f;
+}
+
 .progressring-spiner {
     border-color: #48c3c8;
 }


dmitrylyzo (commented):
Please add `hover` for Apple TV theme. Maybe `rgba(0, 0, 0, 0.1)` for now.

src/themes/blueradiance/theme.css
```diff
@@ -261,6 +261,10 @@ html {
     background: rgba(0, 0, 0, 0.3);
 }
 
+.listItem:hover {
+    background: #10223e;
```
dmitrylyzo, line 265: I am not sure if it will look better, but we could change colors like this:
`focus`: `rgba(0, 0, 0, 0.5)`
`hover`: `rgba(0, 0, 0, 0.3)`
or
`focus`: `rgba(0, 0, 0, 0.3)`
`hover`: `rgba(0, 0, 0, 0.15)`

src/themes/wmc/theme.css
```diff
@@ -244,6 +244,10 @@ html {
     background: #333;
 }
 
+.listItem:hover {
+    background: #0c2a4e;
```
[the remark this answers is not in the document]
Piccar: Due to the blue background the focus or hover colorchange won't be visable.

src/themes/blueradiance/theme.css
Piccar: I liked your second option!

src/themes/dark/theme.css
```diff
@@ -238,6 +238,10 @@ html {
     background: #333;
 }
 
+.listItem:hover {
+    background: #252528;
```
dmitrylyzo, line 242: I think the `Dark` theme should be monochrome, but the buttons (`Sort`, `Filter`, etc.) are blue when hovered, so I dunno :man_shrugging:

Piccar: I know what you mean. But the menu that opens when you click on the three lines has a grey hover effect. I thought that mimicking that behavior would be fitting. Sort and filter are only small icons. Maybe thats why they are blue when hoverd?

dmitrylyzo: I mean, it could be `#252525` (equal channels).

Btw, there is `actionSheetMenuItem:hover` (it should go after `listItem:hover` to work properly):
https://github.com/jellyfin/jellyfin-web/blob/958e938f373b6239a0b46acbc774fe423d286c4d/src/themes/dark/theme.css#L175-L177
For `Dark` we could take `#242424` and remove `actionSheetMenuItem:hover` since `actionSheetMenuItem` is a `listItem`.
For other themes, the background color under the menu may be different, so we need that custom style.

dmitrylyzo:
I looked at `actionSheetMenuItem` (popup menu items) again.

Please remove (now we use the same color):
https://github.com/jellyfin/jellyfin-web/blob/2bdbc93f5bf07788bb37f76ddd0e013bc8edf981/src/themes/light/theme.css#L203-L206

Please move after `.listItem:hover` to make them work:
https://github.com/jellyfin/jellyfin-web/blob/2bdbc93f5bf07788bb37f76ddd0e013bc8edf981/src/themes/appletv/theme.css#L192-L195
https://github.com/jellyfin/jellyfin-web/blob/2bdbc93f5bf07788bb37f76ddd0e013bc8edf981/src/themes/blueradiance/theme.css#L187-L190
https://github.com/jellyfin/jellyfin-web/blob/2bdbc93f5bf07788bb37f76ddd0e013bc8edf981/src/themes/purplehaze/theme.css#L280-L283

dmitrylyzo:
@Piccar Could you finish the PR by resolving [these suggestions](https://github.com/jellyfin/jellyfin-web/pull/3237#issuecomment-1003609109)?
Also, I would squash all commits to the first one. _No need to mention me._

dmitrylyzo:
You forgot to remove the existed `actionSheetMenuItem:hover` in those 3 files - now they are dubbed.

dmitrylyzo (approved):
CodeQL is not relevant.
I would squash to the first commit again.

Issue #2792, opened 2021-07-15
Audio visualizer animation continues when music is paused


SethFalco:
**Describe The Bug**

When listening to music on Jellyfin, the queue page replaces the cover image of the playing track with a GIF.
However, when the music is paused, the animation continues.

![image](https://raw.githubusercontent.com/jellyfin/jellyfin-web/master/src/assets/img/equalizer.gif)
> On the side, this GIF could probably be replaced with an SVG.

**Steps To Reproduce**

1. Go to a music library
2. Click on the play button for an album
3. Click on the bar at the bottom to get taken to `/queue`
4. Pause the music

**Expected Behavior**

I would expect one of the following:
1. The animation pauses.
2. The image is replaced with a pause icon. ⏸️

I think number 2 would be better and easier to implement.
You can probably just use the pause Material Icon ([here)](https://fonts.google.com/icons?selected=Material+Icons&icon.query=pause) since that's already in Jellyfin anyway.

**Logs**

N/A

**Screenshots**

![image](https://user-images.githubusercontent.com/22801583/125751286-442ddf0d-51eb-40e8-9347-24f31b95fcc8.png)
> Just pretend you can see the GIF still moving. ^-^'

**System (please complete the following information):**
 - Platform: Linux
 - Browser: Firefox / [Jellyfin Media Player](https://github.com/jellyfin/jellyfin-media-player)
 - Jellyfin Version: 10.7.6

**Additional Context**

This is just a minor UX concern, it can throw a user off when part of the UI suggests that music should be playing when it's not.

SethFalco:
I made an SVG version of the GIF, it's smaller and looks much smoother in the browser.

| Type | Size |
|---|---
| GIF | 2.6 KB (2,621 bytes)
| SVG | 1.7 KB (1,687 bytes)

```xml
[...]

PR #3239, opened 2021-12-24
Replace equalizer.gif with equalizer.svg

SethFalco:
**Changes**

I made an SVG version of the GIF.
I've tried to keep this as close to the GIF as possible, so they should be near identical.

There are 2 benefits:
* The file size is smaller.
* The animation is **_much_** smoother.

| Type | Size |
|---|---
| GIF | 2.6 KB (2,621 bytes)
| SVG | 1.5 KB (1,541 bytes)

You can see them side-by-side here:
https://codepen.io/SethFalco/pen/LYzOpJv

**Issues**
* Related to https://github.com/jellyfin/jellyfin-web/issues/2792

---

I realize this is a very minor optimization, but this was mostly because I was curious if I could make a nice equalizer vector, tbh.

You're welcome to point out anything you'd like changed about it. It should be pretty easy since it's just a vector now anyway.

As a side-note, you don't want to run `equalizer.svg` through [svgo](https://github.com/svg/svgo). The project has a bug that removes the `to` key frames, which screws up the animation. I've already optimized the SVG with it, then manually fixed that issue.


thornbill:
While this seems like a positive change overall, it will require some testing to verify it works on some of the older browsers shipped to TVs (webOS and Tizen).

SethFalco:
Sounds fair, I've only tested in on Chromium and Firefox for now.
Do you (or someone else on the team) have the means to test them, or will I have to arrange that?

I can test it on the browser on my Samsung TV (Tizen) and report the results, meanwhile.

If you find that it doesn't work somewhere, feel free to just let me know and I can check if there's another way I can achieve the same thing.

SethFalco:
Just a follow-up, so I tested in on my Samsung TV in the default browser, and it worked as expected.

Name: UE40NU7122
Product: https://www.samsung.com/uk/support/model/UE40NU7120KXXU/
OS Version: 1320

SethFalco:
Very minor change that shaves off 2.22% bytes (35 bytes) from the original SVG using [CSSO](https://github.com/css/csso).

Just learned that for CSS properties that take multiple parameters, there's no need for whitespace if the previous parameter ends with a unit.

For example:
```css
clip-path:inset(84% 0 0)
```

Can be written as:
```css
clip-path:inset(84%0 0)
```

thornbill:
I believe @dmitrylyzo tested this on webOS and reported that it was not working on some versions there.

SethFalco:
RIP. I'm a bit surprised at that, though. 🤔 
Would it be possible to get a GIF of what they saw?

dmitrylyzo:
> RIP. I'm a bit surprised at that, though. thinking Would it be possible to get a GIF of what they saw?
> 
> As in, I'd like to know more. Did it show the first frame only? Was the animation just broken? Did anything render at all?

I tested in the emulator. webOS 1/2 uses the old WebKit web engine, and there was no animation at all.
webos-1
_In the Chrome DevTools, I see the animation._

I tried changing `fill` in the keyframes, but it doesn't work as well.

SethFalco:
I'm actually having some trouble with this, so have decided to shelve it.
A bit unfortunate, but RIP.

Seems my initial approach doesn't have enough browser compatibility, and the approach with `scaleY` is yielding expected results for my locally. :thinking:

PR #3244, opened 2021-12-27
Add eslint rule for empty functions

thornbill:
**Changes**
Adds eslint rule to disallow empty functions because sonarqube does not like them

**Issues**
N/A

### DIFF
--- a/src/components/tabbedview/tabbedview.js
+++ b/src/components/tabbedview/tabbedview.js
@@ -23,10 +23,6 @@ function onViewDestroy() {
     this.initialTabIndex = null;
 }
 
-function onBeforeTabChange() {
-    // no-op
-}
-
 class TabbedView {
     constructor(view, params) {
         this.tabControllers = [];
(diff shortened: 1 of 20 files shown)


src/components/tabbedview/tabbedview.js
```diff
@@ -24,7 +24,7 @@ function onViewDestroy() {
 }
 
 function onBeforeTabChange() {
-
+    // no-op
 }
```
dmitrylyzo, line 28: Safe to remove and pass `null` - there is a check:
https://github.com/jellyfin/jellyfin-web/blob/b8d53c67db81acc492d5c2dbd6f5c6f9ff24075e/src/components/maintabsmanager.js#L157-L159

dmitrylyzo: That's why we never get those `destroy` - this is the only place where `tabControllers` is set:
https://github.com/jellyfin/jellyfin-web/blob/07bcb519c607d1701c8cb531e2b268fa567e0126/src/controllers/home.js#L61

PR #3247, opened 2021-12-29
Remove network device option from add library dialog

thornbill:
**Changes**
Removes the "Network" option from the Folder list in the add library dialog since it is no longer supported

**Issues**
N/A

### DIFF
--- a/src/components/directorybrowser/directorybrowser.js
+++ b/src/components/directorybrowser/directorybrowser.js
@@ -32,15 +32,11 @@ function refreshDirectoryBrowser(page, path, fileOptions, updatePathOnError) {
 
     const promises = [];
 
-    if (path === 'Network') {
-        promises.push(ApiClient.getNetworkDevices());
+    if (path) {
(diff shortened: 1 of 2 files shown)


src/components/directorybrowser/directorybrowser.js
```diff
@@ -61,10 +57,6 @@ function refreshDirectoryBrowser(page, path, fileOptions, updatePathOnError) {
                 html += getItem(cssClass, folder.Type, folder.Path, folder.Name);
             }
 
-            if (!path) {
-                html += getItem('lnkPath lnkDirectory', '', 'Network', globalize.translate('ButtonNetwork'));
```
cvium, line 65: What about the translation?

cvium (approved):

src/components/directorybrowser/directorybrowser.js
thornbill: Good call. I thought it might have been used in other places, but it is not. :slightly_smiling_face:

Issue #2602, opened 2021-04-21
Select the # sign


MRDJellyfin:
**Describe The Bug**

**Steps To Reproduce**

**Expected Behavior**

**Logs**

**Screenshots**

**System (please complete the following information):**
 - Platform: [e.g. Linux, Windows, iPhone, Tizen]
 - Browser: [e.g. Firefox, Chrome, Safari]
 - Jellyfin Version: [e.g. 10.6.0]

**Additional Context**

In Windows 10 when the # sign is selected in the alphabet that appears on the right in movies, which is supposed to show all movies that start with numbers like any sign such as ¿? and ¡!, does not select anything. The same happens in any of the versions: Jellyfin Tray App, Jellyfin Desktop (Shim) and Jellyfin Media Player.

perniczech:
Can confirm, filtering by '#' shows no results. 10.7.6 here.

PR #3271, opened 2022-01-04
Fix number option in alphapicker

thornbill:
**Changes**
Fixes the number option in the alphapicker for libraries

**Note:** I was unable to test movie trailer changes, because I do not have test content for them.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-web/issues/2602


crobibero (approved):
I see what you mean about copy&paste :)

PR #3281, opened 2022-01-05
Remove bump-version junk

joshuaboniface:
This didn't work, and won't matter for beta1 so just scrap it.


oddstr13 (approved):

anthonylavado (approved):
As with companion https://github.com/jellyfin/jellyfin/pull/7115, this looks safe enough.

PR #3296, opened 2022-01-05
Add "Select All" action to multiSelect

### LINKED ISSUE #895: Add "select all" to selection menu
After pressing and holding a media entry the selection menu should include "select all" so one does not have to individually select several additional entries.

image

marius-luca-87:
**Changes**

Adds a "Select all" helper to the multi-select menu.

image

**Issues**
Fixes #895

### DIFF
--- a/src/components/multiSelect/multiSelect.js
+++ b/src/components/multiSelect/multiSelect.js
@@ -176,6 +176,12 @@ import itemHelper from '../itemHelper';
                 const menuItems = [];
 
                 menuItems.push({
+                    name: globalize.translate('SelectAll'),
+                    id: 'selectall',
+                    icon: 'select_all'
+                });
+
+                menuItems.push({
                     name: globalize.translate('AddToCollection'),
                     id: 'addtocollection',
                     icon: 'add'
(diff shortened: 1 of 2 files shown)


ferferga (approved):

src/components/multiSelect/multiSelect.js
```diff
@@ -246,6 +252,21 @@ import itemHelper from '../itemHelper';
...
 
                             switch (id) {
+                                case 'selectall':
+                                    {
+                                        const elems = document.querySelectorAll('.itemSelectionPanel');
+                                        for (let i = 0, length = elems.length; i < length; i++) {
+                                            const chkItemSelect = elems[i].querySelector('.chkItemSelect');
+
+                                            if (chkItemSelect) {
+                                                if (!chkItemSelect.classList.contains('checkedInitial') && !chkItemSelect.checked) {
```
thornbill, line 262: I would combine these two if statements.

thornbill (changes requested):

thornbill:
I noticed an unusual behavior when testing this. In this screen there should be 100 items shown and selected, but for some reason there appears to be 226 selected.

all selelcted

marius-luca-87:
Apparently a few cards are left hidden and the way the current system works is by assuming you will never be able to click them.
image
The only quick way I could find would be to check for "chkItemSelect.getBoundingClientRect().width != 0" before selecting them.

thornbill (approved):
Well that is not great... it definitely shouldn't be enabling the multiselect stuff on hidden pages or inactive tabs, but that doesn't relate to your changes really. I think your work around for that is fine. I will see if I can find a fix to prevent multiselect from being enabled on the hidden views after this is merged. :sweat_smile:

PR #3301, opened 2022-01-06
Fix wizard not being translated to selected language

thornbill:
**Changes**
Fixes the wizard not being translated to the language selected by the user.

We have a number of sources for selecting the display language:

- User Language
  - Display language in user settings
- UI Language (server configuration)
  - First screen of the wizard
- Browser Language
- Fallback Language (en-us)

I'm not sure which should take precedence between the UI language and browser language. This change only uses the UI Language for the startup wizard, but we may want to consider using it in all cases.

**Issues**
N/A


thornbill:
After further discussion this probably isn't needed. The web interface will already use the user language (when selected) or the browser language. This setting in the startup wizard is really only used for translating strings server side. We should possibly add some additional context to the screen in the wizard to make that more clear.

PR #3315, opened 2022-01-11
Expose collection generation options for mixed libraries

1337joe:
**Changes**
- Pass "mixed" instead of "" for library type "Other"
  - Mixed is the expected string in the js `setContentType` function.
  - Both fall through to the [default on the server side](https://github.com/jellyfin/jellyfin/blob/307679afef158f3a9952aec40a4221fcea5f8a82/Jellyfin.Api/Controllers/LibraryController.cs#L932) but "mixed" is more descriptive
- Expose collection generation options for mixed libraries

**Issues**
Fixes jellyfin/jellyfin#7158

### DIFF
--- a/src/components/libraryoptionseditor/libraryoptionseditor.js
+++ b/src/components/libraryoptionseditor/libraryoptionseditor.js
@@ -411,7 +411,7 @@ import template from './libraryoptionseditor.template.html';
             parent.querySelector('.chkEnableEmbeddedEpisodeInfosContainer').classList.add('hide');
         }
 
-        parent.querySelector('.chkAutomaticallyAddToCollectionContainer').classList.toggle('hide', contentType !== 'movies');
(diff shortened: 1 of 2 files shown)


src/components/libraryoptionseditor/libraryoptionseditor.js
```diff
@@ -411,7 +411,11 @@ import template from './libraryoptionseditor.template.html';
             parent.querySelector('.chkEnableEmbeddedEpisodeInfosContainer').classList.add('hide');
         }
 
-        parent.querySelector('.chkAutomaticallyAddToCollectionContainer').classList.toggle('hide', contentType !== 'movies');
```
thornbill, line 414: I'm being picky here, but I think I would prefer `toggle` to be kept here and just change the condition to `contentType !== 'movies' && contentType !== 'mixed'` or `!['movies', 'mixed'].includes(contentType)`.

thornbill (approved):

PR #3318, opened 2022-01-12
Added Quit App to main menu

iainsaxonhome:
I'm using the Jellyfin Media Player in TV mode as a full screen app (open via Flex Launcher) and needed a way to close the program via the keyboard navigation. To resolve this I've added a Quit Application link below the Sign Out menu item.

Unfortunately I haven't been able to test this in the JMP yet so requires testing but the menu item exists in the menu and is hidden using the `appHost.supports()` function as per the other menu items. I'm not sure if further conditional checks to determine if this should appear (eg if its an embedded or iOS/Android app the Quit Application link should not appear) need to happen here or if that is determined in the Jellyfin Media Player.

I'm also not sure if this should be a function that requires a user confirmation (eg Are you sure you want to exit?).

It also requires translations for other languages.

Is this something that the Jellyfin Community would consider adding to the application?

**Changes**
Add a Quit Application button to the main menu.

### DIFF
--- a/src/components/apphost.js
+++ b/src/components/apphost.js
@@ -199,7 +199,6 @@ const supportedFeatures = function () {
     if (browser.operaTv || browser.tizen || browser.orsay || browser.web0s) {
         features.push('exit');
     } else {
-        features.push('exitmenu');
         features.push('plugins');
     }
 
--- a/src/controllers/user/menu/index.html
+++ b/src/controllers/user/menu/index.html
@@ -112,6 +112,15 @@
                         </div>
                     </div>
                 </a>
+
+                <a is="emby-linkbutton" data-ripple="false" href="#" style="display:block;padding:0;margin:0;" class="quitApp listItem-border">
+                    <div class="listItem">
+                        <span class="material-icons listItemIcon listItemIcon-transparent close"></span>
+                        <div class="listItemBody">
+                            <div class="listItemBodyText">${ButtonQuitApp}</div>
+                        </div>
+                    </div>
+                </a>
             </div>
         </div>
     </div>
--- a/src/controllers/user/menu/index.html
+++ b/src/controllers/user/menu/index.html
@@ -113,11 +113,11 @@
                     </div>
                 </a>
 
-                <a is="emby-linkbutton" data-ripple="false" href="#" style="display:block;padding:0;margin:0;" class="quitApp listItem-border">
+                <a is="emby-linkbutton" data-ripple="false" href="#" style="display:block;padding:0;margin:0;" class="exitApp listItem-border">
                     <div class="listItem">
                         <span class="material-icons listItemIcon listItemIcon-transparent close"></span>
                         <div class="listItemBody">
-                            <div class="listItemBodyText">${ButtonQuitApp}</div>
+                            <div class="listItemBodyText">${ButtonExitApp}</div>
                         </div>
                     </div>
                 </a>
--- a/src/controllers/user/menu/index.js
+++ b/src/controllers/user/menu/index.js
@@ -17,8 +17,8 @@ export default function (view, params) {
         window.NativeShell.openClientSettings();
     });
 
-    view.querySelector('.quitApp').addEventListener('click', function () {
-        window.NativeShell.quitApp();
+    view.querySelector('.exitApp').addEventListener('click', function () {
+        appHost.exit();
     });
 
     view.addEventListener('viewshow', function () {
@@ -37,8 +37,8 @@ export default function (view, params) {
         const supportsClientSettings = appHost.supports('clientsettings');
         page.querySelector('.clientSettings').classList.toggle('hide', !supportsClientSettings);
 
-        const supportsQuitApp = appHost.supports('quitapp');
-        page.querySelector('.quitApp').classList.toggle('hide', !supportsQuitApp);
+        const supportsExitMenu = appHost.supports('exitmenu');
+        page.querySelector('.exitApp').classList.toggle('hide', !supportsExitMenu);
 
         const supportsMultiServer = appHost.supports('multiserver');
         page.querySelector('.selectServer').classList.toggle('hide', !supportsMultiServer);
--- a/src/controllers/user/menu/index.js
+++ b/src/controllers/user/menu/index.js
@@ -17,6 +17,10 @@ export default function (view, params) {
         window.NativeShell.openClientSettings();
     });
 
+    view.querySelector('.quitApp').addEventListener('click', function () {
+        window.NativeShell.quitApp();
+    });
+
     view.addEventListener('viewshow', function () {
         // this page can also be used by admins to change user preferences from the user edit page
         const userId = params.userId || Dashboard.getCurrentUserId();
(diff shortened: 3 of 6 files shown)


iwalton3:
This feature looks good on my end from a theoretical standpoint but it needs to get tested.

dmitrylyzo:
We already have `exit` function in `appHost`/`NativeShell`.

**Here is how it is solved in the TV apps**
When the user presses `Back` on the remote control or `Escape` on the keyboard while on the Home screen, `appHost` executes `exit` function. It can be overloaded in `NativeShell`.
https://github.com/jellyfin/jellyfin-web/blob/838b8812fdd171fa53d331c826a3f9af0dd3b7e9/src/components/apphost.js#L329-L335
https://github.com/jellyfin/jellyfin-web/blob/838b8812fdd171fa53d331c826a3f9af0dd3b7e9/src/components/apphost.js#L276-L290

This approach may not fit for JMP in Desktop mode, so it is better to have an exit menu item.

iainsaxonhome:
@dmitrylyzo , thanks for pointing out the existing exit functionality. I was unaware of this code before I issued the JMP PR https://github.com/jellyfin/jellyfin-media-player/pull/196. I tried to use the long-press `Escape` via keyboard as suggested in input mapping files I found however it didn't exit the Jellyfin Media Player app and assumed there was no functionality for it yet.

I saw there was a Client Settings menu item that appeared to only be used for the various native applications where `jellyfin-web` is a base so it seemed OK to add the Quit Application link in the same way.

Adding a button to quit the app from the main menu helps because it would allow the user to exit the app (where supported) using any input that can already operate the menus such as a touch screen, gamepad, keyboard, mouse, remote or any emulated device that replicates these like a mini wireless keyboard with trackpad or an Air Mouse.

In the meantime I'll update this PR soon to use "Exit" instead of "Quit" and tie it into the existing code as highlighted in your response for consistency of language and functionality.

@iwalton3 , I'll issue an updated PR to the JMP soon to use the `window.NativeShell.AppHost.exit();` method overload approach described in this issue instead of creating a new function for `window.NativeShell.quitApp()`.

dmitrylyzo:
@iainsaxonhome I think we need to stop pushing `exitmenu` for browsers (just remove) because we can't close the window from the script.
https://github.com/jellyfin/jellyfin-web/blob/838b8812fdd171fa53d331c826a3f9af0dd3b7e9/src/components/apphost.js#L202

Or I could be wrong about usage of `exitmenu`. Yes, it isn't in use, but can be used for showing a confirmation dialog on exit. We can invent a new one though.
- `exit` - can exit/close app (currently by `Back`/`Escape` - TVs)
- `exitmenu` - add a menu entry (current PR, can imply `exit`)
- `exitprompt` - need a confirmation dialog (can imply `exit`)

At some point we need to document all these features :sweat_smile:

iainsaxonhome:
@dmitrylyzo, definitely agree with documenting these features though I'd say that's often as difficult as naming things and clearing cache :) .

Last time I looked at the code you mentioned I did wonder whether the Web project should be trying to do any of these supported feature checks and move it to the various clients that use the Web as a base however now that I look at the conditional statements again its specifically checking the browser type so that's probably OK in most cases.

If browsers can't be instructed to close a tab programmatically then you're right that it shouldn't be added to the feature list here.

I think the idea of having the 3 separate exit related features is OK as each controls a particular aspect and can work together to achieve an ideal UX for each client. 

For example the Jellyfin Media Player app would perhaps have all three enabled so the Exit App menu item appears if its running in TV display mode and full screen, would ideally ask for confirmation to exit and will exit as defined in the AppHost.

On the other hand the Android or iOS apps maybe wouldn't need the exit menu or the confirmation if the exit behaviour is to stash the current data and switch to another app or the home screen (this is a guess on my part).

I am trying to think of a client where the exit menu would be shown, can exit the app but is intended to quick-switch and not require an exit confirmation :thinking: .

dmitrylyzo:
> Last time I looked at the code you mentioned I did wonder whether the Web project should be trying to do any of these supported feature checks and move it to the various clients that use the Web as a base however now that I look at the conditional statements again its specifically checking the browser type so that's probably OK in most cases.

That code is only called for browsers (in apps we overload `appHost.supports`), so we are probably free to remove the exit part entirely.
_Also, I see no usage of the `plugins` feature and don't know its purpose._

> If browsers can't be instructed to close a tab programmatically then you're right that it shouldn't be added to the feature list here.

At the moment, the script can close the window that it has opened using `Window.open`.
https://developer.mozilla.org/en-US/docs/Web/API/Window/close

> For example the Jellyfin Media Player app would perhaps have all three enabled so the Exit App menu item appears if its running in TV display mode and full screen, would ideally ask for confirmation to exit and will exit as defined in the AppHost.

I think this can be achieved by checking `viewdata.mode` in `NativeShell.AppHost.supports` or by having a separate list of supported features for each mode.

> On the other hand the Android or iOS apps maybe wouldn't need the exit menu or the confirmation if the exit behaviour is to stash the current data and switch to another app or the home screen (this is a guess on my part).

Agree, but for some reason iOS has `exit` and `exitmenu` features: https://github.com/jellyfin/jellyfin-expo/blob/3c0d6639d5265e31d322d227c9ac4be27a70a4c9/assets/js/NativeShell.staticjs#L10-L11

Android has only `exit` (back action exit): https://github.com/jellyfin/jellyfin-android/blob/e04d6dded3a38aedcd4014f923340acddc76c9e4/app/src/main/assets/native/nativeshell.js#L7

Probably, the exit action on mobile doesn't actually close the app, but switches to the previous activity.

> I am trying to think of a client where the exit menu would be shown, can exit the app but is intended to quick-switch and not require an exit confirmation.

I would remove that `exitmenu` from the browser (not to bother users) and the feature is ready.
Currently, confirmation is only shown for Tizen (btw, it also has `exitmenu` - I just copied it from an old Android app).
https://github.com/jellyfin/jellyfin-web/blob/afd37d62877a765a2284ae55821f0dff35062080/src/components/apphost.js#L330-L332
We can add `exitprompt` later, but it would definitely be nice to have a roadmap for "exit" behavior.
In general, on TV exit is activated by the back action. Samsung recommends asking the user to confirm the exit. On webOS we show the home screen, so that the user can return back to the app (using the back button again).

If we want the "menu exit" be instant and the back action exit with the confirmation, we can add a param (say `needConfirm` or `isBack`), which `false` by default, to `AppHost.exit` and pass `true` at the `inputManager`.
https://github.com/jellyfin/jellyfin-web/blob/afd37d62877a765a2284ae55821f0dff35062080/src/scripts/inputManager.js#L117

dmitrylyzo:
@iainsaxonhome You forgot to handle click in the navDrawer menu (Library entries are handled automatically).
https://github.com/jellyfin/jellyfin-web/blob/afd37d62877a765a2284ae55821f0dff35062080/src/scripts/libraryMenu.js#L325
```js
        const btnExit = navDrawerScrollContainer.querySelector('.exitApp');
        if (btnExit) {
            btnExit.addEventListener('click', () => {
                appHost.exit();
            });
        }
```

iainsaxonhome:
@dmitrylyzo , thanks for spotting that, I was testing in the TV display mode and had missed that addition to the library menu.

Its a little more verbose but I opted to add the event handling the same way as the existing buttons ;) .

dmitrylyzo (approved):

thornbill:
I want to test this with the iOS app before it is merged.

thornbill:
I have logged an issue in the expo repo to remove this as a supported feature and noted that it will require a new app release for the 10.8 release.

https://github.com/jellyfin/jellyfin-expo/issues/340

PR #3340, opened 2022-01-15
Fix auto focus on Play/Resume

### LINKED ISSUE #3320: [LG webOS] Play Button in Hover State when Exiting Playback
**Describe The Bug**
When testing webOS 6 (see https://github.com/jellyfin/jellyfin-webos/issues/69), I noticed that when returning from a video that was now resumable, the play button would still have its hover state on. After moving the selection cursor left or right from the "start from beginning" button, the highlighting states are correct again. 

**Steps To Reproduce**
Though this happened on webOS, it would likely reproduce on the TV layout on desktop.

1. Play any item, and watch king enough so that it is resumable
2. Press "back" (I suppose clicking the back arrow at the top left works too)
3. See error as per screenshot.

**Expected Behavior**
Only one button should have any kind of highlight or hover state active, and it would best if it was the resume button, not the restart playback button.

**Logs**
N/A

**Screenshots**
image

 - Platform: Linux
 - Browser: webOS app
 - Jellyfin Version: 10.8.0-alpha5

**Additional Context**
There was work done to fix both buttons being highlighted blue, which was in https://github.com/jellyfin/jellyfin-web/pull/2799.

dmitrylyzo:
After stopping video playback we get the item state via WebSocket, so we need to refocus the "most valid" button in the WebSocket handler.

**Changes**
- Add `Resume` as a candidate if `Play` was focused.
- Add `autoFocus` call to WebSocket handler.
- Make ~`fab`~ class mutually exclusive. `fab` is changed to `raised`.

**Issues**
Fixes #3320

### DIFF
--- a/src/components/autoFocuser.js
+++ b/src/components/autoFocuser.js
@@ -59,6 +59,9 @@ import layoutManager from './layoutManager';
                 candidates.push(container.querySelector('.btnPreviousPage'));
             } else if (activeElement.classList.contains('btnSelectView')) {
                 candidates.push(container.querySelector('.btnSelectView'));
+            } else if (activeElement.classList.contains('btnPlay')) {
+                // Resume has priority over Play
+                candidates = candidates.concat(Array.from(container.querySelectorAll('.btnResume')));
             }
 
             candidates.push(activeElement);
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -374,12 +374,12 @@ function reloadPlayButtons(page, item) {
         hideAll(page, 'btnShuffle');
     }
 
-    const btnResume = page.querySelector('.mainDetailButtons .btnResume');
-    const btnPlay = page.querySelector('.mainDetailButtons .btnPlay');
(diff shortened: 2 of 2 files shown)


dmitrylyzo:
Need to fix `fab` class.

dmitrylyzo:
Made the `fab` class mutually exclusive. I believe it was intended that way.

Before:

After:


But the round button in a row of square ones looks a bit odd.

thornbill:
Could we remove the round style of that button? I agree it looks very out of place there.

dmitrylyzo:
> Could we remove the round style of that button? I agree it looks very out of place there.

This simply means to remove these lines:
https://github.com/jellyfin/jellyfin-web/blob/cf55e3341dac05082e46f47eb203c6caa705732b/src/controllers/itemDetails/index.js#L377-L383

- first appearance: #1845
- refactored: #2799 (removed blue highlighting)

thornbill:
I believe we can remove the round styling. If we want to change the shape of the active button, that can be done separately.

thornbill (approved):

PR #3342, opened 2022-01-16
LiveTV Channels paging

dmitrylyzo:
**Changes**
- Add a second paging block on the Channels tab
- Fix autofocus and auto scroll top


thornbill:
@dmitrylyzo what should I be testing here? I don't seem to have enough channels to need pagination, but I don't recall what the other fix was for. :sweat_smile:

dmitrylyzo:
> @dmitrylyzo what should I be testing here? I don't seem to have enough channels to need pagination, but I don't recall what the other fix was for. sweat_smile

Even with `Library page size: 5/10`?

https://github.com/jellyfin/jellyfin-web/blob/827c722ae2b05a68ac23668e36634258dcb4503f/src/controllers/livetv/livetvchannels.js#L127
This limits the focus scope to the current tab, so that the channel card is selected the first time you open it. _I will rephrase the commit._

thornbill (approved):

PR #3343, opened 2022-01-16
Limit transcoding profiles with maximum resolution

dmitrylyzo:
_Save some CPU cycles and network bandwidth when transcoding_ :crossed_fingers:

**Changes**
- Limit transcoding profiles with maximum resolution
- Limit quality settings with maximum resolution
~:warning: If the resolution is in between FHD and UHD, the `4K` options are disabled.~
`Allow next higher resolution`

:warning: Google Cast needs testing.

_Requires https://github.com/jellyfin/jellyfin/pull/7197 and https://github.com/jellyfin/jellyfin/pull/7198_

### DIFF
--- a/src/components/apphost.js
+++ b/src/components/apphost.js
@@ -39,6 +39,25 @@ function getDeviceProfile(item) {
             profile = profileBuilder(builderOpts);
         }
 
+        const maxTranscodingVideoWidth = appHost.screen()?.maxAllowedWidth;
+
+        if (maxTranscodingVideoWidth) {
+            profile.TranscodingProfiles.forEach((transcodingProfile) => {
+                if (transcodingProfile.Type === 'Video') {
+                    transcodingProfile.Conditions = (transcodingProfile.Conditions || []).filter((condition) => {
+                        return condition.Property !== 'Width';
+                    });
+
+                    transcodingProfile.Conditions.push({
+                        Condition: 'LessThanEqual',
+                        Property: 'Width',
+                        Value: maxTranscodingVideoWidth.toString(),
+                        IsRequired: false
+                    });
+                }
+            });
+        }
+
         resolve(profile);
     });
 }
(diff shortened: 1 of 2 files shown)


dmitrylyzo:
The current implementation requires a dependency loop (`appHost -> browserDeviceProfile -> appHost`) to enable the limit in apps (except TVs).

Alternatively, we can post-process the profile: https://github.com/dmitrylyzo/jellyfin-web/commit/e5022000ecd8fc664d65fe02f5747a7b2bcf1de5
As a bonus, this will allow to add the limit to a profile that was not generated by `browserDeviceProfile`.

thornbill:
@dmitrylyzo Could the max width be passed to the profile builder as an "option"? https://github.com/jellyfin/jellyfin-web/blob/master/src/components/apphost.js#L38-L39

thornbill:
👍 I think it's reasonable to expect clients that provide their own profiles to also be responsible for handling the resolution limits. In theory they should be able to determine resolution more accurately anyway.

dmitrylyzo:
> In theory they should be able to determine resolution more accurately anyway.

There is a new `AppHost.screen` function for this. It is used for the profile and the list of quality options.

As for options, I think we should leave the next higher resolution, i.e. if the screen width greater than 1930 (+10 for sure) add 4K options.

We also need to support portrait oriented videos, but we currently lock the screen in landscape.

Shadowghost:
Why exactly do we allow the next bigger resolution too? IMHO exactly that should not be done because afaik it leads to `Auto` choosing 4K when transcoding is required on a local network without bandwith restriction.

dmitrylyzo:
> Why exactly do we allow the next bigger resolution too? IMHO exactly that should not be done because afaik it leads to `Auto` choosing 4K when transcoding is required on a local network without bandwith restriction.

Quality options don't affect the transcoding resolution limit. The next bigger resolution is for:
> If the resolution is in between FHD and UHD, the 4K options are disabled.

For example, you have 2K display, but only 1080p is allowed.

dmitrylyzo:
Personally, I lean toward the post-processing version:
- to eliminate the dependency loop.
- to re-use the `AppHost.screen` function instead of duplicating it (and the limiting logic) in apps.
- to add a resolution limit to webui apps "out of the box".
- to fix/trim/post-process the profiles even if they were not generated by `browserDeviceProfile`.

joshuaboniface (approved):

PR #3344, opened 2022-01-16
Try to determine the number of speakers before falling back to hardcoded values

knackebrot:
This is split from my old PR: https://github.com/jellyfin/jellyfin-web/pull/2511

**Changes**
If the user hasn't specified his desired number of audio channels, try to determine it from the browser. This effectively reduces the number of audio channels from 6 to 2 for most users when transcoding.

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -294,12 +294,55 @@ import browser from './browser';
                     (browser.tizen && isTizenFhd ? 20000000 : null)));
     }
 
+    function getSpeakerCount() {
+        const AudioContext = window.AudioContext || window.webkitAudioContext || false; /* eslint-disable-line compat/compat */
+
+        if (AudioContext) {
+            const audioCtx = new AudioContext();
+
+            return audioCtx.destination.maxChannelCount;
+        }
+
+        return -1;
+    }
+
+    function getPhysicalAudioChannels(options) {
+        const allowedAudioChannels = parseInt(userSettings.allowedAudioChannels(), 10);
+
+        if (allowedAudioChannels > 0) {
+            return allowedAudioChannels;
+        }
+
+        if (options.audioChannels) {
+            return options.audioChannels;
+        }
+
+        const isSurroundSoundSupportedBrowser = browser.safari || browser.chrome || browser.edgeChromium || browser.firefox || browser.tv || browser.ps4 || browser.xboxOne;
+        const speakerCount = getSpeakerCount();
+
+        if (speakerCount > 2) {
+            if (isSurroundSoundSupportedBrowser) {
+                return speakerCount;
+            }
+
+            return 2;
+        }
+
+        if (speakerCount > 0) {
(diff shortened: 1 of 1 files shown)


src/scripts/browserDeviceProfile.js
```diff
@@ -294,12 +294,55 @@ import browser from './browser';
...
+            const audioCtx = new AudioContext();
+
+            return audioCtx.destination.maxChannelCount;
+        }
+
+        return -1;
+    }
+
+    function getPhysicalAudioChannels(options) {
+        const allowedAudioChannels = parseInt(userSettings.allowedAudioChannels() || '-1');
```
dmitrylyzo, line 310: `allowedAudioChannels` has `-1` fallback value and use an explicit radix to be sure.
```suggestion
        const allowedAudioChannels = parseInt(userSettings.allowedAudioChannels(), 10);
```

knackebrot: done, added radix

dmitrylyzo:
@knackebrot Please remove ` || '-1'` there - as I said, `allowedAudioChannels` has it inside.
https://github.com/jellyfin/jellyfin-web/blob/b0a05933805ea31d47d6e68196827babd5b8abdb/src/scripts/settings/userSettings.js#L127

knackebrot:
> @knackebrot Please remove ` || '-1'` at the same line - as I said, `allowedAudioChannels` has it inside.
> 
> https://github.com/jellyfin/jellyfin-web/blob/b0a05933805ea31d47d6e68196827babd5b8abdb/src/scripts/settings/userSettings.js#L127

sorry, fixed now.

dmitrylyzo (approved):
LGTM

src/scripts/browserDeviceProfile.js
```diff
@@ -294,12 +294,55 @@ import browser from './browser';
                     (browser.tizen && isTizenFhd ? 20000000 : null)));
     }
 
+    function getSpeakerCount() {
+        const AudioContext = window.AudioContext || window.webkitAudioContext || false; /* eslint-disable-line compat/compat */
+
+        if (AudioContext) {
+            const audioCtx = new AudioContext();
+
+            return audioCtx.destination.maxChannelCount;
```
dmitrylyzo, line 303: Testing this on the webOS 1.2 emulator, I got `maxChannelCount: 0` and `channelCount: 2`.
I'm not sure if adding ` || audioCtx.destination.channelCount` is right.
It is possible that this is an emulator bug, so it is probably better to just leave it for now.

knackebrot: I believe this should not ever happen, according to [docs](https://webaudio.github.io/web-audio-api/#dom-audionode-channelcount):  
> ...some nodes have additional channelCount constraints on the possible values for the channel count:
> ... AudioContext
> The channel count MUST be between 1 and maxChannelCount. An IndexSizeError exception MUST be thrown for any attempt to set the count outside this range.

Issue #2582, opened 2021-04-11
Book Position Not Remembered


shortspider:
**Describe The Bug**
When I open a book and scroll through the position isn't remembered. When the book is opened again it is at the start of the book.

**Media Information**
N/A

**Screenshots**

**System (please complete the following information):**
 - Platform: Linux, Android
 - Browser: Firefox, Android app
 - Jellyfin Version: [e.g. 10.7.1]

**Additional Context**
So the book player plugin seems to have support for restoring the position:
![image](https://user-images.githubusercontent.com/5712543/114310119-c0d7e880-9ab7-11eb-8452-46950a637d2e.png)
The problem is that `startPositionTicks` is never populated. I'm not sure if this is a web issue or there just isn't support for it on the server side (both maybe?). If this is a server side issue please let me know and I'll open an issue there. Willing to work on this just need a bit of guidance on what exactly needs to be done.

shortspider:
Digging into this a bit more the books are calling `Sessions/Playing/Progress` and sending the current position in the book. This is tested on 10.7.2.

However, if you open a book and move a few pages in, wait for the Progress call to be made and complete, close the book, and then open it again it doesn't start from the position where you left off at. If you repeat the above and refresh the page it still doesn't work, you need to let some time go by before you will get the correct position on open.

I'm not sure if this is a caching issue on the client or something on the server at this point.

Also, is anyone else seeing this or just me?

shortspider:
Just to add to my last comment, here you can see that a `Progress` call is made with the position at not 0:
![image](https://user-images.githubusercontent.com/5712543/115645198-ef429880-a2ed-11eb-960f-10e2094b266c.png)
But when we reload the book the position is still 0:
![image](https://user-images.githubusercontent.com/5712543/115645214-f8cc0080-a2ed-11eb-97bb-4e3fea3afbb1.png)

tariq7t:
I can confirm the same issue is happening to me where the book position is not being remembered.

MinecraftPlaye:
I can confirm that this issue is still present with the following server/clients:

    Platform: Linux, Windows, iPadOS
    Browser: Firefox
    Client: [Jellyfin Media Player 1.6.1](https://github.com/jellyfin/jellyfin-media-player/releases/tag/v1.6.1), Jellyfin Mobile (iOS) v1.4.1 (5.4.1.0) with Expo v39.0.0
    Jellyfin Server Version: 10.7.6

[...]

PR #3034, opened 2021-10-04
Save the progress when reading a Comic Book Archive

MinecraftPlaye:
**Changes**
Save the progress when reading a Comic Book Archive. Thx to the community for the helping me to get this working. :)

Tested against the following Jellyfin Server version: https://github.com/jellyfin/jellyfin/commit/9d46b3695f7cf291f67a542dfa648b0b3e80824e
Does not work with Jellyfin Server 10.7.7

PR #3349, opened 2022-01-18
Save reading progress for pdfs for each page

MinecraftPlaye:
**Changes**
This removes the 10s interval between saving and instead saves the
reading progress for pdf files every time.

As a side effect, the pdf will appear in the "Continue Watching" section
of Jellyfin. Before this commit, there was no indication that someone
started reading the pdf book.

**Issues**
Potentially related to  #2582, but for pdfs only

### DIFF
--- a/src/plugins/pdfPlayer/plugin.js
+++ b/src/plugins/pdfPlayer/plugin.js
@@ -36,6 +36,12 @@ export class PdfPlayer {
     stop() {
         this.unbindEvents();
 
+        const stopInfo = {
+            src: this.item
+        };
+
+        Events.trigger(this, 'stopped', [stopInfo]);
+
         const elem = this.mediaElement;
         if (elem) {
             dialogHelper.close(elem);
@@ -47,6 +53,10 @@ export class PdfPlayer {
 
         // cancel page render
         this.cancellationToken = true;
+    }
+
+    destroy() {
+        // Nothing to do here
     }
 
     currentItem() {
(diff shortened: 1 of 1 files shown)


src/plugins/pdfPlayer/plugin.js
```diff
@@ -265,7 +265,7 @@ export class PdfPlayer {
 
     renderPage(canvas, number) {
         this.book.getPage(number).then(page => {
-            Events.trigger(this, 'timeupdate');
+            Events.trigger(this, 'pause');
```
dmitrylyzo, line 268: I tested the rewind to the first page and found that the event triggering is placed completely wrong.
`renderPage` is called when the page needs to be rendered. It isn't called for some number of previous pages (cached).
For now, I thing `Events.trigger` should be at the end of the `next` and `previous` functions.

dmitrylyzo (commented):
To update the resume button after stopping, do the same as #3034
`stopped` event, `destroy` function and `streamInfo.item`.

As for the `Watched` state, it requires changes on the server side. It currently doesn't extract the page count (`RunTimeTicks` is `TimeSpan.TicksPerSecond: 10000000`), and need to update `UserDataManager.UpdatePlayState` method.
https://github.com/jellyfin/jellyfin/blob/8b36bc0ade036557769270f67c402ce94795b04b/Emby.Server.Implementations/Library/UserDataManager.cs#L216

dmitrylyzo:
Looks good. Could you squash the commits?

MinecraftPlaye:
> To update the resume button after stopping, do the same as #3034: `stopped` event, `destroy` function and `streamInfo.item`.

As I've noticed that the bookPlayer is missing those calls, too, I just created PR #3350 to solve it there.

> Looks good. Could you squash the commits?

Sure.

dmitrylyzo (approved):

PR #3350, opened 2022-01-19
 Update the resume button after stopping

MinecraftPlaye:
Update the resume button after stopping for the book player.

### DIFF
--- a/src/plugins/bookPlayer/plugin.js
+++ b/src/plugins/bookPlayer/plugin.js
@@ -44,6 +44,12 @@ export class BookPlayer {
     stop() {
         this.unbindEvents();
 
+        const stopInfo = {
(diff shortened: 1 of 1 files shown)


dmitrylyzo:
Personally, I see no reason not to change `timeupdate` to `pause`, because the "blank page" issue doesn't relate to it. But this change will help keep progress more accurate.
https://github.com/jellyfin/jellyfin-web/blob/22d6c99f44427658b5ef34efc883c6c27b2c8370/src/plugins/bookPlayer/plugin.js#L315

There is also a problem with calculating of progress from/to book location, but I don't know how to fix that. The problem with the blank page may be related to this.

dmitrylyzo (approved):

PR #3355, opened 2022-01-20
Add Markdown to Login Disclaimer

### LINKED ISSUE #3304: Login disclaimer: Add HTML support
I am trying to use the "Login Disclaimer" to show the links of my 2 web clients (the original one, and the one written in Vue) but I noticed that the text is inserted as such and not as HTML.

This makes it impossible to have a working link (or even just a \<br> tag).

image

I think it would be nice to add support for this type of interactive Login Disclaimer

--------------------------

If you don't like the idea due to script injection, I would propose a separate property in branding to be able to add a script.

In my case I could solve it by simply running:

```$('#loginPage .disclaimer').html($('#loginPage .disclaimer').text())```

**_What do you think?_**

dmitrylyzo:
**Changes**
Add Markdown to Login Disclaimer.

**Issues**
Fixes #3304

`DOMPurify` doesn't seem to work in webOS 1.2 - returns an empty string. :confused: 
`DOMParser.parseFromString` returns `null`.
We may need a polyfill for `text/html`: https://gist.github.com/eligrey/1129031
_I tested - it works._
~A solution from https://github.com/cure53/DOMPurify/issues/625#issuecomment-1019309574 is used.~
Doesn't work in Firefox 97 - https://github.com/jellyfin/jellyfin-web/pull/3355#issuecomment-1046534977
Polyfill with https://gist.github.com/eligrey/1129031

~:question: 
I removed `text-align: center` from `disclaimer`, so users have to align the text themselves.
Or should I keep it, assuming that the text should always be centered?~

### DIFF
--- a/src/controllers/dashboard/general.html
+++ b/src/controllers/dashboard/general.html
@@ -65,7 +65,7 @@
                 <div class="verticalSection">
                     <h2>${HeaderBranding}</h2>
                     <div class="inputContainer">
-                        <input is="emby-input" type="text" id="txtLoginDisclaimer" label="${LabelLoginDisclaimer}" />
+                        <textarea is="emby-textarea" id="txtLoginDisclaimer" label="${LabelLoginDisclaimer}" class="textarea-mono"></textarea>
                         <div class="fieldDescription">${LabelLoginDisclaimerHelp}</div>
                     </div>
                     <div class="inputContainer customCssContainer">
--- a/src/controllers/session/login/index.html
+++ b/src/controllers/session/login/index.html
@@ -50,7 +50,9 @@
                 <span>${ButtonChangeServer}</span>
             </button>
 
-            <p class="disclaimer" style="text-align: center; margin-top: 2em;"></p>
+            <div class="disclaimerContainer">
+                <div class="disclaimer"></div>
+            </div>
         </div>
     </div>
 </div>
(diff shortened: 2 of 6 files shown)


thornbill:
I think we should keep it centered, but I don't feel particularly strongly about it either way. :man_shrugging:

thornbill (changes requested):
Changes look fine and work as expected. I did notice list items look weird due to the `text-align: center` (see screenshot). We should probably center the container div and limit the width so it fits the content.

Screenshot 2022-02-21 at 00-40-34 Jellyfin

dmitrylyzo:
Yes, I noticed that in Chrome.
In Firefox, it's... different.
firefox-1

dmitrylyzo:
chrome-2

But the webOS 1.2 fix doesn't work in Firefox 97 - it requires `<br>` to have `</br>`. So we need to fix this other way.

dmitrylyzo:
I've added https://gist.github.com/eligrey/1129031, but I had to trim it according to our lint rules.
_If this version is fine, I will squash the commits._

thornbill (approved):
:+1: looks good to me now! I don't think it is too big of a deal if `fit-content` is unsupported since it should just fallback to being full width.

dmitrylyzo:
> +1 looks good to me now! I don't think it is too big of a deal if `fit-content` is unsupported since it should just fallback to being full width.

I tested this - webOS 1.2 supports `fit-content` with the `-webkit-` prefix (`autoprefixer` did the job).

PR #3356, opened 2022-01-20
Special Features: Display extraType as secondaryText on card

SenorSmartyPants:
**Changes**

Special features will now display a translated ExtraType description in card secondaryText. Translation strings added to en-us for all ExtraTypes except Unknown

image

### DIFF
--- a/src/strings/en-us.json
+++ b/src/strings/en-us.json
@@ -1607,5 +1607,14 @@
     "UnknownVideoStreamInfo": "The video stream info is unknown",
     "UnknownAudioStreamInfo": "The audio stream info is unknown",
     "DirectPlayError": "There was an error starting direct playback",
-    "SelectAll": "Select All"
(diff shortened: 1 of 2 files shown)


src/strings/en-us.json
```diff
@@ -1607,5 +1607,14 @@
     "UnknownVideoStreamInfo": "The video stream info is unknown",
     "UnknownAudioStreamInfo": "The audio stream info is unknown",
     "DirectPlayError": "There was an error starting direct playback",
-    "SelectAll": "Select All"
+    "SelectAll": "Select All",
+    "Clip": "Featurette",
+    "Trailer": "Trailer",
+    "BehindTheScenes": "Behind the scenes",
```
thornbill, line 1613: I think title case might work better for these.

SenorSmartyPants: Pushed.

PR #3361, opened 2022-01-22
Only use comic pages for internal calculations and ignore metadata files

### LINKED ISSUE #3352: Comic Player uses metadata and other non-page files for reading position calculation
**Describe The Bug**
The comic player starts at an non-existing page.

**Steps To Reproduce**
1. Go to a cbz comic which contains 1 additional non-page file, e.g. a ComicInfo.xml and open it.
2. Open a cbz file that does only contain pages.

Both cases started at the wrong position. [When changing the line](https://github.com/jellyfin/jellyfin-web/blob/bc18ba9885c7e9d004874afed42283ce2972b356/src/plugins/comicsPlayer/plugin.js#L186) to the line below:
`this.currentPage = options.startPositionTicks / 10000 || 1;`
Now the first case starts at page 0 (the cover). The 2nd case starts at the wrong page. However, adding an additional non-page file like the ComicInfo.xml to the 2nd comic and it opens at page 0 (the cover). 

Thus I conclude that the current comicplayer takes into consideration all files within the cbz archive instead of pages only.

**Expected Behavior**
The comic player opens only pages and only uses them to calculate the starting and save position.

Platform: Manjaro Linux
Browser: Firefox 96.0
Jellyfin Version: 10.8.0 Alpha 5

MinecraftPlaye:
**Changes**
A comic book archive stores pages as images, but can store other
information like metadata within the file, too.

This change uses only the image based files as pages and ignores other
files.

The most commonly used image formats seems to be JPEG and PNG, but other formats gif, tiff etc. are sometimes used, too. :O Technically, a comic book archive can use any picture formats as pages.

Thx to @dmitrylyzo for his help with JS.

**Issues**
Fixes #3352

### DIFF
--- a/src/plugins/comicsPlayer/plugin.js
+++ b/src/plugins/comicsPlayer/plugin.js
@@ -259,7 +259,15 @@ class ArchiveSource {
         this.raw = await this.archive.getFilesArray();
         await this.archive.extractFiles();
 
-        const files = await this.archive.getFilesArray();
+        // the comic book archive supports any kind of image format as it's just a zip archive
+        const supportedFormats = ['jpg', 'png', 'avif', 'gif', 'bmp', 'dib', 'tiff', 'tif'];
+
+        let files = await this.archive.getFilesArray();
+        files = files.filter((file) => {
+            const name = file.file.name;
+            const index = name.lastIndexOf('.');
+            return supportedFormats.includes(name.substr(index + 1));
+        });
         files.sort((a, b) => {
             if (a.file.name < b.file.name) {
                 return -1;


src/plugins/comicsPlayer/plugin.js
```diff
@@ -259,7 +259,15 @@ class ArchiveSource {
...
 
-        const files = await this.archive.getFilesArray();
+        // the comic book archive supports any kind of image format as it's just a zip archive
+        const supportedFormats = ['jpg', 'png', 'avif', 'gif', 'bmp', 'dib', 'tiff', 'tif'];
+
+        let files = await this.archive.getFilesArray();
+        files = files.filter((file) => {
+            const name = file.file.name;
+            const index = name.lastIndexOf('.');
+            return supportedFormats.includes(name.substr(index + 1));
```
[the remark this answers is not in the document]
MinecraftPlaye: Nope, we do not need to test for files that do not have a file extension.

src/plugins/comicsPlayer/plugin.js
```diff
@@ -241,6 +241,9 @@ export class ComicsPlayer {
     }
 }
 
+// the comic book archive supports any kind of image format as it's just a zip archive
+const supportedFormats = ['jpg', 'png', 'avif', 'gif', 'bmp', 'dib', 'tiff', 'tif'];
```
thornbill, line 245: What about "jpeg"?

MinecraftPlaye: > What about "jpeg"?

I just checked the file extensions on Wikipedia and it seems like `jpg` and `jpeg` are not the only possible file extensions for a jpeg image. :O
Anyway,  I just added the entire list of possible file extensions for jpeg to the list. ;)

src/plugins/comicsPlayer/plugin.js
```diff
@@ -259,7 +262,12 @@ class ArchiveSource {
         this.raw = await this.archive.getFilesArray();
         await this.archive.extractFiles();
 
-        const files = await this.archive.getFilesArray();
+        let files = await this.archive.getFilesArray();
+        files = files.filter((file) => {
+            const name = file.file.name;
+            const index = name.lastIndexOf('.');
+            return index !== 1 && supportedFormats.includes(name.slice(index + 1));
```
thornbill, line 269: Why is this checking `index !== 1`?

dmitrylyzo: To ignore files like `jpg`, `png` - `slice(-1 + 1)`

thornbill: Shouldn't it be `-1`?

dmitrylyzo: Oh, yes, it should be `-1` as in the original version https://github.com/jellyfin/jellyfin-web/issues/3352#issuecomment-1017757810

thornbill (approved):

PR #3369, opened 2022-01-29
Escape HTML

### LINKED ISSUE #3368: [Issue]: HTML not being sanitized
### Please describe your bug

The html in the names of folders gets put directly onto the webpage, which screws it up

### Jellyfin Version

10.7.7

### if other:

_No response_

### Environment

```markdown
- OS: Artix 5.15
- Virtualization: No
- Clients: Web
- Browser: All
- FFmpeg Version: /
- Playback Method: /
- Hardware Acceleration: /
- Plugins: / 
- Reverse Proxy: No
- Base URL: no
- Networking: /
- Storage: HDD
```

### Jellyfin logs

```shell
not relevant
```

### FFmpeg logs

```shell
not relevant
```

/

untitled

Image censored for you sanity 

### Code of Conduct

dmitrylyzo:
**Changes**
Escape HTML
_I have not tested some of them._
_Home, Libraries, Item details, Libraries+options in Dashboard, MediaInfo are visually fine._

**Issues**
Fixes #3368

`Catch 'Em All` game :sweat_smile: 

#3368 - due to HTML in the `data-path` attribute of the card.

Not sure if we need to escape names (other data) of these:
- LiveTV
- Tuners
- ScheduledTasks
- Notifications
- Users
- TOC in `bookPlayer`
- Sort menu
- Plugins


package.json
```diff
@@ -75,6 +76,7 @@
     "date-fns": "2.28.0",
     "dompurify": "2.3.4",
     "epubjs": "0.3.90",
+    "escape-html": "1.0.3",
```
thornbill, line 79: This is such a small utility maybe it's worth including it locally in jellyfin-web to avoid a "leftpad" situation?

dmitrylyzo: Do you mean to place https://github.com/component/escape-html/blob/master/index.js somewhere in `libraries`?
If there are no (license) problems to us.
_This utility was not updated for a while (it is perfect :stuck_out_tongue: )._

I am also thinking of replacing `htmlescape` with `escapeHtml` as it called in the utility. What do you think?

_offtopic_:
Is this the "leftpad" you're talking about?
https://github.com/left-pad/left-pad/blob/master/index.js

thornbill: Ah good call if we are already pulling it in for other dependencies it probably doesn't make sense to include it directly. I do think `escapeHtml` is better since it better denotes an action is being done.

_offtopic:_
I was referencing this incident where the leftpad package was nuked by the maintainer. :smile:  https://www.theregister.com/2016/03/23/npm_left_pad_chaos/

src/components/listview/listview.js
```diff
@@ -148,11 +149,7 @@ import ServerConnections from '../ServerConnections';
 
             elem.classList.add('listItemBodyText');
 
-            if (textlines[i]) {
-                elem.innerText = textlines[i];
-            } else {
-                elem.innerHTML = '&nbsp;';
-            }
+            elem.innerText = text;
```
thornbill, line 152: Should this be `text || '&nbsp;`?

dmitrylyzo: This is unnecessary because `text` is defined and non-empty:
https://github.com/jellyfin/jellyfin-web/blob/8420d0fef20851d259df891289bebaf5ea342a7c/src/components/listview/listview.js#L133-L135

thornbill (approved):

PR #3376, opened 2022-02-02
Add option to season image downloader to pick from series images

SenorSmartyPants:
**Changes**

When searching for images for a TV Season, new option to display series level images. 

I have several series where there aren't many options at the season level for images. But there are season specific images (based on cast changes, etc) tagged at the series level at image providers. Now users can check the box and pick from those images. Does not affect how automatic image downloads work.

image

### DIFF
--- a/src/components/imageDownloader/imageDownloader.js
+++ b/src/components/imageDownloader/imageDownloader.js
@@ -31,11 +31,16 @@ import template from './imageDownloader.template.html';
     let browsableImageStartIndex = 0;
     let browsableImageType = 'Primary';
     let selectedProvider;
+    let browsableParentId;
 
-    function getBaseRemoteOptions() {
+    function getBaseRemoteOptions(page) {
         const options = {};
 
-        options.itemId = currentItemId;
+        if (page.querySelector('#chkShowParentImages').checked && browsableParentId) {
+            options.itemId = browsableParentId;
+        } else {
+            options.itemId = currentItemId;
+        }
 
         return options;
     }
@@ -43,7 +48,7 @@ import template from './imageDownloader.template.html';
     function reloadBrowsableImages(page, apiClient) {
         loading.show();
 
-        const options = getBaseRemoteOptions();
+        const options = getBaseRemoteOptions(page);
 
         options.type = browsableImageType;
         options.startIndex = browsableImageStartIndex;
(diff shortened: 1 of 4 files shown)


src/components/imageDownloader/imageDownloader.js
```diff
@@ -295,6 +300,12 @@ import template from './imageDownloader.template.html';
             reloadBrowsableImages(page, apiClient);
         });
 
+        page.querySelector('#chkShowParentImages').addEventListener('change', function () {
+            browsableImageStartIndex = 0;
```
thornbill, line 304: SonarQube picked up that there is a lot of duplication in these event handlers. It would be best to factor the common parts out to a single function.

src/components/imageeditor/imageeditor.js
```diff
@@ -280,7 +280,7 @@ import template from './imageeditor.template.html';
 
     function showImageDownloader(page, imageType) {
         import('../imageDownloader/imageDownloader').then((ImageDownloader) => {
-            ImageDownloader.show(currentItem.Id, currentItem.ServerId, currentItem.Type, imageType).then(function () {
+            ImageDownloader.show(currentItem.Id, currentItem.ServerId, currentItem.Type, imageType, currentItem.Type == 'Season' ? currentItem.ParentId : null).then(function () {
```
thornbill, line 283: Very minor, but this line is rather long now.

src/components/imageDownloader/imageDownloader.js
```diff
@@ -31,19 +31,24 @@ import template from './imageDownloader.template.html';
...
     let browsableImageType = 'Primary';
     let selectedProvider;
+    let browsableParentId;
 
-    function getBaseRemoteOptions() {
+    function getBaseRemoteOptions(page) {
         const options = {};
 
-        options.itemId = currentItemId;
+        if (page.querySelector('#chkShowParentImages').checked && browsableParentId != null) {
```
thornbill, line 39: The `browsableParentId != null` comparison seems a bit odd here since `browsableParentId` will likely be undefined rather than null in most cases. It works but only because it is not using the strict equality comparison. I think just using `&& browsableParentId` as a check would be better to avoid this ambiguity.

src/components/imageDownloader/imageDownloader.js
```diff
@@ -336,6 +347,10 @@ import template from './imageDownloader.template.html';
             scrollHelper.centerFocus.on(dlg, false);
         }
 
+        if (browsableParentId != null) {
```
thornbill, line 350: Same case here as above.

thornbill (changes requested):

SenorSmartyPants:
Suggestion made. Commits look a little weird, and gitkraken isn't letting me squash atm.

SenorSmartyPants:
Fixed commits, only one now with all changes.

thornbill (approved):

PR #3381, opened 2022-02-04
Fix queue when playing on remote device (partial)

AER00:
**Changes**
Many queue specific features was missing when using remote play (casting). This commit enables showing queue from remote device, and implements (really) basic queue management. There is still a lot of things to improve.

**Issues**
Unfortunately, it's currently not possible to manage queue fully remotely without implementing some new endpoints in the API, such as go to nth item of current playlist, or move / remove an item without stopping current playback.


danditomaso:
@AER00 I'm helping out the Jellyfin-web team with reviewing some of these older PR's so we can get on top of the backlog. Wanted to follow up with you and see if some of these ESLint warnings can be resolved? If so, it wont take long to get this merged after it has been tested.

thornbill (approved):

PR #3409, opened 2022-02-09
Fix pagination on live tv pages

thornbill:
**Changes**
Fixes pagination for live tv pages. (Requires unstable build of server to display the right start index.)

**Issues**
N/A

### DIFF
--- a/src/controllers/list.js
+++ b/src/controllers/list.js
@@ -736,7 +736,7 @@ class ItemsView {
                     autoplay: true
                 });
             } else {
-                getItems(self, self.params, currentItem, null, null, 300).then(function (result) {
+                getItems(self, self.params, currentItem, null, 0, 300).then(function (result) {
                     playbackManager.play({
                         items: result.Items,
                         autoplay: true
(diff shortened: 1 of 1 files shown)


src/controllers/list.js
```diff
@@ -15,12 +15,12 @@ import { appRouter } from '../components/appRouter';
 
 /* eslint-disable indent */
 
-    function getInitialLiveTvQuery(instance, params) {
+    function getInitialLiveTvQuery(instance, params, startIndex = 0, limit = 300) {
```
dmitrylyzo, line 18: iirc, the default value doesn't work for passed `null`.
In calls like this (play/queue/shuffle):
https://github.com/jellyfin/jellyfin-web/blob/64b3fc80dcbaa93c080b5ca7b10aac8c3a74ed35/src/controllers/list.js#L739
But this probably won't screw anything up if the server replaces `null` with `0` on its side.
Also, this is not related to the original idea of PR.

dmitrylyzo (approved):
Can't test, but LGTM

src/controllers/list.js
thornbill: Good call. I think it is better to explicitly pass `0` for those. I went ahead and made the change here since it is a tiny change. :slightly_smiling_face:

dmitrylyzo (approved):

PR #3414, opened 2022-02-10
Add support for MPEG-1 Layer II audio for Chrome and Firefox

knackebrot:
If these browsers support MP3, they support MP2 as well, at least when muxed with video (or in the mp4 container). Tested on latest Chrome, Edge Chromium, Firefox and Tizen 2020 TV.

**Changes**
- Added Chrome, Edge Chromium and Firefox to browsers supporting MP2 audio with video
- Enabled MP2 in HLS

**Before**
image

**After**
image

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -370,8 +370,15 @@ import browser from './browser';
                                     || videoTestElement.canPlayType('video/mp4; codecs="avc1.640029, mp4a.6B"').replace(/no/, '')
                                     || videoTestElement.canPlayType('video/mp4; codecs="avc1.640029, mp3"').replace(/no/, '');
 
-        // Not sure how to test for this
-        const supportsMp2VideoAudio = browser.edgeUwp || browser.tizen || browser.web0s;
+        let supportsMp2VideoAudio = options.supportsMp2VideoAudio;
+        if (supportsMp2VideoAudio == null) {
+            supportsMp2VideoAudio = browser.edgeUwp || browser.tizen || browser.web0s;
+
+            // If the browser supports MP3, it presumably supports MP2 as well
+            if (supportsMp3VideoAudio && (browser.chrome || browser.edgeChromium || (browser.firefox && browser.versionMajor >= 83))) {
+                supportsMp2VideoAudio = true;
+            }
+        }
 
         /* eslint-disable compat/compat */
         let maxVideoWidth = browser.xboxOne ?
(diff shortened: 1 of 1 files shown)


src/scripts/browserDeviceProfile.js
```diff
@@ -371,7 +371,7 @@ import browser from './browser';
                                     || videoTestElement.canPlayType('video/mp4; codecs="avc1.640029, mp3"').replace(/no/, '');
 
         // Not sure how to test for this
-        const supportsMp2VideoAudio = browser.edgeUwp || browser.tizen || browser.web0s;
+        const supportsMp2VideoAudio = browser.edgeUwp || browser.tizen || browser.web0s || (supportsMp3VideoAudio ? browser.chrome || browser.edgeChromium || browser.firefox : '');
```
dmitrylyzo, line 374: Maybe split this?
```suggestion
        let supportsMp2VideoAudio = browser.edgeUwp || browser.tizen || browser.web0s;

        // If browser supports MP3, it presumably supports MP2 as well
        if (supportsMp3VideoAudio && (browser.chrome || browser.edgeChromium || browser.firefox)) {
            supportsMp2VideoAudio = true;
        }
```

_At some point, I'd like to add the ability to pass support status in `options` as here:_
https://github.com/jellyfin/jellyfin-web/blob/1cc64865459fb669907022d30e3e7c2c8c177352/src/scripts/browserDeviceProfile.js#L435-L443

Also, are you sure this is true for all versions of browser?

knackebrot: I've added what you requested and tested support in browsers.
Firefox 83 is the first version where mp2 files play, it didn't work in earlier ones.
Chrome 49 is the earliest version I tried and it played fine.

src/scripts/browserDeviceProfile.js
```diff
@@ -370,8 +370,15 @@ import browser from './browser';
...
-        const supportsMp2VideoAudio = browser.edgeUwp || browser.tizen || browser.web0s;
+        let supportsMp2VideoAudio = options.supportsMp2VideoAudio;
+        if (supportsMp2VideoAudio == null) {
+            supportsMp2VideoAudio = browser.edgeUwp || browser.tizen || browser.web0s;
+        }
+
+        // If the browser supports MP3, it presumably supports MP2 as well
+        if (supportsMp3VideoAudio && (browser.chrome || browser.edgeChromium || (browser.firefox && browser.versionMajor >= 83))) {
+            supportsMp2VideoAudio = true;
+        }
```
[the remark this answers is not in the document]
knackebrot: Fixed

dmitrylyzo (approved):

Issue #3416, opened 2022-02-10
Fragmented MP4 with hls.js


knackebrot:
Fragmented MP4 significantly reduces the muxing overhead of HLS fragments, so I think it would be worthwhile to try to make it work on browsers without native HLS support. Starting playback from the beginning works fine, but seeking is problematic. When seeking beyond buffered range (when a new transcoding job is started), the new initialization mp4 file doesn't get downloaded and the playback breaks, though new fragments keep getting downloaded as if it worked fine. Needs investigation whether we can configure hls.js to work, or if it would require a fork to handle our arguably weird use case.

PR #2520, opened 2021-03-22
Prefer fmp4 where available

knackebrot:
Fragmented MP4 provides noticeably lower muxing overhead compared to TS, resulting in smaller segments and lower overall bitrate. It works well (at the minimum) in MSE-enabled browsers, iOS 10+, Tizen 4+, Edge UWP and Xbox One.

```bash
ffmpeg -i big_buck_bunny_720p_surround.mp4 -c copy -copyts -avoid_negative_ts disabled -max_muxing_queue_size 2048 -f hls -max_delay 5000000 -hls_time 6 -hls_segment_type mpegts -hls_playlist_type vod -hls_list_size 0 mpegts/bunny.m3u8
ffmpeg -i big_buck_bunny_720p_surround.mp4 -c copy -copyts -avoid_negative_ts disabled -max_muxing_queue_size 2048 -f hls -max_delay 5000000 -hls_time 6 -hls_segment_type fmp4 -hls_fmp4_init_filename fmp4/init.mp4 -hls_playlist_type vod -hls_list_size 0 fmp4/bunny.m3u8

du mpegts fmp4
63936   mpegts
60492   fmp4

ffmpeg -i big_buck_bunny_240p_surround.mp4 -c copy -copyts -avoid_negative_ts disabled -max_muxing_queue_size 2048 -f hls -max_delay 5000000 -hls_time 6 -hls_segment_type mpegts -hls_playlist_type vod -hls_list_size 0 mpegts_240/bunny.m3u8
ffmpeg -i big_buck_bunny_240p_surround.mp4 -c copy -copyts -avoid_negative_ts disabled -max_muxing_queue_size 2048 -f hls -max_delay 5000000 -hls_time 6 -hls_segment_type fmp4 -hls_fmp4_init_filename fmp4_240/init.mp4 -hls_playlist_type vod -hls_list_size 0 fmp4_240/bunny.m3u8

du mpegts_240 fmp4_240
28180   mpegts_240
25500   fmp4_240
```

**Changes**
- Treat MSE-enabled browsers as supporting fmp4
- Remove webOS from fmp4 support list until confirmed it works
- Enable fmp4 by default

PR #3442, opened 2022-02-19
Enable mp4 transcoding when supported

eyezak:
**Changes**
Enable fmp4 in all supported cases
Fix safari15 quirks
Tested safari15, chrome, firefox

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -399,7 +399,10 @@ import browser from './browser';
                 hlsInTsVideoAudioCodecs.push('mp3');
             }
 
-            hlsInFmp4VideoAudioCodecs.push('mp3');
+            // Safari fails to load HLS.mp4 with mp3 audio
+            if (!browser.safari) {
+                hlsInFmp4VideoAudioCodecs.push('mp3');
+            }
         }
 
         // For AC3/EAC3 remuxing.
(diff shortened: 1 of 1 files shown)


thornbill:
We have an option to enable fmp4 support currently on a per-user basis. Do these changes take that into account? Or replace it entirely? I do not see any updates to the setting itself here. Thanks!

dmitrylyzo:
> We have an option to enable fmp4 support currently on a per-user basis. Do these changes take that into account? Or replace it entirely? I do not see any updates to the setting itself here. Thanks!

https://github.com/jellyfin/jellyfin-web/blob/98e8bf3061bb15cc6472fab801518396bc3c6076/src/scripts/browserDeviceProfile.js#L700

There was another PR: https://github.com/jellyfin/jellyfin-web/pull/2520
It was almost good to me.

@eyezak I see no reason to change order then `push`/`unshift` - fMP4 is better if supported and enabled (prefered).
All we need is to find out in what amount it is supported (codecs).

eyezak:
> 

sorry i didn't get back to this, you are correct I make 'prefer' a preference but with https://github.com/jellyfin/jellyfin/pull/7325 the server honours the order correctly. What is the desired behaviour for this flag?
i will take a look at #3442 with this server change

nyanmisaka:
fmp4 seeking is still broken in hls.js that used on non-Safari browsers. That's the reason why I haven't make it as default.

knackebrot:
https://github.com/jellyfin/jellyfin-web/issues/3416

PR #3453, opened 2022-02-22
Add copy buttons to Media Info

### LINKED ISSUE #3075: Can't Copy Mediainfo
It's not possible to copy mediainfo on mobile browser/in mobile app.
Have not tried on desktop with cursor.

dmitrylyzo:
**Changes**
- Extract copy to clipboard function
- Add copy buttons to Media Info
_Info blocks can be copied individually._
_There is currently no way to copy multiple versions at once._

**Issues**
Fixes #3075 

**Screenshots**

results in:
```
test-gray.1080p
Container: mp4
Path: /mnt/hgfs/jellyfin-test/movies/test-gray.1080p.mp4
Size: 2 MB

Video
Title: 1080i H264 SDR
Codec: H264
Codec tag: avc1
AVC: No
Profile: High
Level: 40
Resolution: 1920x1080
Aspect ratio: 16:9
Interlaced: Yes
Framerate: 25
Bitrate: 62 kbps
Bit depth: 8 bit
Video range: SDR
Pixel format: yuv420p
Ref frames: 1
NAL: 0

Subtitle
Title: Английский - ASS
Language: eng
Codec: ASS
Default: No
Forced: No
External: Yes

Subtitle
Title: Русский - SRT
Language: rus
Codec: SRT
Default: No
Forced: No
External: Yes
```


dmitrylyzo:
Can't get it pixel perfect to the original (without buttons).

src/components/itemMediaInfo/itemMediaInfo.js
```diff
@@ -28,12 +37,24 @@ import template from './itemMediaInfo.template.html';
...
         mediaInfoContent.innerHTML = html;
+
+        for (const btn of mediaInfoContent.querySelectorAll('.btnCopy')) {
+            btn.addEventListener('click', () => {
+                const infoBlock = dom.parentWithClass(btn, 'mediaInfoStream') || dom.parentWithClass(btn, 'mediaInfoSource') || mediaInfoContent;
+
+                copy(infoBlock.textContent).then(() => {
+                    toast(globalize.translate('Copied'));
+                }).catch(() => {
+                    console.error('Could not copy text');
```
thornbill, line 48: Maybe we should show a toast on errors also?

src/scripts/clipboard.js
```diff
@@ -0,0 +1,59 @@
...
+ * @returns {Promise<void>} Promise.
+ */
+function textAreaCopy(text) {
+    const textArea = document.createElement('textarea');
+    textArea.value = text;
+    document.body.appendChild(textArea);
+    textArea.focus();
+
+    if (browser.iOS) {
+        // https://stackoverflow.com/a/46858939
```
thornbill, line 15: This workaround wasn't needed for me on iOS 15... I'm not sure if we should limit it to affected versions or not. What do you think?

dmitrylyzo: iOS 13.4 should be fine with `writeText` (https://stackoverflow.com/a/61868028)
I've added a version check for 10.x

thornbill: Sounds like we should test for <13.4. I believe there are some devices that have a max iOS version of 12.

dmitrylyzo: Done.
On the other hand, iOS 13.4+ won't get that check - it supports `Clipboard.writeText`.

thornbill (approved):

PR #3454, opened 2022-02-23
Fix multiselect not being scoped to current container

thornbill:
**Changes**
The current implementation operates globally so every item container rendered to the page is being manipulated by each instance of the multiselect "component". This was discovered in https://github.com/jellyfin/jellyfin-web/pull/3296#pullrequestreview-846761881 where we found it was acting on items from hidden tabs and pages still in the dom.

* Fixes indentation
* Rewrites as class
* Only act on items in instance container
* Removes unused bindOnClick option for multiselect
* General cleanup

> Note: Converted this back to a draft because it does not handle cases like the home screen where selecting from multiple containers on a page is a feature.

**Issues**
N/A

### DIFF
--- a/src/elements/emby-itemscontainer/emby-itemscontainer.js
+++ b/src/elements/emby-itemscontainer/emby-itemscontainer.js
@@ -19,12 +19,9 @@ import Sortable from 'sortablejs';
 
     function onClick(e) {
         const itemsContainer = this;
-        const multiSelect = itemsContainer.multiSelect;
 
-        if (multiSelect) {
-            if (multiSelect.onContainerClick.call(itemsContainer, e) === false) {
-                return;
-            }
+        if (itemsContainer.multiSelect?.onContainerClick.call(itemsContainer, e) === false) {
+            return;
         }
 
         itemShortcuts.onClick.call(itemsContainer, e);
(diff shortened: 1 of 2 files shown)


src/elements/emby-itemscontainer/emby-itemscontainer.js
```diff
@@ -19,12 +19,9 @@ import Sortable from 'sortablejs';
 
     function onClick(e) {
         const itemsContainer = this;
-        const multiSelect = itemsContainer.multiSelect;
 
-        if (multiSelect) {
-            if (multiSelect.onContainerClick.call(itemsContainer, e) === false) {
-                return;
-            }
+        if (itemsContainer.multiSelect?.onContainerClick.call(itemsContainer, e) === false) {
```
dmitrylyzo, line 23:
```suggestion
        if (itemsContainer.multiSelect?.onContainerClick(e) === false) {
```

dmitrylyzo (commented):
For some reason, there is a second click that removes `checkedInitial`.

UPD:
It comes from the item container and this is a pre-existing bug:
https://github.com/jellyfin/jellyfin-web/blob/78889bb01b5aa62c159052f65bb95a17a92c6b70/src/elements/emby-itemscontainer/emby-itemscontainer.js#L23

PR #3456, opened 2022-02-23
Few refactors of remotecontrol

dmitrylyzo:
**Changes**
- Fix vertical button alignment in Safari 10
- Remove unnecessary flexbox.
- Center card icon in Safari 10 (in all cards)
- Remove unnecessary `height`

**Issues**
N/A

**Screenshots**
_The colors are just to distinguish the elements._

Old
_Using `position: absolute` breaks Flexbox._ https://github.com/jellyfin/jellyfin-web/pull/3018#issuecomment-1048455506

New

**Unresolved problem**
Still can't find a proper fix for Safari 10 not being able to shrink poster.
Adding `min-height: 0` makes the poster shrinkable, but without respecting the aspect ratio.
https://github.com/jellyfin/jellyfin-web/blob/bce24b26ba33a02b635797b36df2852b445b6e53/src/components/remotecontrol/remotecontrol.scss#L395-L400

Another approach is to limit the size of `nowPlayingPageImageContainer`:
```css
max-height: 50%; // or calc(100% - ??em);
```
But it is not universal.

### DIFF
--- a/src/components/cardbuilder/card.scss
+++ b/src/components/cardbuilder/card.scss
@@ -389,6 +389,10 @@ button::-moz-focus-inner {
     transform: translate(-50%, -50%);
 }
 
+.cardImageContainer .cardImageIcon {
+    margin: auto; /* 'justify-content: center' doesn't work in Safari 10 */
+}
+
 .cardIndicators {
     right: 0.225em;
     top: 0.225em;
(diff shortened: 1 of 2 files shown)


src/components/cardbuilder/card.scss
```diff
@@ -389,6 +389,10 @@ button::-moz-focus-inner {
     transform: translate(-50%, -50%);
 }
 
+.cardImageContainer .cardImageIcon {
+    margin: auto; /* 'justify-content: center' doesn't work in Safari 10 */
```
thornbill, line 393: I wondered why this was broken, but not enough to actually dig into it. :sweat_smile:

thornbill (approved):

PR #3459, opened 2022-02-24
Add aria-hidden to button icons

dmitrylyzo:
**Changes**
- Add `aria-hidden` to button icons
- Some refactors

**Issues**
N/A


thornbill (approved):
Nice! I didn't realize we were still using some `<i>` and `<em>` elements... it's good to get those cleaned up. :smile:

Issue #3215, opened 2021-12-08
[Issue]: Add accessible names (button labels) for official iOS client


charmarkk:
### Please describe your bug

I have a mobility impairment and use Voice Control for iOS to interact with my iPad most often. There are two ways to access on-screen elements with Voice Control, numbers (which iOS assigns to visual elements) and names (which are taken from the labels of the elements in an app, webpage, etc. as assigned by developers).

See [Deque's](https://www.deque.com/blog/new-in-ios-13-accessibility-voice-control-and-more/) and [Apple's](https://support.apple.com/en-us/HT210417) articles on the topic for more about this really cool accessibility feature!

I usually use the numbers option, as it's usually a bit easier and more reliable than guessing what the accessible name for an object is, or switching voice control to display those, but with media server clients this is tricky since the numbers always change. It can be hard to get them to show up as well, since you have to "tap" the screen to get the player controls to show up, then Voice Control has to scan the objects on screen and assign them numbers. Once it does this, the player control hide delay expires and the controls disappear, making it hard to "tap" an item by number. It's much easier to say "tap pause" once the controls come up.

I've used Plex for ages, but wanted to switch to something more feature rich and open source, but Jellyfin (for iOS at least) appears to be lacking several crucial accessibility features. I could use "tap play" (or whatever) with Plex for the player controls, but it did not work with Jellyfin. When I invoked the "Show names" Voice Control command, it only showed a unicode "?" ( ⍰ ) for most interactive elements on the app, leading me to believe the player control, menu, and all other icons are made with icon fonts or emojis, and their labels are that icon itself (or non-standard emoji) rather than a human readable string.

What should happen is each button should have an [accessible name](https://www.w3.org/TR/wai-aria-1.2/#namecalculation) - a human readable and understandable label (like "Play", "Skip Forward", "Menu", etc.) - which is what would then be made to be interactive when I say "Tap Play", "Tap Skip Forward", etc. and what a blind/low-vision user using VoiceOver would hear when navigating the app via touch. As is, they would not hear anything. 

[...]

PR #3463, opened 2022-02-25
Add more accessibility

dmitrylyzo:
Tbh, I dunno if ARIA-labels work :sweat_smile: 

**Changes**
- Add title to buttons
- Add aria-labels for accessibility

**Issues**
#3215


thornbill:
I believe screen readers should use the title if present, so I don't think both are needed unless the aria label needs to be different. (Using the accessibility tool in Firefox buttons with titles are not flagged as issues which seems to confirm this.)

thornbill (approved):
Changes look fine and it seems to resolve a lot of accessibility issues found in the Firefox dev tools, so let's ship it. I'm of the opinion that virtually anything would be an improvement currently. We can always refine later if specific issues come up. :rocket:

PR #3471, opened 2022-02-28
Fix remaining any type warnings

thornbill:
**Changes**
* Fixes the remaining any type warnings in TypeScript files

**Issues**
N/A


src/components/search/SearchResultsRow.tsx
```diff
@@ -16,10 +16,31 @@ const createScroller = ({ title = '' }) => ({
 </div>`
 });
 
+type CardOptions = {
+    itemsContainer?: HTMLElement,
+    parentContainer?: HTMLElement,
```
dmitrylyzo, line 21: Isn't `HTMLDivElement` better (stricter)?
So we won't be able to pass buttons/images/etc.

thornbill: In my opinion it's best to keep things as generic as possible until there is a reason to be more strict... :man_shrugging:

dmitrylyzo (approved):

PR #3472, opened 2022-02-28
Add google cast unsupported message and update translations

thornbill:
**Changes**
* Adds a message in the casting menu when Google Cast is unsupported (chromecast web plugin is not loaded)
* Updates translation "Cast the media" -> "Cast to Device"
* Updates translations that use "Chromecast" to use "Google Cast"

Screenshot 2022-02-28 at 12-59-46 Jellyfin

**Issues**
N/A


dmitrylyzo:
With a long title:
cast-1

Maybe center it?
cast-2
Now I am not sure :sweat_smile:

thornbill:
I adjusted the styles a bit so the title and text are using the same left margin and reduced the space between them... I think it is an improvement. :sweat_smile:

dmitrylyzo (approved):

PR #3479, opened 2022-03-02
Change buttons that act as links to anchor tags

davidmurdoch:
Buttons (`<button>`) and links (`<a>`) are different and serve different purposes. Links _go somewhere_ while buttons _do something_.

When I click on a movie title to _go to_ the movie page the thing I click should act like a link. If I ctrl+click or middle click it it should still act like a link... by opening in a new tab. It should have a URL. Conversely, if I click a :heart: button it _shouldn't_ take me to a new page; it should _do something_ to the item: by marking it as "hearted".

I often want to open many links in a new tab, especially when initially categorizing/organizing items, but I can't because they aren't links... they are buttons that eventually act like links (they end up changing the URL). There is even CSS to add an underline to the _button text_ on hover like a link :confused:.

There are dozens of other places where buttons are used in places links should be used, and even some cases where buttons act like a link (by changing the URL) only for that URL to not actual route anywhere (like the dialog buttons that open metadata info).

Anyway, let me know if this change is welcomed or not.

### DIFF
--- a/src/components/cardbuilder/cardBuilder.js
+++ b/src/components/cardbuilder/cardBuilder.js
@@ -22,6 +22,7 @@ import './card.scss';
 import '../../elements/emby-button/paper-icon-button-light';
 import '../guide/programs.scss';
 import ServerConnections from '../ServerConnections';
+import { appRouter } from '../appRouter';
 
         const enableFocusTransform = !browser.slow && !browser.edge;
 
@@ -1038,9 +1039,10 @@ import ServerConnections from '../ServerConnections';
                 return text;
             }
 
-            let html = '<button ' + itemShortcuts.getShortcutAttributesHtml(item, serverId) + ' type="button" class="itemAction textActionButton" title="' + text + '" data-action="link">';
+            const url = appRouter.getRouteUrl(item);
+            let html = '<a href="' + url + '" ' + itemShortcuts.getShortcutAttributesHtml(item, serverId) + ' class="itemAction textActionButton" title="' + text + '" data-action="link">';
             html += text;
-            html += '</button>';
+            html += '</a>';
 
             return html;
         }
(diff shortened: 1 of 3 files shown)


thornbill:
In general I am in favor of this change, but I seem to recall encountering some weird edge cases with our router when using anchor tags. Since we are getting close to the 10.8 feature freeze, I would prefer to hold off on merging this until 10.9 gets rolling to give us a full release cycle for testing. 😅

davidmurdoch:
Understandable. There are so many platforms to test on I can't fathom how to do proper due diligence on these sorts of changes.

thornbill:
Feature freeze is over, so whenever you get a chance to rebase to fix the conflicts and fix the eslint issue we can get this reviewed + merged. :slightly_smiling_face:

davidmurdoch:
I used the GitHub UI to resolve the conflict, and I haven't yet retested my changes. Are you okay with the merge commit GitHub made or do you prefer an actual rebase?

thornbill:
I know there a lot of git history purists out there, but I am not one of them... I generally don't care as long as we have good readable commit messages. :sweat_smile:

davidmurdoch:
@thornbill I've tested the changes and things still work. :-) Only issue is that the card's `Year` isn't clickable, but looks clickable when hovering over it, but that's the case on Master now. I'm not sure if it's always been that way though.

src/components/cardbuilder/cardBuilder.js
```diff
@@ -1470,6 +1473,8 @@ import ServerConnections from '../ServerConnections';
             let html = '';
 
             html += '<div class="cardOverlayContainer itemAction" data-action="' + action + '">';
+            const url = appRouter.getRouteUrl(item);
+            html += '<a href="' + url + '" class="cardImageContainer"></a>'
```
davidmurdoch, line 1477:
```suggestion
            html += '<a href="' + url + '" class="cardImageContainer"></a>';
```

thornbill (approved):

PR #3480, opened 2022-03-04
Fix more accessibility issues

thornbill:
**Changes**
* Adds aria-hidden to blurhash canvases
* Adds aria-hidden to the Jellyfin logo in the header
* Adds titles to the emby-scrollbuttons

I think the biggest remaining issue on the home screen as far as text labels go are the card overlays... the way that they are hidden does not prevent them from receiving focus... solving that would probably help a lot.

**Issues**
N/A


src/scripts/libraryMenu.js
```diff
@@ -28,7 +28,7 @@ import Headroom from 'headroom.js';
         html += '<button type="button" is="paper-icon-button-light" class="headerButton headerButtonLeft headerBackButton hide"><span class="material-icons ' + (browser.safari ? 'chevron_left' : 'arrow_back') + '" aria-hidden="true"></span></button>';
         html += '<button type="button" is="paper-icon-button-light" class="headerButton headerHomeButton hide barsMenuButton headerButtonLeft"><span class="material-icons home" aria-hidden="true"></span></button>';
         html += '<button type="button" is="paper-icon-button-light" class="headerButton mainDrawerButton barsMenuButton headerButtonLeft hide"><span class="material-icons menu" aria-hidden="true"></span></button>';
-        html += '<h3 class="pageTitle"></h3>';
+        html += '<h3 class="pageTitle" aria-hidden="true"></h3>';
```
dmitrylyzo, line 31: The library name goes here.
But the tab name is probably more relevant.

dmitrylyzo (approved):
> I think the biggest remaining issue on the home screen as far as text labels go are the card overlays... the way that they are hidden does not prevent them from receiving focus... solving that would probably help a lot.

Also the drawer.

PR #3481, opened 2022-03-04
Sort album tracks by (Parent)IndexNumber, then SortName

Maxr1998:
Track sorting in albums could previously be incorrect if metadata specified a SortName tag.

Also see: jellyfin/jellyfin#7323, jellyfin/jellyfin#7324

### DIFF
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -1319,7 +1319,9 @@ function renderChildren(page, item) {
         Fields: fields
     };
 
-    if (item.Type !== 'BoxSet') {
+    if (item.Type == 'MusicAlbum') {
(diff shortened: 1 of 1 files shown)


src/controllers/itemDetails/index.js
```diff
@@ -1319,7 +1319,9 @@ function renderChildren(page, item) {
         Fields: fields
     };
 
-    if (item.Type !== 'BoxSet') {
+    if (item.Type == 'MusicAlbum') {
```
thornbill, line 1322: It might be more logical to have this on line no. 1344 where we do something similar for `MusicArtist`.

Maxr1998: Maybe all those cases could be merged, yes. I just wanted to keep the two `SortBy` cases together.

thornbill (approved):

PR #3501, opened 2022-03-15
Fix old streamInfo on stream change

dmitrylyzo:
When the client changes the stream (fail to play) `streamInfo` contains the old play session and the next progress report will update the wrong session. The server then responds with the wrong playback method until the next progress report.

**Changes**
Update `streamInfo` before playback start so that the data is up to date.
_Similar to #3373_

**Issues**
N/A


thornbill:
This probably needs to go in 10.8 right?

thornbill (approved):

PR #3503, opened 2022-03-18
Fix birth location link

felixoswald:
**Changes**
Fixed incorrect HTML link handling on actor details page.

before fix:
image

after fix:
image

**Issues**
none

### DIFF
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -669,7 +669,7 @@ function reloadFromItem(instance, page, params, item, user) {
             location = `<a is="emby-linkbutton" class="button-link textlink" target="_blank" href="https://www.openstreetmap.org/search?query=${encodeURIComponent(location)}">${location}</a>`;
         }
         itemBirthLocation.classList.remove('hide');
-        itemBirthLocation.innerText = globalize.translate('BirthPlaceValue', location);
+        itemBirthLocation.innerHTML = globalize.translate('BirthPlaceValue', location);
     } else {
         itemBirthLocation.classList.add('hide');
     }


src/controllers/itemDetails/index.js
```diff
@@ -669,7 +669,7 @@ function reloadFromItem(instance, page, params, item, user) {
             location = `<a is="emby-linkbutton" class="button-link textlink" target="_blank" href="https://www.openstreetmap.org/search?query=${encodeURIComponent(location)}">${location}</a>`;
```
nielsvanvelzen, line 669: Location should be escaped here to prevent an XSS injection.

felixoswald: should be fixed with commit 55ad352d08e4162d1e99114b044176545f012501

src/controllers/itemDetails/index.js
```diff
@@ -666,10 +666,12 @@ function reloadFromItem(instance, page, params, item, user) {
...
         let location = item.ProductionLocations[0];
         if (!layoutManager.tv && appHost.supports('externallinks')) {
-            location = `<a is="emby-linkbutton" class="button-link textlink" target="_blank" href="https://www.openstreetmap.org/search?query=${encodeURIComponent(location)}">${location}</a>`;
+            itemBirthLocation.innerHTML = globalize.translate('BirthPlaceValue', `<a is="emby-linkbutton" class="button-link textlink" target="_blank" href="https://www.openstreetmap.org/search?query=${encodeURIComponent(location)}"></a>`);
+            page.querySelector('#itemBirthLocation > a').innerText = `${location}`;
+        } else {
+            itemBirthLocation.innerText = globalize.translate('BirthPlaceValue', `${location}`);
         }
         itemBirthLocation.classList.remove('hide');
-        itemBirthLocation.innerText = globalize.translate('BirthPlaceValue', location);
```
dmitrylyzo, line 672: You can revert my changes in this line (https://github.com/jellyfin/jellyfin-web/commit/59adbc348a37bccc8f9a277378e863efbe95497a) and use `escapeHtml`:
```js
        let location = item.ProductionLocations[0];
        if (!layoutManager.tv && appHost.supports('externallinks')) {
            location = `<a is="emby-linkbutton" class="button-link textlink" target="_blank" href="https://www.openstreetmap.org/search?query=${encodeURIComponent(location)}">${escapeHtml(location)}</a>`;
        } else {
            location = escapeHtml(location);
        }
        itemBirthLocation.classList.remove('hide');
        itemBirthLocation.innerHTML = globalize.translate('BirthPlaceValue', location);
```

felixoswald: fixed in 2521b0614225db9a6f6a063c2eccf754d74fb15f

felixoswald: shouldn't the links in `renderLinks` (lines 745 and 750) also use `escapeHtml()`?

dmitrylyzo:
https://github.com/jellyfin/jellyfin-web/pull/3503#discussion_r830460067
> shouldn't the links in `renderLinks` (lines 745 and 750) also use `escapeHtml()`?

This, yes (at least no harm):
https://github.com/jellyfin/jellyfin-web/blob/55ad352d08e4162d1e99114b044176545f012501/src/controllers/itemDetails/index.js#L750

There are `href`s that should probably be URI-escaped, but I am not sure (we are not cleaning them yet).

We can more or less trust the translation. If not, we should escape them in `globalize` module.

src/controllers/itemDetails/index.js
dmitrylyzo: I replied in the main thread to mark this conversation as resolved.

felixoswald:
https://github.com/jellyfin/jellyfin-web/pull/3503#issuecomment-1072974521

Should be fixed. Haven't found any other links on the actor page which are not escaped

nielsvanvelzen (approved):

PR #3506, opened 2022-03-20
Provide more friendly transcoding bitrate options.

shenwii:
Changed the bitrate option for transcoding.
*Remove some high bitrate options, such as 1080P 60M
*Add some low bitrate options, for example, for 1080P, there should be lower bitrate options, such as 4M, setting 10M as the lowest bitrate for 1080P is not appropriate, in fact, most 1080P bitrates are less than 10M

Filter out transcoding options that are higher than the source streaming bitrate multiplied by 3.
For example, a 1080P 10M hevc source transcoded to 1080P 40M does not make much sense.

PS: I know that hardware-accelerated transcoding is worse than CPU transcoding, and HEVC encoding has better compression, but even so, the transcoded bitrate is not more than 3 times the original bitrate at about the same image quality, I think.


Shadowghost:
Just as a sidenote: AFAIK currently the resolution isn'tt aken into account at all by the transcoder, only bitrate and display resolution.

src/components/qualityOptions.js
```diff
@@ -32,29 +75,57 @@ export function getVideoQualityOptions(options) {
 
     // Quality options are indexed by bitrate. If you must duplicate them, make sure each of them are unique (by making the last digit a 1)
     if (maxAllowedWidth >= 3800 && hostScreenWidth >= 1930) {
-        qualityOptions.push({ name: '4K - 120 Mbps', maxHeight: 2160, bitrate: 120000000 });
-        qualityOptions.push({ name: '4K - 80 Mbps', maxHeight: 2160, bitrate: 80000000 });
+        qualityOptions2160p.forEach(function(qualityOption) {
+            if (videoBitRate) {
+                if (qualityOption.bitrate < videoBitRate * maxStreamingBitrateMultiplier) {
```
1hitsong, line 80: By running the `videoBitRate * maxStreamingBitrateMultiplier` calculation in all these forEach functions you're calculating the same value 24 times. Better to move it outside of these loops and only calculate it once.

1hitsong (changes requested):

1hitsong:
Functionally, it works well. Just some minor code cleanup / optimization items to address.

thornbill:
Superseded by #6071

PR #3507, opened 2022-03-21
Comics: Add support for .zip & .rar comics

### LINKED ISSUE #3446: Comics: Add support for .zip and .rar comics
**Describe The Bug**

The current behaviour for the comic player is to determine whether a file is a comic based on the extension, either `.cbz`, or `.cbr`

https://github.com/jellyfin/jellyfin-web/blob/29faa698b72a05c1c6ff39a087e2091b10c705df/src/plugins/comicsPlayer/plugin.js#L236

These are simply `.zip` and `.rar` archives, with a different extension to indicate the content type. 

I think it would be helpful to allow these files to be read directly, possibly with a per-library toggle for readign `.zip` and `.rar` files as comics. 

**Media Information**
NA

**Screenshots**
NA

 - Jellyfin Version: 10.7

matthewstrasiotto:
**Changes**

As documented in #3446 , the `.cbz` and `.cbr` extensions simply correspond to `.zip` and `.rar` archives respectively.

Presently, comics can only be played by jellyfin if they use one of the "explicit" extensions - a `.cbr` or `.cbz`. This would allow `.zip`s and `.rar`s to be recognized. 

**Issues**

Resolves #3446

### DIFF
--- a/src/plugins/comicsPlayer/plugin.js
+++ b/src/plugins/comicsPlayer/plugin.js
@@ -233,11 +233,11 @@ export class ComicsPlayer {
     }
 
     canPlayItem(item) {
-        if (item.Path && (item.Path.endsWith('cbz') || item.Path.endsWith('cbr'))) {
-            return true;
-        }
+        const supportedFormats = ['cbz', 'zip', 'cbr', 'rar'];
 
-        return false;
+        const name = item.Path;
+        const index = name.lastIndexOf('.');
+        return index !== -1 && supportedFormats.includes(name.slice(index + 1));
     }
 }
 


thornbill:
This probably deserves some consideration from @jellyfin/core.

thornbill:
The core team discussed this a bit and the overall consensus was this is not something we want to support. In addition to the great points summarized in [this comment](https://github.com/jellyfin/jellyfin-web/issues/3446#issuecomment-1068228979), we have generally been opposed to supporting media in archive formats (zip/rar) with cbz/cbr being the one real exception for books.

PR #3516, opened 2022-03-26
Fix transcode reasons listing

dmitrylyzo:
In https://github.com/jellyfin/jellyfin/pull/7325 `TranscodeReasons` was changed to `TranscodeReason`.

**Changes**
Use `TranscodeReason` and beautfy the code.

**Issues**
No transcode reasons in stats and Dashboard.


dmitrylyzo:
Better to fix the server https://github.com/jellyfin/jellyfin/pull/7496

PR #3522, opened 2022-03-28
Fix issue #3440 where networking menu can get in a bad state.

### LINKED ISSUE #3440: [Issue]: Can't make any changes to the Admin->Networking page, blue save button is not functional
### Please describe your bug

The blue "save button" on the Networking page is broken. This is a brand new installation. If I try to make any changes to the network settings and click the blue "save" button at the bottom of the page, nothing happens. There is no notification of the setting being saved. If  I navigate away from the page all changes are lost.

I also tried the following:

1. Opening the page is a Firefox private window and loading the page.
2. Generating a new Firefox profile with default settings and loading the page.
3. Accessing the page thought https://domain.com/web/index.html#!/networking.html
4. Accessing the page though the http://localhost:8096/web/index.html#!/networking.html

The blue save button does not respond when a change is made and then clicked. There is no way to save changes to the networking page.

Note: The save button on other pages works as expected. I can click on DLNA, make a change, click save and it saves as expected. The network save button however does not work.

### Jellyfin Version

10.7.7

### if other:

_No response_

### Environment

```markdown
- OS:Raspberry Pi (Debian bullseye)
- Virtualization: None
- Clients: Browser
- Browser: Firefox 98.0
- FFmpeg Version: n/a
- Playback Method: n/a
- Hardware Acceleration: n/a
- Plugins: n/a
- Reverse Proxy: apache
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

```shell
There is no event in the log of the attempted configuration change, the save button is broken.
```

### FFmpeg logs

```shell
n/a
```

_No response_

_No response_

### Code of Conduct

gardner:
**Changes**

1. Removed event handler that was hiding part of the menu when a setting was unchecked.
2. Removed "hide" attribute from form items so that they are visible by default.

**Issues**

Fixes #3440

### DIFF
--- a/src/controllers/dashboard/networking.html
+++ b/src/controllers/dashboard/networking.html
@@ -87,36 +87,36 @@
                             </label>
                             <div class="fieldDescription checkboxFieldDescription">${AllowRemoteAccessHelp}</div>
                         </div>
-                        <div class="inputContainer fldExternalAddressFilter hide">
+                        <div class="inputContainer fldExternalAddressFilter">
                             <input is="emby-input" type="text" id="txtExternalAddressFilter" label="${LabelAllowedRemoteAddresses}" />
                             <div class="fieldDescription">${AllowedRemoteAddressesHelp}</div>
                         </div>
-                        <div class="selectContainer fldExternalAddressFilterMode hide">
+                        <div class="selectContainer fldExternalAddressFilterMode">
                             <select is="emby-select" id="selectExternalAddressFilterMode" label="${LabelAllowedRemoteAddressesMode}">
                                 <option value="whitelist">${Whitelist}</option>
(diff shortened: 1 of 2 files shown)


gardner:
Hi there,
  this is my first contribution to the project so I am a bit naive about the best way to do things in this project. Most of my front-end experience is with React.

> Maybe we just aren't resetting the options properly when the remote access option changes?

What do you propose? 

1. Should we validate the form before it is hidden?
2. Reset it to defaults when they uncheck the parent checkbox?
3. Should we store values that were input?

Cheers

thornbill:
I would expect it to reset the default values and not save any entered values.

gardner:
Closing this so someone else can fix it.

PR #3523, opened 2022-03-28
Add some tips for the enhanced NVDEC decoder

nyanmisaka:
**Changes**
- add some tips to prevent users from selecting it in general.

**Issues**
Fixes [#7508](https://github.com/jellyfin/jellyfin/issues/7508)

### DIFF
--- a/src/controllers/dashboard/encodingsettings.html
+++ b/src/controllers/dashboard/encodingsettings.html
@@ -83,6 +83,7 @@
                             <input type="checkbox" is="emby-checkbox" id="chkEnhancedNvdecDecoder" />
                             <span>${EnableEnhancedNvdecDecoder}</span>
                         </label>
+                        <div class="fieldDescription checkboxFieldDescription">${EnableEnhancedNvdecDecoderHelp}</div>
                     </div>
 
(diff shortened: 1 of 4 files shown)


nyanmisaka:
> Do we want this backported to 10.8 to prevent user errors?

Sure, but I can't add backport labels.

gardner (commented):
This just exposes the setting in the UI. Is there code elsewhere that uses the setting?

gardner:
The Spanish version of the text could be:

> Implementación experimental de NVDEC, no habilite esta opción a menos que encuentre errores de decodificación.

nyanmisaka:
> This just exposes the setting in the UI. Is there code elsewhere that uses the setting?

It's used in the server side to switch between two NVDEC impl.

nyanmisaka:
> The Spanish version of the text could be:
> 
> > Implementación experimental de NVDEC, no habilite esta opción a menos que encuentre errores de decodificación.

Thanks. You can also new translation via https://translate.jellyfin.org/

thornbill (approved):
These changes appear to be fine, but I noticed on my server the option appears to be enabled by default.

**Update:** I see it should be disabled by default. This may have just been some artifact from previous testing.

PR #3527, opened 2022-03-29
Fix rewatching next up status

thornbill:
**Changes**
Fixes an issue where the rewatching in next up option shows as enabled when it is not

**Issues**
N/A

### DIFF
--- a/src/scripts/settings/userSettings.js
+++ b/src/scripts/settings/userSettings.js
@@ -468,7 +468,7 @@ export class UserSettings {
             return this.set('enableRewatchingInNextUp', val, false);
         }
 
-        return this.get('enableRewatchingInNextUp', false);
+        return this.get('enableRewatchingInNextUp', false) === 'true';
     }
 
     /**


nielsvanvelzen (approved):

src/scripts/settings/userSettings.js
```diff
@@ -468,7 +468,8 @@ export class UserSettings {
             return this.set('enableRewatchingInNextUp', val, false);
         }
 
-        return this.get('enableRewatchingInNextUp', false);
+        const enabled = this.get('enableRewatchingInNextUp', false);
+        return enabled === 'true';
```
dmitrylyzo, line 472: Why not just
```suggestion
        return this.get('enableRewatchingInNextUp', false) === 'true';
```

thornbill: I debated going this route, but thought it might be harder to read... but it's probably fine, so let's go with that :smile:

dmitrylyzo (approved):

PR #3535, opened 2022-03-31
Fix 'resume' when clicking on item details poster

dmitrylyzo:
`detailImageContainer` contains no `data-action`, so action is always `undefined`.

**Changes**
Find the element with 'action'.

**Issues**
Clicking on the poster doesn't resume.

### DIFF
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -1926,7 +1926,15 @@ export default function (view, params) {
     }
 
     function onPlayClick() {
-        playCurrentItem(this, this.getAttribute('data-action'));
+        let actionElem = this;
+        let action = actionElem.getAttribute('data-action');
+
+        if (!action) {
+            actionElem = actionElem.querySelector('[data-action]') || actionElem;
+            action = actionElem.getAttribute('data-action');
+        }
+
+        playCurrentItem(actionElem, action);
     }
 
     function onInstantMixClick() {


src/controllers/itemDetails/index.js
```diff
@@ -1926,7 +1926,15 @@ export default function (view, params) {
     }
 
     function onPlayClick() {
-        playCurrentItem(this, this.getAttribute('data-action'));
+        let actionElem = this;
+        let action = actionElem.getAttribute('data-action');
+
+        if (!action) {
+            actionElem = actionElem.querySelector('[data-action]') || actionElem;
```
thornbill, line 1933: The fallback to `actionElem` doesn't really make sense here since it would just end up with the action being undefined again.

dmitrylyzo: Originally it was:
```js
const actionElem = this.querySelector('[data-action]') || this;
const action = actionElem.getAttribute('data-action');
```

There are 2 layouts that we currently need to handle:
```
detailImageContainer <-- this
    poster data-action
```
and
```
btnPlay data-action <-- this
```

That worked fine. But I decide to also handle the item card with the overlay - I left the play button in Reactified poster:
```
poster data-action <-- this
    overlayPlay data-action
```

Alternatively, we can search for a parent with `data-action` starting from `e.target` using `dom.parentWithAttribute`.

PR #3536, opened 2022-03-31
Remove setTitle function in app router

thornbill:
**Changes**
This doesn't seem like something that belongs in a router and it was just an abstraction over the library menu function :man_shrugging: 

**Issues**
N/A


dmitrylyzo:
SonarCloud smells import dup.

thornbill:
Fixed the duplicate import and added a rule to prevent that in #3539

dmitrylyzo (approved):

PR #3543, opened 2022-04-03
Restore sort order and fix strm usecase

Shadowghost:
**Changes**
* Restore previous stream order presentation to remedy the stream index changes in  jellyfin/jellyfin#7529
* Allow subtitle selection when no video is present (required for not probed strm files)

Followup on jellyfin/jellyfin#7529 but works without it too.

### DIFF
--- a/src/components/itemHelper.js
+++ b/src/components/itemHelper.js
@@ -331,6 +331,17 @@ export function supportsMediaSourceSelection (item) {
     return true;
 }
 
+export function sortTracks (trackA, trackB) {
+    let cmp = trackA.IsExternal - trackB.IsExternal;
+    if (cmp != 0) return cmp;
+    cmp = trackB.IsForced - trackA.IsForced;
+    if (cmp != 0) return cmp;
+    cmp = trackB.IsDefault - trackA.IsDefault;
+    if (cmp != 0) return cmp;
+
+    return trackA.Index - trackB.Index;
+}
+
 export default {
     getDisplayName: getDisplayName,
     supportsAddingToCollection: supportsAddingToCollection,
@@ -346,5 +357,6 @@ export default {
     canRate: canRate,
     canConvert: canConvert,
     canRefreshMetadata: canRefreshMetadata,
-    supportsMediaSourceSelection: supportsMediaSourceSelection
+    supportsMediaSourceSelection: supportsMediaSourceSelection,
+    sortTracks: sortTracks
 };
--- a/src/components/itemMediaInfo/itemMediaInfo.js
+++ b/src/components/itemMediaInfo/itemMediaInfo.js
@@ -12,6 +12,7 @@ import toast from '../toast/toast';
 import { copy } from '../../scripts/clipboard';
 import dom from '../../scripts/dom';
 import globalize from '../../scripts/globalize';
+import itemHelper from '../../components/itemHelper';
 import loading from '../loading/loading';
 import '../../elements/emby-select/emby-select';
 import '../listview/listview.scss';
@@ -71,8 +72,8 @@ const attributeDelimiterHtml = layoutManager.tv ? '' : '<span class="hide">: </s
             const size = `${(version.Size / (1024 * 1024)).toFixed(0)} MB`;
             html += `${createAttribute(globalize.translate('MediaInfoSize'), size)}<br/>`;
         }
-        for (let i = 0, length = version.MediaStreams.length; i < length; i++) {
-            const stream = version.MediaStreams[i];
+        version.MediaStreams.sort(itemHelper.sortTracks);
+        for (const stream of version.MediaStreams) {
             if (stream.Type === 'Data') {
                 continue;
             }
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -274,7 +274,7 @@ function renderAudioSelections(page, mediaSources) {
     const tracks = mediaSource.MediaStreams.filter(function (m) {
         return m.Type === 'Audio';
     });
-    tracks.sort((a, b) => itemHelper.sortTracks(a, b));
+    tracks.sort(itemHelper.sortTracks);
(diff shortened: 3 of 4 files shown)


src/components/itemHelper.js
```diff
@@ -331,6 +331,30 @@ export function supportsMediaSourceSelection (item) {
     return true;
 }
 
+export function sortTracks (trackA, trackB) {
+    if (trackA.IsExternal && trackB.IsExternal) {
```
dmitrylyzo, line 335: _visually - not tested_
```suggestion
    if (trackA.IsExternal === trackB.IsExternal) {
```
Otherwise, if they are both non-external, `-1` will be returned instead of checking the other properties.

The same for others.

dmitrylyzo (commented):
I usually write this like:
```js
if (x1 < x2) return -1;
if (x2 < x1) return 1;

if (y1 < y2) return -1;
if (y2 < y1) return 1;

return 0;
```

src/components/itemHelper.js
```diff
@@ -331,6 +331,19 @@ export function supportsMediaSourceSelection (item) {
...
 
+export function sortTracks (trackA, trackB) {
+    if (Number(trackA.IsExternal) < Number(trackB.IsExternal)) return -1;
+    if (Number(trackB.IsExternal) < Number(trackA.IsExternal)) return 1;
+
+    if (Number(trackA.IsForced) < Number(trackB.IsForced)) return -1;
+    if (Number(trackB.IsForced) < Number(trackA.IsForced)) return 1;
+
+    if (Number(trackB.IsDefault) < Number(trackA.IsDefault)) return -1;
+    if (Number(trackB.IsDefault) < Number(trackA.IsDefault)) return 1;
```
dmitrylyzo, line 342:
```suggestion
    if (Number(trackA.IsDefault) < Number(trackB.IsDefault)) return 1;
```
I assumed you need `Default` to go first.

dmitrylyzo (commented):
At least easier to read. If you need `Forced` to go first, swap conditions.

I am not sure about using `Number` - `true` and `false` are comparable.
At the same time, you can write:
```js
const cmp = Number(a) - Number(b);
if (cmp !== 0) return cmp;
```
You can change the order by swapping operands.

src/controllers/itemDetails/index.js
```diff
@@ -303,31 +304,26 @@ function renderSubtitleSelections(page, mediaSources) {
...
+        selected = v.Index === selectedId ? ' selected' : '';
+        return '<option value="' + v.Index + '" ' + selected + '>' + v.DisplayTitle + '</option>';
+    }).join('');
 
-        if (tracks.length > 0) {
-            select.removeAttribute('disabled');
-        } else {
-            select.setAttribute('disabled', 'disabled');
-        }
+    if (tracks.length > 1) {
```
dmitrylyzo, line 318: I think there was `> 0` to allow selecting `Off` in case of single subtitles - to disable subtitles.

Shadowghost: Makes senese. I'll change it and clean up the commits if we're sure whe want the sorting done clientside.

src/components/itemHelper.js
```diff
@@ -331,6 +331,17 @@ export function supportsMediaSourceSelection (item) {
...
 
+export function sortTracks (trackA, trackB) {
+    let cmp = trackA.IsExternal - trackB.IsExternal;
+    if (cmp != 0) return cmp;
+    cmp = trackB.IsForced - trackA.IsForced;
+    if (cmp != 0) return cmp;
+    cmp = trackB.IsDefault - trackA.IsDefault;
+    if (cmp != 0) return cmp;
+
+    return 0;
```
dmitrylyzo, line 342: I shuffled the tracks to test the PR. We could use `Index` as the last ordering field, if this doesn't hurt your goal.
```suggestion
    return trackA.Index - trackB.Index;
```

Shadowghost: I don't realy care, did it provide a better result in some way in your test?

dmitrylyzo: After shuffling, I get tracks unordered by `Index`. In my case, this restores the original order.

At the same time, the server response is already ordered by `Index` (after your investigation). So maybe this is unnecessary or just to be sure.

dmitrylyzo (approved):

src/components/itemMediaInfo/itemMediaInfo.js
```diff
@@ -71,8 +72,9 @@ const attributeDelimiterHtml = layoutManager.tv ? '' : '<span class="hide">: </s
             const size = `${(version.Size / (1024 * 1024)).toFixed(0)} MB`;
             html += `${createAttribute(globalize.translate('MediaInfoSize'), size)}<br/>`;
         }
-        for (let i = 0, length = version.MediaStreams.length; i < length; i++) {
-            const stream = version.MediaStreams[i];
+        const streams = version.MediaStreams.sort((a, b) => itemHelper.sortTracks(a, b));
```
thornbill, line 75: Couldn't you simplify these to just `version.MediaStreams.sort(itemHelper.sortTracks)`?

src/components/itemMediaInfo/itemMediaInfo.js
```diff
@@ -71,8 +72,9 @@ const attributeDelimiterHtml = layoutManager.tv ? '' : '<span class="hide">: </s
...
+        for (let i = 0, length = streams.length; i < length; i++) {
+            const stream = streams[i];
```
dmitrylyzo, line 77:
```suggestion
        for (const stream of streams) {
```
It is shorter, but the transpilation makes it larger: `try/catch`ed iterator.

Shadowghost: Should we change it or not? imo both is readable and if the current implementation leads to less transpilation I don't really care what we use.

dmitrylyzo (approved):

thornbill (approved):

Issue #3345, opened 2022-01-17
Comic / Manga navigation broken for the ePub format


MinecraftPlaye:
**Describe The Bug**
Navigation in ePub comics does not work. You can navigate via the table of contents and via pressing the right arrow button (which means going a page back for mangas) on the desktop client. The left arrow button does nothing.
On iPadOS I couldn't even navigate back.

**Steps To Reproduce**
1. Go to the Book library.
2. Click on a manga/comic. It has to be stored via ePub. I specifically tested ePub 3 files.
3. Try to navigate.

**Expected Behavior**
I expect to be able to navigate through the book in all directions.

**System (please complete the following information):**
First Test:

- Platform: Manjaro Linux
- Browser: Firefox 96.0
- Jellyfin Version: 10.8.0 Alpha 5
- Jellyfin Web Version: 10.8.0 Alpha 5

Second Test:

- Platform: Manjaro Linux
- Jellyfin Version: 10.7.7
- Jellyfin Media-Player: 1.6.1

Third Test:

- Platform: iPad
- Jellyfin Version: 10.7.7
- Jellyfin Mobile (iOS) Version: 1.4.1

MinecraftPlaye:
Fourth test:

    Platform: Manjaro Linux
    Browser: Firefox 96.0
    Jellyfin Version: 10.8.0 Beta 1
    Jellyfin Web Version: 10.8.0 Beta 1

Another note: from my understanding, jellyfin-web uses epubjs and the demo on their website suffers from the same problem.

Additional information:
Clicking on a chapter on the toc moves you forward to that chapter. However, you can't continue in the book, only navigate back to the beginning. Thus I believe this issue might be related to the reading direction of mangas, but I'm not exactly certain.

thornbill:
I was going to make some edits to this and flag it as a regression, but I think it maybe covering multiple issues...

In my testing, touch events are broken in the epub player on iOS devices. I do not have a modern Android device to check there, but the events _do_ work with Firefox's touch device emulation. The touch events work for me using the same books in 10.7.7.

MinecraftPlaye:
> In my testing, touch events are broken in the epub player on iOS devices in 10.8.

That's definitely a regression as this worked fine in 10.7.7 on my iPad. 

> I do not have a modern Android device to check there, but the events do work with Firefox's touch device emulation. The touch events work for me using the same books in 10.7.7.

[...]

PR #3547, opened 2022-04-05
Fix touch events in epub player

thornbill:
**Changes**
Fixes a regression in 10.8 where touch events do not work in Safari due to iframe content restrictions added in epub.js

**Issues**
Fixes the touch event issue mentioned in #3345 (but not the manga specific issue)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5243,9 +5243,9 @@
       "dev": true
     },
     "epubjs": {
-      "version": "0.3.90",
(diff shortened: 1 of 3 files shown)


dmitrylyzo (approved):
LGTM.

Except that only the EPUB player doesn't work in the webOS 1.2 emulator, but I am not sure if it ever worked.

But it works in the webOS 4 emulator.

PR #3548, opened 2022-04-05
Remove some functions from the global scope

thornbill:
**Changes**
Just some cleanup of a couple functions that were being attached to the global scope (one of which had a duplicate implementation in the app router)

I did a search and it does not seem like anything outside of jellyfin-web used these, so this should be safe :slightly_smiling_face: 

**Issues**
N/A

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -194,7 +194,6 @@ module.exports = {
                 'DashboardPage': 'writable',
                 'Emby': 'readonly',
                 'getParameterByName': 'writable',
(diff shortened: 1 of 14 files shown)


thornbill:
Okay I simplified these a bit further. We already use `URLSearchParams` in a few places, so I assume it is safe... otherwise we should add a polyfill

dmitrylyzo:
We have
```
location.hash: '#!/movies.html?topParentId=...'
location.search: ''
```
because search parameters go before the hash.

PR #3549, opened 2022-04-06
Show subtitle text color setting

Madh93:
**Changes**
Show text color option in subtitle appearance setting:

imagen

**Issues**
N/A

### DIFF
--- a/src/components/subtitlesettings/subtitlesettings.js
+++ b/src/components/subtitlesettings/subtitlesettings.js
@@ -32,7 +32,7 @@ function getSubtitleAppearanceObject(context) {
     appearanceSettings.dropShadow = context.querySelector('#selectDropShadow').value;
     appearanceSettings.font = context.querySelector('#selectFont').value;
     appearanceSettings.textBackground = context.querySelector('#inputTextBackground').value;
-    appearanceSettings.textColor = context.querySelector('#inputTextColor').value;
+    appearanceSettings.textColor = layoutManager.tv ? context.querySelector('#selectTextColor').value : context.querySelector('#inputTextColor').value;
     appearanceSettings.verticalPosition = context.querySelector('#sliderVerticalPosition').value;
 
     return appearanceSettings;
@@ -57,6 +57,7 @@ function loadForm(context, user, userSettings, appearanceSettings, apiClient) {
         context.querySelector('#selectTextWeight').value = appearanceSettings.textWeight || 'normal';
         context.querySelector('#selectDropShadow').value = appearanceSettings.dropShadow || '';
         context.querySelector('#inputTextBackground').value = appearanceSettings.textBackground || 'transparent';
+        context.querySelector('#selectTextColor').value = appearanceSettings.textColor || '#ffffff';
         context.querySelector('#inputTextColor').value = appearanceSettings.textColor || '#ffffff';
         context.querySelector('#selectFont').value = appearanceSettings.font || '';
         context.querySelector('#sliderVerticalPosition').value = appearanceSettings.verticalPosition;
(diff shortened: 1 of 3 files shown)


Madh93:
> > I have always wondered why this was hidden... man_shrugging
> 
> Probably because you need to know what to write in it.
> 
> I prefer this field to be replaced by the color picker. Btw, there is option for the background color.

Yeah, by default the value for the background color is `transparent`, but if you specify some color the result is kind of weird:

imagen
imagen

dmitrylyzo:
> Yeah, by default the value for the background color is `transparent`, but if you specify some color the result is kind of weird:

I know. This is what they look like in Firefox.
Appearance is slightly better in Chrome:
subs-1

Maybe we should replace it with background opacity?

I'm not talking about doing it immediately - just to think about it.

dmitrylyzo:
As an alternative to the color picker, the text field can be replaced with the select.
Palette? I think we can start with just `white` and `yellow`. At most, we can use a full set of web colors.

alexladeira:
Hi, just a jellyfin user here. I know is not very polite to ask for ETA, but... Can we have someone looking into this?

Madh93:
My apologies for the slow reply @dmitrylyzo @thornbill 

I've replaced the text field by a basic HTML5 color picker. It works great but I don't like how it's show.

Firefox:
imagen

Chrome:
imagen

dmitrylyzo (commented):
Much better.

In webOS 1.2 it displays correctly, but the color selection doesn't work. We could show `actionSheet` with a list of predefined colors.

Madh93:
@dmitrylyzo I've added an `actionSheet` with a list of predefined colors that it must be shown in webOS and related devices. In any other case, browsers will show the color picker.

WebOS 6.0:

webos
webos2

Firefox:

browser

src/components/subtitlesettings/subtitlesettings.js
```diff
@@ -201,6 +204,10 @@ function embed(options, self) {
                 sliderVerticalPosition.classList.add('focusable');
                 sliderVerticalPosition.enableKeyboardDragging();
             }, 0);
+
+            // Replace color picker
+            options.element.querySelector('#inputTextColor').closest('div').classList.add('hide');
+            options.element.querySelector('#selectTextColor').closest('div').classList.remove('hide');
```
dmitrylyzo, line 210: no `closest` in webOS 1.2
```suggestion
            dom.parentWithTag(options.element.querySelector('#inputTextColor'), 'DIV').classList.add('hide');
            dom.parentWithTag(options.element.querySelector('#selectTextColor'), 'DIV').classList.remove('hide');
```

dmitrylyzo (commented):
I was thinking to listen the `click` event and show `actionSheet`, but your way also fine.

My suggestions (other than `closest`) probably require discussion.

src/components/subtitlesettings/subtitleappearancehelper.js
```diff
@@ -62,9 +62,34 @@ function getTextStyles(settings, preview) {
...
 
-    const textColor = settings.textColor || '#ffffff';
-    if (textColor) {
-        list.push({ name: 'color', value: textColor });
+    const hexTextColor = settings.hexTextColor || '#ffffff';
+
+    switch (settings.textColor || 'white') {
+        case 'yellow':
+            list.push({ name: 'color', value: '#ffff00' });
+            break;
```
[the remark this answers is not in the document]
Madh93: Nice approach I didn't know :man_facepalming:

src/components/subtitlesettings/subtitlesettings.js
Madh93: Oh thank you! I tried to install the webOs 1.2 emulator with no luck so I using the webOS 6.0 simulator instead.

garret:
Sorry to ask the obvious but has this PR been accepted into Jellyfin stable? 
I am running the latest docker container image from linuxserver and have also the latest android app version but can't see the possibility to choose a color for the subtitles 🙈

nielsvanvelzen:
This PR is not in 10.8.x, it will be released with 10.9.

bfg100k:
> This PR is not in 10.8.x, it will be released with 10.9.

Any idea when 10.9 will drop and/or why this PR cannot be included in 10.8.x? Being able to change the subtitle color globally will make a big difference to those HDR videos!

nielsvanvelzen:
Only bugfixes go into current releases, everything else will go in the next version. We don't have an ETA, could be a few weeks or multiple months.

YoungbloodTheOne:
please don't forget gray color (#888888) in the list, for oled TVs , this avoid screen burn

hajosattila:
> please don't forget gray color (#888888) in the list, for oled TVs , this avoid screen burn

+1

PR #3551, opened 2022-04-07
Fix url search in hash

thornbill:
**Changes**
Our urls are weird and include search parameters as part of the hash portion of the url. I kind of broke this in https://github.com/jellyfin/jellyfin-web/pull/3548...

**Issues**
N/A


dmitrylyzo (approved):
Seems to work as intended.

PR #3552, opened 2022-04-08
Collection: Display total duration

doc-code-hub:
## Edit

The PR has been converted into a Draft as it was agreed that it is better if the total RunTimeTicks of a collection are calculated server-side, so that the frontend only has to convert them to readable format and display them.

### Edit+
I will complete the PR, if necessary (
_working on my [PR](https://github.com/jellyfin/jellyfin/pull/7661) on jellyfin-server seems unnecessary, but I intend to close this draft only when the that one is completed_ :)
) , once the [server side developments](https://github.com/jellyfin/jellyfin/pull/7661) are completed.

-----------

**Changes**

In the **_itemDetails_** page I added the total duration of a BoxSet, obtained from the RunTimeTicks of the children.

When all the children of a BoxSet are retrieved, I execute a new function that deals with:
- Calculate the total ticks of all children
- Get the DisplayDuration
- Add it to the DOM with a transition (in .itemMiscInfo-secondary)

If a collection contains even just one element that does not have a RunTimeTicks, rendering is avoided (e.g. with boxset inside boxset) in order to show the total duration only if it's perfect. 

_I added the transition due to the fact that informations about children are retrieved in a promise, so they come later, otherwise causing an instant unpleasant change of position._

---
_Here are two videos of the animation_

https://user-images.githubusercontent.com/61033694/162517262-e1fb3547-bbaa-4fad-85b3-ef5a2289204e.mp4

https://user-images.githubusercontent.com/61033694/162517390-30692813-a321-467e-826c-78d2179e8712.mp4

**Issues**
I have not found any open issues regarding the total duration of a BoxSet

---
First time working on my jellyfin-web fork!
_If it needs changes / corrections, I am available_

### DIFF
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -1099,7 +1099,7 @@ function renderMiscInfo(page, item) {
             interactive: true
         });
 
-        if (miscInfo.innerHTML && item.Type !== 'SeriesTimer') {
+        if ((miscInfo.innerHTML && item.Type !== 'SeriesTimer') || item.Type === 'BoxSet') {
             miscInfo.classList.remove('hide');
         } else {
             miscInfo.classList.add('hide');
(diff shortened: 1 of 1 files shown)


thornbill:
I think this is a nice feature, but imho it should probably be handled server side such that the API returns RunTimeTicks at the collection level.

doc-code-hub:
@thornbill, 
I completely agree!

- It would be much more comfortable
- It would make the information available immediately
- It would allow to always have a total effective duration, regardless of the type of content inside (also solving the problem of collection within collection).

In relation to what you suggest I am thinking of:

- Open an issue on jellyfin-server to add the RunTimeTicks information on the collections (and possibly for other types of content from which the data may be missing, if any)
- Convert this PR into a Draft
- Follow the developments on jellyfin-server
- Repair my PR for the frontend implementation (or close the PR if the data should be automatically shown once updated on the server side)

thornbill:
That sounds good to me!

PR #3553, opened 2022-04-09
Refactor NavigationDrawer to Es6-class

grafixeyehero:


src/libraries/navdrawer/navdrawer.js
```diff
@@ -3,358 +3,369 @@
...
                 passive: true
             });
-            dom.addEventListener(mask, 'touchmove', onBackgroundTouchMove, {});
-            dom.addEventListener(mask, 'touchend', onBackgroundTouchEnd, {
+            dom.addEventListener(this.mask, 'touchmove', this.onBackgroundTouchMove.bind(this), {});
+            dom.addEventListener(this.mask, 'touchend', this.onBackgroundTouchEnd.bind(this), {
                 passive: true
             });
-            dom.addEventListener(mask, 'touchcancel', onBackgroundTouchEnd, {
+            dom.addEventListener(this.mask, 'touchcancel', this.onBackgroundTouchEnd.bind(this), {
```
dmitrylyzo, line 362: `bind`s - as above.

src/libraries/navdrawer/navdrawer.js
```diff
@@ -3,358 +3,369 @@
...
 
         if (!options.disableMask) {
-            mask = document.createElement('div');
-            mask.className = 'tmla-mask hide';
-            document.body.appendChild(mask);
-            dom.addEventListener(mask, dom.whichTransitionEvent(), onMaskTransitionEnd, {
+            this.mask = document.createElement('div');
+            this.mask.className = 'tmla-mask hide';
+            document.body.appendChild(this.mask);
+            dom.addEventListener(this.mask, dom.whichTransitionEvent(), this.onMaskTransitionEnd, {
```
dmitrylyzo, line 209: Need to bind (in the constructor) because `this` in the function will be `this.mask` (`div`) instead of navdrawer.

src/libraries/navdrawer/navdrawer.js
```diff
@@ -3,358 +3,369 @@
...
+                        this._edgeSwipeEnabled = false;
+                        dom.removeEventListener(this.edgeContainer, 'touchstart', this.onEdgeTouchStart.bind(this), {
                             passive: true
                         });
-                        dom.removeEventListener(edgeContainer, 'touchend', onEdgeTouchEnd, {
+                        dom.removeEventListener(this.edgeContainer, 'touchend', this.onEdgeTouchEnd.bind(this), {
                             passive: true
                         });
-                        dom.removeEventListener(edgeContainer, 'touchcancel', onEdgeTouchEnd, {
+                        dom.removeEventListener(this.edgeContainer, 'touchcancel', this.onEdgeTouchEnd.bind(this), {
```
dmitrylyzo, line 324: `bind`s - as above.

src/libraries/navdrawer/navdrawer.js
```diff
@@ -3,358 +3,369 @@
...
+            dom.addEventListener(this.mask, 'touchcancel', this.onBackgroundTouchEnd.bind(this), {
                 passive: true
             });
         }
 
-        self.clickMaskClose();
-    };
-
-    return new TouchMenuLA();
+        this.mask.addEventListener('click', this.clickMaskClose.bind(this));
```
dmitrylyzo, line 367: I believe that this `bind` can be left as is, because it won't be removed at runtime. But, at the same time, can be replaced as others.

dmitrylyzo (commented):
After a quick look

src/libraries/navdrawer/navdrawer.js
```diff
@@ -3,358 +3,373 @@
...
                 }
             }
         }
-    };
+    }
 
-    TouchMenuLA.prototype.initialize = function () {
-        options = Object.assign(defaults, options || {});
+    initialize() {
+        const options = Object.assign(this.defaults, this.options || {});
```
dmitrylyzo, line 337: We need to store the new options. Otherwise `handleSize` is `undefined`.
```suggestion
        const options = Object.assign({}, this.defaults, this.options || {});
        this.options = options;
```
Assignment to an empty object, so as not to change `defaults`. The latter can also be extracted as const.

src/libraries/navdrawer/navdrawer.js
```diff
@@ -3,358 +3,373 @@
...
 
-    function onMaskTransitionEnd() {
-        const classList = mask.classList;
+    onMaskTransitionEnd = () => {
+        const classList = this.mask.classList;
 
         if (!classList.contains('backdrop')) {
             classList.add('hide');
         }
-    }
```
dmitrylyzo, line 156: It looks fine, but I can't get it to be called.
We can probably remove this. Maybe not now.

grafixeyehero: this was called when opacity changed
```
.tmla-mask.backdrop {
    opacity: 1;
}
```

dmitrylyzo (approved):
We could improve the code even more (like extracting the defaults, making methods private), but I think migration to React is preferable.

PR #3556, opened 2022-04-11
Remove duplicate appRouter.param function

thornbill:
**Changes**
Removes `appRouter.param` function that is functionally a duplicate of the `getParameterByName` function in url utils.

I did a search and nothing seems to be using this function via the window global, so it should be safe to remove entirely.

**Issues**
N/A


dmitrylyzo (approved):
LGTM

PR #3560, opened 2022-04-12
Move backdrop transparency method to component

thornbill:
**Changes**
* Moves the method for updating the backdrop transparency to the backdrop component.
* Deprecates the `setTransparency` method in the app router until it can be removed from JMP.

**Issues**
N/A

### DIFF
--- a/src/utils/dashboard.js
+++ b/src/utils/dashboard.js
@@ -7,6 +7,7 @@ import baseConfirm from '../components/confirm/confirm';
 import globalize from '../scripts/globalize';
 import * as webSettings from '../scripts/settings/webSettings';
 import datetime from '../scripts/datetime';
+import { setBackdropTransparency } from '../components/backdrop/backdrop';
 import DirectoryBrowser from '../components/directorybrowser/directorybrowser';
 import dialogHelper from '../components/dialogHelper/dialogHelper';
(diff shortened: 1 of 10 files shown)


src/utils/dashboard.js
```diff
@@ -234,7 +235,8 @@ const Dashboard = {
     datetime,
     DirectoryBrowser,
     dialogHelper,
-    itemIdentifier
+    itemIdentifier,
+    setBackdropTransparency
```
thornbill, line 239: I'm not a fan of the `Dashboard` global as a whole, but this seems like the best place for now? I would be  curious what others think about this one.

dmitrylyzo (approved):

iwalton3:
It looks like the Dashboard component needs to be added here to be accessible from JMP plugins: https://github.com/jellyfin/jellyfin-web/blob/1e2f52a28edafeae6441d5e0393488e4c1c5cafb/src/components/pluginManager.js#L79

thornbill:
@iwalton3 the Dashboard component should be available on the window/global scope, but I do think it makes sense to include it in the utilities we use to initialize the plugins... that probably deserves an issue so we don't forget it :sweat_smile:

iwalton3:
@thornbill here is the follow-up for this: https://github.com/jellyfin/jellyfin-web/pull/4373

PR #3563, opened 2022-04-13
Use history for app router

thornbill:
**Changes**
* Removes page.js
* Adds history library for use as router
* Removes unused methods from appRouter

Using history would allow us to have shared routing state with react-router so we can begin using it for routing in react contexts :eyes: 

Minimal proof of concept of react-router routes alongside all the current routes can be seen [here](https://github.com/thornbill/jellyfin-web/commit/29365d51b087eefd6ef3c2f83a8c8cf695966672).

This will require **a lot** of testing...

**Issues**
N/A


src/components/appRouter.js
```diff
@@ -33,56 +34,22 @@ class AppRouter {
     startPages = ['home', 'login', 'selectserver'];
 
     constructor() {
-        // WebKit fires a popstate event on document load
```
thornbill, line 36: I don't _think_ we should hit this issue now since we are just listening to the `history` library events, but it should probably be tested.

dmitrylyzo: I checked the state of `isBack` on webOS 1.2 (closest to Tizen 2.3) and it seems to be valid on the first load - `false` (`action` is `undefined`).

For reference, the old `popstate` approach has `isBack: true` without `!loaded`.

We need Tizen 2.x to be sure, but either way it should be fixed in `history`.

dmitrylyzo (approved):

PR #3567, opened 2022-04-15
Convert Home to react

grafixeyehero:


src/components/pages/HomeView.tsx
```diff
@@ -0,0 +1,184 @@
...
+type ControllerProps = {
+    onResume: (
+        options: OptionsProps
+    ) => void;
+    refreshed: boolean;
+    onPause: () => void;
+    destroy: () => void;
+}
+
+class HomeView extends Component <IProps> {
```
thornbill, line 29: Is there any reason to use a class component instead of a function component for this?

grafixeyehero: No just implemented the way before.

src/components/viewContainer.js
```diff
@@ -236,7 +236,7 @@ import Dashboard from '../utils/dashboard';
     const mainAnimatedPages = document.querySelector('.mainAnimatedPages');
     let allPages = [];
     let currentUrls = [];
-    const pageContainerCount = 3;
+    const pageContainerCount = 1;
```
thornbill, line 239: This should not be necessary should it?

grafixeyehero: the reason I did it b/c of the header tabs. for instance when you change for page to page the tab disappear. 

https://user-images.githubusercontent.com/32230989/165391872-40e08e81-c107-41df-bb04-45fac85b156b.mp4

src/components/pages/HomeView.tsx
grafixeyehero: @thornbill update to function components https://github.com/jellyfin/jellyfin-web/pull/3567/commits/c1ea22268bc956d34e491059ef10da1ead28d68c

src/components/viewContainer.js
grafixeyehero: @thornbill I am thinking moving **header tabs** from **skin header** class to the **page view**?

grafixeyehero: @thornbill updated to react router https://github.com/jellyfin/jellyfin-web/pull/3567/commits/4699e9de608096c889e16f7b7b510d7147e6f746

src/components/Page.tsx
```diff
@@ -63,6 +67,8 @@ const Page: FunctionComponent<PageProps & HTMLAttributes<HTMLDivElement>> = ({
             data-title={title}
             data-backbutton={`${isBackButtonEnabled}`}
             data-menubutton={`${isMenuButtonEnabled}`}
+            data-dom-cache={`${isDomCacheEnabled}`}
+            data-backdroptype={`${backDropType}`}
```
thornbill, line 71: This doesn't need to be wrapped in a string since it is already a string.

thornbill (changes requested):
Sorry this had been sitting so long. I've been looking at other things for a bit. Just one minor comment and I'll do some testing so we can get this in.

src/components/Page.tsx
grafixeyehero: @thornbill updated https://github.com/jellyfin/jellyfin-web/pull/3567/commits/76eff104d8a40e66a32ba07e9cc6b0f75ab16d6e

src/components/Page.tsx
```diff
@@ -61,8 +65,10 @@ const Page: FunctionComponent<PageProps & HTMLAttributes<HTMLDivElement>> = ({
             data-role='page'
             className={`page ${className}`}
             data-title={title}
-            data-backbutton={`${isBackButtonEnabled}`}
-            data-menubutton={`${isMenuButtonEnabled}`}
+            data-backbutton={isBackButtonEnabled}
+            data-menubutton={isMenuButtonEnabled}
+            data-dom-cache={isDomCacheEnabled}
```
thornbill, line 70: Is this actually used for anything? I did a bit of searching but I didn't see anywhere that was reading the `data-dom-cache` attribute. :man_shrugging:

thornbill (changes requested):
Looks great! I noticed the back button was showing up because the `appRouter` would disable it based on the route types it defined. I think we should just disable it explicitly by passing in the prop.

Otherwise it worked fine in testing besides a bug with loading state in the `ConnectionRequired` component, but that is unrelated to your changes. I'll just open a separate PR for it. :rocket:

src/routes/home.tsx
```diff
@@ -0,0 +1,177 @@
...
+            onViewDestroy();
+        };
+    }, [onPause, onResume, onTabChange, onViewDestroy]);
+
+    return (
+        <div ref={element}>
+            <Page
+                id='indexPage'
+                className='mainAnimatedPage homePage libraryPage allLibraryPage backdropPage pageWithAbsoluteTabs withTabs'
+                isDomCacheEnabled= {true}
```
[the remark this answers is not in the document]
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/3567/commits/58513092c3ca370f45d18e5b9da97af7cbefaa21

src/components/Page.tsx
grafixeyehero: yah your are right. I couldn't find it to so I removed it https://github.com/jellyfin/jellyfin-web/pull/3567/commits/328afb0db8dc7892ddd7522d98d6c89aa1379fb0

thornbill (approved):

Issue #3326, opened 2020-11-07
Last subtitle line dont show up


EduFdezSoy:
The last line of the subtitles is not showing, only the last line. 

Tested in:
- Firefox: cuts the last line
- Android (exoplayer): working as expected
- VLC: working as expected  
  
The files tested are all **mkv** with **ass** subtitles inside  

I checked the subtitles but cant see anything wrong inside, here is the last line:  
`Dialogue: 0,0:10:25.58,0:10:29.17,Italics,Honda,0000,0000,0000,,That is the task of a comic bookstore.`  

---
**System:**
 - OS: Debian
 - Virtualization: none
 - Clients: Browser
 - Browser: Firefox 82.0.2 (64-bit)
 - Jellyfin Version: 10.6.4
 - Playback: DirectPlay  
 - Reverse Proxy: none
 - Base URL: none
 - Networking: host, local network under a private wiregard vpn 
 - Storage: local

dmitrylyzo:
> Issue is probably caused by the JavascriptSubtitlesOctopus library.

Yes, this can be fixed with https://github.com/jellyfin/JavascriptSubtitlesOctopus/pull/21
There is a better fix for tag parser in the upstream, but that PR has stalled.

Feynt:
Reporting in, can confirm this is still an ongoing issue.  Inserting an extra garbage line into subtitles will allow me to read the final proper line in the videos.

dmitrylyzo:
Fixed by #3644

Issue #2291, opened 2021-01-01
[10.7] Transparent Subtitles


TheNamelessWonderer:
**Describe The Bug**
Some subtitles are transparent in 10.7 making them hard to read.

**Steps To Reproduce**

1. Play a video
2. Watch as some subtitles are transparent

**Expected Behavior**

The Sub titles shouldn't be transparent

**Logs**
10:6 FFMPEG command: `ffmpeg -ss 00:01:39.000 -noaccurate_seek -f matroska,webm -i file:"/path/to/file" -map_metadata -1 -map_chapters -1 -threads 0 -map 0:0 -map 0:1 -codec:v:0 libx264 -pix_fmt yuv420p -preset veryfast -crf 23 -maxrate 7834386 -bufsize 15668772 -profile:v high -level 4.1 -x264opts:0 subme=0:me_range=4:rc_lookahead=10:me=dia:no_chroma_me:8x8dct=0:partitions=none  -force_key_frames:0 "expr:gte(t,99+n_forced*3)" -g 72 -keyint_min 72 -sc_threshold 0 -vf "scale=trunc(min(max(iw\,ih*dar)\,1920)/2)*2:trunc(ow/dar/2)*2,subtitles='/Volumes/sda1/TV Shows/Jojo'\\\''s Bizarre Adventure/Season 3/[Some-Stuffs]_Jojo'\\\''s_Bizarre_Adventure_Diamond_is_Unbreakable_27_(1920x1080_Blu-Ray_FLAC)_[8B33D737].mkv:si=0'" -start_at_zero -vsync -1 -codec:a:0 libmp3lame -ac 2 -ab 192000  -copyts -avoid_negative_ts disabled -f hls -max_delay 5000000 -hls_time 3 -individual_header_trailer 0 -hls_segment_type mpegts -start_number 33 -hls_segment_filename "/Users/user/.local/share/jellyfin/transcodes/f5f139122bdf44f778c29c36b4d383a8%d.ts" -hls_playlist_type vod -hls_list_size 0 -y "/Users/user/.local/share/jellyfin/transcodes/f5f139122bdf44f778c29c36b4d383a8.m3u8"`

[...]

PR #3589, opened 2022-04-24
Fix various issues with SSA subtitle rendering

Deathspike:
**Changes**
* Removes `renderMode` & `renderAhead` from `SubtitlesOctopusRender`.
* Removes options that are equal to the default for readability from `SubtitlesOctopusRender`.

**Rationale**
This change effectively disables blend rendering in JavascriptSubtitlesOctopus. This rendering mode problematic, which seems to be even more severe in Jellyfin's branch with all of its custom modifications and regressions. For full reference, `jellyfin-vue` exhibits no issues whatsoever and uses `libass-wasm` straight from `npm` (thus https://github.com/libass/JavascriptSubtitlesOctopus) and without using `blend` mode. I would honestly want to see `jellyfin-web` do the same, but I feel that should require some discussion first.

**Issues**
* https://github.com/jellyfin/jellyfin-web/issues/3326
* https://github.com/jellyfin/jellyfin-web/issues/2291

### DIFF
--- a/src/plugins/htmlVideoPlayer/plugin.js
+++ b/src/plugins/htmlVideoPlayer/plugin.js
@@ -1082,12 +1082,10 @@ function tryRemoveElement(elem) {
                 timeOffset: (this._currentPlayOptions.transcodingOffsetTicks || 0) / 10000000,
 
                 // new octopus options; override all, even defaults
-                renderMode: 'blend',
                 libassMemoryLimit: 40,
                 libassGlyphLimit: 40,
                 targetFps: 24,
-                prescaleTradeoff: 0.8,
-                renderAhead: 90
+                prescaleTradeoff: 0.8
             };
             import('libass-wasm').then(({default: SubtitlesOctopus}) => {
                 apiClient.getNamedConfiguration('encoding').then(config => {
--- a/src/plugins/htmlVideoPlayer/plugin.js
+++ b/src/plugins/htmlVideoPlayer/plugin.js
@@ -1083,14 +1083,10 @@ function tryRemoveElement(elem) {
 
                 // new octopus options; override all, even defaults
                 renderMode: 'blend',
-                dropAllAnimations: false,
                 libassMemoryLimit: 40,
                 libassGlyphLimit: 40,
                 targetFps: 24,
                 prescaleTradeoff: 0.8,
-                softHeightLimit: 1080,
-                hardHeightLimit: 2160,
-                resizeVariation: 0.2,
                 renderAhead: 90
             };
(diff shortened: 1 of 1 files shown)


dmitrylyzo:
Upstream is going to use `blend` mode (now `wasm`, unreleased) by default, and it works just fine.

Render ahead mode is supposed to improve the timing on low-end devices (TV), but it has a bug. There is a PR that fixes it, but we wanted to port the features from our fork to upstream, so that we can use npm module (and fewer projects to maintain), but I failed 😞

Deathspike:
Maybe the upstream `libass-wasm` works fine, but the Jellyfin branch sure doesn't. The sister project `jellyfin-vue` works great (using 4.0.0 released in 2020). I don't see any reason `jellyfin-web` should continue to rely on a package that is known to have issues where [it's own branch has a PR request that supposedly fixes things isn't being merged](https://github.com/jellyfin/JavascriptSubtitlesOctopus/pull/21), and upstream [rejects the pull request](https://github.com/libass/JavascriptSubtitlesOctopus/pull/111#issuecomment-1022707471) for reasons pointing to improper atomic commit and *known issues*.

Since those PRs have been stale for ~6 months, I don't believe that the next web release should continue to suffer from these known issues. Look, I truly do appreciate that you want to improve `libass-wasm` for older/slower televisions, but if a merge is not going to happen in its current form, then let's make sure that any modern televisions, mobile devices, and desktop device isn't held hostage by this odd dependency and incorrect subtitle rendering. 

This PR will at least have those players work well.

dmitrylyzo:
`blend` and render ahead are different things. You can zero the `renderAhead` property and get that unreleased mode (it will be the default when they release), and it works better than `normal` mode (current default).

The PR in our fork is ready (the parser can be replaced with their version though).

I'm more inclined to release the scoped module and not rely on slow upstream. I have prepared `new-master` branch in my fork. The questions are versioning and license (do we have the right to release it in this form?).

Deathspike:
Licensing is a non-issue. Everything is MIT, meaning we can do whatever we want with the source code. What, *concretely*, works 'better' is your repository and your `new-master` branch when compared with the upstream libass-wasm? Because, currently, it feels very much like the `libass-wasm` used in `jellyfin-web` is a half-finished tinker project ran by one person which introduces all kinds of issues that don't exist in upstream (or `jellyfin-vue`). And I feel that shouldn't be the case in the de-facto web interface used for almost all Jellyfin clients.

What, *concretely*, needs to be done for your work to be finished, stable, and ready for testing? I'm also still very interested what other people think of this.

dmitrylyzo:
I suppose with #3644 this becomes unnecessary.

PR #3596, opened 2022-04-26
Convert Movies page to react

grafixeyehero:


grafixeyehero:
https://github.com/jellyfin/jellyfin-web/pull/3792

PR #3601, opened 2022-04-27
Fix XSS in repositories list

thornbill:
**Changes**
Fixes a XSS vulnerability in the repositories list in the admin dashboard.

**Issues**
N/A

### DIFF
--- a/src/controllers/dashboard/plugins/repositories/index.js
+++ b/src/controllers/dashboard/plugins/repositories/index.js
@@ -42,39 +42,64 @@ function saveList(page) {
 }
 
 function populateList(options) {
-    let html = '';
+    const paperList = document.createElement('div');
+    paperList.className = 'paperList';
 
-    html += '<div class="paperList">';
-    for (let i = 0; i < options.repositories.length; i++) {
-        html += getRepositoryHtml(options.repositories[i]);
-    }
+    options.repositories.forEach(repo => {
+        paperList.appendChild(getRepositoryElement(repo));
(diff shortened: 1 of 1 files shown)


joshuaboniface (approved):

dmitrylyzo (commented):
The only problem I encountered:
In Chrome 79 `is` attribute is ignored when constructing via DOM, via HTML it is fine.
But the library menu works fine even being constructed via DOM :man_shrugging: 
In Chrome 100, Firefox 72 and Firefox 99 everything works fine.

thornbill:
@dmitrylyzo could it possibly be some other attribute being set here causing issues?

dmitrylyzo:
> @dmitrylyzo could it possibly be some other attribute being set here causing issues?

I was wrong - the link in the library menu doesn't work via DOM either.
https://github.com/jellyfin/jellyfin-web/blob/89a25b1360248d1564a86b1e4871ce982bd6cff4/src/scripts/libraryMenu.js#L673

This also doesn't work: https://github.com/jellyfin/jellyfin-web/blob/89a25b1360248d1564a86b1e4871ce982bd6cff4/src/controllers/session/login/index.js#L294

I wonder if Chrome 79 was a transitional version?

thornbill:
Looks like we should be passing that as an option to `createElement`. I'll open a PR for that in a bit. https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements

PR #3606, opened 2022-04-29
Improve subs

dannymichel:
**Changes**
Improve subtitles by changing default dont to universal system font stack(the same one github uses) and improving stroke for ``uniform`` option.


cvium:
This PR is broken. It looks like you've branched out from `master` when you should've branched out from the release branch.

PR #3611, opened 2022-05-02
Add Unused Languages to Web

rhld16:
**Changes**

I started translating the Welsh strings using Weblate but realised my translations were not appearing on the web client. Upon further inspection I realised that many languages were not being loaded even if added to #2280. I've added any unused language which has a valid date-fns locale: Welsh, Estonian, Latvian and Norwegian.

An issue still remains that some languages in the 'Display Language' aren't functional when selected.


thornbill:
Are these languages present in the 10.8 release branch? If so, I would recommend we change this to target the 10.8 release.

rhld16:
These languages were added to the dropdown at 2861dd which was a part of v10.8.0-beta2

thornbill:
Would you be comfortable rebasing this on the `release-10.8.z` branch so we can get this in the 10.8 release? You may need to create a new branch and cherry-pick your changes to it. Thanks!

rhld16:
Well I messed that one up, will close and open a new issue with this done correctly...

dmitrylyzo:
That was easy to fix without opening a new PR.
You just needed to rebase the branch on release, leaving only your commits, and change the target with the `Edit` button next to the PR title.

PR #3612, opened 2022-05-03
Remove rounded edges from cards

dannymichel:
**Changes**
Remove rounded edges from cards and revert back to jellyfin look rather than emby
Screen Shot 2022-05-03 at 2 41 55 AM


thornbill:
The team prefers the rounded corners... which is why they were added.

Numerous media apps and services use rounded corners including from a quick check Hulu, Netflix (on iOS), Apple TV, Disney+, Movies Anywhere, Pluto, Infuse. What Emby does or doesn't do is completely irrelevant here. This is not some artifact that was leftover from the time of our fork 3.5 years ago.

If you do not like the rounded corners, they should be rather trivial to remove with custom css.

dont-like-that

PR #3620, opened 2022-05-06
Update dialog history handling

thornbill:
**Changes**
This updates our dialog handling to use history state instead of path changes. This fixes the issue where refreshing the page when a dialog is open breaks the app and makes our dialogs compatible with React Router.

**Issues**
N/A?

### DIFF
--- a/src/components/dialogHelper/dialogHelper.js
+++ b/src/components/dialogHelper/dialogHelper.js
@@ -1,4 +1,4 @@
-import { appRouter } from '../appRouter';
+import { history } from '../appRouter';
 import focusManager from '../focusManager';
 import browser from '../../scripts/browser';
 import layoutManager from '../layoutManager';
@@ -49,17 +49,19 @@ import '../../assets/css/scrollstyles.scss';
         self.originalUrl = window.location.href;
         const activeElement = document.activeElement;
         let removeScrollLockOnClose = false;
+        let unlisten;
 
-        function onHashChange() {
-            const isBack = self.originalUrl === window.location.href;
+        function onHashChange({ location }) {
+            const dialogs = location.state?.dialogs || [];
+            const shouldClose = !dialogs.includes(hash);
 
-            if (isBack || !isOpened(dlg)) {
-                window.removeEventListener('popstate', onHashChange);
+            if ((shouldClose || !isOpened(dlg)) && unlisten) {
+                unlisten();
             }
 
-            if (isBack) {
+            if (shouldClose) {
                 self.closedByBack = true;
-                closeDialog(dlg);
+                close(dlg);
             }
         }
 
@@ -68,7 +70,7 @@ import '../../assets/css/scrollstyles.scss';
                 self.closedByBack = true;
                 e.preventDefault();
                 e.stopPropagation();
-                closeDialog(dlg);
+                close(dlg);
             }
         }
 
@@ -77,7 +79,9 @@ import '../../assets/css/scrollstyles.scss';
                 inputManager.off(dlg, onBackCommand);
             }
 
-            window.removeEventListener('popstate', onHashChange);
+            if (unlisten) {
+                unlisten();
+            }
 
             removeBackdrop(dlg);
             dlg.classList.remove('opened');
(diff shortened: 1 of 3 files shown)


grafixeyehero:
when i tested most dialog work fine. 
Issues:
metadataEditor it open at first time but when close it and  reopen again has issue  no error shows in console

grafixeyehero:
From Closing btn in upper right corner

src/components/dialogHelper/dialogHelper.js
```diff
@@ -144,9 +148,19 @@ import '../../assets/css/scrollstyles.scss';
         animateDialogOpen(dlg);
 
         if (isHistoryEnabled(dlg)) {
-            appRouter.show(`/dialog?dlg=${hash}`, { dialogId: hash });
+            const state = history.location.state || {};
+            const dialogs = state.dialogs || [];
+            dialogs.push(hash);
+
+            history.push(
+                `${history.location.pathname}${history.location.search}`,
```
dmitrylyzo, line 156: `Relative pathnames are not supported in hash history.push("!/home.html")` warning. Could this lead to that behavior?

src/components/dialogHelper/dialogHelper.js
```diff
@@ -86,10 +90,10 @@ import '../../assets/css/scrollstyles.scss';
                 document.body.classList.remove('noScroll');
             }
 
-            if (!self.closedByBack && isHistoryEnabled(dlg)) {
-                const state = window.history.state || {};
-                if (state.dialogId === hash) {
-                    appRouter.back();
+            if (isHistoryEnabled(dlg)) {
+                const state = history.location.state || {};
+                if (state.dialogs?.length > 0 && state.dialogs[0] === hash) {
```
dmitrylyzo, line 95: Maybe need the last one?

src/components/dialogHelper/dialogHelper.js
dmitrylyzo: `appRouter.show` is promised, so the previous `back` is handled before opening a new dialog. This is only for history, the dialog is already there, iirc.

src/components/dialogHelper/dialogHelper.js
thornbill: Yeah this was wrong. I was pushing new dialogs to the end of the array, but checking the first entry. Should be fixed now.

grafixeyehero:
@thornbill possible fix for metadataEditor dialog i will create PR

grafixeyehero:
@thornbill here is https://github.com/grafixeyehero/jellyfin-web/pull/214

thornbill:
ooo very nice @grafixeyehero! I will do some testing with your change. Thanks!

src/components/dialogHelper/dialogHelper.js
```diff
@@ -86,10 +88,14 @@ import '../../assets/css/scrollstyles.scss';
...
-                const state = window.history.state || {};
-                if (state.dialogId === hash) {
-                    appRouter.back();
+            if (isHistoryEnabled(dlg)) {
+                const state = history.location.state || {};
+                if (state.dialogs?.length > 0) {
+                    if (state.dialogs[state.dialogs.length - 1] === hash) {
+                        history.back();
+                    } else if (state.dialogs.includes(hash)) {
+                        console.info('[dialogHelper] dialog "%s" was closed, but is not the last dialog opened', hash);
```
dmitrylyzo, line 97: Maybe `warn`?

This shouldn't happen, but shouldn't the dangling hash be removed? - If `removeOnClose: true` (most of the time it is), the dialog is automatically removed from DOM.

thornbill: Yeah good call. I have updated it to warn and replace the current history state to exclude the closed dialog hash.

dmitrylyzo (approved):
AFAIK, `isHistoryEnabled` is currently always `true`.

We probably need to rewind the history on refresh (to clear `dialogs`), but this requires testing.

thornbill:
@dmitrylyzo I believe I found one place in the playback OSD where `isHistoryEnabled` is false depending on the browser when I looked at it previously.

PR #3621, opened 2022-05-08
Append year to title of content in video playback.

jesse231:
**Changes**
This adds the premiere year to the title in the video player which was requested here: https://features.jellyfin.org/posts/1392/append-the-year-to-the-title-in-the-video-player.

**Issues**
N/A

### DIFF
--- a/src/controllers/playback/video/index.js
+++ b/src/controllers/playback/video/index.js
@@ -210,7 +210,18 @@ import { setBackdropTransparency, TRANSPARENCY_LEVEL } from '../../../components
                 itemName = parentName || '';
             }
 
-            LibraryMenu.setTitle(itemName);
+            // Display the item with its premiere date if it has one
+            let title = itemName;
+            if (item.PremiereDate) {
+                try {
+                    const year = datetime.parseISO8601Date(item.PremiereDate).getFullYear();
+                    title += ` (${year})`;
+                } catch (e) {
+                    console.error(e);
+                }
+            }
+
+            LibraryMenu.setTitle(title);
 
             const documentTitle = parentName || (item ? item.Name : null);
 


src/controllers/playback/video/index.js
```diff
@@ -209,8 +209,13 @@ import { setBackdropTransparency, TRANSPARENCY_LEVEL } from '../../../components
             if (!itemName) {
                 itemName = parentName || '';
             }
-
-            LibraryMenu.setTitle(itemName);
+            //Display the item with its premiere date if it has one
```
dmitrylyzo, line 212:
```suggestion

            // Display the item with its premiere date if it has one
```

src/controllers/playback/video/index.js
```diff
@@ -209,8 +209,13 @@ import { setBackdropTransparency, TRANSPARENCY_LEVEL } from '../../../components
...
+            if ('PremiereDate' in item) {
+                const date = datetime.parseISO8601Date(item.PremiereDate).getFullYear();
```
dmitrylyzo, line 214: `parseISO8601Date` is mostly used in the `try/catch` block because it throws an exception.

How about this?
```js
let title = itemName;

if ('PremiereDate' in item) {
    try {
        const year = datetime.parseISO8601Date(item.PremiereDate).getFullYear();
        title += ` ${year}`;
    } catch () {
        // print error
    }
}

LibraryMenu.setTitle(title);
```

jesse231: I see. I'll update the code now.

src/controllers/playback/video/index.js
dmitrylyzo: I don't know where to put this comment, but at least add a blank line before and a space after `//`.

jesse231: Oops, I meant to add that as well.

src/controllers/playback/video/index.js
```diff
@@ -210,7 +210,18 @@ import { setBackdropTransparency, TRANSPARENCY_LEVEL } from '../../../components
                 itemName = parentName || '';
             }
 
-            LibraryMenu.setTitle(itemName);
+            // Display the item with its premiere date if it has one
+            let title = itemName;
+            if ('PremiereDate' in item) {
```
dmitrylyzo, line 215: Check not only the presence, but also the value.
```suggestion
            if (item.PremiereDate) {
```

dmitrylyzo (approved):

PR #3623, opened 2022-05-09
Upgrade hls.js to 1.2.0

knackebrot:
Configured to behave similarly to 0.4.17, according to https://github.com/video-dev/hls.js/blob/v1.1.5/MIGRATING.md .

Contains lots of fixes and low-latency HLS support, which could be useful in the future.


ThibaultNocchi:
Hello,

We've pinned HLS.js to the latest pre v1 version cause of a seeking bug, see #3161
I've also lately tried to play with their latest version and different settings, but in vain. 

Maybe you should try to see if it's present with your PR?

knackebrot:
Hello,

I didn't encounter that error with the video mentioned in the issue nor with any of my XviD videos.

Shadowghost:
If it works maybe worth a backport too (builds seem to currently fail though)?

ThibaultNocchi:
I've been able to reproduce it with this PR, both in Chrome and FF. You gotta seek around a few times (I just forward the cursor by a few % each time) and at one time, it won't resume.

We can see in the network dev console that it loads the first .ts segment, then a bunch of segments 3 or 4 after. So I guess it misses a few of them in between. What's weird is that sometimes it handles that well.

Also, seeking again doesn't resume buffering, so you can only seek inside the current buffer. It'll infinitely load the same 2 segments.
print_screen_2022-05-10-08-04-28

knackebrot:
@ThibaultNocchi I've updated hls.js to 1.2.0, it contains multiple fixes related to fragment loading and buffering, could you test if you still experience this issue, since I can't reproduce it?

ThibaultNocchi:
I've seen the update, and they even closed the issue saying that it should be fixed. I'll try tomorrow :)

ThibaultNocchi:
Ok so I tried it, and I'm still getting those freezes. What I do is let the video play a few seconds, seek a few minutes ahead, wait for buffering, it plays, I seek again, etc. And at one time, even though the fragments are loaded, it will freeze. I just seek ahead again and it works again. So sometimes a random seek freezes.

So I have two questions:

1. Do you experience those? I did on Chrome and FF on Linux, Chrome on MacOS. Though Chrome on Android didn't seem to freeze.
2. Do random freezes that can be fixed by re-seeking warrant to stick on an old beta release of hls.js? Maybe a question for @thornbill

knackebrot:
It was kind of hard to reproduce for me, but after seeking back and forth in quick succession about 20 times, finally it stalled. I don't know if it's the same issue, but the current time gets set to 0:00 and I can see an exception in the console 
image
We could probably handle it by reinitializing Hls.js when it occurs.
@ThibaultNocchi How fast (in FPS) is your transcoding job running?

knackebrot:
It was kind of hard to reproduce for me, but after seeking back and forth in quick succession about 20 times, finally it stalled. I don't know if it's the same issue, but the current time gets set to 0:00 and I can see an exception in the console.
image
We probably could reinitialize hls.js when it occurs.
@ThibaultNocchi How fast (in FPS) is your transcoding job running?

ThibaultNocchi:
I don't have this error on my side. Also when I seek, I seek outside of the buffered zone. My transcode FPS is between 120 and 150 FPS.
I'm really wondering what's wrong on my side if nobody notices it haha

thornbill:
@ThibaultNocchi you said this update was working fine in Vue now right?

ThibaultNocchi:
Yes @thornbill it's been now a month I'm playing with it with the player, and I can confirm I've had 0 occurrences of that problem with latest HLS.js.

thornbill:
@knackebrot would you be able to fix the merge conflicts here and ensure we are updating to the latest hls.js release? Thanks!

PR #3629, opened 2022-05-11
Create option for bold subtitles

dannymichel:
Create option for bold subtitles
167627392-f1eab6a4-8787-406e-b727-f26d081319e7

### DIFF
--- a/src/components/subtitlesettings/subtitlesettings.template.html
+++ b/src/components/subtitlesettings/subtitlesettings.template.html
@@ -73,6 +73,13 @@
         </div>
 
         <div class="selectContainer">
+            <select is="emby-select" id="selectTextWeight" label="${LabelTextWeight}">
+                <option value="normal">${Normal}</option>
+                <option value="bold">${Bold}</option>
+            </select>
+        </div>
+
+        <div class="selectContainer">
             <select is="emby-select" id="selectFont" label="${LabelFont}">
                 <option value="">${Default}</option>
                 <option value="typewriter">${Typewriter}</option>
--- a/src/components/subtitlesettings/subtitlesettings.template.html
+++ b/src/components/subtitlesettings/subtitlesettings.template.html
@@ -74,7 +74,7 @@
 
         <div class="selectContainer">
             <select is="emby-select" id="selectTextWeight" label="${LabelTextWeight}">
-                <option value="normal">${Normal}</option>
+                <option value="">${Normal}</option>
                 <option value="bold">${Bold}</option>
             </select>
         </div>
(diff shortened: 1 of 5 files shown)


dannymichel:
Everything look ok @dmitrylyzo ?

dmitrylyzo (commented):
It is more or less fine.
The question is, if we only have 2 options, wouldn't it be better to use a checkbox?

The posts below are for the current ("select") version.

src/components/subtitlesettings/subtitlesettings.template.html
```diff
@@ -72,6 +72,13 @@ <h2 class="sectionTitle">
             </select>
         </div>
 
+        <div class="selectContainer">
+            <select is="emby-select" id="selectTextWeight" label="${LabelTextWeight}">
+                <option value="">${Normal}</option>
```
dmitrylyzo, line 77: I said "alternative" (pick one).
Let's go my way - "no implicit values".
```suggestion
                <option value="normal">${Normal}</option>
```

dannymichel: Yokes i think i messed up @dmitrylyzo did this close because i changed the name of the branch?

src/strings/en-us.json
```diff
@@ -918,6 +918,8 @@
     "LabelTextBackgroundColor": "Text background color:",
     "LabelTextColor": "Text color:",
     "LabelTextSize": "Text size:",
+    "LabelTextWeight": "Text weight:",
+    "Bold": "Bold:",
```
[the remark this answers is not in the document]
dannymichel: 337ccbc425b827d5dffa0dde6bc855e285ae46bc

src/components/subtitlesettings/subtitlesettings.template.html
dannymichel: dc6bfc8f364be1d02e0ef7e52e09b6edbbe3b655

dannymichel:
Reopened https://github.com/jellyfin/jellyfin-web/pull/3632

src/components/subtitlesettings/subtitlesettings.template.html
dmitrylyzo: > Yokes i think i messed up @dmitrylyzo did this close because i changed the name of the branch?

It seems so. When you rename a branch, a new one is created (a new commit reference) and the old one is removed, which closes the PR.

dannymichel: ok i reopened it here with your suggested commit https://github.com/jellyfin/jellyfin-web/pull/3632

PR #3636, opened 2022-05-13
Use base url prefixed path when fetching plugin images

cfal:
**Changes**
Fix the path for fetching plugin images when a base URL has been specified.

### DIFF
--- a/src/controllers/dashboard/plugins/installed/index.js
+++ b/src/controllers/dashboard/plugins/installed/index.js
@@ -2,6 +2,7 @@ import loading from '../../../../components/loading/loading';
 import libraryMenu from '../../../../scripts/libraryMenu';
 import dom from '../../../../scripts/dom';
 import globalize from '../../../../scripts/globalize';
+import { appRouter } from '../../../../components/appRouter';
 import * as cardBuilder from '../../../../components/cardbuilder/cardBuilder.js';
 import '../../../../components/cardbuilder/card.scss';
 import '../../../../elements/emby-button/emby-button';
@@ -70,7 +71,7 @@ function getPluginCardHtml(plugin, pluginConfigurationPages) {
     }
 
     if (plugin.HasImage) {
-        const imageUrl = ApiClient.getUrl(`/Plugins/${plugin.Id}/${plugin.Version}/Image`);
+        const imageUrl = ApiClient.getUrl(`${appRouter.baseUrl()}/Plugins/${plugin.Id}/${plugin.Version}/Image`);
         html += `<img src="${imageUrl}" style="width:100%" />`;
     } else {
         html += `<div class="cardImage flex align-items-center justify-content-center ${cardBuilder.getDefaultBackgroundClass()}">`;


src/controllers/dashboard/plugins/installed/index.js
```diff
@@ -70,7 +71,7 @@ function getPluginCardHtml(plugin, pluginConfigurationPages) {
     }
 
     if (plugin.HasImage) {
-        const imageUrl = ApiClient.getUrl(`/Plugins/${plugin.Id}/${plugin.Version}/Image`);
+        const imageUrl = ApiClient.getUrl(`${appRouter.baseUrl()}/Plugins/${plugin.Id}/${plugin.Version}/Image`);
```
dmitrylyzo, line 74: `getUrl` already takes `BaseUrl` into account.
On my test server, `BaseUrl` is `/jellyfin` and I see plugin images.

cfal: that's interesting. i have a base path set and see these failing requests every time i go to the plugin page:

```
GET /Plugins/022a3003993f45f1856587d12af2e12a/1.4.1.0/Image HTTP/1.1
GET /Plugins/4fe3201ed6ae4f2e8917e12bda571281/12.0.0.0/Image HTTP/1.1
```

dmitrylyzo: In the browser console, look for `ApiClient serverAddress:` and `Setting server address to`, the URL should contain BaseUrl.

cfal: yes, I can see that line, and it does contain the correct base URL. other pages also work correctly, eg `ApiClient.getJSON(ApiClient.getUrl('Devices'))` is able to fetch `<base path>/Devices`. wonder why it doesn't work on the plugins page.

dmitrylyzo: Place a breakpoint on this line and see what value `window.ApiClient._serverAddress` has.
Or add `console.log('serverAddress:', window.ApiClient._serverAddress);`before the `getUrl` call.

cfal: even at https://github.com/jellyfin/jellyfin-web/blob/master/src/controllers/dashboard/plugins/installed/index.js#L102 - the url is correct and includes the correct base path. but in the inspected html, I see `<img src="/Plugins/.." ..` without the correct base path, so something must be changing window.ApiClient from when L102 gets called, to `imageUrl` being set on L74. guessing it's `ServerConnection.setLocalApiClient` replacing the client or `ApiClient.serverAddress(..)` replacing the address somewhere.. but don't have time to investigate more for now.

in any case, it seems like this PR is probably the incorrect fix, so i'll close it.

cfal: i think this problem should be easy to reproduce if you have a base path by installing any plugin that provides an image?

it occurs on both chrome and firefox.

dmitrylyzo: > but in the inspected html, I see ` so something must be changing window.ApiClient from when L102 gets called, to `imageUrl` being set on L74.

That's why I suggested putting a breakpoint or placing the debug output at L74.

> guessing it's `ServerConnection.setLocalApiClient` replacing the client or `ApiClient.serverAddress(..)` replacing the address somewhere.. but don't have time to investigate more for now.

`ServerConnection.setLocalApiClient` is called twice - when connecting to the server and authenticating the user. Both calls are performed before loading the plugin list.

The only difference I see is that the image url parameter for `getUrl` starts with `/`.

PR #3642, opened 2022-05-15
Change CSS font-family by language preference

taku0:
**Changes**

- Update `lang` attribute of `html` element on user preference change.
- Choose appropriate `font-family` depending on `lang` attribute using attribute selectors in CSS.
- Add `Noto Sans TC`, the Traditional Chinese (Taiwan) variant.
- Fix event listener registration in `globalize.js`.

**Issues**

Fixes https://github.com/jellyfin/jellyfin-web/issues/913.

**How to Test**

1. Add a media with name `返`.
2. Change the language setting in http://localhost:8080/#/mypreferencesdisplay.html to `日本語` (Japanese).
3. Navigate to the media added at the step 1.
4. Compare the displayed name with an image here: https://en.wikipedia.org/wiki/Source_Han_Sans#/media/File:Source_Han_Sans_Version_Difference.svg.
5. Repeat from the step 2 for each languages `한국어` (Korean), `汉语 (简化字)` (Simplified Chinese), `漢語 (繁体字)` (Traditional Chinese - Taiwan), and `廣東話 (香港)` (Traditional Chinese - Hong Kong).

### DIFF
--- a/src/scripts/globalize.js
+++ b/src/scripts/globalize.js
@@ -1,7 +1,7 @@
 import { Events } from 'jellyfin-apiclient';
 import isEmpty from 'lodash-es/isEmpty';
 
-import * as userSettings from './settings/userSettings';
+import { currentSettings as userSettings } from './settings/userSettings';
 
 /* eslint-disable indent */
 
--- a/src/scripts/site.js
+++ b/src/scripts/site.js
@@ -274,11 +274,24 @@ async function onAppReady() {
             }
         };
 
-        Events.on(ServerConnections, 'localusersignedin', handleStyleChange);
-        Events.on(ServerConnections, 'localusersignedout', handleStyleChange);
+        const handleLanguageChange = () => {
+            const locale = globalize.getCurrentLocale();
+
+            document.documentElement.setAttribute('lang', locale);
+        };
+
+        const handleUserChange = () => {
+            handleStyleChange();
+            handleLanguageChange();
+        };
+
+        Events.on(ServerConnections, 'localusersignedin', handleUserChange);
+        Events.on(ServerConnections, 'localusersignedout', handleUserChange);
         Events.on(currentSettings, 'change', (e, prop) => {
             if (prop == 'disableCustomCss' || prop == 'customCss') {
                 handleStyleChange();
+            } else if (prop == 'language') {
+                handleLanguageChange();
             }
         });
 
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -51,6 +51,7 @@
  - [GodTamIt](https://github.com/GodTamIt)
  - [MinecraftPlaye](https://github.com/MinecraftPlaye)
  - [Matthew Jones](https://github.com/matthew-jones-uk)
+ - [taku0](https://github.com/taku0)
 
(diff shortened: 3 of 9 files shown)


src/scripts/globalize.js
```diff
@@ -243,7 +243,7 @@ import * as userSettings from './settings/userSettings';
 
     updateCurrentCulture();
 
-    Events.on(userSettings, 'change', function (e, name) {
+    Events.on(userSettings.currentSettings, 'change', function (e, name) {
```
dmitrylyzo, line 246: Oops, good catch.
You can import it as `import { currentSettings as userSettings } from ...` instead of this.
That bunch of exports is for backward compatibility when we were migrating to ES6, iirc.

src/scripts/site.js
```diff
@@ -246,11 +246,24 @@ async function onAppReady() {
             }
         };
 
-        Events.on(ServerConnections, 'localusersignedin', handleStyleChange);
-        Events.on(ServerConnections, 'localusersignedout', handleStyleChange);
+        const handleLanguageChange = async () => {
```
dmitrylyzo, line 249: No need to be `async`.
```suggestion
        const handleLanguageChange = () => {
```

src/scripts/site.js
```diff
@@ -246,11 +246,24 @@ async function onAppReady() {
...
+            const locale = globalize.getCurrentLocale();
+
+            document.documentElement.setAttribute("lang", locale);
+        };
+
+        const handleUserChange = async () => {
```
dmitrylyzo, line 255: No need to be `async`.
```suggestion
        const handleUserChange = () => {
```

taku0:
Addressed the review comments.

nyanmisaka:
This should also be back ported to 10.8 branch if possible.

src/scripts/site.js
```diff
@@ -246,11 +246,24 @@ async function onAppReady() {
             }
         };
 
-        Events.on(ServerConnections, 'localusersignedin', handleStyleChange);
-        Events.on(ServerConnections, 'localusersignedout', handleStyleChange);
+        const handleLanguageChange = () => {
+            const locale = globalize.getCurrentLocale();
+
+            document.documentElement.setAttribute("lang", locale);
```
dmitrylyzo, line 252:
```suggestion
            document.documentElement.setAttribute('lang', locale);
```

taku0: Fixed.

taku0:
Addressed stylelint errors.

joshuaboniface:
>This should also be back ported to 10.8 branch if possible.

Agreed - @taku0 can you rebase your commit off the `release-10.8.z` branch instead of `master` then force-push it here? The auto-rebase in GitHub didn't work.

joshuaboniface:
Disregard what I suggested, just do this on your feature branch then force push: `git rebase -i upstream/release-10.8.z` lol

ferferga (approved):

joshuaboniface:
@taku0 Get out of my brain I was *just* about to try my own hand at it :rofl:

taku0:
Rebased and changed the target of the PR.

joshuaboniface (approved):

PR #3651, opened 2022-05-17
Remove unused Screenshot code

grafixeyehero:


nielsvanvelzen (approved):

thornbill (changes requested):
You should only make edits to the en-us language file and let weblate deal with the rest to avoid conflicts on the weblate side.

PR #3655, opened 2022-05-18
automatically switch to LocalAddress when accessible

kookxiang:
**Changes**

a server has two addresses: ManualAddress (via internet) + LocalAddress (local only).
when a server LocalAddress is accessible, it should not use ManualAddress.
it seems like this feature was disabled somehow, this PR just enable it

I'd test it on my server and it works well.

image
image

**Issues**

### DIFF
--- a/src/components/ServerConnections.js
+++ b/src/components/ServerConnections.js
@@ -58,7 +58,7 @@ class ServerConnections extends ConnectionManager {
         );
 
         apiClient.enableAutomaticNetworking = false;
-        apiClient.manualAddressOnly = true;
+        apiClient.manualAddressOnly = false;
 
         this.addApiClient(apiClient);
 


dmitrylyzo:
That's old: https://github.com/jellyfin/jellyfin-web/blame/f5c2a331b0f69c92202aa6c622e2885ed3e6e7ba/src/scripts/site.js#L199

I guess it currently works as you described, only in the standalone mode (or bundled apps): https://github.com/jellyfin/jellyfin-tizen/issues/113#issuecomment-1079773421

After the change, UI will still be loaded from the remote (manual) address, right?

thornbill:
I haven't looked at how this is implemented in the apiclient yet, but this may cause issues with docker installs where the local address can be the IP address of the container on the docker network.

dmitrylyzo:
In theory, it will try 3 addresses: https://github.com/jellyfin/jellyfin-apiclient-javascript/blob/e8bd1b0e137068b3874ef35720453787e41aba60/src/connectionManager.js#L618-L652
The bundled app does this all the time, iirc

thornbill (approved):

PR #3662, opened 2022-05-21
Support searching for tv programs

1337joe:
Previously, if a show was tagged as "IsKids" and "IsSeries" it would show up in neither of those categories on search because the search excluded results tagged as anything but the target category. This change removes the requirement that a show be only one type to show up in search.

**Changes**
- Remove false flags from category filters to not exclude things that fall into multiple sections

**Issues**
Related to jellyfin/jellyfin#7204
Requires jellyfin/jellyfin#7780

### DIFF
--- a/src/components/search/LiveTVSearchResults.tsx
+++ b/src/components/search/LiveTVSearchResults.tsx
@@ -83,7 +83,11 @@ const LiveTVSearchResults: FunctionComponent<LiveTVSearchResultsProps> = ({ serv
             // Episodes row
             fetchItems(apiClient, {
                 IncludeItemTypes: 'LiveTvProgram',
-                IsSeries: true
+                IsMovie: false,
+                IsSeries: true,
+                IsSports: false,
+                IsKids: false,
+                IsNews: false
             }).then(result => setEpisodes(result.Items || []));
             // Sports row
             fetchItems(apiClient, {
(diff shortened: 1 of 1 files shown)


src/components/search/LiveTVSearchResults.tsx
```diff
@@ -79,57 +78,31 @@ const LiveTVSearchResults: FunctionComponent<LiveTVSearchResultsProps> = ({ serv
...
             // Programs row
-            fetchItems(apiClient, {
-                IncludeItemTypes: 'LiveTvProgram',
-                IsMovie: false,
-                IsSeries: false,
-                IsSports: false,
-                IsKids: false,
-                IsNews: false
-            }).then(result => setPrograms(result.Items || []));
+            fetchItems(apiClient, { IncludeItemTypes: 'LiveTvProgram' })
```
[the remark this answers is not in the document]
1337joe: Good point. The data set I'm testing with doesn't seem to have uncategorized items so this leaves it blank, but everything in programs being a duplicate of another category doesn't really add anything anyway.

src/components/search/LiveTVSearchResults.tsx
```diff
@@ -79,46 +78,26 @@ const LiveTVSearchResults: FunctionComponent<LiveTVSearchResultsProps> = ({ serv
...
             }).then(result => setMovies(result.Items || []));
             // Episodes row
             fetchItems(apiClient, {
                 IncludeItemTypes: 'LiveTvProgram',
-                IsMovie: false,
-                IsSeries: true,
-                IsSports: false,
-                IsKids: false,
-                IsNews: false
+                IsSeries: true
```
thornbill, line 86: After testing this, it may make sense to keep the exclusions for this one to avoid duplication in this row also... :man_shrugging:

1337joe: That's only the case until jellyfin/Jellyfin.XmlTv#12 hits in the next release of the XmlTv library. There is currently a bug that causes literally everything to identify as a series.

thornbill (approved):

src/components/search/LiveTVSearchResults.tsx
thornbill: Is XmlTv used for SchedulesDirect EPG data also?

1337joe: Glancing over the code it looks like it does not. I haven't used SchedulesDirect so I can't speak to how it's results will behave for search.

thornbill: I do use SD for guide data. That is where I was seeing the duplicates in the Episodes row, so it may need a similar fix. :slightly_smiling_face:

1337joe: Ok, that makes sense. Pushed that fix.

PR #3666, opened 2022-05-21
Fix import

grafixeyehero:


src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -39,7 +39,7 @@ import { setBackdropTransparency, TRANSPARENCY_LEVEL } from '../../components/ba
  */
 function resolveUrl(url) {
     return new Promise((resolve) => {
-        var xhr = new XMLHttpRequest();
+        const xhr = new XMLHttpRequest();
```
dmitrylyzo, line 42: This comes from the release branch (we have no Lint job there).

src/scripts/libraryMenu.js
```diff
@@ -1,6 +1,5 @@
 import escapeHtml from 'escape-html';
 import Headroom from 'headroom.js';
-
```
dmitrylyzo, line 3: I think Bill wanted them to be separated as: global deps, local deps, deps without `from` and styles.

grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/3666/commits/3bc5b58dd83e28755d79c13dfb56719ebd0a7356

dmitrylyzo (approved):

crobibero (approved):

PR #3668, opened 2022-05-22
Show PlaySettings button even if transcoding isn't supported

dmitrylyzo:
**Changes**
Show PlaySettings button even if transcoding isn't supported

**Issues**
There is no way to show play stats, set subtitles offset and aspect ratio if the video is DirectPlay (`!SupportsTranscoding`).


src/controllers/playback/video/index.js
```diff
@@ -868,6 +864,7 @@ import { appRouter } from '../../../components/appRouter';
                         mediaType: 'Video',
                         player: player,
                         positionTo: btn,
+                        quality: state.MediaSource?.SupportsTranscoding,
```
dmitrylyzo, line 867: Can it be that transcoding is allowed (for the user) but not supported?

dmitrylyzo: it probably can - no transcoding profiles

PR #3670, opened 2022-05-23
Fix unexpected remuxing of direct playable media

dmitrylyzo:
**Changes**
Force bitrate auto detection if transcoding is not allowed for the user.

**Issues**
If the user has set a maximum bitrate and transcoding is blocked after that, direct playable video is never (_if its bitrate exeeds the specified one_) played direct because of this limit.
Moreover, the user can't change the bitrate because the UI is hidden.

**Steps To Reproduce**
1. With allowed transcoding, set the maximum bitrate in `Settings/Playback` to have the direct playable video transcoded.
2. Block transcoding.
3. Refresh page / relogin.
4. Start video.
5. Remuxing is used instead of DirectPlay. (_No DirectPlay because bitrate is exceeded_)
And there is no way to change this.

### DIFF
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -1409,13 +1409,22 @@ class PlaybackManager {
 
             const apiClient = ServerConnections.getApiClient(self.currentItem(player).ServerId);
 
-            apiClient.getEndpointInfo().then(function (endpointInfo) {
+            Promise.all([
+                apiClient.getEndpointInfo(),
+                apiClient.getCurrentUser()
+            ]).then(([endpointInfo, user]) => {
                 const playerData = getPlayerData(player);
                 const mediaType = playerData.streamInfo ? playerData.streamInfo.mediaType : null;
 
                 let promise;
                 if (options.enableAutomaticBitrateDetection) {
                     appSettings.enableAutomaticBitrateDetection(endpointInfo.IsInNetwork, mediaType, true);
+                    promise = apiClient.detectBitrate(true);
+                } else if (user?.Policy
+                    && ((mediaType === 'Video' && !user.Policy.EnableVideoPlaybackTranscoding)
+                        || (mediaType === 'Audio' && !user.Policy.EnableAudioPlaybackTranscoding))) {
+                    // If transcoding is not allowed, we cannot use the specified bitrate - force bitrate detection
+                    console.warn('Transcoding is not allowed, cannot use the specified bitrate - force bitrate detection');
                     promise = apiClient.detectBitrate(true);
                 } else {
                     appSettings.enableAutomaticBitrateDetection(endpointInfo.IsInNetwork, mediaType, false);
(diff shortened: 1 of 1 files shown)


src/components/playback/playbackmanager.js
```diff
@@ -1409,14 +1409,23 @@ class PlaybackManager {
...
                 let promise;
                 if (options.enableAutomaticBitrateDetection) {
                     appSettings.enableAutomaticBitrateDetection(endpointInfo.IsInNetwork, mediaType, true);
                     promise = apiClient.detectBitrate(true);
+                } else if (user?.Policy
+                    && ((mediaType === 'Video' && !user.Policy.EnableVideoPlaybackTranscoding)
+                        || (mediaType === 'Audio' && !user.Policy.EnableAudioPlaybackTranscoding))) {
+                    // If transcoding is not allowed, we cannot use the specified bitrate - force bitrate detection
+                    console.warn('Transcoding is not allowed, cannot use the specified bitrate - force bitrate detection');
+                    promise = apiClient.detectBitrate(true);
```
dmitrylyzo, line 1428: We could add `appSettings.enableAutomaticBitrateDetection(endpointInfo.IsInNetwork, mediaType, true);` to stop warning next play.

src/components/playback/playbackmanager.js
```diff
@@ -2143,8 +2152,26 @@ class PlaybackManager {
...
+
+                    if (mediaType === 'Video' || mediaType === 'Audio') {
+                        enableAutomaticBitrateDetection = appSettings.enableAutomaticBitrateDetection(endpointInfo.IsInNetwork, mediaType);
+
+                        if (!enableAutomaticBitrateDetection && user?.Policy
+                            && ((mediaType === 'Video' && !user.Policy.EnableVideoPlaybackTranscoding)
+                                || (mediaType === 'Audio' && !user.Policy.EnableAudioPlaybackTranscoding))) {
+                            // If transcoding is not allowed, we cannot use the specified bitrate - force bitrate detection
+                            console.warn('Transcoding is not allowed, cannot use the specified bitrate - force bitrate detection');
+                            enableAutomaticBitrateDetection = true;
```
dmitrylyzo, line 2170: We could add `appSettings.enableAutomaticBitrateDetection(endpointInfo.IsInNetwork, mediaType, true);` to stop warning next play.

src/components/playback/playbackmanager.js
thornbill, line 1428: How does detecting the bitrate account for not being able to transcode? Does it just return `null`/`undefined` to indicate the source bitrate should be used?

dmitrylyzo: > How does detecting the bitrate account for not being able to transcode? Does it just return `null`/`undefined` to indicate the source bitrate should be used?

It just should be as in `Auto` mode - return detected bitrate or that 140Mbps (in the apiclient), i.e. greater than the original bitrate.
But if it is lower than the original bitrate it will probably fall back to remuxing. :thinking: 
I dunno if we can pass `undefined` as the maximum bitrate to the server. As an alternative, bitrate should be ignored on the server side.

dmitrylyzo:
I tested passing `null` bitrate to `playAfterBitrateDetect` and it seems to work as expected - DirectPlay.

But now I'm leaning towards the idea of ignoring `MaxBitrate` on the server: if the user doesn't have permission to transcode, the server cannot make the bitrate lower.
There is another option: if the bitrate of the media exceeds the maximum network bitrate (detection required), the media is treated as unsupported.

dmitrylyzo:
Server-side alternative: https://github.com/jellyfin/jellyfin/pull/9082

Shadowghost:
I'd prefer doing this for all clients server side.
Nonetheless the drop-down menus should be hidden in web if the user doesn't have transcoding permissions.

thornbill:

PR #3675, opened 2022-05-29
Serve script cast_sender.js locally instead of serving it from google cdn

firstdorsal:
**Changes**

Serve script cast_sender.js locally instead of serving it from google cdn

I hope putting it in the scripts folder is correct.


nielsvanvelzen:
Why? The script loads additional resources from the Google CDN anyway so this change doesn't do much except for adding some obfuscated javascript blob to our sources.

firstdorsal:
This script gets requested on every jellyfin page load from gstatic but its action of loading another script does not get performed until some action (loading something chromecast related i asume) happens. So why allow google to serve and track a resource that is not required by all users?

thornbill:
This is not the best idea imo. We have no way of tracking upstream changes made to the upstream script, so at some point it will break and no one will know why. If you are using Chromecast you should know that you are dependent on Google services.

If you want to use Jellyfin without Chromecast, you can use a browser like Firefox where Chromecast is not supported or you can remove the Chromecast plugin from your local copy of the [web config file](https://github.com/jellyfin/jellyfin-web/blob/master/src/config.json#L47) for your server.

PR #3706, opened 2022-06-14
Add react router and migrate search page

thornbill:
**Changes**
Adds react router, supporting components, and migrates the search page to use react router

This replaces the need for the `pageComponent` support in `appRouter`. All pages in [src/components/pages](https://github.com/jellyfin/jellyfin-web/tree/master/src/components/pages) should be migrated in future PRs.

**Issues**
N/A


src/scripts/routes.js
```diff
@@ -308,12 +308,6 @@ import { appRouter } from '../components/appRouter';
         type: 'home'
```
thornbill, line 308: This file should probably be moved to `routes/` and renamed. Not sure if that would best fit in this PR or a follow up.

src/components/appRouter.js
```diff
@@ -122,7 +122,11 @@ class AppRouter {
                 isBack: action === Action.Pop
```
thornbill, line 122: This file might make more sense in `scripts/`. :man_shrugging:

package.json
```diff
@@ -96,6 +96,7 @@
     "pdfjs-dist": "2.12.313",
     "react": "17.0.2",
     "react-dom": "17.0.2",
+    "react-router-dom": "6.3.0",
```
dmitrylyzo, line 99: Tested in webOS 1.2 emulator: `const` in `main.bundle` from `react-router-dom`.
We need to babel it for legacy browsers.

src/components/HistoryRouter.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * A slightly customized version of the HistoryRouter from react-router-dom.
+ * We need to use HistoryRouter to have a shared history state between react-router and appRouter, but it does not seem
+ * to be properly exported in the upstream package.
+ * We also needed some customizations to handle #! routes.
+ * Refs: https://github.com/remix-run/react-router/blob/v6.3.0/packages/react-router-dom/index.tsx#L222
+ */
+export function HistoryRouter({ basename, children, history }: HistoryRouterProps) {
+    const [state, setState] = React.useState({
```
dmitrylyzo, line 16:
```suggestion
    const [state, setState] = React.useState<Update>({
```
_Lint should warn on type, imo, but it doesn't_ :man_shrugging:

dmitrylyzo (commented):
I'm new to React :sweat_smile: but at least it works in webOS 1.2

src/components/ConnectionRequired.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+            // TODO: should exiting the app be handled here?
+
+            const client = ServerConnections.currentApiClient();
+
+            // If this is a user route, ensure a user is logged in
+            if ((isAdminRequired || isUserRequired) && !client?.isLoggedIn()) {
+                try {
+                    console.warn('[ConnectionRequired] unauthenticated user attempted to access user route');
+                    bounce(await ServerConnections.connect());
```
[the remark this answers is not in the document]
thornbill: I imagine the backdrop will eventually be its own component that uses either the available hooks or [<Routes>](https://reactrouter.com/docs/en/v6/getting-started/overview#multiple-sets-of-routes) to control visibility.

src/components/ConnectionRequired.tsx
```diff
@@ -0,0 +1,167 @@
...
+                            navigate(BounceRoutes.StartWizard);
+                            return;
+                        }
+                    } catch (ex) {
+                        console.error('[ConnectionRequired] checking wizard status failed', ex);
+                    }
+                }
+
+                // Bounce to the correct page in the login flow
+                bounce(firstConnection);
```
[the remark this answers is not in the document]
thornbill: Good catch I missed a `return` when I was porting the logic from appRouter.

src/components/ConnectionRequired.tsx
```diff
@@ -0,0 +1,167 @@
...
+                return;
+            }
+
+            // TODO: should exiting the app be handled here?
```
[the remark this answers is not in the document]
thornbill: I updated the comment here. I mainly just wanted to point out a gap between what is currently covered by appRouter and what is handled in this component.

src/components/Page.tsx
```diff
@@ -0,0 +1,56 @@
...
+        const event = {
+            bubbles: true,
+            cancelable: false,
+            detail: {
+                isRestored: false
+            }
+        };
+        // pagebeforeshow - hides tabs on tabless pages in libraryMenu
+        element.current?.dispatchEvent(new CustomEvent('pagebeforeshow', event));
+        // viewshow - updates state of appRouter
```
grafixeyehero, line 37: `viewbeforeshow` need here for theme

src/components/Page.tsx
```diff
@@ -0,0 +1,56 @@
...
+        element.current?.dispatchEvent(new CustomEvent('viewshow', event));
+        // pageshow - updates header/navigation in libraryMenu
+        element.current?.dispatchEvent(new CustomEvent('pageshow', event));
+    }, [ element ]);
+
+    return (
+        <div
+            ref={element}
```
grafixeyehero, line 45: libraryMenu need `id` to update the selected current view Id

src/components/Page.tsx
thornbill: Good catch! I added the event as well as some additional values in the event details that were required for theme songs/videos and the now playing bar to work correctly.

src/components/Page.tsx
thornbill: `id` is now a required prop for the Page component.

PR #3711, opened 2022-06-15
Punctuation bug fix in right to left languages.

1netanel:
Punctuation is reversed when using subtitles with RTL (right to left) languages.

**Changes**
Forcing RTL (right to left) text direction for RTL languages.


thornbill:
This seems like it should be handled by the server.

1netanel:
> This seems like it should be handled by the server.

The problem is in the <track> element which supports only left to right. The only way to enable RTL is by forcing it from the .vtt file.

PR #3712, opened 2022-06-16
Set permissions for GitHub actions

nathannaveen:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: nathannaveen <42319948+nathannaveen@users.noreply.github.com>

### DIFF
--- a/.github/workflows/commands.yml
+++ b/.github/workflows/commands.yml
@@ -5,8 +5,13 @@ on:
       - created
       - edited
 
+permissions:
+  contents: read
+
 jobs:
   rebase:
+    permissions:
+      contents: none
     name: Rebase
     if: github.event.issue.pull_request != '' && contains(github.event.comment.body, '@jellyfin-bot rebase') && github.event.comment.author_association == 'MEMBER'
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/commands.yml
```diff
@@ -5,8 +5,13 @@ on:
       - created
       - edited
 
+permissions:
+  contents: read
+
 jobs:
   rebase:
+    permissions:
+      contents: none
```
h1dden-da3m0n, line 14: same question as before? (might be me being new to the permissions options)

h1dden-da3m0n (commented):
Other than to satisfy my curiosity and possible using only one setting instead of two. LGTM

nathannaveen:
I don't have cycles at the moment for addressing these comments, would you be able to cherry pick them? Thanks!

PR #3716, opened 2022-06-16
Recognize MIME for transcoded audio in streams

Orhideous:
After jellyfin/jellyfin#6941 MIME library returns IETF-approved type (`application/vnd.apple.mpegurl`) instead of old one (`application/x-mpegURL`), but latter was hardcoded in player plugin.

**Changes**
Use correct MIME-type for transcoded streams.

**Issues**
Fixes: #3663

### DIFF
--- a/src/plugins/htmlAudioPlayer/plugin.js
+++ b/src/plugins/htmlAudioPlayer/plugin.js
@@ -74,7 +74,7 @@ function enableHlsPlayer(url, item, mediaSource, mediaType) {
                 type: 'HEAD'
             }).then(function (response) {
                 const contentType = (response.headers.get('Content-Type') || '').toLowerCase();
-                if (contentType === 'application/x-mpegurl') {
+                if (contentType === 'application/vnd.apple.mpegurl') {
                     resolve();
                 } else {
                     reject();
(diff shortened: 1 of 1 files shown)


crobibero:
I'm no authority on the web client but I would probably allow either to be specified

thornbill:
I agree with @crobibero's comment above. Let's keep both for now just in case someone is connecting to a 10.7 server or some other unusual case is hit here.

src/plugins/htmlAudioPlayer/plugin.js
```diff
@@ -74,7 +74,7 @@ function enableHlsPlayer(url, item, mediaSource, mediaType) {
                 type: 'HEAD'
             }).then(function (response) {
                 const contentType = (response.headers.get('Content-Type') || '').toLowerCase();
-                if (contentType === 'application/x-mpegurl') {
+                if (contentType === 'application/vnd.apple.mpegurl') {
```
nyanmisaka, line 77:
```suggestion
                if (contentType === 'application/vnd.apple.mpegurl' || contentType === 'application/x-mpegurl') {
```

nyanmisaka (changes requested):

nyanmisaka (approved):

crobibero (approved):

thornbill (approved):

Issue #3708, opened 2022-06-14
[Issue]: Anything above 1080p - 60mbit quality options are missing from playback since 10.8.


sarunas-zilinskas:
### Please describe your bug

Ever since 10.7.7 update to 10.8 I noticed that the performance of transcoding has increased significantly, which is very welcome and you did great job guys!

On the sad note, I have noticed a bug - when trying to play 4K file, 4K playback option is missing on the client's side. The highest option available is 1080p 60mbit/s. 4K 80/100/120 mbit options are missing. They are missing in the quality section of the playback and in the profile settings.

After the update I noticed that I had another instance open of the browser from the 10.7.7 version and with the same session I was able to playback the file in 4K, changing quality settings etc. so yes, the file(s, plural, because I tried couple of them which were watched before in 4K on 10.7.7) are definitely 4K definiton and with the older session 4K option was present. Then I wanted to test out if that persists by closing the browser and opening up again. This lead to 4K option missing. In playback info it shows` Video resolution: 1920x800` , however, even in the media info it shows that it is 4K.

Tried it in chrome and firefox. Same behavior on both browsers.

Seems like the issue is as if server assumes client can't handle anything above 4K and does not give an option to select anything higher than 1080p.

### Jellyfin Version

10.8.0

### Environment

```markdown
- OS: Debian 11
- Virtualization: Docker + jellyfin image
- Clients: Chrome 102 and firefox 101.0.1 (Both tested on windows 10)
- Browser: Chrome 102 and firefox 101.0.1 (Both tested on windows 10)
- FFmpeg Version: The one included in jellyfin 10.8 image
- Playback Method: Transcoding (From HEVC to H264)
- Hardware Acceleration: VAAPI
- Plugins: -
- Reverse Proxy: Caddy
- Base URL: none
- Networking:
- Storage: Local
```

### Jellyfin logs

```shell
[2022-06-14 21:06:13.151 +00:00] [INF] [12] Jellyfin.Api.Helpers.MediaInfoHelper: User policy for "sa". EnablePlaybackRemuxing: True EnableVideoPlaybackTranscoding: True EnableAudioPlaybackTranscoding: True
[2022-06-14 21:06:13.152 +00:00] [INF] [12] Jellyfin.Api.Helpers.MediaInfoHelper: RemoteClientBitrateLimit: 150000000, RemoteIp: "10.0.2.254", IsInLocalNetwork: True
[...]

PR #3717, opened 2022-06-16
Add maximum allowed video resolution selector

dmitrylyzo:
In some cases it is hard to get the screen resolution right. Let the user decide.

**Changes**
Add maximum allowed video resolution selector

**Issues**
#3708

**Screenshots**
resolution


src/components/playbackSettings/playbackSettings.js
```diff
@@ -247,6 +257,33 @@ import template from './playbackSettings.template.html';
...
+        fillChromecastQuality(selectChromecastVideoQuality, maxVideoWidth);
+
+        selectVideoInNetworkQuality.value = selectVideoInNetworkQualityValue;
+        selectVideoInternetQuality.value = selectVideoInternetQualityValue;
+        selectChromecastVideoQuality.value = selectChromecastVideoQualityValue;
+
+        // HACK: Fix wrong value if it out of list
+        selectVideoInNetworkQuality.value = selectVideoInNetworkQuality.value;
+        selectVideoInternetQuality.value = selectVideoInternetQuality.value;
+        selectChromecastVideoQuality.value = selectChromecastVideoQuality.value;
```
dmitrylyzo, line 284: SonarCloud doesn't like this. Any ideas?

thornbill: I'm not sure I understand what this is supposed to be doing :sweat_smile:

dmitrylyzo: If the old value is not found in the new list, select becomes empty (its value is undefined). This block sets the existing or undefined value back, 
to fill select.

thornbill: Just so I am clear, setting the value the first time does not accomplish this then?

src/strings/en-us.json
```diff
@@ -708,6 +708,7 @@
     "LabelLibraryPageSizeHelp": "Set the amount of items to show on a library page. Set to 0 in order to disable paging.",
     "LabelMaxDaysForNextUp": "Max days in 'Next Up':",
     "LabelMaxDaysForNextUpHelp": "Set the maximum amount of days a show should stay in the 'Next Up' list without watching it.",
+    "LabelMaxVideoResolution": "Maximum Allowed Video Resolution",
```
thornbill, line 711: Maybe we could call this "Maximum Allowed Video Transcoding Resolution" and remove the help text?

Shadowghost:
Might be due to source video bitrate being low enough to suffice the 720p option

thornbill:
Ah that could be it... looks like the source is only around 2 mbps

> Bitrate 2053 kbps

src/components/playbackSettings/playbackSettings.js
dmitrylyzo: Yes, at least not the way I want. If I set a value that is not in the list, select does not choose default (with undefined `value` attribute - `Auto`), it just becomes empty.

dmitrylyzo:
> I noticed something seems a bit off with the setting while testing...
> 
> 1080p -> Works as expected 720p -> Works as expected 480p -> Shows 720p quality options 360p -> Shows 480p quality options

We just also show next resolution, because that resolutions are just for reference.
https://github.com/jellyfin/jellyfin-web/blob/c8590d37ed4e1a16857940420798bbd55be72a9a/src/components/qualityOptions.js#L43

src/components/playbackSettings/playbackSettings.js
dmitrylyzo: Now also Lint is not satisfied :man_facepalming:

thornbill: What do you think about doing something like this to avoid the hack?

```js
document.querySelector('.selectLabelMaxVideoWidth option[value="0"]')?.selected = true; 
```

dmitrylyzo: I only need to do this if the old value is not in the list. I found solution.

thornbill (approved):
I haven't tested the latest changes, so it might be a good idea for someone to do so, but the code looks fine.

Shadowghost:
I tested the current commits and it worked as intended from how I understood it should work.

joshuaboniface (approved):

PR #3718, opened 2022-06-16
Turn backdrops on by default

unknown-current:
All new users have backdrops disabled by default. Previously, this could be changed by editing the main.*.bundle.js file, but since 10.8.0 that no longer applies.

**Changes**
Change the default setting for backdrops to true.

### DIFF
--- a/src/scripts/settings/userSettings.js
+++ b/src/scripts/settings/userSettings.js
@@ -239,7 +239,7 @@ export class UserSettings {
             return this.set('enableBackdrops', val.toString(), false);
         }
 
-        return toBoolean(this.get('enableBackdrops', false), false);
+        return toBoolean(this.get('enableBackdrops', false), true);
     }
 
     /**
--- a/src/scripts/settings/userSettings.js
+++ b/src/scripts/settings/userSettings.js
@@ -1,6 +1,7 @@
 import appSettings from './appSettings';
 import { Events } from 'jellyfin-apiclient';
 import { toBoolean } from '../../utils/string.ts';
+import browser from '../browser';
 
 function onSaveTimeout() {
     const self = this;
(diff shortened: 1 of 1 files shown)


src/scripts/settings/userSettings.js
```diff
@@ -239,7 +239,7 @@ export class UserSettings {
             return this.set('enableBackdrops', val.toString(), false);
         }
 
-        return toBoolean(this.get('enableBackdrops', false), false);
+        return toBoolean(this.get('enableBackdrops', false), true);
```
dmitrylyzo, line 242: Navigation on the TV is not smooth with backdrops due to the poor performance of the TV hardware. So I would disable it for "slow" browsers.

Import `browser` module and
```suggestion
        return toBoolean(this.get('enableBackdrops', false), !browser.slow);
```

thornbill:
I tagged this as needs discussion since it's a matter of opinion as to whether this should be enabled by default or not...

Personally I'm not a fan of the backdrops in web.

unknown-current:
I think ideally there would be an option to change default user settings per server but that's a bigger undertaking and I know there's a feature request for it already. Personally I like backdrops especially because I use the Ultrachromic theme which has blurred backdrops that I think look quite nice, but I would agree it's a matter of opinion. Before the latest update there was a hacky way to do it with the main.xxxx.bundle.js but that no longer works, which is the main reason I proposed this. Maybe there's a way to change it now without rebuilding but this is my first time really digging into the code so I'm not sure.

When it comes to this PR, I agree further discussion and input from more people is definitely necessary since it's a design choice.

thornbill:
It turns out we explicitly disabled these by default in the 10.6.0 release (https://github.com/jellyfin/jellyfin-web/pull/1288). Much to my surprise, I was even opposed to the change at the time, but the cleaner aesthetic has clearly grown on me since then. :sweat_smile:

I definitely agree with the long term solution being allowing admins to specify a default set of user preferences per server, but in the meantime I don't _think_ we want to make these on by default again.

PR #3719, opened 2022-06-18
Expose DoVi metadata and VideoRangeType

nyanmisaka:
**Server**
https://github.com/jellyfin/jellyfin/pull/7964

**Changes**
- Expose DoVi metadata and VideoRangeType

Please add this to **"stable backport" and "backend"**, thanks!

### DIFF
--- a/src/components/itemMediaInfo/itemMediaInfo.js
+++ b/src/components/itemMediaInfo/itemMediaInfo.js
@@ -113,7 +113,7 @@ const attributeDelimiterHtml = layoutManager.tv ? '' : '<span class="hide">: </s
             if (stream.Profile) {
                 attributes.push(createAttribute(globalize.translate('MediaInfoProfile'), stream.Profile));
             }
-            if (stream.Level && stream.Type === 'Video') {
+            if (stream.Level > 0) {
                 attributes.push(createAttribute(globalize.translate('MediaInfoLevel'), stream.Level));
             }
             if (stream.Width || stream.Height) {
--- a/src/components/itemMediaInfo/itemMediaInfo.js
+++ b/src/components/itemMediaInfo/itemMediaInfo.js
@@ -113,7 +113,7 @@ const attributeDelimiterHtml = layoutManager.tv ? '' : '<span class="hide">: </s
             if (stream.Profile) {
                 attributes.push(createAttribute(globalize.translate('MediaInfoProfile'), stream.Profile));
             }
-            if (stream.Level) {
+            if (stream.Level && stream.Type === 'Video') {
                 attributes.push(createAttribute(globalize.translate('MediaInfoLevel'), stream.Level));
             }
             if (stream.Width || stream.Height) {
@@ -128,7 +128,7 @@ const attributeDelimiterHtml = layoutManager.tv ? '' : '<span class="hide">: </s
                 }
                 attributes.push(createAttribute(globalize.translate('MediaInfoInterlaced'), (stream.IsInterlaced ? 'Yes' : 'No')));
             }
-            if (stream.AverageFrameRate || stream.RealFrameRate) {
+            if ((stream.AverageFrameRate || stream.RealFrameRate) && stream.Type === 'Video') {
                 attributes.push(createAttribute(globalize.translate('MediaInfoFramerate'), (stream.AverageFrameRate || stream.RealFrameRate)));
             }
             if (stream.ChannelLayout) {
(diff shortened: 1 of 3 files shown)


Shadowghost (approved):

src/components/itemMediaInfo/itemMediaInfo.js
```diff
@@ -113,7 +113,7 @@ const attributeDelimiterHtml = layoutManager.tv ? '' : '<span class="hide">: </s
             if (stream.Profile) {
                 attributes.push(createAttribute(globalize.translate('MediaInfoProfile'), stream.Profile));
             }
-            if (stream.Level) {
+            if (stream.Level && stream.Type === 'Video') {
```
dmitrylyzo, line 116: Do we have to check `Type`?
Most attributes don't have such a check.

src/components/playerstats/playerstats.js
```diff
@@ -269,31 +269,10 @@ import ServerConnections from '../ServerConnections';
...
-        }
-
-        if (videoStream.ColorTransfer) {
-            sessionStats.push({
-                label: globalize.translate('LabelColorTransfer'),
-                value: videoStream.ColorTransfer
-            });
-        }
-
-        if (videoStream.ColorPrimaries) {
```
dmitrylyzo, line 293: Removed as unnecessary (for the user)?

src/components/itemMediaInfo/itemMediaInfo.js
nyanmisaka: Yes, only for certain attributes. Such as level should only be shown for video type but not for music and image.

src/components/playerstats/playerstats.js
nyanmisaka: The stats list is too long and it's better to replace these technical infos with "HDR10/HLG/DOVI".

src/components/itemMediaInfo/itemMediaInfo.js
dmitrylyzo: Do we have `Level`, `AverageFrameRate` and `RealFrameRate` for non-video and `BitRate` for non-video/audio?

nyanmisaka: Nope. The first three are video only and bitrate is audio/video only.

dmitrylyzo: Actually, I got `Level` for audio, but it is `0`.
I mean, maybe remove these checks?

nyanmisaka: AKAIK 0 and -99 are usually invalid levels, compared with 51(5.1 h264) or 180(6.0 hevc). Did we ever use level attribute for audio?
https://github.com/jellyfin/jellyfin/blob/50bc41d84d7f3f7ecdccbb6ed1885426f87c0d9a/tests/Jellyfin.MediaEncoding.Tests/Test%20Data/Probing/music_metadata.json

dmitrylyzo: I have video VP9 with `Level: -99`.
Subtitles and audio have `Level: 0`.
I think we can just leave `if (Level)` or check for validity `if (Level > 0)`.

nyanmisaka: Done.

src/components/itemMediaInfo/itemMediaInfo.js
```diff
@@ -128,7 +128,7 @@ const attributeDelimiterHtml = layoutManager.tv ? '' : '<span class="hide">: </s
                 }
                 attributes.push(createAttribute(globalize.translate('MediaInfoInterlaced'), (stream.IsInterlaced ? 'Yes' : 'No')));
             }
-            if (stream.AverageFrameRate || stream.RealFrameRate) {
+            if ((stream.AverageFrameRate || stream.RealFrameRate) && stream.Type === 'Video') {
```
dmitrylyzo, line 131: This check seems necessary - the embedded cover image has `Level: -99` and `Framerate: 90000`.

src/components/itemMediaInfo/itemMediaInfo.js
```diff
@@ -137,7 +137,7 @@ const attributeDelimiterHtml = layoutManager.tv ? '' : '<span class="hide">: </s
             if (stream.Channels) {
                 attributes.push(createAttribute(globalize.translate('MediaInfoChannels'), `${stream.Channels} ch`));
             }
-            if (stream.BitRate && stream.Codec !== 'mjpeg') {
+            if (stream.BitRate && (stream.Type === 'Video' || stream.Type === 'Audio')) {
```
dmitrylyzo, line 140: `BitRate: undefined` for the embedded cover image (`mjpeg`), so I think even the old check is unnecessary.

src/components/itemMediaInfo/itemMediaInfo.js
nyanmisaka: Removed.

src/components/itemMediaInfo/itemMediaInfo.js
nyanmisaka: Removed.

dmitrylyzo (approved):

PR #3720, opened 2022-06-19
Fix device logos for various clients

Shadowghost:
Fix device logos for:
* JellyCon
* Jellyfin iOS (this is Swiftfin)
* Infuse

### DIFF
--- a/src/scripts/imagehelper.js
+++ b/src/scripts/imagehelper.js
@@ -11,6 +11,7 @@
             case 'Sony PS4':
                 return baseUrl + 'playstation.svg';
             case 'Kodi':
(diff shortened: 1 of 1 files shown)


nyanmisaka:
image
 Can you also add the Infuse client to Apple devices?

nielsvanvelzen:
Just wondering: do we add all possible clients, the ones mentioned on our site or just offficial ones?

Shadowghost:
I don't think we have a policy on that.
But IMHO we should not prevent clients from being added (as long a they are actively maintained)

joshuaboniface (approved):

PR #3721, opened 2022-06-19
Fix stream selection remembering

Shadowghost:
Server already exposes and uses this configuration.
Code was adopted to properly handle the decoupling of the setting into audio and subtitle.

I'm not sure if we even need all the code for automatic stream selection for the next item anymore since server should take care of that too. It works as is and I may take a look on moving all the logic to server in the future.


src/components/playback/playbackmanager.js
```diff
@@ -2385,8 +2388,18 @@ class PlaybackManager {
                 playOptions.items = null;
 
                 return getPlaybackMediaSource(player, apiClient, deviceProfile, maxBitrate, item, startPosition, mediaSourceId, audioStreamIndex, subtitleStreamIndex).then(function (mediaSource) {
-                    if (userSettings.enableSetUsingLastTracks())
-                        autoSetNextTracks(prevSource, mediaSource);
+                    const user = apiClient.getCurrentUser();
```
dmitrylyzo, line 2391: `getCurrentUser` returns Promise

Add `async` above:
```
                return getPlaybackMediaSource(player, apiClient, deviceProfile, maxBitrate, item, startPosition, mediaSourceId, audioStreamIndex, subtitleStreamIndex).then(async (mediaSource) => {
```

```suggestion
                    const user = await apiClient.getCurrentUser();
```

src/components/playback/playbackmanager.js
```diff
@@ -2385,8 +2388,18 @@ class PlaybackManager {
...
+                        if (user.Configuration.RememberSubtitleSelections)
+                            autoSetNextTracks(prevSource, mediaSource, true, true);
+                        else
+                            autoSetNextTracks(prevSource, mediaSource, true, false);
+                    } else {
+                        if (user.Configuration.RememberSubtitleSelections)
+                            autoSetNextTracks(prevSource, mediaSource, false, true);
+                        else
+                            autoSetNextTracks(prevSource, mediaSource, false, false);
+                    }
```
dmitrylyzo, line 2402: Why so complicated?
```suggestion
                    autoSetNextTracks(prevSource, mediaSource, user.Configuration.RememberAudioSelections, user.Configuration.RememberSubtitleSelections);
```

src/components/playback/playbackmanager.js
```diff
@@ -2319,8 +2319,11 @@ class PlaybackManager {
                     return;
                 }
 
-                rankStreamType(prevSource.DefaultAudioStreamIndex, prevSource, mediaSource, 'Audio');
-                rankStreamType(prevSource.DefaultSubtitleStreamIndex, prevSource, mediaSource, 'Subtitle');
+                if (audio)
+                    rankStreamType(prevSource.DefaultAudioStreamIndex, prevSource, mediaSource, 'Audio');
+
+                if (subtitle)
+                    rankStreamType(prevSource.DefaultSubtitleStreamIndex, prevSource, mediaSource, 'Subtitle');
```
dmitrylyzo, line 2326: afaik, most of the code is Google styled.
iirc, we allow simple body at the same line.
```suggestion
                if (audio) {
                    rankStreamType(prevSource.DefaultAudioStreamIndex, prevSource, mediaSource, 'Audio');
                }

                if (subtitle) {
                    rankStreamType(prevSource.DefaultSubtitleStreamIndex, prevSource, mediaSource, 'Subtitle');
                }
```

dmitrylyzo (commented):
When tested, I didn't get `Title`, but `DisplayTitle`: https://github.com/jellyfin/jellyfin-web/blob/2b0091eca20f0a8623a50f594c6f3b327d95f16e/src/components/playback/playbackmanager.js#L2278

Shadowghost:
> When tested, I didn't get `Title`, but `DisplayTitle`:
> 
> https://github.com/jellyfin/jellyfin-web/blob/2b0091eca20f0a8623a50f594c6f3b327d95f16e/src/components/playback/playbackmanager.js#L2278

Isn't that a backend bug then?
Applied the suggestions.

dmitrylyzo:
Subtitle selection doesn't work without the audio stream (I tested with video without audio tracks).
https://github.com/jellyfin/jellyfin-web/blob/22a77ce54e74d6ba9c49c7f6ce657e9147f400e0/src/components/playback/playbackmanager.js#L2312-L2320

Will audio be selected if there are no subtitles? :thinking:

Shadowghost:
Not sure if this is wanted though?

dmitrylyzo (approved):

PR #3724, opened 2022-06-20
Fix resume not working on LG WebOS

samcon:
Resume wasn't working for me on LG WebOS (via jellyfin-webos) on Jellyfin 10.8.0, jellyfin-webos 1.0.1 and latest jellyfin-web (master)

**Description**
I believe the issue is that:
- ?t=XXX is ignored on the video element in the WebOS browser
- Flawed logic in setCurrentTimeIfNeeded
- For some reason, changing currentTime in the 'playing' event (of the HTML video element) doesn't seem to do anything. I also tried 'loadeddata', 'loadedmetadata', 'canplay' but none of them worked. Only thing that worked is to delay the execution with setTimeout.
- Also possible that this only effects DirectStream MKV files

**Changes**
- I'm not sure what the original intention was, but `Math.abs(element.currentTime || 0, seconds)` is actually `Math.abs(element.currentTime || 0)`. I interpreted the logic was meant to not skip if the diff is smaller than 1 sec.
- Maybe someone has a less hack-y way to change (i.e. without using setTimeout) currentTime 

**Issues**
Probably solves jellyfin/jellyfin-webos#46

### DIFF
--- a/src/components/htmlMediaHelper.js
+++ b/src/components/htmlMediaHelper.js
@@ -131,7 +131,8 @@ import { Events } from 'jellyfin-apiclient';
     }
 
     function setCurrentTimeIfNeeded(element, seconds) {
-        if (Math.abs(element.currentTime || 0, seconds) <= 1) {
+        // If it's worth skipping (1 sec or less of a difference)
+        if (Math.abs((element.currentTime || 0) - seconds) >= 1) {
             element.currentTime = seconds;
         }
     }
--- a/src/plugins/htmlVideoPlayer/plugin.js
+++ b/src/plugins/htmlVideoPlayer/plugin.js
@@ -1375,6 +1375,9 @@ function tryRemoveElement(elem) {
                         // Can't autoplay in these browsers so we need to use the full controls, at least until playback starts
                         if (!appHost.supports('htmlvideoautoplay')) {
                             html += '<video class="' + cssClass + '" preload="metadata" autoplay="autoplay" controls="controls" webkit-playsinline playsinline>';
+                        } else if (browser.web0s) {
+                            // in webOS, setting peload auto allows resuming videos
+                            html += '<video class="' + cssClass + '" preload="auto" autoplay="autoplay" webkit-playsinline playsinline>';
                         } else {
                             // Chrome 35 won't play with preload none
                             html += '<video class="' + cssClass + '" preload="metadata" autoplay="autoplay" webkit-playsinline playsinline>';
--- a/src/plugins/htmlVideoPlayer/plugin.js
+++ b/src/plugins/htmlVideoPlayer/plugin.js
@@ -1376,7 +1376,7 @@ function tryRemoveElement(elem) {
                         if (!appHost.supports('htmlvideoautoplay')) {
                             html += '<video class="' + cssClass + '" preload="metadata" autoplay="autoplay" controls="controls" webkit-playsinline playsinline>';
                         } else if (browser.web0s) {
-                            // in webOS, setting peload auto allows resuming videos
+                            // in webOS, setting preload auto allows resuming videos
                             html += '<video class="' + cssClass + '" preload="auto" autoplay="autoplay" webkit-playsinline playsinline>';
                         } else {
                             // Chrome 35 won't play with preload none


src/components/htmlMediaHelper.js
```diff
@@ -131,8 +131,8 @@ import { Events } from 'jellyfin-apiclient';
     }
 
     function setCurrentTimeIfNeeded(element, seconds) {
-        if (Math.abs(element.currentTime || 0, seconds) <= 1) {
-            element.currentTime = seconds;
+        if (Math.abs(element.currentTime || 0) - seconds <= 1) {
+            setTimeout(function(element, seconds){ element.currentTime = seconds; }, 500, element, seconds);
```
dmitrylyzo, line 135: I can't test this right now, but I believe that you just need to schedule seeking right after current event.
```suggestion
            setTimeout(() => {
                element.currentTime = seconds;
            }, 0);
```
Maybe also leave a comment.

dmitrylyzo:
What webOS version are you testing?
Older versions can behave oddly, so timeouts are not unusual.

Since this is a bug, please you rebase this branch onto `release-10.8.z` and edit target branch to the same (without reopening a new PR).

src/components/htmlMediaHelper.js
```diff
@@ -131,8 +131,8 @@ import { Events } from 'jellyfin-apiclient';
     }
 
     function setCurrentTimeIfNeeded(element, seconds) {
-        if (Math.abs(element.currentTime || 0, seconds) <= 1) {
-            element.currentTime = seconds;
+        if (Math.abs(element.currentTime || 0) - seconds <= 1) {
```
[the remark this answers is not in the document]
samcon: Ha! Gotta love JS!

samcon:
> What webOS version are you testing? Older versions can behave oddly, so timeouts are not unusual.
> 
> Since this is a bug, please rebase this branch on `release-10.8.z` and edit target branch to the same (without reopening a new PR).



> release-10.8.z

WebOS 4.5

Will do

dmitrylyzo:
Don't forget to reapply suggestions. You are free to squash commits to a first one.

src/components/htmlMediaHelper.js
```diff
@@ -131,8 +131,11 @@ import { Events } from 'jellyfin-apiclient';
     }
 
     function setCurrentTimeIfNeeded(element, seconds) {
-        if (Math.abs(element.currentTime || 0, seconds) <= 1) {
-            element.currentTime = seconds;
+        // If it's worth skipping (1 sec or less of a difference)
+        if (Math.abs((element.currentTime || 0) - seconds) <= 1) {
```
dmitrylyzo, line 135: Looks like we have inverted it - should be `>= 1`.
But looking at this:
https://github.com/jellyfin/jellyfin-web/blob/da2e3d15be6e5ba668b2f2ca2c92fa3486c51149/src/components/htmlMediaHelper.js#L158-L163
I am not sure about which one is right - need to test.
`setCurrentTimeIfNeeded` is currently supposed to seek if `currentTime <= 1` (as you have noticed).

samcon: Changed back to the correct check

I think this comment applies to the previous check (`element.currentTime === 0 && element.duration >= seconds`), but not sure either.

dmitrylyzo: > I think this comment applies to the previous check (`element.currentTime === 0 && element.duration >= seconds`), but not sure either.

Ah, right.

At least with `>= 1` it is logically valid in compare with `Math.abs(element.currentTime || 0) <= 1`, which is `(element.currentTime || 0) <= 1` since `currentTime` is non-negative.

dmitrylyzo (commented):
I have played with timeouts: 50ms worked too (once), but 300 worked more often. So there's a chance that 500ms may fail too. It probably depends on buffering time.

I have tried events `durationchange`, `loadeddata`, `play`, `loadedmetadata`: they all go before `seekOnPlaybackStartIfNeeded` (`play` goes first).
I haven't tried `timeupdate` though 🤔 Something like: seek if needed on the first time update.

src/components/htmlMediaHelper.js
```diff
@@ -131,8 +131,11 @@ import { Events } from 'jellyfin-apiclient';
     }
 
     function setCurrentTimeIfNeeded(element, seconds) {
-        if (Math.abs(element.currentTime || 0, seconds) <= 1) {
-            element.currentTime = seconds;
+        // If it's worth skipping (1 sec or less of a difference)
+        if (Math.abs((element.currentTime || 0) - seconds) >= 1) {
+            // Changing currentTime directly doesn't seek in the video
+            // A <=500 timeout doesn't seem to work for my setup
+            setTimeout(() => { element.currentTime = seconds; }, 500);
```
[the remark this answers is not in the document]
samcon: Probably worth more investigation in the future...
I also tried using the "Using mediaOption to Avoid Needless Preload" [described here](http://webcache.googleusercontent.com/search?q=cache:DCB3EzosswEJ:https://webostv.developer.lge.com/develop/developer-guide/resuming-media-quickly-mediaoption/&hl=en&gl=il&strip=1&vwsrc=0) but to no avail.

Great suggestions! Pushed

dmitrylyzo (approved):
LGTM.

Maybe also add `HACK: ` somewhere to mark it as an unfinished/non-ideal solution?

dmitrylyzo (approved):

samcon:
I did some more testing and noticed subtitles aren't synced to the resumed time. 
If subs aren't supposed to sync automatically when currentTime is set, any pointers on how to "manually" invoke sub sync?

dmitrylyzo:
Tested unmodified jf-web on emulators of webOS 1.2, 2, 4 and 5 - only 4 fails :man_shrugging: 
If this is true for real TVs, we need to check the `web0sVersion`.

> I did some more testing and noticed subtitles aren't synced to the resumed time. If subs aren't supposed to sync automatically when currentTime is set, any pointers on how to "manually" invoke sub sync?

What subs? SRT or ASS, embedded or external.

samcon:
External SRT file

dmitrylyzo:
In webOS 4 emulator `currentTime` and subtitles are synchronized and updates normally, but the video updates slowly.
Progress line (and its text) is out of sync too. Do you have the same?
I tried it from 0s position.

**Test assets**

https://user-images.githubusercontent.com/56478732/175397912-000f0065-7782-4dcd-bd8f-857eefa7ed4b.mp4

[jf-gray_1080p_h264_m40.time.srt.txt](https://github.com/jellyfin/jellyfin-web/files/8970910/jf-gray_1080p_h264_m40.time.srt.txt)

samcon:
Alright, I did some more digging and it turns out that if you change 'metadata' here:
https://github.com/jellyfin/jellyfin-web/blob/70284199a3b755a8d13bd3ba6edbab696b82d6b4/src/plugins/htmlVideoPlayer/plugin.js#L1377-L1378
to 'auto' then resume simply (i.e. w/o the timer) works 🤷‍♂️.

Subtitles still don't sync to the correct position though

dmitrylyzo:
I like hack-free solutions. 👍 

Does it work with transcoding?

`autoplay` is supposed to force a full preload, but taking into account quirks of browsers, it is better to specify a full preload explicitly. Moreover, I think we can set it to `auto` for all cases (**need testing**), because we are definitely going to play it.

> Subtitles still don't sync to the correct position though

Have you tried the sample above? In the emulator subtitles are out of sync even from the beginning.

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1375,6 +1375,9 @@ function tryRemoveElement(elem) {
                         // Can't autoplay in these browsers so we need to use the full controls, at least until playback starts
                         if (!appHost.supports('htmlvideoautoplay')) {
                             html += '<video class="' + cssClass + '" preload="metadata" autoplay="autoplay" controls="controls" webkit-playsinline playsinline>';
+                        } else if (browser.web0s) {
+                            // in webOS, setting peload auto allows resuming videos
```
dmitrylyzo, line 1379:
```suggestion
                            // in webOS, setting preload auto allows resuming videos
```

joshuaboniface (approved):

PR #3730, opened 2022-06-28
CardBuilder: Test episode number in addition to episode name

SenorSmartyPants:
Display episode number on card if present. Some episodes can have no name, but are still numbered (in my EPG data).

**Before**
image

**After**
image


thornbill (approved):

SenorSmartyPants:
Thanks

PR #3731, opened 2022-06-28
Add vscode configuration to fix issues on saving

thornbill:
**Changes**
* Adds project settings for VS Code
  * Adds eslint as a recommended plugin
  * Adds settings to fix eslint issues on save

Similar configuration is already used for `jellyfin-expo`, `jellyfin-sdk-typescript`, and `jellyfin-vue`.

**Issues**
N/A

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,6 @@ config.json
 
 # ide
 .idea
-.vscode
 
 # log
 yarn-error.log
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -3,12 +3,8 @@ dist
 web
 node_modules
 
-# config
-config.json
```
thornbill, line 7: This is a leftover from when we had a default config file in the repo that was named differently and users were expected to copy it to `config.json` to make edits.

grafixeyehero (approved):

PR #3742, opened 2022-07-01
Fix unhandled rejection for invalid response

thornbill:
**Changes**
If a server returns an "OK" response from `/System/Info/Public`, but the response body is not valid JSON an unhandled promise rejection would occur.

**Issues**
N/A

### DIFF
--- a/src/utils/dashboard.js
+++ b/src/utils/dashboard.js
@@ -57,10 +57,12 @@ export async function serverAddress() {
                 }
 
                 return {
-                    url: url,
+                    url,
                     config: await resp.json()
                 };
-            }).catch(() => { /* swallow errors */ });
+            }).catch(error => {
+                console.error(error);
+            });
     });
 
     return Promise.all(promises).then(responses => {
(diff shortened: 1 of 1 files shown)


src/utils/dashboard.js
```diff
@@ -50,30 +50,26 @@ export async function serverAddress() {
...
-        });
+        return fetch(`${url}/System/Info/Public`)
+            .then(async resp => {
+                if (!resp.ok) {
+                    return;
+                }
+
+                return {
+                    url: url,
+                    config: await resp.json()
```
thornbill, line 61: Since we are using async/await any errors thrown will be handled by the `.catch`

src/utils/dashboard.js
```diff
@@ -50,30 +50,26 @@ export async function serverAddress() {
...
+                };
+            }).catch(() => { /* swallow errors */ });
```
cvium, line 63: Logging isn't necessary here?

cvium (approved):

PR #3744, opened 2022-07-03
Add autoplay timer to slideshow.js. 

MCBuckethead:
**Motivation**
I felt the slide show went through the items too quickly and it didn't have any customization.

**Changes**
Added a new button to slideshowBottomBar, btnAutoplayTimer.
Added function triggered by btnAutoplayTimer, showDelayMenu.
This function opens a actionSheet with 4 items. Each item is a different delay option in seconds: 1, 3, 5, 10.
Selecting one of those options will also start the autoplay if it currently is not.

**Before View**
image

**After View**
image
image

**Possible Next Steps**
Could use a cleaner way to add new options to playback, right now it's just hard coded.

### DIFF
--- a/src/components/slideshow/slideshow.js
+++ b/src/components/slideshow/slideshow.js
@@ -296,7 +296,7 @@ export default function (options) {
         },
         {
             id: '5000',
-            name: '5 second'
+            name: '5 seconds'
         },
         {
             id: '10000',
@@ -307,7 +307,7 @@ export default function (options) {
             items,
             positionTo: button
         };
-        
+
         actionSheet.show(actionSheetOptions).then(function (result) {
             if (result) {
                 swiperInstance.params.autoplay.delay = result;
--- a/src/components/slideshow/slideshow.js
+++ b/src/components/slideshow/slideshow.js
@@ -17,6 +17,7 @@ import ServerConnections from '../ServerConnections';
 import { Swiper } from 'swiper/swiper-bundle.esm';
 import 'swiper/swiper-bundle.css';
 import screenfull from 'screenfull';
+import actionSheet from '../actionSheet/actionSheet';
 
 /**
  * Name of transition event.
@@ -192,6 +193,7 @@ export default function (options) {
                 html += '<div class="slideshowBottomBar hide">';
 
                 html += getIcon('play_arrow', 'btnSlideshowPause slideshowButton', true, true);
+                html += getIcon('timer', 'btnAutoplayTimer', true, true);
                 if (appHost.supports('filedownload') && options.user && options.user.Policy.EnableContentDownloading) {
                     html += getIcon('file_download', 'btnDownload slideshowButton', true);
                 }
@@ -222,6 +224,13 @@ export default function (options) {
             const btnPause = dialog.querySelector('.btnSlideshowPause');
             if (btnPause) {
                 btnPause.addEventListener('click', getClickHandler(playPause));
+            }
+
+            const btnAutoplayTimer = dialog.querySelector('.btnAutoplayTimer');
+            if (btnAutoplayTimer) {
+                btnAutoplayTimer.addEventListener('click', function (e) {
+                    showDelayMenu(e.target);
+                });
             }
 
             const btnDownload = dialog.querySelector('.btnDownload');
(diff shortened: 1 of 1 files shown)


src/components/slideshow/slideshow.js
```diff
@@ -276,6 +285,36 @@ export default function (options) {
...
+            id: '3000',
+            name: '3 seconds'
+        },
+        {
+            id: '5000',
+            name: '5 seconds'
+        },
+        {
+            id: '10000',
+            name: '10 seconds'
```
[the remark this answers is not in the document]
MCBuckethead: There was already a "{0} seconds" there so I used that.

src/components/slideshow/slideshow.js
```diff
@@ -276,6 +285,36 @@ export default function (options) {
...
+        }];
+
+        const actionSheetOptions = {
+            items,
+            positionTo: button
+        };
+
+        actionSheet.show(actionSheetOptions).then(function (result) {
+            if (result) {
+                swiperInstance.params.autoplay.delay = result;
```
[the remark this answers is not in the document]
MCBuckethead: Added to userSettings.

src/components/slideshow/slideshow.js
```diff
@@ -344,6 +381,9 @@ export default function (options) {
             slides = currentOptions.items;
         }
 
+        const autoplayDelay = userSettings.get('autoplayDelay', true) || 3000;
+        const autoplay = options.autoplay ? {delay:autoplayDelay} : false;
```
dmitrylyzo, line 385: _Not tested_
```suggestion
        const autoplay = !options.interactive || !!options.autoplay ? { delay: autoplayDelay } : false;
```
But I was also thinking about specifying a delay in the options.

src/components/slideshow/slideshow.js
```diff
@@ -352,7 +392,7 @@ export default function (options) {
                 minRatio: 1,
                 toggle: true
             },
-            autoplay: !options.interactive || !!options.autoplay,
+            autoplay: !options.interactive || autoplay,
```
dmitrylyzo, line 395:
```suggestion
            autoplay,
```

src/components/slideshow/slideshow.js
```diff
@@ -276,14 +287,40 @@ export default function (options) {
...
      * Handles OSD changes when the autoplay is started.
      */
-    function onAutoplayStart() {
+    function onAutoplayStart(autoplayDelay) {
         const btnSlideshowPause = dialog.querySelector('.btnSlideshowPause .material-icons');
         if (btnSlideshowPause) {
             btnSlideshowPause.classList.replace('play_arrow', 'pause');
         }
+        if (autoplayDelay) {
+            swiperInstance.params.autoplay.delay = autoplayDelay;
```
dmitrylyzo, line 322: Isn't initialization enough?

src/components/slideshow/slideshow.js
```diff
@@ -17,6 +17,9 @@ import ServerConnections from '../ServerConnections';
 import { Swiper } from 'swiper/swiper-bundle.esm';
 import 'swiper/swiper-bundle.css';
 import screenfull from 'screenfull';
+import actionSheet from '../actionSheet/actionSheet';
+import globalize from '../../scripts/globalize';
+import * as userSettings from '../../scripts/settings/userSettings';
```
dmitrylyzo, line 22: _Not tested_
```suggestion
import { currentSettings as userSettings } from '../../scripts/settings/userSettings';
```
To avoid those backward compatibility exports.

src/components/slideshow/slideshow.js
MCBuckethead: Starting with shuffle or play all then using the play/pause button works as expected without this, but if the user directly opens an image and then presses play it uses the default delay without this.

src/components/slideshow/slideshow.js
MCBuckethead: Done

src/components/slideshow/slideshow.js
MCBuckethead: Done

src/components/slideshow/slideshow.js
MCBuckethead: Done. I looked into a bit about doing it with options, but I'm not familiar enough with the project to know the best way to do it. The current way seems fine to me for now.

src/components/slideshow/slideshow.js
MCBuckethead: After testing more, there's still something wrong with this. Some part of pausing the slideshow by dragging the image causes the wrong delay to be used after playing. It only happens sometimes though, so I'll keep investigating.

thornbill: Hey @MCBuckethead did you have any luck fixing the issue you mentioned here?

thornbill:
@jellyfin-bot rebase

thornbill:
@MCBuckethead Now that 10.9 is out let's work on getting this in. Could you fix the merge conflicts?

markster1700:
At best default should be more reasonable value eg: 7seconds. Current 3 seconds is not appropriate.
Also, looking at 2022 solution 1,3,5,10 seconds ?? Why would anyone select 1 or 3 seconds ?

src/components/slideshow/slideshow.js
markster1700: Has this being abandoned? I wanted to change the default from 3 to 7 seconds per slide and it was rejected since this is a better option. But if this is going nowhere maybe setting the default to 7 sec is not a bad idea.

PR #3745, opened 2022-07-03
Add chapter markings to video player slider

Viperinius:
**Motivation**
I wanted to be able to see when movie chapters start without moving the slider bubble through the whole timeline. Also inspired by [this feature request](https://features.jellyfin.org/posts/397/chapter-markers-in-timeline).

**Changes**
Added container for chapter marks to the video OSD page.
Added functions to add and calculate the icon positions.
Those calculations are done in updateValues of the slider.
The markings are shown or hidden together with / the same way as the slider bubble.
When the chapter mark has been passed / watched, the colour changes to blue.
The icon is set with CSS and can thus be easily customised by the user.

**What it looks like**
(shown with the Jungle Book from the demo server)
image


src/controllers/playback/video/index.js
```diff
@@ -1574,6 +1574,28 @@ import { setBackdropTransparency, TRANSPARENCY_LEVEL } from '../../../components
             return '<h1 class="sliderBubbleText">' + datetime.getDisplayRunningTime(ticks) + '</h1>';
         };
 
+        nowPlayingPositionSlider.getChapterFractions = function () {
+            showOsd();
```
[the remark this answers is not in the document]
Viperinius: Removed it, that was just a remainder of some copy pasting, whoops.

src/controllers/playback/video/index.js
```diff
@@ -1574,6 +1574,30 @@ import { setBackdropTransparency, TRANSPARENCY_LEVEL } from '../../../components
...
+                const chapterNames = [];
+                const chapterFractions = [];
+                const runtimeDuration = item.RunTimeTicks;
+
+                for (let i = 0, length = item.Chapters.length; i < length; i++) {
+                    const currentChapter = item.Chapters[i];
+
+                    const fraction = currentChapter.StartPositionTicks / runtimeDuration;
+                    chapterFractions.push(fraction);
+                    chapterNames.push(currentChapter.Name);
```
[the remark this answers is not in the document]
Viperinius: Agreed on returning an object instead of the "double array".

I changed all names to be more generic so that the `className` can be used to specify different kinds of markers, if ever needed. Does this new way fit your idea?

src/elements/emby-slider/emby-slider.js
```diff
@@ -130,6 +137,57 @@ import '../emby-input/emby-input';
...
+            const bubbleTrackRect = range.sliderBubbleTrack.getBoundingClientRect();
+            const chapterMarkRect = chapterMark.getBoundingClientRect();
+
+            if (!bubbleTrackRect.width || !chapterMarkRect.width) {
+                // width is not set, most probably because the OSD is currently hidden
+                return;
+            }
+
+            let chapterMarkPos = (bubbleTrackRect.width * valueChapterMark / 100) - chapterMarkRect.width / 2;
+            chapterMarkPos = Math.min(Math.max(chapterMarkPos, - chapterMarkRect.width / 2), bubbleTrackRect.width - chapterMarkRect.width / 2);
```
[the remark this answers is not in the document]
Viperinius: Well, the early return if the width is 0 was just meant to prevent recalculating the marker positions even if the OSD is hidden. This lead to noticeable jumps of the markers when the OSD is shown again and the width is correct again.

About the resizing: Could you elaborate?
Do you mean if the positioning can handle a resize during playback? If so, then yes, as long as updateValues is continued to be called.

src/controllers/playback/video/index.js
```diff
@@ -1574,6 +1574,31 @@ import { setBackdropTransparency, TRANSPARENCY_LEVEL } from '../../../components
             return '<h1 class="sliderBubbleText">' + datetime.getDisplayRunningTime(ticks) + '</h1>';
         };
 
+        nowPlayingPositionSlider.getMarkerInfo = function () {
+            const markers = [];
+
+            const item = currentItem;
+
+            // use markers based on chapters
+            if (item && item.Chapters && item.Chapters.length) {
```
thornbill, line 1583:
```suggestion
            if (item?.Chapters?.length) {
```

src/controllers/playback/video/index.js
```diff
@@ -1574,6 +1574,31 @@ import { setBackdropTransparency, TRANSPARENCY_LEVEL } from '../../../components
...
+                const runtimeDuration = item.RunTimeTicks;
```
thornbill, line 1584: This assignment is unnecessary.

src/controllers/playback/video/index.js
```diff
@@ -1574,6 +1574,31 @@ import { setBackdropTransparency, TRANSPARENCY_LEVEL } from '../../../components
...
+
+                for (let i = 0, length = item.Chapters.length; i < length; i++) {
```
thornbill, line 1586: I think `forEach` would work better here.

```js
                item.Chapters.forEach(currentChapter => { ... });
```

src/controllers/playback/video/index.js
```diff
@@ -1574,6 +1574,31 @@ import { setBackdropTransparency, TRANSPARENCY_LEVEL } from '../../../components
...
+                    const currentChapter = item.Chapters[i];
+
+                    const fraction = currentChapter.StartPositionTicks / runtimeDuration;
```
thornbill, line 1589: No need to assign this to a variable either.

thornbill (changes requested):
First off this is a really cool feature! :tada: 

I did not finish reviewing the `emby-slider.js` changes, because I'm not a fan of how the hiding works currently. I think the markers should just be lines that show up inline with the slider that are just visible anytime the slider is shown. The large icons make me think I should be able to click them, and they are kind intrusive and difficult to get to display.

Viperinius:
> I think the markers should just be lines that show up inline with the slider that are just visible anytime the slider is shown. The large icons make me think I should be able to click them, and they are kind intrusive and difficult to get to display.

Yeah, I get that, especially being able to click them. I quickly experimented with the CSS a bit and got this:
image
Is this what you had in mind?

(As these are just CSS changes, the "icon way" can still be used with custom server CSS.)

thornbill:
That looks great!

Viperinius:
Alright, I added the changed CSS.

Concerning the hiding / positioning, I am open to any suggestions to improve it

thornbill:
The changes look great!

I think I would just leave the markers visible when the OSD is shown instead of requiring hover.

Viperinius:
Good suggestion, saves a bit of code... and done :)

thornbill (approved):

DoTheSneedful:
I really want this feature but I noticed it's not in any release since being merged to master over a year ago, certainly not available in the 10.8 version I'm currently using. I am not familiar with Jellyfin's process, so I hope it's not too much trouble to ask what the ETA for something like this is?

PR #3747, opened 2022-07-04
Feature/add prettier

Squixx:
**Changes**
Modernize code formatting by adding prettier.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,13 +2,7 @@ const restrictedGlobals = require('confusing-browser-globals');
 
 module.exports = {
     root: true,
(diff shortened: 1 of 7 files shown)


Squixx:
yeah i was actually trying to add NX & storybook & cypress. Thought this would be usefull by itself as I ran out of time. I'll circle back to that as that many packages doesnt make sense

ThibaultNocchi:
I'm pretty sure the update to the package-lock file is due to @Squixx maybe using a version of npm >= 7 that updates the lockfile version to 2. You could try resetting both package-*.json files and use npm 6 to adds the package.

PR #3768, opened 2022-07-15
Fix TypeError

yahuli:
**Fix**

Fix when item.Id is a Integer, it will throw a Error.
error

**Issues**


thornbill:
This would probably be good to include in the next 10.8 bugfix release. Do you think you could rebase this change on the `release-10.8.z` branch and change the target branch of the PR to also be that branch? Thanks!

PR #3792, opened 2022-08-05
Convert Movies to react

grafixeyehero:


src/elements/ItemsContainerElement.tsx
```diff
@@ -0,0 +1,28 @@
...
+    >
+    </div>`
+});
+
+type IProps = {
+    id?: string;
+    className?: string;
+}
+
+const ItemsContainerElement: FunctionComponent<IProps> = ({ id, className }: IProps) => {
```
vitorsemeano, line 17: You can use the shorthand FC, it's the same thing.

vitorsemeano: If you define Iprops in the generic FunctionComponent (FC), arguments should be automatically infered by it, so i shouldn't be needed a second declaration right after the first argument.

src/elements/ItemsContainerElement.tsx
```diff
@@ -0,0 +1,28 @@
...
+const createElement = ({ id, className }: IProps) => ({
+    __html: `<div
+        is="emby-itemscontainer"
+        ${id}
+        class="${className}"
+    >
+    </div>`
+});
+
+type IProps = {
```
vitorsemeano, line 12: prefer interface over type for these kind of typings.

src/elements/ItemsScrollerContainerElement.tsx
```diff
@@ -0,0 +1,43 @@
...
+    <div
+        is="emby-itemscontainer"
+        ${id}
+        class="${className}"
+    >
+    </div>
+    </div>`
+});
+
+type IProps = {
```
vitorsemeano, line 19: The same suggestions given in ItemsContainerElement can be applied here

src/routes/movies.tsx
```diff
@@ -0,0 +1,139 @@
...
+};
+
+const Movies: FunctionComponent = () => {
+    const [ searchParams ] = useSearchParams();
+    const currentTabIndex = parseInt(searchParams.get('tab') || getDefaultTabIndex(searchParams.get('topParentId')).toString());
+    const [ selectedIndex, setSelectedIndex ] = useState(currentTabIndex);
+    const element = useRef<HTMLDivElement>(null);
+
+    const getTabs = () => {
+        return [{
```
vitorsemeano, line 47: you can create a const mapping outside of the component to map these indexes to their corresponding translation and Components. Then you only need to iterate that mapping or fetch the component based on the tab id.

src/routes/movies.tsx
```diff
@@ -0,0 +1,139 @@
...
+    }, []);
+
+    useEffect(() => {
+        const page = element.current;
+
+        if (!page) {
+            console.error('Unexpected null reference');
+            return;
+        }
+        mainTabsManager.setTabs(element.current, selectedIndex, getTabs, undefined, undefined, onTabChange);
```
vitorsemeano, line 109:
```suggestion
        mainTabsManager.setTabs(page, selectedIndex, getTabs, undefined, undefined, onTabChange);
```

src/elements/ItemsContainerElement.tsx
thornbill: I'm still pretty new to Typescript. Why would an interface be preferred here? Most of the information I can find says it doesn't really matter, but types might be preferable since they support some things interfaces do not.

vitorsemeano: well interfaces resemble objects/classes/components. The way i see it, when you have components or objects, you can usually represent using a interface. They can also be extended, inheriting the super interface properties.

When you want to represent some helper, like Record, or array or literal strings, you need to use types. In some cases, a interface signature requires types to function property in conjunction with generics, but that's a little more advanced ahah.

I am no expert, but i normally use it that way.

src/view/components/AlphaPickerContainer.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { FunctionComponent, useCallback, useEffect, useRef, useState } from 'react';
+import AlphaPicker from '../../components/alphaPicker/alphaPicker';
+import { IQuery } from './type';
+
+type AlphaPickerProps = {
```
vitorsemeano, line 5: prefer a interface here

vitorsemeano: also, is more readable using full interface/type names than "I" to represent a interface/type.

src/view/components/AlphaPickerContainer.tsx
```diff
@@ -0,0 +1,49 @@
...
+            delete query.NameStartsWith;
+        } else {
+            query.NameStartsWith = newValue;
+            delete query.NameLessThan;
+        }
+        query.StartIndex = 0;
+        reloadItems();
+    }, [query, reloadItems]);
+
+    useEffect(() => {
```
vitorsemeano, line 29: on each change, you are not unmounting the useEffect, removing the event listener in "alphavaluechanged".

If this is invoked 100 times, there will be 100 event listeners in the element.

also if the reference element is used to grab the div with the class 'alphaPicker', you can pass the ref to that dom element.

src/view/components/AlphaPickerContainer.tsx
```diff
@@ -0,0 +1,49 @@
...
+const AlphaPickerContainer: FunctionComponent<AlphaPickerProps> = ({ query, reloadItems }: AlphaPickerProps) => {
+    const [ alphaPicker, setAlphaPicker ] = useState<AlphaPicker>();
+    const element = useRef<HTMLDivElement>(null);
+
+    alphaPicker?.updateControls(query);
+
+    const onAlphaPickerChange = useCallback((e) => {
+        const newValue = (e as CustomEvent).detail.value;
+        if (newValue === '#') {
+            query.NameLessThan = 'A';
```
vitorsemeano, line 19: changing a prop directly like this is not recommended and very prune to errors.

You then call reloadItems, but if someone forgets to call this, in a future update, is very prune to it and can be hard to find the cause if you don't know what you need to call for this param to "react". 

So my suggestion is using useState in the parent component where the query is defined. Then pass the setQuery to the children that needs to modify this state. then call setQuery with a callback and do your logic inside it and return a modified query (you need to create a new object inside that callback so react can detect a new object was changed).

Another approach is using useReducer. Same thing but organized in a different manner. (more centralized so to speak)

src/elements/ItemsContainerElement.tsx
grafixeyehero: @vitorsemeano https://github.com/jellyfin/jellyfin-web/pull/3792/commits/2b40dbd50f39f56924b0d7ee6a3cab2dab38cbe4

src/elements/ItemsContainerElement.tsx
```diff
@@ -0,0 +1,28 @@
...
+    id?: string;
+    className?: string;
+}
+
+const ItemsContainerElement: FunctionComponent<IProps> = ({ id, className }: IProps) => {
+    return (
+        <div
+            dangerouslySetInnerHTML={createElement({
```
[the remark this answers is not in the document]
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/3792/commits/2b40dbd50f39f56924b0d7ee6a3cab2dab38cbe4

src/elements/ItemsScrollerContainerElement.tsx
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/3792/commits/2b40dbd50f39f56924b0d7ee6a3cab2dab38cbe4

src/routes/movies.tsx
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/3792/commits/2b40dbd50f39f56924b0d7ee6a3cab2dab38cbe4

src/routes/movies.tsx
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/3792/commits/2b40dbd50f39f56924b0d7ee6a3cab2dab38cbe4

src/view/components/AlphaPickerContainer.tsx
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/3792/commits/9df95d4fcbea0f59809e56ad4fa4b1c92afe0081

src/view/components/AlphaPickerContainer.tsx
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/3792/commits/9df95d4fcbea0f59809e56ad4fa4b1c92afe0081

src/view/components/AlphaPickerContainer.tsx
vitorsemeano: kind of resolves the issue, but it's still the same, relying on sub states to trigger an update to the query.

You should try to explore useReducer or using "setQuery" (with useState) with a callback and put your logic inside the callback to change or not the query state.

I could do it, but i don't have the time right now.

src/view/components/AlphaPickerContainer.tsx
vitorsemeano: Normally i use the full component name plus Props (ex: AlphaPickerContainerProps). This way you have an interface that represents the properties/props of the component in question. If you have the time and you agree with this, you can change. otherwise this can be marked as resolved.

src/view/components/AlphaPickerContainer.tsx
grafixeyehero: @vitorsemeano https://github.com/jellyfin/jellyfin-web/pull/3792/commits/b88f09c63d7b75b97ebc8c4539f98bc8ecb1d136.

for the rest of query state SelectView, Sort and Filter dialog must convert to react. current implementation it just warps the button

grafixeyehero: @vitorsemeano updated [use single state](https://github.com/jellyfin/jellyfin-web/pull/3792/commits/1b07551f538be313e56865263fdddcecda759801)

src/view/hook/useLocalStorage.tsx
```diff
@@ -0,0 +1,20 @@
+import { useEffect, useState } from 'react';
```
thornbill, line 1: I think we should have a "top" level hooks folder (i.e. `src/hooks`) and move this there.

src/view/movies/CollectionsView.tsx
```diff
@@ -0,0 +1,35 @@
+import React, { FC, useCallback } from 'react';
+
+import ViewItemsContainer from '../components/ViewItemsContainer';
```
thornbill, line 3: I think these should go in `src/routes/movies`... I know they don't currently represent routes really, but imho it would make sense for them to be someday.

src/view/components/AlphaPickerContainer.tsx
```diff
@@ -0,0 +1,53 @@
+import React, { FC, useCallback, useEffect, useRef, useState } from 'react';
```
thornbill, line 1: I think these should go under `/src/components` somewhere. Maybe have an `itemsList` or `itemsView` or just `items` subfolder there?

src/view/components/interface.ts
```diff
@@ -0,0 +1,117 @@
+import type { BaseItemDto } from '@jellyfin/sdk/lib/generated-client';
```
thornbill, line 1: Should this be a type definition file? i.e. `interface.d.ts`

src/view/components/interface.ts
```diff
@@ -0,0 +1,117 @@
...
+
+export interface QueryI extends ViewSettingsI {
```
thornbill, line 3: Can we exclude the `I` prefixes and suffixes? I've never been a fan of that convention in Java either. :sweat_smile:

src/view/components/ViewItemsContainer.tsx
```diff
@@ -0,0 +1,386 @@
...
+        value: 'Runtime,SortName,ProductionYear'
+    }];
+};
+
+const defaultViewSettingsValue: ViewSettingsI = {
+    showTitle: true,
+    showYear: true,
+    imageType: 'primary',
+    viewType: '',
+    cardLayout: true,
```
thornbill, line 103: I believe this should be `false` by default.

thornbill (changes requested):

src/view/components/AlphaPickerContainer.tsx
grafixeyehero: i am thinking like this  /src/components/common/

src/view/components/interface.ts
grafixeyehero: no need here `d.ts` files are treated as an ambient (global) module declarations only if they don't have any imports.

src/view/components/ViewItemsContainer.tsx
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/3792/commits/d7e48d30b6480d3af9083b3aa36eebb4331a2d3f

src/view/components/interface.ts
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/3792/commits/d7e48d30b6480d3af9083b3aa36eebb4331a2d3f

src/view/components/AlphaPickerContainer.tsx
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/3792/commits/d7e48d30b6480d3af9083b3aa36eebb4331a2d3f

src/view/movies/CollectionsView.tsx
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/3792/commits/d7e48d30b6480d3af9083b3aa36eebb4331a2d3f

src/view/hook/useLocalStorage.tsx
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/3792/commits/d7e48d30b6480d3af9083b3aa36eebb4331a2d3f

src/routes/movies/SuggestionsView.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+    const loadSuggestions = useCallback((page, userId) => {
+        const screenWidth = dom.getWindowSize().innerWidth;
+        let itemLimit = 5;
+        if (screenWidth >= 1600) {
+            itemLimit = 8;
+        } else if (screenWidth >= 1200) {
+            itemLimit = 6;
+        }
+        const url = window.window.ApiClient.getUrl('Movies/Recommendations', {
```
thornbill, line 89:
```suggestion
        const url = window.ApiClient.getUrl('Movies/Recommendations', {
```

src/routes/movies/CollectionsView.tsx
```diff
@@ -0,0 +1,35 @@
+import React, { FC, useCallback } from 'react';
+
+import ViewItemsContainer from '../../components/common/ViewItemsContainer';
+
+interface CollectionsViewI {
```
thornbill, line 5: It looks like we could create a `LibraryView` (or similarly named) interface for all these views since they accept the same prop.

thornbill (changes requested):
Just a couple minor comments then hopefully we can finally get this in :crossed_fingers:

grafixeyehero:
@thornbill  which one you prefer for genres view?.

current 
genres

new
genres-2

src/routes/movies/CollectionsView.tsx
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/3792/commits/1c6b1fc4785168b159ae8c42bf89f5d7cc1ed248

src/routes/movies/SuggestionsView.tsx
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/3792/commits/1c6b1fc4785168b159ae8c42bf89f5d7cc1ed248

thornbill:
I prefer the current UI

thornbill:
Looks like the one action failed due to a type error:

```
ERROR in /home/runner/work/jellyfin-web/jellyfin-web/src/routes/movies/GenresView.tsx
./routes/movies/GenresView.tsx 24:21-33
[tsl] ERROR in /home/runner/work/jellyfin-web/jellyfin-web/src/routes/movies/GenresView.tsx(24,22)
      TS2345: Argument of type 'BaseItemDto[] | null | undefined' is not assignable to parameter of type 'SetStateAction<BaseItemDto[]>'.
  Type 'undefined' is not assignable to type 'SetStateAction<BaseItemDto[]>'.
```

src/routes/movies.tsx
```diff
@@ -0,0 +1,139 @@
+import '../elements/emby-scroller/emby-scroller';
```
thornbill, line 1: This should probably be moved to `src/routes/movies/index.tsx` since the directory exists now.

thornbill (changes requested):

src/routes/movies.tsx
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/3792/commits/0914f9b72fcb82d488f62aed3650d7427c0136fe

thornbill (approved):

thornbill:
You can click the headings of the genres.

PR #3799, opened 2022-08-15
Update NPM lockfile to version 2

nielsvanvelzen:
Current Node LTS ships with an NPM version that uses the v2 lockfile format. It's currently extremely annoying to work on jf-web without NPM trying to upgrade the file. I don't see a reason to keep using the old format so this change updates it.

**Changes**
- Update NPM lockfile to version 2

**Issues**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,20825 @@
 {
   "name": "jellyfin-web",
   "version": "10.8.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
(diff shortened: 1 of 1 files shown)


nielsvanvelzen:
All CI on this PR worked fine though

thornbill:
Ah the new format is backwards compatible with npm 6! I presume that is why it is working... I had missed that in the release announcement. :rocket:

thornbill (approved):

PR #3800, opened 2022-08-15
Remove NPM prepare script

nielsvanvelzen:
The prepare script is really slow and annoying. It runs each time you install dependencies (which I always do when opening a project to make sure they are up-to-date). I think we needed it for the build & release process but with 10.9 getting the rewritten CI we should remove it.

**Changes**
- Remove NPM prepare script

**Issues**


thornbill:
This was used by the apps that include a bundled version of web (originally the Android app). If you install an npm dependency via a git url the build script for the dependency is not run... which makes git urls basically unusable without this addition. The last time I checked, the Tizen app still relied on it.

nielsvanvelzen:
https://cs.github.com/?scope=org%3Ajellyfin&scopeName=jellyfin&q=jellyfin-web+language%3AJSON

There are no package.json files which have jellyfin-web as dependency.

thornbill:
Another place we need to check is the build scripts to ensure they all explicitly call the build script instead of relying on the current behavior of build running on install... and we should clean up any usages of the `SKIP_PREPARE` variable.

crobibero (approved):

joshuaboniface (approved):

DumberSoft:
I have a question and how can I convert it to .tmg

PR #3856, opened 2022-08-25
Make book table of contents prettier

rourien:
**Changes**
Changes the way the table of contents looks. The two biggest parts are hiding the bullets in the list, making each link black and hiding the link underline. This makes them look more like buttons than web links. The `#dialogToc` changes resize the table of contents based on the length of the list and the length of the longest chapter name rather than have the table of contents take up most of the screen for a short and small list. The link text size was also increased from 92% to 120%.

### DIFF
--- a/src/plugins/bookPlayer/style.scss
+++ b/src/plugins/bookPlayer/style.scss
@@ -42,6 +42,12 @@
 
 #dialogToc {
     background-color: white;
+    height: fit-content;
+    width: fit-content;
+    max-height: 80%;
+    max-width: 60%;
+    padding-right: 50px;
+    padding-bottom: 15px;
 
     .bookplayerButtonIcon {
         color: black;
@@ -49,5 +55,18 @@
 
     .toc li {
         margin-bottom: 5px;
+
+        list-style-type: none;
+        font-size: 120%;
+
+        a:link {
+            color: #black;
+            text-decoration: none;
+
+        a:active,
+        a:hover {
+            color: #00a4dc;
+            text-decoration: none;
+        }
     }
 }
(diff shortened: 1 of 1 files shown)


src/plugins/bookPlayer/style.scss
```diff
@@ -42,12 +42,31 @@
...
 
     .toc li {
         margin-bottom: 5px;
+
+        list-style-type: none;
+        font-size: 120%;
+
+        a:link {
+            color: #black;
+            text-decoration: none;
```
thornbill, line 64: Looks like you missed a closing brace after this line.

thornbill: Did you push a fix for this? It still seems to be missing...

thornbill:
```suggestion
            text-decoration: none;
        }
```

src/plugins/bookPlayer/style.scss
```diff
@@ -42,12 +42,32 @@
...
     }
 
     .toc li {
         margin-bottom: 5px;
+
+        list-style-type: none;
+        font-size: 120%;
+
+        a:link {
+            color: #black;
```
thornbill, line 63:
```suggestion
            color: #000000;
```

src/plugins/bookPlayer/style.scss
```diff
@@ -42,12 +42,32 @@
...
     }
 
     .toc li {
         margin-bottom: 5px;
+
+        list-style-type: none;
+        font-size: 120%;
+
+        a:link {
+            color: #000000;
```
thornbill, line 63:
```suggestion
            color: #000;
```

thornbill (approved):

PR #3859, opened 2022-08-27
Set mp2 unsupported for android devices

lomion0815:
Some users encounter mp2 audio not decoded on android devices (see https://github.com/jellyfin/jellyfin-android/issues/764).

**Changes**
This fix sets mp2 support on android devices to unsupported and therefor forces transcoding of audio in such cases (usually DVB recordings).

**Issues**
- https://github.com/jellyfin/jellyfin-android/issues/764

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -390,6 +390,9 @@ import browser from './browser';
             if (supportsMp3VideoAudio && (browser.chrome || browser.edgeChromium || (browser.firefox && browser.versionMajor >= 83))) {
                 supportsMp2VideoAudio = true;
             }
+            if (browser.android) {
+                supportsMp2VideoAudio = false;
+            }
         }
 
         /* eslint-disable compat/compat */


dmitrylyzo:
I believe this can be considered as a bugfix.
Could you rebase the branch to the `release-10.8.x` and change the PR target (_no need to open a new one_)?

dmitrylyzo:
Oops, my bad :sweat:  - I did not consider that the MP2 support assumption was not added in 10.8 (_it was merged in the master right after the first beta_). Then, the previous state was right.

lomion0815:
> Oops, my bad 😓 - I did not consider that the MP2 support assumption was not added in 10.8 (_it was merged in the master right after the first beta_). Then, the previous state was right.

In 10.8 there was already a "simple" MP2 support assumption which caused the playback to fail. 
`const supportsMp2VideoAudio = browser.edgeUwp || browser.tizen || browser.web0s;`
I suggest to use this PR for the bugfix in 10.8.

In addition shall I provide another PR for master based on the commit before the rebase?

dmitrylyzo:
> In 10.8 there was already a "simple" MP2 support assumption which caused the playback to fail.
> `const supportsMp2VideoAudio = browser.edgeUwp || browser.tizen || browser.web0s;`

But this doesn't affect Android.

> In addition shall I provide another PR for master based on the commit before the rebase?

I think just revert it back. Sorry for the misdirection.

lomion0815:
Reverted last rebase ;-)

PR #3860, opened 2022-08-27
[BUG] Fixed: itemContextMenu manipulates incorrect item (e.g. song) if there is no image metadata

Callum17:
Desktop web client (Firefox, Chrome)
Jellyfin Version: 10.9.0

itemContextMenu for the nowPlayingBar media bar manipulates incorrect item (e.g. song) if there is no image metadata due to the relevant EventListener not being replaced on transition to the next item played.
Can manifest as the wrong song being added to a playlist, deleted, or receiving user supplied metadata, etc.

**Changes**

Removed `isRefreshing` - no longer longer seems relevant. Rename and small tidy of variable used to track currently played item's metadata image that interacted with `isRefreshing`.

Might be worth considering a future tidy up where all changes occur on update to `state.NowPlayingItem.Id`, rather than tracking and individually handling pieces of UI information (e.g. the image url). Considered such a refactor out of scope for fixing this bug.

jellyfin_bug

**Issues**
N/A


src/components/nowPlayingBar/nowPlayingBar.js
```diff
@@ -524,54 +524,46 @@ import { appRouter } from '../appRouter';
...
-            if (url) {
-                imageLoader.lazyImage(nowPlayingImageElement, url);
-                nowPlayingImageElement.style.display = null;
-                nowPlayingTextElement.style.marginLeft = null;
-            } else {
-                nowPlayingImageElement.style.backgroundImage = '';
-                nowPlayingImageElement.style.display = 'none';
-                nowPlayingTextElement.style.marginLeft = '1em';
-            }
+        if (url && url !== nowPlayingImageUrl) {
```
dmitrylyzo, line 527: With `url !== nowPlayingImageUrl`, poster is not updated when switching (in the queue) from a song without an image to a song with an image.
I think you can leave the previous logic without `isRefreshing`.

dmitrylyzo (commented):
> Might be worth considering a future tidy up where all changes occur on update to `state.NowPlayingItem.Id`, rather than tracking and individually handling pieces of UI information (e.g. the image url). Considered such a refactor out of scope for fixing this bug.

I haven't noticed any flaws in removing `isRefreshing` (e.g., recreating UI elements), but fixing the logic with `nowPlayingItem.Id !== nowPlayingItemId` would be safer, imo.

src/components/nowPlayingBar/nowPlayingBar.js
Callum17: >poster is not updated when switching (in the queue) from a song without an image to a song with an image.

Unable to reproduce from a clean dev deployment (assuming we are referencing the image in the bottom left, within the nowPlayingBar). `url` is [assigned further up](src/components/nowPlayingBar/nowPlayingBar.js), so I would not expect such behavior?

Logic in this PR should be functionally the same as before. `nowPlayingImageUrl` is simply a rename of `currentImgUrl`. I took the opportunity to flatten
```
if (url !== currentImgUrl) {
  currentImgUrl = url;
  isRefreshing = true;
  if (url) {
```

into
```
if (url && url !== nowPlayingImageUrl) {
  nowPlayingImageUrl = url;
```
for readability - the nested scope is no longer necessary.

Would appreciate instructions / images to assist with reproducing this issues if we still think it is relevant.

dmitrylyzo: Logic is not the same. You don't update `nowPlayingImage` when `url` is `null`.

I'm not sure if I was just switching to the next one or switching back and forth. This will probably show an error:
1. Start song without an image.
2. Add song with image to the queue and move it right after the first one.
3. Ensure the nowPlayingBar is visible (open home).
4. Click `Next item` button.
Song 2 starts, image updates.
6. Click `Previous item` button.
Song 1 starts, image removed.
7. Click `Next item` button again.
Song 2 starts, but no image because `url` is equal to the cached one.

Callum17: Looks like the issue was specifically related to navigating back to the same song with an image.
Latest commit should fix the aforementioned issue - `nowPlayingImageUrl ` gets updated to `null` when playing a song with no image.

Could do the assignment after the conditional - check on `url`, but I prefer to do it inside to make intent clear.

Shadowghost:
Does this only happen on current master or also 10.8? If the latter is the case this should be rebased onto the release branch.

src/components/nowPlayingBar/nowPlayingBar.js
dmitrylyzo: Currently it will execute the "null" block if the url has not been changed. This probably does no harm right now, simply because the update function is not called twice for the same item.

So I would leave the whole block as it was (just remove `isRefreshing` part) to avoid possible bug (with sequential call) and to minimize the count of changes. 🤷‍♂️

Callum17:
> Does this only happen on current master or also 10.8? If the latter is the case this should be rebased onto the release branch.

It happens in 10.8.5 too.
Rebased a new PR onto release-10.8.z #3877 

Feel free to close one of these PRs as deprecated - not that familiar with Jellyfin's branching strategy.

src/components/nowPlayingBar/nowPlayingBar.js
Callum17: Expanded conditional again so the DOM update only happens if `url !== nowPlayingImageUrl`.

Would rather avoid introducing `isRefreshing` again as it seems redundant both for readability and functionality.
I've aslso updated PR #3877

thornbill:
> Feel free to close one of these PRs as deprecated - not that familiar with Jellyfin's branching strategy.

This one won't be needed if the other is merged to the 10.8 branch. We cherry-pick changes that are merged there back to master.

thornbill:
Closing since this will be backported from the release branch PR.

Issue #3861, opened 2022-08-28
CBZ fit width issue


meet-k-pandya:
### Describe the bug

Steps:
1. Open CBZ file
2. Page is set to fit height causing page overlap, no option to zoom out as well.

Need options to change display behavior. Fit to width, Fit to height, zoom in, zoom out.
Or if it's too complex, I think fit to height instead of fit to width should solve most issues for CBZ files.

I tried looking in the code, but couldn't how this is being handled.

### Application version

2.4.4

### Where did you install the app from?

Google Play

### Device information

Asus ROG Phone 5

### Android version

Android 12

### Jellyfin server version

10.8.4

### Which video player implementations does this bug apply to?

- [ ] Web player (default)
- [ ] Integrated player (ExoPlayer)
- [ ] External player (VLC, mpv, MX Player)

MinecraftPlaye:
Maybe this might help with the "fit to width" thing: https://stackoverflow.com/a/66833775

meet-k-pandya:
This works, but requires additional handling to manipulate `object-postion` on scroll/touch/up-down arrow keys. Trying to avoid such additional event handling if possible.

Issue #3346, opened 2022-01-17
Comics / Mangas do not scale properly on change of the window size 


MinecraftPlaye:
**Describe The Bug**
ePub comics do not scale properly. Especially noticeable on large screens. Works fine with cbz files.
The scaling breaks even more when a comic / manga changes page size, e.g. one page has the resolution 100x500 and another page has the resolution 800x500.

**Steps To Reproduce**
1. Go to the Book library.
2. Click on a manga/comic. It has to be stored via ePub. I specifically tested ePub 3 files.
3. Change the window size.

**Screenshots**
![image](https://user-images.githubusercontent.com/9168045/149833758-90a37c2b-1fd8-414b-b59d-974c340ed2e2.png)
![image](https://user-images.githubusercontent.com/9168045/149833812-27cbc336-af1f-414d-aacc-023ee0ba0228.png)

**Expected Behavior**
I expect the comic / manga to scale with the window size without distorting  the comic / manga's aspect ratio.

**System (please complete the following information):**
First Test:

- Platform: Manjaro Linux
- Jellyfin Version: 10.7.7
- Jellyfin Media-Player: 1.6.1

Second Test:
Note here, it does not work perfectly, but the result is better than in the pictures above. By this, I mean that it looks like the comic / manga is just moved a bit up creating lot's of free space below the comic / manga instead of properly scaling it. I thus suspect that it looks better because I usually have my browser in maximized mode and not as a small window like the JMP.

- Platform: Manjaro Linux
- Browser: Firefox 96.0
- Jellyfin Version: 10.8.0 Alpha 5 / 10.7.7
- Jellyfin Web Version: 10.8.0 Alpha 5 / 10.7.7

MinecraftPlaye:
No, the issue is still present.

    Platform: Manjaro Linux
    Browser: Firefox 100.0
    Jellyfin Version: 10.8.0 Beta 3
    Jellyfin Web Version: 10.8.0 Beta 3

PR #3898, opened 2022-09-12
ComicsPlayer: Toggle buttons for page direction and double page view

meet-k-pandya:
**Changes**

1. Button to switch between `Left to Right` and `Right to Left` page direction.
2. Button to switch between `Single Page` and `Double Page` view.
3. Persistence of preferences for individual comic.
4. Fix rendering issues 
5. Fix styling/theme issues

**Issues**

Addresses https://features.jellyfin.org/posts/1158/richer-comic-reader
Almost Fixes https://github.com/jellyfin/jellyfin-web/issues/3861 (couldn't figure out Fit to width)
Fixes https://github.com/jellyfin/jellyfin-web/issues/3346

### DIFF
--- a/src/plugins/comicsPlayer/plugin.js
+++ b/src/plugins/comicsPlayer/plugin.js
@@ -5,9 +5,11 @@ import dialogHelper from '../../components/dialogHelper/dialogHelper';
 import keyboardnavigation from '../../scripts/keyboardNavigation';
 import { appRouter } from '../../components/appRouter';
 import ServerConnections from '../../components/ServerConnections';
-// eslint-disable-next-line import/named, import/namespace
-import { Swiper } from 'swiper/swiper-bundle.esm';
-import 'swiper/swiper-bundle.css';
+import * as userSettings from '../../scripts/settings/userSettings';
+//eslint-disable-next-line import/no-unresolved
+import { Swiper } from 'swiper/bundle';
+//eslint-disable-next-line import/no-unresolved
+import 'swiper/css/bundle';
 
 import './style.scss';
 
@@ -27,6 +29,9 @@ export class ComicsPlayer {
         this.currentPage = 0;
         this.pageCount = 0;
 
+        const mediaSourceId = options.items[0].Id;
+        this.comicsPlayerSettings = userSettings.getComicsPlayerSettings(mediaSourceId);
+
         const elem = this.createMediaElement();
         return this.setCurrentSrc(elem, options);
     }
@@ -39,6 +44,9 @@ export class ComicsPlayer {
         };
 
         Events.trigger(this, 'stopped', [stopInfo]);
+
+        const mediaSourceId = this.item.Id;
+        userSettings.setComicsPlayerSettings(this.comicsPlayerSettings, mediaSourceId);
 
         this.archiveSource?.release();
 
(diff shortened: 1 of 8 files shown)


MinecraftPlaye:
That looks really nice. :+1: 
Just tested it on an iPhone and in Firefox on my PC.

To add my 2 cents to this: While this looks like a great start, in the future it would be cool if the Jellyfin server had a value to store the reading direction for each book/comic. That way, the bookshelf plugin could finally save that information and the web-interface could use that information as a default value instead of defaulting always to `ltr`. ;)

src/plugins/comicsPlayer/plugin.js
```diff
@@ -96,11 +157,71 @@ export class ComicsPlayer {
         }
     }
 
+    //Natively available in v8.2.6 [https://github.com/nolimits4web/Swiper/issues/3279]
```
MinecraftPlaye, line 160: Related to #3892

meet-k-pandya: I did try to update the package, but the build was failing.

meet-k-pandya:
> That looks really nice. 👍 Just tested it on an iPhone and in Firefox on my PC.
> 
> To add my 2 cents to this: While this looks like a great start, in the future it would be cool if the Jellyfin server had a value to store the reading direction for each book/comic. That way, the bookshelf plugin could finally save that information and the web-interface could use that information as a default value instead of defaulting always to `ltr`. ;)

Yes, that would be ideal. I was thinking of adding the preferences to the UserData object as the next improvement.

src/plugins/comicsPlayer/plugin.js
meet-k-pandya: Successfully fixed the build, added the changes in this PR

src/plugins/comicsPlayer/plugin.js
```diff
@@ -21,12 +22,28 @@ export class ComicsPlayer {
 
         this.onDialogClosed = this.onDialogClosed.bind(this);
         this.onWindowKeyUp = this.onWindowKeyUp.bind(this);
+
+        const storedPrefs = localStorage.getItem('comicsPlayerPrefs');
+        if (!storedPrefs)
+            localStorage.setItem('comicsPlayerPrefs', '{}');
```
thornbill, line 28: I think this should probably use `userSettings` instead of dealing directly with `localStorage`.

https://github.com/jellyfin/jellyfin-web/blob/master/src/scripts/settings/userSettings.js

meet-k-pandya: Sure, I will try this out.

MinecraftPlaye (approved):
For the comment about using `userSettings` instead of dealing directly with `localStorage`, I´m just referring to @thornbill as nothing changed there between his and this review and I´m not familiar with the difference myself.

The rest of the code looks good to me and seems to work fine during a quick test with Firefox 104 on Fedora. Thought it would probably not hurt for a more experienced web-dev to also take a look at those changes. ;)

src/plugins/comicsPlayer/plugin.js
meet-k-pandya: Updated PR with the requested change, please review this again.

MinecraftPlaye (commented):
Not sure why, but with the latest force-push, the close-button doesn´t seem to work anymore on my system with Firefox 104. The server does get the message that the comic has been closed (`Playback stopped reported by app`), but the comic-player does not close properly. You can still scroll through the comic, but not press any button like close or change reading direction.

meet-k-pandya:
> Not sure why, but with the latest force-push, the close-button doesn´t seem to work anymore on my system with Firefox 104. The server does get the message that the comic has been closed (`Playback stopped reported by app`), but the comic-player does not close properly. You can still scroll through the comic, but not press any button like close or change reading direction.

I have not made any change on the close functionality, let me check what's the issue.

meet-k-pandya:
> Not sure why, but with the latest force-push, the close-button doesn´t seem to work anymore on my system with Firefox 104. The server does get the message that the comic has been closed (`Playback stopped reported by app`), but the comic-player does not close properly. You can still scroll through the comic, but not press any button like close or change reading direction.

Should work now, please check again

MinecraftPlaye (approved):

MinecraftPlaye:
> Should work now, please check again

Yes, it works again. :+1:

PR #3900, opened 2022-09-13
Fix failures to open secondary action sheet in html video player

is343:
**Changes**
- Set the main video menu action sheet to resolve on click.
- Added reject reason text to the close logic.
  - it took awhile to figure out where all of the uncaught promise errors were coming from with no message text being returned

**Issues**
I did not find any issues for this. I found this bug while I was working on [this feature request](https://features.jellyfin.org/posts/936/multi-language-subtitles).

I noticed the secondary menu popups would fail sometimes on chrome. But it happened every time on firefox. This is happening when the clicked action sheet item tries to open another action sheet. 

I'm not sure why the close logic for rejecting the promise does not go down the resolve path. We are definitely clicking a valid menu item in order to trigger the second actionsheet to render.
This might be some weird race condition going on, because I was able to also avoid it by wrapping the second `actionSheet.show` logic in a brief 100ms `setTimeout`. Maybe the second one renders before the first one closes, and ends up closing them both?

Either way, I was able to solve it by passing the `resolveOnClick` option to the main action sheet that has options to open secondary action sheets.

**Before**

*Firefox* - closes every time
Screen Recording 2022-09-12 at 3 08 28 PM

*Chrome* - closes sometimes - still gets reject error every time, even for button clicks
Screen Recording 2022-09-12 at 3 07 47 PM

**After**

*Firefox* - no longer closing - showing reject reason in console
Screen Recording 2022-09-12 at 3 07 47 PM


src/components/playback/playersettingsmenu.js
```diff
@@ -238,6 +238,7 @@ function showWithUser(options, player, user) {
 
     return actionsheet.show({
         items: menuItems,
+        resolveOnClick: true,
         positionTo: options.positionTo
     }).then(function (id) {
```
is343, line 243: Should I also add a `catch` to handle the promise reject? I am not sure what more it will do than `console.error` the message. It's not much different than what it is doing now, but we will not have an unhandled promise showing up in the console like it currently does.

thornbill: I wouldn't worry about it. Currently none of our popups really handle the rejection when clicking outside. I imagine fixing them all would come as part of a larger refactor.

thornbill:
This issue only exists on master/unstable builds correct? I only noticed it after the history changes from #3620  landed in master.

thornbill (approved):

is343:
That's correct. I'm not actually seeing this in my home server instance running a prod build.

> This issue only exists on master/unstable builds correct? I only noticed it after the history changes from #3620 landed in master.

That's correct. I'm not actually seeing this in my home server instance running a prod build.

PR #3939, opened 2022-09-22
Fix PDF reader dark theme

### LINKED ISSUE #3929: PDF reader does not use dark theme
**Describe The Bug**
https://github.com/jellyfin/jellyfin-web/pull/3855 changed the book readers to use a dark theme by default, but the PDF reader/player was not updated. This makes the controls (specifically the close button) not visible.

**Steps To Reproduce**
1. Open the PDF player within Jellyfin

**Expected Behavior**
Controls (close button) should be visible.

**Logs**
N/A

**Screenshots**
Screenshot 2022-09-21 at 14-11-34 Jellyfin

 - Platform: Any
 - Browser: Any
 - Jellyfin Version: Unstable

**Additional Context**

kristiankauffeld:
**Changes**
Change PDF reader background color, so close button is visible.

**Issues**
Fixes #3929

### DIFF
--- a/src/plugins/pdfPlayer/style.scss
+++ b/src/plugins/pdfPlayer/style.scss
@@ -4,7 +4,7 @@
     width: 100%;
     overflow: none;
     z-index: 100;
-    background: #fff;
+    background: #000000;
 }
 
 #canvas {
--- a/src/plugins/pdfPlayer/style.scss
+++ b/src/plugins/pdfPlayer/style.scss
@@ -4,7 +4,7 @@
     width: 100%;
     overflow: none;
     z-index: 100;
-    background: #000000;
+    background: #000;
 }
 
 #canvas {


src/plugins/pdfPlayer/style.scss
```diff
@@ -4,7 +4,7 @@
     width: 100%;
     overflow: none;
     z-index: 100;
-    background: #fff;
+    background: #000000;
```
thornbill, line 7:
```suggestion
    background: #000;
```

Our stylelint rules prefer the short color format when possible.

thornbill (changes requested):

src/plugins/pdfPlayer/style.scss
kristiankauffeld: Okay got it! I've updated the PR

thornbill (approved):

PR #3942, opened 2022-09-23
Add epubjs theme support to the book player

### LINKED ISSUE #3930: EPUBs are not readable with new dark mode
**Describe The Bug**
EPUBs seem to render with black text on a dark background by default since https://github.com/jellyfin/jellyfin-web/pull/3855 was merged.

**Steps To Reproduce**

1. Open the EPUB player within Jellyfin

**Expected Behavior**
Book text should be legible. We should either:

1. Use epub.js theme support to render in dark mode correctly.
2. Force a light background for the EPUB reader.

Preferably we should use epub.js themes in my opinion. I have not found much documentation for that though.

**Logs**
N/A

**Screenshots**
Screenshot 2022-09-21 at 14-10-57 Jellyfin

 - Platform: Any
 - Browser: Any
 - Jellyfin Version: Unstable

**Additional Context**

MinecraftPlaye:
**Changes**
Add a dark theme to the book player via epubjs. For everything but the text, the dark theme from Jellyfin is retained. 

Should any other theme than the dark theme be chosen by the user, epubjs falls back to it's default theme (the one Jellyfin used before the book player got a dark theme).

**Note:** This change assumes that the dark theme is the default theme of Jellyfin as `userSettings.theme()` returns `null` should the user use the default theme.

**Issues**
Fixes #3930

### DIFF
--- a/src/plugins/bookPlayer/plugin.js
+++ b/src/plugins/bookPlayer/plugin.js
@@ -9,6 +9,7 @@ import Screenfull from 'screenfull';
 import TableOfContents from './tableOfContents';
 import dom from '../../scripts/dom';
 import { translateHtml } from '../../scripts/globalize';
+import * as userSettings from '../../scripts/settings/userSettings';
 
 import '../../elements/emby-button/paper-icon-button-light';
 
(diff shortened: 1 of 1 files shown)


thornbill (approved):
:+1: looks fine! I did find some other issues with the book player styles, but they aren't related to this so I'll open a separate PR.

PR #3944, opened 2022-09-25
Metadata loss warning while renaming libraries

Artiume:
**Changes**

Add a warning on metadata loss while renaming libraries. Could've made a confirm window, went with using the label for less code.

**Issues**

https://www.reddit.com/r/jellyfin/comments/pk8evo/comment/ipqyk41/


src/controllers/dashboard/library.js
```diff
@@ -76,7 +76,8 @@ import cardBuilder from '../../components/cardbuilder/cardBuilder';
     function renameVirtualFolder(page, virtualFolder) {
         import('../../components/prompt/prompt').then(({default: prompt}) => {
             prompt({
-                label: globalize.translate('LabelNewName'),
+                title: globalize.translate('LabelNewName'),
+                label: globalize.translate('MessageAreYouSureYouWishToRenameMediaFolder'),
```
thornbill, line 80:
```suggestion
                label: globalize.translate('LabelNewName'),
                description: globalize.translate('MessageAreYouSureYouWishToRenameMediaFolder'),
```

thornbill (changes requested):
I think this will work better. Title is the heading displayed next to the back arrow and label is the input field label.

Screenshot 2022-09-29 at 16-25-39 Jellyfin

src/controllers/dashboard/library.js
nielsvanvelzen: When I look at your screenshot, I don't think it should be a question but a note/warning instead. Something like "Renaming a folder clears all all metadata, proceed with caution".

src/strings/en-us.json
```diff
@@ -1036,6 +1036,7 @@
     "MessageAlreadyInstalled": "This version is already installed.",
     "MessageAreYouSureDeleteSubtitles": "Are you sure you wish to delete this subtitle file?",
     "MessageAreYouSureYouWishToRemoveMediaFolder": "Are you sure you wish to remove this media folder?",
+    "MessageAreYouSureYouWishToRenameMediaFolder": "Are you sure you wish to rename this media folder? All metadata will be lost",
```
thornbill, line 1039:
```suggestion
    "MessageAreYouSureYouWishToRenameMediaFolder": "Renaming a media library will cause all metadata to be lost, proceed with caution.",
```

crobibero (approved):

nielsvanvelzen (approved):

PR #3966, opened 2022-09-29
Add mini css plugin for production builds

thornbill:
**Changes**
Adds the mini css plugin to the webpack config for production builds

**Issues**
N/A


webpack.common.js
```diff
@@ -92,8 +97,10 @@ module.exports = {
             chunks: 'all',
             maxInitialRequests: Infinity,
             cacheGroups: {
-                vendor: {
-                    test: /[\\/]node_modules[\\/]/,
+                node_modules: {
+                    test(module) {
+                        return NODE_MODULES_REGEX.test(module.context);
```
thornbill, line 102: The previous test caused a crash when using the mini css plugin. This test should be basically the same, but doesn't crash.

nielsvanvelzen (approved):

PR #3972, opened 2022-09-30
Upgrade to node 16

thornbill:
**Changes**
Upgrade to node 16

**Issues**
N/A

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,3 @@
+engine-strict=true
 fund=false
 save-exact=true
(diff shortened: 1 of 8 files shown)


.npmrc
```diff
@@ -1,2 +1,3 @@
+engine-strict=true
```
thornbill, line 1: I think I found them all, but this should help track down any I missed. It can be removed once a full build goes through.

joshuaboniface (approved):

PR #3976, opened 2022-10-01
 Move SyncPlay code to a plugin

thornbill:
**Changes**
* Moves SyncPlay code to a plugin
* Removes the custom initialization code for SyncPlay
* Hides the SyncPlay button if the plugin is not loaded
* Removes a check for old server versions

In the future we should try to remove references to the plugin code that are outside of the plugin directory and move any playback logic to the playbackManager.

**Issues**
N/A


nielsvanvelzen:
@iwalton3 I'm not sure how JMP implements SyncPlay but it might need updating when this PR is merged

thornbill:
In theory this one should not affect JMP since it really only changes how the SyncPlay code is loaded... hopefully. 😅

Issue #3494, opened 2022-03-10
Add eslint rules to catch issues flagged by SonarQube


thornbill:
Below is a breakdown of eslint rules that we should add to prevent things that SonarQube considers code smells by their priority in SonarQube.

These rules should be applied and fixed one per PR.

### Critical
* [x] [default-case-last](https://eslint.org/docs/rules/default-case-last) &mdash; #3493 
* [x] [no-void](https://eslint.org/docs/rules/no-void) &mdash; #3509 

### Major
* [x] [no-redeclare](https://eslint.org/docs/rules/no-redeclare) (already included by `eslint:recommended`)
* [x] [no-sequences](https://eslint.org/docs/rules/no-sequences) &mdash; #3609
* [x] [no-shadow](https://eslint.org/docs/rules/no-shadow) &mdash; #4058
* [x] [no-nested-ternary](https://eslint.org/docs/rules/no-nested-ternary) &mdash; #3603
* [ ] [max-params](https://eslint.org/docs/rules/max-params) with a max of 7 &mdash; #4035
* [x] [no-throw-literal](https://eslint.org/docs/rules/no-throw-literal) &mdash; #3798

### Minor
* [x] [no-return-await](https://eslint.org/docs/rules/no-return-await) &mdash; #3796
* [x] [no-unused-vars](https://eslint.org/docs/rules/no-unused-vars)
* [x] [no-duplicate-imports](https://eslint.org/docs/rules/no-duplicate-imports) &mdash; #3539

sancodes:
I would like to help, if this is still open.

thornbill:
Go for it. I only did the first one from the list. 🙂

sancodes:
@thornbill I've added all the rules. Please let me know if I've missed anything. Thank you

sancodes:
@thornbill I have added lint for `no-void`. Please help me review it. If any errors, I'm willing to fix it.

sancodes:
@thornbill thank you for approving changes, if other issues are still open, I can try to tackle it. Let me know.

thornbill:
@sancodes they are still open. I have been updating the list at the top of the issue, so it should be accurate. :slightly_smiling_face:

sancodes:
@thornbill I will start with `major` section. I will open new PR through new branch.

ryan-hartzell:
@thornbill I wanted to help out on these. I added a PR for `no-nested-ternary`. I'm not sure of the logic on all the fixes.

[...]

PR #3980, opened 2022-10-03
Fix sonar issues

thornbill:
**Changes**
Fixes some of the sonarjs issues with few instances and enables the corresponding eslint rules

**Issues**
Related to https://github.com/jellyfin/jellyfin-web/issues/3494


src/components/filtermenu/filtermenu.js
```diff
@@ -124,14 +124,15 @@ function saveValues(context, settings, settingsKey) {
...
-    }
+    // TODO: Add filtering by series status?
+    // const seriesStatuses = [];
+    // elems = context.querySelectorAll('.chkSeriesStatus');
+
+    // for (let i = 0, length = elems.length; i < length; i++) {
+    //     if (elems[i].checked) {
+    //         seriesStatuses.push(elems[i].getAttribute('data-filter'));
+    //     }
+    // }
```
nielsvanvelzen, line 135: Maybe just remove the code instead of leaving commented code?

nielsvanvelzen (approved):

src/components/filtermenu/filtermenu.js
thornbill: I commented it out because I suspected it would work with a minor fix... but I was being lazy and didn't want to test that at the time. Now I did and it works, so I just committed the fix here. :sweat_smile:

nielsvanvelzen (approved):

PR #3982, opened 2022-10-03
Chrome now supports HEVC decoding

knackebrot:
Rolled out for some users in 105 and set to roll out in 107 for everyone. Works only if hardware acceleration is available.

This PR enables DirectPlay. HLS seems to get transcoded to H264 no matter the transcode reason (AudioCodecNotSupported, ContainerNotSupported etc.) Can't confirm whether hls.js actually supports demuxing HEVC from TS. FMP4 works fine, but we still have the seeking issue.

https://chromiumdash.appspot.com/commit/f46f3399d7ff1119f1f1b9026d31110d7e5b9d8a
https://github.com/StaZhu/enable-chromium-hevc-hardware-decoding


nyanmisaka:
FYI hls.js never support HEVC in TS container and it’s buggy when handling fMP4 seeking. We have to use shaka-player and fMP4 instead if HEVC is needed.

thornbill:
Will this need updated based on @nyanmisaka's comment?

nyanmisaka:
> Will this need updated based on @nyanmisaka's comment?

#4041 should supersede this.

PR #3999, opened 2022-10-05
Fix sonarjs no-duplicated-branches

thornbill:
**Changes**
Fixes some more "major" issues in sonar. I ignored the cardbuilder because that file is just a complete mess... there is already a comment in that particular function with a refactor TODO.

**Issues**
Related to #3494

### DIFF
--- a/src/elements/emby-ratingbutton/emby-ratingbutton.js
+++ b/src/elements/emby-ratingbutton/emby-ratingbutton.js
@@ -63,18 +63,6 @@ import ServerConnections from '../../components/ServerConnections';
             }
 
             button.classList.add('ratingbutton-withrating');
-        } else if (likes) {
-            if (icon) {
-                icon.classList.add('favorite');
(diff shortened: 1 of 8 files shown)


src/elements/emby-ratingbutton/emby-ratingbutton.js
```diff
@@ -63,18 +63,6 @@ import ServerConnections from '../../components/ServerConnections';
...
-                icon.classList.add('favorite');
-                icon.classList.remove('ratingbutton-icon-withrating');
-            }
-            button.classList.remove('ratingbutton-withrating');
-        } else if (likes === false) {
-            if (icon) {
-                icon.classList.add('favorite');
-                icon.classList.remove('ratingbutton-icon-withrating');
-            }
-            button.classList.remove('ratingbutton-withrating');
```
thornbill, line 77: These two were identical to the `else` block below.

nielsvanvelzen (approved):

PR #4001, opened 2022-10-06
Fix sonarjs prefer-single-boolean-return

thornbill:
**Changes**
Fixes the most frequent "minor" issue in sonar

**Issues**
Related to #3494

### DIFF
--- a/src/components/itemHelper.js
+++ b/src/components/itemHelper.js
@@ -144,7 +144,7 @@ export function canEditImages (user, item) {
     }
 
     if (itemType === 'UserView') {
-        return user.Policy.IsAdministrator;
+        return !!user.Policy.IsAdministrator;
     }
 
     if (item.Type === 'Recording' && item.Status !== 'Completed') {
@@ -218,13 +218,13 @@ export function canMarkPlayed (item) {
 }
 
 export function canRate (item) {
-    return item.Type === 'Program'
-        || item.Type === 'Timer'
-        || item.Type === 'SeriesTimer'
-        || item.Type === 'CollectionFolder'
-        || item.Type === 'UserView'
-        || item.Type === 'Channel'
-        || !item.UserData;
+    return item.Type !== 'Program'
+        && item.Type !== 'Timer'
+        && item.Type !== 'SeriesTimer'
+        && item.Type !== 'CollectionFolder'
+        && item.Type !== 'UserView'
+        && item.Type !== 'Channel'
+        && item.UserData;
 }
 
 export function canConvert (item, user) {
@@ -255,7 +255,7 @@ export function canConvert (item, user) {
         return false;
     }
 
-    return item.IsPlaceHolder;
+    return !item.IsPlaceHolder;
 }
 
 export function canRefreshMetadata (item, user) {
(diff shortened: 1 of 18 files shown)


src/components/itemHelper.js
```diff
@@ -148,11 +144,7 @@ export function canEditImages (user, item) {
     }
 
     if (itemType === 'UserView') {
-        if (user.Policy.IsAdministrator) {
-            return true;
-        }
-
-        return false;
+        return user.Policy.IsAdministrator;
```
dmitrylyzo, line 147: Maybe cast to boolean in case of missing `IsAdministrator`?
```suggestion
        return !!user.Policy.IsAdministrator;
```

src/components/itemHelper.js
```diff
@@ -218,29 +210,21 @@ export function canMarkPlayed (item) {
...
         || item.Type === 'SeriesTimer'
         || item.Type === 'CollectionFolder'
         || item.Type === 'UserView'
         || item.Type === 'Channel'
-        || !item.UserData) {
-        return false;
-    }
-
-    return true;
+        || !item.UserData;
```
dmitrylyzo, line 227: Logic is inverted.

src/components/itemHelper.js
```diff
@@ -271,11 +255,7 @@ export function canConvert (item, user) {
         return false;
     }
 
-    if (item.IsPlaceHolder) {
-        return false;
-    }
-
-    return true;
+    return item.IsPlaceHolder;
```
dmitrylyzo, line 258:
```suggestion
    return !item.IsPlaceHolder;
```

src/components/itemHelper.js
```diff
@@ -307,11 +287,8 @@ export function supportsMediaSourceSelection (item) {
     if (item.EnableMediaSourceDisplay === false) {
         return false;
     }
-    if (item.EnableMediaSourceDisplay == null && item.SourceType && item.SourceType !== 'Library') {
-        return false;
-    }
 
-    return true;
+    return item.EnableMediaSourceDisplay == null && item.SourceType && item.SourceType !== 'Library';
```
dmitrylyzo, line 291:
```suggestion
    return !(item.EnableMediaSourceDisplay == null && item.SourceType && item.SourceType !== 'Library');
```

src/components/itemHelper.js
```diff
@@ -304,14 +284,11 @@ export function supportsMediaSourceSelection (item) {
...
-    if (item.EnableMediaSourceDisplay === false) {
-        return false;
-    }
-    if (item.EnableMediaSourceDisplay == null && item.SourceType && item.SourceType !== 'Library') {
+    if (!item.EnableMediaSourceDisplay) {
         return false;
     }
 
-    return true;
+    return !item.SourceType || item.SourceType === 'Library';
```
dmitrylyzo, line 291: Looks like it was equal to:
```js
if (item.EnableMediaSourceDisplay != null) {
    return !!item.EnableMediaSourceDisplay;
}

return !item.SourceType || item.SourceType === 'Library';
```
So, you were able to force "MediaSourceSelection" support.

Not sure which is the right one.

thornbill: There was also a check for `EnableMediaSourceDisplay` being false above though.

dmitrylyzo: Yes, the equivalent includes it. If `EnableMediaSourceDisplay` explicitly `false`, return `false`. If it is undefined or null, check `item.SourceType`. Otherwise (`EnableMediaSourceDisplay` is defined and is not `false` and `item.SourceType` is undefined or is a library), return `true`.
_It probably didn't work correctly for `EnableMediaSourceDisplay: 0` (!== false)._

dmitrylyzo (approved):

PR #3676, opened 2022-05-29
Respect safe area (notch) of mobile devices

taku0:
**Changes**

This adds paddings to various containers so that buttons and contents are not hidden by the notch or the Home indicator (the gray bar at the bottom).

https://webkit.org/blog/7929/designing-websites-for-iphone-x/

- `src/assets/css/librarybrowser.scss`
  - `.skinHeader`: menu bar at the top.
  - `.padded-left` and `.padded-right`: the heading of library.

- `src/elements/emby-scroller/emby-scroller.scss`
  - `.emby-scroller`: the library items.

- `src/assets/css/videoosd.scss`
  - `.videoOsdBottom`: the video playback control buttons and the progress bar.

- `src/components/appFooter/appFooter.scss`
  - `.appfooter`: the bottom bar showing the currently playing media.

- `src/components/multiSelect/multiSelect.scss`
  - `.selectionCommandsPanel`: the top bar when selecting multiple media.
    Shown when long-pressing media.

- `src/components/remotecontrol/remotecontrol.scss`
  - `.playlistSectionButton`: the bottom bar for current playlist.
    Shown when touching `.appfooter`.

- `src/components/toast/toast.scss`
  - `.toastContainer`: the toast.  Shown when refreshing metadata.

- `src/index.html`
  - `.mainDrawerHandle`: gesture area for the drawer.
    It seems conflicting with gestures of Safari.

- `src/libraries/navdrawer/navdrawer.scss`
  - `.drawer-open`: the drawer.

- `src/plugins/htmlVideoPlayer/style.scss`
  - `.videoPlayerContainer`: the video area when not fullscreen.
  - `.videoSubtitles`: the video subtitles for some devices, not including iPhone.

- `src/plugins/youtubePlayer/style.scss`
  - `.youtubePlayerContainer` container for the YouTube plugin.  Not tested.

**Issues**

**How to Test**

1. Open Jellyfin with iPhone X or later in landscape orientation.
2. Browse a library and check the buttons and contents are not hidden by the notch.
3. Open the drawer menu and check the menu items are not hidden by the notch.
4. Play a video and check the controls are not hidden by the notch.
5. Play a music and check the controls are not hidden by the notch or the Home indicator.
6. Tap the bottom bar while playing the music to show playlist and check the controls are not hidden by the notch or the Home indicator.
7. Long press a library item and check the top bar is not hidden by the notch.
8. Refresh metadata of a library and check the toast is not hidden by the notch.

I couldn't test with Android because I don't have one with a notch or hole.

PR #4009, opened 2022-10-07
Fix legacy browsers

dmitrylyzo:
**Changes**
- Add `@remix-run/router` to `babel-loader`.
- Turn off `mergeLonghand` (cssnano) because it combines `padding-*`, breaking fallback styles (https://github.com/cssnano/cssnano/issues/1192, https://github.com/cssnano/cssnano/issues/1163).

**Issues**
- webOS 1.2 cannot load Jeff with `const`/`let`.
- webOS 1.2 lost some padding (video OSD) after #3676.

### DIFF
--- a/cssnano.config.js
+++ b/cssnano.config.js
@@ -1,0 +1,10 @@
+module.exports = {
+    preset: [
+        'default',
+        // Turn off `mergeLonghand` because it combines `padding-*` and `margin-*`,
+        // breaking fallback styles.
+        // https://github.com/cssnano/cssnano/issues/1163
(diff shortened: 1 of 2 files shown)


thornbill (approved):

cssnano.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+    preset: [
+        'default',
+        { mergeLonghand: false }
```
thornbill, line 4: Could we add a comment here with a link to the bug for future reference?

thornbill (approved):

PR #4015, opened 2022-10-08
Trim username on login

FarisR99:
**Changes**
Trim the username before attempting to authenticate.

### DIFF
--- a/src/controllers/session/login/index.js
+++ b/src/controllers/session/login/index.js
@@ -231,7 +231,9 @@ import './login.scss';
         view.querySelector('.manualLoginForm').addEventListener('submit', function (e) {
             appSettings.enableAutoLogin(view.querySelector('.chkRememberLogin').checked);
             const apiClient = getApiClient();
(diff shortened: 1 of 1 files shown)


nielsvanvelzen:
I don't think this is a good change. The server technically supports usernames starting/ending with whitespace, this change would block the ability to sign into those accounts.

FarisR99:
@nielsvanvelzen That's fair, I had this consideration in the back of my mind but I didn't confirm whether you *can* have whitespaces + assumed people wouldn't do something as weird as this. I do believe that in the first place, whitespaces before and after names should never be supported, but it is too late now.

The main problem with whitespaces I ran into was that autocomplete on Android would add a space after the completion so you have to backspace on this app, whereas on a typical app it will trim it.

FarisR99:
An alternative solution would be for the backend to do a "best match" first taking into account whitespaces, but worst case matching stored usernames with spaces before/after to the input with no spaces. This is hacky though.

thornbill:
Yeah I think this would need to be handled on the backend. We should probably be trimming usernames when they are created, but for backwards compatibility we would need to check both trimmed and untrimmed.

PR #4016, opened 2022-10-09
Add DownloadAll for season and series

Ongy:
**Changes**

Add a Download All button to Series and Season item types.

**Issues**

This is mostly for the Android app, which uses jellyfin-web.

**To Discuss**
This code still has minor issues:
* The `Download All` string isn't translated
* The functions in the handler should probably be moved somehwere else? Where to?

* On both Firefox and Chromium the DownloadAll button only downloads *some*. It's been 7 or so on both for me. This is due to a parallel download limit in the browsers. Should this code pull in a check whether it's running native before it shows the button?

### DIFF
--- a/src/components/itemContextMenu.js
+++ b/src/components/itemContextMenu.js
@@ -324,6 +324,7 @@ import toast from './toast/toast';
         const apiClient = ServerConnections.getApiClient(serverId);
 
         return new Promise(function (resolve, reject) {
+            // eslint-disable-next-line sonarjs/max-switch-cases
             switch (id) {
                 case 'addtocollection':
                     import('./collectionEditor/collectionEditor').then(({default: CollectionEditor}) => {
@@ -380,8 +381,8 @@ import toast from './toast/toast';
                                     Fields: 'CanDownload,Path'
                                 });
                             }
-                        )).then(seasons => {
-                            downloadEpisodes([].concat.apply([], seasons.map(season => season.Items)));
+                        )).then(seasonData => {
+                            downloadEpisodes([].concat.apply([], seasonData.map(season => season.Items)));
                         });
                     };
 
--- a/src/components/itemContextMenu.js
+++ b/src/components/itemContextMenu.js
@@ -141,6 +141,14 @@ import toast from './toast/toast';
             });
         }
 
+        if (item.Type === 'Season' || item.Type == 'Series') {
+            commands.push({
+                name: 'Download All',
+                id: 'downloadall',
+                icon: 'file_download'
+            });
+        }
+
         if (item.CanDelete && options.deleteItem !== false) {
             if (item.Type === 'Playlist' || item.Type === 'BoxSet') {
                 commands.push({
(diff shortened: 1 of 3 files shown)


nielsvanvelzen:
I think it would be better to add a new downloadBatch function to the native shell to send all items to download at once. That way the implementation in the app can more optimally create the notification(s) and download processes.

Ongy:
That's already what happens. The switch towards native shell is in https://github.com/jellyfin/jellyfin-web/blob/master/src/scripts/fileDownloader.js

And the code here collects all the downloads to send them off to download at once.

nielsvanvelzen:
https://github.com/jellyfin/jellyfin-web/blob/55117972be39b600136c78f4861d743c94da7747/src/scripts/shell.js#L44-L50

It later splits them into multiple downloadFile calls.

Ongy:
I just assumed that that's a Node function. Is native shell on the Android implementation of Jellyfin?

Without type annotations I can't find my way around the codebase, as vscode go to definition doesn't work =.=

Ongy:
Ok, so it ends up in https://github.com/jellyfin/jellyfin-android/blob/b4a4d359d04112e460167b42e620978595265e46/app/src/main/java/org/jellyfin/mobile/utils/SystemUtils.kt#L61

The bulk download could help. But from looking at other discussions the App and web interface shouldn't depend too much on each other.
So the code should check for whether there's a bulk download and fall back to this if there isn't I think.

nielsvanvelzen:
> The bulk download could help. But from looking at other discussions the App and web interface shouldn't depend too much on each other.
> So the code should check for whether there's a bulk download and fall back to this if there isn't I think.

Yeah exactly, check for something like `window.NativeShell.bulkDownload` to exist and if it does, use that, otherwise call downloadFile for each item. We can then take our time to implement it in the Android app (and possibly others)

Ongy:
Is there a description for how to add a new translated word somewhere?

I saw the long json files, but I assume that there's somethingmore involved than just adding a line into them?
After that there's only the lint complaint about going over 30 cases in switch-case open. Is there a prefered way to fix that?

thornbill:
For the translations, just add them to the en-us json file. We use a weblate instance for the community to be able to contribute to translations. It will pick up any new keys there, but editing the other language files can cause conflicts.

If there isn't a good way to reduce the number of cases in this switch statement (there probably isn't), we can just add comments so eslint ignores that error in this section of code.

Ongy:
Hey,

from my point this is done iirc.
The main missing point is the annotation. Can you point me to how to run the lint local, or advice how to set it?

Thanks

thornbill:
The branch was a bit outdated so some of the lint rules that were recently added to the `master` branch were not present.

You can disable the max switch case rule for this particular change by adding `// eslint-disable-next-line sonarjs/max-switch-cases` before this line:
https://github.com/jellyfin/jellyfin-web/blob/44a5d3c57027346608bd79f6233e643b62f25e3c/src/components/itemContextMenu.js#L327

The other issue is that the `seasons` variable needs renamed here since a variable with the same name exists in the outer scope:
https://github.com/jellyfin/jellyfin-web/blob/44a5d3c57027346608bd79f6233e643b62f25e3c/src/components/itemContextMenu.js#L382

src/components/itemContextMenu.js
```diff
@@ -346,6 +355,48 @@ import toast from './toast/toast';
...
+                    const downloadSeasons = seasons => {
+                        Promise.all(seasons.map(seasonItem => {
+                                return apiClient.getEpisodes(seasonItem.SeriesId, {
+                                    seasonId: seasonItem.Id,
+                                    userId: options.user.Id,
+                                    Fields: 'CanDownload,Path'
+                                });
+                            }
+                        )).then(seasonData => {
+                            downloadEpisodes([].concat.apply([], seasonData.map(season => season.Items)));
```
thornbill, line 384: It looks like we should be able to use `array.flat()` since we include the core js polyfills.

```suggestion
                            downloadEpisodes(seasonData.map(season => season.Items).flat());
```

thornbill (approved):

PR #4020, opened 2022-10-10
Add Home Assistant to device images

ctalkington:
**Changes**

I'm working on making Jellyfin integrate better with Home Assistant. This was just an extra detail I noticed.

**Issues**


nielsvanvelzen (approved):
The SVG could probably be optimized a bit with `svgo` but otherwise this change looks fine. Thanks for working on the HA integration!

PR #4028, opened 2022-10-11
Change settings menu item name to "Euskara"

urtzai:
"Euskara" is the native way of naming Basque

**Changes**
Change settings menu item HTML name to "Euskara"

**Issues**
Fixes translation and aesthetic issues


nielsvanvelzen (approved):

thornbill:
This should probably be pointed at the release-10.8.z branch since the other change was backported there.

PR #4033, opened 2022-10-12
Allow zero as MaxParentalRating value

Shadowghost:
jellyfin/jellyfin#8526 allows zero to be a valid content rating level and the current code does not respect that.

**Changes**
Allow zero as value for `MaxParentalRating`.

### DIFF
--- a/src/routes/user/userparentalcontrol.tsx
+++ b/src/routes/user/userparentalcontrol.tsx
@@ -224,7 +224,7 @@ const UserParentalControl: FunctionComponent = () => {
             }
 
             const parentalRating = parseInt((page.querySelector('#selectMaxParentalRating') as HTMLSelectElement).value || '0', 10);
-            user.Policy.MaxParentalRating = parentalRating !== undefined ? parentalRating : null;
+            user.Policy.MaxParentalRating = Number.isNaN(parentalRating) ? null : parentalRating;
             user.Policy.BlockUnratedItems = Array.prototype.filter.call(page.querySelectorAll('.chkUnratedItem'), function (i) {
                 return i.checked;
             }).map(function (i) {
--- a/src/routes/user/userparentalcontrol.tsx
+++ b/src/routes/user/userparentalcontrol.tsx
@@ -168,7 +168,7 @@ const UserParentalControl: FunctionComponent = () => {
         populateRatings(allParentalRatings);
         let ratingValue = '';
 
-        if (user.Policy.MaxParentalRating) {
+        if (user.Policy.MaxParentalRating != null) {
             for (let i = 0, length = allParentalRatings.length; i < length; i++) {
                 const rating = allParentalRatings[i];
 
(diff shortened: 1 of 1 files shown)


src/routes/user/userparentalcontrol.tsx
```diff
@@ -168,7 +168,7 @@ const UserParentalControl: FunctionComponent = () => {
         populateRatings(allParentalRatings);
         let ratingValue = '';
 
-        if (user.Policy.MaxParentalRating) {
+        if (user.Policy.MaxParentalRating !== null) {
```
dmitrylyzo, line 171: Shouldn't we also check for `undefined`?
```suggestion
        if (user.Policy.MaxParentalRating != null) {
```

src/routes/user/userparentalcontrol.tsx
```diff
@@ -223,7 +223,8 @@ const UserParentalControl: FunctionComponent = () => {
                 throw new Error('Unexpected null user.Policy');
             }
 
-            user.Policy.MaxParentalRating = parseInt((page.querySelector('#selectMaxParentalRating') as HTMLSelectElement).value || '0', 10) || null;
+            const parentalRating = parseInt((page.querySelector('#selectMaxParentalRating') as HTMLSelectElement).value || '0', 10);
+            user.Policy.MaxParentalRating = parentalRating !== undefined ? parentalRating : null;
```
nielsvanvelzen, line 227: parseInt will return a number or NaN so checking for undefined shouldn't do anything

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#return_value

Shadowghost: So we need to check for NaN?

nielsvanvelzen: yeah, and because it's JavaScript you need to use the isNaN function, don't do `== NaN` because that's always false

nielsvanvelzen (approved):

src/routes/user/userparentalcontrol.tsx
```diff
@@ -223,7 +223,8 @@ const UserParentalControl: FunctionComponent = () => {
                 throw new Error('Unexpected null user.Policy');
             }
 
-            user.Policy.MaxParentalRating = parseInt((page.querySelector('#selectMaxParentalRating') as HTMLSelectElement).value || '0', 10) || null;
+            const parentalRating = parseInt((page.querySelector('#selectMaxParentalRating') as HTMLSelectElement).value || '0', 10);
+            user.Policy.MaxParentalRating = Number.isNaN(parentalRating) ? null : parentalRating;
```
dmitrylyzo, line 227: I would write:
```js
user.Policy.MaxParentalRating = parseInt((page.querySelector('#selectMaxParentalRating') as HTMLSelectElement).value || '0', 10) || null;
```

Shadowghost: That's how it was before. When checking it in the debugger it did not provide the intended behaviour since parseInt returns `0` which is evaluated as `false` leading to `MaxParentalRating` being set to `null` instead of `0`.

dmitrylyzo: Ah, right.

PR #4035, opened 2022-10-12
added eslint to enable max function parameter

akp111:
**Changes**

Added rule to allow at max 7 parameters in a function. Tested it by including 8 parameters in a function and it gave warning
**Issues**

(Issue link) [https://github.com/jellyfin/jellyfin-web/issues/3494]

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -46,6 +46,7 @@ module.exports = {
         'keyword-spacing': ['error'],
         'no-throw-literal': ['error'],
         'max-statements-per-line': ['error'],
+        'max-params': ["error", 7],
         'no-duplicate-imports': ['error'],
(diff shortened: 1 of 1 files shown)


.eslintrc.js
```diff
@@ -46,6 +46,7 @@ module.exports = {
         'keyword-spacing': ['error'],
         'no-throw-literal': ['error'],
         'max-statements-per-line': ['error'],
+        'max-params': ["error", 7],
```
thornbill, line 49:
```suggestion
        'max-params': ['error', 7],
```

There are a few cases where functions use more than 7 parameters currently that will need to be fixed.

akp111: Shall I make it dynamic such that it takes value for env to set maximum number of parameters?

PR #4050, opened 2022-10-14
Replace apiclient event bus with local version

thornbill:
**Changes**
Moves to using a local implementation of the event bus to start transitioning away from using the apiclient.

**Issues**
N/A

### DIFF
--- a/src/utils/events.ts
+++ b/src/utils/events.ts
@@ -1,0 +1,50 @@
+﻿/* eslint-disable @typescript-eslint/no-explicit-any */
+interface Event {
+    type: string;
+}
+
+type callback = (e: Event, ...args: any[]) => void;
+
+function getCallbacks(obj: any, type: string): callback[] {
+    if (!obj) {
+        throw new Error('obj cannot be null!');
+    }
+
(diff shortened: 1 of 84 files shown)


src/utils/events.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+type callback = (e: Event, ...args: any[]) => void;
+
+function getCallbacks(obj: any, type: string): callback[] {
+    if (!obj) {
+        throw new Error('obj cannot be null!');
+    }
+
+    obj._callbacks = obj._callbacks || {};
```
nielsvanvelzen, line 13: You might want to use a symbol to add data on objects to avoid conflicts with json serializing / spread operators etc. etc.

thornbill: Well that was a fun dive into symbols. I really don't think I have used them before.

Unfortunately we have to keep the property name the same here since events that are triggered by the apiclient on the apiclient or connection manager still need to be handled by listeners attached in web.

nielsvanvelzen (approved):

PR #4056, opened 2022-10-15
Fix button group position on ItemDetails

dmitrylyzo:
In #3743, styles specific to TV layout were missed.

**Changes**
Add missing styles.

**Issues**
Button group is not padded in TV layout.

### DIFF
--- a/src/assets/css/librarybrowser.scss
+++ b/src/assets/css/librarybrowser.scss
@@ -810,6 +810,14 @@
     padding-left: 32.45vw;
 }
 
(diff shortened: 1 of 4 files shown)


nielsvanvelzen (approved):

dmitrylyzo:
After the "mirror" PR there are also bug with actionSheet options in the TV layout.

dmitrylyzo:
Should I split it into `fix styles` and `fix HTML escaping` PRs?

thornbill (approved):

thornbill:
I think this is fine as a single PR.

Issue #4055, opened 2022-10-15
Tizen 2.3: Master branch doesn't work


the4anoni:
**Describe The Bug**
Like in topic. When built with jellyfin-tizen, jellyfin-web master branch just freezes on initial screen

**Steps To Reproduce**
Build jellyfin-tizen and run it

**Expected Behavior**
It works

**Logs**
I don't have any logs

**System (please complete the following information):**
 - Platform: Tizen

**Additional Context**
Using jellyfin-web branch [v10.8.5](https://github.com/jellyfin/jellyfin-web/releases/tag/v10.8.5) fixed issue for me.

dmitrylyzo:
This is usually caused by ES5 incompatible code. It should work after 9d8f9c806c6e8e0c45bb38f87eabdac32bf71bba.
Current master (f0b29fa9cc4e168e499e292176e66073cb38596f) works on webOS 1.2 (~Tizen 2.3).

dmitrylyzo:
@the4anoni Are you on Windows?
https://github.com/jellyfin/jellyfin-tizen/issues/144#issuecomment-1289734592

the4anoni:
> @the4anoni Are you on Windows? [jellyfin/jellyfin-tizen#144 (comment)](https://github.com/jellyfin/jellyfin-tizen/issues/144#issuecomment-1289734592)

Yes, I am on Windows

PR #4057, opened 2022-10-16
Add toPercent utility for compatibility

thornbill:
**Changes**
Should fix compatibility issues with browsers that do not support the `Intl` functions.

**Issues**
Should fix part of https://github.com/jellyfin/jellyfin-web/issues/4055.

### DIFF
--- a/src/utils/number.ts
+++ b/src/utils/number.ts
@@ -5,6 +5,7 @@ function toLocaleStringSupportsOptions() {
 /**
  * Gets the value of a number formatted as a perentage.
  * @param {number} value The value as a number.
+ * @param {string} locale The locale to use for formatting (i.e. en-us).
  * @returns {string} The value formatted as a percentage.
  */
 export function toPercent(value: number | null | undefined, locale: string): string {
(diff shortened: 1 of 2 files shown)


src/utils/number.ts
```diff
@@ -0,0 +1,23 @@
+function toLocaleStringSupportsOptions() {
+    return !!(typeof Intl === 'object' && Intl && typeof Intl.NumberFormat === 'function');
+}
+
+/**
+ * Gets the value of a number formatted as a perentage.
+ * @param {number} value The value as a number.
+ * @returns {string} The value formatted as a percentage.
```
dmitrylyzo, line 8: doc for `locale` param?

dmitrylyzo (approved):
LGTM

If we want to support correct percent formatting on older web engines, we need a real polyfill or some custom implementation.
We could use the translation string as the format (as suggested [here](https://stackoverflow.com/a/66750952)). But that may not be enough, for example, for Farsi - non-Arabic numbers.
Anyway, I'm not an expert. :upside_down_face:

thornbill:
Yeah I think falling back to the previous behavior should be fine for older platforms that do not support the Intl APIs. The issue with the polyfills I saw was that they required importing multiple polyfills and each locale separately... which is a bit of a pain for such a minor usage.

PR #4069, opened 2022-10-18
Fix images sometimes not un-blurring

### LINKED ISSUE #3542: Images Won't Load
**Describe The Bug**
Images don't load

**Screenshots**
Screen Shot 2022-04-02 at 2 34 43 PM
Screen Shot 2022-04-02 at 3 02 41 PM

 - Platform: Mac
 - Browser: All
 - Jellyfin Version: 10.7.x - 10.8.x

### LINKED ISSUE #3693: Sudden row of contents  blur image [Issue]:
### Please describe your bug

When I scroll via browser  sudden  a  blur cover photoed  row I can find. Version: 10.7 and 10.8 both had it. Browser: microsoft edge.
screenshot 
image

### Jellyfin Version

10.8.0

### if other:

_No response_

### Environment

```markdown
- OS: Ubuntu 20.04
- Virtualization: Dedicated server
- Clients: Browser
- Browser: MS Edge Latest
- FFmpeg Version: 4.4.1-Jellyfin
- Playback Method: Direct Play
- Hardware Acceleration: No
- Plugins: default
- Reverse Proxy: nginx
- Base URL: /jellyfin
- Networking: Host
- Storage: Metadata is on ramdisk currently , normal storage also shows the same problem
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

image

### Code of Conduct

peterspenler:
**Changes**
Updates the image loader to use `isIntersecting` instead of `intersectionRatio` to determine if an image intersects the intersection area for lazy loading. If we use the `intersectionRatio` property to determine if an image is in the intersection area there is a case where the IntersectionObserver calls the callback exactly as the image enters the area. This means that the `intersectionRatio` property will still be 0, which is what we would expect when the image is outside of the area, even though the image has just entered the area. When this happens the full images never load to replace the blurred image, since the intersection callback is only called when the image leaves or enters the intersection area. By using the `isIntersecting` property instead we can reliably determine whether or not an image is in the intersection area.

**Issues**
Fixes #3693 
Fixes #3542

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -52,6 +52,7 @@
  - [MinecraftPlaye](https://github.com/MinecraftPlaye)
  - [Matthew Jones](https://github.com/matthew-jones-uk)
  - [taku0](https://github.com/taku0)
+ - [Peter Spenler](https://github.com/peterspenler)
 
 # Emby Contributors
 
--- a/src/components/images/imageLoader.js
+++ b/src/components/images/imageLoader.js
@@ -83,7 +83,7 @@ worker.addEventListener(
             source = entry;
         }
 
-        if (entry.intersectionRatio > 0) {
+        if (entry.isIntersecting) {
             if (source) {
                 fillImageElement(target, source);
             }


dmitrylyzo:
According to [this](https://caniuse.com/intersectionobserver), `isIntersecting` is not implemented in Chrome below 58. So this change needs to be tested on legacy browsers (webOS, Tizen).

thornbill:
I believe the polyfill we use should provide it but it's worth verifying.

dmitrylyzo (approved):

thornbill (approved):

peterzolczer:
I have this very issue on my fresh install on docker, 10.8.10. Blurry or no (cover/artist/poster/etc.) images, doesn't matter whether I use a client or open it in browser:

image

Denveous:
4 years have passed an no one has been able to fix this? Why is this PR'd and closed if it's still a bug in every browser including chrome?

PR #4081, opened 2022-10-21
Add search bar for plugin catalogue

darmiel:
**Changes**

Adds a search bar to the `Catalogue` section, allowing you to filter the available plugins.

This PR differs from #4079 in that, on the one hand, the plugin catalog can be searched (available plugins, not already installed ones) and, on the other hand, the category headings are also hidden.

I tried to make it as less intrusive as possible by adding `display: none` to cards and headers that do not match the filter.

https://user-images.githubusercontent.com/71837281/197134209-3ff4d6f1-c343-414b-b82a-a04bc0e54ff5.mov

### DIFF
--- a/src/controllers/dashboard/plugins/available/index.html
+++ b/src/controllers/dashboard/plugins/available/index.html
@@ -2,8 +2,7 @@
     <div>
         <div class="content-primary">
             <div class="inputContainer">
-                <label class="inputLabel" for="txtSearchPlugins">${Search}</label>
-                <input id="txtSearchPlugins" name="txtSearchPlugins" type="text" is="emby-input" class="emby-input">
+                <input id="txtSearchPlugins" name="txtSearchPlugins" type="text" is="emby-input" label="${Search}" />
             </div>
             <div id="noPlugins" class="hide">${MessageNoAvailablePlugins}</div>
             <div id="pluginTiles" style="text-align:left;"></div>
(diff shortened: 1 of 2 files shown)


src/controllers/dashboard/plugins/available/index.html
```diff
@@ -1,6 +1,10 @@
 <div id="pluginCatalogPage" data-role="page" class="page type-interior pluginConfigurationPage withTabs fullWidthContent">
     <div>
         <div class="content-primary">
+            <div class="inputContainer">
+                <label class="inputLabel" for="txtSearchPlugins">${Search}</label>
+                <input id="txtSearchPlugins" name="txtSearchPlugins" type="text" is="emby-input" class="emby-input">
```
thornbill, line 6:
```suggestion
                <input id="txtSearchPlugins" name="txtSearchPlugins" type="text" is="emby-input" label="${Search}" />
```

The label element and class are added by the webcomponent. Having the label element created manually causes an error to be thrown in the console. Your other PR likely had the same issue, but I missed it when testing.

thornbill (changes requested):

src/controllers/dashboard/plugins/available/index.html
darmiel: Oops, you're right. 
Should I open a new PR to fix the other PR, or should change it in this PR?

thornbill: Either way would have been fine since it's a small fix and somewhat related... but I'll probably go ahead and merge this one so make it a separate PR please. :slightly_smiling_face:

PR #4083, opened 2022-10-21
Refactor ConnectionState to a shared enum

thornbill:
**Changes**
Refactors the ConnectionState values to a shared enum. These `ConnectionState` values are an invention of the apiclient, so they will not exist in the new sdk.

**Issues**
N/A

### DIFF
--- a/src/utils/jellyfin-apiclient/ConnectionState.ts
+++ b/src/utils/jellyfin-apiclient/ConnectionState.ts
@@ -1,0 +1,10 @@
+/**
+ * Server state values for a connected server used by jellyfin-apiclient.
+ */
+export enum ConnectionState {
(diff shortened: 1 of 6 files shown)


src/utils/jellyfin-apiclient/ConnectionState.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Server state values for a connected server used by jellyfin-apiclient.
+ */
+export enum ConnectionState {
```
thornbill, line 4: I'm not sure that `utils` is the best place for this, but we really don't have a dedicated place for constants of this type. :man_shrugging:

dmitrylyzo (approved):

PR #4084, opened 2022-10-21
Fix double routing of root path

thornbill:
**Changes**
This moves the root path handling to react-router to fix an issue where it was being handled by both react-router and the legacy appRouter. This should fix the startup wizard (mostly). Additional work will be needed to fix this same type of double handling from happening on all other legacy routes though.

**Issues**
N/A

### DIFF
--- a/src/components/ConnectionRequired.tsx
+++ b/src/components/ConnectionRequired.tsx
@@ -96,6 +96,9 @@ const ConnectionRequired: FunctionComponent<ConnectionRequiredProps> = ({
                         }
                         const systemInfo = await infoResponse.json();
                         if (!systemInfo?.StartupWizardCompleted) {
+                            // Update the current ApiClient
+                            // TODO: Is there a better place to handle this?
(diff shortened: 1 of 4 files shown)


src/components/ConnectionRequired.tsx
```diff
@@ -96,6 +96,9 @@ const ConnectionRequired: FunctionComponent<ConnectionRequiredProps> = ({
                         }
                         const systemInfo = await infoResponse.json();
                         if (!systemInfo?.StartupWizardCompleted) {
+                            // Update the current ApiClient
+                            // TODO: Is there a better place to handle this?
+                            ServerConnections.setLocalApiClient(firstConnection.ApiClient);
```
thornbill, line 101: This is something the `appRouter` does that we missed when porting the logic over. Without it you can end up with an exception because the current api client is null after navigating to the startup wizard.

nielsvanvelzen (approved):

PR #4086, opened 2022-10-22
Update all documentation links to new website

nielsvanvelzen:
**Changes**
- Update all docs.jellyfin.org links to jellyfin.org/docs
- Update all jellyfin.org/docs links to new link format (mostly removing the .html suffix)
- Update a http link to https in index.html

The links are always the same as the menu on the new website, this means that they don't end with a `/` when there is no subpage. The website itself does add this slash when the page is opened.

Also, vscode automatically trimmed whitespace in the repo-stale.yaml file but the yaml is still valid.

**Issues**

### DIFF
--- a/.github/SUPPORT.md
+++ b/.github/SUPPORT.md
@@ -7,7 +7,7 @@ When looking for support or information, please first search for your
 question in these venues:
 
 * [Jellyfin Forum](https://forum.jellyfin.org)
-* [Jellyfin Documentation](https://docs.jellyfin.org)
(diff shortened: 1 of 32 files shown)


Shadowghost:
We should PR this against 10.8 and backport otherwise upcoming bugfix releases will have broken links too.

nielsvanvelzen:
The old links still work, they have redirects in place. I didn't bother with a backport because the release and master branch have diverged too much.

Bond-009 (approved):

thornbill (approved):

PR #4112, opened 2022-10-27
removed disclaimer target _blank enforcement

derlucas:
My blind try to change the JS code. It's untested. I want to allow "target" attribute on "a" Elements in the disclaimer.

**Changes**
changed session/login/index.js to check if the "target" property on an "<a>" element is present, if not it will be set to "_blank" as it used to be.

### DIFF
--- a/src/controllers/session/login/index.js
+++ b/src/controllers/session/login/index.js
@@ -290,7 +290,9 @@ import './login.scss';
 
                 for (const elem of disclaimer.querySelectorAll('a')) {
                     elem.rel = 'noopener noreferrer';
-                    elem.target = '_blank';
+                    if( typeof elem.target === 'undefined' || !(elem.target)) {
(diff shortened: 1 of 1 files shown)


dmitrylyzo:
The `target` attribute is currently being sanitized by DOMPurify (https://github.com/cure53/DOMPurify/issues/317)
As for the current `target="_blank"`, it looks like we need to add `rel="noopener noreferrer"` (https://github.com/cure53/DOMPurify/issues/317#issuecomment-670850160).

thornbill:
Personally I'm not a fan of this change. It will cause issues in the apps that wrap web if a different site opens in the current context.

thornbill:
Closing due to incompatibility with apps.

derlucas:
Would you accept it, if the feature is configurable to give Admins a choice to use it?

PR #4116, opened 2022-10-28
Add api and user contexts

thornbill:
**Changes**
* Moves React App setup to separate file
* Adds utility to create an SDK Api instance from an ApiClient instance
* Adds api and user contexts
* Updates search placeholder component to use contexts

**Issues**
N/A


src/hooks/useApi.tsx
```diff
@@ -0,0 +1,71 @@
...
+    const [ api, setApi ] = useState<Api>();
+    const [ user, setUser ] = useState<UserDto>();
+
+    useEffect(() => {
+        connections.currentApiClient()
+            .getCurrentUser()
+            .then(newUser => udpateApiUser(null, newUser))
+            .catch(err => {
+                console.warn('[ApiProvider] Could not get current user', err);
+            });
```
grafixeyehero, line 34: why?

grafixeyehero: getCurrentUser() could be null if no stored server available

thornbill: Good catch. I didn't realize the apiclient would throw if there is no current user there.

grafixeyehero:
@thornbill here is a log output `useApi.tsx?9440:41 Uncaught TypeError: Cannot read properties of null (reading 'getCurrentUser')
    at eval (useApi.tsx?9440:41:1)
    at invokePassiveEffectCreate (react-dom.development.js?f8c1:23487:1)
    at HTMLUnknownElement.callCallback (react-dom.development.js?f8c1:3945:1)
    at Object.invokeGuardedCallbackDev (react-dom.development.js?f8c1:3994:1)
    at invokeGuardedCallback (react-dom.development.js?f8c1:4056:1)
    at flushPassiveEffectsImpl (react-dom.development.js?f8c1:23574:1)
    at unstable_runWithPriority (scheduler.development.js?bacd:468:1)
    at runWithPriority$1 (react-dom.development.js?f8c1:11276:1)
    at flushPassiveEffects (react-dom.development.js?f8c1:23447:1)
    at eval (react-dom.development.js?f8c1:23324:1)`

thornbill:
I have no idea what would have caused that since it's just an imported object. There doesn't seem to really be any reason to use a prop for that though so I just moved the import directly into the hook. Hopefully that will fix this. :crossed_fingers:

src/hooks/useApi.tsx
```diff
@@ -0,0 +1,67 @@
...
+export const ApiContext = createContext<JellyfinApiContext>({});
+export const useApi = () => useContext(ApiContext);
+
+export const ApiProvider: FC = ({ children }) => {
+    const [ legacyApiClient, setLegacyApiClient ] = useState<ApiClient>();
+    const [ api, setApi ] = useState<Api>();
+    const [ user, setUser ] = useState<UserDto>();
+
+    useEffect(() => {
+        ServerConnections.currentApiClient()
```
grafixeyehero, line 25:
```suggestion
const apiClient = ServerConnections.currentApiClient();
        if (apiClient) {
            apiClient.getCurrentUser()
                .then(newUser => updateApiUser(undefined, newUser))
                .catch(err => {
                    console.info('[ApiProvider] Could not get current user', err);
                });
        }```

grafixeyehero: @thornbill this solve the problem

thornbill: Should be fixed now. :crossed_fingers: I updated the return types for `currentApiClient()` to include `undefined` and fixed the related issues.

grafixeyehero (approved):

PR #4125, opened 2022-10-31
Fix: Remove duplicated HTML escape in remote-control title

joschahenningsen:
**Changes**

The `remoteControl` screen shows the title of my content escaped twice:

image

This issue occurs because the title is escaped here
https://github.com/jellyfin/jellyfin-web/blob/e0bc17a018056a9c588240df52b74626ed3b1f68/src/components/playback/nowplayinghelper.js#L64
And a second time here:
https://github.com/jellyfin/jellyfin-web/blob/e0bc17a018056a9c588240df52b74626ed3b1f68/src/components/remotecontrol/remotecontrol.js#L89

This PR removes the second one.


dmitrylyzo:
Good catch. :+1:

Since `MediaSession` is also currently "broken" by escaping, I suggest to remove the escaping from `getNowPlayingNames` and add it to `updateNowPlayingInfo`:
https://github.com/jellyfin/jellyfin-web/blob/f8b0c941f5e97d99cd27cf69e6fe3962314577d5/src/components/nowPlayingBar/nowPlayingBar.js#L503
https://github.com/jellyfin/jellyfin-web/blob/f8b0c941f5e97d99cd27cf69e6fe3962314577d5/src/components/nowPlayingBar/nowPlayingBar.js#L510

joschahenningsen:
Sure, can do that tomorrow :)

dmitrylyzo:
Superseded by #4243

PR #4134, opened 2022-11-02
Exclude location.state from effect dependencies

thornbill:
**Changes**
Including the state here causes the current view to try to be restored when dialogs that modify location state are activated

**Issues**
Fixes https://github.com/jellyfin/jellyfin-tizen/issues/149


dmitrylyzo (approved):
#4138 as a continuation

PR #4153, opened 2022-11-06
Fix nowPlayingPageImageContainer height

Aimekram:
**Changes**
I set the container to preferably be square so it now acts like an image placeholder and as soon as the image is square it should work correctly :)

**Issues**
Fixes https://github.com/jellyfin/jellyfin-web/issues/2605


thornbill:
I think this component might also be used when playing content on another device using the "play on"/cast menu in which case the image might not always be square. That would need testing though.

dmitrylyzo:
In mobile layout in portrait orientation, the image container is `flex`, so image is centered and looks fine (`aspect-ratio` doesn't seem to affect flexbox).
We just need to make the height const.

Custom CSS for debugging:
```css
.nowPlayingInfoContainer{background-color:rgba(0,255,0,0.5)!important}
.nowPlayingPageImageContainer{background-color:rgba(255,0,0,0.5)!important}
.nowPlayingInfoControls{background-color:rgba(0,0,255,0.5)!important}
.nowPlayingInfoButtons{background-color:rgba(0,255,255,0.5)!important}
.nowPlayingPageImage,.nowPlayingPageImageAudio{opacity:0.5!important}
```

Aimekram:
Thanks for looking at my changes :) Should I do something more here?

dmitrylyzo:
I played with CSS, this makes the container a fixed area. The downside is that it probably won't upscale a smaller image:
```diff
diff --git a/src/components/remotecontrol/remotecontrol.scss b/src/components/remotecontrol/remotecontrol.scss
index 616ae61c4..6f665c7be 100644
--- a/src/components/remotecontrol/remotecontrol.scss
+++ b/src/components/remotecontrol/remotecontrol.scss
@@ -103,6 +103,7 @@
 }
 
 .nowPlayingPageImageContainer {
+    display: flex;
     width: 16%;
     aspect-ratio: auto 1 / 1;
     position: relative;
@@ -153,12 +154,9 @@
 }
 
 .nowPlayingPageImage {
-    display: block;
-    bottom: 0;
-    left: 0;
-    right: 0;
-    margin: 0 auto;
-    width: 100%;
+    margin: auto;
+    max-height: 100%;
+    max-width: 100%;
     -webkit-box-shadow: 0 0 1.9vh #000;
     box-shadow: 0 0 1.9vh #000;
     border-radius: 0.2em;
@@ -444,13 +442,6 @@
         font-size: 2em;
     }
 
-    .nowPlayingPageImage.nowPlayingPageImageAudio,
-    .nowPlayingPageImage.nowPlayingPageImagePoster {
-        width: auto;
-        max-width: 100%;
-        max-height: 100%;
-    }
-
     .playlistSectionButton .volumecontrol {
         width: 100%;
     }
```
Tested only in Chrome 107 and Firefox 107.

PR #4157, opened 2022-11-07
Remove like/dislike filter

nielsvanvelzen:
There is no way to like/dislike items in jellyfin-web (and it was also removed some time ago in the Android TV app). So it doesn't make sense to have filters for liked/disliked content.

I'm not sure how to deal with the translations in jellyfin-web. The OptionLikes and OptionDislikes keys are now unused, should I remove those from the base language file and have Weblate deal with the translations?

**Changes**
- Remove like/dislike filtering
- Remove unused like/dislike handling in userdatabuttons.js 

**Issues**


thornbill:
> I'm not sure how to deal with the translations in jellyfin-web. The OptionLikes and OptionDislikes keys are now unused, should I remove those from the base language file and have Weblate deal with the translations?

Yes, I believe you just remove them from the base language.

nielsvanvelzen:
Thanks, same as the Android app then. Wasn't sure if jf-web was configured the same. Updated the PR.

thornbill (approved):

PR #4163, opened 2022-11-09
Reduce tv card scaling

Vankerkom:
**Changes**
Reduced the transform scale from 18% to 7% to fix the readability of heading text on focus.

tv-card-size

**Issues**
None

### DIFF
--- a/src/components/cardbuilder/card.scss
+++ b/src/components/cardbuilder/card.scss
@@ -114,7 +114,7 @@ button::-moz-focus-inner {
 }
 
 .card.show-animation:focus > .cardBox {
-    transform: scale(1.18, 1.18);
+    transform: scale(1.07, 1.07);
 }
 
 .cardBox-bottompadded {


thornbill:
@dmitrylyzo I will defer to you on this one since I don't really use the TV display mode..

dmitrylyzo:
Real TVs don't use card scaling (`.show-animation` class) to indicate focus - they are considered "slow". So I can't say much either.

It is harder to distinguish a focused item with reduced scale, imo.

[Here](https://youtu.be/qAwEE7JHEiM?t=26) is how Samsung solves this problem.
When the first item is selected, the row header moves up. It returns back when it isn't overlapped by an item.

At the same time, with +18% scaling, actor's posters on the movie details are cropped on the 4K display.

`1.5em` is not enough:
https://github.com/jellyfin/jellyfin-web/blob/3f7898280202e932a934103b3cec385adbb9151a/src/assets/css/librarybrowser.scss#L1414-L1422

Proposed +7%

thornbill:
Hey @Vankerkom would you be able to resolve the merge conflicts on this? Thanks!

thornbill:
@jellyfin-bot rebase

PR #4166, opened 2022-11-10
Fix keyboard navigation for INPUT and TEXTAREA

dmitrylyzo:
**Changes**
Ignore Left/Right if the interactive element is focused.

**Issues**
Left/Right (move cursor) doesn't work when editing a value in `input`.


Vankerkom (approved):
Clean implementation 👍

Vankerkom (approved):

thornbill (approved):

PR #4170, opened 2022-11-14
Add encoder settings for ffmpeg's segment deletion

DomiStyle:
**Changes**
Exposes the new setting for segment deletion and also makes throttler and segment deletion time configurable in case anybody wants to keep a longer or shorter buffer.

For the related server PR see here: https://github.com/jellyfin/jellyfin/pull/8744

**Issues**
https://github.com/jellyfin/jellyfin/issues/3929


DomiStyle:
Pushed one more commit that prevents the transcoder from starting up twice when seeking due to HLS.js requesting segment 0 and 1 before seeking to the correct position.

Should speed up transcode startup slightly when continuing a video file.

nyanmisaka:
> Pushed one more commit that prevents the transcoder from starting up twice when seeking due to HLS.js requesting segment 0 and 1 before seeking to the correct position.
> 
> Should speed up transcode startup slightly when continuing a video file.

This should only be used on the initial loading. Otherwise this will break the seeking.

DomiStyle:
> This should only be used on the initial loading. Otherwise this will break the seeking.

The HLS player is only created for the first load? The playlist is never loaded again after playback starts unless I'm overseeing something.

Seeking has been working fine in my tests.

nyanmisaka:
> > This should only be used on the initial loading. Otherwise this will break the seeking.
> 
> The HLS player is only created for the first load? The playlist is never loaded again after playback starts unless I'm overseeing something.
> 
> Seeking has been working fine in my tests.

You're right. I happened to try a file that couldn't seek due to DOVI metadata.

nussdk (approved):
LGTM

thornbill:
Server side changes would need merged first

nussdk:
Yeah can we look at the server side change too. https://github.com/jellyfin/jellyfin-web/pull/4170https://github.com/jellyfin/jellyfin/pull/8744 This will let users use RAM as transcoding cache and not have it lock up the server if RAM gets full. maybe we look into adding a %full max to transcoding also to start override oldest files if >90% of disk usage also   Kyle Nussdorfer From: Bill ThorntonSent: Sunday, December 11, 2022 5:22 PMTo: jellyfin/jellyfin-webCc: nussdk; CommentSubject: Re: [jellyfin/jellyfin-web] Add encoder settings for ffmpeg's segment deletion (PR #4170) Server side changes would need merged first—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you commented.Message ID: ***@***.***>

nyanmisaka (approved):

nyanmisaka:
Here are some eslint errors

```
/home/runner/work/jellyfin-web/jellyfin-web/src/controllers/dashboard/encodingsettings.js
Error:   111:43  error    Missing radix parameter                                                                                                                                             radix
Error:   112:41  error    Missing radix parameter                                                                                                                                             radix
```

DomiStyle:
@nyanmisaka Should be fixed.

thornbill (approved):

PR #4181, opened 2022-11-20
Make Height editable on TV Channels

SenorSmartyPants:
**Changes**

Editing metadata for a channel now allows user to set Height. Useful in determining which channels are HD. 

**Issues**

Depends on jellyfin/jellyfin#8777
Useful in conjunction with jellyfin/jellyfin#8768


SenorSmartyPants:
Back end has been merged.

thornbill:
I think from a ux standpoint it would be better to have a drop down menu for common resolutions.

thornbill:
I rebased this because sonarqube was flagging some issues that had been resolved since the PR was originally opened. The code changes look pretty good, I'm just adding a label so I remember to give it a quick test before approving + merging. :slightly_smiling_face:

thornbill:
@jellyfin-bot rebase

SenorSmartyPants:
Rebase, fix merge conflict and force push.

thornbill:
Setting the "height" of a TV channel seems like a really strange concept from a user standpoint. I think it would be better if we call it "resolution" and display standard resolutions as the options with values being the height.

I think I would just make the options `SD, SD (PAL), HD, Full HD, UHD (4K)`.

SenorSmartyPants:
I used height as it is an existing field and kept the label. I have no problem with the concept of updating these labels.

```
                    <select is="emby-select" id="selectHeight" label="${LabelHeight}">
                        <option value="0"></option>
                        <option value="480">SD</option>
                        <option value="576">SD (PAL)</option>
                        <option value="720">HD</option>
                        <option value="1080">Full HD</option>
                        <option value="2160">UHD (4K)</option>
                    </select>
```

Or maybe those should be added as translatable resource strings.

Feel free to modify this PR as you wish.

thornbill:
Yeah to be clear I don't have any issues at all with the height field being used for the backend. I just think it can be presented in a more user friendly manner on the frontend. 🙂

nielsvanvelzen (approved):

PR #4188, opened 2022-11-27
Fix unlimited pagination

ChristopheVandePoel:
**Changes**
Setting page size to 0 in settings disables pagination (expected). Most current master commit sends 0 as an actual page size in the query. This breaks views for those who have pagination disabled (effectively showing empty library).

PR has small change that removes the query limit and disables pagination controls if page size is set to 0.

### DIFF
--- a/src/components/common/Pagination.tsx
+++ b/src/components/common/Pagination.tsx
@@ -16,7 +16,7 @@ const Pagination: FC<PaginationProps> = ({ viewQuerySettings, setViewQuerySettin
     const totalRecordCount = itemsResult.TotalRecordCount || 0;
     const startIndex = viewQuerySettings.StartIndex || 0;
     const recordsEnd = Math.min(startIndex + limit, totalRecordCount);
-    const showControls = limit < totalRecordCount;
+    const showControls = limit > 0 && limit < totalRecordCount;
     const element = useRef<HTMLDivElement>(null);
 
     const onNextPageClick = useCallback(() => {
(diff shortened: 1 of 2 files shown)


dmitrylyzo (commented):
PR works, but I think the `Pagination` component shouldn't use `userSettings.libraryPageSize` directly - it should use the limit of its parent item view.

Just nitpicking.
I think the commit should be called `fix: ...` because it is really a fix. Or `Fix unlimited pagination`.

ChristopheVandePoel:
> PR works, but I think the `Pagination` component shouldn't use `userSettings.libraryPageSize` directly - it should use the limit of its parent item view.

I noticed that as well, and I agree, but I figured that this rewrite to react was part of a larger refactor effort and I didnt want the PR to cause conflicts elsewhere. I simply used what was at hand so that it uses that limit parameter to set things properly for now.

I'll commit the update you suggested. The only gripe I have is that this is sort of an implicit condition. It's not very well documented that "if limit === 0 -> pagination is disabled".

dmitrylyzo (approved):
It would be better to squash the commits and name (amend) it `fix: ...` or `Fix unlimited pagination`. Then force push. :point_left: This is  what I meant in my previous post.

Btw, you currently have different authors in commits, and you might want to use (amend) the Github one.

ChristopheVandePoel:
> It would be better to squash the commits and name (amend) it `fix: ...` or `Fix unlimited pagination`. Then force push. 👈 This is what I meant in my previous post.

Oh wow. Scatterbrain me, I thought you were talking about the PR title here, sorry :)

> Btw, you currently have different authors in commits, and you might want to use (amend) the Github one.

Yeah. Juggling a couple of different profiles at the same time on my machine. I squashed with the github one.

PR #4196, opened 2022-12-09
Show production studio

mark5231:
**Changes**

image

Show studios. Support this feature request: https://features.jellyfin.org/posts/1203/show-production-studio-for-movies-tv-shows
**Issues**


thornbill (changes requested):
Looks great!

Just one minor suggestion to simplify updating the hide class. As far as the placement goes, I think this might be better suited under the genres.

mark5231:
> Looks great!
> 
> Just one minor suggestion to simplify updating the hide class. As far as the placement goes, I think this might be better suited under the genres.

Thanks for your review. I applied your suggestion. Please take a look.

src/controllers/itemDetails/index.js
```diff
@@ -1031,6 +1031,29 @@ function renderDirector(page, item, context) {
...
+            Type: 'Studio',
+            ServerId: item.ServerId,
+            Id: studio.Id
+        }, {
+            context: context
+        }) + '">' + escapeHtml(studio.Name) + '</a>';
+    }).join(', ');
+
+    const studiosLabel = page.querySelector('.studiosLabel');
+    studiosLabel.innerHTML = globalize.translate(studios.length > 1 ? 'Studios' : 'Studio');
```
thornbill, line 1049: `innerText` is safer to use if you are only inserting text (removes the possibility of inadvertently introducing a cross-site scripting vulnerability)

```suggestion
    studiosLabel.innerText = globalize.translate(studios.length > 1 ? 'Studios' : 'Studio');
```

thornbill (changes requested):
I do prefer the general approach taken here over #4261. We will need to clean up the duplication on series pages, but that can always be done in a separate PR.

thornbill (approved):

PR #4198, opened 2022-12-09
Use unstable tag for sdk version

thornbill:
**Changes**
Updates the sdk version to use the unstable tag

**Issues**
N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "@fontsource/noto-sans-sc": "4.5.12",
     "@fontsource/noto-sans-tc": "4.5.12",
     "@jellyfin/libass-wasm": "4.1.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -75,7 +75,7 @@
     "@fontsource/noto-sans-sc": "4.5.12",
     "@fontsource/noto-sans-tc": "4.5.12",
     "@jellyfin/libass-wasm": "4.1.1",
-    "@jellyfin/sdk": "0.7.0",
+    "@jellyfin/sdk": "unstable",
```
thornbill, line 78: I'm not sure how renovate will handle this, but I guess the only way to find out is to try it and then adjust the rules as needed. :sweat_smile:

crobibero (approved):

PR #4199, opened 2022-12-10
Add setting for collection management

Shadowghost:
Provides the interface for jellyfin/jellyfin#8887 and enforces the new setting.

### DIFF
--- a/src/components/itemContextMenu.js
+++ b/src/components/itemContextMenu.js
@@ -15,7 +15,6 @@ import toast from './toast/toast';
         const user = options.user;
 
         const canPlay = playbackManager.canPlay(item);
-        const restrictOptions = (browser.operaTv || browser.web0s) && !user.Policy.IsAdministrator;
 
         const commands = [];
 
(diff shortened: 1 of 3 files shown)


src/components/itemContextMenu.js
```diff
@@ -99,7 +99,7 @@ import toast from './toast/toast';
             });
         }
 
-        if (!restrictOptions) {
+        if (options.EnableCollectionManagement && !restrictOptions) {
```
thornbill, line 102: Should this be requiring admin AND the new permission?

Also should the permission apply to playlists also? or only collections?

Shadowghost: Permission should only apply to collections and not additionally require admin.

thornbill: This `if` statement is wrapping the options to add to collections and playlists and the `restrictOptions` boolean includes an admin check. (I would also guess that we should just be checking for `browser.tv` instead of only checking specific TV OSes for this.)

src/components/itemContextMenu.js
```diff
@@ -99,7 +99,7 @@ import toast from './toast/toast';
             });
         }
 
-        if (!restrictOptions) {
+        if (options.EnableCollectionManagement && !browser.tv) {
```
thornbill, line 102: This is still including the "Add to playlist" menu item as well. :wink:

thornbill (changes requested):

src/components/itemContextMenu.js
Shadowghost: I removed the admin check alltogether. There is no reason why any of those items should only be available for admins.
Replaced it by just checking for `browser.tv` and disabling the entries if we're on TV.

PR #4201, opened 2022-12-11
Show total count when no pagination is applied

h3llrais3r:
If no pagination is applied (or disabled in the settings), there is no view on how many items you have in your library. By showing the pagination text (without the pagination buttons) it's still visible for the user how many items are listed in the library.

Without paging enabled (or you don't reach the configured amount of items for paging):
image

With paging enabled (as-is):
image

### DIFF
--- a/src/components/common/Pagination.tsx
+++ b/src/components/common/Pagination.tsx
@@ -15,7 +15,8 @@ const Pagination: FC<PaginationProps> = ({ viewQuerySettings, setViewQuerySettin
     const limit = userSettings.libraryPageSize(undefined);
     const totalRecordCount = itemsResult.TotalRecordCount || 0;
     const startIndex = viewQuerySettings.StartIndex || 0;
-    const recordsEnd = Math.min(startIndex + limit, totalRecordCount);
+    const recordsStart = totalRecordCount ? startIndex + 1 : 0;
+    const recordsEnd = limit ? Math.min(startIndex + limit, totalRecordCount) : totalRecordCount;
     const showControls = limit > 0 && limit < totalRecordCount;
     const element = useRef<HTMLDivElement>(null);
 
@@ -71,7 +72,7 @@ const Pagination: FC<PaginationProps> = ({ viewQuerySettings, setViewQuerySettin
             <div className='paging'>
                 <div className='listPaging' style={{ display: 'flex', alignItems: 'center' }}>
                     <span>
-                        {globalize.translate('ListPaging', (totalRecordCount ? startIndex + 1 : 0), recordsEnd || totalRecordCount, totalRecordCount)}
+                        {globalize.translate('ListPaging', recordsStart, recordsEnd, totalRecordCount)}
                     </span>
                     {showControls && (
                         <>
--- a/src/components/common/Pagination.tsx
+++ b/src/components/common/Pagination.tsx
@@ -74,7 +74,7 @@ const Pagination: FC<PaginationProps> = ({ viewQuerySettings, setViewQuerySettin
                         {globalize.translate('ListPaging', (totalRecordCount ? startIndex + 1 : 0), recordsEnd || totalRecordCount, totalRecordCount)}
                     </span>
                     {showControls && (
-                        <div style={{ display: 'inline-flex' }}>
+                        <>
                             <IconButtonElement
                                 is='paper-icon-button-light'
                                 className='btnPreviousPage autoSize'
@@ -85,7 +85,7 @@ const Pagination: FC<PaginationProps> = ({ viewQuerySettings, setViewQuerySettin
                                 className='btnNextPage autoSize'
                                 icon='material-icons arrow_forward'
                             />
-                        </div>
+                        </>
                     )}
                 </div>
             </div>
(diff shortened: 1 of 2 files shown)


src/components/common/Pagination.tsx
```diff
@@ -88,6 +88,15 @@ const Pagination: FC<PaginationProps> = ({ viewQuerySettings, setViewQuerySettin
                         />
                     </div>
                 )}
+                {!showControls && (
```
thornbill, line 91: We should be able to just use the same logic as above and move the `showControls` check to wrap the button elements.

src/scripts/libraryBrowser.js
```diff
@@ -90,6 +90,10 @@ export function getQueryPagingHtml (options) {
         html += '<span style="vertical-align:middle;">';
         html += globalize.translate('ListPaging', totalRecordCount ? startIndex + 1 : 0, recordsEnd, totalRecordCount);
         html += '</span>';
+    } else {
```
thornbill, line 93: Similar thing here, but I think we can remove the check entirely and always use the logic above.

thornbill (changes requested):

src/components/common/Pagination.tsx
h3llrais3r: Could do it like this:
image

h3llrais3r:
Logic simplified in last commit.

src/components/common/Pagination.tsx
```diff
@@ -69,25 +69,25 @@ const Pagination: FC<PaginationProps> = ({ viewQuerySettings, setViewQuerySettin
...
-                            icon='material-icons arrow_forward'
-                        />
-                    </div>
-                )}
+                <div className='listPaging' style={{ display: 'flex', alignItems: 'center' }}>
+                    <span>
+                        {globalize.translate('ListPaging', (totalRecordCount ? startIndex + 1 : 0), recordsEnd || totalRecordCount, totalRecordCount)}
+                    </span>
+                    {showControls && (
+                        <div style={{ display: 'inline-flex' }}>
```
thornbill, line 77: I believe you can use a react fragment (`<>`) here instead of adding a `<div>`.

thornbill (changes requested):

src/scripts/libraryBrowser.js
thornbill: It should equal the `totalRecordCount` I think.

```js
const recordsEnd = Math.min(startIndex + limit, totalRecordCount)
```

h3llrais3r: No it's, not, as it's the `Math.min(..., ...)` and the `startIndex + limit` equals to `0` when no paging is enabled.

src/components/common/Pagination.tsx
h3llrais3r: I'll try, but I didn't use it because I don't see it used somewhere else...

h3llrais3r: Fixed.

h3llrais3r:
Any other changes required?

thornbill:
Sorry for the delay. I'm a bit behind on reviews due to illness and holidays. I'll try to take a look at your latest changes this week. Thanks!

h3llrais3r:
> Sorry for the delay. I'm a bit behind on reviews due to illness and holidays. I'll try to take a look at your latest changes this week. Thanks!

No problem, take your time. Health is more important than coding. ;)

src/scripts/libraryBrowser.js
```diff
@@ -86,11 +86,9 @@ export function getQueryPagingHtml (options) {
 
     html += '<div class="listPaging">';
 
-    if (showControls) {
-        html += '<span style="vertical-align:middle;">';
-        html += globalize.translate('ListPaging', totalRecordCount ? startIndex + 1 : 0, recordsEnd, totalRecordCount);
-        html += '</span>';
-    }
+    html += '<span style="vertical-align:middle;">';
+    html += globalize.translate('ListPaging', totalRecordCount ? startIndex + 1 : 0, recordsEnd || totalRecordCount, totalRecordCount);
```
thornbill, line 90: I think I would prefer we handle this when we assign the value of `recordsEnd` to reduce the logic here in the templating code.

```js
const recordsEnd = limit ? Math.min(startIndex + limit, totalRecordCount) : totalRecordCount;
```

h3llrais3r: Will do.
Do you want to create a const for the `recordsStart` as well?
Then we can even make it more simple in the templating logic:
`const recordsStart = totalRecordCount ? startIndex + 1 : 0;`

thornbill: Sure that sounds good.

h3llrais3r: Done

thornbill (approved):

h3llrais3r:
Hi, in which version has this been released? I'm still unable to see the page count.
I'm running version 10.8.9

h3llrais3r:
Thx, will try again when 10.9.0 is released.

PR #4207, opened 2022-12-13
add arctic theme

jakehamilton:
**Changes**

This PR adds a new theme based on [Nord](https://www.nordtheme.com/). The whole application is themed (or nearly) with a soft, modern dark color scheme.

**Issues**

**Screenshots**

<details>
  <summary>View Screenshots</summary>

Screenshot from 2022-12-13 14-54-29
Screenshot from 2022-12-13 14-54-16
Screenshot from 2022-12-13 14-54-08
Screenshot from 2022-12-13 14-54-02
Screenshot from 2022-12-13 14-54-00
Screenshot from 2022-12-13 14-53-54
Screenshot from 2022-12-13 14-53-51
Screenshot from 2022-12-13 14-53-48
Screenshot from 2022-12-13 14-53-44
Screenshot from 2022-12-13 14-53-41
Screenshot from 2022-12-13 14-53-37
Screenshot from 2022-12-13 14-58-57
Screenshot from 2022-12-13 14-58-54
Screenshot from 2022-12-13 14-58-51
Screenshot from 2022-12-13 14-58-39
Screenshot from 2022-12-13 14-57-44
Screenshot from 2022-12-13 14-56-57
Screenshot from 2022-12-13 14-56-20
Screenshot from 2022-12-13 14-56-13
Screenshot from 2022-12-13 14-56-02
Screenshot from 2022-12-13 14-55-52
Screenshot from 2022-12-13 14-54-46

</details>

**Caveats**
I have tested desktop, mobile, and tv layouts on my computer but have not yet actually loaded it onto mobile or tv devices. There are undoubtedly quirks and I'm hoping that others will be able to help test the theme to see what breaks :pray:. Additionally, there are some features of Jellyfin that I don't use (eg. tuners), so this theme may have accidental omissions due to me not having certain parts of the UI visible when theming.

Finally, this is the first draft of the theme. If it looks fine visually and there aren't massive technical issues with it, I'd like to then take another pass at the code to organize and clean it up :+1:.

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -6,27 +6,37 @@
       "name": "Apple TV",
       "id": "appletv",
       "color": "#bcbcbc"
-    }, {
+    },
+    {
       "name": "Blue Radiance",
       "id": "blueradiance",
       "color": "#011432"
-    }, {
+    },
+    {
       "name": "Dark",
       "id": "dark",
       "color": "#202020",
       "default": true
-    }, {
+    },
+    {
       "name": "Light",
       "id": "light",
       "color": "#303030"
-    }, {
+    },
+    {
       "name": "Purple Haze",
       "id": "purplehaze",
       "color": "#000420"
-    }, {
+    },
+    {
       "name": "WMC",
       "id": "wmc",
       "color": "#0c2450"
+    },
+    {
+      "name": "Arctic",
+      "id": "arctic",
+      "color": "#ff0000"
     }
   ],
   "menuLinks": [],
--- a/src/themes/arctic/theme.css
+++ b/src/themes/arctic/theme.css
@@ -1,0 +1,1264 @@
+:root {
+    --nord0: 46, 52, 64;
+    --nord1: 59, 66, 82;
+    --nord2: 67, 76, 94;
+    --nord3: 76, 86, 106;
+    --nord4: 216, 222, 233;
+    --nord5: 229, 233, 240;
+    --nord6: 236, 239, 244;
+    --nord7: 143, 188, 187;
+    --nord8: 136, 192, 208;
+    --nord9: 129, 161, 193;
+    --nord10: 94, 129, 172;
+    --nord11: 191, 97, 106;
+    --nord12: 208, 135, 112;
+    --nord13: 235, 203, 139;
+    --nord14: 163, 190, 140;
+    --nord15: 180, 142, 173;
+
+    --success: var(--nord14);
+    --success-text: var(--nord6);
+
+    --warning: var(--nord13);
+    --warning-text: var(--nord6);
+
+    --error: var(--nord11);
+    --error-text: var(--nord6);
+
+    --primary: var(--nord9);
+    --primary-text: var(--nord6);
+
+    --primary-light: var(--nord8);
+    --primary-text-light: var(--nord6);
+
+    --secondary: var(--nord15);
+    --secondary-text: var(--nord6);
+
+    --tertiary: var(--nord10);
+    --tertiary-text: var(--nord6);
+
+    --bg: var(--nord0);
+    --bg-light: var(--nord1);
+
+    --text: var(--nord6);
+    --text-light: var(--nord4);
+}
+
+h2,
+h3 {
+    font-weight: bold;
+}
+
+.skinHeader,
+html {
+    color: rgb(var(--text));
+}
+
+.wizardStartForm,
+.ui-corner-all,
+.ui-shadow {
+    background-color: #303030;
+}
+
+.emby-collapsible-button {
+    background: rgb(var(--bg-light));
+    border: none;
+    padding-left: 1.5rem !important;
+    padding-right: 1.5rem !important;
+    transition: none !important;
+    border-radius: 999999px;
+}
+
+.collapseContent {
+    margin-top: 0.5rem;
+    border-radius: 2rem;
+    background-color: rgb(var(--bg-light));
+}
+
+/* #region emby-collapse-has */
+/* .emby-collapse:has(.collapseContent.hide) .emby-collapsible-button { */
+/*     border-radius: 999999px; */
+/*     transition: none !important; */
+/* } */
+
+/* .emby-collapse:has(.collapseContent.expanded) .emby-collapsible-button { */
+/*     border-top-left-radius: 2rem; */
+/*     border-top-right-radius: 2rem; */
+/*     border-bottom-left-radius: 0; */
+/*     border-bottom-right-radius: 0; */
+/* } */
+/* .emby-collapse:has(.collapseContent.expanded) .collapseContent { */
+/*     border-bottom-left-radius: 2rem; */
+/*     border-bottom-right-radius: 2rem; */
+/* } */
+/* #endregion */
+
+.formDialogHeader:not(.formDialogHeader-clear),
+.skinHeader-withBackground {
+    color: var(--text);
+    background: rgb(var(--bg-light));
+}
+
+.skinHeader {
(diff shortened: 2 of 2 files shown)


src/config.json
```diff
@@ -6,27 +6,37 @@
       "name": "Apple TV",
       "id": "appletv",
       "color": "#bcbcbc"
-    }, {
+    },
```
jakehamilton, line 9: Looks like this got hit by my formatter. I can revert these formatting changes or we can leave it as-is :)

nielsvanvelzen:
This looks great, I'm a big fan of Nord (I use themes based on the Nord colors for some of my apps!). I have some feedback which is obviously based on my personal taste. Feel free to ignore it :-)

- There's too much rounding in a bunch of places, like the buttons/popups/menus. I think decreasing the radius size might look better
- The background on the cards looks bad (I prefer the transparency behind the titles)
- The top header should not be floating

Another discussion point for the people that actively maintain jellyfin-web; do we want the "official" themes to only be recolors (like most of them pretty much are right now) or do we also want themes that change the layouts, like this one? I feel like the latter might be hard to maintain in the long term.

thornbill:
I'm a bit behind on reviews, so I haven't had a chance to really look over the code yet, but I completely agree the screenshots look pretty awesome! :purple_heart: 

Niels does bring up a good point about the layout changes... thinking about the future of the project and (hopefully) moving towards using a component library like mui, I believe that type of change would probably not be supported. Considering that combined with the maintenance concerns raised by Niels, it may be best to exclude those.

src/config.json
thornbill: Please do revert the formatting changes. This file can cause some conflicts when upgrading Jellyfin, so it would be best to keep the change set smaller.

src/themes/arctic/theme.css
```diff
@@ -0,0 +1,1264 @@
...
+
+    .skinHeader .headerTabs .emby-tabs {
+        padding: 0 !important;
+    }
+}
+
+html {
+    color-scheme: dark;
+    background: rgb(var(--bg));
+    font-family: Nunito, Noto Sans, Noto Sans HK, Noto Sans JP, Noto Sans KR,
```
batkiz, line 188: Very nice theme! However, I found some problems while using it. Some same characters have different glyphs in "CJK", such as ["台" or "骨"](https://fonts.google.com/?preview.text=%E5%8F%B0%E9%AA%A8&preview.text_type=custom&query=Noto+Sans&category=Sans+Serif), so if we specify the fallback order of Noto Sans family here, it may make some characters look weird.

I'm not good at CSS, but I'm wondering could this problem be solved.

jakehamilton:
@nielsvanvelzen and @thornbill thanks for taking the time to give feedback & review!

I think you make some very good points and I was also wondering about layout or even just large changes to styles being merged in as a core theme. As was mentioned, layout changes would likely cause problems if things shift in the future. Having to check a growing number of themes when making any change to the code would get tiresome very quickly. While I do like a lot of the changes I've done in this theme, _many_ are subjective and I can understand not wanting to have diverging layouts between themes. At least from my perspective it then makes sense for themes to be purely recoloring (which fits nicely into Mui later if we were to use its theme support).

Unless we're willing to allow layout modifications into core themes, I think this PR may be unusable in its current form.

---

However, if we're simplifying things to the level of pure recolorings, would it make sense to have a "template" theme with variables that we can tweak to allow us to quickly build out new colorschemes? I'd still love to have a Nord theme available out of the box and I'm sure others would find Dracula, Catpuccin, etc a fun addition.

src/themes/arctic/theme.css
jakehamilton: Ah, yeah this is my bad as I didn't test with non-english fonts. I had thought about it but was going to leave it for later depending on whether this PR would proceed or not. I added Nunito here purely because I love the font, but it can totally be removed here to return to Noto Sans.

I'm not sure if the CJK variants include other glyphs that would overlap with latin characters, but if they don't then you may be right that the order could be swapped around.

VladTheSheep:
Hey, switched to this theme as soon as I found it! It looks great! 

Though enabling backdrops turns the background almost completely white. Digging through the CSS, it seems like this is intentional? 

```
.backgroundContainer.withBackdrop {
    background: -webkit-gradient(
        linear,
        left top,
        left bottom,
        from(rgba(192, 212, 222, 0.94)),
        color-stop(rgba(235, 250, 254, 0.94)),
        color-stop(rgba(227, 220, 212, 0.94)),
        color-stop(rgba(206, 214, 216, 0.94)),
        to(rgba(192, 211, 218, 0.94))
    );
    background: -webkit-linear-gradient(
        top,
        rgba(192, 212, 222, 0.94),
        rgba(235, 250, 254, 0.94),
        rgba(227, 220, 212, 0.94),
        rgba(206, 214, 216, 0.94),
        rgba(192, 211, 218, 0.94)
    );
    background: -o-linear-gradient(
        top,
        rgba(192, 212, 222, 0.94),
        rgba(235, 250, 254, 0.94),
        rgba(227, 220, 212, 0.94),
        rgba(206, 214, 216, 0.94),
        rgba(192, 211, 218, 0.94)
    );
    background: linear-gradient(
        to bottom,
        rgba(192, 212, 222, 0.94),
        rgba(235, 250, 254, 0.94),
        rgba(227, 220, 212, 0.94),
        rgba(206, 214, 216, 0.94),
        rgba(192, 211, 218, 0.94)
    );
}
```

It makes the text above items hard to read, like "Continue Watching" and it's a bit bright for my eyes in general hehe.

Screenshot from 2022-12-23 14-30-09
Screenshot from 2022-12-23 14-30-25

Looks much better with the darker background IMO. Would be nice to have backdrops though! I think it'd be perfect if the gradient was adjusted to match the color of the background, like the last image.

Screenshot from 2022-12-23 14-40-39

Changed all the RGB values to 46, 52, 64 for the gradient. Much easier on the eyes if you ask me :)

Just my 5 cent. Really liking this theme!

jakehamilton:
@VladTheSheep thank you for reporting this! I originally took the stylesheet from one of the other themes (a light one). I must have missed that part which explains why the colors were all wrong.

I've been using the theme for a few days now and I'm finding things that I want to change/fix as well. Still some work to do!

thornbill:
> However, if we're simplifying things to the level of pure recolorings, would it make sense to have a "template" theme with variables that we can tweak to allow us to quickly build out new colorschemes?

This is absolutely something I have been wanting to investigate to reduce the overhead of doing some basic customization. (I even have a card on my [roadmap](https://trello.com/c/xtyCAsNy/15-use-css-variables-in-themes-for-easy-customization) about this :smile:) It doesn't necessarily have to be _just_ recoloring... other existing themes change things like border radius, just no major layout changes.

Ritchie1108:
maybe you can submit it here
https://github.com/awesome-jellyfin/awesome-jellyfin/blob/main/THEMES.md

1hitsong:
Tested in Waterfox on Windows 11

**Live TV / Guide**

The dates are partially cut off and the dark text seems out of place
image

**Live TV / Channel Filter dropdown**

Text overlaps icon
image

1hitsong (changes requested):
2 display issues found

thornbill:
While this is a beautiful theme, it is probably better suited as a third-party theme based on the discussions above. Built-in themes should be limited to more minor style changes.

PedrelliMath:
It's an interesting theme. I and many other people like to “nordify” everything we can.  @jakehamilton, Is there any way to install this as a third party?

Issue #1752, opened 2020-07-23
Search results are limited to 24.


skyrim663:
**Describe the bug**
When I try to search for a keyword, the number of results is always 24.
Can this limit be manually modified?
At least I can change the upper limit number that suits me.
This will make the search function more convenient. 
For too many media files, limiting the search to 24 may not be good news. 

Can this limit be manually modified?

**System (please complete the following information):**
 - OS:Ubuntu 18.04
 - Virtualization: KVM
 - Clients: Browser
 - Browser: Chrome 
 - Jellyfin Version: 10.6.0
 - Playback:
 - Installed Plugins: none
 - Reverse Proxy:none
 - Base URL: none
 - Networking: [e.g. Host, Bridge/NAT]
 - Storage: [e.g. local, NFS, cloud]

**To Reproduce**

1. search for the keyword
2. Only 24 search results

**Expected behavior**

**Logs**
no error

AlfHou:
Thanks. Where is the API endpoint on the server? I think it would make sense to have "infinite scroll" i.e. query the backend for more result every 24th item

MrTimscampi:
> Thanks. Where is the API endpoint on the server? I think it would make sense to implement "infinite scroll" i.e. query the backend for more result every 24th item

We need a rewrite of `emby-itemscontainer.js` to have infinite scrolling. It currently doesn't support it. If you'd be up for doing that, we're all for it, as it's bring proper infinite scrolling to all the library pages.

hawken93:
IMO there needs to be some sort of pagination because 24 is probably a good number for what you can typically display on a page, but continuously fetching more results as the user scrolls or something would be better

SteveDinn:
This bug is critical! There are now 25 James Bond movies :)

StraightCurlyCurves:
> This bug is critical! There are now 25 James Bond movies :)

This is the reason why I landed here! (26 with Never say Never Again) xD

happyguy1:
ditto on the infinite search - or at very least a larger number - for instance 1024

...there are hundreds of movies with the word (let alone string) "big" in the name

SteveDinn:
The same goes for browsing movies by their first letter: I have tons of movies that start with 'T', but I can only see the first 50 or so.

happyguy1:
@SteveDinn - not sure what your UI experience is, but for instance, i have 330 movies that start with S, and the first 100 are displayed on the first page, with the rest displayed on successive pages
 
[...]

PR #4213, opened 2022-12-16
Increasing limit of returned results from searches

bradbeattie:
**Changes**
In the context of https://github.com/jellyfin/jellyfin/pull/8842, many more than 24 items might be relevant. This PR increases the number of returned results per category to accommodate this (e.g. searching for movies with the tag `christmas`).

**Issues**
While this isn't a complete solution to https://github.com/jellyfin/jellyfin-web/issues/1752 (which would likely require infinite horizontal scroll), it's a sufficient short-term measure to relieve the pressure some users might feel.

### DIFF
--- a/src/components/search/SearchResults.tsx
+++ b/src/components/search/SearchResults.tsx
@@ -49,7 +49,7 @@ const SearchResults: FunctionComponent<SearchResultsProps> = ({ serverId = windo
     const getDefaultParameters = useCallback(() => ({
         ParentId: parentId,
         searchTerm: query,
-        Limit: 24,
+        Limit: 100,
         Fields: 'PrimaryImageAspectRatio,CanDelete,BasicSyncInfo,MediaSourceCount',
         Recursive: true,
         EnableTotalRecordCount: false,


lecelmakerman:
For non infinite scroll horizontal, I work whit this:
General - Custom CSS

`.itemsContainer, .vertical-list {
    display: flex;
    flex-wrap: wrap;
}`

bradbeattie:
Working around the PR approval workflow, I can get the effect of this with the following addition to my Dockerfile:

```
RUN sed -i 's/,Limit:24,/,Limit:240,/g' /jellyfin/jellyfin-web/search.*.chunk.js
```

thornbill:
This is not really a viable solution to the problem imho.

There can be 16 different search queries made when searching to search for different item types which at a worst case would mean that 16 x 240 (3,840) results would need to be rendered on the screen. That is just not feasible. Also I doubt anyone wants to horizontal scroll through 240 results.

What we really need is a "view more" option for each section that goes to a paginated results screen for only that item type.

bradbeattie:
Agreed that the worst case is near 4000 entries. Realistically though, I doubt anywhere around this number would ever be displayed. That said, I'm wholeheartedly on board with a paginated or infinite scroll approach. But given the lack of progress on https://github.com/jellyfin/jellyfin-web/issues/1752, I figured a quick workaround would be worthwhile.

240 might be overkill, but ~50-100 might still prove useful especially in the context of tag searching.

fhumayun:
Looking for heroes to resolve this merge conflict 😌

bradbeattie:
Merge conflict resolved. Adjusted the limit from 24 to 240 to 100 as an attempt to compromise on the blocking disagreement.

thornbill (approved):
Really I think the search page should be rewritten... firing off so many requests is clearly not working well, so we should make a single request and handle separating them by type client side.

This seems like a fine compromise in the meantime though. :slightly_smiling_face:

bbeattie-phxlabs:
I'd argue that search should be pushed off to something like https://github.com/meilisearch/meilisearch to help address issues of typo correction, stemming, filtering, etc. But that's a much bigger lift than this change.

Additionally, it looks like the PR approval needs an additional push to move through.

bradbeattie:
I'd argue that search should be pushed off to something like https://github.com/meilisearch/meilisearch to help address issues of typo correction, stemming, filtering, etc. But that's a much bigger lift than this change.

Additionally, it looks like the PR approval needs an additional push to move through.

PR #4214, opened 2022-12-16
Fix sortby Releasedate (Release 10.8.8)

lomion0815:
It found it misleading that when sorting by "Release Date" in fact the programs get sorted by "Production Date" first and as a second criteria sorted by primieredate. Since my "Production Date" values where not correct (they are not set by the TVHeadend plugin) I got an absolute strange ordering.

**Changes**
Simply removed ProductionYear from sorty query since the purpose was not clear to me.

### DIFF
--- a/src/controllers/list.js
+++ b/src/controllers/list.js
@@ -1041,7 +1041,7 @@ class ItemsView {
 
         sortBy.push({
             name: globalize.translate('ReleaseDate'),
-            value: 'ProductionYear,PremiereDate,SortName'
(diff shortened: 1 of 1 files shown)


thornbill:
This should target the master branch. This type of behavioral change could cause a regression (or perceived regression) for some users, so we should not backport it to 10.8.

thornbill:
Closing since this is targeting the wrong branch. Feel free to open one that targets master.

PR #4219, opened 2022-12-20
Trim whitespaces when entering/modifying usernames

Jxiced:
**Changes**
Added `trim()` to the end of `value` on `form.querySelector('#txtUsername')` and `view.querySelector('#txtManualName')` to prevent whitespaces from appearing at the start or end of usernames.
This is implemented on the login page, the first-time-setup wizard, and when editing or adding a user.

**Issues**
Fixes [#8897](https://github.com/jellyfin/jellyfin/issues/8897)

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -57,6 +57,7 @@
  - [Meet Pandya](https://github.com/meet-k-pandya)
  - [Peter Spenler](https://github.com/peterspenler)
  - [Vankerkom](https://github.com/vankerkom)
+ - [Jxiced](https://github.com/jxiced)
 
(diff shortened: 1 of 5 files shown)


thornbill:
This will break any existing logins that have leaking out trailing whitespace. It will probably need a server side change to check for possibly migrate existing user names.

Jxiced:
> Fixes [#8897](https://github.com/jellyfin/jellyfin/issues/8897)

Opened a second request, [#4224](https://github.com/jellyfin/jellyfin-web/pull/4224) maybe something similar to that implementation might work?

cvium:
See my comment on your other PR

Issue #2483, opened 2021-03-06
Cannot send request when too many songs are selected


Ainto:
**Describe the bug**

The request cannot be sent when too many songs are selected in the playlist.
For example: save all songs in the current play queue to a new playlist.

**System (please complete the following information):**
 - OS: QNAP
 - Virtualization: Docker
 - Clients: Browser
 - Browser: Microsoft Edge 88.0.680.1
 - Jellyfin Version: 10.6.3
 - Base URL: none

**To Reproduce**

1. Add some music to the playback queue
2. Click "save icon" (add to playlist) and click "add"

**Expected behavior**

> In the browser devtools network

Request URL: http://192.168.1.xxx:8096/Playlists?Name=XXX&Ids=[Very long]
Request Method: POST
Status Code: 414 URI Too Long

Zagitta:
This also seems to be an issue when using syncplay and attempting to play an episode as 
https://jelly.my.domain/Users/[user uuid]/Items?Ids=[very long] results in a http 414 error

oddstr13:
This issue is also present when attempting to play a series episode from a show with many episodes per season.

EliseRemise:
Thanks Bond-009 this is it. I see it has been here fore a while. Haven't tried Emby but these problems are on there forum too. Altough they claim to have increased the number off tracks.

EliseRemise:
Just saw the fix is waiting on approval: https://github.com/jellyfin/jellyfin-web/pull/4221

dann-merlin:
Please note that the PR #4221 will fix the usage of the `/Users/<hex userid>/Items?...` endpoint, but not the original issue, as that is another endpoint that would require another fix itself. As I'm not affected by the playlist issue, I did not bother to fix this and am also not planning to (I don't have a lot of fun with typescript, sorry ^^). I'm sure however that the fix should be straight forward for anyone doing it themselves as it would pretty much be the same as my PR just for another endpoint.

PR #4221, opened 2022-12-21
Fix too long URLs on getItems() when too many ids are provided

dann-merlin:
This commit fixes bugs caused by calls to the API endpoint `GET /Users/{userId}/Items?Ids=...`.
When too many ids are provided the URL gets too long and the server returns a HTTP Error 414.

**Testing**
I was not able to test all changes, because of a lack of hardware (Chromecast) and limitations of my library (I don't store any music on jellyfin).
I only tested SyncPlay episode streaming (for a Series with many hundreds of episodes), which was successful.
Further testing would be appreciated.

**Changes**
This commit introduces a wrapper for the `getItems` function of the `ApiClient` in `src/utils/jellyfin-apiclient/getItemsHelper.ts`.
This wrapper splits requests with more than `idsPerItemRequestLimit = 25` ids into multiple requests (with max. 25 requests each). 
The results are then merged and returned as if only one request was sent.
Requests with less than `idsPerItemRequestLimit` ids are passed through.
The wrapper function is used in every case where multiple ids might be provided.

**Issues**
Should fix jellyfin/jellyfin#7350 .

Related to #2483.
It does not solve the issue completely, as the example provided in the issue description is not using the same endpoint. A fix for creating playlists with lots of entries would require a different fix.
However this commit should fix playback, which is also mentioned in this issue.


dann-merlin:
Should I maybe strip some of those changes that have not been issues yet, like the chromecast thing, so this can be merged more easily?

src/utils/jellyfin-apiclient/getItemsHelper.ts
```diff
@@ -0,0 +1,63 @@
...
+        merged.TotalRecordCount += result.TotalRecordCount;
+        merged.StartIndex = Math.min(merged.StartIndex, result.StartIndex);
+    }
+    return merged;
+}
+
+export function getItems(apiClient: ApiClient, userId: string, options?: any):
+    Promise<BaseItemDtoQueryResult> {
+    if (options.Ids === undefined ||
+        options.Ids.split(',').ength <= idsPerItemRequestLimit) {
```
thornbill, line 57: Looks like you have a typo here.

```suggestion
        options.Ids.split(',').length <= idsPerItemRequestLimit) {
```

src/utils/jellyfin-apiclient/getItemsHelper.ts
```diff
@@ -0,0 +1,63 @@
+import type {BaseItemDtoQueryResult} from '@jellyfin/sdk/lib/generated-client';
```
thornbill, line 1: I think this file can just be named `getItems.ts`... "Helper" seems a bit redundant. :slightly_smiling_face:

src/utils/jellyfin-apiclient/getItemsHelper.ts
```diff
@@ -0,0 +1,63 @@
...
+    merged.StartIndex = 0;
+
+    for (const result of results) {
+        if (result.Items == null) {
+            console.log(`Retrieved Items array is invalid: ${result.Items}`);
+            continue;
+        }
+        if (result.TotalRecordCount == null) {
+            console.log(`Retrieved TotalRecordCount is invalid: ${
+                result.TotalRecordCount}`);
```
thornbill, line 39: This is a weird line break. I also prefer including the filename in logs for example:

```suggestion
            console.log('[getItems] Retrieved TotalRecordCount is invalid', result.TotalRecordCount);
```

src/utils/jellyfin-apiclient/getItemsHelper.ts
```diff
@@ -0,0 +1,63 @@
...
+    const results = [];
+
+    let nextI;
+    for (let i = 0; i < ids.length && i < options.Limit; i = nextI) {
+        nextI = i + idsPerItemRequestLimit;
+        if (nextI > options.Limit) {
+            nextI = options.Limit;
+        }
+        const idsSlice = ids.slice(i, nextI);
+        console.log(idsSlice);
```
thornbill, line 18: Was this log just included for debugging?

thornbill (changes requested):

src/utils/jellyfin-apiclient/getItemsHelper.ts
dann-merlin: You are correct, which confuses me a lot, because I am currently using this fix on my instance and it does not break O.o
Committing your suggestion...

src/utils/jellyfin-apiclient/getItemsHelper.ts
dann-merlin: That's fine with me! I'll rename it. Is the quoted line just to be able to make a comment? I am confused about that.

src/utils/jellyfin-apiclient/getItemsHelper.ts
dann-merlin: Ooops, yes it was. I removed it now!

dann-merlin:
Thanks for the review already! I appreciate it a lot!

thornbill:
I would leave the Chromecast change in... I'm pretty sure it has came up as being broken at some point also.

dann-merlin:
Is there anything else that needs to be addressed? GitHub says there's still 1 change requested, but I don't see anything.

Shadowghost:
This will only go away if @thornbill approves the PR.

src/utils/jellyfin-apiclient/getItems.ts
```diff
@@ -0,0 +1,61 @@
+import type {BaseItemDtoQueryResult} from '@jellyfin/sdk/lib/generated-client';
+import {ApiClient} from 'jellyfin-apiclient';
+
+const idsPerItemRequestLimit = 25;
+
+function getItemsSplit(apiClient: ApiClient, userId: string, options: any) {
```
thornbill, line 6: Instead of using `any`, adding a type in this file that covers the usage here should be preferred. (The end goal of course would be to use the TS SDK where everything is already typed, but it's not compatible here unfortunately)

src/utils/jellyfin-apiclient/getItems.ts
```diff
@@ -0,0 +1,61 @@
...
+
+function mergeResults(results: BaseItemDtoQueryResult[]) {
+    const merged: BaseItemDtoQueryResult = {};
+    merged.Items = [];
+    merged.TotalRecordCount = 0;
+    merged.StartIndex = 0;
+
+    for (const result of results) {
+        if (result.Items == null) {
+            console.log(`Retrieved Items array is invalid: ${result.Items}`);
```
thornbill, line 33:
```suggestion
            console.log('[getItems] Retrieved Items array is invalid', result.Items);
```

src/utils/jellyfin-apiclient/getItems.ts
```diff
@@ -0,0 +1,61 @@
...
+            continue;
+        }
+        if (result.TotalRecordCount == null) {
+            console.log('[getItems] Retrieved TotalRecordCount is invalid', result.TotalRecordCount);
+            continue;
+        }
+        if (result.StartIndex == null) {
+            console.log(
+                `Retrieved StartIndex is invalid: ${result.StartIndex}`);
```
thornbill, line 42:
```suggestion
            console.log('[getItems] Retrieved StartIndex is invalid', result.StartIndex);
```

src/utils/jellyfin-apiclient/getItems.ts
```diff
@@ -0,0 +1,61 @@
...
+        }
+        merged.Items = merged.Items.concat(result.Items);
+        merged.TotalRecordCount += result.TotalRecordCount;
+        merged.StartIndex = Math.min(merged.StartIndex, result.StartIndex);
+    }
+    return merged;
+}
+
+export function getItems(apiClient: ApiClient, userId: string, options?: any):
+    Promise<BaseItemDtoQueryResult> {
```
thornbill, line 53: It looks like the return type can be inferred here.

```suggestion
export function getItems(apiClient: ApiClient, userId: string, options?: any) {
```

src/utils/jellyfin-apiclient/getItems.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (options.Ids === undefined ||
+        options.Ids.split(',').length <= idsPerItemRequestLimit) {
```
thornbill, line 55:
```suggestion
    if (options.Ids?.split(',').length <= idsPerItemRequestLimit) {
```

thornbill (changes requested):

src/plugins/syncPlay/core/Helper.js
```diff
@@ -88,7 +89,9 @@ export function getItemsForPlayback(apiClient, query) {
         query.EnableTotalRecordCount = false;
         query.CollapseBoxSetItems = false;
 
-        return apiClient.getItems(apiClient.getCurrentUserId(), query);
+        // call getItems from jellyfin-apiclient/getItems.ts instead of apiClient.getItems()
+        // to split up into multiple requests if necessary (URL might get too long)
```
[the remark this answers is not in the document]
dann-merlin: Done!

src/utils/jellyfin-apiclient/getItems.ts
dann-merlin: Is there documentation somewhere on what variables can be defined in options? Because how options looks like heavily depends on the caller (it varies a lot) and is often built through many function calls. I don't think I have a good enough understanding of the variable and how the object could look like. This is why I use this `optionsTemplate` and only change the `Ids` field.

dann-merlin: Actually I just looked it up in the generated docs of the typescript sdk and if I'm not mistaken it is also `any` there... (`docs/classes/generated_client.ItemsApi.md`)

thornbill: Here is what the SDK uses: https://typescript-sdk.jellyfin.org/interfaces/generated_client.ItemsApiGetItemsRequest.html (note the capitalization is camelCase instead of PascalCase which is why I said it wouldn't be compatible)

Since this is just a build time check you should only have to define the properties used in this file though...

```ts
interface GetItemsRequest {
    Ids?: string
    Limit?: number
}
```

Unless you would _want_ to add the whole thing to the (apiclient type definitions file)[https://github.com/jellyfin/jellyfin-web/blob/master/src/apiclient.d.ts].

dann-merlin: I see! I could only implement the necessary parts as an interface!
I'm new to typescript so I did not think of that. Thanks!

I implemented this now and am looking forward to your feedback!

dann-merlin: Ooops, yes. Not it's pushed ^^

src/utils/jellyfin-apiclient/getItems.ts
```diff
@@ -0,0 +1,74 @@
+import type { BaseItemDtoQueryResult } from '@jellyfin/sdk/lib/generated-client';
+import { ApiClient } from 'jellyfin-apiclient';
+
+interface GetItemsRequest {
+    Ids?: string;
+    Limit?: number;
+}
+
+const ITEMS_PER_REQUEST_LIMIT = 25;
```
thornbill, line 9: Based on a maximum url length around 2000 characters, I think we can pretty safely bump this up to `40`. In my testing that put the url length around 1500 characters which should leave plenty of headroom for variable domain / base urls and differing parameter sets.

src/utils/jellyfin-apiclient/getItems.ts
```diff
@@ -0,0 +1,74 @@
...
+
+function getItemsSplit(apiClient: ApiClient, userId: string, options: GetItemsRequest) {
+    const optionsTemplate = { ...options };
+    const ids = options.Ids!.split(',');
```
thornbill, line 13:
```suggestion
    const ids = options.Ids?.split(',') || [];
```

thornbill (changes requested):

thornbill (approved):

src/utils/jellyfin-apiclient/getItems.ts
dann-merlin: Good point. I'm not sure what the smallest default url length of the widely used webservers is. According to a quick google search it should be at least 4kb, so we might be able to bump it even higher, but I don't think the performance issues caused by too many requests should be too big even with "only" 40. I have the original version of this PR in use for a while now with a season of around 1000 episodes and it was never an issue (Syncplay).

I bumped it to 40 now.

dann-merlin:
Happy to see it merged!
Cheers!

PR #4223, opened 2022-12-22
Stop sending unnecessary `Sessions/Playing/Progress` requests to the backend

edvwib:
Currently the client keeps sending these requests on an interval of 10 seconds even when there's no progress being made.


nielsvanvelzen:
I'm quite sure we still need these updates so the server knows the media session is still active and stuff like remote control / sessions on the dashboard still work.

edvwib:
That makes sense
Saw this while messing around, but didn't think that far 😅

PR #4224, opened 2022-12-23
Migrate Usernames to Remove Whitespaces

Jxiced:
**Changes**
Added trim() to the end of value on form.querySelector('#txtUsername') and view.querySelector('#txtManualName') to prevent whitespaces from appearing at the start or end of usernames.
This is implemented on the login page, the first-time-setup wizard, and when editing or adding a user.
Created dialogs to allow the user to migrate to a new username if one is already taken, otherwise whitespaces will just be removed.

**Issues**
Addresses [#8897](https://github.com/jellyfin/jellyfin/issues/8897)


Jxiced:
First time messing around with JavaScript, so open to others cleaning up and using better practise.

cvium:
It should never be the client's responsibility. The server should trim whitespace.

PR #4225, opened 2022-12-23
Fixed scss problems in item detail pages

### LINKED ISSUE #4197: Item details page layout is broken on mobile in master
**Describe The Bug**
The margins/padding for the content of the item details page seem incorrect on mobile causing elements on the page to overlap or not be visible.

**Steps To Reproduce**
1. Open and unstable/master instance on a mobile device
2. Go to the item details screen

**Expected Behavior**
Elements on the page should not be overlapping

**Logs**
N/A

**Screenshots**
| master | 10.8.8 |
|---|---|
| FECF8516-3198-48CF-9095-28A994BCA0BE | 5FD9014D-75D2-4CAE-8C87-F1DFAA783163 |

 - Platform: Mobile
 - Browser: All
 - Jellyfin Version: Master

**Additional Context**
Probably a regression introduced by https://github.com/jellyfin/jellyfin-web/pull/3743

hadicharara:
**Changes**
Problems stem from the fact that `.layout-mobile` should be applied before direction selectors. Reorganized selectors to fix the issue.

**Issues**
Fixes #4197

### DIFF
--- a/src/assets/css/librarybrowser.scss
+++ b/src/assets/css/librarybrowser.scss
@@ -807,22 +807,16 @@
     .layout-tv & {
         display: block;
     }
-}
 
-.layout-desktop [dir="rtl"] .detailPagePrimaryContainer {
-    padding-right: 32.45vw;
-}
+    .layout-desktop [dir="rtl"] &,
+    .layout-tv [dir="rtl"] & {
+        padding-right: 32.45vw;
+    }
 
-.layout-desktop [dir="ltr"] .detailPagePrimaryContainer {
-    padding-left: 32.45vw;
-}
-
-.layout-tv [dir="rtl"] .detailPagePrimaryContainer {
-    padding-right: 32.45vw;
-}
-
-.layout-tv [dir="ltr"] .detailPagePrimaryContainer {
-    padding-left: 32.45vw;
+    .layout-desktop [dir="ltr"] &,
+    .layout-tv [dir="ltr"] & {
+        padding-left: 32.45vw;
+    }
 }
 
 .layout-desktop .detailRibbon {
@@ -839,20 +833,20 @@
     min-width: 0;
     max-width: 100%;
     flex: 1 0 0;
-}
 
-.layout-mobile .infoWrapper {
-    @media all and (max-width: 32em) {
-        position: relative;
+    .layout-mobile & {
+        @media all and (max-width: 32em) {
+            position: relative;
+        }
     }
-}
 
-.layout-mobile [dir="ltr"] .infoWrapper {
-    @include header-poster-padding;
-}
+    .layout-mobile [dir="ltr"] & {
+        @include header-poster-padding;
+    }
 
-.layout-mobile [dir="rtl"] .infoWrapper {
-    @include header-poster-padding-rtl;
+    .layout-mobile [dir="rtl"] & {
+        @include header-poster-padding-rtl;
+    }
 }
 
 .infoText {
--- a/src/assets/css/librarybrowser.scss
+++ b/src/assets/css/librarybrowser.scss
@@ -610,11 +610,6 @@
         }
     }
 
-    .layout-mobile & {
-        padding-left: 5% !important;
-        padding-right: 5% !important;
-    }
-
     [dir="ltr"] & {
         padding-left: 32.45vw;
         padding-right: 2%;
@@ -623,6 +618,11 @@
     [dir="rtl"] & {
         padding-right: 32.45vw;
         padding-left: 2%;
+    }
+
+    .layout-mobile & {
+        padding-left: 5%;
+        padding-right: 5%;
     }
 }
 
--- a/src/assets/css/librarybrowser.scss
+++ b/src/assets/css/librarybrowser.scss
@@ -739,7 +739,7 @@
 .layout-mobile .mainDetailButtons {
     align-items: center;
     justify-content: center;
-    text-align: center !important;
+    text-align: center;
 }
 
 .layout-mobile .mainDetailButtons {
@@ -847,14 +847,6 @@
 .infoText {
     min-width: 0;
     max-width: 100%;
-
-    [dir="ltr"] & {
-        text-align: left;
-    }
-
-    [dir="rtl"] & {
-        text-align: right;
-    }
 }
 
 .detailPageSecondaryContainer {
--- a/src/assets/css/librarybrowser.scss
+++ b/src/assets/css/librarybrowser.scss
@@ -669,11 +669,6 @@
     margin: 1em 0;
 }
 
-.detailButton,
-.mainDetailButtons {
-    display: flex;
-}
-
 .itemName {
     margin: 0.5em 0;
     font-weight: 600;
--- a/src/assets/css/librarybrowser.scss
+++ b/src/assets/css/librarybrowser.scss
@@ -610,6 +610,11 @@
         }
     }
 
+    .layout-mobile & {
+        padding-left: 5% !important;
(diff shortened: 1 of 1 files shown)


dmitrylyzo:
Does moving this part before `.layout-mobile & {` work (no stylelint warnings/errors)?
https://github.com/jellyfin/jellyfin-web/blob/c730a79f5d8ce125da2588c4be27713516ebafc7/src/assets/css/librarybrowser.scss#L614-L622

Here is how it was before RTL mod:
https://github.com/jellyfin/jellyfin-web/blob/941fe35103ad5ecf2359da82eb9a59fd2656c768/src/assets/css/librarybrowser.scss#L473-L479

hadicharara:
When I do that, the result in the browser is the following:

```
[dir=rtl] .detailPageContent {
    padding-left: 2%;
    padding-right: 32.45vw;
}

.layout-mobile .detailPageContent {
    ~~padding-left: 5%;~~
    ~~padding-right: 5%;~~
}
```
The dir selectors get prioritized. I did it this way to avoid using `!important` in `.layout-mobile &` since it's bad practice, though I could.

hadicharara:
I tried to put it after the dir selectors, and it works.

dmitrylyzo:
> I tried to put it after the dir selectors, and it works. As such:
> 
> ```
>     [dir="ltr"] & {
>         padding-left: 32.45vw;
>         padding-right: 2%;
>     }
> 
>     [dir="rtl"] & {
>         padding-right: 32.45vw;
>         padding-left: 2%;
>     }
> 
>     .layout-mobile & {
>         padding-left: 5%;
>         padding-right: 5%;
>     }
> ```

Yes, that's exactly what I meant.

src/assets/css/librarybrowser.scss
```diff
@@ -840,20 +840,20 @@
...
+    @include header-poster-padding;
+}
 
-        @media all and (max-width: 32em) {
-            position: relative;
-        }
-    }
+.layout-mobile [dir="rtl"] .infoWrapper {
+    @include header-poster-padding-rtl;
 }
```
dmitrylyzo, line 857: This can be rewritten as
```scss
.infoWrapper {
    min-width: 0;
    max-width: 100%;
    flex: 1 0 0;

    .layout-mobile [dir="ltr"] & {
        @include header-poster-padding;
    }

    .layout-mobile [dir="rtl"] & {
        @include header-poster-padding-rtl;
    }

    .layout-mobile & {
        @media all and (max-width: 32em) {
            position: relative;
        }
    }
}
```

src/assets/css/librarybrowser.scss
```diff
@@ -759,20 +763,16 @@
...
+    @include header-poster-padding-rtl;
 
-        @media all and (max-width: 32em) {
-            padding-right: 0 !important;
-        }
+    @media all and (max-width: 32em) {
+        margin-bottom: 0;
+        padding-right: 0;
     }
 }
```
dmitrylyzo, line 777: Try the technique I suggested for `.infoWrapper'.

src/assets/css/librarybrowser.scss
hadicharara: This doesn't work. I have no idea why.

hadicharara:
Unfortunately, the linter complains about the order of the selectors. If we put `.layout-mobile` in its own selector, it becomes fine.

src/assets/css/librarybrowser.scss
hadicharara: Can't do it for this one because the linter complains. There is already another `.mainDetailButtons` up in the file.

src/assets/css/librarybrowser.scss
hadicharara: Done.

src/assets/css/librarybrowser.scss
```diff
@@ -627,6 +622,11 @@
...
 
+.layout-mobile .detailPageContent {
+    padding-left: 5%;
+    padding-right: 5%;
+}
+
```
dmitrylyzo, line 629: Since it works.
```suggestion

        .layout-mobile & {
            padding-left: 5%;
            padding-right: 5%;
        }
    }
}

```

dmitrylyzo (commented):
I believe this also requires LTR/RTL:
https://github.com/jellyfin/jellyfin-web/blob/c730a79f5d8ce125da2588c4be27713516ebafc7/src/assets/css/librarybrowser.scss#L1145-L1147

These lines can be removed (in a separate cleanup commit) - `display` is set in the corresponding blocks:
https://github.com/jellyfin/jellyfin-web/blob/c730a79f5d8ce125da2588c4be27713516ebafc7/src/assets/css/librarybrowser.scss#L673-L676

src/assets/css/librarybrowser.scss
hadicharara: Just implemented your suggestions. The .layout-mobile move wouldn't work without !important. Let me know if you have any other suggestions.

dmitrylyzo: That's becasue you have moved this block before the LTR/RTL paddings. In my suggestion, it goes after them.

hadicharara: Fixed.

src/assets/css/librarybrowser.scss
```diff
@@ -738,18 +733,34 @@
...
+    @media all and (max-width: 62.5em) {
+        margin-left: -0.5em;
+    }
+}
+
+[dir=rtl] .mainDetailButtons {
+    @media all and (max-width: 62.5em) {
+        margin-right: -0.5em;
+    }
+}
```
dmitrylyzo, line 746: Tbh, I don't understand the purpose of these margins.

To avoid duplicating `@media`:
```suggestion
@media all and (max-width: 62.5em) {
    [dir=ltr] .mainDetailButtons {
        margin-left: -0.5em;
    }

    [dir=rtl] .mainDetailButtons {
        margin-right: -0.5em;
    }
}
```

src/assets/css/librarybrowser.scss
```diff
@@ -738,18 +733,34 @@
...
+
 .layout-mobile .parentName,
 .layout-mobile .itemName,
 .layout-mobile .itemMiscInfo,
 .layout-mobile .mainDetailButtons {
     align-items: center;
     justify-content: center;
-    text-align: center;
+    text-align: center !important;
```
dmitrylyzo, line 754: To avoid `important`, you can remove:
https://github.com/jellyfin/jellyfin-web/blob/ac105f4c831f3e9e30ea318d06f69af42a18a6bf/src/assets/css/librarybrowser.scss#L862-L869
`text-align` will be `start` (or just unset), which seems to make it aligned to the correct side.

src/assets/css/librarybrowser.scss
hadicharara: Me neither. I tried applying them but I really don't see a difference.

src/assets/css/librarybrowser.scss
hadicharara: This doesn't work. Text should be centered, else it might be under the thumbnail.

src/assets/css/librarybrowser.scss
hadicharara: I removed them in my latest commit.

src/assets/css/librarybrowser.scss
hadicharara: Did it. LMK if that does what you expected.

dmitrylyzo (approved):
LGTM

PR #4262, opened 2023-01-06
Add setting for ParallelImageEncodingLimit

hannesbraun:
**Changes**

This adds the setting for changing the maximum amount of parallel image encodings. See jellyfin/jellyfin#8783 for more details.

**Issues**

*None*


src/controllers/dashboard/general.html
```diff
@@ -80,6 +80,13 @@ <h2>${HeaderBranding}</h2>
                     </div>
                 </div>
 
+                <div class="verticalSection">
+                    <h2>${HeaderPerformance}</h2>
+                    <div class="inputContainer">
+                        <input is="emby-input" id="txtParallelImageEncodingLimit" label="${LabelParallelImageEncodingLimit}" type="number" pattern="[0-9]*" min="1" step="1" />
```
nielsvanvelzen, line 86: Shouldn't min be set to 0?

> Setting this to 0 (or less) will choose a limit based on your system specs.

hannesbraun: I forgot to update that. Thanks for letting me know. Fixed.

src/controllers/dashboard/general.js
```diff
@@ -36,6 +37,7 @@ import alert from '../../components/alert';
             config.MetadataPath = $('#txtMetadataPath', form).val();
             config.MetadataNetworkPath = $('#txtMetadataNetworkPath', form).val();
             config.QuickConnectAvailable = form.querySelector('#chkQuickConnectAvailable').checked;
+            config.ParallelImageEncodingLimit = parseInt(form.querySelector('#txtParallelImageEncodingLimit').value || '1');
```
thornbill, line 40: A radix should be specified when calling `parseInt` to avoid some browser parsing number strings starting with "0" being parsed as a hexadecimal.

```suggestion
            config.ParallelImageEncodingLimit = parseInt(form.querySelector('#txtParallelImageEncodingLimit').value || '1', 10);
```

src/strings/en-us.json
```diff
@@ -791,6 +792,8 @@
     "LabelOriginalName": "Original name:",
     "LabelOriginalTitle": "Original title:",
     "LabelOverview": "Overview:",
+    "LabelParallelImageEncodingLimit": "Parallel image encoding limit",
+    "LabelParallelImageEncodingLimitHelp": "Maximum amount of image encodings that are allowed to run in parallel. Setting this to 0 (or less) will choose a limit based on your system specs.",
```
thornbill, line 796: If the minimum is 0, then we can probably exclude the "or less" part. :wink:

thornbill (changes requested):

src/controllers/dashboard/general.js
hannesbraun: Fixed. Thanks :)

src/controllers/dashboard/general.js
```diff
@@ -36,6 +37,7 @@ import alert from '../../components/alert';
             config.MetadataPath = $('#txtMetadataPath', form).val();
             config.MetadataNetworkPath = $('#txtMetadataNetworkPath', form).val();
             config.QuickConnectAvailable = form.querySelector('#chkQuickConnectAvailable').checked;
+            config.ParallelImageEncodingLimit = parseInt(form.querySelector('#txtParallelImageEncodingLimit').value || '1', 10);
```
thornbill, line 40: It looks like the default in the server PR is `0`, but here you are defaulting to `1`... this should probably be `0` also.

hannesbraun: Fixed.

thornbill (approved):

PR #4263, opened 2023-01-06
Fix change audio track

dmitrylyzo:
**Changes**
- Take into account the container when traverse device profiles.
When an audio track isn't supported, it doesn't show up in `audioTracks`. Therefore, we need to filter the device profiles properly.

- Detect support for secondary audio consistently.
This should be done in #4240.

- Assume webOS 5+ (2020, Chrome 68+) supports secondary audio like Tizen 5.5+

**Issues**
https://github.com/jellyfin/jellyfin-webos/issues/85
https://github.com/jellyfin/jellyfin-webos/issues/118
https://github.com/jellyfin/jellyfin-tizen/issues/21

### DIFF
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -1316,11 +1316,11 @@ class PlaybackManager {
 
             const profiles = deviceProfile.DirectPlayProfiles || [];
 
-            return profiles.filter(function (p) {
+            return profiles.some(function (p) {
                 return p.Type === 'Video'
                     && includesAny((p.Container || '').toLowerCase(), container)
                     && includesAny((p.AudioCodec || '').toLowerCase(), codec);
-            }).length > 0;
+            });
         }
 
         self.setAudioStreamIndex = function (index, player) {
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -11,6 +11,7 @@ import { appHost } from '../apphost';
 import Screenfull from 'screenfull';
 import ServerConnections from '../ServerConnections';
 import alert from '../alert';
+import { includesAny } from '../../utils/container.ts';
 
 const UNLIMITED_ITEMS = -1;
 
@@ -1306,6 +1307,7 @@ class PlaybackManager {
                 return false;
             }
 
+            const container = mediaSource.Container.toLowerCase();
             const codec = (mediaStream.Codec || '').toLowerCase();
 
             if (!codec) {
@@ -1315,20 +1317,9 @@ class PlaybackManager {
             const profiles = deviceProfile.DirectPlayProfiles || [];
 
             return profiles.filter(function (p) {
-                if (p.Type === 'Video') {
-                    if (!p.AudioCodec) {
-                        return true;
-                    }
-
-                    // This is an exclusion filter
-                    if (p.AudioCodec.indexOf('-') === 0) {
-                        return p.AudioCodec.toLowerCase().indexOf(codec) === -1;
-                    }
-
-                    return p.AudioCodec.toLowerCase().indexOf(codec) !== -1;
-                }
-
-                return false;
+                return p.Type === 'Video'
+                    && includesAny((p.Container || '').toLowerCase(), container)
+                    && includesAny((p.AudioCodec || '').toLowerCase(), codec);
             }).length > 0;
         }
 
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -353,6 +353,23 @@ import browser from './browser';
         return 2;
     }
 
+/**
+ * Checks if the web engine supports secondary audio.
+ * @param {HTMLVideoElement} videoTestElement The video test element
+ * @returns {boolean} _true_ if the web engine supports secondary audio.
+ */
+export function canPlaySecondaryAudio(videoTestElement) {
+    // We rely on HTMLMediaElement.audioTracks
+    // It works in Chrome 79+ with "Experimental Web Platform features" enabled
+    return !!videoTestElement.audioTracks
+        // It doesn't work in Firefox 108 even with "media.track.enabled" enabled (it only sees the first audio track)
+        && !browser.firefox
+        // It seems to work on Tizen 5.5+ (2020, Chrome 69+). See https://developer.tizen.org/forums/web-application-development/video-tag-not-work-audiotracks
+        && (browser.tizenVersion >= 5.5 || !browser.tizen)
+        // Assume webOS 5+ (2020, Chrome 68+) supports secondary audio like Tizen 5.5+
+        && (browser.web0sVersion >= 5.0 || !browser.web0sVersion);
+}
+
     export default function (options) {
         options = options || {};
 
@@ -752,13 +769,7 @@ import browser from './browser';
 
         profile.CodecProfiles = [];
 
-        // We rely on HTMLMediaElement.audioTracks
-        // It works in Chrome 79+ with "Experimental Web Platform features" enabled
-        // It doesn't work in Firefox 108 even with "media.track.enabled" enabled (it only sees the first audio track)
-        // It seems to work on Tizen 5.5+ (Chrome 69+). See https://developer.tizen.org/forums/web-application-development/video-tag-not-work-audiotracks
-        const supportsSecondaryAudio = !!videoTestElement.audioTracks
-            && !browser.firefox
-            && (browser.tizenVersion >= 5.5 || !browser.tizen);
+        const supportsSecondaryAudio = canPlaySecondaryAudio(videoTestElement);
 
         const aacCodecProfileConditions = [];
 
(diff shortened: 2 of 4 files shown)


src/components/playback/playbackmanager.js
```diff
@@ -1316,16 +1317,20 @@ class PlaybackManager {
...
+                    if (containers.includes(container)) {
+                        if (!p.AudioCodec) {
+                            return true;
+                        }
 
-                    return p.AudioCodec.toLowerCase().indexOf(codec) !== -1;
+                        // This is an exclusion filter
+                        if (p.AudioCodec.startsWith('-')) {
+                            return !p.AudioCodec.substring(1).toLowerCase().split(',').includes(codec);
+                        }
```
dmitrylyzo, line 1330: Should I also add an exclusive filter to the player?
Or just remove it from everywhere?

src/components/playback/playbackmanager.js
```diff
@@ -1316,16 +1317,20 @@ class PlaybackManager {
 
             return profiles.filter(function (p) {
                 if (p.Type === 'Video') {
-                    if (!p.AudioCodec) {
-                        return true;
-                    }
+                    const containers = (p.Container || '').toLowerCase().split(',');
```
dmitrylyzo, line 1320: How safe is it to test strings directly?
I am currently using arrays and assume that it is joined exactly by `,`.

src/components/playback/playbackmanager.js
```diff
@@ -1306,6 +1306,7 @@ class PlaybackManager {
                 return false;
             }
 
+            const container = mediaSource.Container;
```
dmitrylyzo, line 1309: Can `mediaSource.Container` be combined like `mkv,webm`?
_I currently have just `mkv`._

src/components/playback/playbackmanager.js
dmitrylyzo: Since the server [supports an exclusive list](https://github.com/jellyfin/jellyfin/blob/c795f17fa6ae9e4f68f23cb569a1129de8fd7635/MediaBrowser.Model/Dlna/ContainerProfile.cs#L38-L43), we need to add it for the player as well.

src/components/playback/playbackmanager.js
dmitrylyzo: Inspired by https://github.com/jellyfin/jellyfin/blob/c795f17fa6ae9e4f68f23cb569a1129de8fd7635/MediaBrowser.Model/Dlna/ContainerProfile.cs#L19-L72

src/components/playback/playbackmanager.js
dmitrylyzo: Inspired by https://github.com/jellyfin/jellyfin/blob/c795f17fa6ae9e4f68f23cb569a1129de8fd7635/MediaBrowser.Model/Dlna/ContainerProfile.cs#L19-L72

src/utils/container.ts
```diff
@@ -0,0 +1,42 @@
+/**
+ * Checks if the list includes any value from the search.
+ * @param value The list to search in.
```
dmitrylyzo, line 3:
```suggestion
 * @param list The list to search in.
```

src/utils/container.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (typeof list === 'string') {
+        if (list.startsWith('-')) {
+            negative = true;
+            list = list.substring(1);
+        }
+
+        list = list.split(',');
+    }
+
+    list = list.filter(i => i);
```
thornbill, line 23: I'm not sure what this actually does? :sweat_smile:

src/utils/container.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    if (!list.length) {
+        return true;
+    }
+
+    if (typeof search === 'string') {
+        search = search.split(',');
+    }
+
+    search = search.filter(i => i);
```
thornbill, line 33: Same here.

src/utils/container.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    const holdList = list;
```
thornbill, line 35: Why reassign this?

src/utils/container.ts
```diff
@@ -0,0 +1,42 @@
...
+ * @param list The list to search in.
+ * @param search The values to search.
+ * @returns {boolean} _true_ if the list includes any value from the search.
+ */
+export function includesAny(list: string | string[] | null | undefined, search: string | string[]) {
+    if (!list) {
+        return true;
+    }
+
+    let negative = false;
```
thornbill, line 12: `negative` seems a bit ambiguous... maybe `inverseMatch` or something similar would be more clear? :man_shrugging:

src/scripts/browserDeviceProfile.js
```diff
@@ -345,6 +345,23 @@ import browser from './browser';
         return 2;
     }
 
+/**
+ * Checks if the web engine supports secondary audio.
+ * @param {HTMLVideoElement} videoTestElement The video test element
+ * @returns {boolean} _true_ if the web engine supports secondary audio.
+ */
+export function canSecondaryAudio(videoTestElement) {
```
thornbill, line 353: Maybe this could be renamed to be similar to the other functions?

```suggestion
export function supportsSecondaryAudio(videoTestElement) {
```

thornbill (changes requested):

src/utils/container.ts
dmitrylyzo: Removes empty items

src/utils/container.ts
dmitrylyzo: Lint error in the callback function. Something like `'list' can be 'null'`.

src/scripts/browserDeviceProfile.js
dmitrylyzo: We have variable with the same name in the `browserDeviceProfile` - I was afraid of the shadowing error.

src/utils/container.ts
dmitrylyzo: Hmm, it is not Lint - it is VS Code and TS checker (?) during build: `TS2533: Object is possibly 'null' or 'undefined'.`

src/scripts/browserDeviceProfile.js
thornbill: `canSupportSecondaryAudio`? :sweat_smile:

dmitrylyzo: Renamed to `canPlaySecondaryAudio`

src/utils/container.ts
dmitrylyzo: I used non-null assertion operator and disabled eslint rule.

src/scripts/browserDeviceProfile.js
```diff
@@ -353,6 +353,23 @@ import browser from './browser';
...
+export function canPlaySecondaryAudio(videoTestElement) {
+    // We rely on HTMLMediaElement.audioTracks
+    // It works in Chrome 79+ with "Experimental Web Platform features" enabled
+    // It doesn't work in Firefox 108 even with "media.track.enabled" enabled (it only sees the first audio track)
+    // It seems to work on Tizen 5.5+ (2020, Chrome 69+). See https://developer.tizen.org/forums/web-application-development/video-tag-not-work-audiotracks
+    // Assume webOS 5+ (2020, Chrome 68+) supports secondary audio like Tizen 5.5+
+    return !!videoTestElement.audioTracks
+        && !browser.firefox
+        && (browser.tizenVersion >= 5.5 || !browser.tizen)
+        && (browser.web0sVersion >= 5.0 || !browser.web0s || !window.NativeShell);
```
thornbill, line 370: What is the point of the NativeShell check here?

dmitrylyzo: This means `we are not in the app`.
Since we cannot reliably detect the webOS version in the built-in browser, we simply ignore version checking.

dmitrylyzo: We could do this instead: `browser.web0sVersion >= 5.0 || !browser.web0sVersion`.

What do you think of comments placed like this?
```js
export function canPlaySecondaryAudio(videoTestElement) {
    // We rely on HTMLMediaElement.audioTracks
    // It works in Chrome 79+ with "Experimental Web Platform features" enabled
    return !!videoTestElement.audioTracks
        // It doesn't work in Firefox 108 even with "media.track.enabled" enabled (it only sees the first audio track)
        && !browser.firefox
        // It seems to work on Tizen 5.5+ (2020, Chrome 69+). See https://developer.tizen.org/forums/web-application-development/video-tag-not-work-audiotracks
        && (browser.tizenVersion >= 5.5 || !browser.tizen)
        // Assume webOS 5+ (2020, Chrome 68+) supports secondary audio like Tizen 5.5+
        && (browser.web0sVersion >= 5.0 || !browser.web0sVersion);
}
```

thornbill: Oh breaking the comments down like that seems nice :+1:

thornbill: It looks like in your latest changes the NativeShell check is missing. Was that intentional?

thornbill (approved):

Issue #4265, opened 2023-01-07
[Issue]: Playback setting Audio boost when downmixing cannot be set to 0


ltomes:
https://github.com/jellyfin/jellyfin-web/blob/b0f63e51a2e726b63af50bac41d2917cf65ac53b/src/controllers/dashboard/encodingsettings.html#L257

### Please describe your bug

[Issue]: The playback setting `Audio boost when downmixing` cannot be set to 0 in the UI, the helper text eludes to `1` being a valid value to maintain volume. 

Edit: In retrospect I can see how removing sound at the _transcoding_ level would generally be confusing, and might not be desirable for all of jellyfin's users. I understand if you just want to close this as not an issue.

Location
https://<domain>/web/index.html#!/encodingsettings.html

I expected to be able to enter a value of `0` and save settings. Thanks for your time

### Jellyfin Version

10.8.0

### Environment

```markdown
- OS: Unraid
- Virtualization: Docker/nvidia-docker
- Clients:NR
- Browser:Brave/Chrome/Opera
- FFmpeg Version:NR
- Playback Method:NR
- Hardware Acceleration:Yes
- Plugins:NR
- Reverse Proxy:NR
- Base URL:NR
- Networking:NR
- Storage:NR
```

### Please attach any screenshots here

<img width="639" alt="image" src="https://user-images.githubusercontent.com/4184677/211160006-fa84c75c-d089-485f-af73-bf6fce68b8fd.png">

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #4266, opened 2023-01-07
Audio boost when downmixing should be 0 - 3

ltomes:
This number input should have a range down to `0`

**Changes**
Changed minimum excepted value for downmixing to `0` from `0.5`
Please let me know if this requires changes, this is my first MR here/I just wanted to resolve this issue quick if it doesn't go too deep.

**Issues**

https://github.com/jellyfin/jellyfin-web/issues/4265

### DIFF
--- a/src/controllers/dashboard/encodingsettings.html
+++ b/src/controllers/dashboard/encodingsettings.html
@@ -254,7 +254,7 @@
                     <div class="fieldDescription checkboxFieldDescription">${EnableFallbackFontHelp}</div>
                 </div>
                 <div class="inputContainer">
-                    <input is="emby-input" type="number" id="txtDownMixAudioBoost" pattern="[0-9]*" required="required" min=".5" max="3" step=".1" label="${LabelDownMixAudioScale}" />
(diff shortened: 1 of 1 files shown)


thornbill:
> In retrospect I can see how removing sound at the transcoding level would generally be confusing, and might not be desirable for all of jellyfin's users. I understand if you just want to close this as not an issue.

So setting this value to 0 would effectively remove the sound when downmixing? I'm curious now what your use case for this is. :slightly_smiling_face:

ltomes:
> > In retrospect I can see how removing sound at the transcoding level would generally be confusing, and might not be desirable for all of jellyfin's users. I understand if you just want to close this as not an issue.
> 
> So setting this value to 0 would effectively remove the sound when downmixing? I'm curious now what your use case for this is. 🙂

I don't have a great one, a library of pet videos with poor sound quality that I thought I would  boost the audio out of/didn't want to change the source of, and didn't want to make copies of.

thornbill:
The team discussed this change a bit and the general consensus was we didn't want to enable this option. It mostly seems like it would just create issues when people change it without understanding the impact that it will have.

ltomes:
I agree, I wanted this for a weird local library. but don't see the value for others (obviously I didn't realize that until after submitting the MR though).

Issue #3737, opened 2022-06-29
Login disclaimer not showing on 10.8.1


l4kr:
**Describe The Bug**
Hi! When I change the login disclaimer in server settings it doesn't work even on a clean install.
When tested on 10.7.7 it works perfectly fine. 

**Steps To Reproduce**
1. change login disclaimer

**Expected Behavior**
A disclaimer should show on login page

thornbill:
It seems to be working fine here: https://demo.jellyfin.org/stable

PR #4268, opened 2023-01-08
Rename login disclaimer

dann-merlin:
The HTML element `disclaimerContainer` on the login page is confused by adblockers using the "I don't care about cookies" filters with cookie disclaimers. This results in the disclaimer being blocked when using popular adblocker extensions like uBlock Origin.

**Changes**
This commit renames
`disclaimerContainer` -> `loginDisclaimerContainer`
`disclaimer` -> `loginDisclaimer`
to be more verbose and circumvent that filter.

**Issues**
This was most likely the reason for #3737


dmitrylyzo:
https://demo.jellyfin.org/stable works with uBlock Origin 1.46.0 (Firefox 108 and Chrome 108)

What kind of filter affects this?

dann-merlin:
I thought it was the i-dont-care-about-cookies.txt. Specifically I believe this line to be the culprit (found it in my uBlock Origin filters list):
```
##.disclaimerContainer:not(body):not(html):not(.grid__unit)
```

I might be wrong though. The demo site does not work for me (UngoogledChromium 108 with uBlock Origin 1.40.8 and Librewolf 108 with uBlock Origin 1.46.0).

I tried it with normal Firefox aswell now and there it works (Firefox 108 with uBlock Origin 1.46.0).

I don't know why this is the case tbh.

The list you mentioned also introduces a matching filter:
```
##.disclaimerContainer
```
but I could not find this one in my uBlock Origin filters list.

thornbill:
This seems like it could become almost a game of whack-a-mole with these plugins randomly hiding stuff. Maybe we should go with something like `loginMessage` to avoid the "disclaimer" bit entirely.

Issue #4234, opened 2023-01-02
Web client tries to play nonexistent media when clicking on actor photo


davemsh:
**Describe The Bug**
In all the Jellyfin web clients I've tried, when I click/tap on an actor under a show's Cast & Crew heading, it brings me to the details page for that actor (as expected). When I then clicked the actor's portrait photo, I get a popup dialog with the error message "No player found for the requested media."

**Steps To Reproduce**
1. Go to an actor's details page
2. Click/tap on the actor's main portrait on the left
3. Error is displayed

**Expected Behavior**
I expected it to show a bigger version of that picture.

**Logs**
There are no pertinent log messages.

**Screenshots**
![image](https://user-images.githubusercontent.com/63111368/210193790-97c823aa-6fce-4ac8-9d11-9bab43207977.png)

**System (please complete the following information):**
 - Platform: Windows, Android, iOS
 - Browser: Firefox 108.0.1, official Android app, official iOS app
 - Jellyfin Version: 10.8.8

**Additional Context**
N/A

PR #4272, opened 2023-01-10
Check if it's actually possible to play the item before trying to play it from itemDetails

edvwib:
This will simply not play the item if the playbackManager says it can't.
The issue described the expected behaviour to be to show a bigger version of the image, however I did not find an easy way to do that right now 😅 

This should however be better than displaying the error dialog.

**Issues**
Fixes: #4234

### DIFF
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -1886,11 +1886,13 @@ export default function (view, params) {
 
         const apiClient = getApiClient();
 
-        Promise.all([getPromise(apiClient, pageParams), apiClient.getCurrentUser()]).then(([item, user]) => {
+        return Promise.all([getPromise(apiClient, pageParams), apiClient.getCurrentUser()]).then(([item, user]) => {
             currentItem = item;
             reloadFromItem(instance, page, pageParams, item, user);
+            return item;
         }).catch((error) => {
             console.error('failed to get item or current user: ', error);
+            throw error;
         });
     }
 
@@ -2052,7 +2054,6 @@ export default function (view, params) {
         bindAll(view, '.btnCancelSeriesTimer', 'click', onCancelSeriesTimerClick);
         bindAll(view, '.btnCancelTimer', 'click', onCancelTimerClick);
         bindAll(view, '.btnDownload', 'click', onDownloadClick);
-        view.querySelector('.detailImageContainer').addEventListener('click', onPlayClick);
         view.querySelector('.trackSelections').addEventListener('submit', onTrackSelectionsSubmit);
         view.querySelector('.btnSplitVersions').addEventListener('click', function () {
             splitVersions(self, view, apiClient, params);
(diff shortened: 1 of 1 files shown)


thornbill:
Would it be possible to have this checked before adding the click handler?

edvwib:
Good point. It's possible, but if I understand it correctly we then have to wait for the reload function to finish.

I've updated the PR, the reload function now returns a promise, so that way we can then check if we can play the item, and only then add the eventlistener.

dmitrylyzo (commented):
There is also a pointer cursor that goes from the card CSS.
https://github.com/jellyfin/jellyfin-web/blob/856e0938cc3b0123fd50cc0638ad78501cbb0762/src/components/cardbuilder/card.scss#L23
https://github.com/jellyfin/jellyfin-web/blob/856e0938cc3b0123fd50cc0638ad78501cbb0762/src/components/cardbuilder/card.scss#L210
It would be nice to make it `unset` or `default` if the item isn't supported.

edvwib:
Yeah it would be nice to get rid of the cursor. Of course it's possible, but I haven't figured out a nice way to do it... 😕

thornbill:
Yeah unfortunately a lot of things are pretty hacked together in the older code. Maybe you could just add a class to the element if it should show the pointer cursor?

dmitrylyzo:
It looks like we can open a copy of the actor's details in a new tab using the poster - it is an "anchor" with the correct `href`.
And it is uncontrollable: https://github.com/jellyfin/jellyfin-web/blob/69053a131f1c01b6ce018795ade07fd6adeddb08/src/components/cardbuilder/cardBuilder.js#L1352

edvwib:
That's the page we're already on though, isn't it? The link is valid, but has no effect since it's the same url.

dmitrylyzo:
> That's the page we're already on though, isn't it? The link is valid, but has no effect since it's the same url.

There is no (visual) effect when you click the left mouse button, but clicking the middle mouse button opens a new tab (or open the context menu and click "Open in a new tab").

I don't think we need such feature on the poster. But if we leave it, we should leave the pointer cursor.

thornbill:
I haven't tested it, but I had an idea for a potentially easier approach... could we change this to use a `<span>` when it can't be played? There is a `canPlay`check a few lines below that could be combined with this.

https://github.com/jellyfin/jellyfin-web/blob/master/src/components/cardbuilder/cardBuilder.js#L1459

thornbill:
> I haven't tested it, but I had an idea for a potentially easier approach... could we change this to use a `<span>` when it can't be played? There is a `canPlay`check a few lines below that could be combined with this.
> 
> https://github.com/jellyfin/jellyfin-web/blob/master/src/components/cardbuilder/cardBuilder.js#L1459

I messed with this a bit but couldn't get it to work properly... so I guess the only thing remaining is dmitry's feedback above. :slightly_smiling_face:

thornbill:
Superseded by #4881

PR #4274, opened 2023-01-11
Fix back action on Hisense TV

dmitrylyzo:
**Changes**
Interpret `Backspace` as `Back` on Hisense TV.

**Issues**
For some reason, Hisense TV uses the `Backspace` key code for the `Back` button.

### DIFF
--- a/src/scripts/keyboardNavigation.js
+++ b/src/scripts/keyboardNavigation.js
@@ -3,6 +3,7 @@
  * @module components/input/keyboardnavigation
  */
 
+import browser from './browser';
 import inputManager from './inputManager';
 import layoutManager from '../components/layoutManager';
 import appSettings from './settings/appSettings';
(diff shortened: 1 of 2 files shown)


src/scripts/keyboardNavigation.js
```diff
@@ -120,6 +121,15 @@ export function enable() {
                 inputManager.handleCommand('back');
                 break;
 
+            // HACK: Hisense TV uses Backspace for Back action
+            case 'Backspace':
+                if (browser.tv && browser.hisense) {
```
nielsvanvelzen, line 126: Should this also check if there is no focus on an input like with the arrow keys?

dmitrylyzo: The user said that on-screen keyboard captures pressing `Back`, so we don't receive the click.

But now I am thinking about adding a check on the VIDAA OS, because Hisense could be Android, and it could have different behavior.

Also, I don't think it will work with a keyboard connected to the TV.

nielsvanvelzen (approved):

Issue #4296, opened 2023-01-22
Gamepad support broken on Steam Deck


KingIzzymon:
**Describe The Bug**
No gamepad support when accessing Jellyfin web client via Google Chrome flatpak on Steamdeck using Steam's gamepad template. while in Steam Deck Game Mode

**Steps To Reproduce**
1. Turn on steam deck and load into game mode
2. Open google chrome from steam using the "Gamepad" steam controller template
3. Access the Jellyfin web client
4. Ensure "Gamepad support" is enabled by clicking on your profile (upper right) -> Controller options -> enable gamepad support then reload the web page

**Expected Behavior**
The steamdeck controls should be able to navigate the Jellyfin web client, accessing media libraries, selecting content and controlling the web media player.

**Logs**
No errors

**Screenshots**
Not screenshotable

**System (please complete the following information):**
 - Browser: Google Chrome flatpak (latest as of this posting)
 - Platform: Docker image: jellyfin/jellyfin:latest
 -- Jellyfin Version: Version: 10.8.8
 -- Operating System: Linux
 -- Architecture: X64

dmitrylyzo:
Try changing the `Display mode` (in User settings / Display) to `TV`.

KingIzzymon:
Ah, that did the trick! Would you mind if I made a PR to add a note next to the "Enable gamepad" setting to say that display needs to be in TV mode?

There are several people who have posted about this confusion on various websites.

dmitrylyzo:
We need to (try to) detect Steam Deck (`browser.js`) and treat it as a TV (probably as not "slow") to automatically change the display mode.

We can also automatically activate a gamepad for it, as this is its main control. :thinking:

KingIzzymon:
Sounds like a plan. I'll submit a quick PR this afternoon to add the label then I'll take a peek at trying to detect the steam deck.

KingIzzymon:
So javascript's "navigator.userAgent" function is only able to read that I'm using a linux operating system:

navigator.userAgent => `User-agent header: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) CHrome/109.0.0.0 Safari/537.36`

Because the steam deck is just a handheld linux computer, I'm not sure theres anything we can do to autodetect the steam deck at this time like we do for the xbox and ps4.

nyanmisaka:
[...]

PR #4301, opened 2023-01-23
Update docs - gamepad requirement

KingIzzymon:
It was found that Jellyfin-web does not automatically identify the steam deck as a game console so the 'Display -> Display Mode' setting needs to be changes from 'Auto' to 'TV' explicitly for the gamepad control (even if already enabled) to work. 

Note: webpage needs manual refresh after setting to TV and enabling gamepad control for the changes to be applied.

At this time, it is unclear on how to have javascript auto-detect the steam deck since it is technically a linux desktop system, so for now, a small note is being added to the "Enable gamepad" setting that TV display mode is required.

**Changes**
Update en-us.json & en-gb.json to include "gamepad enable" setting requires TV display mode.

**Issues**
Reference Issue #4296 Gamepad support broken on Steam Deck

### DIFF
--- a/src/strings/en-gb.json
+++ b/src/strings/en-gb.json
@@ -1518,7 +1518,7 @@
     "MessageChromecastConnectionError": "Your Google Cast receiver is unable to contact the Jellyfin server. Please check the connection and try again.",
     "Framerate": "Framerate",
     "DirectPlayHelp": "The source file is entirely compatible with this client, and the session is receiving the file without modifications.",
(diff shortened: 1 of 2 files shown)


dmitrylyzo (approved):

KingIzzymon:
To update this branch, should I "Update with Rebase" or just "Update with Merge Commit"?

dmitrylyzo:
I prefer "with Rebase" - cleaner git history.

PR #4327, opened 2023-02-04
Add toggle to show/hide remaining video time in video player

DVDAndroid:
**Changes**
Added a click listener to the remaining time label that toggles between show the total video length and the remaining video time.

When toggle is in remaining video time mode, `-` is added as a prefix.

It imitates the same behavior in VLC.

Toggle preference as local storage user setting.

### DIFF
--- a/src/controllers/playback/video/index.js
+++ b/src/controllers/playback/video/index.js
@@ -787,12 +787,18 @@ import { setBackdropTransparency, TRANSPARENCY_LEVEL } from '../../../components
                     nowPlayingPositionText.classList.add('hide');
                 }
 
-                const leftTicks = runtimeTicks - positionTicks;
-                if (leftTicks >= 0) {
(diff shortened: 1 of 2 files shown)


src/controllers/playback/video/index.js
```diff
@@ -871,6 +877,19 @@ import { setBackdropTransparency, TRANSPARENCY_LEVEL } from '../../../components
             elem.innerHTML = html;
         }
 
+        function nowPlayingDurationTextClick() {
+            if (userSettings.enableVideoRemainingTime()) {
+                userSettings.enableVideoRemainingTime(false);
+            } else {
+                userSettings.enableVideoRemainingTime(true);
+            }
```
thornbill, line 885:
```suggestion
            userSettings.enableVideoRemainingTime(!userSettings.enableVideoRemainingTime());
```

thornbill (approved):

PR #4330, opened 2023-02-06
Fix navigation for some types of INPUT

dmitrylyzo:
Regression from #4166

**Changes**
Don't constrain the keyboard navigation for some types of INPUT.

**Issues**
The keyboard navigation in the filter of listview is broken.

### DIFF
--- a/src/scripts/keyboardNavigation.js
+++ b/src/scripts/keyboardNavigation.js
@@ -50,6 +50,11 @@ const NavigationKeys = ['ArrowLeft', 'ArrowRight', 'ArrowUp', 'ArrowDown'];
  */
 const InteractiveElements = ['INPUT', 'TEXTAREA'];
 
+/**
+ * Types of INPUT element for which navigation shouldn't be constrained.
+ */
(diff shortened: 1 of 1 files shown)


src/scripts/keyboardNavigation.js
```diff
@@ -50,6 +50,11 @@ const NavigationKeys = ['ArrowLeft', 'ArrowRight', 'ArrowUp', 'ArrowDown'];
  */
 const InteractiveElements = ['INPUT', 'TEXTAREA'];
 
+/**
+ * Types of INPUT element for which navigation shouldn't be constrained.
+ */
+const NonInteractiveInputElements = ['button', 'checkbox', 'color', 'file', 'hidden', 'image', 'radio', 'reset', 'submit'];
```
dmitrylyzo, line 56: Or maybe go the opposite way and list "text" fields? :thinking:

thornbill: It does seem like that would make for a much shorter list to check. :sweat_smile:

thornbill (approved):

PR #4345, opened 2023-02-12
Fix subtitles desync

dmitrylyzo:
**Changes**
- Fix MediaReady.
`onMediaReady` callback is not called if `seconds` is `0` and when using progressive streaming.
- Fix removing listeners.
They remain connected when using progressive streaming because `currentTime` is not `0` when `duration` reaches `seconds`.
- Fix subtitle desync when progressive streaming.
Copy the timestamps because the seeking is not accurate (I-frame start?).
- Fix time offset in JSO constructor _(in fact, cosmetic)_.
- Fix requested position when progressive streaming.
`seconds` must correspond to `video.currentTime` which begins with `0` relative to the transcoding offset.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-web/issues/4346

### DIFF
--- a/src/components/htmlMediaHelper.js
+++ b/src/components/htmlMediaHelper.js
@@ -151,14 +151,17 @@ import { Events } from 'jellyfin-apiclient';
                 // update video player position when media is ready to be sought
                 const events = ['durationchange', 'loadeddata', 'play', 'loadedmetadata'];
                 const onMediaChange = function(e) {
-                    if (element.currentTime === 0 && element.duration >= seconds) {
-                        // seek only when video position is exactly zero,
-                        // as this is true only if video hasn't started yet or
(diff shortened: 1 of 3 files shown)


dmitrylyzo:
This may also require `copyts`:
https://github.com/jellyfin/jellyfin-web/blob/871ec1ff6b7da0aee34c2d2fb38169344fe609ca/src/scripts/browserDeviceProfile.js#L762-L769
MKV:
https://github.com/jellyfin/jellyfin-web/blob/871ec1ff6b7da0aee34c2d2fb38169344fe609ca/src/scripts/browserDeviceProfile.js#L705-L713

dmitrylyzo:
The main idea (`copyts`) became unnecessary after #4385
Other fixes can be applied - separate PRs.

Issue #977, opened 2020-03-24
Chrome 80 tab freeze feature breaks Chromecast control


mooninite:
**Describe the bug**

I use the Chrome browser in Android for casting to my Chromecast. Chrome 78 introduced a feature to freeze inactive tabs after 5 minutes, which was disabled by default. Unfortunately in Chrome 80 they turned it on by default and removed the flag. This now freezes the Jellyfin tab after 5 minutes of inactivity and breaks the Chromecast connection.

**To Reproduce**

1. Open Chrome on Android
2. Cast Jellyfin to a Chromecast
3. Let Jellyfin tab go inactive for 5 minutes
4. Open Jellyfin tab and try to control Chromecast

**Expected behavior**

Playtime bar should increase time. Stop, pause buttons should work.

The tab freeze feature causes the playtime bar to freeze and stop/pause buttons to not work.
**Logs**

**Screenshots**

**System (please complete the following information):**
 - OS: Android 10
 - Browser: Chrome 80
 - Jellyfin Version: 10.5.0

**Additional context**

A workaround is to refresh the tab. Jellyfin is able to reconnect to the Chromecast without disrupting the playing video and control is returned.

My complaint may be with Chrome instead of Jellyfin unless there is another workaround.

YouKnowBlom:
Unfortunately I don't think there's much we can do. Best approach is probably asking them directly for a way to disable it.

Just cause I'm curious, do you prefer accessing jellyfin via chrome rather than the app?

mooninite:
My Chrome usage started with Emby when their Android app required a subscription. Besides that, Chrome allows picture-in-picture.

Chrome bug filed: https://bugs.chromium.org/p/chromium/issues/detail?id=1065654

JustAMan:
I wonder if this is related to https://github.com/jellyfin/jellyfin-web/issues/35

mooninite:
It is not related.

JustAMan:
How come? That scenario looks the same to yours - connection to a Chromecast is lost upon page reload.

mooninite:
I know how to reproduce issue #35 as I am familiar with it and am affected by it. They are not the same issue as the connection is not immediately lost upon page reload. Please re-read my reproducing steps.

Issue #38:
- Open Jellyfin
- Cast to Chromecast
- Navigate to Server Dashboard page
- Chromecast connection is 'lost'

This issue:
- Open Jellyfin
- Cast to Chromecast
[...]

PR #4347, opened 2023-02-12
Fix Blurred pdf

Zourlo:
Blurred pdf may happen if the screen is small. I modified pdfPlayer to avoid it.

**Changes**

I modified pdfPlayer / renderPage so the issue I found is fixed. I increased the scale and changed the canvas size to render the PDF within the size of the screen.

**Issues**

Fixes jellyfin-android  Books (pdf) #977

### DIFF
--- a/src/plugins/pdfPlayer/plugin.js
+++ b/src/plugins/pdfPlayer/plugin.js
@@ -278,16 +278,22 @@ export class PdfPlayer {
 
     renderPage(canvas, number) {
         this.book.getPage(number).then(page => {
-            const original = page.getViewport({ scale: 1 });
+            const width = dom.getWindowSize().innerWidth;
+            const height = dom.getWindowSize().innerHeight;
+			
+            const viewport = page.getViewport({ scale: 5 });
             const context = canvas.getContext('2d');
-
-            const widthRatio = dom.getWindowSize().innerWidth / original.width;
-            const heightRatio = dom.getWindowSize().innerHeight / original.height;
-            const scale = Math.min(heightRatio, widthRatio);
-            const viewport = page.getViewport({ scale: scale });
-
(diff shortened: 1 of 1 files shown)


src/plugins/pdfPlayer/plugin.js
```diff
@@ -278,16 +278,21 @@ export class PdfPlayer {
 
     renderPage(canvas, number) {
         this.book.getPage(number).then(page => {
-            const original = page.getViewport({ scale: 1 });
+            const width = dom.getWindowSize().innerWidth;
+            const height = dom.getWindowSize().innerHeight;
+            const viewport = page.getViewport({ scale: 5 });
```
thornbill, line 283: Should the scale be the pixel density of the screen?

Zourlo: Thank you for your comment.
I am not sure. My phone's screen has more pixel density than my monitor. 
By scaling high and resizing the pdf depending of the screen size, fixed the problem. I put 5 but 3 was enough for my device.

src/plugins/pdfPlayer/plugin.js
```diff
@@ -279,16 +279,21 @@ export class PdfPlayer {
...
```
thornbill, line 284: I would recommend trying this instead of picking an arbitrary value for the scale.

```suggestion
            const scale = Math.ceil(window.devicePixelRatio || 1);
            const viewport = page.getViewport({ scale });
```

thornbill (changes requested):

git
thornbill, line 1: This shouldn't be here... :wink:

thornbill (changes requested):

thornbill (approved):

PR #4349, opened 2023-02-12
Fix stuck text track cues for firefox

is343:
This PR is regarding a bug that I found while working on #3906. The bug is that the current visible subtitle track would not clear when setting an offset. Any text on the screen when the offset changes would remain after the offset changed, and would be stuck visible on the screen. This seems to currently only affect firefox, and it seems to only happen on development builds via the `start` script. I haven't been able to reproduce this in the current prod build.

The fix I implemented in #3906 did not completely fix the issue and it was still reproducible. Because the fix was out of scope of that PR, we decided to pull this out into it's own PR and work on it separately.

This has still not been solved at the time of opening this draft PR. This draft will allow us to continue to debug the issue and contribute.

Once a fix has been found, I will convert this from draft.

**Changes**

**Issues**

This is fixing a bug that I found while working on #3906.

#### The below screenshots and descriptions were taken from the original PR.

Screen Shot 2023-02-09 at 7 36 08 PM

Firefox - (before fix) showing the bug where the current visible track does not clear when changing the offset. You can see the currently visible text (music notes) remains rendered until it reaches the point where it would be rendered again with the new offset timing
Screen Recording 2022-09-14 at 6 10 24 PM

Firefox - (after fix) no longer showing the the bug from above
Screen Recording 2022-09-14 at 8 01 33 PM


is343:
@dmitrylyzo I am now able to force-clear the active cues for firefox. Please verify on your end that you can no longer reproduce the issue.

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -560,17 +566,67 @@ function tryRemoveElement(elem) {
...
+            }
+        }
+
         /**
          * @private
          */
         setTextTrackSubtitleOffset(currentTrack, offsetValue) {
             if (currentTrack.cues) {
                 offsetValue = this.updateCurrentTrackOffset(offsetValue);
+                const shouldClearActiveCues = offsetValue && this.requiresClearingActiveCuesOnOffsetChange();
```
dmitrylyzo, line 616:
```suggestion
                const shouldClearActiveCues = offsetValue && this.requiresHidingActiveCuesOnOffsetChange();
```

is343: That's embarrassing. I must have tried to change the name but must have not saved correctly or something.

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -560,17 +566,67 @@ function tryRemoveElement(elem) {
...
+         * Forces the active cue to clear by disabling then re-enabling the track.
+         * The track mode is reverted inside of a 0ms timeout to free up the track
+         * and allow it to disable and clear the active cue.
+         * The track needs to be temporarily stored in order for us to access it
+         * inside the timeout. The stored value is reset after it is used.
+         * @private
+         */
+        forceClearTextTrackActiveCues(currentTrack) {
+            if (currentTrack.activeCues) {
+                this.#currentTextTrack = currentTrack;
```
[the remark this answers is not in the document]
is343: I originally stored it temporarily because I was oddly getting strange errors where it was losing the reference. But I'm trying this again by just referencing the `currentTrack` that we pass in, and I am not getting those errors anymore.
You are correct, the closure is keeping access open for us.
The error I was getting in the console before must have been something unrelated.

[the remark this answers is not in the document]
is343: Thanks for questioning this and getting me to look at it again.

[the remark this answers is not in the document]
is343: I confirmed that this PR fix in it's current state works with the secondary subtitles.

dmitrylyzo (commented):
In general, everything works fine.

But when you rapidly click the slider, the tracks (_I'm testing with secondary subtitles PR_) becomes empty because they all are disabled before the timeout is triggered.
https://github.com/jellyfin/jellyfin-web/blob/1bf97318f802fd1aaca222b13f79b6cafd94ab03/src/plugins/htmlVideoPlayer/plugin.js#L537
and we fall here and lose the offset change:
https://github.com/jellyfin/jellyfin-web/blob/1bf97318f802fd1aaca222b13f79b6cafd94ab03/src/plugins/htmlVideoPlayer/plugin.js#L544

Theoretically, the timeout could also break the subtitle change by enabling the previous track.

is343:
> In general, everything works fine.
> 
> But when you rapidly click the slider, the tracks (_I'm testing with secondary subtitles PR_) becomes empty because they all are disabled before the timeout is triggered.
> 
> https://github.com/jellyfin/jellyfin-web/blob/1bf97318f802fd1aaca222b13f79b6cafd94ab03/src/plugins/htmlVideoPlayer/plugin.js#L537
> 
> and we fall here and lose the offset change:
> 
> https://github.com/jellyfin/jellyfin-web/blob/1bf97318f802fd1aaca222b13f79b6cafd94ab03/src/plugins/htmlVideoPlayer/plugin.js#L544
> 
> Theoretically, the timeout could also break the subtitle change by enabling the previous track.

I wonder if it makes sense to try implementing a some sort of debounce to prevent too many offset requests happening at once.

dmitrylyzo:
> I wonder if it makes sense to try implementing a some sort of debounce to prevent too many offset requests happening at once.
> 
> Under normal use, I don't think we should have this problem. But for situations like this, we can prevent it from breaking and ruining the user experience.

Maybe I'm just being overprotective, and the answer to the user is _"don't do that"_. :thinking:

is343:
> Maybe I'm just being overprotective, and the answer to the user is _"don't do that"_. 🤔

That is probably the simplest approach for now. I'm looking into what we would need to do to implement some sort of throttle/debounce from lodash and it seems like a lot. It will be even more complicated once the secondary subtitle PR gets merged, since both tracks call the same set method, so we would have to make sure the second one doesn't get debounced/throttled. We'd probably have to refactor again once that PR also gets merged.

Maybe we can advise people to not rapid click the button if this issue comes up. And if it does become a problem, we can find a more permanent solution at that time.

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -523,10 +524,12 @@ function tryRemoveElement(elem) {
             }
         }
 
+        setSubtitleOffset = debounce(this._setSubtitleOffset, 500);
+
```
is343, line 528: @dmitrylyzo Let me know if this looks like something we want to implement to deal with this issue. 
It seems to work for me (even with secondary subtitles) when rapidly changing the offset value.

Even though it was already in the dependency list, I don't want to start importing from `lodash-es` if this is something we want to avoid for the repo in general

dmitrylyzo: > Even though it was already in the dependency list, I don't want to start importing from `lodash-es` if we are trying to avoid using lodash in the repo in general

`lodash-es` is already in use, so feel free to use it.

I would change `500` ms with `100` ms.

is343: Great sounds good. I will make that change now.

dmitrylyzo:
With `debounce` [this](https://github.com/jellyfin/jellyfin-web/blob/89b82e6d5214d37d94deed23dbf2acc7c7f838fc/src/components/subtitlesync/subtitlesync.js#L137) produces `No available track, cannot apply offset:  0` message, i.e. we are doing something (resetting subtitle offset) to the player we are releasing. Perhaps there is no harm, but it smells.

I guess we can cancel the debounced function when we are destroying the player. :thinking:

is343:
> With `debounce` [this](https://github.com/jellyfin/jellyfin-web/blob/89b82e6d5214d37d94deed23dbf2acc7c7f838fc/src/components/subtitlesync/subtitlesync.js#L137) produces `No available track, cannot apply offset: 0` message, i.e. we are doing something (resetting subtitle offset) to the player we are releasing. Perhaps there is no harm, but it smells.
> 
> I guess we can cancel the debounced function when we are destroying the player. 🤔

You are right, I do see it on my end too. I was not aware that we reset the value (it makes sense though), so I didn't test with navigating away from the player. 

I think `cancel` will stop all pending. `flush` seems to call immediately. 
I tried adding `this.setSubtitleOffset.flush();` to the start of the html player `destroy` method, and that seems to work. It is calling `setSubtitleOffset` with an offset of `0` when navigating away.
https://github.com/jellyfin/jellyfin-web/blob/4d311181e0baa275fade43e523a232c446c64485/src/plugins/htmlVideoPlayer/plugin.js#L746

dmitrylyzo:
I think we should just cancel it because we are destroying the player and changing the offset is useless.

Moreover, we don't want to trigger that timeout (in `forceClearTextTrackActiveCues`). It can still be triggered if we are "lucky" to call `destroy` right after `debounce` is triggered. And even then there won't be much harm :crossed_fingers:

dmitrylyzo (approved):

PR #4350, opened 2023-02-13
Fix library options missing when editing an existing Mixed-type library

1337joe:
**Changes**
- Add checks for `undefined` `contentType` and treat it the same as 'mixed'

On library creation the dialog sees the type as 'mixed' but when queried for existing libraries the server provides mixed-type libraries with a null `CollectionType`, which doesn't populate on the javascript side and fails the check when only testing for 'mixed'.

**Issues**
Fixes jellyfin/jellyfin#8538

### DIFF
--- a/src/components/libraryoptionseditor/libraryoptionseditor.js
+++ b/src/components/libraryoptionseditor/libraryoptionseditor.js
@@ -398,7 +398,7 @@ import template from './libraryoptionseditor.template.html';
             parent.querySelector('.chkEnablePhotosContainer').classList.add('hide');
         }
 
-        if (contentType !== 'tvshows' && contentType !== 'movies' && contentType !== 'homevideos' && contentType !== 'musicvideos' && contentType !== 'mixed') {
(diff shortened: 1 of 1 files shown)


thornbill:
Why is null being returned though? "mixed" should be a valid type. :thinking:

1337joe:
You're not wrong. I expected `mixed` to be a defined type as well, made debugging this one more tricky than I expected.

Here's where it should be but isn't: [CollectionType.cs](https://github.com/jellyfin/jellyfin/blob/master/MediaBrowser.Model/Entities/CollectionType.cs)

This is the logic that determines contents of the library: https://github.com/jellyfin/jellyfin/blob/54cd3e6d551d797bace33d65334cf1e98669676d/Jellyfin.Api/Controllers/LibraryController.cs#L905

Note that the default case is the type list for `mixed`, so it hits that by falling through everything else instead of having a defined item.

I assumed there was some history that led to this state, but looking into the git blame on `LibraryController` it looks like we inherited this from Emby. I'll try to find some time this weekend to look into cleaning this up on the server side.

PR #4356, opened 2023-02-14
Babelify @jellyfin/libass-wasm

dmitrylyzo:
**Changes**
- Simplify adding modules to babel-loader
- Babelify @jellyfin/libass-wasm

**Issues**
`libass-wasm` must be transpiled because of the forgotten `const`. _This will also be helpful when migrating to ES6._
https://github.com/jellyfin/jellyfin-webos/issues/31#issuecomment-1095401463

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -97,6 +97,7 @@ module.exports = {
             {
                 test: /\.(js|jsx)$/,
                 include: [
+                    path.resolve(__dirname, 'node_modules/@jellyfin/libass-wasm'),
                     path.resolve(__dirname, 'node_modules/@uupaa/dynamic-import-polyfill'),
                     path.resolve(__dirname, 'node_modules/blurhash'),
                     path.resolve(__dirname, 'node_modules/date-fns'),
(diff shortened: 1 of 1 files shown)


thornbill (approved):

etu:
Heya, I've ran into problems after the update of Jellyfin(-web) to 10.8.10, had no idea what was going on because Jellyfin worked on my phone and in my browser. However, it didn't load on my LG TV that runs WebOS.

So without being sure about anything I opened this ticket https://github.com/jellyfin/jellyfin-webos/issues/154 and rolled my system back to a version prior.

Now when I had some time over I tried to roll back only jellyfin to 10.8.9 (not jellyfin-web, it was still on 10.8.10) or anything else and it was still broken so I assumed it was something else in the system, maybe nginx or something.

Some hours later of bisecting nixpkgs/nixos I found that the breakage were due to jellyfin-web and not jellyfin. 

So I spent some time bisecting jellyfin-web and trying out specific commits, bisect log available here: https://github.com/jellyfin/jellyfin-webos/issues/154#issuecomment-1529570746

So... yeah, that pointed me here. This PR broke Jellyfin-web on LG WebOS 5 TV's. Not sure *why* though, I'm not sure if I have a way to inspect the browser or so. I don't even have developer tools on it enabled... :slightly_smiling_face:

PR #4357, opened 2023-02-15
feat: migrate from JavascriptSubtitlesOctopus to JASSUB

ThaUnknown:
**Changes**

Migrated from JSSO to JASSUB, the 2 libraries are very similar in what they do and how they do it, but JASSUB is a lot more modern. The changes themselves are very simple as JASSUB implements all the old JSSO functionality, with some changed paths and function names in a few cases.
**Issues**

- https://github.com/jellyfin/JavascriptSubtitlesOctopus/pull/46

**Specific code changes**
JASSUB handles rendering a LOT differently from JSSO and as such it implements a lot of new rendering modes:
- on demand render - this renders subtitles ONLY when a new video frame is about to be painted on the user's screen, this does away with the whole timestamp and event system JSSO used, and it's immensely accurate allowing for pretty much flawless rendering of any custom TS [example](https://thaunknown.github.io/jassub/jassub/burn-in/index.html) [not that any english text u see on that video is TS, which might not be obvious at first glance], this is also of amazing benefit for[ very high framerate](https://thaunknown.github.io/jassub/jassub/high/index.html) videos and [variable framerate](https://thaunknown.github.io/jassub/jassub/variable/index.html) videos, as it renders together with the video frame, not at some abstract 24 fps [note the yellow square is the subtitle, the white one is the video, this showcases tracking]
- offscreen render - this uses the offscreen canvas API to render EVERYTHING fully on the worker, meaning the main web app itself won't lag no matter how heavy the subtitles are
- blend modes - the reason JSSO's JS render was slow was because it did decoding on JS, JASSUB does this in WASM, which means JS render will be a LOT lighter and marginally faster than WASM blend, but this only applies if the user has a dedicated GPU, it might be good to switch JS blend to WASM blend when we know the target device doesn't have a GPU

JASSUB handles fonts completely differently from JSSO, which makes it a lot faster
- you don't even need any fonts for the library to init, which means you don't need any data files for the library to work
- it supports dynamic font loading [not added in this PR, see availableFonts in typedefinions of JASSUB]
- it can load fonts directly from memory instead of from url [this is also not added in this PR as I couldn't find any type definions for `mediaSource.MediaAttachments` so I'm not sure if they implement direct binary data]
- it can load fonts directly from the users device if his browser [supports it](https://caniuse.com/mdn-api_window_querylocalfonts)

JASSUB handles video resizing completely differently
- videos can resize dynamically at any point, JASSUB handles that individually per frame, and updates the canvas size once a new video frame is about to render, this is in contrast to JSSO which just used timeouts which cleared the canvas and caused a lot of performance issues
- videos can be anamorphic which JSSO also didn't handle, JASSUB scales the subtitles correctly on non-firefox browsers [ff doesn't expose anamorphic metadata for videos]

JASSUB is compiled differently
- it uses lembind instead of webIDL, this leads to a roughly 500kb size reducement of the output WASM and JS as the bindings are dynamic
- since it drops webIDL it can use newer versions of emscripten [emsdk:2.0.34 -> emsdk:3.1.24]
- since a lot of old garbage was removed we can use features like minimal runtime, nomain, and compiled streaming which means the WASM itself is loaded and executed a LOT quicker, we're speaking >x2 speed

What JASSUB doesn't have compared to jellyfin's JSSO:
- ~~smart WASM blend [simply didn't have to implement it yet, soonTM]~~ added in 1.5.0
- renderAhead - I simply got rid of it, as onDemandRender already renders 1 frame ahead which is usually enough anyways, and from my testing it provided 0 benefits for heavy TS over not running it, and used more CPU when it really didn't help 

The library is overall a LOT smaller, with legacy workers included its 1283653 bytes smaller and ignoring legacy its 667793 bytes smaller

Note: the worker might need to be babel'ed as it uses modern syntax


ThaUnknown:
@dmitrylyzo

ThaUnknown:
im a bit confused, all of jassub is meant to be babell'ed yet the worker doesn't get babelled, I assume the copy plugin copies the worker before babel gets to it, any ideas @dmitrylyzo ?

ThaUnknown:
do note, that the main reason that JF takes so long to load ASS subtitles is because of the insane amt of fonts it tries to load:
image

if my dynamic font loading was implemented this would be cut down to close to 0

ThaUnknown:
I excluded the worker from the ESCheck as other workers are also excluded for it, additionally the pdf worker uses syntax newer than JASSUB workers, so I deemed it fine, if this is an issue I can fix it

ThaUnknown:
tests pass, everything works flawless:
image

nielsvanvelzen (approved):
The code changes look fine and the jassub demo works well for me. I have not tested the jellyfin-web build locally though.

dmitrylyzo (commented):
The main reason our JSO fork exists is to support older devices that cannot render subtitles fast enough. RenderAhead (main feature of our fork) is just buffering: render subtitles once and show them in time (the canvas changes only once per subtitle change). The less frequently subtitles are changed, the more subtitles can be rendered ahead.

I guess JASSUB is focused on modern web engines, i.e. no TV will get most of the improvements listed.

It currently does not work in webOS 1.2:
```
Unhandled promise rejection 
Error {stack: "../node_modules/jassub/src/jassub.js@http://192.16…les.core-js.bundle.js?0a94992cd65445704567:7027:6", line: 1097, sourceURL: "http://192.168.1.17:8096/web/node_modules.jassub.89b5285f60becdbf8beb.chunk.js"}
line: 1097
message: "Can't find variable: EventTarget"
sourceURL: "http://192.168.1.17:8096/web/node_modules.jassub.89b5285f60becdbf8beb.chunk.js"
stack: "../node_modules/jassub/src/jassub.js@http://192.168.1.17:8096/web/node_modules.jassub.89b5285f60becdbf8beb.chunk.js:1097:45↵__webpack_require__@http://192.168.1.17:8096/web/runtime.bundle.js?0a94992cd65445704567:23:46↵__webpack_require__@[native code]↵callReaction@http://192.168.1.17:8096/web/node_modules.core-js.bundle.js?0a94992cd65445704567:12297:25↵http://192.168.1.17:8096/web/node_modules.core-js.bundle.js?0a94992cd65445704567:12322:19↵flush@http://192.168.1.17:8096/web/node_modules.core-js.bundle.js?0a94992cd65445704567:4840:9↵apply@[native code]↵call@[native code]↵call@[native code]↵http://192.168.1.17:8096/web/node_modules.core-js.bundle.js?0a94992cd65445704567:7042:12↵run@http://192.168.1.17:8096/web/node_modules.core-js.bundle.js?0a94992cd65445704567:7016:7↵eventListener@http://192.168.1.17:8096/web/node_modules.core-js.bundle.js?0a94992cd65445704567:7027:6"
__proto__: Error
```
WebKit doesn't seem to have `EventTarget`. :thinking: 

> im a bit confused, all of jassub is meant to be babell'ed yet the worker doesn't get babelled, I assume the copy plugin copies the worker before babel gets to it, any ideas @dmitrylyzo ?

Yes, CopyPlugin just copies the workers as is. Workers are currently expected to be "babeled" by hand (emscripten doesn't transpilate, it just includes polyfills, iiuc).

I noticed that JASSUB doesn't seem to have a default font - if you have no fonts listed in the subtitles, the latter don't appear.

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1257,11 +1248,17 @@ function tryRemoveElement(elem) {
...
                     } else {
-                        this.#currentSubtitlesOctopus = new SubtitlesOctopus(options);
+                        this.#currentJASSUB = new JASSUB(options);
                     }
+
+                    this.#currentJASSUB.addEventListener('error', ()=>{
+                        this.#currentJASSUB.destroy();
+                        this.#currentJASSUB = null;
+                        onErrorInternal(this, 'mediadecodeerror');
+                    }, { once: true });
```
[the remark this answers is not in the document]
ThaUnknown: uuuh I'm missing something because I don't really understand what you mean

ThaUnknown:
@dmitrylyzo I honestly would have expected babel to polyfill event target, is it configured wrong or does it not polyfill api's?

Indeed JASSUB is focused on modern engines and insanely heavy subtitles, so for smart TV's it would *mostly* stay without change compared to JSO, outside a few decoding optimizations and the improved bindings.

Oops minor oversight with the default font, I assumed jf included them in available fonts, will look at a fix

src/plugins/htmlVideoPlayer/plugin.js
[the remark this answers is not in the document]
dmitrylyzo: The event subscription is performed before async creation of JASSUB here: https://github.com/jellyfin/jellyfin-web/blob/bcc4f876052427e3d7216c737cc4d7368a788792/src/plugins/htmlVideoPlayer/plugin.js#L1251

dmitrylyzo:
> @dmitrylyzo I honestly would have expected babel to polyfill event target, is it configured wrong or does it not polyfill api's?

We just approximated the old WebKit with Chrome 27. But `EventTarget` has existed since Chrome 1, maybe that's why it's not polyfiled. And we luckily didn't use it.
If this cannot be solved by the browserlist, then it should be explicitly polyfiled.

> Oops minor oversight with the default font, I assumed jf included them in available fonts, will look at a fix

JSO has a built-in fallback font. It is used for any font family that is missing, iiuc. In the upstream it has become configurable (I haven't pulled it to our fork).

ThaUnknown:
yup, all good I already looked at all this and know how to fix it, it's just gonna take a while as I'm currently writing an EBML parser, because I came up with the brilliant idea of enabling multi-track video/audio playback in browser on the fly on the client, so I've put off this for a day or two

ThaUnknown:
@dmitrylyzo I fixed the issues you found, tho I'm certain there will be more, mainly with the worker, but I have no way of testing this for webOS, any chance you could check again?

ThaUnknown:
I've fixed the worker issues by babel'ing the worker, but turns out babel'ing a 1MB worker uses 2GB of RAM, which crashes the pipelines, wonderous....

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1266,37 +1264,33 @@ function tryRemoveElement(elem) {
...
-                    setTimeout(() => {
-                        onErrorInternal(htmlVideoPlayer, 'mediadecodeerror');
-                    }, 0);
-                },
+                fallbackFont: 'liberation sans',
+                availableFonts: {'liberation sans': `${appRouter.baseUrl()}/libraries/default.woff2`},
+                // eslint-disable-next-line compat/compat
+                workerUrl: `${appRouter.baseUrl()}/${new URL('jassub/dist/jassub-worker.js', import.meta.url)}`,
+                // eslint-disable-next-line compat/compat
+                legacyWorkerUrl: `${appRouter.baseUrl()}/${new URL('jassub/dist/jassub-worker-legacy.js', import.meta.url)}`,
```
dmitrylyzo, line 1276: This results in incorrect URLs, such as `http://192.168.1.17:8096/web/http://192.168.1.17:8096/web/bd8c93ca1a6e55cd85f3.js`.
Reproducible in Chrome.

dmitrylyzo (commented):
`event-target-polyfill` should be babeled, but it seems to have a bug - `EventTarget` is not `Event`:
https://github.com/benlesh/event-target-polyfill/blob/4d4f2b626f5870ab46e802b8d7d71a2bb700ec58/index.js#L34
So I get an error [here](https://github.com/benlesh/event-target-polyfill/blob/4d4f2b626f5870ab46e802b8d7d71a2bb700ec58/index.js#L79-L83) because JASSUB throws `ErrorEvent`.

src/plugins/htmlVideoPlayer/plugin.js
ThaUnknown: ah shite, I expected webpack to compile the URL away to a string like it used to, I guess something changed

ThaUnknown: I fixed this by using pathname, in theory it should be fine to just use the new URL standalone, but I decided not to as the baseUrl method might be used for cases which I'm not aware of so I kept it in

ThaUnknown:
@dmitrylyzo okay this is the [`final_final_finalv3.patch`](https://twitter.com/patreon/status/1554183857343107073 aka I really don't see how it can have any more issues, this works flawlessly now, except for the fact that recently some hls changes broke playback for me, so testing this got more annoying, but it works regardless

dmitrylyzo:
I found why it thinks that `ErrorEvent` isn't `Event`: `event-target-polyfill` polyfills `Event` even if it exists.
https://github.com/benlesh/event-target-polyfill/blob/4d4f2b626f5870ab46e802b8d7d71a2bb700ec58/index.js#L15
`typeof root.Event` is `'object'` and `isConstructor(root.Event)` is `true`.

```diff
-if (typeof root.Event !== "function" || !isConstructor(root.Event)) {
+if (typeof root.Event !== "function" && !isConstructor(root.Event)) {
```
or maybe:
```diff
-if (typeof root.Event !== "function" || !isConstructor(root.Event)) {
+if (!isConstructor(root.Event)) {
```
https://github.com/benlesh/event-target-polyfill/blob/4d4f2b626f5870ab46e802b8d7d71a2bb700ec58/index.js#L28
`typeof root.EventTarget` is `'undefined'`.
```diff
-if (typeof root.EventTarget === "undefined" || !isConstructor(root.Event)) {
+if (typeof root.EventTarget === "undefined") {
```
or maybe:
```diff
-if (typeof root.EventTarget === "undefined" || !isConstructor(root.Event)) {
+if (!isConstructor(root.EventTarget)) {
```

src/plugins/htmlVideoPlayer/plugin.js
ThaUnknown: you said webos 1.2, that's so ancient it doesn't have let and const, the non legacy is transpiled for consistency sake

ThaUnknown: i don't really know why the URL is incorrect then, I'm gonna remove baseurl and you try that, cuz the current code works on chrome

ThaUnknown:
> I found why it thinks that `ErrorEvent` isn't `Event`: `event-target-polyfill` polyfills `Event` even if it exists. https://github.com/benlesh/event-target-polyfill/blob/4d4f2b626f5870ab46e802b8d7d71a2bb700ec58/index.js#L15 `typeof root.Event` is `'object'` and `isConstructor(root.Event)` is `true`.
> 
> ```diff
> -if (typeof root.Event !== "function" || !isConstructor(root.Event)) {
> +if (typeof root.Event !== "function" && !isConstructor(root.Event)) {
> ```
> 
> or maybe:
> 
> ```diff
> -if (typeof root.Event !== "function" || !isConstructor(root.Event)) {
> +if (!isConstructor(root.Event)) {
> ```
> 
> https://github.com/benlesh/event-target-polyfill/blob/4d4f2b626f5870ab46e802b8d7d71a2bb700ec58/index.js#L28 `typeof root.EventTarget` is `'undefined'`.
> 
> ```diff
> -if (typeof root.EventTarget === "undefined" || !isConstructor(root.Event)) {
> +if (typeof root.EventTarget === "undefined") {
> ```
> 
> or maybe:
> 
> ```diff
> -if (typeof root.EventTarget === "undefined" || !isConstructor(root.Event)) {
> +if (!isConstructor(root.EventTarget)) {
> ```

so the polyfill should check against root.Event not Event? I'll PR that, and change the install source to the PR'ed one for now

ThaUnknown:
@dmitrylyzo try this final_final_final_v4_final

src/plugins/htmlVideoPlayer/plugin.js
dmitrylyzo: > you said webos 1.2, that's so ancient it doesn't have let and const, the non legacy is transpiled for consistency sake

I mean, it should already be prepared for legacy browsers. Emscripten devs discussed transpilation, iirc, but I don't remember their decision.
`let/const` from [Chrome 49](https://caniuse.com/const).
WASM from [Chrome 57](https://caniuse.com/wasm).
So, at least for Chrome, we can allow `let/const` for the non-legacy worker.

> I'm gonna remove baseurl

It seems to work.

ThaUnknown: I compile the 2orkwr differently, and also use modern syntax, so I'm keeping it as it is

great, anything else to address or is this now mergable?

dmitrylyzo:
It works in Chrome 111, but still doesn't work on webOS 1.2/2/3/4

`typeof root.Event` is `object`, so the first condition is `true` [here](https://github.com/ThaUnknown/event-target-polyfill/blob/5152219bdee9bc18ab0a1cccddab48a813c20b81/index.js#L15). `&&` should be used or just `isConstructor`.
And yes, `isConstructor` was incorrect :+1: 

After `event-target-polyfill` fixed, I got `TypeError: 'undefined' is not an object (evaluating 'Ri.instantiateStreaming')` [here](https://github.com/ThaUnknown/jassub/blob/edd1b747121475d5530528f87a494a665ad388c9/src/worker.js#L25).

Added `Ri&&` (`WebAssembly &&`) and in webOS 1.2 got
```
BindingError: Cannot register public name '' twice
```
_Hard to say which line._
```js
var Fe = void 0;

function C2(U) {
    throw new Fe(U); // BindingError: Cannot register public name '' twice
}
```
I think it would be better if you get old Chrome <49 (say, 38 - [webOS 3](https://webostv.developer.lge.com/develop/specifications/web-api-and-web-engine)) and try it yourself.

ThaUnknown:
> BindingError: Cannot register public name '' twice

oh no bindings fail.... that's insanely bad

> download chrome 34

easy to say, but the oldest version of chrome I could find in google's storage was 48, and it doesn't even install ^^

ThaUnknown:
@dmitrylyzo how the hell did you even test this, I can't even open jellyfin because webpack uses arrow functions which are unsupported

ThaUnknown:
okay @dmitrylyzo that was quite the [nasty bug](https://github.com/emscripten-core/emscripten/issues/18994), but it's fixed, this time it was emscripten that was at fault, not me, this effectively means that legacy worker has almost no benefits over JSO, except maybe a bit of filesize

legacy worker now works and renders and I applied the fixes u mentioned for eventtarget, but I can't get JF to work on chrome 34 which is what webos 1.2's node version equals, but I can't see this not working anymore

dmitrylyzo:
> @dmitrylyzo how the hell did you even test this, I can't even open jellyfin because webpack uses arrow functions, const and fetch which are unsupported

I don't use `npm run serve` - I replace jellyfin-web on the server (build development version) + proxy with `no-cache` header and use webOS emulator + CLI to run the app and Chrome 79 (newer doesn't work for webOS 4+).

ThaUnknown:
anyways this should finally work, this time I tested everything

dmitrylyzo:
You should leave test for `EventTarget` [here](https://github.com/ThaUnknown/event-target-polyfill/blob/8becf1fbdd9ec547c1ea2097d2cb492f6e4d3984/index.js#L28).
The original idea was probably to polyfill `EventTarget` if `Event` also requires polyfilling. In this case `isConstructor(root.Event, 'event')` should be stored in a variable, because once `Event` is polyfilled, we cannot check the opposite.

Some progress:
```
detected that ImageData is not constructable despite browser saying so 
Detected a browser having issue with transparent pixels, applying workaround
overridden console
Failed to load resource: the server responded with a status of 404 (Not Found) http://192.168.1.17/jellyfin/web/undefined
Failed to load resource: Not Found
jso: setting total libass memory limits to: glyph=40 MiB, bitmap cache=40 MiB 
JASSUB: fontselect: failed to find any fallback with glyph 0x0 for font: (Source Sans Pro Semibold, 700, 0) 
playbackmanager playback error type: mediadecodeerror
undefined
InvalidStateError: DOM Exception 11: An attempt was made to use an object that is not, or is no longer, usable. 
InvalidStateError: DOM Exception 11: An attempt was made to use an object that is not, or is no longer, usable.
```
`InvalidStateError: DOM Exception 11: An attempt was made to use an object that is not, or is no longer, usable.` [here](https://github.com/ThaUnknown/jassub/blob/8d3de0075472a344f72936885b83dd06c2b1edc5/src/polyfill.js#L56).
Temporarily commented `postMessage` block (probably an induced error).
`ReferenceError: Can't find variable: buffer` [here](https://github.com/ThaUnknown/jassub/blob/8d3de0075472a344f72936885b83dd06c2b1edc5/src/worker.js#L227).

ThaUnknown:
> Failed to load resource: the server responded with a status of 404 (Not Found) http://192.168.1.17/jellyfin/web/undefined
Failed to load resource: Not Found

I hope that's not me

> JASSUB: fontselect: failed to find any fallback with glyph 0x0 for font: (Source Sans Pro Semibold, 700, 0) 

That happens on first frame, harmless mainly because the WASM loads before any fonts load so the library isn't cockblocked by fonts loading, fixes itself once the fonts are loaded and the video demands a new frame to paint

> undefined

uuuuuuuuuuuuuuuuuuh?

> InvalidStateError: DOM Exception 11: An attempt was made to use an object that is not, or is no longer, usable.

uuuUUUuuUUUh?... the fuck? the code for console has not changed from what JSO used, I'd assume these are XHR errors?

ThaUnknown:
okay, network errors and main thread rendering also fixed some suppression for some common logs and event target

dmitrylyzo:
> I've read this about 4 times now and I still don't know what you mean. Currently to polyfill eventtarget either event or eventtarget has to be unconstructable, because if ET works but Event doesn't then you can't dispatch the custom events as they arent DOM events just some custom objects, I do not see any fault in code here?

It looks like `new EventTarget('')` fails (but `EventTarget` exists) and `isConstructor(root.EventTarget)` returns `false`. So it polyfills `EventTarget`.
When I prevent this, JASSUB fails on [`new JASSUB`](https://github.com/jellyfin/jellyfin-web/blob/062ee9d5240795cd514b866cdff15136eed1ec8d/src/plugins/htmlVideoPlayer/plugin.js#L1324).
But if I allow to polyfill (don't touch), it fails on WASM compiling as webOS 5, i.e. no `Illegal constructor` error.

Forget this for now:
> > I think it should be checked using `typeof` as it was (`const hasTarget = typeof root.EventTarget !== "undefined"`).

I am not sure, but at least it goes further.

> ....what? the browser supports WASM but can't compile it? I found no issues like that by googling the error or even the build param so that's interesting.... ideas?

[This commits](https://github.com/jellyfin/jellyfin-web/pull/4357/commits/132825962aea0dce523aad9f0a09988f573e9ef5) works.
Btw, this also works in webOS 4 (WASM).

ThaUnknown:
@dmitrylyzo okay everything but webOS 2 should work? I can't imagine webOS 2 working tbh unless we somehow check for webOS2 inside the plugin.js file? is that possible? does JF expose what webOS version its running, do you know how I'd check for that? I'd simply disable onDemandRender and that'd fix it

dmitrylyzo:
It works now on (emulators):
- webOS 1.2
- webOS 2
- webOS 4
- webOS 5
_There is a color shift with `Railgun S Karaoke Test` subtitles: blue lightnings become red._

I haven't tested the performance, maybe only that it starts rendering earlier (_new font processing_) in compare to our JSO fork.

ThaUnknown:
> It works now on (emulators):
> 
> * webOS 1.2
> * webOS 2
> * webOS 4
> * webOS 5
>   _There is a color shift with `Railgun S Karaoke Test` subtitles: blue lightnings become red._
> 
> I haven't tested the performance, maybe only that it starts rendering earlier (_new font processing_) in compare to our JSO fork.

is the transparency fix being applied when the color shift happens? that's pretty much the only thing that's changed

ThaUnknown:
I'm struggling to reproduce said color issues, like in any way, shape or form, also comparing the code from what JSO does, I can't find anything conditional which could cause this

dmitrylyzo:
It works now on (emulators):
- webOS 1.2
- webOS 2
- webOS 4
- webOS 5
_There is a color shift with `Railgun S Karaoke Test` subtitles: blue lightnings become red._

I haven't tested the performance, maybe only that it starts rendering earlier (_new font processing_) in compare to our JSO fork.

> is the transparency fix being applied when the color shift happens? that's pretty much the only thing that's changed, if yes could you try to force disable the transparency fix?

`hasAlphaBug` is `false` for both JASSUB and JSO.
Disabling async rendering fixes this bug. Transparency fix should probably be applied to the worker as well.

ThaUnknown:
transparency fix on worker is kinda useless as it needs a constructed array to manipulate, which JSO doesn't provide, as the worker works purely on buffers without constructing arrays, because it's slow, it's one of the optimisations it has over JSO, and the transparency correction only works when the browser doesnt have createImageBitmap [quite new] or offscreenCanvas [chrome only] so it's unlikely that a browser modern enough that has those features has that bug

anyways that doesn't matter, JSO also didn't perform alpha correction on the worker, and didn't perform it for asyncRender either so the issue must lie somewhere else

ThaUnknown:
@dmitrylyzo as I can't reproduce this like at all, I've blindly changed the only difference between my code and what JSO does, try to see if it fixes it, if it doesn't please provide me with the exact files and reproduction steps you used because I can't reproduce this error at all

dmitrylyzo:
> @dmitrylyzo as I can't reproduce this like at all, I've blindly changed the only difference between my code and what JSO does, try to see if it fixes it, if it doesn't please provide me with the exact files and reproduction steps you used because I can't reproduce this issue at all

Replacing `subarray` with `slice` fixes the color shift.
Now it seems to work on webOS 1.2, 2.0, 4.0, 5.0, 6.0

ThaUnknown:
> > @dmitrylyzo as I can't reproduce this like at all, I've blindly changed the only difference between my code and what JSO does, try to see if it fixes it, if it doesn't please provide me with the exact files and reproduction steps you used because I can't reproduce this issue at all
> 
> Replacing `subarray` with `slice` fixes the color shift. Now it seems to work on webOS 1.2, 2.0, 4.0, 5.0, 6.0

that is really weird and really unfortunate, because under heavy load that slice operation can be up to 12% slower than subarray, I assume the createImageBitmap is a reference to the imagedata which is a reference to the uint8c, which gets mutated when the next paint call gets ran? lots of speculation, but oh well at least this is now functional and mergable, feel free to ship it

ThaUnknown:
@dmitrylyzo can you review/approve this then? or do you want to benchmark and stuff

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -209,7 +209,7 @@ function tryRemoveElement(elem) {
         /**
          * @type {any | null | undefined}
          */
-        #currentSubtitlesOctopus;
+        #currentJASSUB;
```
thornbill, line 212: I think it would be better to use generic variable names where possible to avoid needing to change this again at some point. Maybe something like `currentAssRenderer`?

package.json
```diff
@@ -85,12 +84,14 @@
     "dompurify": "3.0.1",
     "epubjs": "0.3.93",
     "escape-html": "1.0.3",
+    "event-target-polyfill": "github:ThaUnknown/event-target-polyfill",
```
thornbill, line 87: I realize this should be temporary, but I'm not a fan of installing directly from github here. It's not a great solution, but if it's not available on npm perhaps we should include it in `src/legacy` with some of the other polyfills for now. :man_shrugging:

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1266,59 +1264,68 @@ function tryRemoveElement(elem) {
...
-                    // HACK: Clear JavascriptSubtitlesOctopus: it gets disposed when an error occurs
-                    htmlVideoPlayer.#currentSubtitlesOctopus = null;
-
-                    // HACK: Give JavascriptSubtitlesOctopus time to dispose itself
-                    setTimeout(() => {
-                        onErrorInternal(htmlVideoPlayer, 'mediadecodeerror');
-                    }, 0);
-                },
+                fallbackFont: 'liberation sans',
+                availableFonts: {'liberation sans': `${appRouter.baseUrl()}/default.woff2`},
```
thornbill, line 1272: afaik we don't ship this font?

thornbill (commented):
I can't test the functionality of this, but left a few comments on the code changes. :slightly_smiling_face:

src/plugins/htmlVideoPlayer/plugin.js
ThaUnknown: I doubt any1's batshit insane enough to make another ASS renderer, speaking from experience of how much sanity I've lost over trying to implement mine, so I don't think there's actually a need?

package.json
ThaUnknown: no followup on this, and since this is already done previously in this package i'm gonna assume its fine

src/plugins/htmlVideoPlayer/plugin.js
ThaUnknown: no followup, so i'm gonna assume its not needed

thornbill: You wouldn't think this would happen often, but this would be the third ass/ssa rendering library we've used in 4 years.

ThaUnknown: lets hope it's the last you'll need lol, renamed it @thornbill

thornbill (approved):

Issue #4352, opened 2023-02-13
Cast & Crew listing on Episode page does not include Guest Stars


bugfixin:
When viewing the details of an episode, the Cast & Crew listing doesn't include people of type "guest star". I believe it should - there may be other categories of People that it should include and doesn't.

bugfixin:
I believe this stems from the filtering of the cast/crew to type "actor":

https://github.com/jellyfin/jellyfin-web/blob/master/src/controllers/itemDetails/index.js#L1846 

Seeing as the category is called "Cast & Crew" and not just "Cast" - why filter to actors-only? This behavior is notably different from the android TV app, which does display crew as well as actors.

thornbill:
Looks like a duplicate of #4148

PR #4361, opened 2023-02-17
Show all Cast&Crew, not just Actors

bugfixin:
**Changes**
On item details pages, only actors are currently shown in Cast & Crew listings - this should be expanded to all cast & crew

**Issues**
Addresses https://github.com/jellyfin/jellyfin-web/issues/4352

### DIFF
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -1804,9 +1804,7 @@ function renderSpecials(page, item, user) {
 }
 
 function renderCast(page, item) {
-    const people = (item.People || []).filter(function (p) {
-        return p.Type === 'Actor';
-    });
+    const people = item.People || [];
 
     if (!people.length) {
         page.querySelector('#castCollapsible').classList.add('hide');
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -1,3 +1,4 @@
+import { PersonKind } from '@jellyfin/sdk/lib/generated-client/models/person-kind';
 import { intervalToDuration } from 'date-fns';
 import DOMPurify from 'dompurify';
 import escapeHtml from 'escape-html';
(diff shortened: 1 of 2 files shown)


thornbill:
We probably should filter out director, producer, etc. that are listed separately.

bugfixin:
That doesn't make sense to me - why call it cast & crew if it doesnt include "and crew"? If this is the behavior on other apps I don't see this as an enhancement, it's a bug fix.

thornbill:
Director(s) and writer(s) are already listed which is what I was referencing. It would sense to add producer(s) in the same fashion in my opinion.

image

danbracey:
Please have a look at https://github.com/jellyfin/jellyfin-web/pull/5211 in conjunction with this one please :)

1hitsong:
> Director(s) and writer(s) are already listed which is what I was referencing. It would make sense to add producer(s) in the same fashion in my opinion.

Personally, I like them listed in the Cast & Crew row. I like the pretty pictures.

✅ Confirmed this PR shows both cast and crew in the Cast & Crew section.

1hitsong (approved):

src/controllers/itemDetails/index.js
```diff
@@ -1842,9 +1842,7 @@ function renderSpecials(page, item, user) {
 }
 
 function renderCast(page, item) {
-    const people = (item.People || []).filter(function (p) {
-        return p.Type === 'Actor';
```
1hitsong, line 1846: Since this PR was originally created, the Cast & Crew and Guest Star lists have been separated on the page.

This means we need to filter out people whose `Type` is 'GuestStar'

1hitsong (changes requested):

thornbill:
Just for (probably my own) future reference, this worked previously but was inadvertently broken (in 2020!) when changing how writers are displayed.

Ref: https://github.com/jellyfin/jellyfin-web/commit/660e4a0a3cf94d50d9872f62234da99c003d092f

1hitsong (approved):

PR #4362, opened 2023-02-18
Fix subtitle offset reset when seeking progressive stream

dmitrylyzo:
**Changes**
Add an option to prevent the subtitle offset from being reset.

**Issues**
Subtitle offset resets when seeking progressive stream.
See https://github.com/jellyfin/jellyfin-web/pull/613


thornbill (approved):
wow that's an old one... nice find!

PR #4387, opened 2023-03-01
Move public pages to react-router

thornbill:
**Changes**
Moves the last of the app router pages to using react-router.

**Issues**
N/A


src/components/appRouter.js
```diff
@@ -124,7 +124,7 @@ class AppRouter {
                 isBack: action === Action.Pop
             });
         } else {
-            console.info('[appRouter] "%s" route not found', normalizedPath, location);
+            // The route is not registered here, so it should be handled by react-router
```
thornbill, line 127: All routes should be handled by react-router now, so no need to log ones that are not handled by app router.

src/index.jsx
```diff
@@ -39,6 +38,10 @@ import { currentSettings } from './scripts/settings/userSettings';
 import taskButton from './scripts/taskbutton';
 import App from './App.tsx';
 
+import './styles/livetv.scss';
+import './styles/dashboard.scss';
+import './styles/detailtable.scss';
```
thornbill, line 43: These styles were loaded in the routes file previously... for reasons...

nielsvanvelzen (approved):

PR #4389, opened 2023-03-02
Remove old Emby (3.x) version checks

JRaspass:
As per https://trello.com/c/sA25wd46/8-remove-old-emby-3x-version-checks

### DIFF
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -1156,12 +1156,7 @@ function renderMoreFromArtist(view, item, apiClient) {
     const section = view.querySelector('.moreFromArtistSection');
 
     if (section) {
-        if (item.Type === 'MusicArtist') {
-            if (!apiClient.isMinServerVersion('3.4.1.19')) {
-                section.classList.add('hide');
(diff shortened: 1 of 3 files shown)


src/controllers/itemDetails/index.js
```diff
@@ -1157,10 +1157,7 @@ function renderMoreFromArtist(view, item, apiClient) {
 
     if (section) {
         if (item.Type === 'MusicArtist') {
```
thornbill, line 1159: It looks like this `if` block can just be removed now.

JRaspass: I presume it would need folding into the next if, so something like?
```js
if (item.Type !== 'MusicArtist' && (item.Type !== 'MusicAlbum' || !item.AlbumArtists || !item.AlbumArtists.length)) {
```
holy length batman!

thornbill: Yeah that looks right. We don't enforce a line length limit (yet) so you can find some that are rather long. 😅

thornbill (approved):

PR #4391, opened 2023-03-04
Support markdown in item descriptions

nielsvanvelzen:
**Changes**
- Support markdown in item description
  - HTML is still supported
  - Mirrors Android TV app behavior

**Issues**

### DIFF
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -1,5 +1,6 @@
 import { intervalToDuration } from 'date-fns';
 import DOMPurify from 'dompurify';
+import { marked } from 'marked';
 import escapeHtml from 'escape-html';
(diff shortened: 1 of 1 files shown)


thornbill (approved):
I think this is fine. I know at one point we had discussed markdown support and didn't see much point since no metadata providers seem to support it, but we have since added markdown support to other areas. :man_shrugging:

nielsvanvelzen:
The metadata providers don't support it because web doesn't support it because the metadata providers don't support it because web doesn't support it because the metadata providers don't support it because web doesn't support it because...

There's also the case for manual editing, not all of our users use the metadata provided descriptions, or they modify them. Being able to add a few new lines to create a paragraph or making text bold easily is nice to have.

PR #4399, opened 2023-03-07
Update hls.js to 1.3.4

knackebrot:
This replaces https://github.com/jellyfin/jellyfin-web/pull/3623


thornbill:
It looks like hls.js will require being babelified now: https://github.com/jellyfin/jellyfin-web/blob/master/webpack.common.js#L149

thornbill (approved):

Issue #4402, opened 2023-03-08
Library Page Issues - LiveTV


jedthe3rd:
**Describe The Bug**
When I have Library Pages enabled and try to go to the next page in the On Now Live TV view (library?) it won't go to the next page. All the other Live TV views handle paging correctly. 
Then if I try to disable Library Pages it won't load all the content available. This is true for all Live TV views.

**Steps To Reproduce**
1. Go here: <server uri>/web/index.html#!/mypreferencesdisplay.html?userId=61a5e50717a440fd8fb0602a5a88d246
2. Change Library page size to 200 (might work with smaller values if you don't have that many on now channels).
3. Make sure you have over 200 (might work with smaller values if you don't have that many on now channels) On Now content.
4. Visit: <server uri>/web/index.html#!/list.html?type=Programs&IsAiring=true&serverId=dab9b10fde7541cbb6cfd3e5a46f201a
5. Hit the next page button and the page will reload without moving to the next page.
(url it calls: <serveruri>/LiveTv/Programs/Recommended?UserId=61a5e50717a440fd8fb0602a5a88d246&StartIndex=200&Fields=ChannelInfo%2CPrimaryImageAspectRatio%2CPrimaryImageAspectRatio&Limit=200&HasAired=false&IsAiring=true&SortBy=SortName&SortOrder=Descending&ImageTypeLimit=1")
6. Notice that the page didn't change.

**Expected Behavior**
Page advances to the next page with the rest of the On Now shows or if paging is 0 all content is shown on the first page.

**Logs**
n/a

**Screenshots**
Clicking -> doesn't do anything:
![image](https://user-images.githubusercontent.com/7817335/223829046-4110091b-7530-4ac6-a2ae-934556f15625.png)
Page Size: 0
![image](https://user-images.githubusercontent.com/7817335/223829812-5fb16413-b1bc-461b-a8fe-bca32127dc91.png)
Page Size: 1000
![image](https://user-images.githubusercontent.com/7817335/223830195-6153de9b-4f5d-4fee-9a38-1bac6808f4bb.png)
Page Size: 200
![image](https://user-images.githubusercontent.com/7817335/223831041-6607b102-5b59-4af7-9c0d-87319e539e2b.png)

**System (please complete the following information):**
 - Platform: Windows 10
 - Browser: All
 - Jellyfin Version: 10.8.9

**Additional Context**
I am loading some free IPTV content which is why I have so many channels.

[...]

PR #4406, opened 2023-03-12
Fix LiveTV lists pagination

jedthe3rd:
Remove use of LiveTvRecommendedPrograms api for LiveTV page lists. Increased library page size limit to 10K.

**Issues**
Fixes: #4402

### DIFF
--- a/src/components/displaySettings/displaySettings.template.html
+++ b/src/components/displaySettings/displaySettings.template.html
@@ -224,7 +224,7 @@
     </h2>
 
     <div class="inputContainer inputContainer-withDescription">
-        <input is="emby-input" type="number" id="txtLibraryPageSize" pattern="[0-9]*" required="required" min="0" max="10000" step="1" label="${LabelLibraryPageSize}" />
+        <input is="emby-input" type="number" id="txtLibraryPageSize" pattern="[0-9]*" required="required" min="0" max="1000" step="1" label="${LabelLibraryPageSize}" />
         <div class="fieldDescription">${LabelLibraryPageSizeHelp}</div>
     </div>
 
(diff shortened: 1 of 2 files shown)


src/components/displaySettings/displaySettings.template.html
```diff
@@ -224,7 +224,7 @@ <h2 class="sectionTitle">
     </h2>
 
     <div class="inputContainer inputContainer-withDescription">
-        <input is="emby-input" type="number" id="txtLibraryPageSize" pattern="[0-9]*" required="required" min="0" max="1000" step="1" label="${LabelLibraryPageSize}" />
+        <input is="emby-input" type="number" id="txtLibraryPageSize" pattern="[0-9]*" required="required" min="0" max="10000" step="1" label="${LabelLibraryPageSize}" />
```
thornbill, line 227: 10k is really not a practical limit for this. It will cause pages to load too slow especially on lower end servers.

jedthe3rd: Hmm, ya, I get that. I also get letting the user define a setting that could overload the server isn't good. I didn't really think about that. 

So 1000 is too small if you use the many free legal IPTV sources since it shows times for shows multiple times. Maybe a grouping system with sub lists might be a better option and grouping by show will make these lists more user friendly as the end goal is to find content you want to watch in the future doesn't really matter when it is playing. I will think on this and make a PR once I get a good solution. That is out of scope for this PR though so I am good with leaving the user defined max as it was.

thornbill: Is this just a workaround for pagination not working correctly then?

jedthe3rd: > Is this just a workaround for pagination not working correctly then?

No specific change was an enhancement on top of fixing the pagination. The other change in lists.js should be a fix for the issue I describe in my ticket. 

Actually I just realized that I missed adding the fix for user config librarypagesize = 0. Committing that now.

thornbill:
@jellyfin-bot rebase

PR #4408, opened 2023-03-12
fix(hls): playback should work without bundling hls

ThibaultNocchi:
**Changes**
Revert the webpack change to bundle hls.js from #4399 

**Issues**
Playback didn't work and raised a `_typeof` error.

video-dev/hls.js#1939 gave me the correct hint.


thornbill:
This is blocked by https://github.com/video-dev/hls.js/issues/5288

thornbill (approved):

PR #4431, opened 2023-03-18
Fix missing port number in German translation

fritzmg:
See https://github.com/jellyfin/jellyfin-media-player/issues/288#issuecomment-1474852988

### DIFF
--- a/src/strings/de.json
+++ b/src/strings/de.json
@@ -634,7 +634,7 @@
     "LabelSerialNumber": "Seriennummer:",
     "LabelSeriesRecordingPath": "Serien Aufnahmepfad:",
     "LabelServerHost": "Adresse:",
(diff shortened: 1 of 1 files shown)


fritzmg:
Ah, I just noticed that translations are managed here: https://translate.jellyfin.org/projects/jellyfin/jellyfin-web/
I am assuming this needs to be changed there instead?

thornbill:
Yes that is right. Any translation changes should be made via weblate unless it is a change in the primary language file (en-us).

fritzmg:
I submitted the change there (not sure if there is anything else to do, first time using weblate).

thornbill:
There shouldn't be anything else you need to do. It should be automatically committed to master the next time it syncs over.

PR #4435, opened 2023-03-20
Add video volume change on mouse scroll

robert-hamilton36:
**Changes**
Creates onWheel listener, which changes volume levels based on direction the mouse wheel is scrolled
Adds onWheel listener to document during 'viewshow' event
Removes onWheel listener from document during 'viewbeforehide' event

**Issues**

**Notes**
Found idea in [features requests](https://features.jellyfin.org/)
Assumes the video player always takes up 100% of the screen (Which I've always found to be the case)


thornbill (approved):

Karkunow:
@robert-hamilton36 @thornbill this functionality is impacting LG TV users. We have no sound, as you can see in the issue above. Could this be resolved?

PR #4446, opened 2023-03-26
Add Select context menu

robert-hamilton36:
**Changes**
1. multiSelect.js 
Adds external function to multiselect.js, which starts the multiselect feature on a given card.

2. itemContextMenu.js
Imports dom script to use dom.parentWithClass. This allows for copying of the same logic from multiSelect.js for selecting a card element(Which is ran on long clicks).
Adds the context menu command select, for all context menus positioned to a card component. 
Adds multiSelect case to execute command, which runs the startMultiSelect on the card the context menu is positioned to.

3. en-us.json
Add translation for 'select'

**Notes**
Found on trello board, which would need to be updated


src/components/itemContextMenu.js
```diff
@@ -116,6 +117,16 @@ import toast from './toast/toast';
             }
         }
 
+        // Multiselect is currrently only ran on long clicks of card components
```
thornbill, line 120: I don't believe the UI for the multiselect will work on TVs either. I would suggest adding this inside the check on line 102 and have it be the first item in this group (preceding add to collection and playlist).

thornbill (changes requested):
Looks like this works as expected! I just had one small suggestion.

src/components/itemContextMenu.js
robert-hamilton36: Thanks! Good catch, I didn't consider tvs.

I made the changes locally and then tried push back into my branch.
I think this was a mistake as its closed the pull request.

PR #4463, opened 2023-03-30
Fix previousTrack click eventlistener to avoid erroneous calls

### LINKED ISSUE #4462: Redundant calls to PlaybackManager in nowPlayingBar and remotecontrol previousTrack eventhandler
I am working on a player plugin, but the previous track function in jellyfin-web for a music album appears to have a flaw.

Double clicking the previous track button is captured by the two previousTrackButton eventlisteners in nowPlayingBar.js, for 'click' and 'dblclick'. When debugging, it appears that this results in three calls to PlaybackManager.previousTrack in stead of just one.

This is due to PlaybackManager.previousTrack method being called in two 'if' statements in the 'previousTrackButton' 'click' eventlistener in nowPlayingBar.js:

- https://github.com/jellyfin/jellyfin-web/blob/1451e4ece03a8c35ac7c456b96a65530143649aa/src/components/nowPlayingBar/nowPlayingBar.js#L172
- https://github.com/jellyfin/jellyfin-web/blob/1451e4ece03a8c35ac7c456b96a65530143649aa/src/components/nowPlayingBar/nowPlayingBar.js#L174

It looks like the eventlistener expects the PlaybackManager.previousTrack method in those 'if' statements to just return true if a previous track exists, but that is not what this method does: PlaybackManager.previousTrack is not a function with a return value, but performs the operation of loading and playing the previous track.

Note that this eventlistener also calls PlaybackManager.previousTrack as expected:
- https://github.com/jellyfin/jellyfin-web/blob/1451e4ece03a8c35ac7c456b96a65530143649aa/src/components/nowPlayingBar/nowPlayingBar.js#L182

The same issue occurs in remotecontrol.js, as this contains the same code.

**Steps To Reproduce**
I reproduced this behavior of redundant calls by adding console.debug stmts in the concerning eventhandlers.

**Expected Behavior**
The solution is to adapt the above 'if' statements to perform the check for the presence of a previous track in a different way, which seems quite easy to do. In that case PlaybackManager.previousTrack is also called just once, in the way it is expected to be called.  I can submit a PR for that.

 - Platform: Windows 10
 - Browser: Edge
 - Jellyfin-web current github version (10.9.0)

**Additional Context**
This is my first PR, issue so please let me know if I need to take additional steps.

Sky-High:
**Changes**
Have previousTrack eventlistener use playbackManager.getCurrentPlaylistIndex() to check for current track number, in stead of erroneously calling playbackManager.previousTrack for that.

**Issues**
Fixes #4462


thornbill:
@Sky-High it's not a big deal, but there isn't typically a need to keep merging `master` back into your branch unless a merge conflict comes up. :slightly_smiling_face:

src/components/nowPlayingBar/nowPlayingBar.js
```diff
@@ -169,18 +169,22 @@ import { appRouter } from '../appRouter';
...
-                    // Cancel this event if doubleclick is fired
-                    if (e.detail > 1 && playbackManager.previousTrack(currentPlayer)) {
+                if (lastPlayerState.NowPlayingItem.MediaType === 'Audio') {
+                    // Cancel this event if doubleclick is fired. The actual previousTrack will be processed by the 'dblclick' event
+                    if (e.detail > 1 ) {
+                        return;
+                    }
+                    // Return to start of track, unless we are already (almost) at the beginning. In the latter case, continue and move
+                    // to the previous track, unless we are at the first track so no previous track exists.
+                    if (currentPlayer._currentTime >= 2 || playbackManager.getCurrentPlaylistIndex(currentPlayer) <= 1) {
```
thornbill, line 179: Why the change from 5 to 2 here? That seems like it could be a little low.

Sky-High: Ok, thanks for that! My interpretation was that the 5 seconds from the start of a tune are meant to 'simulate' a long 'doubleclick', but it felt like a little too long to me as doubleclick.
But maybe I am just impatient, or maybe there is a technical reason for a 'long' doubleclick, so no problem of course to use 5 seconds after all.
I'll submit an update.

thornbill (approved):

PR #4464, opened 2023-03-30
Update centos docker image from 8 to stream8

cewert:
I noticed the [build for centos is failing](https://dev.azure.com/jellyfin-project/jellyfin/_build/results?buildId=44659&view=logs&jobId=5404562b-245d-5392-77db-b338601c3067&j=5404562b-245d-5392-77db-b338601c3067&t=221a3f35-6cfc-5f51-3267-0fbd49d6df2c) because it is still [using the `centos:8` image](https://github.com/jellyfin/jellyfin-web/blob/master/deployment/Dockerfile.centos#L1) instead of the newer `centos:stream8` image.

This PR probably needs more work but since this repo uses azure for CI instead of github I wasn't able to test them on my fork 🤷 

**Changes**
- Use `centos:stream8` docker image instead of deprecated `centos:8` image


joshuaboniface (approved):

cewert:
Well that was way easier than I thought it would be 🎉

PR #4477, opened 2023-04-06
Remove all trailing colons from translations

nadir-yttrium:
**Changes**
I removed all trailing colons from the translation files as requested on the [jellyfin-web roadmap](https://trello.com/c/QcdCFU8o)

I wasn't quite sure if this is actually everything that was requested or if adding the colons back in in code was implied to be part of the task. 

As colons are currently - as mentioned on the roadmap - pretty inconsistently applied, I figured it wouldn't make much sense to add them all  back in without some clear guidance on when that's appropriate.


thornbill:
Translation changes should only be made to the default language file (en-us). The other languages are managed through our weblate instance.

nadir-yttrium:
That makes sense. Fixed so only en-us gets changed.

thornbill (approved):

PR #4491, opened 2023-04-13
Add Ratings to both Video Player and Up Next Dialog Window

FireGamer3:
**Changes**
This Adds Ratings to the Video Player OSD. and adds ratings to the Up Next Dialog window.

**Issues**
[Roadmap issue](https://trello.com/c/EGDmjVtE/55-add-rating-to-video-player-osd-and-next-up-popover)
[Feature Request](https://features.jellyfin.org/posts/1094/show-episode-movie-rating-while-playing)


FireGamer3:
image
image

thornbill:
That's awesome @FireGamer3! Just be patient with me, I'm a little behind on reviews at the moment. 😅

src/controllers/playback/video/index.js
```diff
@@ -733,6 +734,24 @@ import { PluginType } from '../../../types/plugin.ts';
             return (currentTimeMs - programStartDateMs) / programRuntimeMs * 100;
         }
 
+        function updateRatingDisaply(itemInfo) {
```
grafixeyehero, line 737: no need to add new function here you can use mediaInfo.getMediaInfoHtml(item, {
                officialRating: false,
                criticRating: true,
                starRating: true,
                endsAt: false,
                year: false,
                programIndicator: false,
                runtime: false,
                subtitles: false
            });

FireGamer3: Ill give it a try to see if I can work it in there, I originally opted for this way to have custom css classes that were specifically labeled for the osd, but I could add the mediainfo classes to the osd scss instead

FireGamer3: Alright,

I modified it to use mediaInfo, and adjusted the scss file to look for the mediainfo classes

FireGamer3: sorry will have one more as I had some css classes I could remove since they were not really being used

grafixeyehero: yes u can use osdMediaInfo classes

FireGamer3: Alright It seems like everything on this commit is good, everything lines up and works properly, hides with the time when the viewport is small, and I didn't have to worry about a brand new function with slightly odd behavior ha ha
I added
```
originalAirDate: false,
episodeTitle: false
```
to the list of options as when I went to a show I had, it showed the original air date so I went through the function to find any missing options and these were it.

grafixeyehero (approved):

PR #4494, opened 2023-04-15
Add Collections list to Item Details

FireGamer3:
**Changes**
Adds Collections to an Item detail page. I saw this todo and started messing around with it (and with it learning a bit of the API client). 

So I tried to get a list of BoxSets with the parent ID set to the item being rendered however that didn't work and just returned all box sets so I just looped through them, if this is something we want to wait for on the server side so that getting this information is easier and not have to do a bunch of API calls on the client to search through the collections then let me know and I can close this PR until that is done.

I also assumed that only Movies, Trailers, Series, Programs, Recordings, MusicAlbums, MusicArtists, and Playlists were what you can add to a collection, let me know if I need to add or remove any content types

**Issues**
[Roadmap Issue](https://trello.com/c/cH7IMbst/47-add-collections-to-item-details-screen)
[Feature Request](https://features.jellyfin.org/posts/540/list-all-collections-that-a-movie-belong-to-in-movie-details)

### DIFF
--- a/src/controllers/itemDetails/index.html
+++ b/src/controllers/itemDetails/index.html
@@ -231,6 +231,13 @@
                     </div>
                 </div>
 
+                <div id="collectionsItemIsInCollapsible" class="verticalSection detailVerticalSection hide">
+                    <h2 class="sectionTitle sectionTitle-cards padded-right">${Collections}</h2>
+                    <div is="emby-scroller" class="padded-top-focusscale padded-bottom-focusscale" data-centerfocus="true">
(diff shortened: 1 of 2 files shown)


FireGamer3:
image

FireGamer3:
Yea now that I'm thinking about it I may just sit on this one a while, its not 100% foolproof right now because of nesting promises lol so there could be instances where this could not work and it would be random on if it worked or not.

thornbill:
I fully expected this to be available via the api, but you are absolutely right that it doesn't seem to be currently. 😅

PR #4496, opened 2023-04-15
Fix info header not updated on version change

### LINKED ISSUE #4495: Info-Header is not updated when version of multiple version movies is changed.
**Describe The Bug**
The "Info-Header" only shows the information (year, duration, 3D flag, rating, end time etc.) of the primary item of multi version movies.
If the version is changed the displayed data stays the same, even if the other version has a different playtime or is a 3D version of the movie.
Since the selection for available videos, audio and subtitles is updated at least some of the data is already present and send to the frontend.

**Steps To Reproduce**
1. Have a movie with multiple version that differ in playtime or 2D/3D versions
2. Change the version
3. duration and 3D flag doesn't change

**Expected Behavior**
displayed information should match the selected version

**Screenshots**
Duration of the "Extended Cinema Cut" should be `2h 51m` instead of `2h 58m` which is the duration of the "Extended Collector's Cut"
jellyfin_infoheader

 - Platform: Docker
 - Browser: Firefox
 - Jellyfin Version: 10.8.9

TheMelmacian:
**Changes**
Re-render miscInfo if the selected version (mediaSource) changes.
To provide the updated data the primary item (currentItem) is cloned and updated with the data from the selected MediaSource (via destructuring).
Data that is not present in the MediaSource item keeps its original value from the primary item.
Added a `getSelectedMediaSource` function to minimize code duplication.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-web/issues/4495

### DIFF
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -113,6 +113,11 @@ function getProgramScheduleHtml(items, action = 'none') {
     });
 }
 
+function getSelectedMediaSource(page, mediaSources) {
+    const mediaSourceId = page.querySelector('.selectSource').value;
+    return mediaSources.filter(m => m.Id === mediaSourceId)[0];
+}
+
 function renderSeriesTimerSchedule(page, apiClient, seriesTimerId) {
     apiClient.getLiveTvTimers({
         UserId: apiClient.getCurrentUserId(),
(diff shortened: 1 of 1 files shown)


src/controllers/itemDetails/index.js
```diff
@@ -113,6 +113,11 @@ function getProgramScheduleHtml(items, action = 'none') {
     });
 }
 
+function getSelectedMediaSource(page, mediaSources) {
```
thornbill, line 116: :+1: Nice job pulling out this duplicated code!

thornbill (approved):
:rocket: lgtm!

PR #4507, opened 2023-04-17
Add guest cast on episode detail page

### LINKED ISSUE #4148: Episode detail page doesn't show guest stars
On detail pages for TV episodes, only people with the "Actor" role are shown, not people with the "GuestStar" role.

I can open a PR for this (I already have it working locally) but there are two ways to do it, and I'm not sure which you would prefer.

1. Add another "Guest Cast" section before or after the "Cast & Crew" section. I like this better because I don't need to click the arrows to "scroll" the main cast section over to the guest stars, and usually I am more interested in the guest stars than the main cast on a specific episode, since I already know who the main cast are. For the same reason I prefer the additional section to appear first, like on TVmaze, but it could also come after.

2. Show guest stars in the "Cast & Crew" section after the main cast actors.

fffrankieh:
**Changes**
I added a separate Guest Cast section below Cast & Crew on TV episode detail pages. Before, people marked as Guest Star were not shown at all.

**Issues**
Fixes #4148

### DIFF
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -751,9 +751,6 @@ function setPeopleHeader(page, item) {
     } else {
         page.querySelector('#peopleHeader').innerHTML = globalize.translate('HeaderCastAndCrew');
     }
-    if (item.Type === 'Episode') {
-        page.querySelector('#guestCastHeader').innerHTML = globalize.translate('HeaderGuestCast');
-    }
 }
 
 function renderNextUp(page, item, user) {
@@ -1810,9 +1807,7 @@ function renderCast(page, item) {
 }
 
 function renderGuestCast(page, item) {
-    const people = (item.People || []).filter(function (p) {
-        return p.Type === 'GuestStar';
-    });
+    const people = (item.People || []).filter(p => p.Type === 'GuestStar');
 
     if (!people.length) {
         page.querySelector('#guestCastCollapsible').classList.add('hide');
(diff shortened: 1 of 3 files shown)


thornbill:
I did a bit of digging and the missing guest stars is actually the result of an (almost) 3 year old bug introduced by another contributor. The person type filter used to only exclude directors since they are listed elsewhere on the page. This also broke displaying people for other media (music, books, etc) that have different people types.

I'm fine with guest stars having a dedicated section, but we should probably fix `renderCast` also.

src/strings/en-us.json
```diff
@@ -373,6 +373,7 @@
     "HeaderFetchImages": "Fetch Images",
     "HeaderForKids": "For Kids",
     "HeaderFrequentlyPlayed": "Frequently Played",
+    "HeaderGuestCast": "Guest Cast",
```
thornbill, line 376: I suspect you went with "Guest Cast" here because the other section is "Cast & Crew", but that section needs overhauled, so let's just go with "Guest Stars"

src/controllers/itemDetails/index.js
```diff
@@ -1821,11 +1825,36 @@ function renderCast(page, item) {
     });
 }
 
+function renderGuestCast(page, item) {
+    const people = (item.People || []).filter(function (p) {
+        return p.Type === 'GuestStar';
+    });
```
thornbill, line 1831:
```suggestion
    const people = (item.People || []).filter(p => p.Type === 'GuestStar');
```

thornbill (changes requested):
Seems to work as expected, just a couple code suggestions. :slightly_smiling_face:

src/strings/en-us.json
thornbill, line 376:
```suggestion
    "HeaderGuestCast": "Guest Stars",
```

thornbill (approved):

PR #4509, opened 2023-04-18
Fix centos docker build maybe

thornbill:
**Changes**
At least locally this seems to fix the centos docker builds... I have no idea why those packages were needed or if they should be replaced with something else, but it built for me at least :man_shrugging: 

**Issues**
N/A


crobibero (approved):
#yolo

PR #4511, opened 2023-04-19
Remove node max space setting

thornbill:
**Changes**
Removes the configuration of the `max_old_space_size` from the node build scripts and adds it to CI configuration as needed. This should be set locally based on the available ram of an individual system if needed.

**Issues**
N/A


nielsvanvelzen (approved):

thornbill:
I'm going to close this for now. This may help avoid errors with people getting started with web (although if they have <6GB of ram it will not). I think we need to take a deeper dive into our tooling... using some more modern build tooling might help remove the excessive ram usage. :crossed_fingers:

PR #4533, opened 2023-04-24
Replace babel-loader with swc-loader

thornbill:
**Changes**
Replaces babel-loader with swc-loader for faster build times. The build output passes our ES5 compatibility check and works fine in a modern version of Firefox, but it would probably be good to some testing in webOS/Tizen specifically.

**Results**
Current master with `max_old_space_size` limitation removed:
```
webpack 5.79.0 compiled with 3 warnings in 168598 ms
```

This PR:
```
webpack 5.79.0 compiled with 3 warnings in 84066 ms
```

**Issues**
N/A


nielsvanvelzen (approved):

thornbill:
In webOS 1.2 (WebKit) there is a bug with equal names of functions and arguments in strict mode:

```
createUserByName: function createUserByName (createUserByName,
options) {
```

Babel wasn't in a strict mode, afaik

```
SyntaxError:
Cannot declare a parameter
named 'createUserByName' in strict mode
```

Also, `includes` function requires a polyfill.

ndom91:
I couldn't comment on the roadmap issue for whatever reason, so sorry for posting this here. 

But what do you think about swapping webpack for vite? I also have an experimental branch open locally that's got most of it up and running

ferferga:
@ndom91 Just by chance I came across this comment.

Absolutely open a PR for this if you have it already running! We discussed in chat a few times about pursuing this migration, but we weren't aware anyone already had it! I recall @thornbill had some troubles when he quickly attempted it in the past, maybe with your part everything glues together!

thornbill:
Whoops I had definitely overlooked the last comment here. Sorry!

The issue I had with vite (and basically every other modern bundler) was they failed to support the _very old_ browser engines that we target for TV platform support. Always happy to reevaluate if that has changed though! At the time this PR was created we didn't use the root import alias yet iirc.

I've also been curious if rspack might work better for us since I believe the transpiling is still managed by babel... but I've only briefly looked at the documentation for rspack so I might be wrong there.

I'm going to close this PR since swc didn't work out, but feel free to continue the discussion here. 😉

ndom91:
@thornbill just fyi - you can tell Vite (using esbuild to build) to target any platform, see: https://vitejs.dev/config/build-options

Takes a browserlist type array of options like `[ 'chrome33', 'safari4', 'node4.1' ]`, etc.

thornbill:
We have to load some manual polyfills for really old versions of Tizen/webos but I don't recall that being the issue. I would have to see if I have any notes from my prior attempts.

ndom91:
Yeah no worries, just figured I'd pass some more info along :+1:

sakgoyal:
did anyone check if the `plugin-legacy` will work for the old browsers?

thornbill:
Yes my previous attempt did use the legacy plugin.

PR #4534, opened 2023-04-24
Seamless playing in multipart episodes

sleepycatcoding:
**Changes**
This makes behavior with multiparts episode playing behavior consistent with movies. Previously only movies auto played other parts.

**Issues**
None


thornbill:
Could you provide some instructions on what type of file setup would be needed to test this? Thanks!

sleepycatcoding:
The file setup I used to test this was:

PR #4338, opened 2023-02-08
Add "AllowedTags" option to parental controls

Jpuc1143:
**Changes**
Adds an option in Parental Controls to only allow media with selected tags for viewing.

**Issues**
Client side configuration of jellyfin/jellyfin#9139

PR #4535, opened 2023-04-25
 Add "AllowedTags" option to parental controls

davidangel:
(copied from https://github.com/jellyfin/jellyfin-web/pull/4338 - updates to frontend + merge conflict resolution)

Changes
Adds an option in Parental Controls to only allow media with selected tags for viewing.

Issues
Client side configuration of https://github.com/jellyfin/jellyfin/pull/9139

### DIFF
--- a/src/apps/dashboard/routes/users/parentalcontrol.tsx
+++ b/src/apps/dashboard/routes/users/parentalcontrol.tsx
@@ -6,7 +6,7 @@ import escapeHTML from 'escape-html';
 import globalize from '../../../../scripts/globalize';
 import LibraryMenu from '../../../../scripts/libraryMenu';
 import AccessScheduleList from '../../../../components/dashboard/users/AccessScheduleList';
-import BlockedTagList from '../../../../components/dashboard/users/BlockedTagList';
+import TagList from '../../../../components/dashboard/users/TagList';
 import ButtonElement from '../../../../elements/ButtonElement';
 import SectionTitleContainer from '../../../../elements/SectionTitleContainer';
 import SectionTabs from '../../../../components/dashboard/users/SectionTabs';
@@ -16,6 +16,7 @@ import { getParameterByName } from '../../../../utils/url';
 import CheckBoxElement from '../../../../elements/CheckBoxElement';
 import SelectElement from '../../../../elements/SelectElement';
 import Page from '../../../../components/Page';
+import prompt from '../../../../components/prompt/prompt';
 
 type UnratedItem = {
     name: string;
@@ -28,6 +29,7 @@ const UserParentalControl: FunctionComponent = () => {
     const [ parentalRatings, setParentalRatings ] = useState<ParentalRating[]>([]);
     const [ unratedItems, setUnratedItems ] = useState<UnratedItem[]>([]);
     const [ accessSchedules, setAccessSchedules ] = useState<AccessSchedule[]>([]);
+    const [ allowedTags, setAllowedTags ] = useState([]);
     const [ blockedTags, setBlockedTags ] = useState([]);
 
     const element = useRef<HTMLDivElement>(null);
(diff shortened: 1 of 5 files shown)


adoreparler:
@davidangel  you still working on this? Has the server side been updated?

Shadowghost:
Server functionality was already merged into master

adoreparler:
Sorry new to this project. How do we get it through the SonarCloud check?

adoreparler:
Testing this out and it does not seem to be working. I added a tag to the allowed tags section for another user. Logged in as that user and they still see everything. I will keep testing and look at the code

adoreparler:
Seeing everything was a cached thing, I refreshed and it disappeared. But its still not working, I add a tag to a video and to the allowed tags for another user and log in with them and refresh and nothing shows. Will keep playing around

xangelix:
I have not tested this PR but suspect it is operating on the same logic here https://github.com/jellyfin/jellyfin-web/pull/4338#issuecomment-1483906272
You have to also add the allowed tag to each containing layer, such as the library itself.

davidangel:
Updated this. I'm testing against my own "prod" backend, so `AllowedTags` doesn't seem to be populating, but I think this should be working when the backend provides this data.

adoreparler:
I'll try and pull a fresh copy of the back end and this PR later tonight and test. 

It was not working when I tested it previously. I was testing this at a per item basis, so add a new tag like "kids" to a user allowed tags, and then add that tag to a bunch of videos and then login as that user and those videos tagged with kids should only show up for that user.

Siress:
Thank you for working on this!

src/apps/dashboard/routes/users/parentalcontrol.tsx
```diff
@@ -270,31 +270,59 @@ const UserParentalControl: FunctionComponent = () => {
...
+            }).then(function (value) {
+                const tags = getAllowedTagsFromPage();
+
+                if (tags.indexOf(value) == -1) {
+                    tags.push(value);
+                    loadAllowedTags(tags);
+                }
+            }).catch(() => {
+                // prompt closed
+            }).catch(err => {
```
thornbill, line 291: This `catch` would never get called right? Since the dynamic `import` has been removed, it can probably just be removed.

src/apps/dashboard/routes/users/parentalcontrol.tsx
```diff
@@ -270,31 +270,59 @@ const UserParentalControl: FunctionComponent = () => {
...
+            }).then(function (value) {
+                const tags = getBlockedTagsFromPage();
+
+                if (tags.indexOf(value) == -1) {
+                    tags.push(value);
+                    loadBlockedTags(tags);
+                }
+            }).catch(() => {
+                // prompt closed
             }).catch(err => {
```
thornbill, line 314: Same thing here.

src/strings/en-gb.json
```diff
@@ -1007,8 +1007,9 @@
     "LabelArtistsHelp": "Separate multiple artists with a semicolon.",
     "LabelArtists": "Artists",
     "LabelAppName": "App name",
-    "LabelAllowedRemoteAddressesMode": "Remote IP address filter mode",
-    "LabelAllowedRemoteAddresses": "Remote IP address filter",
+    "LabelAllowContentWithTags": "Allow items with tags:",
+    "LabelAllowedRemoteAddressesMode": "Remote IP address filter mode:",
+    "LabelAllowedRemoteAddresses": "Remote IP address filter:",
```
thornbill, line 1012: Only the `en-us` language should be updated in PRs otherwise it can cause conflicts with weblate (translate.jellyfin.org).

thornbill (changes requested):

girishso:
Not working, I get an empty media library.  I add a tag to a video and to the allowed tags for another user and log in with them and refresh and nothing shows.

thornbill (approved):
These frontend changes seem to work as expected, although it did expose a bug with the server behavior.

PR #4481, opened 2023-04-09
revert asyncRoutes to legacyRoutes

grafixeyehero:
Preparing for experimental layout mode 
@thornbill  as we discuses https://github.com/jellyfin/jellyfin-web/pull/4176#issuecomment-1474486651

PR #4556, opened 2023-05-02
Cleanup routes

thornbill:
**Changes**
* Moves `appRouter` and `HistoryRouter` to `components/router`
* Moves all routes in `routes/` to the appropriate stable/experimental app directory
* Partially reverts the user management route changes from #4481 since we already use the react versions in 10.8.x (see https://github.com/jellyfin/jellyfin-web/blob/release-10.8.z/src/scripts/routes.js#L438-L483)

**Issues**
N/A


thornbill:
@grafixeyehero I'm curious what you think about this one. It seems fine to me to use the react views for the user management stuff since it is already out in 10.8, but I would like to know if you have a different opinion. :slightly_smiling_face:

grafixeyehero:
@thornbill its fine with me. since it doesn't affect our previous conversation👌👌👌

joshuaboniface (approved):

cvium (approved):

PR #4581, opened 2023-05-08
Fix white background on video refresh

Xavier-Do:
When refreshing a /video page, since no player is available, `getCurrentPlayer` will be empty and `bindToPlayer` will crash leading to a redirect with `appRouter.goHome();`
This is also visible when doing back then next in the browser.

Before that that, the background is set in `viewbeforeshow` using `setBackdropTransparency(TRANSPARENCY_LEVEL.Full);`

This will leave unexpected classes in the dom leading to a white background on the main page.

This fixes proposes to reset the transparency before going to home using `setBackdropTransparency(TRANSPARENCY_LEVEL.None);` Anyway, this looks a little fragile and a deeper fix may be preferable in the future.

Another solution would be to move the
`setBackdropTransparency(TRANSPARENCY_LEVEL.Full);` when the player is ready but my understanding of the flow is not enough to now if it would have any unexpected results.

### DIFF
--- a/src/controllers/playback/video/index.js
+++ b/src/controllers/playback/video/index.js
@@ -1496,6 +1496,7 @@ export default function (view) {
                 dom.addEventListener(document, 'click', onClickCapture, { capture: true });
             }
         } catch (e) {
+            setBackdropTransparency(TRANSPARENCY_LEVEL.None); // reset state set in viewbeforeshow
             appRouter.goHome();
         }
     });


thornbill (commented):
I think this is fine for now. Ideally we would be able to restore the video player state when refreshing, but that will probably not happen until it is essentially rewritten.

I would appreciate it if you could add a comment in the code explaining why this is necessary though. Thanks!

Xavier-Do:
> I would appreciate it if you could add a comment in the code explaining why this is necessary though. Thanks!

I can do that. I was also wondering if it shouldn't be in the home `viewbeforeshow` but i'm not familiar enough with the general lifeflow of jellyfin/react.

> I think this is fine for now. Ideally we would be able to restore the video player state when refreshing, but that will probably not happen until it is essentially rewritten.

I'm actually working on this but since I'm still learning jelly-fin architecture I may need some help to make it work will keeping it clean. Just opened a pr on my fork wince it is still a wip. https://github.com/Xavier-Do/jellyfin-web/pull/1/files

thornbill (approved):

PR #4599, opened 2023-05-13
Add episode selector in video player

DualScorch:
**Changes**
Adds a episode and season selector accessible from the player.
Support for desktop, mobile and TV.
 
desktop_episodes
desktop_seasons
tv_episodes
tv_seasons

**Issues**
N/A


src/components/playback/playerepisodeselector.js
thornbill, line 1: TypeScript would be preferred for new files if possible.

PR #4613, opened 2023-05-17
fix: update JASSUB

ThaUnknown:
**Changes**
This PR updates JASSUB to latest.

This adds a new [optional] modern WASM module, which adds 2.1MB to the bundle size. This module adds SIMD support [TLDR fast math operations] which libass utilizes to calculate blur quicker, this makes blur calculations almost 40% faster, and on blur heavy TS the render times can be sped up by anywhere from 15% to 35%.

If the user's browser doesn't support SIMD it will disable blur, on typesetting, this is a little bit aggressive, but it brings heavy TS rendering performance well over what the SIMD WASM can do. [Should this be removed?]

Fixes offscreen rendering for firefox.

Adds support for colorspace mangling. If the subtitle's colorspace didn't match that of the current video's the subtitles would often be incorrectly colored, leading to poor masking, this is now supported and handled correctly:
image
image

**Issues**
https://github.com/jellyfin/jellyfin-web/pull/4585

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,7 +36,7 @@
         "history": "5.3.0",
         "hls.js": "1.4.3",
         "intersection-observer": "0.12.2",
-        "jassub": "1.7.0",
+        "jassub": "1.7.1",
         "jellyfin-apiclient": "1.10.0",
         "jquery": "3.7.0",
         "jstree": "3.3.15",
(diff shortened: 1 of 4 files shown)


thornbill:
@jellyfin-bot rebase

(rebasing to trigger stuck GH action)

thornbill (approved):

PR #4615, opened 2023-05-18
Extract display mirror handling from component

thornbill:
**Changes**
Moves the display mirror logic out of the component to remove the dependency on the component being loaded for display mirroring to function. Also cleans up some promise usage warnings and unnecessary function calls.

**Issues**
N/A

### DIFF
--- a/src/components/playback/playerSelectionMenu.js
+++ b/src/components/playback/playerSelectionMenu.js
@@ -1,4 +1,3 @@
-import appSettings from '../../scripts/settings/appSettings';
 import Events from '../../utils/events.ts';
 import browser from '../../scripts/browser';
 import loading from '../loading/loading';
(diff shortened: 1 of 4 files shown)


src/components/playback/playerSelectionMenu.js
```diff
@@ -140,10 +113,14 @@ export function show(button) {
                 })[0];
 
                 playbackManager.trySetActivePlayer(target.playerName, target);
-
-                mirrorIfEnabled();
```
thornbill, line 144: Calling `mirrorIfEnabled` with no parameters does nothing.

src/components/playback/playerSelectionMenu.js
```diff
@@ -287,23 +268,6 @@ function onAutoCastChange() {
...
-    if (item && item.ServerId) {
-        mirrorIfEnabled({
-            item: item
-        });
-    }
-});
-
-Events.on(appSettings, 'change', function (e, name) {
-    if (name === 'displaymirror') {
-        mirrorIfEnabled();
```
thornbill, line 303: Same thing here... calling `mirrorIfEnabled` with no parameters does nothing.

joshuaboniface (approved):

PR #4632, opened 2023-05-22
Improve keyboard control

dmitrylyzo:
https://github.com/jellyfin/jellyfin-webos/issues/125#issuecomment-1411588082

**Changes**
- Don't show OSD for `Fullscreen` and `Mute`.
- Focus on corresponding button.
_Samsung checklist requires this, iirc._
- Make keys multi-purpose:
  - `Left/Right` activates dragging the position slider when OSD is hidden.
  - `Enter` pauses the playback when OSD is hidden.
- Add `KeyboardEvent` constructor polyfill.
`KeyboardEvent` is not constructible in WebKit (webOS 1.2/2.0, Tizen 2.3/2.4) - `KeyboardEventConstructor` is an object (not a function).
I have tested a few existing polyfills, but they rely on `initKeyboardEvent` method, which is broken in WebKit - it doesn't populate some properties (`keyCode`, `which`) that are read-only.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-webos/issues/125

_Feature_ or _Bugfix_? :thinking:

### DIFF
--- a/src/controllers/playback/video/index.js
+++ b/src/controllers/playback/video/index.js
@@ -1056,6 +1056,21 @@ import { appRouter } from '../../../components/appRouter';
                 return;
             }
 
+            if (layoutManager.tv && !currentVisibleMenu) {
+                // Change the behavior of some keys when the OSD is hidden
+                switch (key) {
+                    case 'ArrowLeft':
+                    case 'ArrowRight':
+                        showOsd(nowPlayingPositionSlider);
+                        nowPlayingPositionSlider.dispatchEvent(new KeyboardEvent(e.type, e));
+                        return;
+                    case 'Enter':
+                        playbackManager.playPause(currentPlayer);
+                        showOsd(btnPlayPause);
+                        return;
+                }
+            }
+
             if (layoutManager.tv && keyboardnavigation.isNavigationKey(key)) {
                 showOsd();
                 return;
--- a/src/controllers/playback/video/index.js
+++ b/src/controllers/playback/video/index.js
@@ -1,4 +1,5 @@
 import escapeHtml from 'escape-html';
+import debounce from 'lodash-es/debounce';
 import { playbackManager } from '../../../components/playback/playbackmanager';
 import SyncPlay from '../../../components/syncPlay/core';
 import browser from '../../../scripts/browser';
@@ -217,9 +218,9 @@ import { appRouter } from '../../../components/appRouter';
 
         let mouseIsDown = false;
 
-        function showOsd() {
+        function showOsd(focusElement) {
             slideDownToShow(headerElement);
-            showMainOsdControls();
+            showMainOsdControls(focusElement);
             resetIdle();
         }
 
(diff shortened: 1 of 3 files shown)


src/controllers/playback/video/index.js
```diff
@@ -1125,7 +1147,7 @@ import { appRouter } from '../../../components/appRouter';
                     // Ignores gamepad events that are always triggered, even when not focused.
                     if (document.hasFocus()) { /* eslint-disable-line compat/compat */
                         playbackManager.fastForward(currentPlayer);
-                        showOsd();
+                        showOsd(btnFastForward);
                     }
                     break;
```
dmitrylyzo, line 1152: Hmm, we can probably remove the gamepad bindings as they are mapped to the arrow keys.

src/controllers/playback/video/index.js
```diff
@@ -1035,26 +1040,46 @@ import { appRouter } from '../../../components/appRouter';
...
                 return;
             }
 
+            if (layoutManager.tv && !currentVisibleMenu) {
+                // Change the behavior of some keys when the OSD is hidden
+                switch (key) {
+                    case 'ArrowLeft':
+                    case 'ArrowRight':
+                        showOsd(nowPlayingPositionSlider);
+                        nowPlayingPositionSlider.dispatchEvent(new KeyboardEvent(e.type, e));
```
dmitrylyzo, line 1065: As an alternative to the event constructor polyfill, we can use a wrapper to create an event.

thornbill (commented):
This seems a little large to port back to 10.8. Do you think it is safe enough?

dmitrylyzo (commented):
I think it's pretty safe (_but now I'm thinking about small changes below_). The most "dangerous" part is KeyboardEvent polyfill: it is only required for old WebKit and should not be used anywhere else.

I tested it on webOS 1.2 and Chrome 114 (Firefox 🤔 don't remember). It would be nice if someone tested it on Safari (iOS and MacOS).

Since we have different indents in the master, I was planning to port it manually, keeping commits separately.

If you think it should be considered as a feature (altering control on TV, focusing corresponding button), we can defer it for 10.9

thornbill:
Ok yeah I'm fine with merging it to 10.8 then and I can do some testing on iOS Safari.

thornbill:
I tested playback in Safari on iOS 16 and did not notice any issues.

thornbill (approved):
:+1: code changes look fine to me. Just let me know if it is ready to merge or if you are still doing work/testing.

dmitrylyzo:
> 👍 code changes look fine to me. Just let me know if it is ready to merge or if you are still doing work/testing.

Yesterday I tested it again, and it may not be the best behavior:
1. Every time you perform an action, the focus jumps there and `Enter` clicks on the last action performed.
2. When you press `Space` while the progress slider is focused, the focus jumps to the `Pause` button and you will have to navigate to the slider again.

Samsung's built-in player behaves the same way, but we can remove focus movement when the OSD is already visible by removing this block: https://github.com/jellyfin/jellyfin-web/blob/11ae2ff43ffe2fdd563bc39dac24f0cfb5f7c3d3/src/controllers/playback/video/index.js#L293-L295

Also, it's "fun" to use this in Desktop mode - we don't have any focus indication. But similar behavior already exists - clicking buttons with the mouse moves focus to the last button clicked.

dmitrylyzo:
If there are no complaints about the current behavior (https://github.com/jellyfin/jellyfin-web/pull/4632#issuecomment-1720063829), it is ready to merge.

thornbill:
👍 even with the quirks it seems like an improvement over the current state

PR #4642, opened 2023-05-24
Fix movies Shuffle button

davidmurdoch:
The Movies `Shuffle` button doesn't do anything anymore, as `topParentId` is `undefined`, which isn't permitted by `ApiClient.getItem`. I've fixed it by using the `SortBy: Random` query option and now pick the first item from the list that it returns.

### DIFF
--- a/src/controllers/movies/movies.js
+++ b/src/controllers/movies/movies.js
@@ -30,11 +30,11 @@ export default function (view, params, tabContent, options) {
     }
 
     function shuffle() {
-        ApiClient.getItem(
-            ApiClient.getCurrentUserId(),
-            params.topParentId
-        ).then((item) => {
-            playbackManager.shuffle(item);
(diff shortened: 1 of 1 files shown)


dmitrylyzo:
Checked the unstable (8ffe7fec2feeb334c867c9eb477fa1678238b317): `topParentId` is the id of the library, and shuffle works (unless a bad item is caught). :man_shrugging:

davidmurdoch:
@dmitrylyzo I did as well. `topParentId` was always `undefined`.  🤔

thornbill:
Do you happen to be using an extension like clear urls that mangles url parameters? That has been an issue that has came up in the past.

1hitsong:
Like others, I'm unable to recreate this issue. Shuffle in the context menu starts playing a shuffled playlist of my movies.

PR #4654, opened 2023-05-29
Add confirmation for 3rd party repos

### LINKED ISSUE #4653: Plugin repository and plugin installs should inform about security risks
**Describe The Bug**
Currently, our confirmation for plugins is limited to at the actual plugin install, and makes no mention of the potential *security* risks involved, only possible stability. As the recent Emby incident shows and as reported to us by @ScribblerCoder, we should absolutely warn about the possible security implications of both 3rd party repos in general, and of non-official plugins.

**Steps To Reproduce**
N/A

**Expected Behavior**
1. We should warn about 3rd party repo risk when adding the repository.
2. We should warn about the security risks when installing untrusted plugins.
3. We should use a better validation method than the "owner" field in the manifest for confirming the above, since this could be faked by a malicious repo. A check based on the URL (looking for "https://repo.jellyfin.org") would be more suitable.

**Logs**
N/A

**Screenshots**
N/A

N/A

**Additional Context**
N/A

joshuaboniface:
**Changes**

Adds a confirmation similar to the one performed during plugin installation, when adding a 3rd party repository.

The safe domain is hardcoded to be "repo.jellyfin.org" as this is very stable and we have no plans to change it. Individual mirrors don't need to be specified since this is user-input content and they should be using the main URL not the URL of a specific mirror.

The confirmation message makes explicit mention of the possibility of malicious code from 3rd party repositories as well as updates that may bring it in, and suggests only adding 3rd parties from trusted people.

The plugin install confirmation is also modified to use the same conditional and an altered message similar to the above, again to emphasize the potential security risks of 3rd party plugins.

I am open to modifications to the wording of both messages in this PR, though I think this wording captures it fairly well.

Finally, some additional information is added to the Developer Info section of the plugin page; specifically, the name of the repository the plugin is sourced from as well as its URL. How this is obtained is a hack, since these should probably be part of the main information about the plugin and not each specific version, but this is worked around by only showing the information from the first (i.e. newest) version.

**Issues**
Fixes #4653


src/controllers/dashboard/plugins/add/index.html
```diff
@@ -34,7 +34,9 @@ <h2 class="sectionTitle">${HeaderInstall}</h2>
             <div class="readOnlyContent">
                 <div is="emby-collapse" title="${HeaderDeveloperInfo}">
                     <div class="collapseContent">
-                        <p id="developer"></p>
+                        <p>Developer: <span id="developer"></span></p>
+                        <p>Repository Name: <span id="repositoryName"></span></p>
+                        <p>Repository URL: <span id="repositoryUrl"</span></p>
```
nielsvanvelzen, line 39: ^ add to 18n

src/controllers/dashboard/plugins/add/index.js
```diff
@@ -65,6 +65,11 @@ function renderPackage(pkg, installedPlugins, page) {
 
     $('#description', page).text(pkg.description);
     $('#developer', page).text(pkg.owner);
+    // This is a hack; the repository name and URL should be part of the global values
+    // for the plugin, not each individual version. So we just use the top (latest)
+    // version to get this information.
+    $('#repositoryName', page).text(pkg.versions[0].repositoryName);
+    $('#repositoryUrl', page).text(pkg.versions[0].repositoryUrl);
```
nielsvanvelzen, line 72: This can fail if the manifest does not have any versions

src/controllers/dashboard/plugins/add/index.js
```diff
@@ -82,6 +87,7 @@ function alertText(options) {
 
 function performInstallation(page, name, guid, version) {
     const developer = $('#developer', page).html().toLowerCase();
+    const repository_url = $('#repositoryUrl', page).html().toLowerCase();
```
nielsvanvelzen, line 90:
```suggestion
    const repositoryUrl = $('#repositoryUrl', page).html().toLowerCase();
```

src/controllers/dashboard/plugins/repositories/index.js
```diff
@@ -166,14 +167,36 @@ export default function(view) {
         dialog.querySelector('.newPluginForm').addEventListener('submit', e => {
             e.preventDefault();
 
-            repositories.push({
-                Name: dialog.querySelector('#txtRepositoryName').value,
-                Url: dialog.querySelector('#txtRepositoryUrl').value,
-                Enabled: true
-            });
+            const repository_url = dialog.querySelector('#txtRepositoryUrl').value.toLowerCase();
```
nielsvanvelzen, line 170:
```suggestion
            const repositoryUrl = dialog.querySelector('#txtRepositoryUrl').value.toLowerCase();
```

nielsvanvelzen (changes requested):

src/controllers/dashboard/plugins/add/index.js
joshuaboniface: I also can't see a (valid) scenario where there's a plugin entry in a repo with zero versions, and then at worst this would throw and error and be empty, so I think it's fine.

nielsvanvelzen: I don't it's fine because the webui might become unresponsive. I believe the server only returns versions with a compatible targetAbi so it is possible to not receive any versions.

You can check if there is at least one version before setting the properties like so;

```suggestion
if (pkg.versions.length) {
    $('#repositoryName', page).text(pkg.versions[0].repositoryName);
    $('#repositoryUrl', page).text(pkg.versions[0].repositoryUrl);
 }
```

src/controllers/dashboard/plugins/add/index.js
```diff
@@ -65,6 +65,16 @@ function renderPackage(pkg, installedPlugins, page) {
...
     $('#developer', page).text(pkg.owner);
+    // This is a hack; the repository name and URL should be part of the global values
+    // for the plugin, not each individual version. So we just use the top (latest)
+    // version to get this information. If it's missing (no versions), then say so.
+    if (pkg.versions.length) {
+        $('#repositoryName', page).text(pkg.versions[0].repositoryName);
+        $('#repositoryUrl', page).text(pkg.versions[0].repositoryUrl);
+    } else {
+        $('#repositoryName', page).text('Unknown (no versions)');
+        $('#repositoryUrl', page).text('Unknown (no versions)');
```
thornbill, line 76: We should probably allow translations for the fallback text.

thornbill (approved):

PR #4660, opened 2023-06-01
Refactor routes to use layout elements

thornbill:
**Changes**
This is a small refactor to restructure our route components. This is the first step towards adopting the new `RouterProvider` APIs from `react-router`. https://reactrouter.com/en/main/upgrading/v6-data

**Issues**
N/A


dmitrylyzo:
Visually, fine. There is only a warning:
> You rendered descendant <Routes> (or called `useRoutes()`) at "/" (under <Route path="">) but the parent route path has no trailing "*". This means if you navigate deeper, the parent won't match anymore and therefore the child routes will never render.
>
> Please change the parent <Route path=""> to <Route path="/*">.

thornbill:
@dmitrylyzo The warning should be fixed now. It was coming from the `<Routes>` in the tabs and drawer. The sonar failure can be ignored as it's something that was fixed in master after this branch was created.

dmitrylyzo (approved):

PR #4667, opened 2023-06-04
Dashboard activity

thornbill:
**Changes**
Rewrites the activity log in the admin dashboard using mui's data grid component.

~~Currently marked as a draft because the old page view should be removed and it depends on #4660~~

I left the old page view as is for the time being. The entire admin dashboard should be moved to its own application directory since there is no need to support TV devices.

Screenshot 2023-06-02 at 12-02-33 Jellyfin

**Issues**
N/A


src/controllers/dashboard/dashboard.html
```diff
@@ -44,7 +44,7 @@ <h3>${HeaderActiveDevices}</h3>
 
             <div class="dashboardColumn dashboardColumn-2-40 dashboardColumn-3-27">
                 <div class="dashboardSection">
-                    <a is="emby-linkbutton" href="#/serveractivity.html?useractivity=true" class="button-flat sectionTitleTextButton">
+                    <a is="emby-linkbutton" href="#/dashboard/activity?useractivity=true" class="button-flat sectionTitleTextButton">
```
crobibero, line 47: Does this change the non-experimental route to go to the experimental route from the default dashboard?

thornbill: Nope the same route is used for both. There are separate sets of route definitions for the experimental and stable layouts so we can have the same route render the appropriate view depending on which app is active.

crobibero: :magic_wand:

crobibero (approved):

PR #4675, opened 2023-06-06
refactor: suggestionview and genresview

grafixeyehero:


thornbill:
There are a couple things I noticed when I first tested this branch:

1. The first time I visit the genres tab, I get a _bunch_ of errors. Seems like it's probably related to our code that handles image loading on scroll (I'm guessing).
Screenshot 2023-06-22 at 14-30-21 Jellyfin
2. The caching is kind of weird when navigating back to the genres tab since you get a different set of results when the data is refreshed due to the random sorting.

src/apps/experimental/components/library/RecommendationContainer.tsx
```diff
@@ -0,0 +1,61 @@
...
+            break;
+    }
+
+    return (
+        <SectionContainer
+            sectionTitle={escapeHTML(title)}
+            items={recommendation.Items || []}
+            cardOptions={{
+                shape: 'overflowPortrait',
+                showYear: true
```
thornbill, line 55: There seem to be some options missing here compared to the other sections. I specifically noticed the text wasn't centered and the title was not displayed.

src/apps/experimental/routes/movies/SuggestionsView.tsx
```diff
@@ -1,160 +1,51 @@
...
+                            'MessageNoMovieSuggestionsAvailable'
+                        )}
+                    </p>
+                </div>
+            ) : (
+                movieRecommendationsItems.map((recommendation, index) => {
+                    return (
+                        <RecommendationContainer
+                            // eslint-disable-next-line react/no-array-index-key
+                            key={index} // use a unique id return value may have duplicate id
```
thornbill, line 42: Maybe this could use something like this?

```suggestion
                            key={`${recommendation.CategoryId}-${index}`} // use a unique id return value may have duplicate id
```

thornbill (commented):
I made a first pass through most of the code changes...

I really like how this code is looking overall. :purple_heart:

thornbill (approved):

PR #4697, opened 2023-06-22
Fix library filter state not being saved on client

### LINKED ISSUE #40: Library Filter Does Not Save At All
On the webapp (going to my jellyfin instance through a browser), when I use a filter on a library (filter by unplayed, specifically) it doesn't save its state. This issue existed in Emby, but didn't exist in the Android app. I'm going to try building the app again to see if I can reproduce this on the Jellyfin app.

EDIT: Can confirm that this issue exists with the app as well.

ricci2511:
**Changes**
Added saveQuery function calls on several `filterchange` event callbacks to properly save the filter state on the client.

**Issues**
Fixes #40

### DIFF
--- a/src/controllers/movies/movies.js
+++ b/src/controllers/movies/movies.js
@@ -297,6 +297,7 @@ export default function (view, params, tabContent, options) {
             });
             Events.on(filterDialog, 'filterchange', () => {
                 query.StartIndex = 0;
+                userSettings.saveQuerySettings(savedQueryKey, query);
                 itemsContainer.refreshItems();
             });
             filterDialog.show();
--- a/src/controllers/movies/movietrailers.js
+++ b/src/controllers/movies/movietrailers.js
@@ -190,6 +190,7 @@ export default function (view, params, tabContent) {
             });
             Events.on(filterDialog, 'filterchange', function () {
                 getQuery().StartIndex = 0;
+                libraryBrowser.saveQueryValues(getSavedQueryKey(), getQuery());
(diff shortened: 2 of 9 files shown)


ricci2511:
Maybe to avoid less repetition it would be worth it to abstract the `this.showFilterMenu` function callback to its own function, e.g. like this:

```
function showFilterMenuCb(query, mode, filterchangeCb) {
    import('../../components/filterdialog/filterdialog').then(({ default: filterDialogFactory }) => {
        const filterDialog = new filterDialogFactory({
            query,
            mode,
            serverId: ApiClient.serverId()
        });
        Events.on(filterDialog, 'filterchange', filterchangeCb);
        filterDialog.show();
    });
}
```

And then e.g. in `movies.js` we could replace the current `this.showFilterMenu` implementation with this:

```
    this.showFilterMenu = showFilterMenuCb.bind(this, query, 'movies', () => {
        query.StartIndex = 0;
        userSettings.saveQuerySettings(savedQueryKey, query);
        itemsContainer.refreshItems();
    });
```

thornbill:
I have a concern with this... our current UI does not have any indication when a filter is active. I can see this causing an influx of issues where people have forgotten they set a filter and are confused why some of their content is not being displayed.

ricci2511:
Understandable, I guess I could look into implementing such indicator if desired.

thornbill:
This was implemented in https://github.com/jellyfin/jellyfin-web/pull/5717

PR #4703, opened 2023-07-04
Cleanup network config

Shadowghost:
**Changes**
* Update unstable SDK
* Cleanup network configuration after https://github.com/jellyfin/jellyfin/pull/8147

**Comments**
A lot of the configuration values weren't even used by the server in 10.8.


thornbill:
Could we leave the translation labels as is and only do the removals in the `en-us` file? I'm afraid this will make weblate very mad as is. :sweat_smile:

src/controllers/dashboard/networking.js
```diff
@@ -48,27 +48,21 @@ function onSubmit(e) {
...
                 config.EnableHttps = form.querySelector('#chkEnableHttps').checked;
                 config.RequireHttps = form.querySelector('#chkRequireHttps').checked;
                 config.EnableUPnP = enableUpnp;
                 config.BaseUrl = form.querySelector('#txtBaseUrl').value;
                 config.EnableRemoteAccess = form.querySelector('#chkRemoteAccess').checked;
                 config.CertificatePath = form.querySelector('#txtCertificatePath').value || null;
                 config.CertificatePassword = form.querySelector('#txtCertPassword').value || null;
-                config.UPnPCreateHttpPortMap = form.querySelector('#chkCreateHttpPortMap').checked;
                 config.AutoDiscovery = form.querySelector('#chkAutodiscovery').checked;
                 config.AutoDiscoveryTracing = form.querySelector('#chkAutodiscoveryTracing').checked;
```
thornbill, line 63: It looks like this line should be removed also?

thornbill (approved):
This should also fix #3440 since the problematic fields have been removed.

PR #4706, opened 2023-07-07
Use a random backdrop image for video osd

### LINKED ISSUE #4710: Use a random backdrop image for the video osd background
If an item has multiple backdrops, we should pick a random one for the video “poster” instead of always using the first one.

TheMelmacian:
**Changes**
- Moved the code to generate the backdrop image url to the url-utils and extended it so that it randomly returns the url of one of the images from the list instead of only the first one.
- Use random backdrop images in the item details page
- Use random backdrop images for the video poster

**Features**
Fixes #4710

### DIFF
--- a/src/utils/jellyfin-apiclient/BackdropImage.ts
+++ b/src/utils/jellyfin-apiclient/BackdropImage.ts
@@ -24,12 +24,9 @@ export interface ScaleImageOptions {
  * @returns The url of the first or a random backdrop image of the provided item.
  */
 export const getItemBackdropImageUrl = (apiClient: ApiClient, item: BaseItemDto, random = false, options: ScaleImageOptions = {}): string | undefined => {
-    let imgUrl;
-
-    if (item.BackdropImageTags?.length) {
+    if (item.Id && item.BackdropImageTags?.length) {
         const backdropImgIndex = random ? randomInt(0, item.BackdropImageTags.length - 1) : 0;
-        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-        imgUrl = apiClient.getScaledImageUrl(item.Id!, {
+        return apiClient.getScaledImageUrl(item.Id, {
             type: 'Backdrop',
             index: backdropImgIndex,
             tag: item.BackdropImageTags[backdropImgIndex],
@@ -37,19 +34,19 @@ export const getItemBackdropImageUrl = (apiClient: ApiClient, item: BaseItemDto,
         });
     } else if (item.ParentBackdropItemId && item.ParentBackdropImageTags?.length) {
         const backdropImgIndex = random ? randomInt(0, item.ParentBackdropImageTags.length - 1) : 0;
-        imgUrl = apiClient.getScaledImageUrl(item.ParentBackdropItemId, {
+        return apiClient.getScaledImageUrl(item.ParentBackdropItemId, {
             type: 'Backdrop',
             index: backdropImgIndex,
             tag: item.ParentBackdropImageTags[backdropImgIndex],
             ...options
         });
-    } else if (item.ImageTags?.Primary) {
-        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-        imgUrl = apiClient.getScaledImageUrl(item.Id!, {
+    } else if (item.Id && item.ImageTags?.Primary) {
+        return apiClient.getScaledImageUrl(item.Id, {
             type: 'Primary',
             tag: item.ImageTags.Primary,
             ...options
         });
+    } else {
+        return undefined;
     }
-    return imgUrl;
 };
--- a/src/utils/jellyfin-apiclient/BackdropImage.ts
+++ b/src/utils/jellyfin-apiclient/BackdropImage.ts
@@ -1,0 +1,55 @@
+import { BaseItemDto } from '@jellyfin/sdk/lib/generated-client';
+import { randomInt } from '../number';
+import { ApiClient } from 'jellyfin-apiclient';
+
+export interface ScaleImageOptions {
+    maxWidth?: number;
+    width?: number;
+    maxHeight?: number;
+    height?: number;
+    fillWidth?: number;
+    fillHeight?: number;
+    quality?: number;
+}
+
+/**
+ * Returns the url of the first or a random backdrop image of an item.
+ * If the item has no backdrop image, the url of the first or a random backdrop image of the parent item is returned.
+ * Falls back to the primary image (cover) of the item, if neither the item nor it's parent have at least one backdrop image.
+ * Returns undefined if no usable image was found.
+ * @param apiClient The ApiClient to generate the url.
+ * @param item The item for which the backdrop image is requested.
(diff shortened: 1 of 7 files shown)


thornbill:
I haven't tested this yet but I suspect there is an issue with using this for the item details page... if backdrops are enabled then the pages loads two different instances of the image. One that is used in the banner and one that is used for the rest of the page. This is really hacky and we should fix it to only use a single image at some point and just change the opacity of the different areas of the page. 😅 I suspect that because of this you could end up with two different images being used for the different sections of the page.

TheMelmacian:
I looked into the code and tested it again.
The current second backdrop cycles trough all existing backdrops of the item (except in firefox -> comment mentions high cpu usage). So the problem with the two different backdrops already exists without this change.

I see two easy ways to solve this:
- revert the changes for the item details:
  - different pictures are still possible (outside of firefox) but it allows the user to set a "main" (the first) backdrop as header image
- if the backdrop option is activated the header image will be deactivated:
  - if the backdrop-option is not enabled a random header picture is used whenever the details page is (re-)loaded
  - when the backdrop-option is activated only one picture is displayed at the same time

TheMelmacian:
I made the header backdrop static again.

src/utils/url.ts
```diff
@@ -33,3 +37,55 @@ export const getParameterByName = (name: string, url?: string | null | undefined
...
+ * Falls back to the primary image (cover) of the item, if neither the item nor it's parent have at least one backdrop image.
+ * Returns undefined if no usable image was found.
+ * @param apiClient The ApiClient to generate the url.
+ * @param item The item for which the backdrop image is requested.
+ * @param random If set to true and the item has more than one backdrop, a random image is returned.
+ * @param options Optional; allows to scale the backdrop image.
+ * @returns The url of the first or a random backdrop image of the provided item.
+ */
+export const getItemBackdropImageUrl = (apiClient: ApiClient, item: BaseItemDto, random = false, options: ScaleImageOptions = {}): string | undefined => {
+    let imgUrl;
```
thornbill, line 63: There is no need for this variable/assignment, just return the value directly.

src/utils/url.ts
```diff
@@ -33,3 +37,55 @@ export const getParameterByName = (name: string, url?: string | null | undefined
...
+
+    if (item.BackdropImageTags?.length) {
+        const backdropImgIndex = random ? randomInt(0, item.BackdropImageTags.length - 1) : 0;
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
thornbill, line 67: Please add a check for `item.Id` in the if statements like you did for the `ParentBackdropItemId` instead of using the non-null assertion.

thornbill (changes requested):
Sorry for the delay! I just noticed a couple improvements that I think can be made. Also I think this would work better in a new file under `utils/jellyfin-apiclient` since this file is more generic functions for manipulating urls.

Thanks for the PR!

src/utils/jellyfin-apiclient/BackdropImage.ts
```diff
@@ -0,0 +1,52 @@
...
+ * If the item has no backdrop image, the url of the first or a random backdrop image of the parent item is returned.
+ * Falls back to the primary image (cover) of the item, if neither the item nor it's parent have at least one backdrop image.
+ * Returns undefined if no usable image was found.
+ * @param apiClient The ApiClient to generate the url.
+ * @param item The item for which the backdrop image is requested.
+ * @param random If set to true and the item has more than one backdrop, a random image is returned.
+ * @param options Optional; allows to scale the backdrop image.
+ * @returns The url of the first or a random backdrop image of the provided item.
+ */
+export const getItemBackdropImageUrl = (apiClient: ApiClient, item: BaseItemDto, random = false, options: ScaleImageOptions = {}): string | undefined => {
```
thornbill, line 26: Not a big deal, but I think changing the order of these parameters would be slightly better.
```suggestion
export const getItemBackdropImageUrl = (apiClient: ApiClient, item: BaseItemDto, options: ScaleImageOptions = {}, random = false): string | undefined => {
```

src/utils/jellyfin-apiclient/BackdropImage.ts
```diff
@@ -0,0 +1,52 @@
...
+    if (item.Id && item.BackdropImageTags?.length) {
+        const backdropImgIndex = random ? randomInt(0, item.BackdropImageTags.length - 1) : 0;
+        return apiClient.getScaledImageUrl(item.Id, {
+            type: 'Backdrop',
```
thornbill, line 30: There are enum values available for the image types in the SDK.

```ts
import { ImageType } from '@jellyfin/sdk/lib/generated-client/models/image-type';
```

src/utils/jellyfin-apiclient/BackdropImage.ts
```diff
@@ -0,0 +1,52 @@
...
+        });
+    } else if (item.Id && item.ImageTags?.Primary) {
+        return apiClient.getScaledImageUrl(item.Id, {
+            type: 'Primary',
+            tag: item.ImageTags.Primary,
+            ...options
+        });
+    } else {
+        return undefined;
+    }
```
thornbill, line 51:
```suggestion
    }
    return;
```

src/utils/jellyfin-apiclient/BackdropImage.ts
```diff
@@ -0,0 +1,52 @@
+import { BaseItemDto } from '@jellyfin/sdk/lib/generated-client';
+import { randomInt } from '../number';
+import { ApiClient } from 'jellyfin-apiclient';
```
thornbill, line 3: It looks like a couple of these could be type imports and it's generally better for the final bundle size to use the full path for SDK imports (although this is less relevant for a type import).

```suggestion
import { type BaseItemDto } from '@jellyfin/sdk/lib/generated-client/models/base-item-dto';
import { type ApiClient } from 'jellyfin-apiclient';

import { randomInt } from '../number';
```

src/utils/jellyfin-apiclient/BackdropImage.ts
thornbill, line 1: We try to keep file names in `camelCase` unless it is a React component or const/enum value.

thornbill (changes requested):
Looking great! I just had a few minor suggestions. :slightly_smiling_face:

src/utils/jellyfin-apiclient/BackdropImage.ts
TheMelmacian: A simple return statement triggers an eslint error: eslint[sonarjs/no-redundant-jump](https://github.com/SonarSource/eslint-plugin-sonarjs/blob/master/docs/rules/no-redundant-jump.md).

So I can either remove the last return statement completely
```suggestion
    }
```
or explicitly return undefined.
```suggestion
    }
    return undefined;
```

Personally I prefer the latter since it makes clear that the result is undefined on purpose.

thornbill: Either would be fine for me.

thornbill (approved):

PR #4725, opened 2023-07-16
Add JellyBook client to admin dashboard

Kara-Zor-El:
**Changes**
I added my app [JellyBook](https://github.com/Kara-Zor-El/JellyBook) as a app that can be see in the admin dashboard under active devices


src/assets/img/devices/jellybook.svg
```diff
@@ -0,0 +1,226 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
nielsvanvelzen, line 1: You can export an SVG in Inkscape as optimized so it doesn't contain information for editing, additionally a tool like `svgo` can optimize the SVG even more to decrease the file size.

Kara-Zor-El: Hi, I just saved the svg as a optimized svg then optimized it with `svgo`

src/assets/img/devices/jellybook.svg
thornbill, line 1: We typically ask third party app icons to not use the Jellyfin icon (mainly to avoid confusion about where they should go for support).

Kara-Zor-El: That makes sense. Is there any possibility that my app could somehow become a first party app? In the mean time I am currently in the process of trying to design a new icon for JellyBook

thornbill:
We are exploring some ideas to allow third party apps to provide their own icons without requiring them to be added here. I will update here when we have a solid alternative.

thornbill:
#4884 added the ability for clients to provide an image url by including an `IconUrl` value in the `Capabilities` when posting the full capabilities via the session api.

https://typescript-sdk.jellyfin.org/classes/generated_client.SessionApi.html#postFullCapabilities

PR #4731, opened 2023-07-26
Fix major code smells

sttatusx:
Hi,
I fixed some major code smells.

@thornbill

### DIFF
--- a/src/scripts/autocast.js
+++ b/src/scripts/autocast.js
@@ -26,7 +26,7 @@ export function isEnabled() {
     const playerId = localStorage.getItem('autocastPlayerId');
     const currentPlayerInfo = playbackManager.getPlayerInfo();
 
-    return (currentPlayerInfo && playerId && currentPlayerInfo.id === playerId);
+    return currentPlayerInfo?.id === playerId
 }
 
 function onOpen() {
--- a/src/scripts/autocast.js
+++ b/src/scripts/autocast.js
@@ -26,7 +26,7 @@ export function isEnabled() {
     const playerId = localStorage.getItem('autocastPlayerId');
     const currentPlayerInfo = playbackManager.getPlayerInfo();
 
-    return currentPlayerInfo?.id === playerId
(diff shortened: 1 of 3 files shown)


src/scripts/autocast.js
```diff
@@ -26,7 +26,7 @@ export function isEnabled() {
     const playerId = localStorage.getItem('autocastPlayerId');
     const currentPlayerInfo = playbackManager.getPlayerInfo();
 
-    return (currentPlayerInfo && playerId && currentPlayerInfo.id === playerId);
+    return currentPlayerInfo?.id === playerId
```
thornbill, line 29:
```suggestion
    return currentPlayerInfo?.id === playerId;
```

sttatusx:
Hi,
can you check this again?
@thornbill

sttatusx:
Hi, @thornbill

sttatusx:
@thornbill

src/scripts/autocast.js
```diff
@@ -26,7 +26,7 @@ export function isEnabled() {
     const playerId = localStorage.getItem('autocastPlayerId');
     const currentPlayerInfo = playbackManager.getPlayerInfo();
 
-    return (currentPlayerInfo && playerId && currentPlayerInfo.id === playerId);
+    return currentPlayerInfo?.id === playerId;
```
thornbill, line 29: This would actually behave differently in the case that both `currentPlayerInfo` and `playerId` are undefined... I think we need to leave one of the checks to prevent that...
```suggestion
    return playerId && currentPlayerInfo?.id === playerId;
```

thornbill (changes requested):
Sorry for the delay! I took a small break from working on this project. I did notice one (potential) issue with these changes that I noted below. Thanks for the PR!

sttatusx:
> Sorry for the delay! I took a small break from working on this project. I did notice one (potential) issue with these changes that I noted below. Thanks for the PR!

That was a good point.
Thank you

thornbill (approved):

PR #4746, opened 2023-08-08
fix-cast-with-localhost-server

### LINKED ISSUE #4745: Webclient fails to play on chromecast using localhost server url
**Describe The Bug**
Running jellyfin-web using localhost to connect to jellyfin server on the same (Windows) pc fails to play to a chromecast receiver.

**Steps To Reproduce**

1. download and build the jellyfin from jellyfin and jellyfin-web repositories, for instance release 10.8.10
2. start jellyfin server and run jellyfin web with Chrome or Edge using http://localhost:8096
3. setup with the installation wizard and create a media library with some audio/video media
4. login, start a track without connecting to the cast device, and observe that the track plays as expected
5. stop the track, and connect to the chromecast receiver using the cast menu
6. start the track again, and observe that it is not played. Even no playbar is shown.

**Expected Behavior**
This setup might not be used a lot, but it is the default Jellyfin installation on Windows, and anyhow should just work.

**Logs**
n.a.

**Screenshots**
n.a.

 - Platform: Windows 10
 - Browser: Chrome, Edge
 - Jellyfin Version: current github version 10.9.0

**Analysis**

To play media on a chromecast device (receiver), jellyfin-web (the sender application) sends a 'play' command message to the receiver with the url of the media file on the jellyfin server, based on the lan address of the jellyfin server. Jellyfin-web was already aware of this server lan address because the app connected to jellyfin server at startup.

This scenario has one caveat however if jellyfin-web is started on the same Windows pc that runs jellyfin server and hosts this jellyfin-web. By default jellyfin-web will then establish 'http://localhost:8096' as url for the server.

This url works fine for the connection between jellyfin-web and jellyfin server, but obviously can not be used in a media url for the chromecast receiver device, as for the cast receiver 'localhost' is not a valid address for the jellyfin server.

This caveat only occurs when jellyfin-web is run

- from the webserver hosted by jellyfin server, and
- from a browser on the same pc as jellyfin server, and
- using localhost as url for starting the jellyfin-web application.

**Hack to test the solution base**

Basically, the solution is relatively straightforward: although jellyfin-web can communicate fine with jellyfin server using 'localhost', commands from jellyfin-web to the chromecast receiver should not use localhost as url for jellyfin server, but a url with the real lan address. Commands are sent to the cast receiver with media url in the sendmessage method:
  https://github.com/jellyfin/jellyfin-web/blob/eca79feb3efe60ab3d6939e1a5293dce3ab93f01/src/plugins/chromecastPlayer/plugin.js#L342

In the default situation this parameter will contain the url based on localhost. If your Jellyfin server has lan address 192.168.12.34, you can adapt above line such that the message becomes:
```
        message = Object.assign(message, {
            userId: apiClient.getCurrentUserId(),
            deviceId: apiClient.deviceId(),
            accessToken: apiClient.accessToken(),
            serverAddress: 'http://192.168.12.34:8096',
            serverId: apiClient.serverId(),
            serverVersion: apiClient.serverVersion(),
            receiverName: receiverName
        });
```
With this modification, rebuild jellyfin-web, clear the browser cache and restart the jellyfin-web as before, with http://localhost:8096.
Repeat the reproduction steps 4 to 6, and observe in step 6 that now the track does play on the cast device as expected.

**The solution**

The actual solution should use a real lan address for the server in the above message assignment. Within javascript code in a browser, however, it is not possible to find the lan address of the pc. But jellyfin server can find out, so the first step of the solution is an enhancement to jellyfin server to obtain its lan address and pass it to the client. This step is registered as issue https://github.com/jellyfin/jellyfin/issues/10098 for the jellyfin server repository.

The second step is to adapt jellyfin-web to request this lan address from jellyfin-server and use that in the message assignment.

Sky-High:
jellyfin-web needs the ip LAN address of the server to submit play commands to a chromecast receiver. This ip address usually is the address used in jellyfin-web to jellyfin-server communication. But this fails when the latter communication uses localhost, as this is not a usable server address for the chromecast device.

**Changes**
When communcation uses localhost, use a ULA (Unique Lan Address) for command messages to chromecast in src/plugins/chromecastPlayer/plugin.js

**Issues**
Fixes #4745


src/plugins/chromecastPlayer/plugin.js
```diff
@@ -335,16 +333,26 @@ class CastPlayer {
             apiClient = ServerConnections.currentApiClient();
         }
 
+        /* If serverAddress is localhost,this address can not be used for the cast receiver device.
+         * Use the local address (ULA, Unique Local Address) in that case.
+	 */
+        const srvAddress = apiClient.serverAddress();
+        const prefix = 'http' + ':' + '//';
+        const checkLocalhost = srvAddress.startsWith(prefix + 'localhost') || srvAddress.startsWith(prefix + '127.') || srvAddress.startsWith(prefix + '[::1]');
```
thornbill, line 341: Sorry for the delay! This is an interesting issue that you found. I would suggest using the URL API to handle some of the parsing for this.
```suggestion
        const serverAddress = apiClient.serverAddress();
        const hostname = (new URL(serverAddress)).hostname;
        const isLocalhost = hostname === 'localhost' || hostname.startsWith('127.') || hostname === '[::1]';
```

You may need to ignore an eslint warning for compatibility when using the URL constructor, but it's polyfilled by core-js so it should be fine to use.

thornbill (approved):

Issue #4747, opened 2023-08-11
Google Cast Unsupported on Motorola Edge 30


minigt:
**Describe The Bug**
Bug derive from
https://github.com/jellyfin/jellyfin-android/issues/1147

The android app on moto edge can't cast to Chromecast, despite other apps(Netlix, Dysne+, Star+) being able to cast.

**Steps To Reproduce**
On a moto edge phone

1) install jellyfin from google play
2) open app
3)connect to server via https
4)click cast button
5)see "(Google Cast Unsupported)"

**Expected Behavior**
Should be able to cast to a chromecast device

minigt:
I found hat the issue is on line

https://github.com/jellyfin/jellyfin-web/blob/26b83efc95f24ddc10c595d8e472682342764d0b/src/index.jsx#L113

as it should be 

`} else if (!browser.chrome && !browser.edgeChromium && !browser.opera && !browser.edge) {`

ryannathans:
browser.edge?

minigt:
yeap, apparently XD

minigt:
I am trying to push the bugfix branch but it says I don't have p
ermission

ryannathans:
oh no I think the browser detection is seeing the "motorola edge 30" in the user agent and identifying the browser as edge...

ryannathans:
> I am trying to push the bugfix branch but it says I don't have permission

you probably want to fork the repo, branch off master, make the changes, push to your repo and then open a pull request

minigt:
you where right, it was the edge on moto edge

here
https://github.com/jellyfin/jellyfin-web/blob/26b83efc95f24ddc10c595d8e472682342764d0b/src/scripts/browser.js#L187

`
const uaMatch = function (ua) {
    ua = ua.toLowerCase();
    console.log(ua);
    const match = /(edg)[ /]([\w.]+)/.exec(ua)
        || /(edga)[ /]([\w.]+)/.exec(ua)
        || /(edgios)[ /]([\w.]+)/.exec(ua)
        || /(edge)[ /]([\w.]+)/.exec(ua)
        || /(opera)[ /]([\w.]+)/.exec(ua)
        || /(opr)[ /]([\w.]+)/.exec(ua)
        || /(chrome)[ /]([\w.]+)/.exec(ua)
        || /(safari)[ /]([\w.]+)/.exec(ua)
        || /(firefox)[ /]([\w.]+)/.exec(ua)
[...]

PR #4748, opened 2023-08-11
Change regex for browser detection

minigt:
**Changes**
Move chrome regex up in the regex chain so "moto edge X" phone doesn't infer that is using edge as browser

**Issues**
https://github.com/jellyfin/jellyfin-web/issues/4747


thornbill (approved):

src/scripts/browser.js
rIDclickup, line 1: way to go! BG THX!

PR #4749, opened 2023-08-11
Remove the blinking caret in full screen playback.

plafreniere:
Changed the caret color to transparet for the main drawer.

In full screen mode, during video playback, the caret would blink in the top left corner. The caret is not needed in the drawer. Unsure why I am the only one seeing it. 

Running chrome 115.0.5790.171.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -116,6 +116,7 @@
             z-index: 1;
             width: 0.8em;
             padding-left: env(safe-area-inset-left);
+            caret-color: transparent;
         }
 
         [dir="ltr"] .mainDrawerHandle {


1hitsong:
I figured out how to recreate this issue.

You likely have caret browsing enabled. In your browser press F7. It should show a confirmation popup asking if you'd like to enable caret browsing.

Screenshot is me playing a video with the option enabled.

image

1hitsong (approved):
✅Confirmed this hides the caret during video playback when caret navigation is enabled and the user clicks on the far left side of the window.

PR #4751, opened 2023-08-12
Make backdrop screensaver interval configurable

newo-2001:
Make backdrop screensaver interval configurable.

**Changes**
Makes the interval between slides when using the backdrop screensaver configurable in display settings as the current speed seems inappropriately fast for a screensaver.

**Issues**
This functionality was mentioned in [this feature request](https://features.jellyfin.org/posts/1879/ability-to-set-the-interval-for-backdrop-screensavers-additional-screensaver-config).

### DIFF
--- a/src/scripts/settings/userSettings.js
+++ b/src/scripts/settings/userSettings.js
@@ -436,7 +436,7 @@ export class UserSettings {
      */
     backdropScreensaverInterval(val) {
         if (val !== undefined) {
-            return this.set('backdropScreensaverInterval', parseInt(val, 10), false);
+            return this.set('backdropScreensaverInterval', val.toString(), false);
         }
 
         return parseInt(this.get('backdropScreensaverInterval', false), 10) || 5;
@@ -449,7 +449,7 @@ export class UserSettings {
      */
     libraryPageSize(val) {
         if (val !== undefined) {
-            return this.set('libraryPageSize', parseInt(val, 10), false);
+            return this.set('libraryPageSize', val.toString(), false);
         }
 
         const libraryPageSize = parseInt(this.get('libraryPageSize', false), 10);
@@ -468,7 +468,7 @@ export class UserSettings {
      */
     maxDaysForNextUp(val) {
         if (val !== undefined) {
-            return this.set('maxDaysForNextUp', parseInt(val, 10), false);
+            return this.set('maxDaysForNextUp', val.toString(), false);
         }
 
         const maxDaysForNextUp = parseInt(this.get('maxDaysForNextUp', false), 10);
@@ -487,7 +487,7 @@ export class UserSettings {
      */
     enableRewatchingInNextUp(val) {
         if (val !== undefined) {
-            return this.set('enableRewatchingInNextUp', val, false);
+            return this.set('enableRewatchingInNextUp', val.toString(), false);
         }
 
         return toBoolean(this.get('enableRewatchingInNextUp', false), false);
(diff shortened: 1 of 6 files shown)


src/strings/en-us.json
```diff
@@ -566,6 +566,8 @@
     "LabelAutomaticDiscoveryHelp": "Allow applications to automatically detect Jellyfin by using UDP port 7359.",
     "LabelBaseUrl": "Base URL",
     "LabelBaseUrlHelp": "Add a custom subdirectory to the server URL. For example: <code>http://example.com/<b>&lt;baseurl&gt;</b></code>",
+    "LabelBackdropScreensaverInterval": "Backdrop Screensaver Interval",
+    "LabelBackdropScreensaverIntervalHelp": "The time in seconds a between different backdrops when using the backdrop screensaver.",
```
dmitrylyzo, line 570:
```suggestion
     "LabelBackdropScreensaverIntervalHelp": "The time in seconds between different backdrops when using the backdrop screensaver.",
```
If I am wrong, just drop this suggestion.

src/scripts/settings/userSettings.js
```diff
@@ -429,6 +429,19 @@ export class UserSettings {
         return this.get('screensaver', false);
     }
 
+    /**
+     * Get or set the interval between backdrops when using the backdrop screensaver.
+     * @param {number|undefined} val - The interval between backdrops in milliseconds.
+     * @return {number} The interval between backdrops in milliseconds.
```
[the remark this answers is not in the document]
newo-2001: The 1000 multiplier you mentioned is actually because swiper expects the delay milliseconds. Alternatively it would be possible to let the user enter it in milliseconds and remove the multiplier but I think that would degrade the user experience.

src/scripts/settings/userSettings.js
```diff
@@ -429,6 +429,19 @@ export class UserSettings {
...
     }
 
+    /**
+     * Get or set the interval between backdrops when using the backdrop screensaver.
+     * @param {number|undefined} val - The interval between backdrops in seconds.
+     * @return {number} The interval between backdrops in seconds.
+     */
+    backdropScreensaverInterval(val) {
+        if (val !== undefined) {
+            return this.set('backdropScreensaverInterval', parseInt(val, 10), false);
```
dmitrylyzo, line 439: Do we need `parseInt`?
Since `backdropScreensaverInterval` expects a `number` (and make it consistent with the others):
```suggestion
            return this.set('backdropScreensaverInterval', val.toString(), false);
```

newo-2001: I agree that the type annotation in this case is misleading at best. I think it best to leave this as is for now to be consistent with the other settings as they would need changing too. I would prefer consistency over correctness.

dmitrylyzo: It is already inconsistent. :man_shrugging: In fact, 50/50.

These two are old.
https://github.com/jellyfin/jellyfin-web/blob/ea6c8f01cfec600be45544d8ac939a8de11f85b1/src/scripts/settings/userSettings.js#L361
https://github.com/jellyfin/jellyfin-web/blob/ea6c8f01cfec600be45544d8ac939a8de11f85b1/src/scripts/settings/userSettings.js#L374

These are newer.
:point_down: _This is the first of such kind._
https://github.com/jellyfin/jellyfin-web/blob/ea6c8f01cfec600be45544d8ac939a8de11f85b1/src/scripts/settings/userSettings.js#L452
https://github.com/jellyfin/jellyfin-web/blob/ea6c8f01cfec600be45544d8ac939a8de11f85b1/src/scripts/settings/userSettings.js#L471

newo-2001: Shall I just go ahead and make all of them consistent with the older style?

dmitrylyzo (approved):

PR #4754, opened 2023-08-18
Change Subtitle Sync slider to go from -300 to 300

### LINKED ISSUE #4753: LG WebOS Subtitle offset jumps by 0.6s
**Describe The Bug**
When on LG WebOS and changing subtitle offset, it jumps in steps of 0.6s making it hard to accurately select a good offset.  

**Steps To Reproduce**

1. Start media with subtitles
2. Click on the cogwheel (Settings)
3. Click on to 'Subtitle Offset'
4. Change offset
5. See offset jump by 0.6s

**Expected Behavior**

It's expected to jump in increments of 0.1s.

 - Platform: LG WebOS
 - Jellyfin Version: 10.8.10

mkanilsson:
**Changes**

This patch changes the subtitle sync from using a precentage to an 'slider value' which ranges from -300 to 300 (ms offset). The reasons for this is that LG WebOS doesn't jump in 0.1 slider increments as specified in the `subtitlesync.template.html` file but instead jumps in 1.0 increments. That results in Subtitle Sync jumping 0.6s per increment. Using a value from -300 to 300 makes LG WebOS jump 0.1s instead.

There is no noticeable effect on any other platform.

**Issues**
Fixes #4753

### DIFF
--- a/src/components/subtitlesync/subtitlesync.template.html
+++ b/src/components/subtitlesync/subtitlesync.template.html
@@ -3,7 +3,7 @@
         <button type="button" is="paper-icon-button-light" class="subtitleSync-closeButton"><span class="material-icons close" aria-hidden="true"></span></button>
         <div class="subtitleSyncTextField" contenteditable="true" spellcheck="false">0s</div>
         <div class="sliderContainer subtitleSyncSliderContainer">
-            <input is="emby-slider" type="range" step=".1" min="0" max="100" value="50" class="subtitleSyncSlider" data-slider-keep-progress="true" />
+            <input is="emby-slider" type="range" step="1" min="-300" max="300" value="0" class="subtitleSyncSlider" data-slider-keep-progress="true" />
         </div>
     </div>
 </div>
--- a/src/components/subtitlesync/subtitlesync.js
+++ b/src/components/subtitlesync/subtitlesync.js
@@ -54,7 +54,7 @@ function init(instance) {
                 playbackManager.setSubtitleOffset(inputOffset, player);
                 // synchronize with slider value
                 subtitleSyncSlider.updateOffset(
-                    getPercentageFromOffset(inputOffset));
+                    getSliderValueFromOffset(inputOffset));
             } else {
                 this.textContent = (playbackManager.getPlayerSubtitleOffset(player) || 0) + 's';
             }
@@ -79,17 +79,17 @@ function init(instance) {
         }
     };
 
-    subtitleSyncSlider.updateOffset = function (percent) {
-        // default value is 0s = 50%
-        this.value = percent === undefined ? 50 : percent;
+    subtitleSyncSlider.updateOffset = function (sliderValue) {
+        // default value is 0s = 0ms
+        this.value = sliderValue === undefined ? 0 : sliderValue;
     };
 
     subtitleSyncSlider.addEventListener('change', function () {
         // set new offset
-        playbackManager.setSubtitleOffset(getOffsetFromPercentage(this.value), player);
+        playbackManager.setSubtitleOffset(getOffsetFromSliderValue(this.value), player);
         // synchronize with textField value
         subtitleSyncTextField.updateOffset(
-            getOffsetFromPercentage(this.value));
+            getOffsetFromSliderValue(this.value));
     });
 
     subtitleSyncSlider.getBubbleHtml = function (value) {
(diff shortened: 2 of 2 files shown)


src/components/subtitlesync/subtitlesync.template.html
```diff
@@ -3,7 +3,7 @@
         <button type="button" is="paper-icon-button-light" class="subtitleSync-closeButton"><span class="material-icons close" aria-hidden="true"></span></button>
         <div class="subtitleSyncTextField" contenteditable="true" spellcheck="false">0s</div>
         <div class="sliderContainer subtitleSyncSliderContainer">
-            <input is="emby-slider" type="range" step=".1" min="0" max="100" value="50" class="subtitleSyncSlider" data-slider-keep-progress="true" />
+            <input is="emby-slider" type="range" step=".1" min="0" max="600" value="300" class="subtitleSyncSlider" data-slider-keep-progress="true" />
```
dmitrylyzo, line 6: Have you tried setting `min="-300" max="300"`? Iirc, `emby-slider` supports negative values. In that case, there will be less math.

mkanilsson: I'll try it out, if it works, I'll push it.

mkanilsson: It worked, I've pushed it.

src/components/subtitlesync/subtitlesync.template.html
```diff
@@ -3,7 +3,7 @@
         <button type="button" is="paper-icon-button-light" class="subtitleSync-closeButton"><span class="material-icons close" aria-hidden="true"></span></button>
         <div class="subtitleSyncTextField" contenteditable="true" spellcheck="false">0s</div>
         <div class="sliderContainer subtitleSyncSliderContainer">
-            <input is="emby-slider" type="range" step=".1" min="0" max="100" value="50" class="subtitleSyncSlider" data-slider-keep-progress="true" />
+            <input is="emby-slider" type="range" step=".1" min="-300" max="300" value="0" class="subtitleSyncSlider" data-slider-keep-progress="true" />
```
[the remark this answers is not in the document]
mkanilsson: I feel like `emby-slider` keyboard-step should step by what is specified in the `step` attribute if `keyboardStepDown` and/or `keyboardStepUp` hasn't been set, but that's for another PR.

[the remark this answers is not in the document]
dmitrylyzo: > I feel like `emby-slider` keyboard-step should step by what is specified in the `step` attribute if `keyboardStepDown` and/or `keyboardStepUp` hasn't been set, but that's for another PR.

:+1: 
I am currently in the process of implementing this.

dmitrylyzo (approved):
LGTM + it is good to be moved to `release-10.8.z` branch.

We still can improve it by making slider value to subtitle offset ratio 1:1 (`step=".1" min="-30" max="30" value="0"`).
Possible options:
- Wait https://github.com/jellyfin/jellyfin-web/pull/4758
- Or add a call  `subtitleSyncSlider.setKeyboardSteps(0.1);` after [this line](https://github.com/jellyfin/jellyfin-web/blob/aa02a52c566c123aedbcb30f33fd932d64b8cf31/src/components/subtitlesync/subtitlesync.js#L24).

Given the above, we can rely on the slider: [this](https://github.com/jellyfin/jellyfin-web/blob/28ab879f3c99f8c7831c05309199ef2740c66d4a/src/components/subtitlesync/subtitlesync.js#L118-L127) becomes redundant and  [this](https://github.com/jellyfin/jellyfin-web/blob/28ab879f3c99f8c7831c05309199ef2740c66d4a/src/components/subtitlesync/subtitlesync.js#L49-L57) can be simplified.

https://github.com/jellyfin/jellyfin-web/pull/4760 will allow to simplify `getBubbleHtml` and drop [`getOffsetFromPercentage`](https://github.com/jellyfin/jellyfin-web/blob/28ab879f3c99f8c7831c05309199ef2740c66d4a/src/components/subtitlesync/subtitlesync.js#L110-L116).

thornbill:
@dmitrylyzo is this PR good to merge or should it wait on the other changes you have mentioned? Thanks!

dmitrylyzo:
This PR definitely fixes stepping and makes it even.

I didn't want to burden the contributor too much, but it would be better to make the slider in 1:1 ratio - this won't change UI/UX (relative to PR), but makes the code cleaner.

FYI, #4758 fixes stepping (in the slider itself), but it's uneven due to rounding.

As for `getBubbleHtml` thing, I will do it in my PR later - this is just enhancement anyway.

PR #4756, opened 2023-08-22
Disable WebVTT subtitles on Windows

### LINKED ISSUE #4742: Styles not applied for WebVTT subtitles on Windows using Chrome/Edge
**Describe The Bug**

It is known that the WebVTT specification subtitles are style-adjusted through the css ::cue selector, but [Chromium and Edge developers applied](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/WebVTTCaptionStyling/explainer.md) the OS's setting value first if there is a closed caption setting value of the OS.

So, if you modified the Closed Caption settings at least once in Windows 10 or 11, the subtitle-related settings in the Jellyfin web client will not work at all in Chrome and Edge.

After touching the closed caption setting even once, the related key value(HKCU\Software\Microsoft\Windows\CurrentVersion\ClosedCaptioning) is created in the registry and cannot be rolled back in the usual way.

The only workaround for this problem is to delete that registry key.

**Steps To Reproduce**

Open Windows Settings and change anything at Accessibility -> Captions.
Open the Jellyfin web client in Chrome/Edge and play the video with subtitles in an uncomplicated format (.srt, .smi, etc.).

**Expected Behavior**

All subtitle settings in Jellyfin web client don't work at all.

**Logs**

**Screenshots**

Screenshot 1
Screenshot 2
Screenshot 3

 - Platform: [e.g. Linux, Windows, iPhone, Tizen] Windows 10 / 11
 - Browser: [e.g. Firefox, Chrome, Safari] Edge / Chrome
 - Jellyfin Version: [e.g. 10.6.0] 10.8.10

**Additional Context**

It was confirmed that the size control of the subtitle works when the font-size is directly adjusted in the block without using the ::cue selector in Windows. (See second screenshot)

When tested with the latest web client code snapshot and Firefox, subtitles were displayed through WebVTT. Nevertheless, subtitle settings were applied and worked well in Firefox regardless of OS settings.

When opening the 10.8.10 version of the web client in Firefox, subtitles were implemented via div blocks.

It would be nice, in my opinion, to extend the fallback implementation used in Firefox to all desktop browsers on Windows, or to have the option to choose between WebVTT and html implementations. Disney+'s subtitles implementation relies on Windows' captioning capabilities, so it's hard not to use this feature.

I also tested it on Mac, but the OS setting was not applied in Edge and Chrome, and it was applied only in Safari, but it was applied simultaneously with the properties of .htmlvideoplayer::cue.

HiSkyZen:
**Changes**

Returns the value of requiresCustomSubtitlesElement of Edge and Chrome on Windows as true.

**Issues**

Fixes #4742


src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1350,6 +1350,20 @@ export class HtmlVideoPlayer {
...
+                return true;
+            }
+
+            if (browser.edgeUwp) {
+                return true;
+            }
+
+            if (browser.edg) {
+                return true;
+            }
```
thornbill, line 1364:
```suggestion
            if (browser.chrome || browser.edgeUwp || browser.edg) {
                return true;
            }
```

thornbill:
Similar to https://github.com/jellyfin/jellyfin-web/pull/5737#issuecomment-2286999425 I wonder if we should have an option for native vs custom styling.

PR #4758, opened 2023-08-23
Fix slider step

### LINKED ISSUE #4753: LG WebOS Subtitle offset jumps by 0.6s

dmitrylyzo:
**Changes**
Use the value of the `step` attribute if no keyboard steps are specified.

**Issues**
Fixes #4753

_Requires manual backporting (indentation change in the master)._

### DIFF
--- a/src/elements/emby-slider/emby-slider.js
+++ b/src/elements/emby-slider/emby-slider.js
@@ -21,6 +21,27 @@ import '../emby-input/emby-input';
     }
 
     /**
+     * Returns normalized slider step.
+     *
+     * @param {HTMLInputElement} range slider itself
+     * @param {number|undefined} step step
+     * @returns {number} normalized slider step.
+     */
+    function normalizeSliderStep(range, step) {
+        if (step > 0) {
+            return step;
+        }
+
+        step = parseFloat(range.step);
+
+        if (step > 0) {
+            return step;
+        }
+
+        return 1;
+    }
+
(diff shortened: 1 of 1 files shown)


src/elements/emby-slider/emby-slider.js
```diff
@@ -20,6 +20,27 @@ import '../emby-input/emby-input';
...
+            return step;
+        }
+
+        step = parseFloat(range.step);
+
+        if (step > 0) {
+            return step;
+        }
+
+        return 1;
```
thornbill, line 41: What do you think about this?

```suggestion
        return Math.max(step, 1);
```

dmitrylyzo: `parseFloat(range.step)` returns `NaN` for `'auto'` or an invalid value. `Math.max` returns `NaN` in this case.

thornbill (approved):

PR #4761, opened 2023-08-28
HLS direct play

### LINKED ISSUE #4479: jellyfin is doing unnecessary HLS transcoding ?
m3u8 is detected as `hls` container on server, but jellyfin-web didn't send `hls` in its **DirectPlayProfiles**;
so direct play check on server fails with **ContainerNotSupported**, and it fallback to _transcoding_.

since we already use _hls.js_, m3u8 can be directly played despite [HTMLMediaElement.canPlayType ](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/canPlayType) says no.

i am new to jellyfin, this transcoding seems unnecessary to me. currently i just hard coded the following lines in _browserDeviceProfile.js_:

` profile.DirectPlayProfiles.push({
            Container: 'hls',
            Type: 'Video',
            VideoCodec: 'h264,hevc',
            AudioCodec: 'aac,ac3,eac3,flac'
        });
`

**how to reproduce:**

1. create a _.strm_ file with a m3u8 video URL as only content, there are many usable m3u8 link on[ hls.js demo site](https://hlsjs.video-dev.org/demo/) 
2. scan Libraries and play the added video

dmitrylyzo:
**Changes**
- Add HLS direct play profile.
- Fix HLS detection (don't rely on the file extension).

**Issues**
Fixes #4479

_It works on 10.8 too._

### DIFF
--- a/src/plugins/htmlVideoPlayer/plugin.js
+++ b/src/plugins/htmlVideoPlayer/plugin.js
@@ -35,6 +35,7 @@ import { setBackdropTransparency, TRANSPARENCY_LEVEL } from '../../components/ba
 import { PluginType } from '../../types/plugin.ts';
 import Events from '../../utils/events.ts';
 import { includesAny } from '../../utils/container.ts';
+import { isHls } from '../../utils/mediaSource.ts';
 import debounce from 'lodash-es/debounce';
 
 /**
@@ -74,7 +75,7 @@ function enableNativeTrackSupport(mediaSource, track) {
         return true;
     }
 
-    if (browser.firefox && (mediaSource?.TranscodingSubProtocol || mediaSource?.Container) === 'hls') {
+    if (browser.firefox && isHls(mediaSource)) {
         return false;
     }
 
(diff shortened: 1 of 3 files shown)


src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -69,12 +69,12 @@ function tryRemoveElement(elem) {
...
 }
 
-function enableNativeTrackSupport(currentSrc, track) {
+function enableNativeTrackSupport(mediaSource, track) {
     if (track?.DeliveryMethod === 'Embed') {
         return true;
     }
 
-    if (browser.firefox && (currentSrc || '').toLowerCase().includes('.m3u8')) {
+    if (browser.firefox && (mediaSource?.TranscodingSubProtocol || mediaSource?.Container) === 'hls') {
```
thornbill, line 77: What if we made this hls test a function? It could reduce a _little_ duplication.

src/utils/mediaSource.ts
```diff
@@ -0,0 +1,10 @@
+import { MediaSourceInfo } from '@jellyfin/sdk/lib/generated-client';
+
+/**
+ * Checks if the media source is an HLS stream.
+ * @param mediaSource The media source.
+ * @returns _true_ if the media source is an HLS stream, _false_ otherwise.
+ */
+export function isHls(mediaSource: MediaSourceInfo|null|undefined): boolean {
+    return (mediaSource?.TranscodingSubProtocol || mediaSource?.Container) === 'hls';
```
dmitrylyzo, line 9: Hmm, https://github.com/jellyfin/jellyfin/pull/10153 will change the type of `TranscodingSubProtocol`.

thornbill: I think it will be fine as long as `hls` matches the enum value.

thornbill (approved):

Issue #558, opened 2019-11-01
Inconsistent size of subtitles


purcell:
**Describe the bug**

When choosing "extra large" subtitles, the subtitles are not actually large when viewed in full screen, and do not scale with the overall viewport.

**To Reproduce**

- Select "extra large" subtitles
- Play a video with soft subs in a smallish browser window. The subtitles will seem relatively large.
- Make the window full-screen to view the video from a distance. The subtitles will remain the same absolute size, and not be reasily eadable.

**Expected behavior**

I would expect the large subtitles to be _relatively_ large, so that when a player window is made bigger for convenient viewing at a distance, the subtitles are correspondingly larger.

**Context/notes**

The corresponding HTML is the following:
```html
<div class="videoSubtitles">
  <div class="videoSubtitlesInner" style="font-size: 2.2em; text-shadow: rgb(0, 0, 0) 0px 0px 7px; background-color: transparent; color: rgb(255, 255, 255); font-family: inherit; font-variant: none;">
    <font size="36">A VERY SECRET SERVICE</font>
  </div>
</div>
```
There are 2 impediments here to correct scaling. Firstly, the `<font>` element wrapper appears to defeat the font style of the wrapper element, which is clearly designed to cascade onto the sub text itself.

Secondly, the `videoSubtitlesInner` element specifies a font size that is effectively fixed in height, since the default font size for the page is fixed, and not relative to the viewport height.

The suggested fix would be as follows:
- Remove the `<font>` wrapper element. I couldn't see where that's added, sorry.
- Set the font size of the `videoSubtitles` to a fraction of the viewport height so that it is a useful relative basis height for scaling the subs it contains.

I did the above in the browser dev tools and got a good result by effectively adding the style rule

```css
.videoSubtitles { font-size: 4vh; }
```

(As a more general suggestion, since Jellyfin offers an "in-browser player app" experience it might make sense to set the font size for the entire UI in `vh` units. That's beyond the scope of this bug report, though!)

**System:**

 - OS: macOS
 - Browser: Firefox
 - Jellyfin Version: 10.4

Hope that helps, and many thanks for Jellyfin!

JustAMan:
shoo bot
also cc @redSpoutnik as one of subs contributors

redSpoutnik:
The font-sizes of subtitles are defined here:

[...]

PR #4767, opened 2023-08-31
feat: scale subtitles with larger displays

demo-exe:
Adresses https://github.com/jellyfin/jellyfin-web/issues/558
From what I found, using "vh" would not be preferable because:
- it gets its "100%" not from browser window height but from video player height that can change on fullscreen
- scaling browser (and video player) can cause subtitles to become unreadably small

This solution is not perfect because:
- This should be handled with HiDPI displays https://github.com/jellyfin/jellyfin-media-player/issues/119
- Moving browser window between screens with different resolution does not update subtitles (reload is required)

### DIFF
--- a/src/components/subtitlesettings/subtitleappearancehelper.js
+++ b/src/components/subtitlesettings/subtitleappearancehelper.js
@@ -6,25 +6,34 @@
 function getTextStyles(settings, preview) {
     const list = [];
 
+    let scale = 1;
+    if (window.screen.height > 1080) {
+        scale = (window.screen.height / 1080);
+    }
+
+    const scaledFontSizeEm = (val) => {
+        list.push({ name: 'font-size', value: (scale * val).toString() + 'em' });
+    };
+
     switch (settings.textSize || '') {
         case 'smaller':
-            list.push({ name: 'font-size', value: '.8em' });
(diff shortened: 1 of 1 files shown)


src/components/subtitlesettings/subtitleappearancehelper.js
```diff
@@ -6,25 +6,34 @@
 function getTextStyles(settings, preview) {
     const list = [];
 
+    let scale = 1;
```
thornbill, line 9: Could we just use `devicePixelRatio` here?

https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio

ArgentumCation: IME that tends to be hardcoded to a platform specific value that isn't the most helpful

purcell:
> it might be worth looking into using `vmin` or `vmax` instead of `vh` since that could handle portrait mode better

That makes sense. Perhaps a fresh PR is called for here? — I don't know who would make the casting decision, but at least a fresh PR with `vh` or `vmin`/`vmax` would make the choice concrete. Those solutions would make the scaling work like VLC, with no obvious downsides, which I'd think should be uncontroversial.

PR #4769, opened 2023-09-05
Prefer unplayed items when shuffling a folder

DavidMikeSimon:
**Changes**
When shuffling a folder, sort by `"IsPlayed,Random"` instead of just `"Random"`, causing unplayed items to always be ahead of played items.

The idea here is that when a user hits the "Shuffle" button in a video context (e.g. a show, a library, a collection), they are probably not interested in seeing items that they have recently already played via shuffle. Tying the priority to the "Played" flag gives the user some control over this behavior, e.g. if they want to "reset" their shuffle exclusion, they can mark a folder as unplayed.

Tested this behavior against my home Jellyfin server (v10.9.0), verified that it shoves items with `Played: true` to the bottom of the response (or after the end of the response, if there's more than the default limit of 300 items).

**Issues**
[Feature Request 717](https://features.jellyfin.org/posts/717/shuffle-weight-recently-played-media-should-be-less-likely-to-be-next)

**Errata**
I originally tried to do this on the backend in https://github.com/jellyfin/jellyfin/pull/10011, but it was pointed out that prioritizing unplayed items was not _always_ the best move (e.g. when playing music it's not helpful). I then realized that it would be easier to just request this behavior on the frontend, when appropriate, using the existing multi-column sort feature.

### DIFF
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -1884,7 +1884,7 @@ class PlaybackManager {
             } else if (firstItem.IsFolder) {
                 let sortBy = null;
                 if (options.shuffle) {
-                    sortBy = 'Random';
+                    sortBy = 'IsPlayed,Random';
                 } else if (firstItem.Type !== 'BoxSet') {
                     sortBy = 'SortName';
                 }


thornbill:
I'm not sure if this makes sense... :thinking:

DavidMikeSimon:
@thornbill I'll explain a little more about the problem I'm trying to solve here. Maybe we can come up with a different approach.

My family likes to watch old TV shows in random order, so that we get a surprise each time about what comes up next. Shuffle mode is great for this. However, once we've seen an episode randomly this way, we don't want to see it again right for a while. Instead we'd rather see an episode that we haven't seen in recent memory. As the [original feature request](https://features.jellyfin.org/posts/717/shuffle-weight-recently-played-media-should-be-less-likely-to-be-next) points out, with purely random picks we are very unlikely to get this behavior.

One solution would be to construct a random playlist in advance. But, this has some issues:
* It kills the surprise, since when we open the playlist we can immediately see the upcoming list of items.
* It is takes a bit of scrolling around to jump back into the playlist where we left off.
* We sometimes have more or less specific moods. For example, sometimes our request is "a random episode from season 3 of a particular show", and other times our request is "a random episode from any show in the Science Fiction collection". If those are two separate but partially overlapping playlists, then we aren't able to avoid repeats as effectively.

Prioritizing unplayed items over played items in random shuffle is a nice simple way to get this behavior.

thornbill:
Perhaps a better option would be to fix the shuffle / play all buttons not respecting the filters set on the page.... then you could filter to only show unplayed and then shuffle to achieve basically the same functionality.

DavidMikeSimon:
@thornbill That sounds pretty good to me! But unless I'm missing something, there doesn't seem to be a view that lists all the leaf items in a nested folder in a filterable way. So the approach you suggest wouldn't work for e.g. shuffling the episodes in a multi-season show, or the episodes in a custom collection of shows or show seasons.

What if there were a user option to control whether shuffle prioritized unplayed items?

PR #4775, opened 2023-09-13
Remove unused private class members

thornbill:
**Changes**
Remove unused private class members. Fixes these related issues in sonarcloud: https://sonarcloud.io/project/issues?impactSeverities=MEDIUM&resolved=false&rules=javascript%3AS1068&types=CODE_SMELL&id=jellyfin_jellyfin-web

**Issues**
N/A

### DIFF
--- a/src/plugins/htmlVideoPlayer/plugin.js
+++ b/src/plugins/htmlVideoPlayer/plugin.js
@@ -168,133 +168,140 @@ const SECONDARY_TEXT_TRACK_INDEX = 1;
 
 export class HtmlVideoPlayer {
     /**
-         * @type {string}
-         */
+     * @type {string}
+     */
     name;
     /**
-         * @type {string}
-         */
+     * @type {string}
+     */
     type = PluginType.MediaPlayer;
     /**
-         * @type {string}
-         */
+     * @type {string}
+     */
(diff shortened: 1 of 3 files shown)


grafixeyehero (approved):

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -277,10 +277,6 @@ export class HtmlVideoPlayer {
          * @type {number | null | undefined}
          */
     #currentTime;
-    /**
-         * @type {any | undefined}
-         */
-    #flvPlayer;
```
grafixeyehero, line 283: https://github.com/jellyfin/jellyfin-web/blob/c506d0b753df12eacbd227dc14232c0a6c642c0c/src/plugins/htmlVideoPlayer/plugin.js#L411

thornbill: Yeah it turns out it this is actually referenced [here](https://github.com/jellyfin/jellyfin-web/blob/master/src/components/htmlMediaHelper.js#L237) but that must have been missed when it was converted to being a private member.

grafixeyehero (approved):

PR #4777, opened 2023-09-15
Migrate quick connect page to react

thornbill:
**Changes**
* Simplifies adding redirects
* Updates `InputElement` to support specifying classes for the container div, call a callback function on value changes, provide an initial value, and prevent re-renders that break the webcomponent
* Migrates the quick connect page to React + TS SDK
* Allows specifying the code as a url param to allow for clients to link to the form prefilled

**Issues**
N/A


nielsvanvelzen:
Is this just a 1 to 1 conversion or are there changes to the UI/UX? I'm asking because I've never been satisfied with the UX once a (valid) quick connect code was entered. The toast showing up in the bottom-left corner is not notable at all.

thornbill:
Well it was a 1:1 conversion except for being able to prefill from a url parameter, but updating the ui sounds like a good idea too lol

thornbill:
How about this? It's still not really great, but there really isn't much in the way of existing components for the stable app. :sweat_smile: 

Screenshot 2023-09-16 at 01-50-45 Jellyfin

Screenshot 2023-09-16 at 02-48-33 Jellyfin

nielsvanvelzen:
Way better!

src/apps/stable/routes/_redirects.ts
```diff
@@ -0,0 +1,6 @@
+import type { Redirect } from 'components/router/Redirect';
+
+export const REDIRECTS: Redirect[] = [
+    { from: 'mypreferencesquickconnect.html', to: '/quickconnect' },
```
grafixeyehero, line 4: `{ from: 'quickconnect', to: '/quickconnect' }`

grafixeyehero (changes requested):

src/apps/stable/routes/_redirects.ts
thornbill: The "from" value is the old url of the page. This way if someone tries to go to the old url directly they get bounced to the new url instead of landing on a broken page.

grafixeyehero (approved):

PR #4781, opened 2023-09-16
Add conditional build USE_SYSTEM_FONTS

dmitrylyzo:
To be able to discard unused fonts for embedding in a Tizen app.
```sh
USE_SYSTEM_FONTS=1 npm run build:production
```

I am currently filtering them using [an unreliable method](https://github.com/jellyfin/jellyfin-tizen/blob/3ce22e1a33ff8d3b1da1f2ae87a5fb521bd24467/gulpfile.js#L39-L49). And some of them (default font for JASSUB, Material Design Icons font) become missing.

**Changes**
Add conditional build `USE_SYSTEM_FONTS`.

**Issues**
There is no normal way to discard unused fonts.

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -92,9 +92,13 @@ function onGlobalizeInit() {
     if (browser.tv && !browser.android) {
         console.debug('using system fonts with explicit sizes');
         import('./styles/fonts.sized.scss');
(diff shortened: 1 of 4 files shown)


thornbill (changes requested):
Looks good it just needs the new variable added as a global to the eslint config!

thornbill (approved):

PR #4793, opened 2023-09-20
Fix error on unmount in syncplay menu

thornbill:
**Changes**
Fixes a React error when `setGroups` is called after the component is unmounted.

Using an `AbortController` might be technically better, but it looks like we would have to polyfill.

**Issues**
React error printed to console


grafixeyehero (approved):

grafixeyehero:
@thornbill isn't it better if we use tanstack query?

thornbill:
Ah that's a great point! I didn't even think that it would probably eliminate this issue also...

grafixeyehero:
Here is https://tanstack.com/query/latest/docs/react/guides/query-cancellation

joshuaboniface (approved):

PR #4802, opened 2023-09-25
Update to node 20

### LINKED ISSUE #4808: Upgrade to latest npm version (and latest node lts version)
- [x] Node 16: https://github.com/jellyfin/jellyfin-web/pull/3972
- [ ] ~Node 18~
- [ ] Node 20: https://github.com/jellyfin/jellyfin-web/pull/4802

Renari:
**Changes**
Updates builds and docker images to use node 18. Also changes nodesource entries to use what is recommended by their [current docs](https://github.com/nodesource/distributions) rather than the depreciated scripts.

**Issues**
Closes #4808

### DIFF
--- a/.ci/azure-pipelines-build.yml
+++ b/.ci/azure-pipelines-build.yml
@@ -16,7 +16,7 @@ jobs:
   - task: NodeTool@0
     displayName: 'Install Node'
     inputs:
-      versionSpec: '16.x'
+      versionSpec: '20.x'
 
   - task: Cache@2
     displayName: 'Cache node_modules'
--- a/deployment/Dockerfile.debian
+++ b/deployment/Dockerfile.debian
@@ -12,10 +12,12 @@ ENV IS_DOCKER=YES
 
 # Prepare Debian build environment
 RUN apt-get update \
-  && apt-get install -y debhelper mmv git curl \
-  && curl -fsSL https://deb.nodesource.com/setup_16.x | bash - \
+  && apt-get install -y debhelper mmv git curl gnupg ca-certificates \
+  && mkdir -p /etc/apt/keyrings \
+  && curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg \
+  && echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_20.x nodistro main" | tee /etc/apt/sources.list.d/nodesource.list \
+  && apt-get update \
   && apt-get install -y nodejs
-
 
 # Link to build script
 RUN ln -sf ${SOURCE_DIR}/deployment/build.debian /build.sh
--- a/fedora/Makefile
+++ b/fedora/Makefile
@@ -7,7 +7,7 @@ RELEASE := $(shell set -x; sed -ne '/^Release:/s/.*  *\(.*\)%{.*}.*/\1/p' $(DIR)
 SRPM    := jellyfin-web-$(subst -,~,$(VERSION))-$(RELEASE)$(shell rpm --eval %dist).src.rpm
 TARBALL :=$(NAME)-$(subst -,~,$(VERSION)).tar.gz
 
-epel-7-x86_64_repos := https://rpm.nodesource.com/pub_16.x/el/\$$releasever/\$$basearch/
+epel-7-x86_64_repos := https://rpm.nodesource.com/pub_20.x/nodistro/\$$basearch/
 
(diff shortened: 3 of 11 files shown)


.ci/azure-pipelines-build.yml
```diff
@@ -16,7 +16,7 @@ jobs:
   - task: NodeTool@0
     displayName: 'Install Node'
     inputs:
-      versionSpec: '16.x'
+      versionSpec: '18.18.x'
```
Renari, line 19: It may be preferred to leave this as `18.x` and change `package.json` to target `18.17`.

thornbill: :+1: I would prefer sticking with `18.x`

thornbill (changes requested):
It also looks like we need to update the following:

- [Fedora Makefile](https://github.com/jellyfin/jellyfin-web/blob/master/fedora/Makefile#L10)
- [Fedora Spec](https://github.com/jellyfin/jellyfin-web/blob/master/fedora/jellyfin-web.spec#L17-L20)
- [TypeScript Action](https://github.com/jellyfin/jellyfin-web/blob/master/.github/workflows/tsc.yml#L21)

There are going to be a couple places in the server repo also, but those can be updated after this is merged.

nielsvanvelzen:
Node 20 will be LTS in a few weeks, wouldn't it make more sense to upgrade to that version instead?

thornbill:
Good point @nielsvanvelzen, if there aren't any issues found going straight from 16 -> 20 let's go that route.

fedora/Makefile
```diff
@@ -7,7 +7,7 @@ RELEASE := $(shell set -x; sed -ne '/^Release:/s/.*  *\(.*\)%{.*}.*/\1/p' $(DIR)
 SRPM    := jellyfin-web-$(subst -,~,$(VERSION))-$(RELEASE)$(shell rpm --eval %dist).src.rpm
 TARBALL :=$(NAME)-$(subst -,~,$(VERSION)).tar.gz
 
-epel-7-x86_64_repos := https://rpm.nodesource.com/pub_16.x/el/\$$releasever/\$$basearch/
+epel-7-x86_64_repos := https://rpm.nodesource.com/pub_20.x/nodistro/\$$basearch/
```
Renari, line 10: I'm not sure that this will function, the nodesource current docs recommend installing an rpm, which adds a new repo in `/etc/yum.repos.d` that looks like this:
```
[nodesource-nsolid]
name=Nsolid Packages for Linux RPM based distros - $basearch
baseurl=https://rpm.nodesource.com/pub_20.x/nodistro/nsolid/$basearch
priority=8
enabled=1
gpgcheck=1
gpgkey=file:///etc/pki/rpm-gpg/NODESOURCE-NSOLID-GPG-SIGNING-KEY-EL

[nodesource-nodejs]
name=Node.js Packages for Linux RPM based distros - $basearch
baseurl=https://rpm.nodesource.com/pub_20.x/nodistro/nodejs/$basearch
priority=9
enabled=1
gpgcheck=1
gpgkey=file:///etc/pki/rpm-gpg/NODESOURCE-NSOLID-GPG-SIGNING-KEY-EL
```
I believe the missing gpgkey will be an issue when installing this rpm.

Renari: On Fedora 38 I wasn't able to install this package nor the previous one using yum, was this already broken?

deployment/Dockerfile.debian
```diff
@@ -12,9 +12,11 @@ ENV IS_DOCKER=YES
...
 # Prepare Debian build environment
 RUN apt-get update \
-  && apt-get install -y debhelper mmv git curl \
-  && curl -fsSL https://deb.nodesource.com/setup_16.x | bash - \
-  && apt-get install -y nodejs
+  && apt-get install -y debhelper mmv git curl gnupg ca-certificates \
+  && mkdir -p /etc/apt/keyrings \
+  && curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg
+  && echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_20.x nodistro main" | tee /etc/apt/sources.list.d/nodesource.list \
+  && apt-get update
```
thornbill, line 19: Is there an install command missing here?

Renari:
Fixed, I also added nodesource to the fedora dockerfile, it seems the one in their package manager is node 18.

fedora/Makefile
thornbill: Probably... fedora tends to be perpetually broken :roll_eyes:

thornbill (approved):

PR #4804, opened 2023-09-26
Split and extract video range support detection

dmitrylyzo:
Small refactoring + [these](https://github.com/jellyfin/jellyfin-webos/blob/a498a2aae9e7978fc1edecab295010f9c2c62861/frontend/js/webOS.js#L80-L81) options will be used.

**Changes**
Move each video color range test into a separate function.

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -188,6 +188,27 @@ function supportsVc1(videoTestElement) {
     return browser.tizen || browser.web0s || browser.edgeUwp || videoTestElement.canPlayType('video/mp4; codecs="vc-1"').replace(/no/, '');
 }
 
+function supportsHdr10(options) {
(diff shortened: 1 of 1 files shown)


src/scripts/browserDeviceProfile.js
```diff
@@ -188,6 +188,27 @@ function supportsVc1(videoTestElement) {
     return browser.tizen || browser.web0s || browser.edgeUwp || videoTestElement.canPlayType('video/mp4; codecs="vc-1"').replace(/no/, '');
 }
 
+function supportsHdr10(options) {
+    return options.supportsHdr10 ?? (false
```
dmitrylyzo, line 192: `false` here just to make future changes easier.

thornbill: ok I have no objections to that... just need to add comments to suppress the eslint error for these lines :+1:

thornbill (approved):

PR #4806, opened 2023-09-27
Refactor app layouts and common components

thornbill:
**Changes**
This is mostly prep for splitting the dashboard to a separate "app".

* Layouts are restructured so common components are not duplicated in the different apps.
* Shared components are moved to `src/components` from the experimental app directory.
* The mui theme is moved to `src/themes`.
* Redirects are moved so they do not "render" within app layouts.

I'm trying to breakdown the [dashboard changes](https://github.com/jellyfin/jellyfin-web/compare/master...thornbill:jellyfin-web:dashboard-app) so they are less huge.

**Issues**
N/A


src/RootApp.tsx
```diff
@@ -1,37 +1,53 @@
...
 import { History } from '@remix-run/router';
 import React from 'react';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import { ReactQueryDevtools } from '@tanstack/react-query-devtools';
 
-import StableApp from './apps/stable/App';
-import { HistoryRouter } from './components/router/HistoryRouter';
-import { ApiProvider } from './hooks/useApi';
-import { WebConfigProvider } from './hooks/useWebConfig';
+import StableApp from 'apps/stable/App';
```
grafixeyehero, line 8: Can this be load by ladable?

thornbill: It doesn't make too big of a difference since we aren't loading any additional large libraries for the stable layout (like mui in the experimental layout), but every little bit probably helps! :sweat_smile:

grafixeyehero (approved):

PR #4822, opened 2023-10-02
Return boolean through direct comparison

Goku-04:
**Changes**
* Remove ternary operator and simply rely on return value of strict equality (===) 

**Issues**
N/A


src/apps/experimental/components/library/GenresSectionContainer.tsx
```diff
@@ -70,8 +70,8 @@ const GenresSectionContainer: FC<GenresSectionContainerProps> = ({
             centerText: true,
             cardLayout: false,
             shape: itemType === BaseItemKind.MusicAlbum ? 'overflowSquare' : 'overflowPortrait',
-            showParentTitle: itemType === BaseItemKind.MusicAlbum ? true : false,
-            showYear: itemType === BaseItemKind.MusicAlbum ? false : true
+            showParentTitle: (itemType === BaseItemKind.MusicAlbum),
+            showYear: (itemType !== BaseItemKind.MusicAlbum)
```
thornbill, line 74: The parentheses really aren't necessary

```suggestion
            showParentTitle: itemType === BaseItemKind.MusicAlbum,
            showYear: itemType !== BaseItemKind.MusicAlbum
```

thornbill:
We should add this eslint rule to the config to prevent any other issues of this type being added in the future: https://eslint.org/docs/latest/rules/no-unneeded-ternary

thornbill:
@Goku-04 would you care to add that eslint rule as part of this PR? Thanks!

.eslintrc.js
```diff
@@ -63,6 +63,8 @@ module.exports = {
         '@typescript-eslint/no-shadow': ['error'],
         'no-throw-literal': ['error'],
         'no-trailing-spaces': ['error'],
+        'no-unneeded-ternary': ['error'],
+        '@typescript-eslint/no-unneeded-ternary': ['error'],
```
thornbill, line 67: This line isn't needed.

There are some rules that need to be specifically enabled for typescript and disabled in the default rules due to syntax differences in typescript vs javascript. This is not one of those so only the default rule needs enabled. 🙂

Goku-04: so remove the 67th line?

thornbill: yup

Goku-04: done

Goku-04:
The lint rule created some erros in other directories and I think some of them are false errors.

dmitrylyzo:
> The lint rule created some erros in other directories and I think some of them are false errors.

All of them (9) are using unnecessary ternary: `x = condition ? true : false` => `x = condition`. In some cases `!!` (double negation) may be required to cast to boolean.

src/components/playback/playersettingsmenu.js
```diff
@@ -46,7 +46,7 @@ function showQualityMenu(player, btn) {
         const bitrate = parseInt(id, 10);
         if (bitrate !== selectedBitrate) {
             playbackManager.setMaxStreamingBitrate({
-                enableAutomaticBitrateDetection: bitrate ? false : true,
+                enableAutomaticBitrateDetection: !!bitrate,
```
dmitrylyzo, line 49:
```suggestion
                enableAutomaticBitrateDetection: !bitrate,
```

thornbill (approved):

PR #4824, opened 2023-10-02
Add loading indicator and no results message for search

GGProGaming:
Resolves a search not returning anything with some text. A loading circle was also added in case search takes a long while.

**Changes**
- Text added. Example: 
image

- Loading circle added (may be hard to spot since search is fast)

Feedback is more than welcome!

https://github.com/jellyfin/jellyfin-web/assets/84688324/fbef0203-12c0-413f-a8b0-c5aecea34952


src/components/search/SearchResults.tsx
```diff
@@ -112,101 +113,124 @@ const SearchResults: FunctionComponent<SearchResultsProps> = ({ serverId = windo
         setBooks([]);
         setPeople([]);
         setCollections([]);
+        setIsLoading(true);
```
JPVenson, line 116: You could just put `setIsLoading` after the if to reduce the number of calls to it

JPVenson (changes requested):

src/components/search/SearchResults.tsx
GGProGaming: Changed it! Nice catch

JPVenson (approved):

src/components/search/SearchResults.tsx
```diff
@@ -216,93 +241,104 @@ const SearchResults: FunctionComponent<SearchResultsProps> = ({ serverId = windo
...
-                title={globalize.translate('Collections')}
-                items={collections}
-            />
-            <SearchResultsRow
-                title={globalize.translate('People')}
-                items={people}
-                cardOptions={{ coverImage: true }}
-            />
+            {isLoading ? (
+                <div className='loading-circle'></div>
```
thornbill, line 245: You could probably use the `<Loading />` component for this?

https://github.com/jellyfin/jellyfin-web/blob/master/src/components/loading/LoadingComponent.tsx

It will show the global loading component we use in other places.

GGProGaming: Oh I didn't even know that existed! I'll switch that over.

src/components/search/SearchResults.tsx
```diff
@@ -216,93 +242,104 @@ const SearchResults: FunctionComponent<SearchResultsProps> = ({ serverId = windo
...
+                        items={collections}
+                    />
+                    <SearchResultsRow
+                        title={globalize.translate('People')}
+                        items={people}
+                        cardOptions={{ coverImage: true }}
+                    />
+
+                    {allEmpty && !isLoading && (
+                        <div className='sorry-text'>Sorry! No results found for &quot;{query}&quot;</div>
```
thornbill, line 338: This text should be translatable. (Add to `en-us.json` and use `globalize.translate()` to retrieve the appropriate translation.)

src/components/search/searchfields.scss
```diff
@@ -9,3 +9,34 @@
...
+    font-family: inherit;
+    width: 100%;
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%);
+}
+
+// New styles for loading circle
+.loading-circle {
```
thornbill, line 25: Loading styles should no longer be needed?

thornbill (changes requested):

src/components/search/SearchResults.tsx
GGProGaming: My mistake x2!

src/components/search/searchfields.scss
GGProGaming: Fixed.

src/components/search/SearchResults.tsx
GGProGaming: Fixed.

GGProGaming:
Seems to be all good to go

src/strings/en-us.json
```diff
@@ -1427,6 +1427,7 @@
     "SearchForMissingMetadata": "Search for missing metadata",
     "SearchForSubtitles": "Search for Subtitles",
     "SearchResults": "Search Results",
+    "SearchResultsEmpty": "Sorry! No results found for ",
```
thornbill, line 1430: You can use substitutions in translations since some languages may require a different order:

```suggestion
    "SearchResultsEmpty": "Sorry! No results found for \"{0}\"",
```

src/components/search/SearchResults.tsx
```diff
@@ -216,93 +242,104 @@ const SearchResults: FunctionComponent<SearchResultsProps> = ({ serverId = windo
...
+                        items={collections}
+                    />
+                    <SearchResultsRow
+                        title={globalize.translate('People')}
+                        items={people}
+                        cardOptions={{ coverImage: true }}
+                    />
+
+                    {allEmpty && !isLoading && (
+                        <div className='sorry-text'>{globalize.translate('SearchResultsEmpty')}&quot;{query}&quot;</div>
```
thornbill, line 338:
```suggestion
                        <div className='sorry-text'>{globalize.translate('SearchResultsEmpty', query)</div>
```

thornbill (changes requested):

src/components/search/SearchResults.tsx
```diff
@@ -216,93 +242,104 @@ const SearchResults: FunctionComponent<SearchResultsProps> = ({ serverId = windo
...
+                        items={collections}
+                    />
+                    <SearchResultsRow
+                        title={globalize.translate('People')}
+                        items={people}
+                        cardOptions={{ coverImage: true }}
+                    />
+
+                    {allEmpty && !isLoading && (
+                        <div className='sorry-text'>{globalize.translate('SearchResultsEmpty', query)</div>
```
thornbill, line 338: I think I missed this in my previous suggestion :sweat_smile: 

```suggestion
                        <div className='sorry-text'>{globalize.translate('SearchResultsEmpty', query)}</div>
```

src/components/search/SearchResults.tsx
```diff
@@ -216,93 +242,104 @@ const SearchResults: FunctionComponent<SearchResultsProps> = ({ serverId = windo
...
+                        title={globalize.translate('Collections')}
+                        items={collections}
+                    />
+                    <SearchResultsRow
+                        title={globalize.translate('People')}
+                        items={people}
+                        cardOptions={{ coverImage: true }}
+                    />
+
+                    {allEmpty && !isLoading && (
```
thornbill, line 337: This prevents the page crashing on load. The translate function does not handle being passed `undefined` as a parameter.

```suggestion
                    {allEmpty && query && !isLoading && (
```

PR #4825, opened 2023-10-02
Refactor home screen

thornbill:
**Changes**
Refactor home screen code to use separate files per home screen section and remove some unused home screen types

**Issues**
N/A

### DIFF
--- a/src/components/homesections/sections/liveTv.ts
+++ b/src/components/homesections/sections/liveTv.ts
@@ -1,0 +1,181 @@
+import type { BaseItemDto } from '@jellyfin/sdk/lib/generated-client/models/base-item-dto';
+import type { UserDto } from '@jellyfin/sdk/lib/generated-client/models/user-dto';
+import type { ApiClient } from 'jellyfin-apiclient';
+
+import { appRouter } from 'components/router/appRouter';
+import cardBuilder from 'components/cardbuilder/cardBuilder';
+import layoutManager from 'components/layoutManager';
+import ServerConnections from 'components/ServerConnections';
+import globalize from 'scripts/globalize';
+import { getBackdropShape } from 'utils/card';
+
+import type { SectionContainerElement, SectionOptions } from './section';
+
+function getOnNowFetchFn(
+    serverId: string
+) {
+    return function () {
+        const apiClient = ServerConnections.getApiClient(serverId);
+        return apiClient.getLiveTvRecommendedPrograms({
+            userId: apiClient.getCurrentUserId(),
+            IsAiring: true,
+            limit: 24,
+            ImageTypeLimit: 1,
+            EnableImageTypes: 'Primary,Thumb,Backdrop',
+            EnableTotalRecordCount: false,
+            Fields: 'ChannelInfo,PrimaryImageAspectRatio'
+        });
+    };
+}
+
+function getOnNowItemsHtmlFn(
+    { enableOverflow }: SectionOptions
+) {
+    return (items: BaseItemDto[]) => (
+        cardBuilder.getCardsHtml({
+            items: items,
+            preferThumb: 'auto',
+            inheritThumb: false,
+            shape: (enableOverflow ? 'autooverflow' : 'auto'),
+            showParentTitleOrTitle: true,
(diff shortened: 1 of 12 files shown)


src/components/homesections/sections/liveTv.ts
```diff
@@ -0,0 +1,181 @@
...
+            showAirDateTime: false,
+            showAirEndTime: true,
+            defaultShape: getBackdropShape(enableOverflow),
+            lines: 3,
+            overlayPlayButton: true
+        })
+    );
+}
+
+function buildSection(
```
ZenenTreadwell, line 57: This seems rather un-reactive. I'm sure there's a better way to do this with JSX.

src/components/homesections/sections/nextUp.ts
```diff
@@ -0,0 +1,106 @@
...
+    };
+}
+
+export function loadNextUp(
+    elem: HTMLElement,
+    apiClient: ApiClient,
+    userSettings: UserSettings,
+    options: SectionOptions
+) {
+    let html = '';
```
ZenenTreadwell, line 67: Same here - I don't know if optionally concatenating HTML text is the way to go here when JSX supports conditional rendering etc.

src/components/homesections/sections/recentlyAdded.ts
```diff
@@ -0,0 +1,165 @@
...
+}
+
+function renderLatestSection(
+    elem: HTMLElement,
+    apiClient: ApiClient,
+    user: UserDto,
+    parent: BaseItemDto,
+    options: SectionOptions
+) {
+    let html = '';
```
ZenenTreadwell, line 93: needs JSX

src/components/homesections/sections/recentlyAdded.ts
```diff
@@ -0,0 +1,165 @@
...
+    apiClient: ApiClient,
+    user: UserDto,
+    userViews: BaseItemDto[],
+    options: SectionOptions
+) {
+    elem.classList.remove('verticalSection');
+    const excludeViewTypes = ['playlists', 'livetv', 'boxsets', 'channels'];
+    const userExcludeItems = user.Configuration?.LatestItemsExcludes ?? [];
+
+    for (let i = 0, length = userViews.length; i < length; i++) {
```
ZenenTreadwell, line 142: This seems like a good place to use ES6's userViews.forEach() syntax

src/components/homesections/sections/resume.ts
```diff
@@ -0,0 +1,105 @@
...
+    titleLabel: string,
+    mediaType: BaseItemKind,
+    userSettings: UserSettings,
+    options: SectionOptions
+) {
+    let html = '';
+
+    const dataMonitor = dataMonitorHints[mediaType] ?? 'markplayed';
+
+    html += '<h2 class="sectionTitle sectionTitle-cards padded-left">' + globalize.translate(titleLabel) + '</h2>';
```
ZenenTreadwell, line 84: JSX

ZenenTreadwell (commented):
Awesome job doing the refactoring work here! I see a couple places where the code could be improved for readability & conforming to React development standards.

Thanks for putting the effort in!

src/components/homesections/sections/liveTv.ts
grhallenbeck: The source files these are being split off from are legacy and unfortunately not React. Splitting the legacy pieces out this way should make it easier to divide work on migrating to React going forward, though!

ZenenTreadwell: Oops! thanks for saying something

thornbill: Thanks @grhallenbeck that is exactly the idea! I just had not made it back to this to comment myself. :sweat_smile:

grafixeyehero (approved):

bradbeattie:
Regression of https://github.com/jellyfin/jellyfin-web/pull/1134 as a result of this change. Would have expected "Recently Added in Movies" to continue to use the Primary image instead of Backdrops.

image

PR #4826, opened 2023-10-02
Clean backdrops when tab is not selected

### LINKED ISSUE #3916: Backdrops not cleaned up
**Describe The Bug**
Backdrop image containers are retained when rotating to the next image, even though only one is displayed at a time.
The old/unused `div`s are (apparently) not re-used.
This causes performance hiccups when waiting on the home page for a while, especially when having the tab open a for while in the background and then returning.

**Steps To Reproduce**

1. Go to User Preferences
2. Click on Display
3. Scroll down to Backdrops
4. Make sure backdrops are enabled
5. Return to home screen
6. Wait a few rotations of the backdrop images
7. See in `body > div.backdropContainer` (`/html/body/div[1]`) that there are multiple `div.backdropImage`, even though only one is displayed.

**Expected Behavior**
Clean up unused `div`s in `body > div.backdropContainer`.

**Logs**
not applicable

**Screenshots**

 - Platform: Windows 10
 - Browser: Brave 1.43.93 Chromium: 105.0.5195.127
 - Jellyfin Version: 10.8.4

**Additional Context**

Kaysera:
**Changes**

When the tab is unselected, the function `existingBackdropImage.parentNode.remove(existingBackdropImage)` failed silently. Instead, now the parent `backdropContainer` is located and all the divs except the first one are removed. 

**Issues**

Fixes #3916

### DIFF
--- a/src/components/backdrop/backdrop.js
+++ b/src/components/backdrop/backdrop.js
@@ -37,8 +37,11 @@ class Backdrop {
             parent.appendChild(backdropImage);
 
             if (!enableAnimation()) {
-                if (existingBackdropImage?.parentNode) {
-                    existingBackdropImage.parentNode.removeChild(existingBackdropImage);
+                if (existingBackdropImage) {
+                    const backdropContainer = getBackdropContainer();
+                    while (backdropContainer.childNodes.length > 1) {
+                        backdropContainer.removeChild(backdropContainer.firstChild);
+                    }
                 }
                 internalBackdrop(true);
                 return;
@@ -51,8 +54,11 @@ class Backdrop {
                 if (backdropImage === self.currentAnimatingElement) {
                     self.currentAnimatingElement = null;
                 }
-                if (existingBackdropImage?.parentNode) {
-                    existingBackdropImage.parentNode.removeChild(existingBackdropImage);
+                if (existingBackdropImage) {
+                    const backdropContainer = getBackdropContainer();
+                    while (backdropContainer.childNodes.length > 1) {
+                        backdropContainer.removeChild(backdropContainer.firstChild);
+                    }
                 }
             };
 


thornbill:
You don't need to do anything yet... just be patient with me. I'll get to it as soon as I can. 😅

Kaysera:
Sure, thank you for your time.

src/components/backdrop/backdrop.js
```diff
@@ -37,8 +37,11 @@ class Backdrop {
             parent.appendChild(backdropImage);
 
             if (!enableAnimation()) {
-                if (existingBackdropImage?.parentNode) {
-                    existingBackdropImage.parentNode.removeChild(existingBackdropImage);
+                if (existingBackdropImage) {
+                    const backdropContainer = getBackdropContainer();
+                    while (backdropContainer.childNodes.length > 1) {
+                        backdropContainer.removeChild(backdropContainer.firstChild);
```
thornbill, line 43: It looks like another alternative would be to set `textContent = ''`. I suspect this would be more performant than iterating over children.

Kaysera: I have been busy, I will try to do this this week and see if it works.

xtreemze: I agree @Kaysera, perhaps we can use this to not add more backdrops if the tab is not visible using document.visibilityState === 'hidden':


`
function onRotationInterval() {
    if (playbackManager.isPlayingLocally(['Video']) || document.visibilityState === 'hidden') {
        return;
    }

    let newIndex = currentRotationIndex + 1;
    if (newIndex >= currentRotatingImages.length) {
        newIndex = 0;
    }

    currentRotationIndex = newIndex;
    setBackdropImage(currentRotatingImages[newIndex]);
}
`

PR #4827, opened 2023-10-03
Make the series and season Play button play the next episode

### LINKED ISSUE #2570: Play button on thumbnail starts at show's pilot
**Describe the bug**

When clicking the play button of a TV Show a user has already started watching, Jellyfin will start playing from the beginning instead of playing the episode the user is currently at.

 - OS: Debian
 - Virtualization: None, service
 - Clients: Browser
 - Browser: Chrome Version 84.0.4147.89 (Official build) (64-bits)
 - Jellyfin Version: 10.6.0
 - Playback: Transcode

**To Reproduce**

1. Go to TV Show library.
2. Click on the play button that appears when hovering over a TV show the user has already started.
3. Jellyfin will start playing the first episode of the first season, instead of the episode the user is currently at.

**Expected behavior**

Jellyfin should continue playing the next not-watched episode for that specific user.

beef331:
**Changes**

Instead of starting the series or season at the first episode this queues up the next unwatched episode. If the selection is fully watched it marks all unplayed and starts over.

**Issues**

Partially Fixes #2570 but it does not continue a partially watched episode.

### DIFF
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -1848,6 +1848,56 @@ class PlaybackManager {
                     SortBy: options.shuffle ? 'Random' : 'SortName',
                     MediaTypes: 'Audio'
                 });
+            } else if (firstItem.Type === 'Series' || firstItem.Type === 'Season') {
+                const apiClient = ServerConnections.getApiClient(firstItem.ServerId);
+
+                promise = apiClient.getEpisodes(firstItem.SeriesId || firstItem.Id, {
+                    IsVirtualUnaired: false,
+                    IsMissing: false,
+                    UserId: apiClient.getCurrentUserId(),
+                    Fields: 'Chapters'
+                }).then(function (episodesResult) {
+                    const originalResults = episodesResult.Items;
+                    const isSeries = firstItem.Type === 'Series';
+
+                    let foundItem = false;
+
+                    episodesResult.Items = episodesResult.Items.filter(function (e) {
+                        if (foundItem) {
+                            return true;
+                        }
+
+                        if (!e.UserData.Played && (isSeries || e.SeasonId === firstItem.Id)) {
+                            foundItem = true;
+                            return true;
+                        }
+
+                        return false;
+                    });
+
+                    if (episodesResult.Items.length === 0) {
+                        if (isSeries) {
+                            episodesResult.Items = originalResults;
+                        } else {
+                            episodesResult.Items = originalResults.filter(function (e) {
+                                if (foundItem) {
+                                    return true;
+                                }
+
+                                if (e.SeasonId === firstItem.Id) {
+                                    foundItem = true;
+                                    return true;
+                                }
+
+                                return false;
+                            });
+                        }
+                    }
+
+                    episodesResult.TotalRecordCount = episodesResult.Items.length;
+
+                    return episodesResult;
+                });
             } else if (firstItem.IsFolder && firstItem.CollectionType === 'homevideos') {
                 promise = getItemsForPlayback(serverId, mergePlaybackQueries({
                     ParentId: firstItem.Id,


src/components/playback/playbackmanager.js
```diff
@@ -1848,7 +1848,33 @@ class PlaybackManager {
...
+                                    return true;
+                                }
+
+                                return false;
+                            });
+                            episodesResult.TotalRecordCount = episodesResult.Items.length;
+                            resolve(episodesResult);
+                        }, reject);
+                });
+
```
dmitrylyzo, line 1876: I know you took this from the `Episode` block, but could you make it without using Promise nesting?

dmitrylyzo (commented):
I like the new version for its sleek appearance. But we will miss episodes when watch status is:
```
S01E01  v
S01E02  v
S01E03     <-- we unmarked it as 'Played', playback starts from here
S02E01  v  <-- this will be skipped
S02E02
S02E03
```
If this behavior is acceptable, below are some necessary changes.

If no, your first attempt, but without the outer `Promise`.

beef331:
> Please revert the changes to the first attempt, but without the outer Promise. Actually, that's what I expected.

Given that the `resolve` parameter of the `Promise` function was used I'm very much uncertain how you expect to drop the outer promise. I clearly do not have a great understanding of this  code base nor JS :stuck_out_tongue:

src/components/playback/playbackmanager.js
```diff
@@ -1848,6 +1848,31 @@ class PlaybackManager {
                     SortBy: options.shuffle ? 'Random' : 'SortName',
                     MediaTypes: 'Audio'
                 });
+            } else if (firstItem.Type === 'Series') {
+                promise = new Promise(function (resolve, reject){
+                    const apiClient = ServerConnections.getApiClient(firstItem.ServerId);
+                    apiClient.getEpisodes(firstItem.Id, {
```
dmitrylyzo, line 1854: Like this. I'm typing on my phone, so I can't check.
```suggestion
                const apiClient = ServerConnections.getApiClient(firstItem.ServerId);
                promise = apiClient.getEpisodes(firstItem.Id, {
```

src/components/playback/playbackmanager.js
```diff
@@ -1848,6 +1848,31 @@ class PlaybackManager {
...
+                                foundItem = true;
+                                return true;
+                            }
+
+                            return false;
+                        });
+                        episodesResult.TotalRecordCount = episodesResult.Items.length;
+                        resolve(episodesResult);
+                    }, reject);
+                });
```
dmitrylyzo, line 1875:
```suggestion
                        return episodesResult;
                });
```

beef331: Ah thanks, though when it comes to `'Season'` the interaction has to be different since one should queue all episodes after the present one. A separate future PR makes sense methinks.

src/components/playback/playbackmanager.js
```diff
@@ -1848,6 +1848,15 @@ class PlaybackManager {
                     SortBy: options.shuffle ? 'Random' : 'SortName',
                     MediaTypes: 'Audio'
                 });
+            } else if (firstItem.Type === 'Series') {
```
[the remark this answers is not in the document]
dmitrylyzo: https://github.com/jellyfin/jellyfin-web/pull/4827#discussion_r1348019090
> Ah thanks, though when it comes to `'Season'` the interaction has to be different since one should queue all episodes after the present one. A separate future PR makes sense methinks.

This logic can be implemented like:
```js
// Use SeriesId if exists
promise = apiClient.getEpisodes(firstItem.SeriesId || firstItem.Id, {
...
// Skip other seasons
const isSeries = firstItem.Type === 'Series';
...
if (!e.UserData.Played
    && (isSeries || e.SeasonId === firstItem.Id)) {
    foundItem = true;
    return true;
}
```
But it doesn't work if the season is complete (watched) - no items.
We probably should restart it.

Also, maybe for seasons, only the selected one should be played. :thinking:

dmitrylyzo:
Similar to my suggestion for seasons, series currently fails if it is complete.
We probably should restart it.

src/components/playback/playbackmanager.js
```diff
@@ -1848,6 +1848,51 @@ class PlaybackManager {
...
+                        }
+
+                        return false;
+                    });
+
+                    if (episodesResult.Items.length === 0) {
+                        const userId = apiClient.getCurrentUserId()
+                        if (isSeries) {
+                            episodesResult.Items = originalResults;
+                            apiClient.markUnplayed(userId, firstItem.Id, {});
```
dmitrylyzo, line 1878: I don't think we should remove the "Played" mark - the server should do it for us, but right now it doesn't always do it. Moreover, there is a [discussion](https://github.com/jellyfin/jellyfin-meta/discussions/56) about this.

I suggest removing it for now, even though it will cause some inconsistency when rewatching.

dmitrylyzo (approved):
LGTM
I've fixed style to my taste :smile: 

I guess we will need to change `UserData.Played` to something else when that discussion is resolved.

> but it does not continue a partially watched episode.

If you want to fix resuming, you need to pass `startPositionTicks` in `options` somehow:
https://github.com/jellyfin/jellyfin-web/blob/550ad476af38bde1556b2ce3ef1c92cd3626703c/src/components/playback/playbackmanager.js#L2395
But this can be done in another PR.

Also, for the future, it's better to create a separate branch for PR.

PR #4830, opened 2023-10-03
Added artist list concatenation for albums that have over 10 artists

### LINKED ISSUE #4228: If there are quite a few artists in one album they overflow the UI.
**Describe The Bug**
When an album i have in my music library has quite a few artists (about 15) they overflow off the UI bar where the album buttons are. (Heart, Shuffle, Mix, Play)

**Expected Behavior**
I don't expect anything really, besides for it to not overflow lol.
maybe if there are over 10 artists: say something along the lines of "and {} others..." after it says the first 10? 
A quick and dirty inspect element job kinda shows what i mean: 
image

**Screenshots**
image

 - Platform: Linux, ChromeOS
 - Browser: Chrome, Firefox
 - Jellyfin Version: 10.8.7

alfred-delacosta:
**Changes**
Added condition to determine if the artist count of music is over 10 and if so, the displaying text will be changed to only display the first 10 artists and will have "and X other artists." concatenated to the end of it.

**Issues**
Fixes #4228

### DIFF
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -399,8 +399,7 @@ function getArtistLinksHtml(artists, serverId, context) {
     }
 
     if (numberOfArtists > 10) {
-        const remainingNumberOfArtists = numberOfArtists - 10;
-        html.push(`${globalize.translate('AndOtherArtists', remainingNumberOfArtists)}`);
+        html.push(globalize.translate('AndOtherArtists', numberOfArtists - 10));
     }
 
     return html.join(' / ');
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -386,14 +386,29 @@ function reloadUserDataButtons(page, item) {
 
 function getArtistLinksHtml(artists, serverId, context) {
     const html = [];
+    const numberOfArtists = artists.length;
 
-    for (const artist of artists) {
-        const href = appRouter.getRouteUrl(artist, {
-            context: context,
-            itemType: 'MusicArtist',
-            serverId: serverId
-        });
-        html.push('<a style="color:inherit;" class="button-link" is="emby-linkbutton" href="' + href + '">' + escapeHtml(artist.Name) + '</a>');
+    if (numberOfArtists < 10) {
+        for (const artist of artists) {
+            const href = appRouter.getRouteUrl(artist, {
+                context: context,
+                itemType: 'MusicArtist',
+                serverId: serverId
+            });
+            html.push('<a style="color:inherit;" class="button-link" is="emby-linkbutton" href="' + href + '">' + escapeHtml(artist.Name) + '</a>');
+        }
+    } else {
+        for (let i = 0; i < 10; i++) {
+            const artist = artists[i];
+            const href = appRouter.getRouteUrl(artist, {
+                context: context,
+                itemType: 'MusicArtist',
+                serverId: serverId
+            });
+            html.push('<a style="color:inherit;" class="button-link" is="emby-linkbutton" href="' + href + '">' + escapeHtml(artist.Name) + '</a>');
+        }
+        const remainingNumberOfArtists = numberOfArtists - 10;
+        html.push(`${globalize.translate('AndOtherArtists', remainingNumberOfArtists)}`);
     }
 
     return html.join(' / ');
@@ -441,20 +456,8 @@ function renderName(item, container, context) {
     if (parentNameHtml.length) {
         if (parentNameLast) {
             // Music
-            // Determine if there are over 10 artists in the album. If so, concat with 'and X others.'
-            const artists = parentNameHtml[0].split(' / ');
-            const otherArtistCount = artists.length - 10;
-            let newParentNameHtml = artists.slice(0, 10).join(' / ');
-            newParentNameHtml = `${newParentNameHtml} ${globalize.translate('AndOtherArtists', otherArtistCount)}</br>`;
-
             if (layoutManager.mobile) {
-                if (artists.length > 10) {
-                    html = '<h3 class="parentName musicParentName ">' + newParentNameHtml + '</h3>';
-                } else {
-                    html = '<h3 class="parentName musicParentName">' + parentNameHtml.join('</br>') + '</h3>';
-                }
-            } else if (artists.length > 10) {
-                html = '<h3 class="parentName musicParentName">' + newParentNameHtml + '</h3>';
+                html = '<h3 class="parentName musicParentName">' + parentNameHtml.join('</br>') + '</h3>';
             } else {
                 html = '<h3 class="parentName musicParentName focuscontainer-x">' + parentNameHtml.join(' - ') + '</h3>';
             }
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -441,8 +441,20 @@ function renderName(item, container, context) {
     if (parentNameHtml.length) {
         if (parentNameLast) {
             // Music
+            // Determine if there are over 10 artists in the album. If so, concat with 'and X others.'
+            const artists = parentNameHtml[0].split(' / ');
+            const otherArtistCount = artists.length - 10;
+            let newParentNameHtml = artists.slice(0, 10).join(' / ');
+            newParentNameHtml = `${newParentNameHtml} and ${otherArtistCount} other artists.</br>`;
+
             if (layoutManager.mobile) {
-                html = '<h3 class="parentName musicParentName">' + parentNameHtml.join('</br>') + '</h3>';
+                if (artists.length > 10) {
+                    html = '<h3 class="parentName musicParentName ">' + newParentNameHtml + '</h3>';
+                } else {
+                    html = '<h3 class="parentName musicParentName">' + parentNameHtml.join('</br>') + '</h3>';
+                }
+            } else if (artists.length > 10) {
+                html = '<h3 class="parentName musicParentName">' + newParentNameHtml + '</h3>';
             } else {
                 html = '<h3 class="parentName musicParentName focuscontainer-x">' + parentNameHtml.join(' - ') + '</h3>';
             }
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -401,7 +401,7 @@ function getArtistLinksHtml(artists, serverId, context) {
     let fullHtml = html.join(' / ');
 
     if (numberOfArtists > 10) {
-        fullHtml += globalize.translate('AndOtherArtists', numberOfArtists - 10);
+        fullHtml = globalize.translate('AndOtherArtists', fullHtml, numberOfArtists - 10);
     }
 
     return fullHtml;
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -445,7 +445,7 @@ function renderName(item, container, context) {
             const artists = parentNameHtml[0].split(' / ');
             const otherArtistCount = artists.length - 10;
             let newParentNameHtml = artists.slice(0, 10).join(' / ');
-            newParentNameHtml = `${newParentNameHtml} and ${otherArtistCount} other artists.</br>`;
+            newParentNameHtml = `${newParentNameHtml} ${globalize.translate('AndOtherArtists', otherArtistCount)}</br>`;
 
             if (layoutManager.mobile) {
                 if (artists.length > 10) {
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -388,25 +388,17 @@ function getArtistLinksHtml(artists, serverId, context) {
     const html = [];
     const numberOfArtists = artists.length;
 
-    if (numberOfArtists < 10) {
-        for (const artist of artists) {
-            const href = appRouter.getRouteUrl(artist, {
-                context: context,
-                itemType: 'MusicArtist',
-                serverId: serverId
-            });
-            html.push('<a style="color:inherit;" class="button-link" is="emby-linkbutton" href="' + href + '">' + escapeHtml(artist.Name) + '</a>');
-        }
-    } else {
-        for (let i = 0; i < 10; i++) {
-            const artist = artists[i];
-            const href = appRouter.getRouteUrl(artist, {
-                context: context,
-                itemType: 'MusicArtist',
-                serverId: serverId
-            });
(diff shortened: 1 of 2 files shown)


alfred-delacosta:
Screenshot of fix:
image

src/controllers/itemDetails/index.js
```diff
@@ -439,8 +439,20 @@ function renderName(item, container, context) {
     if (parentNameHtml.length) {
         if (parentNameLast) {
             // Music
+            // Determine if there are over 10 artists in the album. If so, concat with 'and X others.'
+            const artists = parentNameHtml[0].split(' / ');
+            const otherArtistCount = artists.length - 10;
+            let newParentNameHtml = artists.slice(0, 10).join(' / ');
+            newParentNameHtml = `${newParentNameHtml} and ${otherArtistCount} other artists.</br>`;
```
grhallenbeck, line 446: I think we'd probably want to add an l10n string for this and then use `globalize.translate()` here. The l10n strings are defined under `src/strings`: https://github.com/jellyfin/jellyfin-web/blob/master/src/strings/en-us.json

It looks like the l10n lib supports string templates, so that should be possible with one string definition.

oddstr13:
Please review https://jellyfin.org/docs/general/contributing/development#pull-request-guidelines

Pull-request titles will eventually end up in the release change-log, so make sure it's short and descriptive :smiley_cat:

src/controllers/itemDetails/index.js
alfred-delacosta: Okay. I haven't worked with the l10n library before so I am going to look at other code on how to do it. I will submit my PR shortly. Let me know if it looks correct.

src/controllers/itemDetails/index.js
```diff
@@ -439,8 +439,20 @@ function renderName(item, container, context) {
...
+
             if (layoutManager.mobile) {
-                html = '<h3 class="parentName musicParentName">' + parentNameHtml.join('</br>') + '</h3>';
+                if (artists.length > 10) {
+                    html = '<h3 class="parentName musicParentName ">' + newParentNameHtml + '</h3>';
+                } else {
+                    html = '<h3 class="parentName musicParentName">' + parentNameHtml.join('</br>') + '</h3>';
+                }
+            } else if (artists.length > 10) {
+                html = '<h3 class="parentName musicParentName">' + newParentNameHtml + '</h3>';
```
grhallenbeck, line 455: I'm not familiar with the styles here so maybe someone else with some more familiarity with the project can answer: Is the `focuscontainer-x` classname needed here like it's applied to the `h3` in the `else` below?

thornbill: It should probably be included. The `focuscontainer` class is used to manage focus for keyboards/remotes. `focuscontainer-x` would "trap" the focus in this element when navigating on the x axis (until the last focusable element in the container is reached).

alfred-delacosta: Should I mark this conversation as resolved?

src/controllers/itemDetails/index.js
```diff
@@ -439,8 +439,20 @@ function renderName(item, container, context) {
     if (parentNameHtml.length) {
         if (parentNameLast) {
             // Music
+            // Determine if there are over 10 artists in the album. If so, concat with 'and X others.'
+            const artists = parentNameHtml[0].split(' / ');
+            const otherArtistCount = artists.length - 10;
+            let newParentNameHtml = artists.slice(0, 10).join(' / ');
```
thornbill, line 445: This approach of parsing the generated html doesn't seem like a great approach... I think it would be best to handle this in the `getArtistLinksHtml` function:

https://github.com/jellyfin/jellyfin-web/blob/b7f341ce087d1e5a4448121bfb7a743717b2e1ef/src/controllers/itemDetails/index.js#L385-L398

thornbill (changes requested):

src/controllers/itemDetails/index.js
alfred-delacosta: Yeah I would agree with that. I will update the code.

alfred-delacosta: Code updated in latest push.

src/controllers/itemDetails/index.js
```diff
@@ -384,14 +384,29 @@ function reloadUserDataButtons(page, item) {
...
+            const href = appRouter.getRouteUrl(artist, {
+                context: context,
+                itemType: 'MusicArtist',
+                serverId: serverId
+            });
+            html.push('<a style="color:inherit;" class="button-link" is="emby-linkbutton" href="' + href + '">' + escapeHtml(artist.Name) + '</a>');
+        }
+        const remainingNumberOfArtists = numberOfArtists - 10;
+        html.push(`${globalize.translate('AndOtherArtists', remainingNumberOfArtists)}`);
     }
```
thornbill, line 410:
```suggestion
    for (let i = 0; i < Math.min(numberOfArtists, 10); i++) {
        const artist = artists[i];
        const href = appRouter.getRouteUrl(artist, {
            context,
            itemType: 'MusicArtist',
            serverId
        });
        html.push('<a style="color:inherit;" class="button-link" is="emby-linkbutton" href="' + href + '">' + escapeHtml(artist.Name) + '</a>');
    }

    if (numberOfArtists > 10) {
        const remainingNumberOfArtists = numberOfArtists - 10;
        html.push(`${globalize.translate('AndOtherArtists', remainingNumberOfArtists)}`);
    }
```

src/controllers/itemDetails/index.js
```diff
@@ -384,16 +384,23 @@ function reloadUserDataButtons(page, item) {
...
             itemType: 'MusicArtist',
-            serverId: serverId
+            serverId
         });
         html.push('<a style="color:inherit;" class="button-link" is="emby-linkbutton" href="' + href + '">' + escapeHtml(artist.Name) + '</a>');
     }
 
+    if (numberOfArtists > 10) {
+        const remainingNumberOfArtists = numberOfArtists - 10;
+        html.push(`${globalize.translate('AndOtherArtists', remainingNumberOfArtists)}`);
```
thornbill, line 401:
```suggestion
        html.push(globalize.translate('AndOtherArtists', numberOfArtists - 10));
```

thornbill (approved):
Code looks good now. It would be nice if someone could confirm the functionality works as expected... I don't seem to have any albums available that have enough artists to test. :sweat_smile:

oddstr13:
Before
---------
image

After
-------
image

image

#### TV mode
image

#### Mobile
image

image


#### Experimental
image
image

---
The optimal number of entries seem to be highly dependent on resolution, display mode and potentially theme.
There does not seem to be a way to see all album artists with this PR?
From an UX perspective, I sort of expected the "and # other artists" to be interactive in some way, to actually be able to see them.

src/controllers/itemDetails/index.js
```diff
@@ -384,16 +384,22 @@ function reloadUserDataButtons(page, item) {
...
+            serverId
         });
         html.push('<a style="color:inherit;" class="button-link" is="emby-linkbutton" href="' + href + '">' + escapeHtml(artist.Name) + '</a>');
     }
 
+    if (numberOfArtists > 10) {
+        html.push(globalize.translate('AndOtherArtists', numberOfArtists - 10));
+    }
+
     return html.join(' / ');
```
thornbill, line 403:
```suggestion
    let fullHtml = html.join(' / ');

    if (numberOfArtists > 10) {
        fullHtml += globalize.translate('AndOtherArtists', numberOfArtists - 10);
    }

    return fullHtml;
```

thornbill (changes requested):
This should fix the extra slash showing up before the "and other artists" text.

twihno:
I don't think this PR solves the issue. It's a good feature to have if you **do** have more than 10 **album** artists but not if it's a compilation.
I wrote a quick overview of my interpretation in the linked issue. I don't want to duplicate the comment, so here's the link: #4228

felix920506:
I just tested this PR using the version deployed to Cloudflare pages and there are 2 items that would need to be addressed. 
1. There should be a space between the last artist and "and xxx other artists"
Screenshot from 2024-04-22 22-58-36

2. This falls apart when the browser window is resized to something narrow. I think this s an acceptable compromise for this PR as it is already a significant improvement over the current situation. This should however have a new issue opened for it.
image

src/controllers/itemDetails/index.js
```diff
@@ -385,17 +385,25 @@ function reloadUserDataButtons(page, item) {
...
+            serverId
         });
         html.push('<a style="color:inherit;" class="button-link" is="emby-linkbutton" href="' + href + '">' + escapeHtml(artist.Name) + '</a>');
     }
 
-    return html.join(' / ');
+    let fullHtml = html.join(' / ');
+
+    if (numberOfArtists > 10) {
+        fullHtml += globalize.translate('AndOtherArtists', numberOfArtists - 10);
```
felix920506, line 403:
```suggestion
		fullHtml += ' '
        fullHtml += globalize.translate('AndOtherArtists', numberOfArtists - 10);
```
This should fix the problem I pointed out above

felix920506 (changes requested):

src/controllers/itemDetails/index.js
felix920506: I just realized this way of doing things is problematic for RTL languages localization

felix920506, line 403:
```suggestion
        fullHtml = globalize.translate('AndOtherArtists', fullHtml, numberOfArtists - 10);
```

src/strings/en-us.json
```diff
@@ -44,6 +44,7 @@
     "AllowTonemappingHelp": "Tone-mapping can transform the dynamic range of a video from HDR to SDR while maintaining image details and colors, which are very important information for representing the original scene. Currently works only with 10bit HDR10, HLG and DoVi videos. This requires the corresponding GPGPU runtime.",
     "AlwaysPlaySubtitles": "Always Play",
     "AlwaysPlaySubtitlesHelp": "Subtitles matching the language preference will be loaded regardless of the audio language.",
+    "AndOtherArtists": "and {0} other artists.",
```
felix920506, line 47:
```suggestion
    "AndOtherArtists": "{0} and {1} other artists.",
```

felix920506 (changes requested):
These 2 changes should address my comment above and improve compatibility with RTL languages when doing localization.

1hitsong:
Instead of updating the number of artists shown to a hard-coded value, can we instead just use CSS to limit the displayed artists to 2 lines?

```css
display: -webkit-box;
-webkit-line-clamp: 2;
-webkit-box-orient: vertical;  
overflow: hidden;```

**Comment by alfred-delacosta** [comment]
> These 2 changes should address my comment above and improve compatibility with RTL languages when doing localization.
> 
> This image is with the changes below applied image
> 
> Changing the order of which the lists of artists appear in the strings image using source string `{1} artists and {0}`

So the first image shows what it would look like in a LTR language setting and the second image shows would it would look like in a RTL language setting? I just want to confirm that's my understanding of your post and changes.

**Comment by thornbill** [comment]
> So the first image shows what it would look like in a LTR language setting and the second image shows would it would look like in a RTL language setting? I just want to confirm that's my understanding of your post and changes.

Yes that is correct

**Comment by alfred-delacosta** [comment]
> > So the first image shows what it would look like in a LTR language setting and the second image shows would it would look like in a RTL language setting? I just want to confirm that's my understanding of your post and changes.
> 
> Yes that is correct

Fantastic, then I am going to commit those changes.

**Comment by alfred-delacosta** [comment]
> > So the first image shows what it would look like in a LTR language setting and the second image shows would it would look like in a RTL language setting? I just want to confirm that's my understanding of your post and changes.
> 
> Yes that is correct

Fantastic then I am going to commit those changes.

**Review by alfred-delacosta — COMMENTED** [review]
I'm pretty sure I accepted these changes...

**Review by alfred-delacosta — COMMENTED** [review]
I guess I missed these at some point?...

**Comment by alfred-delacosta** [comment]
@thornbill Any idea as to why I can't get rid of this message on the PR? I already committed the changes and did so a while ago. At least I thought I did.
image

File: src/strings/en-us.json [hunk 9]
```diff
@@ -46,6 +46,7 @@
     "AllowTonemappingHelp": "Tone-mapping can transform the dynamic range of a video from HDR to SDR while maintaining image details and colors, which are very important information for representing the original scene. Currently works only with 10bit HDR10, HLG and DoVi videos. This requires the corresponding GPGPU runtime.",
     "AlwaysPlaySubtitles": "Always Play",
     "AlwaysPlaySubtitlesHelp": "Subtitles matching the language preference will be loaded regardless of the audio language.",
+    "AndOtherArtists": "{0} and {1} other artists.",
```
dmitrylyzo, line 49: In some languages, the plural has more than one form.

For Russian:
`... and 1 other artists.` can be translated as `... и ещё 1 артист`. In fact, here we need singular form.
`... and 2 other artists.` can be translated as `... и ещё 2 артиста`.
`... and 4 other artists.` can be translated as `... и ещё 4 артиста`.
`... and 5 other artists.` can be translated as `... и ещё 5 артистов`.
`... and 6 other artists.` can be translated as `... и ещё 6 артистов`.
`... and 20 other artists.` can be translated as `... и ещё 20 артистов`.
`... and 21 other artists.` can be translated as `... и ещё 21 артист`.
`... and 22 other artists.` can be translated as `... и ещё 22 артиста`.
`... and 23 other artists.` can be translated as `... и ещё 23 артиста`.
`... and 25 other artists.` can be translated as `... и ещё 25 артистов`.

I.e. we need:
- singular form: `{0} and {1} other artist.`.
- plural form 1: `{0} and {1} other artists.` for numbers >20 that match`[0-9]+1`.
- plural form 2: `{0} and {1} other artists.` for numbers >20 that match `[0-9]+[234]`.
- plural form 3: `{0} and {1} other artists.` for everything else.

nielsvanvelzen: Isn't this a limitation of our i18n code in general? I believe most strings don't have proper plural support.

dmitrylyzo: It is. I have no idea other than introducing 3 forms (according to Qt docs - we can merge singular and form 1 from the above) and select string by those rules manually.

Or we need some more featured i18n library. If we want proper localization, of course. 🤷

thornbill: This is a general issue with our translation code. We should investigate moving to a "real" library that handles this properly, but it shouldn't be a blocker for this.

thornbill (approved):

PR #4831, opened 2023-10-03
Refactor reloadData function

1337Nerd:
**Changes**
Fix code smell from sonarCloud, reduce code complexity. Possible unintended effect is the check for startIndex on line 80, but this is currently how it functions so it has been kept for consistency.

**Issues**
N/A

### DIFF
--- a/src/components/activitylog.js
+++ b/src/components/activitylog.js
@@ -61,20 +61,13 @@ function renderList(elem, apiClient, result) {
 }
 
 function reloadData(instance, elem, apiClient, startIndex, limit) {
-    if (startIndex == null) {
-        startIndex = parseInt(elem.getAttribute('data-activitystartindex') || '0', 10);
-    }
-
-    limit = limit || parseInt(elem.getAttribute('data-activitylimit') || '7', 10);
+    startIndex ??= parseInt(elem.getAttribute('data-activitystartindex') || '0', 10);
+    limit ||= parseInt(elem.getAttribute('data-activitylimit') || '7', 10);
     const minDate = new Date();
     const hasUserId = toBoolean(elem.getAttribute('data-useractivity'), true);
-
+    const dateOffset = hasUserId ? 1 : 7; // one day back if user id, otherwise one week
     // TODO: Use date-fns
-    if (hasUserId) {
-        minDate.setTime(minDate.getTime() - 24 * 60 * 60 * 1000); // one day back
-    } else {
-        minDate.setTime(minDate.getTime() - 7 * 24 * 60 * 60 * 1000); // one week back
-    }
+    minDate.setTime(minDate.getTime() - dateOffset * 24 * 60 * 60 * 1000);
 
     ApiClient.getJSON(ApiClient.getUrl('System/ActivityLog/Entries', {
         startIndex: startIndex,
@@ -84,16 +77,9 @@ function reloadData(instance, elem, apiClient, startIndex, limit) {
     })).then(function (result) {
         elem.setAttribute('data-activitystartindex', startIndex);
         elem.setAttribute('data-activitylimit', limit);
-        if (!startIndex) {
+        if (startIndex === 0) {
             const activityContainer = dom.parentWithClass(elem, 'activityContainer');
-
-            if (activityContainer) {
-                if (result.Items.length) {
-                    activityContainer.classList.remove('hide');
-                } else {
-                    activityContainer.classList.add('hide');
-                }
-            }
+            activityContainer?.classList.toggle('hide', result.Items.length === 0);
         }
 
         instance.items = result.Items;


src/components/activitylog.js
```diff
@@ -61,20 +61,13 @@ function renderList(elem, apiClient, result) {
...
-        startIndex = parseInt(elem.getAttribute('data-activitystartindex') || '0', 10);
-    }
-
-    limit = limit || parseInt(elem.getAttribute('data-activitylimit') || '7', 10);
+    startIndex ??= parseInt(elem.getAttribute('data-activitystartindex') || '0', 10);
+    limit ||= parseInt(elem.getAttribute('data-activitylimit') || '7', 10);
     const minDate = new Date();
     const hasUserId = toBoolean(elem.getAttribute('data-useractivity'), true);
-
+    const dateOffset = hasUserId ? 1 : 7; // one day back if user id, otherwise one week
```
thornbill, line 68:
```suggestion
    const daysOffset = hasUserId ? 1 : 7; // one day back if user id, otherwise one week
```

src/components/activitylog.js
```diff
@@ -84,16 +77,9 @@ function reloadData(instance, elem, apiClient, startIndex, limit) {
     })).then(function (result) {
         elem.setAttribute('data-activitystartindex', startIndex);
         elem.setAttribute('data-activitylimit', limit);
-        if (!startIndex) {
+        if (startIndex === 0) {
             const activityContainer = dom.parentWithClass(elem, 'activityContainer');
```
thornbill, line 81: it looks like this assignment can be removed now and just combine the element selection with the class toggle statement

src/components/activitylog.js
```diff
@@ -61,20 +61,13 @@ function renderList(elem, apiClient, result) {
 }
 
 function reloadData(instance, elem, apiClient, startIndex, limit) {
-    if (startIndex == null) {
-        startIndex = parseInt(elem.getAttribute('data-activitystartindex') || '0', 10);
-    }
-
-    limit = limit || parseInt(elem.getAttribute('data-activitylimit') || '7', 10);
+    startIndex ??= parseInt(elem.getAttribute('data-activitystartindex') || '0', 10);
+    limit ||= parseInt(elem.getAttribute('data-activitylimit') || '7', 10);
```
thornbill, line 65: the `??=` and `||=` operators probably produce some gnarly code when babelified which is something we've been trying to avoid lately

thornbill (changes requested):

src/components/activitylog.js
dmitrylyzo: `x ||= y` is probably fine : `x || (x = y)`

`x ??= y` produces `var _x; (_x = x) !== void 0 && _x !== null ? _x : (x = y)`. So I'd only use it if `||` doesn't fit. And probably if `y` is harder to calculate than checking 2 conditions.

1337Nerd: Converted it to use ||=

src/components/activitylog.js
```diff
@@ -84,17 +77,8 @@ function reloadData(instance, elem, apiClient, startIndex, limit) {
...
-            if (activityContainer) {
-                if (result.Items.length) {
-                    activityContainer.classList.remove('hide');
-                } else {
-                    activityContainer.classList.add('hide');
-                }
-            }
-        }
+        const activityContainer = dom.parentWithClass(elem, 'activityContainer');
+        activityContainer?.classList.toggle('hide', result.Items.length === 0 && startIndex === 0);
```
thornbill, line 81: Sorry if it wasn't clear, this is what I meant:

```suggestion
        if (startIndex === 0) {
            dom.parentWithClass(elem, 'activityContainer')?.classList
                .toggle('hide', result.Items.length === 0 && startIndex === 0);
        }
```

1337Nerd: Is the if statement around the class toggle needed? Since it's checking for startIndex === 0 in the toggle statement

dmitrylyzo: I'd say yes to avoid calling `parentWithClass`.

thornbill: This is covered by may suggested change I think... it's basically updating the visibility of the container when the index is `0`.

thornbill: @dmitrylyzo do you agree that my recommended change should work here?

dmitrylyzo: > @dmitrylyzo do you agree that my recommended change should work here?

In terms of refactor, it is good because the original logic is preserved.
And probably the only correct one. See below.

> What if we get `result.Items.length: 0` with `startIndex: 0` on update (server notification)? We should probably hide the container anyway. 🤔 In `finally`?

I just realized I said the opposite of what I meant to say. :sweat_smile: 
What if we get `result.Items.length === 0` with **`startIndex !== 0`** on update (server notification)?

I tried this (with the original logic):
1. Remove `hide` from the "Alerts" container (it is usually empty).
2. Change `data-activitystartindex` of its child list to `1`. _Think of this as using pagination._
3. Run music in another tab -> logs are going to be updated.
4. "Alerts" container remains visible.

The original logic (which will be preserved by Bill's suggestion) changes visibility only at the start (`startIndex === 0`) and keeps the container visible if there are no more items.

`result.Items.length === 0 && startIndex === 0` will **show** the block just because `startIndex !== 0` - even worse.

Anyway, these blocks don't have pagination and I think we just need to not break what we have. So I'm in favor of leaving `if (startIndex === 0)`.

PR #4833, opened 2023-10-03
Remove unnecessary renaming

tehciolo:
**Changes**
- remove `startMultiSelect` renaming to address code smell below

**Issues**
Fixes [this code smell](https://sonarcloud.io/project/issues?cleanCodeAttributeCategories=INTENTIONAL&resolved=false&rules=javascript%3AS6650&id=jellyfin_jellyfin-web&open=AYljc2a2J8YZ6p3GEUmv)

----

I do not know how to manually test this in the UI, but I will happily do so if given minor pointers.

That being said, the change should be SAFE.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -66,6 +66,7 @@ module.exports = {
         'no-unused-expressions': ['off'],
         '@typescript-eslint/no-unused-expressions': ['error', { 'allowShortCircuit': true, 'allowTernary': true, 'allowTaggedTemplates': true }],
         'no-unused-private-class-members': ['error'],
+        'no-useless-rename': ['error'],
         'no-useless-constructor': ['off'],
         '@typescript-eslint/no-useless-constructor': ['error'],
         'no-var': ['error'],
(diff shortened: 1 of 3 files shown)


grhallenbeck:
This removes a redundant rename so would be a very low risk change, but here's what I can see for testing:

This case is hit when choosing the `Select` option after right clicking on an item in a menu. So you should be able to test it still works by performing that select option and making sure it still triggers the multiselect flow as expected.

thornbill:
Please add this rule to our eslint config to prevent any new instances of this being added in the future: https://eslint.org/docs/latest/rules/no-useless-rename

tehciolo:
> This removes a redundant rename so would be a very low risk change, but here's what I can see for testing:
> 
> This case is hit when choosing the `Select` option after right clicking on an item in a library. So you should be able to test it still works by performing that select option and making sure it still triggers the multiselect flow as expected.

Thanks! Managed to manually test and all seems good.

tehciolo:
> Please add this rule to our eslint config to prevent any new instances of this being added in the future: https://eslint.org/docs/latest/rules/no-useless-rename

Done.

thornbill (approved):

PR #4834, opened 2023-10-04
Migrate React Router to React Data Router pattern

grhallenbeck:
**Changes**

Migrates routing from React Router components to a data router using `createHashRouter`. There was logic previously in a custom HistoryRouter implementation that handles syncing a router history between the experimental React application and the legacy router, as well as handling legacy hash-bang routes, which has been moved into a `useLegacyRouterSync` hook. This allows syncing changes from one router to the other since data routers don't expose their internal history or allow an externally-initialized history.

There is some risk here with the history sync logic around potential double history entries, which should be managed with checks against both histories to make sure they aren't already the same before updating one in response to the other. This would probably be a good early target for some unit testing once a testing library has been chosen.

**Issues**


src/apps/experimental/routes/routes.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { toAsyncPageRoute } from 'components/router/AsyncRoute';
+import { toViewManagerPageRoute } from 'components/router/LegacyRoute';
+import { toRedirectRoute } from 'components/router/Redirect';
+import AppLayout from '../AppLayout';
+import { ASYNC_USER_ROUTES } from './asyncRoutes';
+import { LEGACY_PUBLIC_ROUTES, LEGACY_USER_ROUTES } from './legacyRoutes';
+
+export const EXPERIMENTAL_APP_ROUTES: RouteObject[] = [
+    {
+        path: '/',
```
grafixeyehero, line 16: as mention on this https://github.com/jellyfin/jellyfin-web/pull/4660#issuecomment-1580741902
```suggestion
        path: '/*',
```

grafixeyehero (changes requested):
LGTM

src/apps/experimental/routes/routes.tsx
grhallenbeck: Good catch! I didn't notice any warnings about this, but there's quite a bit of logging going on in the console to sift through. I'll update when I resolve conflicts.

grhallenbeck: Updated!

grafixeyehero (approved):

thornbill (approved):

PR #4836, opened 2023-10-04
Fix FilterMenu not saving simple filters

dmitrylyzo:
Regression 6341a71fec882726bdffa4ed3460a6cbcf252d6e

**Changes**
- Move "simpleFilter" query to where it is used.
- Move queries into loops.

**Issues**
FilterMenu doesn't save simple filters because of `elems` override.

### DIFF
--- a/src/components/filtermenu/filtermenu.js
+++ b/src/components/filtermenu/filtermenu.js
@@ -103,7 +103,7 @@ function onInputCommand(e) {
     }
 }
 function saveValues(context, settings, settingsKey, setfilters) {
-    let elems = context.querySelectorAll('.simpleFilter');
+    let elems;
 
     // Video type
     const videoTypes = [];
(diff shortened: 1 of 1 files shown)


src/components/filtermenu/filtermenu.js
```diff
@@ -103,35 +103,27 @@ function onInputCommand(e) {
...
-    let elems = context.querySelectorAll('.simpleFilter');
-
     // Video type
     const videoTypes = [];
-    elems = context.querySelectorAll('.chkVideoTypeFilter');
-
-    for (let i = 0, length = elems.length; i < length; i++) {
-        if (elems[i].checked) {
-            videoTypes.push(elems[i].getAttribute('data-filter'));
+    for (const elem of context.querySelectorAll('.chkVideoTypeFilter')) {
```
dmitrylyzo, line 108: `forEach` is more portable (for legacy) than `for..of`. :thinking:

thornbill: It sounds like we should setup some eslint rules to prevent us from using some of these language features that make generated code more complex since it has came up a few times now. I can also disable related rules on the sonar side as needed.

dmitrylyzo: Tried `eslint-plugin-no-for-of-loops`: 147 errors :sweat_smile:

thornbill: Do you want to change the usages here or leave it as is?

grafixeyehero (approved):

PR #4843, opened 2023-10-05
Use cast receiver application from user configuration

nielsvanvelzen:
Touching this code made me cry.

**Changes**
- Delay cast plugin initialization until user is signed in
- Use the cast receiver id from the user configuration
- Update playback settings page to use new cast receiver magic

~~The user configuration page will be in a separate PR because touching this code is dirty and I can't look at it for too long.~~

**Issues**

META jellyfin/jellyfin-meta#45
SERVER jellyfin/jellyfin#10270

### DIFF
--- a/src/plugins/chromecastPlayer/plugin.js
+++ b/src/plugins/chromecastPlayer/plugin.js
@@ -58,11 +58,6 @@ const PLAYER_STATE = {
     'ERROR': 'ERROR'
 };
 
-// production version registered with google
-// replace this value if you want to test changes on another instance
-const applicationStable = 'F007D354';
-const applicationUnstable = '6F511C87';
-
 const messageNamespace = 'urn:x-cast:com.connectsdk';
 
 class CastPlayer {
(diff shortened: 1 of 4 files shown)


src/plugins/chromecastPlayer/plugin.js
```diff
@@ -106,18 +101,21 @@ class CastPlayer {
...
         }
 
-        let applicationID = userSettings.chromecastVersion();
-        if (applicationID === 'stable') applicationID = applicationStable;
-        if (applicationID === 'unstable') applicationID = applicationUnstable;
+        const apiClient = ServerConnections.currentApiClient();
+        const userId = apiClient.getCurrentUserId();
+
+        apiClient.getUser(userId).then(user => {
+            const applicationID = user.Configuration.CastReceiverId;
```
crobibero, line 108: What happens if `CastReceiverId` is null? Does the session request just error out?

nielsvanvelzen: I kinda forgot about that so I'll probably need to make some changes. It should probably also disable the cast button in the play to menu.

nielsvanvelzen: I've added a check to stop the cast init in case of a falsy value, this will automatically hide the cast button from the play to menu 👍

thornbill (approved):

PR #4847, opened 2023-10-06
Remove cache from docker images after installing packages

arminius-smh:
**Changes**
Trims the docker images for centos and fedora by ~200MB, debian and portable by ~20MB and docker by ~2MB by deleting cache after installing packages during the build process of the docker images.

**Issues**
N/A


deployment/Dockerfile.debian
```diff
@@ -17,7 +17,8 @@ RUN apt-get update \
   && curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg \
   && echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_20.x nodistro main" | tee /etc/apt/sources.list.d/nodesource.list \
   && apt-get update \
-  && apt-get install -y nodejs
+  && apt-get install -y nodejs \
+  && rm -rf /var/lib/apt/lists/*
```
joshuaboniface, line 21: Looks good but should also probably do an `apt-get clean` and/or `rm -r /var/cache/apt/archives` too. Not sure the equivalent in Fedora/CentOS but I imagine that will make them even smaller unless `dnf/yum clean` already does that.

arminius-smh: Sure I've added that as well and also the equivalent for Fedora and CentOS, apt-get clean should be run by every debian container automatically, if I understand `docker run --rm debian cat /etc/apt/apt.conf.d/docker-clean` correctly, so that shouldn't be needed:

```
[...]
# so we get this fun creation that's essentially just "apt-get clean".
DPkg::Post-Invoke { "rm -f /var/cache/apt/archives/*.deb /var/cache/apt/archives/partial/*.deb /var/cache/apt/*.bin || true"; };
APT::Update::Post-Invoke { "rm -f /var/cache/apt/archives/*.deb /var/cache/apt/archives/partial/*.deb /var/cache/apt/*.bin || true"; };

Dir::Cache::pkgcache "";
Dir::Cache::srcpkgcache "";
```

joshuaboniface (approved):

PR #4848, opened 2023-10-06
Clarify where fallback fonts are used

felix920506:
Currently the fallback fonts are only used by the web player when rendering subtitles.
JMP and swiftfin use the client system fonts as far as I can tell, and burning in subtitles use server side system fonts.

This PR intends to properly document this setting and avoid confusion.

Related Issues:
[jellyfin/#5008](https://github.com/jellyfin/jellyfin/issues/5008)
[jellyfin/#9273](https://github.com/jellyfin/jellyfin/issues/9273)


nielsvanvelzen:
I think it would be better to link to the documentation at jellyfin.org because if another client does start using the fonts we cannot easily update this help text without a new release.

felix920506:
I think this is a fair point. However I don't see a suitable place for this in the docs currently.

felix920506:
I am closing this one and making another PR using the link in docs instead.

PR #4854, opened 2023-10-08
Add icon to item context menu that allows users to remove item from 'next up'

BrennanWoodbury:
**Changes**

Added a button to the Item Context Menu to call a new backend api endpoint ExcludeFromNextUp, which is included in backend PR [10367](https://github.com/jellyfin/jellyfin/pull/10367) (which should be treated as a dependency for this PR). This essentially skips an episode from the next up UI.


src/strings/en-us.json
grhallenbeck, line 1: Seems like maybe an unintentional 4->2 space indentation change here?

GGProGaming: Seems to be so.

src/components/itemContextMenu.js
```diff
@@ -516,6 +534,22 @@ function executeCommand(item, id, options) {
...
+                    getResolveFunction(resolve, id, true)();
+                });
+                break;
+            case 'removefromnextup':
+                apiClient.ajax({
+                    url: apiClient.getUrl('Users/' + options.user.Id + '/ExcludeNextUp/' + item.Id, {}),
+                    type: 'POST'
+                }).then(function () {
+                    getResolveFunction(resolve, id, true)();
+                });
```
dmitrylyzo, line 551: IMO, API calls should be moved to utils (say `src/utils/jellyfin-apiclient/nextUp.ts`)

dmitrylyzo (commented):
Why so many commits?

thornbill:
This is blocked until the discussion for the _planned_ implementation in the meta repo is resolved. Making code changes before that happens is not the correct process.

BrennanWoodbury:
@thornbill We marked the meta discussion as resolved.

src/components/itemContextMenu.js
BrennanWoodbury: That was not the standard followed elsewhere in the code -- API calls are declared in their own functions in the same file, or called inline like we did, there doesn't seem to be a consistent standard of putting calls in utils that's been followed for past PR's. In the style guide it says to respect the surrounding code, and as the surrounding code doesn't define the API functions in the utils dir, nor does the style guide state to, I disagree. 

I don't disagree with this idea, I think it's a good idea, but I think that needs to be a part of a greater refactor where  all the API logic is moved into ..../src/utils/jellyfin-api, or wherever the code should live. 
Example: 
image

BrennanWoodbury:
> Why so many commits?

@dmitrylyzo Multiple contributors working on the same feature, we committed and rebased often. We're new to the project as a whole, and fairly unexperienced with git in a group setting, so if there's something else we should have done, please let us know, we're eager to learn! We did our best to follow the development guide, but it's entirely possible we missed something.

thornbill:
Closing per https://github.com/jellyfin/jellyfin/pull/10367#issuecomment-1874492343

PR #4856, opened 2023-10-08
Snap card sliders to card boundaries

### LINKED ISSUE #4677: [Issue]: 'More from Season' window does not snap to episodes
### Please describe your bug

When using the `More from Season <number>` window, using the `<` and `>` arrows moves the current view '1 window width' further. If this window view does not align to the episodes it will simply cuts off the episode (see included gif).

I think it would be better that the view snaps to the start of the episode instead.

Attempt at communicating via ASCII art:

```
// Currently
[1][2][3][4]    <episodes>
|------|        <position 1>     
       |------| <position 2> // Now 'starts' after position 1, but cuts off episode 3 awkwardly
```
```
// Better (I think/would argue) episode snapping behaviour
[1][2][3][4]    <episodes>
|------|        <snap 1>     
      |------|  <snap 2> // Now 'starts' at episode 3 -> Everything is nicely readable
```

### Jellyfin Version

Other

### if other:

10.8.10

### Environment

```markdown
- OS: Ubuntu 20.04.6 LTS
- Linux Kernel: 5.4.0-149-generic
- Virtualization:Docker
- Clients: Browser
- Browser: Firefox 113.0.2 (64-bit
- FFmpeg Version: 5.1.3-Jellyfin
- Playback Method: Not Applicable
- Hardware Acceleration: Not Applicable
- GPU Model: Not Applicable
- Plugins: None
- Reverse Proxy: nginx
- Base URL: jellyfin.<domain>.com
- Networking: Bridge/NAT
- Storage: Local
- Monitor resolution: 2560 x 1440
(It would be super nice to have a 'export github issue environment' button to fill (most) of these)
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

firefox_wLEdjVJ0YG

### Code of Conduct

grhallenbeck:
**Changes**

Fixes the scroller implementation to take into account card sizes when calculating the next or previous scroll positions. Previosly the scrollers would shift by a full page-width, which caused some cards to be cut off both before and after the page change and never fully shown. This snaps the pages to the card edges for both LTR and RTL views, and implements the fix in both legacy and React.

**Issues**

Fixes #4677


src/utils/scroller.ts
```diff
@@ -0,0 +1,101 @@
...
+    // Rounding down here will give us the first item index which is fully visible. We want the first partially visible
+    // index so we'll subtract one.
+    const firstVisibleIndex = Math.max(Math.floor(currentScrollPos / itemWidth) - 1, 0);
+    // Rounding up will give us the last index which is at least partially visible (overflows at container end).
+    const lastVisibleIndex = Math.floor((currentScrollPos + scrollerWidth) / itemWidth);
+
+    return [firstVisibleIndex, lastVisibleIndex];
+}
+
+function scrollToNextVisibleWindow(scroller: ScrollerFactory, items: HTMLElement[], scrollState: ScrollState) {
```
thornbill, line 49: There is a decent amount of repetition between this and `scrollToPreviousVisibleWindow`. I think if we pass the scroll direction as a parameter we could handle both in a single function.

src/utils/scroller.ts
thornbill, line 1: This file might be better alongside the scrollbutton elements... I don't imagine anything else will be using this.

thornbill (changes requested):

src/utils/scroller.ts
grhallenbeck: Good call! The logic got simpler as I refined it, but I hadn't quite taken it to the point of consolidating the functions.

src/utils/scroller.ts
```diff
@@ -0,0 +1,101 @@
+import ScrollerFactory from 'libraries/scroller';
+import globalize from 'scripts/globalize';
+
+export enum ScrollDirection {
+  RIGHT = 'right',
+  LEFT = 'left',
+}
+
+interface ScrollState {
+  scrollPos: number;
```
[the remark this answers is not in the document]
grhallenbeck: The `scrollPos` naming comes out of the `emby-scroller` legacy scroller implementation and the `Scroller` React component, which pass the scroll state in here. Combing through and renaming the state prop across both scroller implementations seemed like a bit too much refactor scope creep to add into these changes.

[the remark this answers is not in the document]
thornbill: No worries, this is fine!

thornbill (approved):

PR #4860, opened 2023-10-08
Fix language/subtitle switcher when using gamepad

### LINKED ISSUE #3755: Unable to change language or subs from movie menu using gamepad
**Describe The Bug**
Using Gamepad on Chromium 100, with TV layout enabled, I'm unable to interact with drop-down list to select language or subtitles. The field is highlighted but if I press A on the gamepad nothing happens. All other buttons (Play, Trailer, Back, etc) works fine.

**Steps To Reproduce**

1. Enable TV layout in jellyfin web
2. Enable gamepad support
3. Using gamepad, open a movie menu with more than one language and/or subtitles

**Expected Behavior**
I should be able to select audio track or subtitles

 - Platform: Windows 10
 - Browser: Chromium
 - Jellyfin Version: 10.8.1

cbe:
**Changes**
Raising the event code `13` (Enter) should be a lot more stable since `0` is just the default and not assigned to any actual key [1]. This keycode also has been standardized early enough to change it at this crucial part of gamepad handling in my opinion as all browsers should support
this.

[1] https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

**Issues**
Fixes #3755


crobibero:
I can confirm that I am able to select Audio track and Subtitle track using this PR's cloudflare pages, but not current unstable 👍

thornbill (approved):

dmitrylyzo:
Maybe backport to 10.8?

PR #4863, opened 2023-10-09
Link documentation in fallback fonts helptext

felix920506:
This PR intends to document where fallback fonts are used in Jellyfin within the web client.

related issues:
[jellyfin/#5008](https://github.com/jellyfin/jellyfin/issues/5008)
[jellyfin/#9273](https://github.com/jellyfin/jellyfin/issues/9273)


thornbill (commented):
We should really avoid having html in translations. I think the best option would be to have the documentation link be separate from the help text, so the html for the link can be part of the code. Maybe just "Read the documentation" or "Read more"?

felix920506:
I have moved the link into code. I will open a PR on web as well since there are some things that aren't there but was in this string before.

src/strings/en-us.json
```diff
@@ -1624,7 +1624,7 @@
     "HeaderSelectFallbackFontPath": "Select Fallback Font Folder Path",
     "HeaderSelectFallbackFontPathHelp": "Browse or enter the path of the fallback font folder to use for rendering ASS/SSA subtitles.",
     "LabelFallbackFontPath": "Fallback font folder path",
-    "LabelFallbackFontPathHelp": "Specify a path containing fallback fonts for rendering ASS/SSA subtitles. The maximum allowed total font size is 20 MB. Lightweight and web-friendly font formats such as woff2 are recommended.",
+    "LabelFallbackFontPathHelp": "These fonts are used by some clients to render subtitles. Please refer to the documentation for more info.",
```
thornbill, line 1627:
```suggestion
    "LabelFallbackFontPathHelp": "These fonts are used by some clients to render subtitles when transcoding. Please refer to the documentation for more information.",
```

felix920506: Your suggested change is inaccurate. These fonts aren't used if the server is burning in the subtitles. They are only currently used by the web client to render subtitles client side.

thornbill: :exploding_head: well that makes no sense for a setting in the encoding settings... no wonder people find it confusing

felix920506: I agree that this setting and its description make absolutely no sense, but this is how it is currently and should be documented.

thornbill (approved):

PR #4871, opened 2023-10-10
Fix libraries maybe not being displayed in TV layout

### LINKED ISSUE #4450: [Issue]: No way to change libraries in "TV" mode
### Please describe your bug

When using the "TV" mode theme, there is no way to go to the different libraries (Music, TV Shows, Music, etc) unless you manually add them to the Home screen.

In my case, I had removed them as they were entirely redundant when using my PC (they're in the menu, why would I have them here too).

When transitioning to my TV, the menu doesn't appear, and so there is no way to go to a library.

The only way to use Jellyfin in this setup is by using another device (phone, computer) and using the remote play option, through which the TV Jellyfin instance will navigate and play content.

To be clear, this is the Menu button I am referencing.
image

In TV mode, this is all that is visible.

image

### Jellyfin Version

10.8.0

### if other:

_No response_

### Environment

```markdown
- OS: PS4 (latest as of writing)
- Virtualization: Official Docker image
- Clients:
- Browser: PS4 webkit
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

As an aside, the menu _really_ should just always be visible (Plex does this and it works fine).
Without this, the first thing I have to do is explain to people that the 'funny dashes in the corner' is the key to using the application.

It seems there is a real lack of cohesion in the UI design.

thornbill:
**Changes**
Ensures that libraries are visible on the home screen when using the TV layout since the hamburger menu is unavailable

**Issues**
Fixes #4450

### DIFF
--- a/src/components/homesections/homesections.js
+++ b/src/components/homesections/homesections.js
@@ -1,3 +1,4 @@
+import layoutManager from 'components/layoutManager';
 import globalize from 'scripts/globalize';
 import { DEFAULT_SECTIONS, HomeSectionType } from 'types/homeSectionType';
 import Dashboard from 'utils/dashboard';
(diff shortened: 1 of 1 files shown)


dmitrylyzo (approved):
A long time ago, I tried enabling the menu drawer on TV. It worked, iirc, but exiting from it (navigate to the right) was glitchy.

PR #4879, opened 2023-10-14
Hide currently selected episode in 'More From Season' section

jmqm:
**Changes**
Filter out currently selected episode in 'More From Season' section that's visible in an episode's detailed view.

Before
image

After
image

### DIFF
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -1095,7 +1095,7 @@ function renderMoreFromSeason(view, item, apiClient) {
             section.classList.remove('hide');
             section.querySelector('h2').innerText = globalize.translate('MoreFromValue', item.SeasonName);
             const itemsContainer = section.querySelector('.itemsContainer');
-            cardBuilder.buildCards(result.Items, {
+            cardBuilder.buildCards(result.Items.filter((i) => i.Id !== item.Id), {
                 parentContainer: section,
                 itemsContainer: itemsContainer,
                 shape: 'autooverflow',


src/controllers/itemDetails/index.js
```diff
@@ -1095,7 +1095,7 @@ function renderMoreFromSeason(view, item, apiClient) {
             section.classList.remove('hide');
             section.querySelector('h2').innerText = globalize.translate('MoreFromValue', item.SeasonName);
             const itemsContainer = section.querySelector('.itemsContainer');
-            cardBuilder.buildCards(result.Items, {
+            cardBuilder.buildCards(result.Items.filter((i) => i.Id !== item.Id), {
```
thornbill, line 1098: Let's move the filtering to be at the start of the function in the `.then()`.

```ts
const items = result.Items.filter(i => i.Id !== item.Id);
```

The if statement checking the length will need updated also.

thornbill:
@jmqm did you mean to close this?

jmqm:
Yes. Implemented your suggestion but I noticed a problem.

Originally, if I'm on episode 12/12 for example, `More From Season` scrolls to the episode. When I removed the currently-selected episode from `More From Season`, it stopped doing that.

I was debugging it for a while and eventually I didn't want to continue further. I was working on an additional feature, to blur unwatched episodes, but it was too hectic and causing problems on other areas. I lost interest in both, the codebase is too unorganized for me, and didn't want to continue.

Apologies for the time wasted.

PR #4884, opened 2023-10-16
Use icon url from client capabilities

thornbill:
**Changes**
* Uses the `IconUrl` from client capabilities as the fallback for the device icon in the dashboard if specified. This allows clients to specify their own icon that is **not** bundled in web.
* Converts image helper to typescript and moves it to `src/utils`.

**Issues**
N/A

Related to #4725

### DIFF
--- a/src/components/cardbuilder/cardBuilder.js
+++ b/src/components/cardbuilder/cardBuilder.js
@@ -11,8 +11,8 @@ import browser from 'scripts/browser';
 import datetime from 'scripts/datetime';
 import dom from 'scripts/dom';
 import globalize from 'scripts/globalize';
-import imageHelper from 'scripts/imagehelper';
 import { getBackdropShape, getPortraitShape, getSquareShape } from 'utils/card';
+import imageHelper from 'utils/image';
 import { randomInt } from 'utils/number';
 
 import focusManager from '../focusManager';
--- a/src/components/homesections/sections/libraryButtons.ts
+++ b/src/components/homesections/sections/libraryButtons.ts
@@ -4,7 +4,7 @@ import escapeHtml from 'escape-html';
 import imageLoader from 'components/images/imageLoader';
 import { appRouter } from 'components/router/appRouter';
 import globalize from 'scripts/globalize';
-import imageHelper from 'scripts/imagehelper';
+import imageHelper from 'utils/image';
 
 function getLibraryButtonsHtml(items: BaseItemDto[]) {
     let html = '';
--- a/src/controllers/dashboard/dashboard.js
+++ b/src/controllers/dashboard/dashboard.js
@@ -12,7 +12,7 @@ import playMethodHelper from '../../components/playback/playmethodhelper';
 import cardBuilder from '../../components/cardbuilder/cardBuilder';
 import imageLoader from '../../components/images/imageLoader';
 import ActivityLog from '../../components/activitylog';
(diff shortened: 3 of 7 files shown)


nielsvanvelzen:
We should probably proxy the custom client icon via the server to avoid any issues with CORS/CSP.

thornbill:
Maybe... but we have no way of doing that currently and since the last method of doing so was removed as a security vulnerability it will probably take some time to design and implement a solution for that.

CORS would really be dependent on the app developer properly configuring their server hosting the image.

Personally I find setting a restrictive CSP for an application like Jellyfin pretty silly. It already breaks loading images from metadata providers and who knows what else. 🤷‍♂️

ferferga:
Isn't this a hole for XSS attacks?

nielsvanvelzen:
> Maybe... but we have no way of doing that currently and since the last method of doing so was removed as a security vulnerability it will probably take some time to design and implement a solution for that.
> 
> CORS would really be dependent on the app developer properly configuring their server hosting the image.
> 
> Personally I find setting a restrictive CSP for an application like Jellyfin pretty silly. It already breaks loading images from metadata providers and who knows what else. 🤷‍♂️

I don't think doing it via the server should be too hard, we can probably add an endpoint with the device id as input and the image as output. But you're right that it will pretty much only affect installs with CSP which our documentation also has disabled in the sample configs.

> Isn't this a hole for XSS attacks?

Looks like some usages of the `getDeviceIcon` function inject it directly in HTML building strings so that needs fixing.

ferferga:
Why are icons so necessary though?
I think we're better off not shipping any icon (changing the way we present clients of course).

I mentioned this other times, but I even think we shouldn't do any browser parsing at all (user agents are being deprecated and they can be easily mocked, plus we don't handle cases where data is imported from another browser) and just create a random name, like GitHub does with Codespaces with something like https://www.npmjs.com/package/unique-names-generator

Adding an endpoint just for this is overkill imo.

nielsvanvelzen (approved):

thornbill:
That was a great catch for the XSS possibilities. I updated it so it tries parsing as a URL and falls back to the default if it fails, which should be safe... (If you are connecting with a malicious client you probably have other issues too. :sweat_smile:)

---

@nielsvanvelzen:

> I don't think doing it via the server should be too hard, we can probably add an endpoint with the device id as input and the image as output. But you're right that it will pretty much only affect installs with CSP which our documentation also has disabled in the sample configs.

Now that you have described it, an endpoint for this would be way easier than the metadata image issue. I wouldn't be opposed to going that route either. Could be a follow up enhancement...

---

@ferferga:

> Why are icons so necessary though? I think we're better off not shipping any icon (changing the way we present clients of course).

I don't know, I think they add a lot to the dashboard. I think it's much more challenging to quickly identify sessions without that visual assistance.  :man_shrugging: 

> I mentioned this other times, but I even think we shouldn't do any browser parsing at all (user agents are being deprecated and they can be easily mocked, plus we don't handle cases where data is imported from another browser) and just create a random name, like GitHub does with Codespaces with something like https://www.npmjs.com/package/unique-names-generator

Ideally I think you are right, but browsers can't seem to get their :hankey:  together enough to reliably generate playback profiles without exceptions based on the parsing. We **definitely** should be using an upstream library instead of rolling our own solution for browser detection. This is kind of outside the scope of this PR though.

PR #4889, opened 2023-10-18
Change search page to support query param

bradbeattie:
**Changes**

* Augmenting `search.tsx` to support a linked search (e.g. `/#/search.html?q=foobar`).
  *  Changes to the search input are dynamically represented in the URL
  *  Changes to the URL are dynamically represented in the search input
* Demonstrating functionality of this by additionally altering `itemTags` to link to predefined searches.

**Issues**

https://features.jellyfin.org/?query=Search%20parameters%20in%20URL (ironically, features.jellyfin.org supports linked searches!)

### DIFF
--- a/src/apps/stable/routes/search.tsx
+++ b/src/apps/stable/routes/search.tsx
@@ -19,11 +19,12 @@ function usePrevious(value: string) {
 
 const Search: FunctionComponent = () => {
     const [ searchParams ] = useSearchParams();
-    const [ query, setQuery ] = useState<string>(searchParams.get('query') || '');
+    const urlQuery = searchParams.get('query') || '';
+    const [ query, setQuery ] = useState<string>(urlQuery);
     const prevQuery = usePrevious(query);
 
-    if (query == prevQuery && searchParams.get('query') != query) {
-        setQuery(searchParams.get('query') || '');
+    if (query == prevQuery && urlQuery != query) {
+        setQuery(urlQuery);
     }
 
     useEffect(() => {
--- a/src/apps/stable/routes/search.tsx
+++ b/src/apps/stable/routes/search.tsx
@@ -11,6 +11,7 @@ import globalize from '../../../scripts/globalize';
 const Search: FunctionComponent = () => {
     const [ query, setQuery ] = useState<string>();
     const [ searchParams ] = useSearchParams();
+    if (!query && searchParams.get('q')) setQuery(searchParams.get('q') || '');
 
     return (
         <Page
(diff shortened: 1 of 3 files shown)


nielsvanvelzen:
Some quick test results:

1) Doesn't update the URL when using the search bar - it should do that so the link can be shared easily (use history replace function to accomplish this, there is probably a correct way to do it with React)
2) Opening the search page with ?query=a and then manually changing the url to ?query=b doesn't update the page

You mentioned both of these issues already but they should be addressed in this PR.

bradbeattie:
1. Now updates the URL when using the search bar
2. Opening the search page with `?query=a` and then manually changing the url to `?query=b` now updates the page

src/apps/stable/routes/search.tsx
```diff
@@ -7,18 +7,42 @@ import SearchResults from '../../../components/search/SearchResults';
...
+        ref.current = value;
+    });
+    return ref.current;
+}
 
 const Search: FunctionComponent = () => {
-    const [ query, setQuery ] = useState<string>();
     const [ searchParams ] = useSearchParams();
+    const [ query, setQuery ] = useState<string>(searchParams.get('query') || '');
+    const prevQuery = usePrevious(query);
```
bradbeattie, line 23: The ability to know what `query` was previously set to allows us to detect which direction the change came from (a change to the URL or a change to the input field).

bradbeattie:
Hrmm... Running into an odd unexpected error when loading the page with no state. Digging...

bradbeattie:
Lemme know if there's anything else you need from me to push this forward.

nielsvanvelzen:
Testing was successful but unfortunately I can't help with the code review as I have 0 React knowledge.

src/controllers/itemDetails/index.js
```diff
@@ -1261,11 +1261,11 @@ function renderTags(page, item) {
     }
 
     for (let i = 0, length = tags.length; i < length; i++) {
-        tagElements.push(tags[i]);
+        tagElements.push('<a href="#/search.html?query=' + encodeURIComponent(tags[i]) + '" class="button-link emby-button" is="emby-linkbutton">' + tags[i] + '</a>');
     }
 
     if (tagElements.length) {
-        itemTags.innerText = globalize.translate('TagsValue', tagElements.join(', '));
+        itemTags.innerHTML = globalize.translate('TagsValue', tagElements.join(', '));
```
Fube, line 1268: nitpick: This could open the path for some XSS, though I do not believe it is a realistic threat as the malicious payload would need to be in the tag

Fube (commented):
possible-bug: Having something in the search input and clicking the search icon in the navigation will cause the input & the URL reset to nothing.

I am unsure if this is the expected behaviour or not.

https://github.com/jellyfin/jellyfin-web/assets/8064235/e0e66195-2824-4681-a389-8e54928e4db6

src/controllers/itemDetails/index.js
thornbill: This should be fine since the tag is url encoded.

thornbill:
Included in #5203

PR #4895, opened 2023-10-19
Use LTR for default subtitle direction

### LINKED ISSUE #4179: Right-to-left languages subtitles having punctuation in the wrong order.
**Describe The Bug**
Right-to-left languages subtitles having punctuation in the wrong order.

**Screenshots**

The problem is due to current web browsers rendering the every VTT file in the same way and ingonring right to left.

Screenshot 2022-11-19 205059

The subtitles should be renders like this (with the punctuation to the left side); 

> וארבעה ילדים אשר...

thornbill:
**Changes**
It seems that subtitle files have traditionally written RTL subs in LTR format to account for a lack of proper RTL support. Browser subtitle rendering will try to intelligently guess the proper text direction based on the first character in the text to prevent this behavior we need to prepend the LTR mark character to force the direction. I did find some precedence for this where Kodi have had to do something similar in the past. :man_shrugging: 

**Issues**
Fixes #4179

### DIFF
--- a/src/plugins/htmlVideoPlayer/plugin.js
+++ b/src/plugins/htmlVideoPlayer/plugin.js
@@ -143,7 +143,9 @@ function normalizeTrackEventText(text, useHtml) {
     const result = text
         .replace(/\\N/gi, '\n') // Correct newline characters
         .replace(/\r/gi, '') // Remove carriage return characters
-        .replace(/{\\.*?}/gi, ''); // Remove ass/ssa tags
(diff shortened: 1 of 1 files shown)


Ignio-dev:
Hey, just checked, works great!

pypp:
It works great, just tried it.

nielsvanvelzen (approved):

thornbill:
No the Android TV app does not use jellyfin-web in any way.

PR #4896, opened 2023-10-20
Continue watch progress for Series and Season play buttons

beef331:
**Changes**

Fetches the information for the episodes and resumes the content from where left off.

**Issues**

### DIFF
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -1852,11 +1852,14 @@ class PlaybackManager {
             } else if (firstItem.Type === 'Series' || firstItem.Type === 'Season') {
                 const apiClient = ServerConnections.getApiClient(firstItem.ServerId);
 
-                promise = apiClient.getEpisodes(firstItem.SeriesId || firstItem.Id, {
+                promise = getItemsForPlayback(firstItem.ServerId, {
                     IsVirtualUnaired: false,
                     IsMissing: false,
                     UserId: apiClient.getCurrentUserId(),
-                    Fields: 'Chapters'
+                    Fields: 'Chapters',
+                    Recursive: true,
+                    MediaTypes: 'Video',
+                    ParentId: firstItem.SeriesId ?? firstItem.Id
                 }).then(function (episodesResult) {
                     const originalResults = episodesResult.Items;
                     const isSeries = firstItem.Type === 'Series';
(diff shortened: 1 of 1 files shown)


src/components/playback/playbackmanager.js
```diff
@@ -1895,6 +1898,12 @@ class PlaybackManager {
                         }
                     }
 
+                    for (let i = 0; i < episodesResult.Items.length; i++) {
+                        if (i > 0) {
+                            episodesResult.Items[i].UserData.PlaybackPositionTicks = 0;
+                        }
+                    }
```
thornbill, line 1905: So this is ensuring that all episodes in the queue start at the beginning?

src/components/playback/playbackmanager.js
```diff
@@ -2443,7 +2452,7 @@ class PlaybackManager {
         }
 
         function playAfterBitrateDetect(maxBitrate, item, playOptions, onPlaybackStartedFn, prevSource) {
-            const startPosition = playOptions.startPositionTicks;
+            const startPosition = item.UserData.PlaybackPositionTicks && playOptions.isFirstItem ? item.UserData.PlaybackPositionTicks : playOptions.startPositionTicks;
```
dmitrylyzo, line 2455: Doesn't this prevent us from restarting playback?

I haven't thought much about this, but we could use `playOptions.startPositionTicks` to control resume/restart/seek.
If `playOptions.startPositionTicks` isn't set, we use `item.UserData.PlaybackPositionTicks` - resuming.
Otherwise we use `playOptions.startPositionTicks` - restarting or using a time link.

src/components/playback/playbackmanager.js
beef331: That is the idea, but in testing it seems that hitting the season/series play button does not keep this information, at the end of an episode it seems to just load the next one. In the other procedure `isFirstItem` seems to always be true.

src/components/playback/playbackmanager.js
beef331: Yea it does seem that way, I'm mostly lacking information on where to pipe the information to reason what to do.

thornbill:
Should this be converted to a draft since it seems like it is still a work in progress?

beef331:
I'll just close it for now, requires too many changes and much understanding of what needs to be changed.

PR #4897, opened 2023-10-20
Fix wrong cursor & right click abilities in library

orrnithogalum:
**Changes**
- Users were able to hover and right click on a card while being between two cards. 

_Before:_
Screencast from 2023-10-20 15-21-29

_After:_
Screencast from 2023-10-20 15-21-13

- Also fixed my line in CONTRIBUTORS.md cause I managed to fuck that up somehow

**Issues**
No issues were created related to this change

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -70,6 +70,7 @@
  - [tehciolo](https://github.com/tehciolo)
  - [scampower3](https://github.com/scampower3)
  - [LittleBigOwI] (https://github.com/LittleBigOwI/)
+ - [Nate G](https://github.com/GGProGaming)
 
 # Emby Contributors
 
--- a/src/components/search/SearchResults.tsx
+++ b/src/components/search/SearchResults.tsx
@@ -6,6 +6,7 @@ import React, { FunctionComponent, useCallback, useEffect, useState } from 'reac
 import globalize from '../../scripts/globalize';
 import ServerConnections from '../ServerConnections';
 import SearchResultsRow from './SearchResultsRow';
+import Loading from '../loading/LoadingComponent';
 
 type SearchResultsProps = {
     serverId?: string;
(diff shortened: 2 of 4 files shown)


src/components/cardbuilder/card.scss
```diff
@@ -834,6 +833,7 @@ button::-moz-focus-inner {
 }
 
 .card-hoverable:hover .cardOverlayContainer {
+    cursor: pointer;
     opacity: 1;
 }
```
dmitrylyzo, line 838: There may be no overlay: https://github.com/jellyfin/jellyfin-web/blob/c0fb9eca82bee9bda8f117d1e3ef38b21ad27721/src/components/cardbuilder/cardBuilder.js#L1206-L1208

Could you check this in TV layout?

src/components/cardbuilder/cardBuilder.js
```diff
@@ -1207,7 +1204,28 @@ function buildCard(index, item, apiClient, options) {
...
+            + positionTicksData
+            + collectionIdData
+            + playlistIdData
+            + contextData
+            + parentIdData
+            + startDate
+            + endDate
+        + ' data-prefix="' + escapeHtml(prefix) + '"';
+
+    cardImageContainerOpen = `<div ${itemData} class="${cardBoxClass}">${cardImageContainerOpen}`;
```
dmitrylyzo, line 1226: Item data has been moved from `card` to `cardBox`. I wonder if it is still working.

src/components/cardbuilder/card.scss
LittleBigOwI: Hi, sorry still new to this so this might sound stupid - What would be the best way to test the TV layout?

src/components/cardbuilder/cardBuilder.js
LittleBigOwI: Indeed the TV layout isn't working as of now. I have marked this PR as a Draft.

I did fix some issues but can't get right clicking to work for now.

src/components/cardbuilder/card.scss
LittleBigOwI: This seems to work fine - Unless I am missing something.

PR #4898, opened 2023-10-20
Handle json parse error when testing urls

thornbill:
**Changes**
Fixes an issue where the app will crash if a tested url returns a successful response that is not valid json. This was particularly noticeable on the cloudflare pages deployments where `/System/Info/Public` is returning the `index.html` file by default.

**Issues**
N/A


nielsvanvelzen (approved):
Appears to be working again! 🎉

PR #4902, opened 2023-10-21
Remove unused prop type in AccessScheduleList

Davoud-Ghargozlou:
Hi,
I fixed [this](https://sonarcloud.io/project/issues?open=AYr_7x6Y4nroI6nze3Zu&id=jellyfin_jellyfin-web) issue in SonarCloud.


dmitrylyzo:
In fact, it appears to be in use - see build error.

thornbill:
Actually the component never uses the prop, so if the assignment is removed this should be good to go.

https://github.com/jellyfin/jellyfin-web/blob/master/src/apps/dashboard/routes/users/parentalcontrol.tsx#L417

thornbill:
Fixed in #5154

PR #4907, opened 2023-10-23
Migrate datetime.js to TypeScript and add unit tests

Pithaya:
**Changes**
- Moves the datetime script file to utils and migrates it to typescript
- Adds unit tests and updates the vite configuration
- Fixes timezone calculation in `parseISO8601Date`
- Fixes [cognitive complexity issue](https://sonarcloud.io/project/issues?open=AXepJq-uYTlMEX6vFILu&id=jellyfin_jellyfin-web) in `getDisplayTime`

**Issues**
/


vitest.setup.ts
```diff
@@ -0,0 +1,4 @@
+export const setup = () => {
+    // UTC−09:00
+    process.env.TZ = 'Etc/GMT+9';
```
thornbill, line 3: I think using Eastern Standard Time might make more sense for tests considering the project's eastern Canadian roots? 🍁

src/utils/datetime.ts
```diff
@@ -0,0 +1,370 @@
...
+ * @param options - An optional object containing formatting options.
+ * @returns The time part of the date as a string in the current locale.
+ * @throws {Error} If the value parameter is null.
+ */
+export function toLocaleTimeString(date: DateOrNull, options?: Intl.DateTimeFormatOptions): string {
+    if (!date) {
+        throw new Error('date cannot be null');
+    }
+
+    options = options || {};
```
thornbill, line 246: This could be a default parameter value instead.

src/utils/datetime.ts
```diff
@@ -0,0 +1,370 @@
...
+    return date.toLocaleTimeString();
+}
+
+/**
+ * Checks if the given value is a string.
+ * @param value - The value to check.
+ * @returns True if the value is a string, false otherwise.
+ */
+function isString(value: Date | string): value is string {
+    return (typeof value).toString().toLowerCase() === 'string';
```
thornbill, line 265: Why would we not just use `typeof value === 'string'`? This seems like a legacy code smell.

thornbill (changes requested):

PR #4912, opened 2023-10-24
Fix imports

dmitrylyzo:
**Changes**
- Fix imports in RemoteControl.
- Fix imports in Playback/Queue controller.
_According to the template._

**Issues**
In webOS 1.2, `master` crashes because `emby-slider` is not loaded.
10.8 doesn't crash (probably due to a poor tree-shaking), but I don't think we should leave a potential bug in the build.


src/components/remotecontrol/remotecontrol.js
```diff
@@ -13,9 +13,11 @@ import * as userSettings from '../../scripts/settings/userSettings';
 import cardBuilder from '../cardbuilder/cardBuilder';
 import itemContextMenu from '../itemContextMenu';
 import '../cardbuilder/card.scss';
+import '../../elements/emby-button/emby-button';
 import '../../elements/emby-itemscontainer/emby-itemscontainer';
```
dmitrylyzo, line 17: `emby-itemscontainer` can be removed (no direct use in `remotecontrol.js`), but I didn't want to risk :sweat_smile:

thornbill (approved):

PR #4914, opened 2023-10-24
For webOS, block DOVI for non mp4 containers

viktory36:
This change will block non-mp4 containers from direct playing DOVI media on webOS browsers with DOVI support. This, paired with the "Prefer fMP4-HLS container" client playback setting will enable DOVI Profile 5 playback on webOS.

For DOVI Profile 8 (dovi with hdr10/hlg fallback), we will require some changes in the server side (jellyfin/jellyfin#10468), or we could even just remove HDR10/HLG support as well from non-mp4 containers, that should kickoff the remuxing of that media to mp4 and play in DoVi.
 
**Changes**
Where browser is webOS and has DOVI support, push a CodecProfile that says containers other than mp4 cannot direct play DOVI.

**Issues**
jellyfin/jellyfin-webos#159


src/scripts/browserDeviceProfile.js
```diff
@@ -1125,6 +1125,23 @@ export default function (options) {
...
 
+    if (browser.web0s) {
+        if (supportsDolbyVision(options)) {
+            // Disallow direct playing of DOVI media in containers not mp4.
+            // This paired with the "Prefer fMP4-HLS Container" client playback setting enables DOVI playback on webOS.
+            profile.CodecProfiles.push({
+                Type: 'Video',
+                Container: "-mp4",
+                Codec: 'hevc',
+                Conditions: hevcCodecProfileConditions.map(condition =>
```
dmitrylyzo, line 1136: Only necessary conditions can be left here - server checks all codec profiles that match codec+container.

I see that you haven't added `'mp4': 'SDR|HLG|HDR10|DOVI'`. But this is probably unnecessary:
- There will be `'-mp4': 'SDR|HLG|HDR10'` (matches non-MP4) and `'': 'SDR|HLG|HDR10|DOVI'` (matches any container).
- The first condition (`-mp4`) would fail for nonMP4+DoVi and the whole direct play test would fail (_as we want_).

Note: `'mp4': 'SDR|HLG|HDR10|DOVI'` + `'': 'SDR|HLG|HDR10'` won't work - specializations can only use a subset of the general profile.

src/scripts/browserDeviceProfile.js
```diff
@@ -918,22 +918,22 @@ export default function (options) {
     let vp9VideoRangeTypes = 'SDR';
     let av1VideoRangeTypes = 'SDR';
 
+    if (supportsDolbyVision(options)) {
+        hevcVideoRangeTypes += '|DOVI';
+    }
+
```
dmitrylyzo, line 924: I prefer not to make unnecessary changes. Please move it back to its original location (your regex won't break).

src/scripts/browserDeviceProfile.js
```diff
@@ -918,22 +918,22 @@ export default function (options) {
...
     if (supportsHdr10(options)) {
-        hevcVideoRangeTypes += '|HDR10';
+        hevcVideoRangeTypes += '|HDR10|DOVI_with_HDR10_Fallback';
```
dmitrylyzo, line 926: If we can just stream DoVi (8?) to an HDR10-only client and it will automatically fallback to HDR10 (client-side), then we don't need a special value.

nyanmisaka: Correct. https://github.com/jellyfin/jellyfin/issues/10468#issuecomment-1778532412

viktory36: I added some more context regarding this addition on the above thread, [here](https://github.com/jellyfin/jellyfin/issues/10468#issuecomment-1780029285). I don't think DoVi profile 8 playback can be enabled on webOS without changes in the server (or ffprobe), so added my comments there.

viktory36:
I messed up stuff with rebase. I'll close this PR and open a clean one.

PR #4916, opened 2023-10-25
Adding DOVI support on webOS

viktory36:
-- resubmitting fresh/clean PR --

This change will block non-mp4 containers from direct playing DOVI media on webOS browsers with DOVI support. This, paired with the "Prefer fMP4-HLS container" client playback setting will enable DOVI Profile 5 playback on webOS.

For DOVI Profile 8 (dovi with hdr10/hlg fallback), I think we will require some changes in the server side or in ffprobe. Please see [jellyfin/jellyfin#10468#issuecomment-1780029285](https://github.com/jellyfin/jellyfin/issues/10468#issuecomment-1780029285)
 
**Changes**
Where browser is webOS and has DOVI support, push a CodecProfile that says containers other than mp4 cannot direct play DOVI.

**Issues**
jellyfin/jellyfin-webos#159

--
I've incorporated the latest review comments from dmitrylyzo from the old PR in this one, and happy to know your thoughts here.
Should we wait for clarity on dovi-8 related issues before merging this?


src/scripts/browserDeviceProfile.js
```diff
@@ -1125,6 +1125,24 @@ export default function (options) {
...
+        // This paired with the "Prefer fMP4-HLS Container" client playback setting enables DOVI playback on webOS.
+        profile.CodecProfiles.push({
+            Type: 'Video',
+            Container: '-mp4',
+            Codec: 'hevc',
+            Conditions: [
+                {
+                    Condition: 'EqualsAny',
+                    Property: 'VideoRangeType',
+                    Value: 'SDR|HDR10|HLG',
```
dmitrylyzo, line 1139: Your regex was fine.
```suggestion
                    Value: hevcVideoRangeTypes.replace(/\|DOVI[^|]*/g, ''),
```

src/scripts/browserDeviceProfile.js
```diff
@@ -1125,6 +1125,24 @@ export default function (options) {
         Conditions: h264CodecProfileConditions
     });
 
+    if (browser.web0s && supportsDolbyVision(options)) {
```
dmitrylyzo, line 1128: At some point we probably need to make `supportsDolbyVision` (and others) return a list of containers and codecs in which DOVI is supported.
Boolean would mean "any container and codec".

src/scripts/browserDeviceProfile.js
viktory36: I actually like this version more, stemming from your comment _"Only necessary conditions can be left here "_ on the original PR for this block. It makes sense that for a codec profile that talks about what is not allowed, that codec profile will never be the one used for a playback, right?

I assumed this was the flow the server does-
1. see that a codecprofile says (not mp4 + dovi) = no go
2. do the transcode process
3. check codecprofile again to find the base [hevcCodecProfileConditions](https://github.com/viktory36/jellyfin-web/blob/317c8485ceb703e4654b121edc1a92f76b94231d/src/scripts/browserDeviceProfile.js#L1146-L1150) (that we always push) and all conditions satisfied, which is why we didn't require a codecprofile that mentioned `Container: 'mp4'`.

Unless I'm wrong, and the below conditions that are presently a part of the base hevcCodecProfileConditions also need to be a part of the `-mp4` codec profile?
`{
    "Type": "Video",
    "Codec": "hevc",
    "Conditions": [
        {
            "Condition": "NotEquals",
            "Property": "IsAnamorphic",
            "Value": "true",
            "IsRequired": false
        },
        {
            "Condition": "EqualsAny",
            "Property": "VideoProfile",
            "Value": "main|main 10",
            "IsRequired": false
        },
        {
            "Condition": "EqualsAny",
            "Property": "VideoRangeType",
            "Value": "SDR|HDR10|HLG|DOVI",
            "IsRequired": false
        },
        {
            "Condition": "LessThanEqual",
            "Property": "VideoLevel",
            "Value": "183",
            "IsRequired": false
        }
    ]
}`

During my tests on webOS, I found no difference between keeping all the conditions in the `-mp4` profile vs current version - both led to the expected behavior of starting a remux. I also found the `Container: 'mp4'` profile was not necessary.


Are you okay with staying on this version, on the account it's a lot simpler and in my tests gives the same results as before?

dmitrylyzo: OK, I got it - it will fail one of the tests. But it looks tricky.

> I assumed this was the flow the server does-

Step 2 goes last (if no direct profile is found):
HEVC in MP4: server tests only big one.
HEVC in MKV/TS: server tests both.
Then "do the transcode process".

src/scripts/browserDeviceProfile.js
viktory36: Hmm, I'm not able to fully grasp the implications of supportsDolbyVision working that way. We would push a codecprofile for each container that supportsDolbyVision returns as being okay, instead of the current logic where we are removing whatever is not supported?

src/scripts/browserDeviceProfile.js
viktory36: Ah, I see. 

There is one other issue in this block - for Dovi profile 8 support on webOS, if nothing changes in the server to specify a unique video range type for DOVI+HDR10 (currently these media are reported as just HDR10/HLG), we will need to remove even HDR10/HLG direct play support for non-mp4, on the off-chance that file actually contains dovi data which would need remuxing.
 ```suggestion
                    Value: 'SDR',
```
I'm happy to do that on my setup, but I don't know if we can speak for every webos+DOVI user in this.

Or is there another way I'm not thinking of?

File: src/scripts/browserDeviceProfile.js (hunk 2 above)
Comment (dmitrylyzo) [reply src/scripts/browserDeviceProfile.js:1128]: This way we can specify containers (and probably codecs) that support DOVI from the apps (we can check the TV model if it matters). And we could still probably use the trick you used.

For example:
```
supportsDolbyVision -> [] | true
append DOVI to the "base" codec profile:
    '': '...|DOVI|...'
```
```
supportsDolbyVision -> false | null | undefined
don't append DOVI to any codec profile
```
```
supportsHdr10 -> [] | true
append HDR10 to the "base" codec profile:
    '': '...|HDR10|...'
```
```
supportsDolbyVision -> ['mp4', 'mkv']
append DOVI to the "base" codec profile:
    '': '...|DOVI|...'
add codec profile:
    '-mp4,mkv': 'SDR|HLG|HDR10'
```
```
supportsHdr10 -> ['mp4', 'ts']
append HDR10 to the "base" codec profile:
    '': '...|HDR10|...'
add codec profile:
    '-mp4,ts': 'SDR|HLG|DOVI'
```

FYI, we can split the conditions into multiple codec profiles because, as I said, they will all be tested regardless of order.

At the same time, this approach looks complicated/tricky.

File: src/scripts/browserDeviceProfile.js (hunk 1 above)
Comment (dmitrylyzo) [reply src/scripts/browserDeviceProfile.js:1139]: I wouldn't fix this in clients - the media should be detected as DoVi8 (aka DOVI+HDR10) on the server side. AFAIK, media info has an extra field for a profile.

File: src/scripts/browserDeviceProfile.js (hunk 2 above)
Comment (viktory36) [reply src/scripts/browserDeviceProfile.js:1128]: I see. I'm thinking we could consider making this when another browser comes around with it's own set of conditions for a scenario similar to this, so in that case having a routine to build their specific profiles make sense. Are there any other unrelated issues or functions where this mode of function would have helped?

I'm reading posts from like 2years ago about how nvidia shield also has this problem, but I'm not sure where they are at with this now.

Comment (dmitrylyzo) [reply src/scripts/browserDeviceProfile.js:1128]: > I'm thinking we could consider making this when another browser comes around with it's own set of conditions for a scenario similar to this, so in that case having a routine to build their specific profiles make sense.

Agree. I didn't say that it should be done right now. Moreover, this idea wasn't discussed with team.

> Are there any other unrelated issues or functions where this mode of function would have helped?

I don't think so. The only purpose of this form is to circumvent inconsistent support.

**Comment by Schaka** [comment]
I decided to try this today - I have an android player and don't need it but I know some people on webos are struggling.

I can't confirm it's working for sure. What I can say is that when playing mkv with DV and HDR layer, it'll play the HDR straight from mkv.

I would expect this to play the DV layer as a preference always. Maybe if that's not covered in this PR, it should be.

**Comment by viktory36** [comment]
@Schaka, this PR enables playback of DV media that does not have HDR fallback (DV profile 5) by forcing a transmux, like you describe in the recently linked issue. For profile 8, things have to change on the server side (please see linked issue in OP).

@dmitrylyzo hey, I think this PR can be merged in it's current state. While it does not fully enable all dolby vision playback, it will resolve the green-purple issue which still helps a considerable chunk of users.

**Comment by Schaka** [comment]
Do you have a sample DV5 file I could use to test this?
If you guys still need confirmation, that is. On the I'm on the C1.

**Comment by viktory36** [comment]
There is one here https://kodi.wiki/view/Samples Dolby Vision Mystery Box (Profile 5)

**Comment by Schaka** [comment]
I tested this PR with the DV5 (in mkv container) and it works perfectly, just transmuxing as expected.

If I wanted to hackfix DV playback for profile 7 and 8, I'd have to remove HDR, HDR10 and HLG from mkv, so that it would force transmux to mp4 as well?

**Comment by dmitrylyzo** [comment]
> Am I correct in understanding that as of now, transmuxing P7 or P8 loses DV information and thus will only play back HDR/HLG/whatever fallback?

With jellyfin-ffmpeg5, I am able to remux to MP4 keeping DoVi8. So I suppose the DV info should remain when remuxing.

File: src/scripts/browserDeviceProfile.js [hunk 3]
```diff
@@ -1141,6 +1141,24 @@ export default function (options) {
...
+        // This paired with the "Prefer fMP4-HLS Container" client playback setting enables DOVI playback on webOS.
+        profile.CodecProfiles.push({
+            Type: 'Video',
+            Container: '-mp4',
+            Codec: 'hevc',
+            Conditions: [
+                {
+                    Condition: 'EqualsAny',
+                    Property: 'VideoRangeType',
+                    Value: 'SDR|HDR10|HLG',
```
dmitrylyzo, line 1155: In theory, this should work too (for DoVi without fallback). It would be nice if you could check (_I only did a rough check_).
```suggestion
                    Condition: 'NotEquals',
                    Property: 'VideoRangeType',
                    Value: 'DOVI',
```
With `NotEquals`, `Value` cannot be a list - we need a new `NotEqualsAny` condition.

FYI, there is an `ApplyConditions` field to activate conditions. I think we should read Dolby Vision with HDR10 fallback as `DOVI,HDR10` when checking the conditions and check both (_requires server-side changes_). So if we need to disable any fallback (_user settings_), we can add (_not tested_):
```js
{
    ApplyConditions: [
        {
            Condition: 'EqualsAny',
            Property: 'VideoRangeType',
            Value: 'DOVI', // with server-side changes, 'DOVI,HDR10' will match
            IsRequired: false
        }
    ],
    Conditions: [
        {
            Condition: 'Equals',
            Property: 'VideoRangeType',
            Value: '', // disable any non-mp4 video with DoVi
            IsRequired: false
        }
    ]
}
```

dmitrylyzo (commented):
I can't really test this (no LG TV), but I agree that for DoVi without fallback, it should work.

src/scripts/browserDeviceProfile.js
viktory36: Setting "NotEquals" to DOVI indeed works to kickoff transmux for DOVI content, identical to how the current lines behave. With this and "Enable fMP4-HLS", I am able to view DV Profile 5 media.

It appears the server-side is aligning to implement this fix by introducing the new VideoRangeTypes "DOVIWithHDR10" and "DOVIWithHLG" for fallback-able DV. With that, I thought I'd just update lines 930 and 936 to include them by default for all HDR/HLG supporting clients, and leave this block as-is.

But looking into ApplyConditions, I don't understand how it's to be used. For instance, doing this works (both P8 and P5 remuxed and playback in DOVI, and HDR direct played as expected -- with the latest patches from the server's PR in place)
 ``` 
        // Disallow direct playing of DOVI media in containers not mp4.
        // This paired with the "Prefer fMP4-HLS Container" client playback setting enables DOVI playback on webOS.
        profile.CodecProfiles.push({

            ApplyConditions: [
                {
                    Condition: 'EqualsAny',
                    Property: 'VideoRangeType',
                    Value: 'DOVI|DOVIWithHDR10|DOVIWithHLG', // with server-side changes, 'DOVI,HDR10' will match
                    IsRequired: false
                }
            ],
            Conditions: [
                {
                    Condition: 'Equals',
                    Property: 'VideoRangeType',
                    Value: '', // disable any non-mp4 video with DoVi
                    IsRequired: false
                }
            ]
        });
    }
```
But I don't fully understand how this is working. Can you please show me some more examples of it's usage, and let me know if what I'm doing here is the same as what you wrote about?

Comment (dmitrylyzo) [reply src/scripts/browserDeviceProfile.js:1154]: > It appears the server-side is aligning to implement this fix by introducing the new VideoRangeTypes "DOVIWithHDR10" and "DOVIWithHLG" for fallback-able DV.

I've been thinking about how we can get around the introduction of a new video range.

If the server will support combined video ranges, we can replace `DOVIWithHDR10` with `[VideoRangeType.DOVI, VideoRangeType.HDR10]`. This means that  the video has a Dolby Vision stream with HDR10 fallback.

Having `HDR10` in the `Condition` will allow playing HDR10 fallback at least.
But if we want to only play Dolby Vision, we can disallow HDR10 if there is also DOVI.

To apply the condition if there is DOVI, we can add the `ApplyCondition`:
```
ApplyConditions: [
    {
        Condition: 'EqualsAny',
        Property: 'VideoRangeType',
        Value: 'DOVI',
        IsRequired: false
    }
]
```
Then, to block any video range:
```
Conditions: [
    {
        Condition: 'Equals',
        Property: 'VideoRangeType',
        Value: '', // not sure if this will work
        IsRequired: false
    }
]
```

> But I don't fully understand how this is working. Can you please show me some more examples of it's usage, and let me know if what I'm doing here is the same as what you wrote about?

I haven't found any example. I only have the idea how it works (https://github.com/jellyfin/jellyfin/blob/231daa38dda8a572d41693fa458f74940ca53949/MediaBrowser.Model/Dlna/StreamBuilder.cs#L1583):

- First, the server checks the "apply" conditions.
- If the video passes them, the server checks the conditions.

In theory, your code (with `DOVIWithHDR10` and `DOVIWithHLG`) can be used to proof the idea.
You need these properties in the condition (I just omitted them):
```
profile.CodecProfiles.push({
    Type: 'Video',
    Container: '-mp4',
    Codec: 'hevc',
...
```
Otherwise, the condition applies to any media (audio, video) and any codec.

Also, I think this condition is in addition to the condition from this PR and should be controlled by the user through the playback settings.

**Comment by dmitrylyzo** [comment]
Just in case. This PR can be considered as "feature complete" - no need to add more.

My suggestion of using `NotEquals` is not really necessary, but it might be useful when more video ranges are added in the future.
On the other hand, it doesn't work if multiple video ranges are required - need another negation condition that supports a list of values.

DOVI+HDR10/HLG should be implemented in the new PR as it requires changes to the server.

File: src/scripts/browserDeviceProfile.js (hunk 3 above)
Comment (viktory36) [reply src/scripts/browserDeviceProfile.js:1154]: Ah, need to target Video. I was also thrown off by not needing a -mp4 with this approach. I can confirm the setup does work with below-
```
        profile.CodecProfiles.push({
            Type: 'Video',
            Container: '-mp4', //even without this line it works
            Codec: 'hevc',
            ApplyConditions: [
                {
                    Condition: 'EqualsAny',
                    Property: 'VideoRangeType',
                    Value: 'DOVI|DOVIWithHDR10|DOVIWithHLG', 
                    IsRequired: false
                }
            ],
            Conditions: [
                {
                    Condition: 'Equals',
                    Property: 'VideoRangeType',
                    Value: '', // disable whatever ended up matching in ApplyConditions from direct play, right? this will not apply if none of ApplyCondition match
                    IsRequired: false
                }
            ]
        });
    }
```
--
I remember taking a look at server to see if multiple videorangetypes could be supplied, but I found it too complex to change -- I remember too much of the code relied on there being one videorangetype. I felt the changes were not worth the undertaking to allow this feature to work

As a WebOS user, I don't think I would ever want to change the settings away from how it works right now - perhaps only making the "Enable fMP4" option a default for WebOS. If we're okay to take that up in a separate PR, do you recommend the above approach over the current implementation for this PR?

viktory36:
Awesome. @GeorgeH005, do you want to take that one up?

src/scripts/browserDeviceProfile.js
dmitrylyzo: > even without this line it works

Without `Container: '-mp4'` Jellyfin won't direct play MP4+DOVI (will always remux), no?

> As a WebOS user, I don't think I would ever want to change the settings away from how it works right now - perhaps only making the "Enable fMP4" option a default for WebOS.

Some users may only want direct playback, even if it's an HDR10/HLG fallback - they just don't want remux.

> If we're okay to take that up in a separate PR, do you recommend the above approach over the current implementation for this PR?

Current implementation is fine. Other options require changes in the server.

viktory36:
No problem - we will just need the new types added in server also added here:
https://github.com/jellyfin/jellyfin-web/blob/713ac0e130bfd452debf1c2a1b25a64b5bcaa01f/src/scripts/browserDeviceProfile.js#L930
https://github.com/jellyfin/jellyfin-web/blob/713ac0e130bfd452debf1c2a1b25a64b5bcaa01f/src/scripts/browserDeviceProfile.js#L936

So we will have 3 prs total. This one with no more changes, another one needed in jellyfin-web that deals with with DOVI+HDR10/HLG implementation that modifies above lines + others if necessary, and the one in server that introduces the new types.

With these 3 PRs merged, we will end up with full DOVI support on WebOS, and all other clients will now see "DOVIWithHDR10" instead of just "HDR10" when they're playing DOVI media.

thornbill:
The PR currently has merge conflicts. I believe it should be good to go once those are resolved.

viktory36:
@dmitrylyzo @thornbill resolved the conflict. Merge, please?

iwatch-x:
wow this would be great ! would this work on 10.9 then?

thornbill (approved):

PR #4918, opened 2023-10-27
Add crash reporter

thornbill:
**Changes**
Adds a crash reporting script that submits logs to the currently connected server when unhandled errors occur.

I wasn't sure where the best place to add this code would be... so suggestions are welcome if you think there is a better place for this. :sweat_smile: 

[Sample log](https://gist.github.com/thornbill/a1968a8c42ede0fb25bf4991797a93cc)

**Issues**
N/A

### DIFF
--- a/src/crashReporter/index.ts
+++ b/src/crashReporter/index.ts
@@ -1,0 +1,75 @@
+import { getClientLogApi } from '@jellyfin/sdk/lib/utils/api/client-log-api';
+
+import ServerConnections from 'components/ServerConnections';
+import { getSDK, toApi } from 'utils/jellyfin-apiclient/compat';
+
+import { buildLogTemplate } from './template';
+
+/** Firefox supports additional properties on the Error object */
+interface NonstandardError extends Error {
+    fileName?: string
+    columnNumber?: number
+    lineNumber?: number
+}
+
+interface OnUnhandledRejectionHandler {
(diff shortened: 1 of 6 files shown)


nielsvanvelzen:
We should make sure there is an opt-in/out for the crash reporting, I'd suggest adding a simple boolean to window.localStorage for that.

src/crashReporter/index.ts
```diff
@@ -0,0 +1,75 @@
...
+        });
+};
+
+const rejectionReporter: OnUnhandledRejectionHandler = (event) => {
+    const error = event.reason as NonstandardError;
+    const message = event.reason as string;
+    reporter(error.message ?? message, error.fileName, error.lineNumber, error.columnNumber, error);
+};
+
+window.onerror = reporter;
```
grhallenbeck, line 74: It might be good to build some guardrails in around a crash reporter listening to `window.onerror`, since it fires any time any resource fails to load or encounters an execution error regardless of source. If an extension crashes, that's not typically useful information since it's outside of the control of the app.

I don't know that it's possible to filter out everything, especially with Apple starting to completely mask paths of any extension code, but a starting point might be filtering out trace paths like:
* `chrome:`
* `chrome-extension:`
* `moz-extension:`
* `safari-extension:`

PR #4923, opened 2023-10-29
Randomize backdrop images in screensaver

Yenof:
~~Added 'Backdrop,Backdrop1,Backdrop2,Backdrop3,Backdrop4,Backdrop5' to imagetypes and Enableimagetypes.~~

~~Added "Season" to Includeitemtypes.~~

~~Increased `Limit` to 400.~~

Changed Slideshow.js to include `RandomInt` to pull random image instead of first. 

Increased `ImageTypeLimit` to 10.

This provides a more varied experience viewing the backdrop screensaver.

### DIFF
--- a/src/components/slideshow/slideshow.js
+++ b/src/components/slideshow/slideshow.js
@@ -14,6 +14,7 @@ import 'material-design-icons-iconfont';
 import '../../elements/emby-button/paper-icon-button-light';
 import ServerConnections from '../ServerConnections';
 import screenfull from 'screenfull';
+import { randomInt } from '../../utils/number.ts';
 
 /**
  * Name of transition event.
@@ -71,7 +72,8 @@ function getBackdropImageUrl(item, options, apiClient) {
     }
 
     if (item.BackdropImageTags?.length) {
-        options.tag = item.BackdropImageTags[0];
+        options.index = randomInt(0, item.BackdropImageTags.length - 1);
+        options.tag = item.BackdropImageTags[options.index];
         return apiClient.getScaledImageUrl(item.Id, options);
     }
 
--- a/src/plugins/backdropScreensaver/plugin.js
+++ b/src/plugins/backdropScreensaver/plugin.js
@@ -18,7 +18,7 @@ class BackdropScreensaver {
             SortBy: 'Random',
             Recursive: true,
             Fields: 'Taglines',
-            ImageTypeLimit: 1,
+            ImageTypeLimit: 10,
             StartIndex: 0,
             Limit: 200
         };


src/plugins/backdropScreensaver/plugin.js
```diff
@@ -12,15 +12,15 @@ class BackdropScreensaver {
     }
     show() {
         const query = {
-            ImageTypes: 'Backdrop',
-            EnableImageTypes: 'Backdrop',
-            IncludeItemTypes: 'Movie,Series,MusicArtist',
+            ImageTypes: 'Backdrop,Backdrop1,Backdrop2,Backdrop3,Backdrop4,Backdrop5',
+            EnableImageTypes: 'Backdrop,Backdrop1,Backdrop2,Backdrop3,Backdrop4,Backdrop5',
```
nielsvanvelzen, line 16: These image types don't exist, why add them?

Yenof: I was hoping someday it could work like this. I'd love to be able to have multiple backdrop images and it be able to use them each at random. 

I suppose atm there is no reason for it, but it doesn't seem cause issues to my client.

dmitrylyzo: Supporting multiple images of the same type doesn't require introducing of extra types.

Actual image retrieval is here:
https://github.com/jellyfin/jellyfin-web/blob/69bb25553efe7f0598786603acb25cbad088a334/src/components/slideshow/slideshow.js#L75
I haven't checked if `item.BackdropImageTags` contains more than one item.

Yenof: Okay I removed all the Backdrop# entries.

dmitrylyzo: > I haven't checked if `item.BackdropImageTags` contains more than one item.

Checked - it is filled with multiple backdrops.

If you want random backdrops, you need increase [this](https://github.com/jellyfin/jellyfin-web/blob/24929051b2b01f3c7ce1030b9afae5fe08524bd9/src/plugins/backdropScreensaver/plugin.js#L21) and change [this](https://github.com/jellyfin/jellyfin-web/blob/24929051b2b01f3c7ce1030b9afae5fe08524bd9/src/components/slideshow/slideshow.js#L74).

Yenof: Oh awesome thank you! I'd like to change the `ImageTypeLimit` to 5, but I'm not sure what to replace the `backdropImageTags[0]`; with. 
Could I use?
```
        const randomIndex = Math.floor(Math.random() * item.BackdropImageTags.length);
        options.tag = item.BackdropImageTags[randomIndex];
```

dmitrylyzo: It could be [`randomInt(0, item.BackdropImageTags.length - 1)`](https://github.com/jellyfin/jellyfin-web/blob/2d353e1845d9eecf5f9b8c58cfe9ebc8f3e00e6a/src/utils/number.ts#L11) but yours should work too.

src/components/slideshow/slideshow.js
```diff
@@ -71,7 +72,8 @@ function getBackdropImageUrl(item, options, apiClient) {
     }
 
     if (item.BackdropImageTags?.length) {
-        options.tag = item.BackdropImageTags[0];
+        const randomIndex = randomInt(0, item.BackdropImageTags.length - 1);
+        options.tag = item.BackdropImageTags[randomIndex];
```
dmitrylyzo, line 76: It turned out that we also need the `index` field.
```suggestion
        options.index = randomInt(0, item.BackdropImageTags.length - 1);
        options.tag = item.BackdropImageTags[options.index];
```

dmitrylyzo (commented):
The only problem (_maybe not_) is that `getBackdropImageUrl` is called once, so the backdrops are only randomized on screensaver (slideshow) startup.

src/components/slideshow/slideshow.js
Yenof: Thank you so much!

dmitrylyzo (approved):
LGTM
Maybe only the change `200` -> `400` is questionable.

Need to rebase and squash.

src/plugins/backdropScreensaver/plugin.js
```diff
@@ -18,9 +18,9 @@ class BackdropScreensaver {
             SortBy: 'Random',
             Recursive: true,
             Fields: 'Taglines',
-            ImageTypeLimit: 1,
+            ImageTypeLimit: 10,
             StartIndex: 0,
-            Limit: 200
+            Limit: 400
```
thornbill, line 23: 400 really does seem excessive... larger query sizes will inherently make things slower

Yenof: Okay, I went back to 200.

Yenof:
I do not think I did the squash or rebase thing correctly. Sorry about that, not really sure what I am doing in case that is unclear lol.

Yenof:
Yeah I really messed something up while trying to rebase and/or squash, it now says there are like 5000 changes. 

I thought I had followed the squash instructions I found correctly but that seems unlikely. 
Although instructions I found for "revert" seem straightforward, I do not want to potentially irreparably mess anything up, and think I should stop running commands I don't understand haha.

If anyone can fix this, that would be cool, but I am way out of the range of my current understanding and am currently unable to.

dmitrylyzo:
I will look tomorrow.

thornbill (approved):

PR #4930, opened 2023-11-02
Fix quality check workflows

thornbill:
**Changes**
Reverts the quality workflow trigger to use `pull_request` and moves the eslint suggestions to a separate workflow that explicitly checks out the PR SHA

**Issues**
None but the quality workflow is currently checking out master instead of the PR content :grimacing: :sweat_smile:


crobibero (approved):

thornbill:
CF Pages failure is due to their API being down: https://www.cloudflarestatus.com/

Issue #3243, opened 2021-12-27
Notification permission prompt interrupts input on first use


freshgum-bubbles:
**Describe The Bug**

Upon signing into Jellyfin, the user is interrupted with a popup to allow notifications
when entering their username. This happens on the first character of the username.

**Steps To Reproduce**

1. Log out, clear permissions from the page, etc.
2. Get to the sign in page, enter a character of your username.
3. The notification popup is shown when you type the first character of your username.

**Expected Behavior**

I think Jellyfin should wait for me to finish typing in my username and password (and possibly
even login) before prompting me asking for permission for notifications. This would mean I am
not interrupted when typing in my username (which is obviously quite annoying). Notifications are
an auxiliary thing and can wait; typing in my credentials is much more important.

**Logs**

**Screenshots**

**System (please complete the following information):**
 - Platform: [e.g. Linux, Windows, iPhone, Tizen]
 - Browser: [e.g. Firefox, Chrome, Safari]
 - Jellyfin Version: [e.g. 10.6.0]

**Additional Context**

It only happens in HTTPS due to the web platform not allowing the permission
to be garnered by HTTP websites. It also specifically happens when the first
character of the username is clicked, leading me to believe Jellyfin is waiting for
any form of user input then throwing up the notification permission popup as soon
as possible.

freshgum-bubbles:
Ahh, [here is the code](https://github.com/jellyfin/jellyfin-web/blob/5c09077a2f12ae44e7b7d356ea8d9f35862b0b60/src/components/notifications/notifications.js#L6) which seems to be causing all the trouble :D

thornbill:
In the code linked in the comment above, the event listeners should be added within an event listener for the `localusersignedin`. Examples of listening for this event can be found in several other places of the code.

PR #4935, opened 2023-11-03
Wait for login before ask for notification permission

SimonTen:
**Changes**
Previously the user was interrupted on first login with a popup to allow notifications.

Now the popup component checks if there is a current user and if not it will listen to the login event, to register the eventlisteners wich trigger the permission request

**Issues**
Fixes: #3243

### DIFF
--- a/src/components/notifications/notifications.js
+++ b/src/components/notifications/notifications.js
@@ -28,7 +28,7 @@ function initPermissionRequest() {
     const apiClient = ServerConnections.currentApiClient();
     if (apiClient) {
         apiClient.getCurrentUser()
-            .then(user => user && registerOneDocumentClickHandler())
+            .then(user => registerOneDocumentClickHandler())
             .catch(() => {
                 Events.on(ServerConnections, 'localusersignedin', registerOneDocumentClickHandler);
             });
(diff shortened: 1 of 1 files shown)


src/components/notifications/notifications.js
```diff
@@ -16,9 +17,27 @@ function onOneDocumentClick() {
         Notification.requestPermission();
     }
 }
+function registerOneDocumentClickHandler() {
```
dmitrylyzo, line 20:
```suggestion
}

function registerOneDocumentClickHandler() {
```

src/components/notifications/notifications.js
```diff
@@ -16,9 +17,27 @@ function onOneDocumentClick() {
...
-document.addEventListener('keydown', onOneDocumentClick);
+    document.addEventListener('click', onOneDocumentClick);
+    document.addEventListener('keydown', onOneDocumentClick);
+}
+
+function initPermissionRequest() {
+    const apiClient = ServerConnections.currentApiClient();
+    if (apiClient) {
+        apiClient.getCurrentUser()
+            .then(user => registerOneDocumentClickHandler())
```
dmitrylyzo, line 31:
```suggestion
            .then(() => registerOneDocumentClickHandler())
```

thornbill (approved):

PR #4938, opened 2023-11-04
Remove FLAC Audio in Video Support for Tizen

rafma0:
I opened this issue here https://github.com/jellyfin/jellyfin/issues/10504 where I explain it better (wrong repo).

Basically, FLAC support on Samsung is weird. In their documentation, they don't claim to support it, but you can play it. However, you will probably get audio and video that are not synced most of the time. I also tested and found that Plex transcodes FLAC audio in videos for Samsung, and I found a post in the Emby forums (linked in the issue I opened) where they update it to do the same.

**Changes**
Remove FLAC audio in video support for Tizen

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -536,7 +536,7 @@ export default function (options) {
         }
     }
 
-    if (canPlayAudioFormat('flac')) {
+    if (canPlayAudioFormat('flac') && !browser.tizen) {
         videoAudioCodecs.push('flac');
         hlsInFmp4VideoAudioCodecs.push('flac');
     }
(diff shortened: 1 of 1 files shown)


src/scripts/browserDeviceProfile.js
```diff
@@ -536,7 +536,7 @@ export default function (options) {
         }
     }
 
-    if (canPlayAudioFormat('flac')) {
+    if (canPlayAudioFormat('flac') && !browser.tizen) {
```
dmitrylyzo, line 539: Maybe add a comment that FLAC in video is delayed on Tizen?

dmitrylyzo (commented):
On my NU7400 (Tizen 4) FLAC has ~1s delay in MKV and no sound in MP4.

src/scripts/browserDeviceProfile.js
rafma0: done!

dmitrylyzo (approved):

PR #4942, opened 2023-11-05
Disable the broken AV1 decoding on Edge 117+

nyanmisaka:
Edge Chromium broke the AV1 hardware decoding in 117+, and it can't decode AV1 in software as a fallback either.

**Changes**
- Disable the broken AV1 decoding on Edge 117+
- Re-enabled the HDR playback on Edge 117+
- Fix a typo in Chrome version check
- Update fMP4-HLS descriptions for AV1

**Issues**
https://www.reddit.com/r/AV1/comments/16mmm35/av1_and_edge/
https://techcommunity.microsoft.com/t5/discussions/av1-not-working-and-dropping-frames-on-yt-videos/m-p/3946087

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -200,7 +200,8 @@ function supportsHdr10(options) {
             || browser.web0s
             || browser.safari && ((browser.iOS && browser.iOSVersion >= 11) || browser.osx)
             // Chrome mobile and Firefox have no client side tone-mapping
-            // Edge Chromium on Nvidia is known to have color issues on 10-bit video
+            // Edge Chromium 117+ fixed the tone-mapping color issue on Nvidia
+            || browser.edgeChromium && browser.versionMajor >= 117
             || browser.chrome && !browser.mobile
     );
 }
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -28,6 +28,12 @@ function canPlayAv1(videoTestElement) {
         return true;
     }
 
+    // Edge Chromium broke the AV1 hardware decoding in 117+,
+    // and it can't decode AV1 in software as a fallback either.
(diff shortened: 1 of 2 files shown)


nyanmisaka:
Not at all, a Chromium developer told me that MS Edge 117+ has fully moved to `Media Foundation For Clear`, and AV1 is very unstable with it, so MS has decided to not re-enable AV1 decoding any time soon.

You need to check in 'stats for nerds' whether the video you play titled 8K AV1 is encoded in AV1.

msimic:
It doesn't even offer you the AV1 options in Youtube.
Can you try streaming directly an AV1 video in Edge?

[Stream1_AV1_4K_8.5mbps.webm](https://www.elecard.com/storage/video/Stream1_AV1_4K_8.5mbps.webm)

This works perfectly for me even on 12 gen integrated graphics.

msimic:
> I know about this option, but it's experimental and the default is disabled. So you can't expect users to manually enable it.

Makes sense. I have to use it at least on the laptop tho, since the chromium default drop me frames on any HDR video if I don't use MF. It is much more performant at least on the iGPU. The standard decoding maxxes the 3D part of the GPU and hardly uses any video decoding capabilities.

msimic:
After a bit of investigation I can confirm that Edge 119 without having **Media Foundation Clear flag** enabled is totally unable to play AV1 and stops as soon as you try it. After a bit of GPU stress it silently fails.

However the **videoElement.canPlayType("video/webm; codecs=av01")** correctly returns empty string.
For **video/webm; codecs=vp9** also returns correctly "probably"

On the other hand if you DO use Media Foundation Clear flag it works flawlessly, even on an 12 gen integrated Iris Xe graphics.
I made my own player with stats to avoid youtube and plugins to unblock it formats and check directly.

image

thornbill:
@nyanmisaka any thoughts on the last comment here?

Issue #4940, opened 2023-11-05
Black screen on Tizen 5.0 (2019) after #35429ca


rafma0:
**Describe The Bug**
A black screen appears on Tizen 5.0 (2019) after commit  https://github.com/jellyfin/jellyfin-web/commit/35429ca61f223932e86c597a6aec745783b0b7c5

**Steps To Reproduce**
- Compile latest jellyfin-web and jellyfin-tizen
- Deploy to TV and run

Result: The splash screen is displayed, followed by a black screen. Pressing the return button only displays the quit message once.

dmitrylyzo:
> Uncaught (in promise) TypeError: Cannot read property 'removeEventListener' of undefined

PR #4958, opened 2023-11-09
Fix legacy

dmitrylyzo:
Regression https://github.com/jellyfin/jellyfin-web/pull/4834

**Changes**
- Remove `!` from `start_url`.
- Add `AbortController` polyfill.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-web/issues/4940

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -3,7 +3,7 @@
     "description": "The Free Software Media System",
     "lang": "en-US",
     "short_name": "Jellyfin",
-    "start_url": "index.html#!/home.html",
+    "start_url": "index.html#/home.html",
     "theme_color": "#101010",
     "background_color": "#101010",
     "display": "standalone",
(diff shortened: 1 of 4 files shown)


src/manifest.json
```diff
@@ -3,7 +3,7 @@
     "description": "The Free Software Media System",
     "lang": "en-US",
     "short_name": "Jellyfin",
-    "start_url": "index.html#!/home.html",
+    "start_url": "index.html#/home.html",
```
thornbill, line 6: Either way this should be updated, but are `#!` routes not working? We _should_ be automatically updating `#!` routes to the `#` equivalent.

dmitrylyzo: It tries to match `!/home.html` with available routes. Maybe initial navigation isn't normalized somewhere?

thornbill: Ok I'll do some additional testing on those paths but that doesn't block this. 👍

thornbill (approved):

rafma0:
@dmitrylyzo I'm still seeing a back screen on my samsung, can you please test on yours? thank you

Issue #4794, opened 2023-09-20
JASSUB doesn't work in legacy browsers


dmitrylyzo:
**Describe The Bug**
JASSUB doesn't work in legacy browsers: worker fails on webOS 1.2, requiring `fetch`, `Promise` (probably more).

**Steps To Reproduce**
1. Run app in webOS 1.2 emulator using `ares-inspect`.
2. Open ares-inspect address in the Chrome 79 (newer may work too).
3. Start playback of video with ASS subtitles.
4. See error in the console.

**Expected Behavior**
No error and see subtitles.

**System (please complete the following information):**
 - Platform: webOS 1.2
 - Browser: Jellyfin webOS app
 - Jellyfin Version: 10.9

**Additional Context**
Promise in webOS 2 has a weird implementation, so it also requires testing.
JASSUB 1.5.13 works fine, because it was built with compatibility flags (browser version).

ThaUnknown:
this is.... unusual?

jassub only uses fetch for engines which support streamed WASM compilation, which REQUIRES fetch in the engine to work, as it uses fetch responses, so if fetch doesnt exist, streamed compilation will never run, and never call fetch, so it's a non-issue?

ThaUnknown:
![image](https://github.com/jellyfin/jellyfin-web/assets/6506529/837b75f0-3d48-400a-9196-2424a33209fa)
also I can't get JF to even play any media to test this :&

ThaUnknown:
I couldn't get JF-web to work, so I simply [copied its webpack config](https://github.com/ThaUnknown/jassub/blob/gh-pages/webpack.config.js) and hosted it on [my test page](https://thaunknown.github.io/jassub/jassub/legacy/index.html), it works on chrome 76, but I can't really test webOS:
![image](https://github.com/jellyfin/jellyfin-web/assets/6506529/f20a5886-2635-47aa-9769-07749bf00b02)

ThaUnknown:
the worker always required transpiration, [this isn't new](https://github.com/jellyfin/jellyfin-web/pull/4357/files#diff-3e31f6f375484b2dfd4629faed9e5183ac6568ebdbc67ec752e10c41a35250c6R185-R198), I'll look into missing globals when using strict, or I could remove use strict which I think is added by vite [as always causes problems]

also I don't see why it would cry about WebAssembly missing, if it doesn't exist it should error inside the trycatch, and polyfill it, and the next time it's ran it's already defined

[...]

PR #4971, opened 2023-11-09
fix: legacy JASSUB wasm

ThaUnknown:
production mode makes Terser run on ALL files even statically copied ones, and since it has dead code removal it can remove dynamic code [bad!]
**Changes**

Used a custom TerserPlugin configuration, which copies [webpack's default](https://github.com/webpack/webpack/blob/main/lib/config/defaults.js#L1339), and simply excludes `.wasm.js` files which terser was breaking. Regrettably I had to undo the import based URL loading for the `.wasm.js` file in question, and instead use static copy as I couldn't get it to work via imports, not sure why, this could be improved upon
**Issues**

Maybe #4794 ? Needs testing.


dmitrylyzo:
[`Math.clz32`](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Math/clz32) and [`Math.fround`](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Math/fround) polyfills are required. Who knows what else. But it started working on webOS 1.2

That's probably because you have removed the chrome version option from the emscripten command when making the combined build.

ThaUnknown:
iirc the min versions were removed because they forcefully enabled emscripten's closure compiler, which removed their own code and prevented the library from working at all

guess I'll spend a few more hours debugging their code once again, if not I'll have to polyfill all this stuff manually

dmitrylyzo:
> the polyfills are missing but the library works fine? or was it after you simply added all of core.js?

I added them to `jassub-worker.wasm.js` in `dist`.
Btw, this also works for the current `master` (without this PR).

ThaUnknown:
is that when you copy over the wasm.js or is it when u add said polyfills to the worker?

ThaUnknown:
> [`Math.clz32`](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Math/clz32) and [`Math.fround`](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Math/fround) polyfills are required. Who knows what else. But it started working on webOS 1.2
> 
> That's probably because you have removed the chrome version option from the emscripten command when making the combined build.

I've re-introduced min browser versions, the issues they caused were fixed in the version of emscripten i've updated to a while back

dmitrylyzo:
> is that when you copy over the wasm.js or is it when u add said polyfills to the worker?

1. Checkout `jellyfin-web:master`.
2. Install modules.
3. Install (add) `jassub@1.7.12`.
4. Add 2 polyfills to `dist/jassub-worker.wasm.js` at the beginning.
5. Build jellyfin-web.

ThaUnknown:
> > is that when you copy over the wasm.js or is it when u add said polyfills to the worker?
> 
> 1. Checkout `jellyfin-web:master`.
> 2. Install modules.
> 3. Install (add) `jassub@1.7.12`.
> 4. Add 2 polyfills to `dist/jassub-worker.wasm.js` at the beginning.
> 5. Build jellyfin-web.

this just flat out doesnt work

I get JF master, build it, copy it over to my JF installation, no playback works on any browser

rafma0:
hmm, this is whats happening when I try to set 2 different ASS tracks on Tizen and then turn it off, using latest master or this PR. am I missing something @dmitrylyzo ?

image

PR #4973, opened 2023-11-09
Add abortable fetch polyfill

dmitrylyzo:
`AbortController` is not enough (https://github.com/jellyfin/jellyfin-web/pull/4958).

**Changes**
Add abortable fetch polyfill.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-web/issues/4940

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -1,11 +1,11 @@
 import 'core-js/stable';
 import 'regenerator-runtime/runtime';
 import 'jquery';
-import 'abortcontroller-polyfill/dist/abortcontroller-polyfill-only';
 import 'fast-text-encoding';
 import 'intersection-observer';
 import 'classlist.js';
 import 'whatwg-fetch';
+import 'abortcontroller-polyfill'; // requires fetch
 import 'resize-observer-polyfill';
 import './styles/site.scss';
 import React, { StrictMode } from 'react';


thornbill (approved):

rafma0:
@dmitrylyzo sorry if I'm being annoying but it's still not working. I tried running the debug build in the 5.0 emulator, to see if it would show any messages on device manager, then I tried running in on tizen ide to see if I could debug the js but I couldn't make it launch. Where I'm at right now is I can insert the server ip, and it gets stuck on a screen with the clock on top.

image

dmitrylyzo:
I tested in Tizen 5 emulator too and it worked. 😕 
But I usually test the development build.

Also, this may be due to a recent dependency update.

To run Tizen debug you need Chrome 79- (or 80- ?) and start debugging from the command line. See https://github.com/jellyfin/jellyfin-tizen/issues/79

rafma0:
Thank you for the link to the guide, I managed to make debug work with latest Chrome (119) opening `chrome://inspect/#devices`, in Discover network targets click `Configure...`, add tv-ip:debug-port, and in ~5 secs it shows as a new Remote Target. Then click `inspect`.

Here's the results after I put the ip and click Connect

image

I also deleted node_modules from jellyfin(web|tizen) and `npm ci --no-audit` again. I'm probably just too tired. I'll try again tomorrow :)

dmitrylyzo:
Yes, webOS 1.2 fails again. :weary:
> node_modules.core-js.bundle.js?bf36f1725aebd62b4fa5: Invalid flags supplied to RegExp constructor.

rafma0:
I'm not really experienced with webpack and babel but I tried a few things to see if I could change how marked and/or core-js are transpiled... and I didn't have any success 😆. The latest version of marked that seems to be working is 5.0.5. Not sure if we should set marked to this version and fixate it in renovate, or if there's a better simpler solution.

dmitrylyzo:
> I'm not really experienced with webpack and babel

Me, too 😅 
I tried [@babel/plugin-transform-unicode-regex](https://babeljs.io/docs/babel-plugin-transform-unicode-regex). But it doesn't transform `RegExp('...', 'u')`.

rafma0:
Another option would be replacing `marked` with `markdown-it`. This lib seems to be maintaned and nearly as popular. I did some tests here https://github.com/rafma0/jellyfin-web/tree/replace-markdown-lib and it seems to work the same, and looks like the maintaner care about keeping compatibility with old browsers https://github.com/markdown-it/markdown-it/issues/948#issuecomment-1736152234, should I send a PR?

PR #4984, opened 2023-11-11
Add option to enable replaygain scanning

TelepathicWalrus:
Adds option to enable/disable scanning for replaygain tags. This is means that existing replaygain tags can be used instead of jellyfin scanning each track again.

Backend: https://github.com/jellyfin/jellyfin/pull/10566

### DIFF
--- a/src/strings/en-us.json
+++ b/src/strings/en-us.json
@@ -971,6 +971,8 @@
     "LabelTypeText": "Text",
     "LabelUnstable": "Unstable",
     "LabelUser": "User",
+    "LabelUseReplayGainTagsHelp": "Scan audio files for replaygain tags and use them instead of computing LUFS value. (Uses less computing power. Will override 'LUFS Scan' option)",
+    "LabelUseReplayGainTags": "Use ReplayGain Tags",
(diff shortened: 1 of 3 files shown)


TelepathicWalrus:
This has been merged server side now, so can be merged once reviewed :)

Gauvino (approved):
LGTM !

src/strings/en-us.json
```diff
@@ -971,6 +971,8 @@
     "LabelTypeText": "Text",
     "LabelUnstable": "Unstable",
     "LabelUser": "User",
+    "LabelUseReplayGainTagsHelp": "Scan audio files for replaygain tags and use them instead of computing LUFS value. (Uses less computing power. Will override 'LUFS Scan' option)",
```
thornbill, line 974: I found this wording a little confusing... does it mean that if replaygain tags are found then the LUFS scan will be skipped for _that_ file?

TelepathicWalrus: Yeah that is correct

thornbill (approved):

PR #4995, opened 2023-11-14
Replace markdown lib

rafma0:
I'm opening this PR, but I'm not sure this is the best option to solve the problem. As discussed here https://github.com/jellyfin/jellyfin-web/pull/4973, after this PR https://github.com/jellyfin/jellyfin-web/pull/4975 `marked` stopped working on legacy browsers. The latest version that seems to work is `5.0.5`. This is because they started using the `u` flag in `RegExp`.

We tried to perform some babel magic but had no success, so I decided to try another lib, `markdown-it`. This lib seems well-maintained and works the same as `marked`, using the latest version and in legacy browsers. Feel free to close this if you prefer to solve it any other way :)

**Changes**
Replace `marked` with `markdown-it`


webpack.common.js
```diff
@@ -196,7 +196,7 @@ const config = {
                     path.resolve(__dirname, 'node_modules/flv.js'),
                     path.resolve(__dirname, 'node_modules/is-what'),
                     path.resolve(__dirname, 'node_modules/libarchive.js'),
-                    path.resolve(__dirname, 'node_modules/marked'),
+                    path.resolve(__dirname, 'node_modules/markdown-it'),
```
dmitrylyzo, line 199:
```suggestion
                    path.resolve(__dirname, 'node_modules/markdown-it'),
                    path.resolve(__dirname, 'node_modules/punycode'),
```

dmitrylyzo (commented):
It works on webOS 1.2, but doesn't support HTML tags (`marked` does).

For example `h1`:
```
<h1>Disclaimer</h1>

Jellyfin is the volunteer-built media solution that puts you in control of your media. Stream to any device from your own server, with no strings attached. Your media, your server, your way.

- Jellyfin is the volunteer-built media solution that puts you in control of your media. Stream to any device from your own server, with no strings attached. Your media, your server, your way.
- second test line 1  
second test line 2

[https://jellyfin.org/](https://jellyfin.org/)

1. one
2. two
3. three
```

webpack.common.js
rafma0: I just added a param for markdownIt to handle html, seems to work correctly, please test

for your example, it generated
```html
<h1>Disclaimer</h1>
<p>Jellyfin is the volunteer-built media solution that puts you in control of your media. Stream to any device from your own server, with no strings attached. Your media, your server, your way.</p>
<ul>
<li>Jellyfin is the volunteer-built media solution that puts you in control of your media. Stream to any device from your own server, with no strings attached. Your media, your server, your way.</li>
<li>second test line 1<br>
second test line 2</li>
</ul>
<p><a href="https://jellyfin.org/">https://jellyfin.org/</a></p>
<ol>
<li>one</li>
<li>two</li>
<li>three</li>
</ol>
```

rafma0: just a side note, I'm seeing sonarcloud complaining about possible xss etc, looks like a false positive since it still goes through dompurify sanitize

rafma0: ah ok I added punycode to webpack config now

dmitrylyzo (approved):

Letrab:
There is still a security risk flagged: https://sonarcloud.io/project/security_hotspots?id=jellyfin_jellyfin-web&pullRequest=4995&resolved=false&types=SECURITY_HOTSPOT&tab=how_to_fix

Is the `{ html: true }` needed? If so, an external sanitizer should be used:
https://github.com/markdown-it/markdown-it/blob/master/docs/security.md

dmitrylyzo:
> There is still a security risk flagged: https://sonarcloud.io/project/security_hotspots?id=jellyfin_jellyfin-web&pullRequest=4995&resolved=false&types=SECURITY_HOTSPOT&tab=how_to_fix
> 
> Is the `{ html: true }` needed? If so, an external sanitizer should be used: https://github.com/markdown-it/markdown-it/blob/master/docs/security.md

There is `DOMPurify.sanitize`.

PR #5011, opened 2023-11-21
Add chapter selection to OSD

TheMelmacian:
**Changes**
I added a selection menu to the OSD which allows to directly select a chapter.
It shows the number, chapter name and the start time.
If no name is set, a name in the form `Chapter 01` is generated.
Th currently active chapter is calculated based on playtime and chapter start/end times and shown in the menu.

~~To prevent the volume from changing when scrolling through the menu, the mouse wheel listener is deactivated while the menu is visible.~~
Edit: This was fixed with https://github.com/jellyfin/jellyfin-web/pull/5401, so I adjusted my code accordingly.

For the button I used the ordered list icon:
jellyfin_chapter_selection

jellyfin_chapter_selection_menu


Gauvino:
Possible to add on 10.9 ?

TheMelmacian:
The 10.9 release is currently prepared and no more features are added: https://jellyfin.org/posts/testing-10.9.0
Probably 10.10 if the PR gets accepted.

1hitsong:
I like the idea. This resembles the same menu I created on Roku 🤘 

If I click the Chapters button and click off the menu to close it instead of clicking on a chapter, I get the following error:
image

thornbill:
That error is an artifact of our weird dialog code that rejects when closed... a couple places depend on it though so probably just need a catch to suppress it here.

TheMelmacian:
The settings, audio selection and subtitle selection are also affected by this.
Adding a catch to the actionsheet promise is easy but the question is should all errors be supressed
```javascript
.catch(() => {
    // prevent 'ActionSheet closed without resolving' error
})
```
or only the `ActionSheet closed without resolving` error
```javascript
.catch((error) => {
    // supress 'ActionSheet closed without resolving' error
    // log and throw all other errors
    if (error !== 'ActionSheet closed without resolving') {
        console.error('ActionSheet error', error);
        throw new Error(error);
    }
})
```

just let me now and I will provide a fix for this.

thornbill:
I would just catch all for now.

1hitsong (dismissed):
Tested in Windows 11.

Functionality works well. Code looks good.

Possible followup idea: Display the chapter images to the side of the chapter names.

TheMelmacian:
since this didn't make the cut for 10.10, any plans to add it in 10.11 ?

thornbill:
Thanks for your patience @TheMelmacian! I will see if our UI/UX lead has feedback.

thornbill:
Sorry again for the delay on this. I reviewed the change with our ui/ux lead and I think we are going to hold off and add this functionality as part of the full redesign effort.

TheMelmacian:
As long as the redesign is planned for the next major release, I have no problem with that.

ItsMCB:
I strongly value this chapter list feature as someone who frequently seeks a particular moment in my media, such as a specific performance from a four-hour concert. Is there an ETA for the full redesign effort that would include this feature?

TheMelmacian:
> Sorry again for the delay on this. I reviewed the change with our ui/ux lead and I think we are going to hold off and add this functionality as part of the full redesign effort.

Sadly the redesign wasn't completed with 10.11.z and we still have to wait for this feature.
It's now the third major release (and nearly two years) this PR was not included and the advertised redesign that should include the same functionality wasn't finished either.
Why holding off this feature for two years? It doesn't include side effects or major code changes that could influence a redesign in any meaningful way. The change in the GUI is also minimal. And who knows when the redesign is finally finished and released?

thornbill:
It's fair to say the redesign work has stalled a bit... In light of that can you help me understand what need this is trying to solve? Chapters are already listed on the item details screen and in the video player timeline (similar to YouTube and other streaming platforms). What is the advantage in displaying them in a third way/place?

TheMelmacian:
The later added "chapters on timeline" feature was never an argument against this PR. Only that the same feature would be part of the redesign.
Having the chapters directly on the timeline has its advantages, but there are also some shortcomings that the chapter list has not. I think both features complement each other quite well.
MPC-HC for example has both, and the user can choose based on preferences and his current use case.

- the current timeline is lacking chapter names
- the timeline doesn't give a good and quick overview of all existing chapters (with names and times)
- its way harder to search and jump to a specific chapter
- to see chapter times I have to hover over the timeline at the the specific marker
- the timeline doesn't work well with media with a very high number of (short) chapters

See other comments:
> I strongly value this chapter list feature as someone who frequently seeks a particular moment in my media, such as a specific performance from a four-hour concert. Is there an ETA for the full redesign effort that would include this feature?

I don't see any chapter information on the item details screen. I'm not sure where they're supposed to be.
But any chapter information outside of the player is irrelevant for selecting a chapter during playback.

PR #5015, opened 2023-11-26
Fixed issue with displaying very long artist names

runbing:
The artist name field gets messy when I add a music album with a really long artist name. I tweaked the desktop layout's CSS to show the artist's name as the item name, cutting off the lengthy text with an ellipsis.

Before:

before

After:

after

### DIFF
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -440,12 +440,12 @@ function renderName(item, container, context) {
         if (parentNameLast) {
             // Music
             if (layoutManager.mobile) {
-                html = '<h3 class="parentName musicParentName">' + parentNameHtml.join('</br>') + '</h3>';
+                html = '<h3 class="parentName infoText musicParentName">' + parentNameHtml.join('</br>') + '</h3>';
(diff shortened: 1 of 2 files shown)


1hitsong:
Same root issue is addressed here as https://github.com/jellyfin/jellyfin-web/pull/4830

Same as I said there, I like the CSS fix over updating code to a hard-coded value. My 1 suggestion was to allow 2 rows of names.

ryanwwest:
I hope that other PR can solve this, or if not this can be merged. I frequently have the same problem when there are tons of Artists to display.

thornbill:
Closing in favor of #4830

ryanwwest:
Does https://github.com/jellyfin/jellyfin-web/pull/4830 handle if there are 10 (or 2) very _long_ artist names that still cause display overflow? I know that would in general solve the issue here, but couldn't there still be instances when 10 or 2 artists cause overflow?

thornbill:
No it would not solve for that, but I think it's an edge case of an edge case to have enough artists with ridiculously long names that it would be an issue.

PR #5018, opened 2023-11-28
Replace "Object.assing()" With an Object Spread

Adam4376:
This is a change requested by [this issue](https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&id=jellyfin_jellyfin-web&open=AYnuoirDJnGD1tMStw4d&tab=code) on Sonar Cloud

### DIFF
--- a/src/components/metadataEditor/metadataEditor.js
+++ b/src/components/metadataEditor/metadataEditor.js
@@ -167,7 +167,7 @@ function onSubmit(e) {
         })
     };
 
-    item.ProviderIds = Object.assign({}, currentItem.ProviderIds);
+    item.ProviderIds = { ...currentItem.ProviderIds };
 
     const idElements = form.querySelectorAll('.txtExternalId');
     Array.prototype.map.call(idElements, function (idElem) {


seeplusplus (approved):

thornbill:
Typically what I look for in these type of cleanup PRs is that they fix all instances of the issue and add an eslint rule (if possible) to prevent new instances from being added.

* [all reported violations of this rule](https://sonarcloud.io/project/issues?resolved=false&rules=javascript%3AS6661&id=jellyfin_jellyfin-web)
* [eslint rule](https://eslint.org/docs/latest/rules/prefer-object-spread)

Thanks!

seeplusplus (approved):

Adam4376:
> Typically what I look for in these type of cleanup PRs is that they fix all instances of the issue and add an eslint rule (if possible) to prevent new instances from being added.
> 
>     * [all reported violations of this rule](https://sonarcloud.io/project/issues?resolved=false&rules=javascript%3AS6661&id=jellyfin_jellyfin-web)
> 
>     * [eslint rule](https://eslint.org/docs/latest/rules/prefer-object-spread)
> 
> 
> Thanks!

I believe I've changed all the usages. Added the eslint rule as well.

src/components/cardbuilder/peoplecardbuilder.js
```diff
@@ -7,15 +7,17 @@
 import cardBuilder from './cardBuilder';
 
 export function buildPeopleCards(items, options) {
-    options = Object.assign(options || {}, {
+    options = {
+        ...options,
```
grhallenbeck, line 11: I'm not sure why there was a default before, but we probably need to keep it to be safe. If `options` is undefined and we try to spread it, it will throw an exception. There are a few other places throughout that this change would need to be made as well.

```suggestion
        ...(options || {}),
```

thornbill: Good catch! For this instance at least, you could provide a default value for the parameter as an alternative too.

```ts
export function buildPeopleCards(items, options = {}) {
```

Adam4376: Appreciate the suggestion. I believe I've added them all back.

PR #5020, opened 2023-11-29
Update mui based layout ui

thornbill:
**Changes**
* Makes the drawer permanent on larger screens
* Adds drawer header link to replace toolbar home link
* Fixes spacing of drawer icons
* Fixes top padding on home screen and item details screen

Screenshot 2023-11-28 at 10-35-27 Jellyfin

**Issues**
N/A

### DIFF
--- a/src/apps/dashboard/AppLayout.tsx
+++ b/src/apps/dashboard/AppLayout.tsx
@@ -1,7 +1,8 @@
 import AppBar from '@mui/material/AppBar';
 import Box from '@mui/material/Box';
-import { useTheme } from '@mui/material/styles';
-import React, { FC, useCallback, useEffect, useState } from 'react';
+import { type Theme } from '@mui/material/styles';
+import useMediaQuery from '@mui/material/useMediaQuery';
+import React, { FC, useCallback, useState } from 'react';
 import { Outlet, useLocation } from 'react-router-dom';
 
 import AppBody from 'components/AppBody';
@@ -9,7 +10,6 @@ import AppToolbar from 'components/toolbar/AppToolbar';
 import ElevationScroll from 'components/ElevationScroll';
 import { DRAWER_WIDTH } from 'components/ResponsiveDrawer';
 import { useApi } from 'hooks/useApi';
-import { useLocalStorage } from 'hooks/useLocalStorage';
 
 import AppDrawer from './components/drawer/AppDrawer';
 
@@ -19,34 +19,17 @@ interface AppLayoutProps {
     drawerlessPaths: string[]
 }
 
-interface DashboardAppSettings {
-    isDrawerPinned: boolean
-}
-
-const DEFAULT_APP_SETTINGS: DashboardAppSettings = {
-    isDrawerPinned: false
-};
-
 const AppLayout: FC<AppLayoutProps> = ({
     drawerlessPaths
 }) => {
-    const [ appSettings, setAppSettings ] = useLocalStorage<DashboardAppSettings>('DashboardAppSettings', DEFAULT_APP_SETTINGS);
-    const [ isDrawerActive, setIsDrawerActive ] = useState(appSettings.isDrawerPinned);
+    const [ isDrawerActive, setIsDrawerActive ] = useState(false);
     const location = useLocation();
-    const theme = useTheme();
     const { user } = useApi();
 
-    const isDrawerAvailable = !drawerlessPaths.some(path => location.pathname.startsWith(`/${path}`));
+    const isSmallScreen = useMediaQuery((t: Theme) => t.breakpoints.up('sm'));
+    const isDrawerAvailable = !isSmallScreen
+        && !drawerlessPaths.some(path => location.pathname.startsWith(`/${path}`));
     const isDrawerOpen = isDrawerActive && isDrawerAvailable && Boolean(user);
-
-    useEffect(() => {
-        if (isDrawerActive !== appSettings.isDrawerPinned) {
-            setAppSettings({
-                ...appSettings,
-                isDrawerPinned: isDrawerActive
-            });
-        }
-    }, [ appSettings, isDrawerActive, setAppSettings ]);
 
     const onToggleDrawer = useCallback(() => {
         setIsDrawerActive(!isDrawerActive);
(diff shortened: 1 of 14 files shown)


src/apps/dashboard/AppLayout.tsx
```diff
@@ -1,15 +1,15 @@
 import AppBar from '@mui/material/AppBar';
 import Box from '@mui/material/Box';
-import { useTheme } from '@mui/material/styles';
-import React, { FC, useCallback, useEffect, useState } from 'react';
+import { type Theme } from '@mui/material/styles';
```
seeplusplus, line 3: Hi, just getting started working with jellyfin and I don't work with React (and hence nor mui), so apologies if I'm missing something here. I haven't ever seen `import type` actually used and the docs make it sound unnecessary for most users:

> This feature is something most users may never have to think about

and 

>  In conjunction with import type, TypeScript 3.8 also adds a new compiler flag to control what happens with imports that won’t be utilized at runtime: [importsNotUsedAsValues](https://www.typescriptlang.org/tsconfig#importsNotUsedAsValues). This flag takes 3 different values:
> 
>     remove: this is today’s behavior of dropping these imports. It’s going to continue to be the default, and is a non-breaking change.

Are the `type`s here needed?

seeplusplus: fwiw I wanted to see if it made a difference, so I did production builds of your branch with and without the `import types`. I checksum'd each and they are the same (I'm not sure if this is true for dev builds).

seeplusplus:
looks like the drawer doesn't automatically dismiss itself when playing media
image

src/apps/dashboard/AppLayout.tsx
nielsvanvelzen: Using `input type` helps with transpilers (like esbuild/babel) as they can entirely ignore the import statement (and skip reading the file) because they don't do type checking. In large projects this can reduce build times significantly.

seeplusplus: Thank you for sharing, that's good to know

src/apps/experimental/AppLayout.tsx
```diff
@@ -1,93 +1,72 @@
...
-            />
+            {
+                user && (
+                    <AppDrawer
+                        open={isDrawerOpen}
+                        onClose={onToggleDrawer}
+                        onOpen={onToggleDrawer}
+                    />
+                )
+            }
```
grafixeyehero, line 63:
```suggestion
            {
                isDrawerAvailable && (
                    <AppDrawer
                        open={isDrawerOpen}
                        onClose={onToggleDrawer}
                        onOpen={onToggleDrawer}
                    />
                )
            }
```

src/apps/experimental/AppLayout.tsx
```diff
@@ -1,93 +1,72 @@
...
             </ElevationScroll>
 
-            <AppDrawer
-                open={isDrawerOpen}
-                onClose={onToggleDrawer}
-                onOpen={onToggleDrawer}
-            />
+            {
+                user && (
+                    <AppDrawer
```
grafixeyehero, line 57:
```suggestion
                isDrawerAvailable && (
```

src/apps/experimental/AppLayout.tsx
```diff
@@ -1,93 +1,72 @@
...
                 </AppBar>
             </ElevationScroll>
 
-            <AppDrawer
-                open={isDrawerOpen}
-                onClose={onToggleDrawer}
-                onOpen={onToggleDrawer}
-            />
+            {
+                user && (
```
grafixeyehero, line 56:
```suggestion
                isDrawerAvailable && (
```

grafixeyehero (changes requested):

src/apps/experimental/AppLayout.tsx
thornbill: :+1: good call... the dashboard (specifically the metadata editor) had the same issue also.

seeplusplus:
minor detail on the metadata manager after the latest fix: the back button doesn't realign to  the left of the header
image

src/apps/dashboard/AppLayout.tsx
```diff
@@ -19,82 +19,61 @@ interface AppLayoutProps {
...
-            <ElevationScroll elevate={isDrawerOpen}>
+            <ElevationScroll elevate={false}>
                 <AppBar
                     position='fixed'
-                    sx={{ zIndex: (muiTheme) => muiTheme.zIndex.drawer + 1 }}
+                    sx={{
+                        width: {
+                            xs: '100%',
+                            sm: `calc(100% - ${DRAWER_WIDTH}px)`
+                        },
```
seeplusplus, line 47:
```suggestion
                        width: {
                            xs: '100%',
                            ...(isDrawerOpen ? { sm: `calc(100% - ${DRAWER_WIDTH}px)` } : {})
                        },
```

seeplusplus: I believe fixes https://github.com/jellyfin/jellyfin-web/pull/5020#issuecomment-1877924295, testing now.

src/apps/dashboard/AppLayout.tsx
```diff
@@ -19,82 +19,61 @@ interface AppLayoutProps {
...
+                        ml: {
+                            xs: 0,
+                            sm: `${DRAWER_WIDTH}px`
```
seeplusplus, line 50:
```suggestion
                             ...(isDrawerOpen ? { sm: `${DRAWER_WIDTH}px)` } : {})
```

src/apps/dashboard/AppLayout.tsx
thornbill: Good catch... I applied the same change that is used in the experimental app layout to fix.

grafixeyehero (approved):
LGTM, In my opinion, it would be better if the breaking points should be 'lg'

thornbill:
Yeah I agree things definitely look off between the SM and MD breakpoints at least. I think MD looks okay in most places... the item details screen could use some work, but I was planning on tweaking it a bit anyway. :sweat_smile:

I have updated the breakpoints to use MD instead of SM now.

grafixeyehero (approved):

Issue #4908, opened 2023-10-23
Firefox opens "Find in page" when j, k, l is pressed in playing video


Erutuon:
**Describe The Bug**
In Firefox, when you press j, k, or l while playing a video, the desired action, but also opens up a basic "Find in page" dialog at the bottom of the page with the letter in the input box.

**Steps To Reproduce**
1. Open Jellyfin in Firefox
2. Play video
3. Press j, k, or l

**Expected Behavior**
The desired action should happen and "Find in page" should not open.

**Screenshots**
After pressing j (top and bottom of screen cropped out for privacy):
![image](https://github.com/jellyfin/jellyfin-web/assets/5840197/d0b9112a-4113-4739-b9c0-09ce44e998e2)

**System (please complete the following information):**
 - Platform: Windows (operating system on which I'm viewing the video; the server runs Linux, but that probably has no relevance for the web interface)
 - Browser: Firefox
 - Jellyfin Version: 10.8.11

GAntoine:
Just tried it on Firefox (Dev Ed) on macOS and the finder bar doesn't appear, maybe this is a FF-on-Windows only issue?

cpmsmith:
I'm guessing this requires the "search for text when you start typing" setting to be checked to happen?

<img width="511" alt="image" src="https://github.com/jellyfin/jellyfin-web/assets/9437625/b989fc2f-5c3c-4d9f-b1af-1d08edc6f692">

cpmsmith:
Yep, I can confirm it happens on macOS when that setting is enabled. I expect it just needs some judicious `e.preventDefault`s in [`onKeyDown` over here](https://github.com/jellyfin/jellyfin-web/blob/05e2186882c5c0d35bdb6ad80195e379b276c5b7/src/controllers/playback/video/index.js#L1176).

PR #5046, opened 2023-12-06
prevent default in video player on key down

NomadicDeveloper22:
add `e.preventDefault();` when pressing 'j' 'k' or 'l' while playing video. Prevents Firefox from opening "find in page" when "Search for text when you start typing" is enabled in browser settings

**Fix:** #4908


dmitrylyzo:
What about `f`, `m`, `p`, `n`, `0..9`?

I think we can add `preventDefault` for any browser (omit `if (browser.firefox)`).

dmitrylyzo:
> Would it make sense to just add a single `preventDefault` for the whole function? Or at least just before the switch itself. Instead of adding a line for each switch case.

Probably, otherwise browser will start searching on other keys. But this at least breaks `Enter`, `F5`, `F11`, `F12`.

> Or would that have a negative impact on certain cases like for gamepads (don't have one to test myself)

Only if gamepad has bindings in the browser or system. :man_shrugging:

TheRedSpy15:
I opted for specifying which cases need the `e.preventDefault()` (letters and nums) to prevent any issue with other keys

I also rearranged the switch cases so the letters and numbers are after everything else. Just to make it a bit easier when looking at

src/controllers/playback/video/index.js
```diff
@@ -1227,50 +1233,22 @@ export default function (view) {
                     e.stopPropagation();
                 }
                 break;
-            case 'k':
-                playbackManager.playPause(currentPlayer);
-                showOsd(btnPlayPause);
-                break;
             case 'ArrowUp':
-            case 'Up':
```
dmitrylyzo, line 1235: I don't think it should be split up.

src/controllers/playback/video/index.js
```diff
@@ -1227,50 +1233,22 @@ export default function (view) {
...
                 playbackManager.volumeUp(currentPlayer);
                 break;
             case 'ArrowDown':
-            case 'Down':
```
dmitrylyzo, line 1239: same

src/controllers/playback/video/index.js
```diff
@@ -1227,50 +1233,22 @@ export default function (view) {
...
                 playbackManager.volumeDown(currentPlayer);
                 break;
-            case 'l':
```
dmitrylyzo, line 1242: IMO, it is better to keep one command handler - less code to maintain. Also, there is no harm in preventing default action for arrow keys.

src/controllers/playback/video/index.js
```diff
@@ -1227,50 +1233,22 @@ export default function (view) {
...
             case 'ArrowRight':
             case 'Right':
                 playbackManager.fastForward(currentPlayer);
                 showOsd(btnFastForward);
                 break;
-            case 'j':
```
dmitrylyzo, line 1248: same

src/controllers/playback/video/index.js
```diff
@@ -1219,6 +1219,12 @@ export default function (view) {
             case 'Enter':
                 showOsd();
                 break;
+            case 'PageUp':
+                playbackManager.nextChapter(currentPlayer);
+                break;
+            case 'PageDown':
+                playbackManager.previousChapter(currentPlayer);
+                break;
```
dmitrylyzo, line 1227: I'm curious. Why did they climb so high?

TheRedSpy15: No functional reason, other than it was easier on the eyes to organize by which needed the preventDefault and which didn't

src/controllers/playback/video/index.js
TheRedSpy15: Ah that's my mistake. I just added a new commit to fix this and the other similar mistakes

dmitrylyzo (commented):
Right now, I'm torn between preventing the default action for handled commands only (current implementation) and preventing it for all alphanumeric keys.

cpmsmith:
> Right now, I'm torn between preventing the default action for handled commands only (current implementation) and preventing it for all alphanumeric keys.

I definitely think it's more proper to `preventDefault` on handled commands only. We don't know what users have bound to any given key. In particular since it doesn't always check modifier keys, if we unconditionally `preventDefault` we'd be preventing most common keyboard shortcuts.

PR #5048, opened 2023-12-06
Update CollectionType values for new enum

TelepathicWalrus:
Changed the values of collectionType to be the same as the new jellyfin enum CollectionType. This has basically changed a few strings to have capital letters. 
Not sure if this will be 'fixed' on the jellyfin server side, but this seems to work. It does seem to make sense that capital letters are used for the enum on server side. I assume that collectiontype.ts will be able to go when the sdk is updated to include collectiontype as part of the api

Thanks

**Changes**
Changes collectionType to use openapi values 

**Issues**


TelepathicWalrus:
> The casing change should be reverted server side when someone figures out the best way to continue using the enum and return lower case values in the api. The change was unintentional and ended up breaking compatibility with most clients in some way.
> 
> That said we _could_ do the following here to get things working in the meantime:
> 
>     1. Update the typescript sdk to the latest unstable version. This should include the new enum from the api.
> 
>     2. Update any usages of the `CollectionType` typescript enum to use the new one in the sdk.
> 
>     3. Replace any hardcoded string checks for collection type with the new enum values from the sdk.
> 
>     4. Remove the `CollectionType` typescript enum here.
> 
> 
> Then once the api is fixed we would only need to update the sdk again and we would eliminate all these hardcoded strings.

I have updated my sdk to the unstable now but getting errors with another part of api.  
`ERROR in ./src/apps/dashboard/routes/users/profile.tsx:257:66
TS2345: Argument of type '"" | UserPolicy' is not assignable to parameter of type 'UserPolicy'.
  Type 'string' is not assignable to type 'UserPolicy'.
    255 |
    256 |             window.ApiClient.updateUser(user).then(() => (
  > 257 |                 window.ApiClient.updateUserPolicy(user.Id || '', user.Policy || '')
        |                                                                  ^^^^^^^^^^^^^^^^^
    258 |             )).then(() => {
    259 |                 onSaveComplete();
    260 |             }).catch(err => {`
i believe this is becuase user.Policy now requires a string value for PasswordResetProviderId and AuthenticationProviderId. I have  included this in the latest commit too so that it will build. Please let me know what you think. 

Sorry about the force push but for some reason it wouldn't bloody work! :P

thornbill:
Looks like I was too slow and you have got it mostly sorted now. 😅 Force pushes are normally fine until someone starts actively reviewing because then it can be hard to track what has changed.

TelepathicWalrus:
Okay so ive updated the package-lock.json to the latest version of sdk unstable. Hopefully this will build and run :)

TelepathicWalrus:
I'll get there in the end :face_exhaling:

TelepathicWalrus:
@thornbill Should be good to go now if you want to review 👍

thornbill:
Hey @TelepathicWalrus sorry PRs have been sitting for a bit. I'm basically the only person regularly reviewing things here and I got backed up around the holidays. Was there a particular reason why you closed this one?

TelepathicWalrus:
Heya, that's okay, enjoy your holidays!! Life comes first!! Yeah wasn't sure it was needed anymore as the server API spec got changed. If you think it is still worth while I can reopen. Cheers

thornbill:
Ah cool I missed that server PR going through. It would still be good to use the enum values instead of hardcoded strings so I'm in favor of still including this change.

TelepathicWalrus:
No problem. Would make it easier if future changes happen I guess 👍 I'll reopen

package-lock.json
```diff
@@ -3610,9 +3610,9 @@
       "dev": true
     },
     "node_modules/@jellyfin/sdk": {
-      "version": "0.0.0-unstable.202307130502",
-      "resolved": "https://registry.npmjs.org/@jellyfin/sdk/-/sdk-0.0.0-unstable.202307130502.tgz",
-      "integrity": "sha512-1+GXATaJLP5akFnUrpxYzoshLtTPZXJEdy/ozhY1g/DkULlz4LthLTaTJ2qImF0mb8Ayk7LNbh00n4ATk0JycA==",
+      "version": "0.0.0-unstable.202312060501",
```
thornbill, line 3613: This probably needs updated again to have the new (old) values. :upside_down_face:

src/apps/experimental/components/LibraryIcon.tsx
```diff
@@ -11,7 +11,7 @@ import Collections from '@mui/icons-material/Collections';
 import Queue from '@mui/icons-material/Queue';
 import Folder from '@mui/icons-material/Folder';
 import React, { FC } from 'react';
-import { CollectionType } from 'types/collectionType';
+import { CollectionType } from '@jellyfin/sdk/lib/generated-client/';
```
thornbill, line 14: Imports from the SDK should use the full file path. We have had issues in the past with tree-shaking not working and the entire SDK being included in the final JS bundle.

```suggestion
import { CollectionType } from '@jellyfin/sdk/lib/generated-client/models/collection-type';
```

thornbill (changes requested):

package-lock.json
TelepathicWalrus: Ah yes that would probably help, I'm away from PC ATM but I'll update values tomorrow 😀

src/apps/experimental/components/LibraryIcon.tsx
thornbill: I only commented on this one, but the rest should be updated also. 😅

TelepathicWalrus: Yeah sorry, needed PC to update all, I've hopefully done it now and not missed any :P

package-lock.json
TelepathicWalrus: Should be updated now :+1:

TelepathicWalrus:
Seems like build is failing because openapi is not using PascalCase for enum names

thornbill:
I think we will have to update the values here... :confused: it's a bit unfortunate, but seems to be a side effect of the enums being all lower case in the server.

nielsvanvelzen:
The typescript-axios generator does have an `enumPropertyNaming` option that already defaults to PascalCase. But the typescript SDK is using an ancient version of openapi-generator (5.4, latest is 7.2) so perhaps that could be the cause.
So this might be fixable in the SDK.

thornbill:
I was assuming the generator was not capable of splitting words from all lowercase values. It is capitalizing the first letter but not the first letter of each word.

nielsvanvelzen:
Right that precious is lost when generating the spec... my bad

src/components/router/appRouter.js
```diff
@@ -638,7 +639,7 @@ class AppRouter {
         }
 
         if (context !== 'folders' && !itemHelper.isLocalItem(item)) {
-            if (item.CollectionType == 'movies') {
+            if (item.CollectionType == 'Movies') {
```
thornbill, line 642: missed one :wink:

src/components/search/SearchResults.tsx
```diff
@@ -20,11 +21,11 @@ const ensureNonNullItems = (result: BaseItemDtoQueryResult) => ({
...
 });
 
-const isMovies = (collectionType: string) => collectionType === 'movies';
+const isMovies = (collectionType: string) => collectionType === CollectionType.Movies;
 
-const isMusic = (collectionType: string) => collectionType === 'music';
+const isMusic = (collectionType: string) => collectionType === CollectionType.Music;
 
-const isTVShows = (collectionType: string) => collectionType === 'tvshows' || collectionType === 'tv';
+const isTVShows = (collectionType: string) => collectionType === CollectionType.Tvshows;
```
thornbill, line 28: Is `'tv'` no longer present? I'm assuming that's just a legacy thing.

thornbill (changes requested):

src/components/router/appRouter.js
TelepathicWalrus: Awwww damn :P

src/components/search/SearchResults.tsx
TelepathicWalrus: Well i couldn't see tv in the collectiontype enum so assumed it was removed? i can re add it if you reckon it is needed.

thornbill: Probably just a legacy thing... should be fine to drop

thornbill (approved):

PR #5057, opened 2023-12-21
Add custom icon support for custom links

grhallenbeck:
**Changes**

Updates experimental app custom links in the app drawer to support custom icon definitions. MUI Icons supports dynamically defining the icon by icon ID with the `<Icon>icon_name</Icon>` syntax.

Screenshot 2023-12-21 143340


grafixeyehero (approved):

thornbill (approved):
Well that was easier than I expected... and there seems to be enough overlap between the names of the other icon font and the mui icons that it shouldn't be a **major** issue... :rocket:

PR #5059, opened 2023-12-25
Fix queue next / last in session player

rasmuslos:
**Changes**

Fixed a bug in the session player that was causing items to be queued at the wrong position

**Issues**

/


rasmuslos:
I found this while implementing the session player in my app. It is a niche feature, but in my opinion really cool. 

To reproduce the bug you have to:

1. Have two supported separate sessions open at the same time. I used the web client for both, one of them was open in private browsing mode to get a second session.
2. Take control of one of the sessions using the other client (Chromecast icon in the top right hand corner)
3. Start music playback using either client
4. Queue tracks now / later using the context menu using the controlling client (not the one doing the actual playback)
5. Check the queue (only possible using the controlled client)
6. The tracks are not where they are supposed to

I also found the cause for #5012, the session object received from the server does not contain a `Shuffled` property

thornbill (approved):

PR #5065, opened 2023-12-31
Migrate appSettings and userSettings to TypeScript

terite:
In an attempt to keep code changes minimal, I added some FIXME comments to relevant places instead of changing behavior.

**Changes**
* Migrate AppSettings to TypeScript
* Migrate UserSettings to TypeScript
* Remove the `UserSettings.getData` method
* Remove the `UserSettings.importFrom` method

**Issues**

### DIFF
--- a/src/scripts/settings/userSettings.ts
+++ b/src/scripts/settings/userSettings.ts
@@ -1,37 +1,57 @@
 import appSettings from './appSettings';
-import Events from '../../utils/events.ts';
-import { toBoolean } from '../../utils/string.ts';
+import Events from '../../utils/events';
+import { toBoolean } from '../../utils/string';
+import type { ApiClient } from 'jellyfin-apiclient';
+import type { DisplayPreferencesDto, UserConfiguration } from '@jellyfin/sdk/lib/generated-client';
 
-function onSaveTimeout() {
-    const self = this;
-    self.saveTimeout = null;
-    self.currentApiClient.updateDisplayPreferences('usersettings', self.displayPrefs, self.currentUserId, 'emby');
+interface DisplayPreferences extends DisplayPreferencesDto {
+    CustomPrefs: Record<string, string>;
 }
 
-function saveServerPreferences(instance) {
+function onSaveTimeout(this: UserSettings): Promise<void> {
+    this.saveTimeout = null;
+    // FIXME: correctly handle cases where instance properties are not set
+    /* eslint-disable-next-line @typescript-eslint/no-non-null-assertion */
+    return this.currentApiClient!.updateDisplayPreferences('usersettings', this.displayPrefs!, this.currentUserId!, 'emby');
+}
+
+function saveServerPreferences(instance: UserSettings) {
     if (instance.saveTimeout) {
         clearTimeout(instance.saveTimeout);
     }
 
+    // FIXME: handle onSaveTimeout promise rejection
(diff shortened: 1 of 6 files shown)


src/scripts/settings/userSettings.ts
```diff
@@ -421,25 +482,29 @@ export class UserSettings {
...
-    libraryPageSize(val) {
+    libraryPageSize(): number;
+    libraryPageSize(val: number): void;
+    libraryPageSize(val?: number): number | void {
         if (val !== undefined) {
             return this.set('libraryPageSize', val.toString(), false);
         }
 
-        const libraryPageSize = parseInt(this.get('libraryPageSize', false), 10);
+        const libraryPageSize = parseInt(this.get('libraryPageSize', false) || '', 10);
```
thornbill, line 507: Should this be `'0'` like the next instance?

```suggestion
        const libraryPageSize = parseInt(this.get('libraryPageSize', false) || '0', 10);
```

src/scripts/settings/userSettings.ts
```diff
@@ -278,7 +317,9 @@ export class UserSettings {
      * @param {string|undefined} val - Language.
      * @return {string} Language.
      */
-    customCss(val) {
+    customCss(): string | null;
+    customCss(val: string): void;
+    customCss(val?: string): string | null | void {
```
thornbill, line 322: SonarCloud seems to prefer a type alias to be used instead of three unions for `string | null | void`.

src/scripts/settings/userSettings.ts
```diff
@@ -106,13 +114,20 @@ export class UserSettings {
...
-    serverConfig(config) {
-        const apiClient = this.currentApiClient;
+    serverConfig(): Promise<UserConfiguration>;
+    serverConfig(config: UserConfiguration): Promise<void>;
+    serverConfig(config?: UserConfiguration): Promise<UserConfiguration | void> {
+        // FIXME: correctly handle missing currentApiClient and currentUserId
+        /* eslint-disable-next-line @typescript-eslint/no-non-null-assertion */
+        const apiClient = this.currentApiClient!;
+        /* eslint-disable-next-line @typescript-eslint/no-non-null-assertion */
+        const currentUserId = this.currentUserId!;
```
thornbill, line 124: Should we log an error in these cases?

thornbill (changes requested):
Looks good! Just a couple questions below.

src/scripts/settings/userSettings.ts
dmitrylyzo: With `'0'`, we won't get the default `100`.

src/scripts/settings/userSettings.ts
terite: I have no preference. I chose to preserve existing behavior for simplicity.

src/scripts/settings/userSettings.ts
terite: Done in [1503113](https://github.com/jellyfin/jellyfin-web/pull/5065/commits/150311324335d353d89ac100c651905afb6dd1d6)

thornbill:
Hey @terite! Now that 10.9 is finally out the door, could you update this branch and we'll get it merged? Thanks!

danditomaso:
@terite I'm helping out the Jellyfin-web team with reviewing some of these older PR's so we can get on top of the backlog. Wanted to follow up to see if 10.9 could be pulled into this branch, then we can get it merged.

PR #5067, opened 2024-01-01
Migrate loading/loading to TypeScript

terite:
This also simplifies the styling to key children's active states on `.mdlSpinnerActive` instead of their own individual `-active` classes in order to simplify code.

I didn't change any of the imports to import from `...loading/loading.ts` since everything seems to work locally. Is there something I'm missing, or can I omit the `.ts` extension when importing TypeScript from JavaScript?

**Changes**
Migrate loading/loading to TypeScript


thornbill (approved):
:+1: lgtm

PR #5070, opened 2024-01-03
Fix iOS 12 crashing in scroller

thornbill:
**Changes**
I still don't fully understand _why_ this was breaking on iOS 12, but this fixes a regression introduced by https://github.com/jellyfin/jellyfin-web/pull/4714.

The string utility changes ended up _not_ being used, but I don't really want to pull it out to a separate PR since it is mostly just adding tests... :sweat_smile: 

**Issues**
Issue reported in matrix


Bond-009 (approved):
Fixes the issue :+1:

PR #5076, opened 2024-01-05
feat: (preferences) migrate user display settings to react

grhallenbeck:
**Changes**

Updates the Display Settings submenu under User Preferences to be React when using the experimental layout. This interfaces with the legacy settings and layout managers to load and save user display preferences, while rendering the UI and managing the updates using React.

Screenshot 2024-01-04 232500
Screenshot 2024-01-04 232616
Screenshot 2024-01-04 232634


src/apps/experimental/routes/user/display/LibraryPreferences.tsx
```diff
@@ -0,0 +1,101 @@
...
+                    label={globalize.translate('ThemeVideos')}
+                    name='enableLibraryThemeVideos'
+                    onChange={onChange}
+                />
+                <FormHelperText id='display-settings-lib-theme-videos-description'>
+                    {globalize.translate('EnableThemeVideosHelp')}
+                </FormHelperText>
+            </FormControl>
+
+            { !browser.tizen && !browser.web0s && (
```
grhallenbeck, line 85: This logic is migrated over directly from the legacy menu builder, but if anyone has context on why tizen and webOS don't have the "display missing episodes" option I'd love to add a doc comment here with a short explanation.

nielsvanvelzen: IMO we should remove browser checks and make them feature flags. The wrappers for jellyfin-web can already specify which features they do and do not support (https://github.com/jellyfin/jellyfin-android/blob/master/app/src/main/assets/native/nativeshell.js)

src/apps/experimental/routes/user/display/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+export default function UserDisplayPreferences() {
+    useEffect(() => {
+        /**
+         * NOTE: LEGACY SUPPORT
+         *
+         * Hide any legacy views which were shown when this component mounts. The legacy view manager automatically
+         * hides the previous view when a new view is shown. Since this is a React page not managed by the legacy view
+         * manager, we need to manually hide the previous view to prevent it from remaining visible on the screen.
+         */
+        viewManager.hideView();
```
thornbill, line 27: Using the `<Page>` component should handle all the unpleasantness of dealing with the legacy view code for you.

https://github.com/jellyfin/jellyfin-web/blob/master/src/components/Page.tsx

grhallenbeck: Oh nice, I hadn't noticed that. I'll take a look at integrating it this evening.

thornbill: You should be able to remove this effect now since it is handled in the `<Page>`component.

src/apps/experimental/AppLayout.tsx
```diff
@@ -59,40 +60,47 @@ const AppLayout = () => {
                     />
                 </AppBar>
             </ElevationScroll>
+            {/*
+                An extra toolbar is needed here without fixed positioning to take up space at the top of the page and
+                ensure other elements flow properly instead of being covered by the fixed AppBar.
+                https://mui.com/material-ui/react-app-bar/#fixed-placement
+             */}
+            <Toolbar variant='dense' />
```
thornbill, line 68: We have handled this with padding in the style overrides. There are some complications since some pages have the standard toolbar, some have the toolbar + tabs, and some have content rendered behind the toolbar (mainly the item details page).

thornbill (changes requested):
Looks great overall!

We have started (slowly) utilizing react-query to handle server state, so I wonder if we should be using `useMutation` when posting the form changes. (https://tkdodo.eu/blog/mastering-mutations-in-react-query)

This does deviate somewhat from the previous structure we have used by having components/hooks scoped under the route... which is not necessarily a bad thing if these don't end up being used outside of this route. I am curious about @grafixeyehero's opinion on this though, just so we can all be aligned. :slightly_smiling_face:

src/apps/experimental/routes/user/display/LibraryPreferences.tsx
thornbill: Let's just remove it for now. I have no idea why this check exists.

src/apps/experimental/routes/user/display/index.tsx
grhallenbeck: Oof, hit by that post-work brain fog and forgot to remove the effect. 😅

src/apps/experimental/AppLayout.tsx
grhallenbeck: Ah that makes sense if there are places where there are things intentionally rendered in that space. I reached for this because I've frequently run into problems trying to account for space of fixed elements, and it's often easiest to reason about it if the space reservation is controlled by the fixed element or a wrapper, or at least is handled in the same place.

Regardless, I ended up reverting this file back to what's in master since none of the other changes here were necessary with more recent changes that made the drawer static.

src/apps/experimental/routes/user/display/LibraryPreferences.tsx
grhallenbeck: > IMO we should remove browser checks and make them feature flags

Totally onboard with approaching this type of thing rather than explicit browser checks. Is there documentation anywhere or an SDK that clients which wrap the web client use? I started down the road of trying to understand what's there for feature flags currently, but didn't find much for context on the purpose of each flag and what they're actually controlling.

For now I removed this check following @thornbill's advice.

grhallenbeck:
> We have started (slowly) utilizing react-query to handle server state, so I wonder if we should be using useMutation when posting the form changes. (https://tkdodo.eu/blog/mastering-mutations-in-react-query)

I've only dabbled with react-query enough to get my toes wet, and I've only taken a glance at this blog post so far, but it looks like a similar concept to ApolloClient for GraphQL which I've used before. It seems like it might be something good to implement at the user settings level that this form can hook into for sending the updates to the API.

If it's not a priority to make that change here, I might take that as followup task to this PR. It looks like the updates the hooks here are doing go into a debounced API call, so they're all sent to the API as one batched update. Maybe a little bit of a strange way of doing things, but I don't think there's much risk to keep that functionality as-is in the short term since it only results in a single API call.

src/apps/experimental/routes/user/display/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+    const {
+        loading,
+        submitChanges,
+        updateField,
+        values
+    } = useDisplaySettingForm();
+
+    const handleSubmitForm = useCallback((e: React.FormEvent<HTMLFormElement>) => {
+        e.preventDefault();
+        void submitChanges();
```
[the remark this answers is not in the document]
grhallenbeck: Ah good question! It's usually a good practice to handle any promise response from functions either with an `await` or by returning the promise. If we're intentionally not waiting for a promise, we can use the `void` keyword. This also helps communicate intent to future devs by showing the promise response is intentionally ignored rather than a mistake. The project lint rules actually check for this and require that the promise is either awaited, returned, or marked with the `void` keyword.

As for sonarcloud, it seems like it's unable to tell that the `submitChanges` function is async. Maybe something to do with it being an async function wrapped in `useCallback`? I didn't have a chance to get into the nitty gritty of what sonarcube is actually seeing for this line.

src/apps/experimental/routes/user/display/hooks/useServerThemes.ts
```diff
@@ -0,0 +1,30 @@
...
+        async function getServerThemes() {
+            const loadedThemes = await themeManager.getThemes();
+
+            setThemes(loadedThemes ?? []);
+        }
+
+        if (!themes) {
+            void getServerThemes();
+        }
+    });
```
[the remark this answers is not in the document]
grhallenbeck: Oops, yes. Meant to run this once, not every time.

src/apps/experimental/routes/user/display/types.ts
```diff
@@ -0,0 +1,29 @@
+export interface JellyfinTheme {
```
thornbill, line 1: We should reuse the Theme interface in `webConfig`: https://github.com/jellyfin/jellyfin-web/blob/master/src/types/webConfig.ts

grhallenbeck: Ah nice. I did some minor edits there to match and removed this in favor of that one.

src/apps/experimental/routes/user/display/LibraryPreferences.tsx
```diff
@@ -0,0 +1,98 @@
...
+            </FormControl>
+
+            <FormControl fullWidth>
+                <FormControlLabel
+                    aria-describedby='display-settings-lib-backdrops-description'
+                    control={<Checkbox defaultChecked={values.enableLibraryBackdrops} />}
+                    label={globalize.translate('Backdrops')}
+                    name='enableLibraryBackdrops'
+                    onChange={onChange}
+                />
```
grafixeyehero, line 52: defaultChecked: https://mui.com/material-ui/api/checkbox/#checkbox-prop-defaultChecked
```suggestion
                <FormControlLabel
                    aria-describedby='display-settings-lib-backdrops-description'
                    control={
                        <Checkbox
                            checked={values.enableLibraryBackdrops}
                            name='enableLibraryBackdrops'
                            onChange={onChange}/>
                    }
                    label={globalize.translate('Backdrops')}
                />
```

thornbill (approved):

PR #5081, opened 2024-01-06
Migrate preferences menu to react

dkacperski97:
**Changes**
Updates the User Preferences menu to be React when using the experimental layout. I didn't migrate admin and user sections, because they weren't displayed on this page for the experimental layout anyway (they are now located in the AppUserMenu). I'm not sure if ripple should be disabled for buttons, but that how it is on the legacy page.

image

### DIFF
--- a/src/apps/experimental/routes/user/menu/hooks/usePreferencesMenuUser.tsx
+++ b/src/apps/experimental/routes/user/menu/hooks/usePreferencesMenuUser.tsx
@@ -1,0 +1,26 @@
+import { useEffect, useState } from 'react';
+import type { UserDto } from '@jellyfin/sdk/lib/generated-client';
+
+import { useApi } from 'hooks/useApi';
+import { getUserApi } from '@jellyfin/sdk/lib/utils/api/user-api';
+
+interface UsePreferencesMenuUserParams {
+    userId?: string | null;
+}
+
+export function usePreferencesMenuUser({ userId }: UsePreferencesMenuUserParams) {
+    const [user, setUser] = useState<UserDto>();
+    const { api, user: currentUser } = useApi();
+
+    useEffect(() => {
+        void (async () => {
+            const preferencesMenuUser =
+                !userId || userId === currentUser?.Id ?
+                    currentUser :
+                    api && (await getUserApi(api).getUserById({ userId })).data;
+            setUser(preferencesMenuUser);
+        })();
+    }, [api, currentUser, userId]);
+
+    return { user };
+}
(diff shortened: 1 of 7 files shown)


src/apps/experimental/routes/user/menu/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+                                <ListItemIcon>
+                                    <PersonIcon />
+                                </ListItemIcon>
+                                <ListItemText primary={globalize.translate('Profile')} />
+                            </ListItemButton>
+                        </ListItemMui>
+                        <ListItemMui disablePadding divider>
+                            <ListItemButton
+                                component='a'
+                                href={'#/mypreferencesdisplay.html?userId=' + userId}
```
grhallenbeck, line 65: For string building like this, it's generally a better practice to use template literals like `` `#/mypreferencesdisplay.html?userId=${userId}` ``

src/apps/experimental/routes/user/menu/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+    }, []);
+
+    return (
+        <Page
+            id='myPreferencesMenuPage'
+            className='libraryPage userPreferencesPage noSecondaryNavPage'
+            title={globalize.translate('Settings')}
+        >
+            <div className='padded-left padded-right padded-bottom-page padded-top'>
+                <div className='readOnlyContent' style={{ margin: '0 auto' }}>
```
grhallenbeck, line 45: This `readOnlyContent` doesn't seem to actually have anything to do with read only content, and just sets some responsive max widths. I wonder if we could just replace this `div` and class name with a [Box](https://mui.com/system/react-box/) component.

src/apps/experimental/routes/user/menu/index.tsx
dkacperski97: I didn't see the need to use template literals for strings that only have one variable at the end of them, but I changed them to template literals.

src/apps/experimental/routes/user/menu/index.tsx
dkacperski97: I don't think that the Box component would help here. By default it's just div that doesn't have any breakpoints, so we would still need to define max-width somehow, rather via a className, since the `readOnlyContent` class is used in several components.

src/apps/experimental/routes/user/menu/hooks/usePreferencesMenuUser.tsx
grafixeyehero, line 1: @dkacperski97 it's a good place to replace the legacy Api to the new SDK?

dkacperski97:
Sorry for the mess, by mistake I force pushed the wrong branch to mine.

After fixing that, I rebased my changes on top of the current master, I replaced the legacy api with the new sdk in the usePreferencesMenuUser hook, and I added type keyword to two imports.

PR #5092, opened 2024-01-10
Use react-query for UserViews queries

thornbill:
**Changes**
Uses react-query for **all** UserViews queries. This allows us to eliminate duplicate API requests in React _and_ the legacy js using functionality built-in to react-query.

In this case the UserViews API is queried 2-3 times on page load and this change reduces that to a single API request. :rocket: 

**Issues**
N/A

### DIFF
--- a/src/utils/query/queryClient.ts
+++ b/src/utils/query/queryClient.ts
@@ -1,0 +1,3 @@
+import { QueryClient } from '@tanstack/react-query';
+
+export const queryClient = new QueryClient();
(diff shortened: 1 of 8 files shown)


src/utils/query/queryClient.ts
```diff
@@ -0,0 +1,3 @@
+import { QueryClient } from '@tanstack/react-query';
+
+export const queryClient = new QueryClient();
```
thornbill, line 3: I'm not sure if this is the best place for this... I was also considering a subdirectory in `src/hooks` :man_shrugging:

grafixeyehero (approved):
LGTM

Issue #3783, opened 2022-07-30
Subtitle offset text field is clamped to slider min/max values


ambroisie:
**Describe The Bug**

Currently, when one tries to have an offset greater than 30 seconds in either directions, the text field "helpfully" resets it to +/-30 seconds
because those are the bounds of the slider.

**Steps To Reproduce**

1. Open a video with subtitles.
2. Use Subtitle Offset.
3. Write a value outside of [-30, 30] in the text entry field.
4. Press enter to have it be taken into account by the player.
5. Watch your value be overwritten automatically.
 
**Expected Behavior**

If a user enters a value by hand, I would expect the software to respect it, instead of trying to keep it synchronized with the slider which is decidedly not big enough for the value I want to enter.

I think it would make sense that it keeps the value I enter, moves the slider to try and reflect that value: if it is outside of the slider range, just show it clamped at either end. If the user interacts with the slider later on, then synchronize the text field as usual.

**Screenshots**

<details>
<summary>Trying to enter a 45s offset</summary>

![image](https://user-images.githubusercontent.com/12465195/181924726-1fb0f1f2-24ea-4f94-8759-676be52a8f23.png)

</details>

<details>
<summary>Offset is reset to 30s</summary>

![image](https://user-images.githubusercontent.com/12465195/181924735-13e99131-242a-4b14-9ba6-0fbcd09108f4.png)

</details>

**System (please complete the following information):**
 - Platform: Linux (NixOS)
 - Browser: Firefox (100)
 - Jellyfin Version: 10.8.1a

ambroisie:
This is still an issue.

ambroisie:
This is still relevant to me.

iChristGit:
Upping this as its an issue for some of me as well.

SrGesus:
This is a real annoying issue

grepson:
Any updates on that one? It's silly that the offset is force capped to 30 seconds.

PR #5094, opened 2024-01-11
Fix subtitle offset textfield being clamped to 30s

MP430:
Fix for #3783.
Subtitle offset textfield now allows values outside -+30s range.

**Changes**

**Issues**

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -77,6 +77,7 @@
 - [Grady Hallenbeck](https://github.com/grhallenbeck)
 - [DinuD](https://github.com/DinuD)
 - [Kevin Tan (Valius)](https://github.com/valius)
+- [MP430](https://github.com/MP430)
 
 ## Emby Contributors
 
--- a/src/components/subtitlesync/subtitlesync.js
+++ b/src/components/subtitlesync/subtitlesync.js
@@ -47,7 +47,11 @@ function init(instance) {
                 inputOffset = inputOffset[0];
                 inputOffset = parseFloat(inputOffset);
 
+                // synchronize with slider value
                 subtitleSyncSlider.updateOffset(inputOffset);
+
+                subtitleSyncTextField.updateOffset(inputOffset);
+                updateSubtitleOffset(inputOffset);
             } else {
                 this.textContent = (playbackManager.getPlayerSubtitleOffset(player) || 0) + 's';
             }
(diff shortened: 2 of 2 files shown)


dmitrylyzo:
The slider and the field are supposed to be bound to each other.
The value must be within some range, otherwise the user can enter an enormous value.

MP430:
Hi Dmitry,

I can set a limit on the textbox if that'd be better.
How about a limit of +-3600?

Thanks

dmitrylyzo:
> How about a limit of +-3600?

For subtitles that are 1 hour desynced? I don't think such subtitles are worth using.

On TV, the "text field" is not accessible, so you won't be able to enter a raw value.
Why not just increase the range of the slider a bit?
Otherwise, we need to make the "text box" work on the TV.

MP430:
My thinking was just that I wouldn't want to not be able to use subtitles
just because the limit wasn't long enough. I don't see that there's any
reason to unnecessarily limit the text field on other platforms.

I think if you just extend the slider, it'd inevitably end up not being
enough

nielsvanvelzen:
If the subtitles are out of sync by more than 30s you most likely have the wrong subs. I can imagine some offset because of some intro or something but that doesn't really exceed 30s

MP430:
The reason I made this PR is because yesterday I watched a film where I was
unable to use subtitles because there were almost a full minute of logos at
the start. While I could edit them using external software, this isn't what
I really want to do when I'm trying to watch a film.

Feel free to close the pull request if this isn't what you wanted

nielsvanvelzen:
> The reason I made this PR is because yesterday I watched a film where I was unable to use subtitles because there were almost a full minute of logos at the start.

I think increasing the limit could make sense for this usecase. But not making it infinite.

ScribblerCoder:
> If the subtitles are out of sync by more than 30s you most likely have the wrong subs. I can imagine some offset because of some intro or something but that doesn't really exceed 30s

gotta disagree here. I have subtitles out of sync by 60s due to intros/recaps.

thornbill:
The subtitles not matching your media is not a Jellyfin problem.

SrGesus:
> The subtitles not matching your media is not a Jellyfin problem.

If that mattered why would there be an offset feature to begin with?
It's quite silly to forbid a user from having a subtitle offset of more than 30 seconds for no reason

schumi2004:
I would say that a maximum of 60 (maybe 90) is sufficient. Not every is skilled enough to recreate media files therefore it's easier to adjust offset but 30 is in some rare cases not enough.
For example: I have media files that included multiple subs which were extracted by radarr/sonarr but it messed up the timing, now I have subs that need an offset of 32 which I can't set.

PR #5098, opened 2024-01-12
Disable network path field

thornbill:
**Changes**
Disables the network path configuration field

**Issues**
N/A

### DIFF
--- a/src/components/mediaLibraryCreator/mediaLibraryCreator.js
+++ b/src/components/mediaLibraryCreator/mediaLibraryCreator.js
@@ -107,7 +107,6 @@ function onAddButtonClick() {
     import('../directorybrowser/directorybrowser').then(({ default: DirectoryBrowser }) => {
         const picker = new DirectoryBrowser();
         picker.show({
-            enableNetworkSharePath: true,
             callback: function (path, networkSharePath) {
                 if (path) {
                     addMediaLocation(page, path, networkSharePath);
--- a/src/components/mediaLibraryEditor/mediaLibraryEditor.js
+++ b/src/components/mediaLibraryEditor/mediaLibraryEditor.js
@@ -166,7 +166,6 @@ function showDirectoryBrowser(context, originalPath, networkPath) {
     import('../directorybrowser/directorybrowser').then(({ default: DirectoryBrowser }) => {
         const picker = new DirectoryBrowser();
         picker.show({
-            enableNetworkSharePath: true,
             pathReadOnly: originalPath != null,
             path: originalPath,
             networkSharePath: networkPath,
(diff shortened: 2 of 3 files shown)


src/components/directorybrowser/directorybrowser.js
```diff
@@ -124,7 +124,7 @@ function getEditorHtml(options, systemInfo) {
         html += '<div class="results paperList" style="max-height: 200px; overflow-y: auto;"></div>';
     }
     if (options.enableNetworkSharePath) {
-        html += '<div class="inputContainer" style="margin-top:2em;">';
+        html += '<div id="networkPathContainer" class="inputContainer hide" style="margin-top:2em;">';
```
grafixeyehero, line 127: is there a particular reason?
this can be achieved by passing `enableNetworkSharePath: false` on the options

thornbill: lol I didn't even notice that :man_facepalming:

grafixeyehero (approved):

jconnop:
@thornbill 
Is there discussion somewhere on why this was removed, and maybe a recommended alternative for Kodi users who depended on it?

jasonmhite:
Sorry to resurrect an old MR, but I just noticed this change when I was messing with my libraries. I was relying on the network path because performance was a lot better through the Kodi addon and it helps Kodi pick up some other files the normal Jellyfin playback mode does not (like the files the intro skipper plugin generates), but I like having Jellyfin manage my metadata. 

I'm having a hard time finding any context for why this feature is being hidden, is the intent to drop the ability to set a network path entirely long term? Or is this just pushing it out to a hidden setting in the XML? Just wanna know if I eventually need to find a workaround, since removing this breaks the native playback feature in the Kodi addon. Sorry if this is the wrong place to ask and thanks for your work on Jellyfin.

PrplHaz4:
> Sorry to resurrect an old MR, but I just noticed this change when I was messing with my libraries. I was relying on the network path because performance was a lot better through the Kodi addon and it helps Kodi pick up some other files the normal Jellyfin playback mode does not (like the files the intro skipper plugin generates), but I like having Jellyfin manage my metadata.
> 
> I'm having a hard time finding any context for why this feature is being hidden, is the intent to drop the ability to set a network path entirely long term? Or is this just pushing it out to a hidden setting in the XML? Just wanna know if I eventually need to find a workaround, since removing this breaks the native playback feature in the Kodi addon. Sorry if this is the wrong place to ask and thanks for your work on Jellyfin.

More discussion on this is in the below PR.  I believe the short of it is it's causing some challenges and proving hard to maintain as-is.  Unfortunately, it's unlikely that re-adding the feature "properly" will ever get prioritized, and any integration using it would have to be updated to use the fix.

https://github.com/jellyfin/jellyfin/pull/12446

PR #5108, opened 2024-01-13
Move volumeControl, buttonMute and volumeSlider from remotecontrol to separate components

dkacperski97:
**Changes**
Move volumeControl, buttonMute and volumeSlider from remotecontrol to separate components and migrate these components to Typescript to make them easier to convert to React in the future.

`VolumeControl.updatePlayerVolumeState` permanently changes the value of `showMuteButton` and `showVolumeSlider` to false if conditions are met. I copied this logic from remotecontrol, but I'm wondering if `showMuteButton` and `showVolumeSlider` shouldn't be set to true in `updatePlayerVolumeState`, before conditions are checked.


src/components/remotecontrol/sliderContainer.ts
```diff
@@ -0,0 +1,60 @@
+import { playbackManager } from 'components/playback/playbackmanager';
```
dmitrylyzo, line 1: Why not `volumeSlider`? `sliderContainer` is for something more generic.

dkacperski97: Yeah, you're right, I changed the name.

src/types/player.ts
```diff
@@ -0,0 +1,13 @@
+import { PluginType } from './plugin';
+
+export interface Player {
```
thornbill, line 3: This should extend the `Plugin` type. :slightly_smiling_face:

dkacperski97: Fixed

dkacperski97:
I renamed the Player interface to PlayerPlugin as suggested in https://github.com/jellyfin/jellyfin-web/pull/5110#issuecomment-2156216330

src/components/remotecontrol/volumeControl.ts
```diff
@@ -0,0 +1,111 @@
+import ButtonMute from './buttonMute';
+import VolumeSlider from './volumeSlider';
+import { playbackManager } from 'components/playback/playbackmanager';
+import { appHost } from 'components/apphost';
+import { PlayerPlugin } from 'types/player';
```
thornbill, line 5:
```suggestion
import type { PlayerPlugin } from 'types/player';
```

thornbill (changes requested):

dkacperski97:
I applied requested changes

PR #5109, opened 2024-01-14
Refactor Dockerfile

Gauvino:
- Changed CentOS Stream version from 8 to 9
- Changed Debian version from normal to slim version
- Dropped "Yum" for "Dnf"
- Removed unused packages on all Dockerfiles
- Fix building on CentOS/Fedora


Gauvino:
image
Proof of building

Gauvino:
@joshuaboniface does this PR is still needed since jellyfin web would now be built on the jellyfin-sandbox repo ?

PR #5110, opened 2024-01-15
Move RemoteControlSection and subcomponents from remotecontrol to separate components

dkacperski97:
**Changes**
Move RemoteControlSection, SendMessageSection, SendTextSection, NavigationSection and PlaybackCommandButton from remotecontrol to separate components and migrate these components to Typescript to make them easier to convert to React in the future. Continuation of the work on refactoring the remotecontrol component started in https://github.com/jellyfin/jellyfin-web/pull/5108

I have some thoughts about the code that I moved from remotecontrol. 
`updateSupportedCommands` is only called when the player changes, but it depends on `supportedCommands`. Shouldn't this function be also called in `updatePlayerState`? Or maybe `supportedCommands` can't change unless the player changes. But is it then necessary to change the value of `currentPlayerSupportedCommands` in `updatePlayerState`? 
To get to the bottom of this I would have to study the code of playbackManager, but that would take a lot of time, so right now I left this logic as it was. This would also answer the questions I had about `showMuteButton` and `showVolumeSlider` in https://github.com/jellyfin/jellyfin-web/pull/5108


src/types/player.ts
```diff
@@ -0,0 +1,9 @@
+import { Plugin } from './plugin';
+
+export interface Player extends Plugin {
```
thornbill, line 3:
```suggestion
export interface PlayerPlugin extends Plugin {
```

thornbill:
Please rebase this on the latest from master. Thanks!

PR #5136, opened 2024-01-18
Add support for redirecting to the requested page after login

### LINKED ISSUE #3240: When sharing a Link, the Login page Should Redirect to the Content after login
**Describe The Bug**
When I **share a link**, I don't want the recipient to **enter** the **link twice** in the browser:  

If an authenticated session **is not** in progress, the shared link returns to the login page, 
but after logging in you are **redirected to the home**, instead of the desired content.  

This **forces** the user to have to **re-enter** the link received.

**Steps To Reproduce**

1. Go to some content
2. Click on Menu (3 dots icon) > Share > Copy Link
3. Paste the link into an incognito tab
4. You should be redirected to login page
5. Login
6. Redirect to home (problem)

**Expected Behavior**
1. Go to some content
2. Click on Menu (3 dots icon) > Share > Copy Link
3. Paste the link into an incognito tab
4. You should be redirected to login page
5. Login
6. Redirect to the desired content

I think it is appropriate to set a query param when the frontend redirects due to lack of authentication, 
and to modify the login page to redirect to the previously requested path, if present in the query params of the URL.

**System:**
 - Platform: Windows 11
 - Browser: Chrome 96.0.4664.110 (Official Build) (64-bit)
 - Jellyfin Version: 10.7.7

thornbill:
**Changes**
Adds support for redirecting to the requested page after being bounced to the login page and successfully logging in.

Definitely needs to be tested thoroughly :sweat_smile: 

**Issues**
Fixes #3240

### DIFF
--- a/src/components/ConnectionRequired.tsx
+++ b/src/components/ConnectionRequired.tsx
@@ -47,8 +47,9 @@ const ConnectionRequired: FunctionComponent<ConnectionRequiredProps> = ({
                 if (location.pathname === BounceRoutes.Login) {
                     setIsLoading(false);
                 } else {
-                    console.debug('[ConnectionRequired] not logged in, redirecting to login page');
-                    navigate(`${BounceRoutes.Login}?serverid=${connectionResponse.ApiClient.serverId()}`);
+                    console.debug('[ConnectionRequired] not logged in, redirecting to login page', location);
+                    const url = encodeURIComponent(location.pathname + location.search);
+                    navigate(`${BounceRoutes.Login}?serverid=${connectionResponse.ApiClient.serverId()}&url=${url}`);
                 }
                 return;
             case ConnectionState.ServerSelection:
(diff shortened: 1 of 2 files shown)


rdmchr:
I think this implementation is fine since the redirect in done in the front end, so no credentials should be leaked via for example an open redirect attack. But I think this can still be used to phish the user with a setup similar to this:
**User A**: "Hey User B check out this movie: `https://trusted-server.com/login?url=https://evil-server.com/login`"
**User B**: Checks just the hostname and sees a trusted server. After the login they are forwarded to the evil server with maybe an error like "Invalid password" they enter their password again and are sent back to the original server, where they originally logged in.
Should be pretty simple to protect against this by checking if the URL has the same hostname as the current URL. Sorry if there is already a protection against this in place. <3

nielsvanvelzen:
It does not allow redirection to a different host though

thornbill:
It shouldn't but I did not test that...

thornbill:
I did a quick test and verified that providing a full URL will not redirect to another domain

nielsvanvelzen (approved):

PR #5142, opened 2024-01-19
Replace deprecated initEvent()

KasinSparks:
**Changes**

Replaced deprecated initEvent() with recommended event constructor, Event(), as per MDN web docs specification. https://developer.mozilla.org/en-US/docs/Web/API/Event/initEvent#browser_compatibility


src/legacy/keyboardEvent.js
```diff
@@ -17,9 +17,7 @@
         const KeyboardEvent = function (eventName, options) {
             options = options || {};
 
-            const event = document.createEvent('Event');
-
-            event.initEvent(eventName, !!options.bubbles, !!options.cancelable);
+            const event = new Event('Event', { bubbles: !!options.bubbles, cancelable: !!options.cancelable });
```
grafixeyehero, line 20:
```suggestion
            const event = new Event(eventName, { bubbles: !!options.bubbles, cancelable: !!options.cancelable });
```

src/scripts/gamepadtokey.js
```diff
@@ -190,8 +190,7 @@ function raiseEvent(name, key, keyCode) {
         return;
     }
 
-    const event = document.createEvent('Event');
-    event.initEvent(name, true, true);
+    const event = new Event('Event', { bubbles: true, cancelable: true });
```
grafixeyehero, line 193:
```suggestion
    const event = new Event(name, { bubbles: true, cancelable: true });
```

grafixeyehero (changes requested):

src/legacy/keyboardEvent.js
KasinSparks: Oops, sorry about that! I have updated this commit with the two changes you suggested.

grafixeyehero (approved):

PR #5146, opened 2024-01-21
Fix handling of single click on the previous track button

dkacperski97:
**Changes**
Fix handling of single click on the previous track button. 

Currently the player starts playing previous song after the button has been click once only if 5ms (or less 🙃) of the song has passed. `currentPlayer._currentTime` (which returns the time in seconds) has been replaced in https://github.com/jellyfin/jellyfin-web/pull/4891 by `playbackManager.currentTime(currentPlayer)` (which returns the time in milliseconds), but the value to which it is compered has not changed.

Also, currently the previous song only starts playing if we are playing the third or later song in the queue (indexes returned by `getCurrentPlaylistIndex` starts from 0, not 1). This check has been changed in https://github.com/jellyfin/jellyfin-web/pull/4463

### DIFF
--- a/src/components/nowPlayingBar/nowPlayingBar.js
+++ b/src/components/nowPlayingBar/nowPlayingBar.js
@@ -180,7 +180,7 @@ function bindEvents(elem) {
                 // to the previous track, unless we are at the first track so no previous track exists.
                 // currentTime is in msec.
 
-                if (playbackManager.currentTime(currentPlayer) >= 5 || playbackManager.getCurrentPlaylistIndex(currentPlayer) <= 1) {
+                if (playbackManager.currentTime(currentPlayer) >= 5 * 1000 || playbackManager.getCurrentPlaylistIndex(currentPlayer) <= 0) {
                     playbackManager.seekPercent(0, currentPlayer);
                     // This is done automatically by playbackManager, however, setting this here gives instant visual feedback.
                     // TODO: Check why seekPercent doesn't reflect the changes inmmediately, so we can remove this workaround.
--- a/src/components/remotecontrol/remotecontrol.js
+++ b/src/components/remotecontrol/remotecontrol.js
@@ -784,7 +784,7 @@ export default function () {
                     // to the previous track, unless we are at the first track so no previous track exists.
                     // currentTime is in msec.
 
-                    if (playbackManager.currentTime(currentPlayer) >= 5 || playbackManager.getCurrentPlaylistIndex(currentPlayer) <= 1) {
+                    if (playbackManager.currentTime(currentPlayer) >= 5 * 1000 || playbackManager.getCurrentPlaylistIndex(currentPlayer) <= 0) {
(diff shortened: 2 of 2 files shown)


src/components/nowPlayingBar/nowPlayingBar.js
```diff
@@ -180,7 +180,7 @@ function bindEvents(elem) {
                 // to the previous track, unless we are at the first track so no previous track exists.
                 // currentTime is in msec.
 
-                if (playbackManager.currentTime(currentPlayer) >= 5 || playbackManager.getCurrentPlaylistIndex(currentPlayer) <= 1) {
+                if (playbackManager.currentTime(currentPlayer) >= 5 * 1000 || playbackManager.getCurrentPlaylistIndex(currentPlayer) <= 0) {
```
dmitrylyzo, line 183: I just prefer less computation.
```suggestion
                if (playbackManager.currentTime(currentPlayer) >= 5000 || playbackManager.getCurrentPlaylistIndex(currentPlayer) <= 0) {
```
Or maybe introduce a const.

src/components/remotecontrol/remotecontrol.js
```diff
@@ -784,7 +784,7 @@ export default function () {
                     // to the previous track, unless we are at the first track so no previous track exists.
                     // currentTime is in msec.
 
-                    if (playbackManager.currentTime(currentPlayer) >= 5 || playbackManager.getCurrentPlaylistIndex(currentPlayer) <= 1) {
+                    if (playbackManager.currentTime(currentPlayer) >= 5 * 1000 || playbackManager.getCurrentPlaylistIndex(currentPlayer) <= 0) {
```
dmitrylyzo, line 787: same
```suggestion
                    if (playbackManager.currentTime(currentPlayer) >= 5000 || playbackManager.getCurrentPlaylistIndex(currentPlayer) <= 0) {
```

src/components/nowPlayingBar/nowPlayingBar.js
dkacperski97: It will be optimized during build stage when building for production:
image
I think it improves readability a bit, but I can change it if you don't think so.

dmitrylyzo: > It will be optimized during build stage when building for production:
> image

OK then.

dmitrylyzo (approved):

thornbill:
What's up with the last force push here? It doesn't seem like any changes were made?

dkacperski97:
@thornbill Yeah, I didn't make any changes to the last force push, I just synchronized the branch with the master. Sorry, I probably shouldn't have done that.

thornbill:
No worries, but please avoid doing that once a PR is being actively reviewed or approved. GitHub really doesn't track review progress when a branch gets force pushed.

thornbill (approved):

PR #5151, opened 2024-01-23
Add move to top and bottom context menu options

kevgrig:
**Changes**

After adding an item to a playlist, I often want to move it to the top and it's tedious to drag and drop if the playlist is large. This adds 'Move to Top' and 'Move to Bottom' options to a playlist item context menu.

**Issues**

None

### DIFF
--- a/src/components/shortcuts.js
+++ b/src/components/shortcuts.js
@@ -121,7 +121,7 @@ function showContextMenu(card, options = {}) {
                     }
                     index++;
                 }
-                item.PlaylistItems = index;
+                item.PlaylistItemCount = index;
             }
         }
 
--- a/src/components/shortcuts.js
+++ b/src/components/shortcuts.js
@@ -110,6 +110,19 @@ function showContextMenu(card, options = {}) {
         if (playlistId) {
             const elem = dom.parentWithAttribute(card, 'data-playlistitemid');
             item.PlaylistItemId = elem ? elem.getAttribute('data-playlistitemid') : null;
+
+            const itemsContainer = dom.parentWithAttribute(card, 'is', 'emby-itemscontainer');
+            if (itemsContainer) {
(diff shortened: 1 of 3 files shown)


kevgrig:
Bump

kevgrig:
Screenshot_20240512_215815

1hitsong (changes requested):
Testing in Waterfox in Windows 11

Display of options works as expected.

Clicking on Move to Top / Move to Bottom results in an error and items in playlist aren't moved.
image

src/components/shortcuts.js
```diff
@@ -110,6 +110,19 @@ function showContextMenu(card, options = {}) {
...
+            if (itemsContainer) {
+                let index = 0;
+                for (const listItem of itemsContainer.querySelectorAll('.listItem')) {
+                    const playlistItemId = listItem.getAttribute('data-playlistitemid');
+                    if (playlistItemId == item.PlaylistItemId) {
+                        item.PlaylistIndex = index;
+                    }
+                    index++;
+                }
+                item.PlaylistItems = index;
```
1hitsong, line 124: Let's change this property name. Elsewhere in the code we already use `PlaylistItems` to mean the actual objects in a playlist. Here it's used as a count of the items in the playlist.

I suggest something like `PlaylistItemCount`.

1hitsong (dismissed):

src/components/shortcuts.js
kevgrig: Apologies for the delay; moving houses. Pushed a new commit with the changed variable name.

thornbill (approved):
The context menu is getting way too complex, but fixing that will require more substantial changes. This works as expected in my testing and I don't want to block it since the menu is already a mess. :upside_down_face:

PR #5152, opened 2024-01-23
Remove unused elements from item details html

thornbill:
**Changes**
I couldn't find any references to these elements anywhere else in the code, so I'm assuming they are remnants of an old design or some work that was never completed

**Issues**
N/A

### DIFF
--- a/src/controllers/itemDetails/index.html
+++ b/src/controllers/itemDetails/index.html
@@ -90,8 +90,6 @@
             <div class="detailPageContent">
                 <div class="detailPagePrimaryContent padded-right">
                     <div class="detailSection">
-                        <div class="itemMiscInfo nativeName hide"></div>
-
                         <form class="trackSelections hide focuscontainer-x">
                             <div class="selectContainer selectSourceContainer hide trackSelectionFieldContainer flex-shrink-zero">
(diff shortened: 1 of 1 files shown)


src/controllers/itemDetails/index.html
```diff
@@ -90,8 +90,6 @@
             <div class="detailPageContent">
                 <div class="detailPagePrimaryContent padded-right">
                     <div class="detailSection">
-                        <div class="itemMiscInfo nativeName hide"></div>
```
dmitrylyzo, line 93: Introduced in https://github.com/jellyfin/jellyfin-web/pull/459 but never used.

src/controllers/itemDetails/index.html
```diff
@@ -110,8 +108,6 @@
                         <div class="recordingFields hide" style="margin: 0.5em 0 1.5em;"></div>
 
                         <div class="detailSectionContent">
-                            <div class="itemLastPlayed hide"></div>
```
dmitrylyzo, line 113: Introduced in https://github.com/jellyfin/jellyfin-web/pull/459
Removed in https://github.com/jellyfin/jellyfin-web/pull/949

joshuaboniface (approved):

PR #5179, opened 2024-02-01
Reduce the number of events triggered by sessionPlayer

dkacperski97:
**Changes**
Reduce the number of events triggered by sessionPlayer. 

Currently, every time we receive new player data, "statechange" event is triggered which forces the entire component (e.g. nowPlayingBar or remotecontrol) to update. This PR fixes that.

### DIFF
--- a/src/plugins/sessionPlayer/plugin.js
+++ b/src/plugins/sessionPlayer/plugin.js
@@ -3,6 +3,7 @@ import serverNotifications from '../../scripts/serverNotifications';
 import ServerConnections from '../../components/ServerConnections';
 import { PluginType } from '../../types/plugin.ts';
 import Events from '../../utils/events.ts';
+import isEqual from 'lodash-es/isEqual';
 
 function getActivePlayerId() {
     const info = playbackManager.getPlayerInfo();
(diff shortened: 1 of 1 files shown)


src/plugins/sessionPlayer/plugin.js
```diff
@@ -115,21 +116,58 @@ function processUpdatedSessions(instance, sessions, apiClient) {
...
+        }
+    };
+
+    prepareOldData(oldPlayerData.PlayState, newPlayerData.PlayState, 'PositionTicks');
+    if (oldPlayerData.TranscodingInfo) {
+        // TranscodingInfo.CompletionPercentage and TranscodingInfo.Framerate change with time
+        // so it's enough if we only trigger 'timeupdate' event
+        prepareOldData(oldPlayerData.TranscodingInfo, newPlayerData.TranscodingInfo, 'CompletionPercentage');
+        prepareOldData(oldPlayerData.TranscodingInfo, newPlayerData.TranscodingInfo, 'Framerate');
+    }
```
dkacperski97, line 158: Other fields from the "TranscodingInfo" object are not used in places where we watch player events, so maybe it is enough to overwrite the entire "TranscodingInfo" object here? To be safe, I only overwrote fields that I noticed that change over time. I'm not sure when "TranscodingInfo.BitRate" changes, but in the worst case, getChangedEvents will return statechange as it did before.

dkacperski97:
I fixed merge conflict. Sorry it took so long, i have been really busy.

thornbill (approved):

PR #5184, opened 2024-02-06
Add stricter checks for DolbyVision in HEVC

nyanmisaka:
**Changes**
- Add stricter checks for DolbyVision in HEVC

Using `ios` and `osx` version checking is not reliable. Also Apple has adopted AV1 but hasn't enabled it in HLS yet, so we may need to add checks for DolbyVision in AV1 in the future.

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -209,6 +209,14 @@ function supportsDolbyVision(options) {
     );
 }
 
+function canPlayDolbyVisionHevc(videoTestElement) {
+    // Profiles 5/7/8 4k@60fps
+    return !!videoTestElement.canPlayType
+        && (videoTestElement.canPlayType('video/mp4; codecs="dvh1.05.09"').replace(/no/, '')
+        && videoTestElement.canPlayType('video/mp4; codecs="dvh1.07.09"').replace(/no/, '')
+        && videoTestElement.canPlayType('video/mp4; codecs="dvh1.08.09"').replace(/no/, ''));
+}
+
 function getDirectPlayProfileForVideoContainer(container, videoAudioCodecs, videoTestElement, options) {
     let supported = false;
     let profileContainer = container;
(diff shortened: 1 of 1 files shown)


thornbill (approved):

src/scripts/browserDeviceProfile.js
```diff
@@ -930,7 +938,7 @@ export default function (options) {
         av1VideoRangeTypes += '|HLG';
     }
 
-    if (supportsDolbyVision(options)) {
+    if (supportsDolbyVision(options) && canPlayDolbyVisionHevc(videoTestElement)) {
```
dmitrylyzo, line 941: Iirc, `canPlayType` doesn't work on all platforms. So we probably shouldn't call `canPlayDolbyVisionHevc` on Tizen and webOS.

nyanmisaka: IIRC Samsung only seems to be focused on promoting their HDR10 Plus, if they enable DolbyVision in the future we could add it back.

nyanmisaka: I'm wondering how the TV client utilizes this boolean `options.supportsDolbyVision`? Is there a button I didn't notice?

dmitrylyzo: It is filled by native client (app only).
https://github.com/jellyfin/jellyfin-webos/blob/3cf2c835fed6017bbe61ea81ad962ed1827c0d0d/frontend/js/webOS.js#L80
https://github.com/jellyfin/jellyfin-webos/blob/3cf2c835fed6017bbe61ea81ad962ed1827c0d0d/frontend/js/index.js#L20

viktory36: Hey, just chiming in to confirm this works fine on my 2022 webos.

legosoff:
Hello. I checked version 1.5.7 of LG webOS hls.js does not respond to Dolby Vision.

Shadowghost:
This change so currently not available in stable branch and therefore not in the We is  app.
It will be part of 10.9

PR #5190, opened 2024-02-07
bandaid fix for m4b on Safari

felix920506:
Enables m4b to play on iOS and Safari ~~at the cost of force transcoding aac in m4b for everyone (Yes transcoding, not remuxing)~~

The force transcoding for not safari has been addressed. However aac in m4b would still transcode for Safari or iOS browsers, which can be overkill in many cases. This would need to be addressed in the future.

**Changes**
Make web report m4b containers aren't supported

**Issues**
Bandaid fix for https://github.com/jellyfin/jellyfin/issues/6986

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,6 @@
 {
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
+    "source.fixAll.eslint": "explicit"
   },
   "eslint.format.enable": true,
   "editor.formatOnSave": false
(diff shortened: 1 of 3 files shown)


RayBB:
@felix920506 thank you so much for working on this! Will be much appreciated by many I'm sure :)
Good timing as I was just posting on [the forum](https://forum.jellyfin.org/t-m4b-aac-audiobook-support-with-jellyfin-web-on-safari-for-ios) about this recently.

One naive question for me as I'm learning about how jellyfin works, is there a way to test actually playing the m4b and it fails and then the player reports to the server to request a transcoding?

thornbill:
If Safari is the issue, why not just wrap this in a `if (!browser.safari) { ... }`?

felix920506:
A better fix has been identified, will close this one in favor of the alternative

felix920506:
@RayBB Yes

PR #4041, opened 2022-10-13
Add ShakaPlayer and AV1 HLS support

nyanmisaka:
**Changes**
- Add ShakaPlayer for non-native fMP4
- Enable AV1 support in HLS playback

**Issues**
- Fixes #3982

The Hls.js used on non-Apple devices cannot handle fmp4 seeking correctly and it lacks support for many codecs.
Shaka also supports Dash and AV1 transcoding/remuxing for future use.

 Player/Feature | TS   | fMP4 seek  | H.264 | HEVC | AV1 | AAC | MP3 | AC3/EAC3 | FLAC | OPUS
--------------- | ---- | ---------- | ----- | ---- | --- | --- | --- | -------- | ---- | ----
Shaka           | ✔️   | ✔️        | ✔️    | ✔️   | ✔️  | ✔️  | ✔️  | ✔️       | ✔️   | ✔️
Hls.js          | ✔️   | ❌        | ✔️    | ✔️   | ❌  | ✔️  | ✔️  | ❌       | ❌   | ❌

image

PR #5200, opened 2024-02-12
Add trickplay support

nicknsy:
This is the associated web code required to get trickplay (https://github.com/jellyfin/jellyfin/pull/9554) working on the frontend.

The main additions are the updateBubbleHtml function in the video player, as well as all the corresponding trickplay configuration & library options.


ferferga:
@nicknsy Given it was mentioned in the server PR, did you check if it works with hls.js alone? Because having them in the hls stream is the correct approach imo. 

The endpoint it's still useful ofc for other clients or other users (i.e posters in resume section maybe?)

nicknsy:
> Given it was mentioned in the server PR description (but not during the review anymore) what's the state of them being included in the hls stream? Did you check if it works with hls.js alone? Because having them in the hls stream is the most correct approach imo.
> 
> The endpoint it's still useful ofc for other clients or other users (i.e posters in resume section maybe?)
> 
> _Reviewing this on phone, so sorry if I missed something_

Doesn't look like hls.js supports trickplay natively: https://github.com/video-dev/hls.js/issues/5392. Because of that, I haven't personally been able to test if the HLS playlist file generated by the server actually works or not (it should though). I believe Roku should support HLS trickplay playlists natively so it might be trivial to implement / test on there?

nyanmisaka:
Google's [Shaka-player](https://github.com/shaka-project/shaka-player#hls-features) supports `EXT-X-IMAGE-STREAM-INF` for [thumbnails](https://github.com/shaka-project/shaka-player/blob/83c02b8a5cfe48e2c028efb3aca6d3fd9026b062/test/hls/hls_parser_unit.js#L1945). They have a [demo page](https://index-dot-shaka-player-demo.appspot.com/#) for easy debugging.

nyanmisaka:
> > Google's [Shaka-player](https://github.com/shaka-project/shaka-player#hls-features) supports `EXT-X-IMAGE-STREAM-INF` for [thumbnails](https://github.com/shaka-project/shaka-player/blob/83c02b8a5cfe48e2c028efb3aca6d3fd9026b062/test/hls/hls_parser_unit.js#L1945). They have a [demo page](https://index-dot-shaka-player-demo.appspot.com/#) for easy debugging.
> 
> Thanks, looks like there was an issue with the playlist not doing relative paths correctly, however after a one-line fix it works great with Shaka. Is Shaka-player something currently supported by jellyfin-web?

Great! I've used shaka in the past as a replacement for our hls.js, but when hls.js fixed their code shaka was unnecessary. If you're interested, you can pick it up and add it as an alternate option. https://github.com/jellyfin/jellyfin-web/pull/4041

ferferga:
@nicknsy Just to clarify, so server-side's hls implementation is working right, correct?

nicknsy:
> @nicknsy Just to clarify, so server-side's hls implementation is working right, correct? 

Once https://github.com/jellyfin/jellyfin/pull/11000 is merged the hls implementation will be working, as tested with Shaka-player.

src/apps/dashboard/routes/_redirects.ts
```diff
@@ -36,5 +36,6 @@ export const REDIRECTS: Redirect[] = [
     { from: 'usernew.html', to: '/dashboard/users/add' },
     { from: 'userparentalcontrol.html', to: '/dashboard/users/parentalcontrol' },
     { from: 'userpassword.html', to: '/dashboard/users/password' },
-    { from: 'userprofiles.html', to: '/dashboard/users' }
+    { from: 'userprofiles.html', to: '/dashboard/users' },
+    { from: 'trickplayconfiguration.html', to: '/dashboard/playback/trickplay' }
```
thornbill, line 40: This is only for existing paths that have changed so no need to add it for a new page.

src/controllers/playback/video/index.js
```diff
@@ -1681,6 +1756,43 @@ export default function (view) {
...
+        ticks /= 100;
+        ticks *= value;
+
+        if (item?.Trickplay) {
+            const mediaSourceId = currentPlayer?.streamInfo?.mediaSource?.Id;
+            const trickplayResolutions = item.Trickplay[mediaSourceId];
+
+            if (!trickplayResolutions) return false;
+
+            // Prefer highest resolution <= 20% of total screen resolution width
```
nicknsy, line 1773: Only do this once on player load and not every update

src/strings/en-us.json
```diff
@@ -1625,5 +1625,38 @@
...
+    "LabelJpegQuality": "JPEG Quality",
+    "LabelJpegQualityHelp": "The JPEG compression quality for trickplay images.",
+    "LabelQscale": "Qscale",
+    "LabelQscaleHelp": "The quality scale of images output by ffmpeg, with 2 being the highest quality and 31 being the lowest.",
+    "LabelTrickplayThreads": "FFmpeg Threads",
+    "LabelTrickplayThreadsHelp": "The number of threads to pass to the '-threads' argument of ffmpeg.",
+    "OptionExtractTrickplayImage": "Enable trickplay image extraction",
+    "ExtractTrickplayImagesHelp": "Trickplay images are similar to chapter images, except they span the entire length of the content and are used to show a preview when scrubbing through videos.",
+    "LabelExtractTrickplayDuringLibraryScan": "Extract trickplay images during the library scan",
+    "LabelExtractTrickplayDuringLibraryScanHelp": "Generate trickplay images when videos are imported during the library scan. Otherwise, they will be extracted during the chapter images scheduled task. If generation is set to non-blocking this will not affect the time a library scan takes to complete."
```
thornbill, line 1661: It seems there is a dedicated scheduled task for extracting trickplay images. Should this be referencing the chapter image task?

src/controllers/dashboard/trickplay.html
```diff
@@ -0,0 +1,68 @@
...
+                </div>
+                <div class="inputContainer">
+                    <input is="emby-input" type="number" id="txtProcessThreads" pattern="[0-9]*" required="" label="${LabelTrickplayThreads}">
+                    <div class="fieldDescription"></div>
+                </div>
+                <div><button is="emby-button" type="submit" class="raised button-submit block"><span>${Save}</span></button></div>
+            </form>
+        </div>
+    </div>
+</div>
```
thornbill, line 68: Missing eof newline

src/apps/dashboard/routes/_legacyRoutes.ts
```diff
@@ -145,5 +145,11 @@ export const LEGACY_ADMIN_ROUTES: LegacyRoute[] = [
             view: 'dashboard/streaming.html',
             controller: 'dashboard/streaming'
         }
+    }, {
+        path: 'playback/trickplay',
```
thornbill, line 149: We really shouldn't be adding new pages using the legacy view/controller functionality. Any new code should be preferably written in typescript + react, but given we are wanting this in for the imminent 10.9 release I think we can let it slide.

thornbill (changes requested):
Looks good overall. I don't have any trickplay images generated yet to test it in action though. I would have expected some overlap between this and the existing chapter bubble functionality, but I don't see any evidence of that in these changes?

src/strings/en-us.json
ebkalderon: I'm not the pull request author (apologies for the drive-by comment; I'm just excited for this 10.9 feature), but I believe this and the corresponding `jellyfin/jellyfin` PR that was previously merged are based on the [Jellyscrub](https://github.com/nicknsy/jellyscrub) plugin. This plug-in features a background task for generating BIF previews, which is separate from the existing chapter image preview task. I'm not sure if this is what is being referenced by this particular string?

nicknsy: > It seems there is a dedicated scheduled task for extracting trickplay images. Should this be referencing the chapter image task?

These string specifically are for the library settings that allow you to enable/disable trickplay generation and generation on scan per library. This is very similar to chapter images however trickplay images take much longer to generate and are functionally a bit different since they have no concept of segments and always span the length of the media (whereas chapters have a start and stop time).

On the server, the [trickplay task](https://github.com/jellyfin/jellyfin/blob/master/MediaBrowser.Providers/Trickplay/TrickplayImagesTask.cs) is also separate from chapter images so users can choose to enable/disable whichever they prefer.

src/apps/dashboard/routes/_legacyRoutes.ts
nicknsy: > We really shouldn't be adding new pages using the legacy view/controller functionality. Any new code should be preferably written in typescript + react, but given we are wanting this in for the imminent 10.9 release I think we can let it slide.

Oh, that's my bad. I wrote this a while ago to go alongside the server PR so I never saw the pinned issue about using React on jellyfin-web. I've never used React but I see some of the user dashboard pages have already been converted so I'll take a shot at converting this too.

nicknsy:
> Looks good overall. I don't have any trickplay images generated yet to test it in action though. I would have expected some overlap between this and the existing chapter bubble functionality, but I don't see any evidence of that in these changes?

I think the main place there would be overlap would be the getBubbleHtml function however it was problematic for trickplay images (really it is even for chapter images) since it deletes the image container from the DOM on every update causing bad pop-in/out effects. I think ideally the chapter images could also be migrated to updateBubbleHtml which is the function I made for trickplay but initially I didn't want to touch that and potentially break something.

Besides that, trickplay generation takes a lot longer and has more options for things like HW accel and tonemapping to the point where it needs a separate config page from chapter images IMO.

thornbill:
I was mainly questioning the potential overlap of the "bubble" code, so that explanation makes sense. :+1:

src/apps/dashboard/routes/playback/trickplay.tsx
```diff
@@ -0,0 +1,291 @@
+import type { ProcessPriorityClass, ServerConfiguration, TrickplayScanBehavior } from '@jellyfin/sdk/lib/generated-client';
+import React, { FunctionComponent, useCallback, useEffect, useRef } from 'react';
```
thornbill, line 2:
```suggestion
import React, { type FunctionComponent, useCallback, useEffect, useRef } from 'react';
```

src/components/dashboard/playback/trickplay/LinkTrickplayAcceleration.tsx
```diff
@@ -0,0 +1,33 @@
+import React, { FunctionComponent } from 'react';
```
thornbill, line 1:
```suggestion
import React, { type FunctionComponent } from 'react';
```

src/components/dashboard/playback/trickplay/LinkTrickplayAcceleration.tsx
```diff
@@ -0,0 +1,33 @@
...
+    </a>`
+});
+
+const LinkTrickplayAcceleration: FunctionComponent<IProps> = ({ className, title, href }: IProps) => {
+    return (
+        <div
+            dangerouslySetInnerHTML={createLinkElement({
+                className: className,
+                title: globalize.translate(title),
+                href: href
```
thornbill, line 27:
```suggestion
                className,
                title: globalize.translate(title),
                href
```

src/controllers/playback/video/index.js
```diff
@@ -1681,6 +1775,27 @@ export default function (view) {
         }
     });
 
+    nowPlayingPositionSlider.updateBubbleHtml = function(bubble, value) {
+        showOsd();
+
+        const item = currentItem;
+        let ticks = currentRuntimeTicks;
+        ticks /= 100;
+        ticks *= value;
```
thornbill, line 1784: I don't see any reason to break this into separate lines.

```suggestion
        const ticks = currentRuntimeTicks * value / 100;
```

src/strings/en-us.json
```diff
@@ -1625,5 +1625,38 @@
     "MachineTranslated": "Machine Translated",
     "ForeignPartsOnly": "Forced/Foreign parts only",
     "HearingImpairedShort": "HI/SDH",
-    "LabelIsHearingImpaired": "For hearing impaired (SDH)"
+    "LabelIsHearingImpaired": "For hearing impaired (SDH)",
```
thornbill, line 1628: Just for future reference, having the translations in alphabetical order is preferred since it can make them easier to find and helps avoid conflicts between different PRs. It's fine to leave as is here though. :slightly_smiling_face:

thornbill (changes requested):
I just have a few minor suggestions. This is looking great though! I was able to generate trickplay images and test in the player now also.

src/strings/en-us.json
nicknsy: I had originally sorted the strings into the file using "Sort Lines" in VS Code but I believe there were a bunch of other strings that weren't properly sorted beforehand so it ended up sorting those too and causing frequent merge conflicts for me.

src/controllers/playback/video/index.js
```diff
@@ -146,6 +146,24 @@ export default function (view) {
             btnUserRating.classList.add('hide');
             btnUserRating.setItem(null);
         }
+
+        // Update trickplay data
+        trickplayResolution = null;
+
+        const mediaSourceId = currentPlayer.streamInfo.mediaSource.Id;
+        let trickplayResolutions;
+        if (item.Trickplay && (trickplayResolutions = item.Trickplay[mediaSourceId])) {
```
thornbill, line 155:
```suggestion
        let trickplayResolutions = item.Trickplay?.[mediaSourceId];
        if (trickplayResolutions) {
```

src/controllers/playback/video/index.js
```diff
@@ -146,6 +146,24 @@ export default function (view) {
...
+            // Prefer highest resolution <= 20% of total screen resolution width
+            let bestWidth;
+            const maxWidth = window.screen.width * window.devicePixelRatio * 0.2;
+            for (const [, info] of Object.entries(trickplayResolutions)) {
+                if (!bestWidth
+                        || (info.Width < bestWidth && bestWidth > maxWidth) // Objects not guaranteed to be sorted in any order, first width might be > maxWidth.
+                        || (info.Width > bestWidth && info.Width <= maxWidth)) bestWidth = info.Width;
```
thornbill, line 162:
```suggestion
                        || (info.Width > bestWidth && info.Width <= maxWidth)) {
                    bestWidth = info.Width;
                }
```

src/apps/dashboard/routes/playback/trickplay.tsx
```diff
@@ -0,0 +1,291 @@
...
+            options.ProcessPriority = (page.querySelector('#selectProcessPriority') as HTMLSelectElement).value as ProcessPriorityClass;
+            options.Interval = Math.max(1, parseInt((page.querySelector('#txtInterval') as HTMLInputElement).value || '10000', 10));
+            options.WidthResolutions = (page.querySelector('#txtWidthResolutions') as HTMLInputElement).value.replace(' ', '').split(',').map(Number);
+            options.TileWidth = Math.max(1, parseInt((page.querySelector('#txtTileWidth') as HTMLInputElement).value || '10', 10));
+            options.TileHeight = Math.max(1, parseInt((page.querySelector('#txtTileHeight') as HTMLInputElement).value || '10', 10));
+            options.Qscale = Math.min(31, parseInt((page.querySelector('#txtQscale') as HTMLInputElement).value || '4', 10));
+            options.JpegQuality = Math.min(100, parseInt((page.querySelector('#txtJpegQuality') as HTMLInputElement).value || '90', 10));
+            options.ProcessThreads = parseInt((page.querySelector('#txtProcessThreads') as HTMLInputElement).value || '1', 10);
+
+            window.ApiClient.updateServerConfiguration(config).then(() => {
```
thornbill, line 81:
```suggestion
            const apiClient = ServerConnections.currentApiClient();
            if (!apiClient) {
                console.error('[PlaybackTrickplay] No current apiclient instance');
                return;
            }

            apiClient.updateServerConfiguration(config).then(() => {
```

src/apps/dashboard/routes/playback/trickplay.tsx
```diff
@@ -0,0 +1,291 @@
...
+                onSaveComplete();
+            }).catch(err => {
+                console.error('[playbacktrickplay] failed to update config', err);
+            });
+        };
+
+        const onSubmit = (e: Event) => {
+            loading.show();
+
+            window.ApiClient.getServerConfiguration().then(function (config) {
```
thornbill, line 91:
```suggestion
            const apiClient = ServerConnections.currentApiClient();
            if (!apiClient) {
                console.error('[PlaybackTrickplay] No current apiclient instance');
                return;
            }

            apiClient.getServerConfiguration().then(function (config) {
```

src/apps/dashboard/routes/playback/trickplay.tsx
```diff
@@ -0,0 +1,291 @@
...
+        (page.querySelector('#txtJpegQuality') as HTMLInputElement).value = options.JpegQuality;
+        (page.querySelector('#txtProcessThreads') as HTMLInputElement).value = options.ProcessThreads;
+
+        loading.hide();
+    }, []);
+
+    const loadData = useCallback(() => {
+        loading.show();
+
+        window.ApiClient.getServerConfiguration().then(function (config) {
```
thornbill, line 49:
```suggestion
        ServerConnections.currentApiClient()?.getServerConfiguration().then(function (config) {
```

thornbill (changes requested):
Should be good after this last round of changes I think. :rocket:

thornbill (approved):

PR #5203, opened 2024-02-13
Add support for searching via a url parameter

thornbill:
**Changes**

* Adds support for searching via a query url parameter
* Syncs search updates with url parameter
* Adds a react implementation of emby-input to correct webcomponent + react lifecycle weirdness
* Makes tags links that open search (I'm not a huge fan of this, but probably fine as a short term solution :upside_down_face:)

Supersedes https://github.com/jellyfin/jellyfin-web/pull/4889

**Issues**
N/A

### DIFF
--- a/src/apps/stable/routes/search.tsx
+++ b/src/apps/stable/routes/search.tsx
@@ -19,11 +19,12 @@ function usePrevious(value: string) {
 
 const Search: FunctionComponent = () => {
     const [ searchParams ] = useSearchParams();
-    const [ query, setQuery ] = useState<string>(searchParams.get('query') || '');
+    const urlQuery = searchParams.get('query') || '';
+    const [ query, setQuery ] = useState<string>(urlQuery);
     const prevQuery = usePrevious(query);
 
-    if (query == prevQuery && searchParams.get('query') != query) {
-        setQuery(searchParams.get('query') || '');
+    if (query == prevQuery && urlQuery != query) {
+        setQuery(urlQuery);
     }
 
     useEffect(() => {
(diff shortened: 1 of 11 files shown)


src/apps/stable/routes/search.tsx
```diff
@@ -1,39 +1,70 @@
...
-const Search: FunctionComponent = () => {
-    const [ query, setQuery ] = useState<string>();
-    const [ searchParams ] = useSearchParams();
+const COLLECTION_TYPE_PARAM = 'collectionType';
+const PARENT_ID_PARAM = 'parentId';
+const QUERY_PARAM = 'query';
+const SERVER_ID_PARAM = 'serverId';
+
+const Search: FC = () => {
+    const navigate = useNavigate();
```
grafixeyehero, line 18: no need this can be removed?

src/components/search/SearchResults.tsx
```diff
@@ -335,8 +339,8 @@ const SearchResults: FunctionComponent<SearchResultsProps> = ({ serverId = windo
                         cardOptions={{ coverImage: true }}
                     />
 
-                    {allEmpty && query && !isLoading && (
-                        <div className='sorry-text'>{globalize.translate('SearchResultsEmpty', query)}</div>
+                    {allEmpty && debouncedQuery && !isLoading && (
+                        <div className='sorry-text'>{globalize.translate('SearchResultsEmpty', debouncedQuery)}</div>
```
grafixeyehero, line 343: displaying no result it bit off
```suggestion
                        <div className='noItemsMessage centerMessage'>{globalize.translate('SearchResultsEmpty', debouncedQuery)}</div>
```

grafixeyehero (approved):

PR #5207, opened 2024-02-15
Change webOS version needed for audioTracks to 4.0

### LINKED ISSUE #4661: HDR content with multiple audio sources won't switch them at all
When HDR content with multiple audio sources is being steamed, and Jellyfin detects that the device can deal with HDR content without needing to deal with tone mapping, you are not able to switch audio sources. It will show as if it is using another, but it will only actually use the default one.

1. Stream multiple audio sourced HDR content on a device that deals with HDR, usually most modern TVs will work, if it shows an HDR icon when it starts playing, that qualifies.
2. Try switching the audio source away from the default 
3. The audio should change and it doesn't.

I would expect for that to not happen and the other audio track to be used.

 - Platform: LG webOS, both in Jellyfin App and on the browser
 - Browser: Native
 - Jellyfin Version: 10.8.10

I would like to state I am not *sure* if the issue is by this cause exactly, that's just my theory because other content works just fine when switching audio.  If anyone has a similar configuration with content that is HDR and a device that supports it as well, and also with multiple audio sources to see if the issue also happens to you.

### LINKED ISSUE #4729: canPlaySecondaryAudio return false for webos 4 and DTS transcoding issue
**Describe The Bug**
I have a LG C9 TV, running webos 4 and i'm using the jellyfin-webos app. 
Webos 4 support DTS and HTMLMediaElement audioTracks property 
If I try to play a video that has a DTS audio track, the file get transcoded (remuxed if i play a non DTS track, audio transcoding DTS to AC3 if I select the DTS track)

**Steps To Reproduce**
1. Play a movie with more than one audio track of which at least one is DTS
2. Notice the movie get remuxed or transcoded ` ( PlayMethod=Transcode, TranscodeReason=SecondaryAudioNotSupported )`

**Expected Behavior**
It should play the movie without remuxing or transcoding

 - Platform: Webos
 - Browser: Webos app
 - Jellyfin Version: 10.8.10
 - 
**Additional Context**
The issue can be fixed by changing `browser.web0sVersion >= 5.0` to `browser.web0sVersion >= 4.0` in https://github.com/jellyfin/jellyfin-web/blob/master/src/scripts/browserDeviceProfile.js#L358C9-L358C9
with that line changed the movie is correcly played directly.

Ignoring the `SecondaryAudioNotSupported` issue, I would still have expected that when playing DTS audio (supported in webos4) it would simply remux and not transcode the audio track.

The issue seems to be caused by the request the app makes  `media:/videos/b08076b6-68eb-2d41-280a-81a9b6d181ca/master.m3u8?MediaSourceId=b08076b668eb2d41280a81a9b6d181ca&VideoCodec=h264,hevc,h264&AudioCodec=aac,mp3,ac3,eac3&AudioStreamIndex=3&VideoBitrate=139360000&AudioBitrate=640000&MaxFramerate=23.976025&api_key` 
I would expect `AudioCodec=aac,mp3,ac3,eac` to include DTS)

andrasmaroy:
**Changes**

Implementation of the proposed solution in #4729. Tested on an LG 55OLEDB9 (WebOS4.5, 05.40.10 FW), running Jellyfin 10.8.11 patched with this change. Tested with an mp4 file with an aac and an eac3 audio track and with an mkv with 2 eac3 tracks. Both files direct played with both audio tracks.

This also solved the issue described in jellyfin/jellyfin-webos#198 where Dolby Vision content failed to trigger DoVi on the second audio track because it was remuxed server-side because of this issue.

**Issues**

Fixes #4729 
Fixes #4661 
Fixes jellyfin/jellyfin-webos#198

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -366,8 +366,8 @@ export function canPlaySecondaryAudio(videoTestElement) {
         && !browser.firefox
         // It seems to work on Tizen 5.5+ (2020, Chrome 69+). See https://developer.tizen.org/forums/web-application-development/video-tag-not-work-audiotracks
         && (browser.tizenVersion >= 5.5 || !browser.tizen)
-        // Assume webOS 5+ (2020, Chrome 68+) supports secondary audio like Tizen 5.5+
-        && (browser.web0sVersion >= 5.0 || !browser.web0sVersion);
+        // Assume webOS 4+ (2018, Chrome 68+) supports secondary audio like Tizen 5.5+
+        && (browser.web0sVersion >= 4.0 || !browser.web0sVersion);
 }
 
     export default function (options) {
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -53,6 +53,7 @@
  - [Matthew Jones](https://github.com/matthew-jones-uk)
  - [taku0](https://github.com/taku0)
  - [Peter Spenler](https://github.com/peterspenler)
+ - [András Maróy](https://github.com/andrasmaroy)
 
 # Emby Contributors
 


src/scripts/browserDeviceProfile.js
```diff
@@ -366,8 +366,8 @@ export function canPlaySecondaryAudio(videoTestElement) {
         && !browser.firefox
         // It seems to work on Tizen 5.5+ (2020, Chrome 69+). See https://developer.tizen.org/forums/web-application-development/video-tag-not-work-audiotracks
         && (browser.tizenVersion >= 5.5 || !browser.tizen)
-        // Assume webOS 5+ (2020, Chrome 68+) supports secondary audio like Tizen 5.5+
-        && (browser.web0sVersion >= 5.0 || !browser.web0sVersion);
+        // Assume webOS 4+ (2018, Chrome 68+) supports secondary audio like Tizen 5.5+
```
dmitrylyzo, line 369: This comment is no longer relevant since you have tested audio switching.
```suggestion
```

dmitrylyzo (commented):
I can't test this in webOS 4 emulator - audio only works in music :confused: 

I don't think we are going to make another 10.8 release. Since you didn't create a branch for PR, it will look weird after rebasing. Could you create a branch from `master` and recreate PR?

dmitrylyzo:
> Sure, just to be clear: new PR to master or to a new feature branch?

`new branch with your changes from master` to `master`.

andrasmaroy:
Closing in favor of #5214.

PR #5210, opened 2024-02-17
Add Favorites View

grafixeyehero:
**Changes**
- Add Favorites view


thornbill (commented):
Why add favorites as a dedicated tab when it is available as a filter?

grafixeyehero:
> Why add favorites as a dedicated tab when it is available as a filter?

1. Having a dedicated tab for Favorites makes it quicker and more convenient to access these items directly. Users don't have to apply a filter every time they want to see their favorite items; they can simply click on the Favorites tab.
2. In previous music implementation the favorites were under a "Suggestions" tab wasn't ideal. so, I moved it to a dedicate tab

PR #5219, opened 2024-02-21
Convert `toast` and `confirm` to TypeScript

StableCrimson:
**Changes**
`toast` converted to TypeScript
`confirm` converted to TypeScript

**Issues**
None


src/components/confirm/confirm.ts
```diff
@@ -12,15 +27,15 @@ function useNativeConfirm() {
...
     if (typeof options === 'string') {
         options = {
             title: '',
             text: options
-        };
+        } as ConfirmOptions;
     }
 
-    const text = (options.text || '').replaceAll('<br/>', '\n');
+    const text = (options.text ?? '').replace('<br/>', '\n');
```
dmitrylyzo, line 38: No need to change `||` to `??` unless necessary. `??` produces more code when transpiled (_we need to support older browsers that don't support `??`_).

StableCrimson: Fixed

src/components/confirm/confirm.ts
```diff
@@ -12,15 +27,15 @@ function useNativeConfirm() {
...
     if (typeof options === 'string') {
         options = {
             title: '',
             text: options
-        };
+        } as ConfirmOptions;
     }
 
-    const text = (options.text || '').replaceAll('<br/>', '\n');
+    const text = (options.text || '').replace('<br/>', '\n');
```
thornbill, line 38: This doesn’t seem like it should be changed?

thornbill: @StableCrimson could you take a look at this question when you have a minute? Thanks!

StableCrimson: Sorry for the late response. Is there ever a case in which we will be replacing more than one occurrence of <br />? I can undo if necessary

thornbill: I believe there could be.

StableCrimson: Updated!

thornbill:
Whoops looks like the tsconfig needs updated to support using ES2021 features to use `replaceAll`.

src/components/confirm/confirm.ts
```diff
@@ -12,15 +27,15 @@ function useNativeConfirm() {
...
     if (typeof options === 'string') {
         options = {
             title: '',
             text: options
-        };
+        } as ConfirmOptions;
     }
 
-    const text = (options.text || '').replaceAll('<br/>', '\n');
+    const text = (options.text ?? '').replace(/<br\/>/g, '\n');
```
StableCrimson, line 38: Replace by default will only replace one instance, but you can use a regex pattern to do a global replace.

StableCrimson:
> Whoops looks like the tsconfig needs updated to support using ES2021 features to use `replaceAll`.

Gotcha. TS replace can be used so long as we use a regex pattern instead of a string pattern.

src/components/confirm/confirm.ts
```diff
@@ -31,27 +46,24 @@ async function nativeConfirm(options) {
...
-    } else {
-        options = text;
     }
 
-    const items = [];
+    const items: OptionItem[] = [];
 
     items.push({
-        name: options.cancelText || globalize.translate('ButtonCancel'),
+        name: options.cancelText ?? globalize.translate('ButtonCancel'),
```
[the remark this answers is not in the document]
danditomaso: One thing to be aware of is both || and ?? are now available in Javascript and neither requires a transpiler step to change ?? to || in case that's what you meant by your "(and portable)" comment?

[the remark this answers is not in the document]
dmitrylyzo: We need to support old web engines that don't support many modern features.
For example, Tizen 2.x and webOS 1/2 use the old WebKit (from 2015, iirc). There's no `let`, `const`, arrow functions, nullish coalescing operator, optional chaining operator and so on.
Because of that, we need to transpile the code. But in some cases it is better to use the most portable syntax - less additional code is produced.

thornbill (approved):

PR #5213, opened 2024-02-19
make a React/TypeScript version of emby-checkbox

terite:
`CheckBoxElement`  was a wrapper around `emby-checkbox` which worked by having React create the custom component via `dangerouslySetInnerHTML`. This has a fair amount of quirks, quickly becomes less than typesafe, results in components that are difficult to attach event listeners to, and among other things.

Instead of wrapping the custom emby components, let's create idiomatic react versions of them.
* They *react* better to updates.
* No more having to escape html.
* Want data attributes? Just set them directly.
* Supports both controlled and uncontrolled components.

PR #5226, opened 2024-02-24
use LinkButton instead of dangerouslySetInnerHTML

terite:
Let's use the component that already exists for similar reasons to here: https://github.com/jellyfin/jellyfin-web/pull/5213


src/components/search/SearchSuggestions.tsx
```diff
@@ -66,13 +54,12 @@ const SearchSuggestions: FunctionComponent<SearchSuggestionsProps> = ({ parentId
 
             <div className='searchSuggestionsList padded-left padded-right'>
                 {suggestions.map(item => (
-                    <div
```
terite, line 69: This removes the wrapping `<div>` which does change the look of search suggestions, but I believe the new look is what was intended.

thornbill:
👍 I like the direction of both PRs. Using the "dangerous" prop was never intended to be a long term solution. We should probably also migrate to using the react-router `<Link>` component.

terite:
> We should probably also migrate to using the react-router `<Link>` component.

[react-router-dom's` link implementation](https://github.com/remix-run/react-router/blob/44b391ad27dacc07bb1298bd79315c00845d5a2c/packages/react-router-dom/index.tsx#L904) is a bit more complex than one might expect. I'd rather not swap it in without additional testing. I'd also like to keep this PR more focused.

thornbill (approved):

src/components/search/SearchSuggestions.tsx
thornbill: The previous appearance was intentional so I rolled this change back.

PR #5242, opened 2024-03-01
Update crf help text

felix920506:
According to Gnattu, CRF settings aren't used when using hardware acceleration. This PR notes this in the help text

**Changes**
Update help text to reflect reality

**Issues**
None

### DIFF
--- a/src/strings/en-us.json
+++ b/src/strings/en-us.json
@@ -304,7 +304,7 @@
     "Guide": "Guide",
     "GuideProviderLogin": "Login",
     "GuideProviderSelectListings": "Select Listings",
-    "H264CrfHelp": "The 'Constant Rate Factor' (CRF) is the default quality setting for the x264 and x265 encoder. You can set the values between 0 and 51, where lower values would result in better quality (at the expense of higher file sizes). Sane values are between 18 and 28. The default for x264 is 23, and for x265 is 28, so you can use this as a starting point. These settings will be ignored when hardware acceleration is enabled.",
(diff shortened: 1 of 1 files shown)


src/strings/en-us.json
```diff
@@ -304,7 +304,7 @@
     "Guide": "Guide",
     "GuideProviderLogin": "Login",
     "GuideProviderSelectListings": "Select Listings",
-    "H264CrfHelp": "The 'Constant Rate Factor' (CRF) is the default quality setting for the x264 and x265 encoder. You can set the values between 0 and 51, where lower values would result in better quality (at the expense of higher file sizes). Sane values are between 18 and 28. The default for x264 is 23, and for x265 is 28, so you can use this as a starting point.",
+    "H264CrfHelp": "The 'Constant Rate Factor' (CRF) is the default quality setting for the x264 and x265 encoder. You can set the values between 0 and 51, where lower values would result in better quality (at the expense of higher file sizes). Sane values are between 18 and 28. The default for x264 is 23, and for x265 is 28, so you can use this as a starting point. These settings will be ignored when hardware acceleration is enabled.",
```
thornbill, line 307: How about this?

```suggestion
    "H264CrfHelp": "The 'Constant Rate Factor' (CRF) is the default quality setting for the x264 and x265 software encoders. You can set the values between 0 and 51, where lower values would result in better quality (at the expense of higher file sizes). Sane values are between 18 and 28. The default for x264 is 23, and for x265 is 28, so you can use this as a starting point. Hardware encoders do not use these settings.",
```

thornbill (approved):

PR #4858, opened 2023-10-08
Add edit and display of persons on books

Pithaya:
Allow adding persons on books, and show them on the item details page.
Related PR on the server: [#10371](https://github.com/jellyfin/jellyfin/pull/10371)

**Changes**
- Show the "people" section of the metadata editor for book items
- Add new values to the person type select
- Add new translation keys
- Show "book specific" person types (author, editor and translator) below genres, and "comics specific" person types (illustrator, inker, letterer, etc.) as cards
- Update CONTRIBUTORS.md

**Issues**

PR #5251, opened 2024-03-04
Add missing options for people metadata editor

scampower3:
**Changes**
Add missing options for people metadata editor mainly for music. Album Artist and Artist.

**Issues**

### DIFF
--- a/src/components/metadataEditor/personEditor.template.html
+++ b/src/components/metadataEditor/personEditor.template.html
@@ -26,6 +26,8 @@
                 <option value="Engineer">${Engineer}</option>
                 <option value="Mixer">${Mixer}</option>
                 <option value="Remixer">${Remixer}</option>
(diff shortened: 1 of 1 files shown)


thornbill (commented):
We should probably iterate over all `PersonKind` values to make this list (probably excluding "Unknown").

https://github.com/jellyfin/jellyfin-sdk-typescript/blob/openapi-unstable/src/generated-client/models/person-kind.ts#L22

scampower3:
Would it be better if the it generates the options dynamically according to the personkind instead of hardcoding the options into the template file?

thornbill:
I just noticed this change is included in https://github.com/jellyfin/jellyfin-web/pull/4858 also

thornbill (approved):
`AlbumArtist` was not included in the other PR. In the future, we should look at iterating over the enum of possible types from the SDK, but this is fine for a quick fix.

PR #5254, opened 2024-03-04
Add Usersetting for toggling the visibility of the Series Names in Playlists.

Arch-Storm:
**Changes**

Simply added a usersetting that can toggle whether Series Names should be displayed in Playlists above the Season and Episode number. This was already mostly implemented through `showParentTitle` in `listView.getListViewHtml()` so I just needed to add a way to use that functionality.

**Issues**

Only found this on the old tracker:
https://features.jellyfin.org/posts/1319/whan-adding-tv-shows-to-a-playlist


thornbill:
I'm not sure this needs to be a setting... I think it just makes sense to show the show name also. 🤷‍♂️

PR #5256, opened 2024-03-04
Fix depricated DOMException code property

jp-roisin:
**Changes**

The PR removes the use of the `DOMException.code` property in `userprofile.tsx`. According to the [Mozilla documentation](https://developer.mozilla.org/en-US/docs/Web/API/DOMException/code), we should use `DOMException.name` instead. Here, I've only changed one usage, but there are some more instances. I'm open to fixing them here or in a separate PR.

**Issues**

[sonarcloud-code-smell](https://sonarcloud.io/project/issues?languages=ts&resolved=false&id=jellyfin_jellyfin-web&open=AYNgV3DGl6NLO0JqGQma&tab=code)


src/apps/stable/routes/user/userprofile.tsx
```diff
@@ -80,11 +80,11 @@ const UserProfile: FunctionComponent = () => {
 
         const onFileReaderError = (evt: ProgressEvent<FileReader>) => {
             loading.hide();
-            switch (evt.target?.error?.code) {
-                case DOMException.NOT_FOUND_ERR:
+            switch (evt.target?.error?.name) {
+                case 'NotFoundError':
```
thornbill, line 84: I would create an enum for these values in `src/types`.

thornbill: @jp-roisin do you plan to address this feedback?  Thanks!

PR #5272, opened 2024-03-09
Change default AutoLogin

RoboMagus:
**Changes**
Changes the default for `enableAutoLogin` to False when login without password is used and multiple users exist.

Based on these suggestions in https://github.com/jellyfin/jellyfin-web/pull/5271#pullrequestreview-1926196757:

> 1. Use sessionStore (or session cookies) to store the credentials.
> 2. Make Remember Me unchecked by default. This requires 1. Otherwise, opening links in the new tab doesn't sign in.
> 3. Uncheck Remember Me when password-less graphical login is used.

But only affecting the RememberMe behavior.

### DIFF
--- a/src/controllers/session/login/index.js
+++ b/src/controllers/session/login/index.js
@@ -229,6 +229,12 @@ export default function (view, params) {
                 showManualForm(context, true);
             } else if (haspw == 'false') {
                 authenticateUserByName(context, getApiClient(), getTargetUrl(), name, '');
+                getApiClient().getPublicUsers().then((users) => {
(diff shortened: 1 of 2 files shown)


tylers-username:
@thornbill I’m wondering whether it makes sense to introduce middleware for this, or if the `enableAutoLogin` method itself should contain the check for multiple users.

Without middleware, or if `enableAutoLogin` does not contain the check, maintainers will need to remember to re-implement the multi-user check every time `.enableAutoLogin()` is called.

PR #5274, opened 2024-03-10
Add more media playback error messages

thornbill:
**Changes**
* Adds new media error enum
* Adds unique messages for different error types
* Fixes an issue with an error not propagating
* Fixes case sensitivity of streaming protocol checks

**Issues**
N/A


dmitrylyzo (approved):
LGTM.
I couldn't get most of the errors, but at least now we have more messages for the future.

HLS error handler doesn't stop playback. It rejects the `play` promise, but nobody seems to care about that.
https://github.com/jellyfin/jellyfin-web/blob/7ab9b653e64658eebf3a9702d8b0b6a3d1039a0a/src/components/htmlMediaHelper.js#L293-L295

UPD:
Probably because it's not being returned:
https://github.com/jellyfin/jellyfin-web/blob/7ab9b653e64658eebf3a9702d8b0b6a3d1039a0a/src/components/playback/playbackmanager.js#L2261
I tried to fix that by rewriting this block in a linear way.

UPD2:
Rejecting `play` is actually handled, but HLS seems to be rejecting something else.
https://github.com/jellyfin/jellyfin-web/blob/7ab9b653e64658eebf3a9702d8b0b6a3d1039a0a/src/components/playback/playbackmanager.js#L2550-L2557

PR #5282, opened 2024-03-16
Complement recent DoVi fallback handling changes server-side

GeorgeH005:
This pull request modifies the client to comply with the recent changes regarding proper Dolby Vision with compatibility format handling on webOS and probably other platforms as well. (jellyfin/jellyfin#10469)

**Changes**
Modify supported hevcVideoRangeTypes to conform with VideoRangeType additions server-side.

**Issues**
Not aware of any in this repository.
Fixes jellyfin/jellyfin#10468

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -216,10 +216,14 @@ function supportedDolbyVisionProfilesHevc(videoTestElement) {
     if (videoTestElement.canPlayType) {
         if (videoTestElement
             .canPlayType('video/mp4; codecs="dvh1.05.09"')
-            .replace(/no/, '')) supportedProfiles.push(5);
+            .replace(/no/, '')) {
+            supportedProfiles.push(5);
+        }
         if (videoTestElement
             .canPlayType('video/mp4; codecs="dvh1.08.09"')
-            .replace(/no/, '')) supportedProfiles.push(8);
+            .replace(/no/, '')) {
+            supportedProfiles.push(8);
+        }
     }
     return supportedProfiles;
 }
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -217,7 +217,7 @@ function supportedDolbyVisionProfilesHevc(videoTestElement) {
         if (videoTestElement
             .canPlayType('video/mp4; codecs="dvh1.05.09"')
             .replace(/no/, '')) supportedProfiles.push(5);
-        if ( videoTestElement
+        if (videoTestElement
             .canPlayType('video/mp4; codecs="dvh1.08.09"')
             .replace(/no/, '')) supportedProfiles.push(8);
     }
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -212,7 +212,7 @@ function supportsDolbyVision(options) {
 
 function supportedDolbyVisionProfilesHevc(videoTestElement) {
     const supportedProfiles = [];
-    // Profiles 5/7/8 4k@60fps
+    // Profiles 5/8 4k@60fps
     if (videoTestElement.canPlayType) {
         if (videoTestElement
             .canPlayType('video/mp4; codecs="dvh1.05.09"')
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -933,17 +933,17 @@ export default function (options) {
     if (supportsHdr10(options)) {
         hevcVideoRangeTypes += '|HDR10';
         // Should also be able to play DoVi with HDR10 fallback
-        hevcVideoRangeTypes += '|DOVIWithHDR10'
+        hevcVideoRangeTypes += '|DOVIWithHDR10';
         vp9VideoRangeTypes += '|HDR10';
         av1VideoRangeTypes += '|HDR10';
     }
 
     if (supportsHlg(options)) {
-        hevcVideoRangeTypes += "|HLG";
+        hevcVideoRangeTypes += '|HLG';
         // Should also be able to play DoVi with HLG fallback
-        hevcVideoRangeTypes += "|DOVIWithHLG";
-        vp9VideoRangeTypes += "|HLG";
-        av1VideoRangeTypes += "|HLG";
+        hevcVideoRangeTypes += '|DOVIWithHLG';
+        vp9VideoRangeTypes += '|HLG';
+        av1VideoRangeTypes += '|HLG';
     }
 
     if (supportsDolbyVision(options) && canPlayDolbyVisionHevc(videoTestElement)) {
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -218,9 +218,6 @@ function supportedDolbyVisionProfilesHevc(videoTestElement) {
             .canPlayType('video/mp4; codecs="dvh1.05.09"')
             .replace(/no/, '')) supportedProfiles.push(5);
         if ( videoTestElement
-            .canPlayType('video/mp4; codecs="dvh1.07.09"')
-            .replace(/no/, '')) supportedProfiles.push(7);
-        if ( videoTestElement
             .canPlayType('video/mp4; codecs="dvh1.08.09"')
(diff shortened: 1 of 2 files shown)


src/scripts/browserDeviceProfile.js
```diff
@@ -932,12 +932,16 @@ export default function (options) {
 
     if (supportsHdr10(options)) {
         hevcVideoRangeTypes += '|HDR10';
+        // Should also be able to play DoVi with HDR10 fallback
+        hevcVideoRangeTypes += '|DOVIWithHDR10';
```
nyanmisaka, line 936: As [mentioned before](https://github.com/jellyfin/jellyfin/pull/10469/files#diff-ab579a24be3882376b30a1f50886a7fdd3e80b5c4b18ec33e20f83ee8caafe66R1809-R1810), you cannot add them globally, which will cause Server to enable `dvh1` fourcc for HLS.js in Chrome and thus playback will fail.

src/scripts/browserDeviceProfile.js
```diff
@@ -932,12 +932,16 @@ export default function (options) {
...
         vp9VideoRangeTypes += '|HDR10';
         av1VideoRangeTypes += '|HDR10';
     }
 
     if (supportsHlg(options)) {
         hevcVideoRangeTypes += '|HLG';
+        // Should also be able to play DoVi with HLG fallback
+        hevcVideoRangeTypes += '|DOVIWithHLG';
```
nyanmisaka, line 944: Same as above.

nyanmisaka (changes requested):

src/scripts/browserDeviceProfile.js
nyanmisaka: The code here means that DOVI Profile 7/8 support is also reported for devices that only support HDR10/HLG. This is not true. The report will be wrong. You should split the three checks in [canPlayDolbyVisionHevc](https://github.com/jellyfin/jellyfin-web/blob/master/src/scripts/browserDeviceProfile.js#L213-L219) and report `DOVIWithHDR10/HLG` accordingly.

nyanmisaka: web report (HDR10) => server tag DV 8.1 as `hvc1` and drop the dovi metadata => web play as HDR10

src/scripts/browserDeviceProfile.js
```diff
@@ -942,8 +951,14 @@ export default function (options) {
         av1VideoRangeTypes += '|HLG';
     }
 
-    if (supportsDolbyVision(options) && canPlayDolbyVisionHevc(videoTestElement)) {
-        hevcVideoRangeTypes += '|DOVI';
+    if (supportsDolbyVision(options)) {
+        const profiles = supportedDolbyVisionProfilesHevc(videoTestElement);
+        if (profiles.includes(5) || profiles.includes(7)) {
```
nyanmisaka, line 956: Profile 7 is backward compatible with HDR10.

nyanmisaka: > Maybe, but i don't think we are tagging it correctly anyways. Server-side we are only handling cross-compatibility IDs 1 and 4 separately and the rest are being classified as "DOVI". As far as I can tell from dolby's pdf, profile 7 usually has a cross-compatibility ID of 6, for which we do nothing special.

Indeed, I seem to have missed this. We should label this case as `DOVIWithHDR10`. Moreover, we only add `DOVI*` for profiles that support HDR. Profiles that only support SDR are ignored. More accurately we should annotate `VideoRangeType.DOVI` in the server as a type that does not contain any fallback layer.

> 1.9. * Profile and Level for Dolby Vision MUST be Profile 5 (single layer 10-bit HEVC) and less than or equal to Level 7.

> 1.9b. + For maximum compatibility, some Dolby Vision variants SHOULD be less than or equal to Level 7.

We use the HLS documentation [provided by Apple](https://developer.apple.com/documentation/http-live-streaming/hls-authoring-specification-for-apple-devices#Video) as a reference. Profile 7 over fMP4-HLS is supported.

nyanmisaka: I did more research and found that usually DV Profile 7 videos are not tagged as `dvh1` but as `hvc1`. When Profile 7 needs to be streamed via HLS, it needs to be converted to Profile 8 to ensure compatibility. So instead of adding a new enum, we force Profile 7 to be marked as hvc1 because currently FFmpeg does not have the ability to convert DV Profiles.

src/scripts/browserDeviceProfile.js
```diff
@@ -210,12 +210,18 @@ function supportsDolbyVision(options) {
     );
 }
 
-function canPlayDolbyVisionHevc(videoTestElement) {
+function supportedDolbyVisionProfilesHevc(videoTestElement) {
+    const supportedProfiles = [];
     // Profiles 5/7/8 4k@60fps
```
nyanmisaka, line 215: The 7 in the comment can be removed as it will be handled as HDR10.

nyanmisaka:
> @nyanmisaka Do you have review or merging rights here?

Nope. cc @thornbill

nyanmisaka:
> Is there a reason this hasn't been acknowledged yet? Any concerns/problems that need correcting?

The main maintainer of the web side is currently AFK. I estimate he will be able to respond this weekend or next Monday.

jesstech:
It would be great to get these 3 PRs into the next weekly build on Monday!  I just switched to the nightly tag and would love to test this functionality before 10.9 reaches freeze.

thornbill:
Please keep in mind that I am one person with a backlog here of around 90 PRs and due to some changes in my personal life my time has been more limited the past couple weeks.

Is this PR dependent on the linked [server PR](https://github.com/jellyfin/jellyfin/pull/10469)?

thornbill:
I glanced over this and the server PR and the server PR should definitely be merged first. Just comment here once that happens and I'll remove the "server" label.

crobibero:
I have spoken

src/scripts/browserDeviceProfile.js
```diff
@@ -210,12 +210,18 @@ function supportsDolbyVision(options) {
...
-        && videoTestElement.canPlayType('video/mp4; codecs="dvh1.07.09"').replace(/no/, '')
-        && videoTestElement.canPlayType('video/mp4; codecs="dvh1.08.09"').replace(/no/, ''));
+function supportedDolbyVisionProfilesHevc(videoTestElement) {
+    const supportedProfiles = [];
+    // Profiles 5/8 4k@60fps
+    if (videoTestElement.canPlayType) {
+        if (videoTestElement
+            .canPlayType('video/mp4; codecs="dvh1.05.09"')
+            .replace(/no/, '')) supportedProfiles.push(5);
+        if ( videoTestElement
```
nyanmisaka, line 220:
```suggestion
        if (videoTestElement
```

src/scripts/browserDeviceProfile.js
```diff
@@ -210,12 +210,18 @@ function supportsDolbyVision(options) {
...
+function supportedDolbyVisionProfilesHevc(videoTestElement) {
+    const supportedProfiles = [];
+    // Profiles 5/8 4k@60fps
+    if (videoTestElement.canPlayType) {
+        if (videoTestElement
+            .canPlayType('video/mp4; codecs="dvh1.05.09"')
+            .replace(/no/, '')) supportedProfiles.push(5);
+        if (videoTestElement
+            .canPlayType('video/mp4; codecs="dvh1.08.09"')
+            .replace(/no/, '')) supportedProfiles.push(8);
```
thornbill, line 222: The bodies of these long multi-line if statements should be on a separate line (and wrapped in braces).

thornbill (changes requested):

thornbill (approved):

src/scripts/browserDeviceProfile.js
```diff
@@ -1250,4 +1266,3 @@ export default function (options) {
 
     return profile;
 }
-
```
dmitrylyzo, line 1253: I would undo the removal of the last line. AFAIK, every file in the project has a caret return on the last line.

dmitrylyzo: Oh, it's a duplicate. Okay, then.

thornbill (approved):

PR #5289, opened 2024-03-19
Enable 'Prefer fMP4-HLS Container' by default on certain platforms

nyanmisaka:
fMP4 is enabled by default only on verified patforms. This
largely avoids transcoding HEVC and AV1, instead letting
the client handle them.

**Changes**
- Enable 'Prefer fMP4-HLS Container' by default on certain platforms
- Increase the minimum version of Tizen that supports native fMP4 to 5

**Issues**
- Many users are unaware that this option exists and it was disabled by default.

### DIFF
--- a/src/scripts/settings/userSettings.js
+++ b/src/scripts/settings/userSettings.js
@@ -1,4 +1,5 @@
 import appSettings from './appSettings';
+import browser from '../browser';
 import Events from '../../utils/events.ts';
 import { toBoolean } from '../../utils/string.ts';
 
@@ -140,7 +141,8 @@ export class UserSettings {
             return this.set('preferFmp4HlsContainer', val.toString(), false);
         }
 
-        return toBoolean(this.get('preferFmp4HlsContainer', false), false);
+        // Enable it by default only for the platforms that play fMP4 for sure.
+        return toBoolean(this.get('preferFmp4HlsContainer', false), browser.safari || browser.firefox || browser.chrome || browser.edgeChromium);
     }
 
     /**
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -68,7 +68,7 @@ function canPlayNativeHls() {
 }
 
 function canPlayNativeHlsInFmp4() {
-    if (browser.tizenVersion >= 3 || browser.web0sVersion >= 3.5) {
+    if (browser.tizenVersion >= 5 || browser.web0sVersion >= 3.5) {
         return true;
     }
 


src/scripts/settings/userSettings.js
```diff
@@ -140,7 +140,7 @@ export class UserSettings {
             return this.set('preferFmp4HlsContainer', val.toString(), false);
         }
 
-        return toBoolean(this.get('preferFmp4HlsContainer', false), false);
+        return toBoolean(this.get('preferFmp4HlsContainer', false), true);
```
dmitrylyzo, line 143: We need to disable fMP4 in device profile on some platforms. _Then we can remove this option (always enable fMP4)._
At least on Tizen 4, I've never been able to get it to work - it always waits for the remuxing job to complete.

nyanmisaka: So disabled for this version of Tizen? Or only enable fMP4 by default only for HLS.js and Safari?

dmitrylyzo: We can trim https://github.com/jellyfin/jellyfin-web/blob/1196ae992f3afc27f8cef8e74367f98bb321be31/src/scripts/browserDeviceProfile.js#L71

but agree
> It is unlikely that we will test every version of Tizen and WebOS one by one. I think they should be left to the user's choice.

As an option, `true` -> `!browser.tizen && !browser.web0s` or list the platforms that play fMP4 for sure. In this case it will be disabled by default on some platforms, but the user can still enable it.

Tbh, I don't know why it works that way on my TV. :shrug: Iirc, I tested fMP4 with AVPlay - same bug.

nyanmisaka: I increased the minimum version from 3 to 5 based on your experience.
```diff
-browser.tizenVersion >= 3
+browser.tizenVersion >= 5
```

Do these older Tizen versions support MSE? If so you can enable HLS.js as a replacement.

nyanmisaka: Yeah. Even Tizen 3.0 has been marked as the first version to support fMP4-HLS. Have you also tried the online m3u8 link provided by Apple? If the fMP4 in it cannot be played normally, then this should be a system issue and should be reported to Samsung.

https://developer.samsung.com/smarttv/develop/specifications/general-specifications.html#Streaming-Feature-Support

nyanmisaka:
> For Tizen, this should be the relevant page: https://developer.samsung.com/smarttv/develop/specifications/general-specifications.html#Streaming-Feature-Support
> 
> Sorry if I restated the obvious, found it useful to hhave the relevant links here.

Unfortunately according to dmitrylyzo's testing, the actual behavior on older Tizen contradicts the documentation. So until someone can fully test it on every Tizen version (3.0+) and WebOS, we're not going to risk breaking playback by enabling it by default.

dmitrylyzo (approved):
I am not sure about Tizen version, but I think it is fine for now. Tizen doesn't really need fMP4.

ikheetjeff:
I've been sticking to version 10.8.13 for a while because higher versions had issues with out-of-sync subtitles. After trying many things, it turns out that disabling Prefer fMP4-HLS Container on Firefox and iOS (Jellyfin Mobile) mostly solves the desync.

After a lot of rewinding and fast-forwarding, desync does occur again, but that also happens on 10.8.13. However, it happens much less quickly and only occasionally. With this feature enabled by default, it almost always happens.

I would like to suggest not enabling this feature by default because, in my opinion, it doesn't work well. I've already made a custom build for myself because otherwise, this version is unusable and I would be forced to revert to 10.8.13. Or maybe I'm missing something crucial?

PR #5293, opened 2024-03-21
Fix searching in tv shows

scampower3:
**Changes**

Searching in tv shows will not return any results.
Replaced collectionType=tv to collectionType=tvshows
**Issues**

### DIFF
--- a/src/controllers/shows/tvrecommended.js
+++ b/src/controllers/shows/tvrecommended.js
@@ -13,6 +13,7 @@ import { LibraryTab } from 'types/libraryTab';
 import { getBackdropShape } from 'utils/card';
 import Dashboard from 'utils/dashboard';
 import Events from 'utils/events';
+import { CollectionType } from '@jellyfin/sdk/lib/generated-client/models/collection-type';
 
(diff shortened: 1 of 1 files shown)


dmitrylyzo (approved):

src/controllers/shows/tvrecommended.js
```diff
@@ -332,7 +332,7 @@ export default function (view, params) {
     function onInputCommand(e) {
         if (e.detail.command === 'search') {
             e.preventDefault();
-            Dashboard.navigate('search.html?collectionType=tv&parentId=' + params.topParentId);
+            Dashboard.navigate('search.html?collectionType=tvshows&parentId=' + params.topParentId);
```
thornbill, line 335: We should pull the value from the SDK to make this future proof.

```js
import { CollectionType } from '@jellyfin/sdk/lib/generated-client/models/collection-type';
```

```suggestion
            Dashboard.navigate(`search.html?collectionType=${CollectionType.Tvshows}&parentId=${params.topParentId}`);
```

thornbill (approved):

PR #5295, opened 2024-03-22
Log current version to console to help out with issue triage and debugging

nielsvanvelzen:
This is a small change that adds some additional logging. As soon as the frontend initializes (via the init function) it will print a message with the package name, package version and build version. This is helpful when sharing browser logs as it could indicate which specific unstable build is used when there is no dashboard access. It could also be helpful to know if an older cached version of the frontend is loading.

There are some other messages in the logs before this one because some files contain top-level log statements. It didn't really matter if I placed this log statement in the init function or also make it top-level, so I opted for the side-effect free variant.

Sample output in dev mode:
> `jellyfin-web version 10.9.0 build Dev Server`

**Changes**
- Log current version to console to help out with issue triage and debugging

**Issues**

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -265,7 +265,9 @@ module.exports = {
                 // Build time definitions
                 __JF_BUILD_VERSION__: 'readonly',
                 __USE_SYSTEM_FONTS__: 'readonly',
-                __WEBPACK_SERVE__: 'readonly'
+                __WEBPACK_SERVE__: 'readonly',
+                __PACKAGE_JSON_NAME__: 'readonly',
+                __PACKAGE_JSON_VERSION__: 'readonly'
             },
             rules: {
                 '@typescript-eslint/prefer-string-starts-ends-with': ['error']
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -264,10 +264,10 @@ module.exports = {
                 'Windows': 'readonly',
                 // Build time definitions
                 __JF_BUILD_VERSION__: 'readonly',
-                __USE_SYSTEM_FONTS__: 'readonly',
-                __WEBPACK_SERVE__: 'readonly',
                 __PACKAGE_JSON_NAME__: 'readonly',
-                __PACKAGE_JSON_VERSION__: 'readonly'
+                __PACKAGE_JSON_VERSION__: 'readonly',
+                __USE_SYSTEM_FONTS__: 'readonly',
+                __WEBPACK_SERVE__: 'readonly'
             },
             rules: {
                 '@typescript-eslint/prefer-string-starts-ends-with': ['error']
--- a/src/global.d.ts
+++ b/src/global.d.ts
@@ -18,4 +18,6 @@ export declare global {
     const __JF_BUILD_VERSION__: string;
     const __USE_SYSTEM_FONTS__: string;
     const __WEBPACK_SERVE__: string;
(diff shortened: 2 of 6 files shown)


src/index.jsx
```diff
@@ -38,6 +38,7 @@ import './legacy/vendorStyles';
 import { currentSettings } from './scripts/settings/userSettings';
 import taskButton from './scripts/taskbutton';
 import RootApp from './RootApp.tsx';
+import { name as WEB_NAME, version as WEB_VERSION } from '../package.json';
```
thornbill, line 41: Does webpack complain about this import? It does on the dashboard page so I was planning to change it at some point...

nielsvanvelzen: There are two possible solutions for that warning:

1) Like apphost.js, don't use named export but just import the complete document and read from that
2) Add build defines for the relevant fields

I opted for option 2 and just pushed that

.eslintrc.js
```diff
@@ -265,7 +265,9 @@ module.exports = {
                 // Build time definitions
                 __JF_BUILD_VERSION__: 'readonly',
                 __USE_SYSTEM_FONTS__: 'readonly',
-                __WEBPACK_SERVE__: 'readonly'
+                __WEBPACK_SERVE__: 'readonly',
+                __PACKAGE_JSON_NAME__: 'readonly',
+                __PACKAGE_JSON_VERSION__: 'readonly'
```
dmitrylyzo, line 270: Sort?
```suggestion
                __PACKAGE_JSON_NAME__: 'readonly',
                __PACKAGE_JSON_VERSION__: 'readonly',
                __WEBPACK_SERVE__: 'readonly'
```

src/global.d.ts
```diff
@@ -18,4 +18,6 @@ export declare global {
     const __JF_BUILD_VERSION__: string;
     const __USE_SYSTEM_FONTS__: string;
     const __WEBPACK_SERVE__: string;
+    const __PACKAGE_JSON_NAME__: string;
+    const __PACKAGE_JSON_VERSION__: string;
```
dmitrylyzo, line 22: Sort?
```suggestion
    const __PACKAGE_JSON_NAME__: string;
    const __PACKAGE_JSON_VERSION__: string;
    const __WEBPACK_SERVE__: string;
```

webpack.common.js
```diff
@@ -52,7 +53,9 @@ const config = {
                     'Dev Server' :
                     process.env.JELLYFIN_VERSION || 'Release'),
             __USE_SYSTEM_FONTS__: JSON.stringify(!!process.env.USE_SYSTEM_FONTS),
-            __WEBPACK_SERVE__: JSON.stringify(!!process.env.WEBPACK_SERVE)
+            __WEBPACK_SERVE__: JSON.stringify(!!process.env.WEBPACK_SERVE),
+            __PACKAGE_JSON_NAME__: JSON.stringify(packageJson.name),
+            __PACKAGE_JSON_VERSION__: JSON.stringify(packageJson.version)
```
dmitrylyzo, line 58: Sort?
```suggestion
            __PACKAGE_JSON_NAME__: JSON.stringify(packageJson.name),
            __PACKAGE_JSON_VERSION__: JSON.stringify(packageJson.version),
            __WEBPACK_SERVE__: JSON.stringify(!!process.env.WEBPACK_SERVE)
```

thornbill (approved):

PR #5300, opened 2024-03-23
Upgrade action version

Gauvino:


thornbill:
Just for some context, normally renovate takes care of bumping these versions it just had too many open PRs already.

PR #5312, opened 2024-03-24
Fix mediainfo display for lyrics and non-video

crobibero:
Fixes: https://github.com/jellyfin/jellyfin-web/issues/5273

### DIFF
--- a/src/components/itemMediaInfo/itemMediaInfo.js
+++ b/src/components/itemMediaInfo/itemMediaInfo.js
@@ -84,6 +84,7 @@ function getMediaSourceHtml(user, item, version) {
             case 'Data':
             case 'Subtitle':
             case 'Video':
+            case 'Lyric':
                 translateString = stream.Type;
                 break;
             case 'EmbeddedImage':
@@ -145,10 +146,10 @@ function getMediaSourceHtml(user, item, version) {
         if (stream.BitDepth) {
             attributes.push(createAttribute(globalize.translate('MediaInfoBitDepth'), `${stream.BitDepth} bit`));
         }
-        if (stream.VideoRange) {
+        if (stream.VideoRange && stream.Type === 'Video') {
             attributes.push(createAttribute(globalize.translate('MediaInfoVideoRange'), stream.VideoRange));
         }
-        if (stream.VideoRangeType) {
+        if (stream.VideoRangeType && stream.Type === 'Video') {
             attributes.push(createAttribute(globalize.translate('MediaInfoVideoRangeType'), stream.VideoRangeType));
         }
         if (stream.VideoDoViTitle) {
--- a/src/strings/en-us.json
+++ b/src/strings/en-us.json
@@ -951,6 +951,7 @@
     "LogLevel.None": "None",
     "Logo": "Logo",
     "LogoScreensaver": "Logo Screensaver",
+    "Lyric": "Lyric",
     "Lyricist": "Lyricist",
     "ManageLibrary": "Manage library",
     "ManageRecording": "Manage recording",


src/components/itemMediaInfo/itemMediaInfo.js
```diff
@@ -84,6 +84,7 @@ function getMediaSourceHtml(user, item, version) {
             case 'Data':
             case 'Subtitle':
             case 'Video':
+            case 'Lyric':
                 translateString = stream.Type;
                 break;
```
dmitrylyzo, line 89: Not requesting any changes, just wondering why not
```js
let translateString = stream.Type;
if (stream.Type === 'EmbeddedImage') translateString = 'Image'; // `switch` may be used as well
```

src/components/itemMediaInfo/itemMediaInfo.js
```diff
@@ -145,13 +146,13 @@ function getMediaSourceHtml(user, item, version) {
         if (stream.BitDepth) {
             attributes.push(createAttribute(globalize.translate('MediaInfoBitDepth'), `${stream.BitDepth} bit`));
         }
-        if (stream.VideoRange) {
+        if (stream.VideoRange && stream.Type === 'Video') {
```
dmitrylyzo, line 149: Isn't `stream.VideoRange` check enough?

src/strings/en-us.json
```diff
@@ -1663,5 +1663,6 @@
     "OptionExtractTrickplayImage": "Enable trickplay image extraction",
     "ExtractTrickplayImagesHelp": "Trickplay images are similar to chapter images, except they span the entire length of the content and are used to show a preview when scrubbing through videos.",
     "LabelExtractTrickplayDuringLibraryScan": "Extract trickplay images during the library scan",
-    "LabelExtractTrickplayDuringLibraryScanHelp": "Generate trickplay images when videos are imported during the library scan. Otherwise, they will be extracted during the trickplay images scheduled task. If generation is set to non-blocking this will not affect the time a library scan takes to complete."
+    "LabelExtractTrickplayDuringLibraryScanHelp": "Generate trickplay images when videos are imported during the library scan. Otherwise, they will be extracted during the trickplay images scheduled task. If generation is set to non-blocking this will not affect the time a library scan takes to complete.",
+    "Lyric": "Lyric"
```
dmitrylyzo, line 1667: It would be ideal to place new entries in alphabetical order - less conflict.

src/components/itemMediaInfo/itemMediaInfo.js
nyanmisaka: When the video stream does not exist (audio only), its default value is `VideoRange.Unknown`. So this makes sense to me.

src/components/itemMediaInfo/itemMediaInfo.js
crobibero: I originally did that change but reverted when I realized there was localization.
This way we are forced to add localization :smile:

src/components/itemMediaInfo/itemMediaInfo.js
```diff
@@ -145,13 +146,13 @@ function getMediaSourceHtml(user, item, version) {
...
             attributes.push(createAttribute(globalize.translate('MediaInfoVideoRange'), stream.VideoRange));
         }
-        if (stream.VideoRangeType) {
+        if (stream.VideoRangeType && stream.Type === 'Video') {
             attributes.push(createAttribute(globalize.translate('MediaInfoVideoRangeType'), stream.VideoRangeType));
         }
-        if (stream.VideoDoViTitle) {
+        if (stream.VideoDoViTitle && stream.Type === 'Video') {
```
dmitrylyzo, line 155: `VideoDoViTitle` is [`undefined`](https://github.com/jellyfin/jellyfin/blob/b1870792b1310b58f84089c1c50deab1b7d12f24/MediaBrowser.Model/Entities/MediaStream.cs#L214) for non-video streams. So I think it is safe not to check the stream type.
```suggestion
        if (stream.VideoDoViTitle) {
```

dmitrylyzo (approved):
LGTM. Maybe squash commits?

PR #5322, opened 2024-03-25
Limit Video Level to 4.2 in fMP4 for iOS < 13

Bond-009:
Fixes a playback error I experienced with a Level 5 video on iOS 12 in Safari with `Prefer fMP4-HLS Media Container`  enabled

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -1147,7 +1147,7 @@ export default function (options) {
 
     // On iOS 12.x, for TS container max h264 level is 4.2
     if (browser.iOS && browser.iOSVersion < 13) {
-        const codecProfile = {
+        const codecProfileTS = {
             Type: 'Video',
             Codec: 'h264',
             Container: 'ts',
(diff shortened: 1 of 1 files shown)


src/scripts/browserDeviceProfile.js
```diff
@@ -1156,14 +1156,32 @@ export default function (options) {
...
             IsRequired: false
         });
 
-        profile.CodecProfiles.push(codecProfile);
+        profile.CodecProfiles.push(codecProfileTS);
+
+        const codecProfileMp4 = {
+            Type: 'Video',
+            Codec: 'h264',
+            Container: 'mp4',
```
dmitrylyzo, line 1171: You can omit `Container` if you need to apply the condition to H264 in any container.

nyanmisaka: Indeed. If streaming through HLS fails, directplay may fail too. Directplay is usually more strict on Safari.

Bond-009: Same video direct plays in an mov container (without this patch)

thornbill (approved):

PR #5326, opened 2024-03-26
fix: only display mjpeg hw encoding in qsv and vaapi

gnattu:
**Changes**

Hardware-accelerated MJPEG encoding is currently only implemented in QSV and VAAPI, and this option will have no effect if other hardware acceleration methods are in use. Displaying this option will confuse our users when they are using other hardware acceleration methods, therefore hide the option.

**Issues**

### DIFF
--- a/src/controllers/dashboard/encodingsettings.html
+++ b/src/controllers/dashboard/encodingsettings.html
@@ -135,7 +135,7 @@
                             <span>${AllowAv1Encoding}</span>
                         </label>
                     </div>
-                    <div class="checkboxList">
+                    <div class="checkboxList allowMjpegEncodingOption hide">
                         <label>
                             <input type="checkbox" is="emby-checkbox" id="chkAllowMjpegEncoding" />
                             <span>${AllowMjpegEncoding}</span>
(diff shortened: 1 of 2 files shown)


nyanmisaka (approved):

thornbill:
We mention mjpeg transcoding in the trickplay settings which might be confusing also

nyanmisaka:
> We mention mjpeg transcoding in the trickplay settings which might be confusing also

There was a discussion I raised about this https://github.com/jellyfin/jellyfin/pull/9554#discussion_r1151901354 But it seems I missed it later. Actually the HEVC option is used to enable/disable both HEVC software and hardware encoding.

If possible, it is best to remove the MJPEG box. Trickplay can still borrow the codec settings in Playback, but whether to enable MJPEG hardware encoding needs to be set in the Trickplay tab. Eventually in the future, ffmpeg will be used to detect whether the hardware supports MJPEG, etc., to achieve automation. cc@nicknsy

gnattu:
I plan to have a more proper fix for this and will require a server change. I will move this option into the Trickplay panel and rename the options, then add some helper texts.

Issue #5288, opened 2024-03-18
Track gets skiped after closing a book


lomenzel:
**Describe The Bug**
If listening to music while reading a book, when closing the book the current treck ends and a new one is started

**Steps To Reproduce**
Start a music song
Open a Book
Close the Book
see that next song is started

**Expected Behavior**
I would expect the song to finish before the next one is played

**Logs**

**System (please complete the following information):**
 - Platform: Linux (NixOS)
 - Browser: All
 - Jellyfin Version: Version: 10.8.13

**Additional Context**

scottross123:
Hey, I was able to recreate this bug, same version and on Linux. I think I have a fix and made a PR, it seems be due this line of code : `Events.trigger(this, 'stopped', [stopInfo]);` in `src/plugins/bookPlayer/plugin.js`.
https://github.com/jellyfin/jellyfin-web/blob/fce1e7f173e38c499b3fc98f5f50413a12e29592/src/plugins/bookPlayer/plugin.js#L68

https://github.com/jellyfin/jellyfin-web/blob/fce1e7f173e38c499b3fc98f5f50413a12e29592/src/plugins/bookPlayer/plugin.js#L64C9-L69C1

 This code seems to dispatch an event which causes the media player to stop and go to the next track, and is not related to the actual closing of the book. Removing it fixes the issue and does not affect the open/closing of books.

I am a new contributor to Jellyfin though so I would first like a more senior contributor or just someone more familiar with the code base to confirm this.

#5329

thornbill:
Copied from the PR:

This is currently working as intended. As far as I know the stopped event is needed for playback tracking. The playback manager would need a major overhaul to support playing multiple things at once in the same session.

PR #5329, opened 2024-03-27
Fix track getting skipped when open book is closed

scottross123:
**Changes**

Removed `Events.trigger(this, 'stopped', [stopInfo]);`  from `src/plugins/bookerPlayer/plugin.js`. This line creates a bug where closing a book triggers this function which cause the current track to be skipped if music is also playing while reading a book. Does not affect the actual opening/closing of a book.

**Issues**

#5288

### DIFF
--- a/src/plugins/bookPlayer/plugin.js
+++ b/src/plugins/bookPlayer/plugin.js
@@ -61,12 +61,6 @@ export class BookPlayer {
     stop() {
         this.unbindEvents();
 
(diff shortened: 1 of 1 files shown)


thornbill:
This is currently working as intended. As far as I know the stopped event is needed for playback tracking. The playback manager would need a major overhaul to support playing multiple things at once in the same session.

PR #5332, opened 2024-03-27
Disable ffmpeg path setting for security reasons

gnattu:
**Changes**

This forward-ports the security patch we implemented for 10.8.13 into master. Disables the ffmpeg path setting in GUI.

**Issues**

### DIFF
--- a/src/controllers/dashboard/encodingsettings.html
+++ b/src/controllers/dashboard/encodingsettings.html
@@ -251,9 +251,8 @@
                 <div class="inputContainer fldEncoderPath">
                     <div style="display: flex; align-items: center;">
                         <div style="flex-grow:1;">
-                            <input is="emby-input" class="txtEncoderPath" label="${LabelffmpegPath}" autocomplete="off" dir="ltr" />
+                            <input is="emby-input" class="txtEncoderPath" label="${LabelffmpegPath}" autocomplete="off" dir="ltr" disabled/>
                         </div>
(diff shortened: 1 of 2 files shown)


nyanmisaka (approved):

src/controllers/dashboard/encodingsettings.html
```diff
@@ -251,9 +251,8 @@ <h3 class="checkboxListLabel">${LabelHardwareEncodingOptions}</h3>
                 <div class="inputContainer fldEncoderPath">
                     <div style="display: flex; align-items: center;">
                         <div style="flex-grow:1;">
-                            <input is="emby-input" class="txtEncoderPath" label="${LabelffmpegPath}" autocomplete="off" dir="ltr" />
+                            <input is="emby-input" class="txtEncoderPath" label="${LabelffmpegPath}" autocomplete="off" dir="ltr" disabled/>
```
nyanmisaka, line 254: Maybe add update the description to inform users that they need to modify the ffmpeg path in the config file if necessary?

gnattu: I think we will want to have this in the changelog.

Currently the ffmpeg path order is command line -> config -> system, and most of our packaging method will use command line to specify ffmpeg so the config file would not have effect and the user will want to change the cli in this case.

thornbill (approved):

PR #5341, opened 2024-03-29
ignore .DS_Store

felix920506:
ignores .DS_Store file created by macOS finder
https://en.wikipedia.org/wiki/.DS_Store

**Issues**
None

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,6 @@ yarn-error.log
 # build artifacts
 fedora/jellyfin-web-*.src.rpm
 fedora/jellyfin-web-*.tar.gz
+
+# Finder
+.DS_Store


thornbill:
You should probably configure this (and any other OS specific ignores) globally.

felix920506:
I saw it in the main repo .gitignore so I figured it should probably be here as well
https://github.com/jellyfin/jellyfin/blob/master/.gitignore#L201-L202

thornbill:
Personally I think it's better for people to deal with the specifics of their systems in the global config to avoid adding cruft to every repository they contribute to. 🤷‍♂️

vreelb:
That's certainly the ideal, but for repositories that accept public contributions having the `.gitignore` cover something that a novice might accidentally commit saves reviewers the need to comment about it on their PRs.

PR #5348, opened 2024-03-29
Use SeriesSortName in SortBy

dmitrylyzo:
`SeriesName` is not a valid value for `SortBy`.

**Changes**
Use `SeriesSortName` in `SortBy`.

**Issues**
N/A

### DIFF
--- a/src/controllers/favorites.js
+++ b/src/controllers/favorites.js
@@ -133,7 +133,7 @@ function getFetchDataFn(section) {
     return function () {
         const apiClient = this.apiClient;
         const options = {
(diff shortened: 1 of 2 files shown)


nielsvanvelzen (approved):
Could also use one of these:
https://github.com/jellyfin/jellyfin-sdk-typescript/blob/master/src/models/api/item-sort-by.ts
https://github.com/jellyfin/jellyfin-sdk-typescript/blob/openapi-unstable/src/generated-client/models/item-sort-by.ts

thornbill (approved):

PR #5361, opened 2024-04-01
Show slideshow controls when touched

tcely:
It appears that a previous commit changed the blocks so that only mouse events were recognized.

**Changes**

Moved the `showOsd` call outside of the "mouse" block.

**Issues**


tcely:
A PR for the newer branch has been opened as #6016 instead of fixing the merge problems created by changing the base branch.

Issue #5374, opened 2024-04-05
[Issue]: Playing Track >= 301 will play track 1 instead


felix920506:
### Please describe your bug

When playing track >= 301, track 1 will play instead. If I play track 300 I cannot skip to the next track.

### Reproduction Steps

1. Add an album with more than 300 tracks in a single "disc" to a music library and rescan
2. Play track >= 301
3. See Error (variant 1)
4. Play track 300
5. Skip to next track
6. See Error (variant 2)

### Jellyfin Version

Unstable (master branch)

### if other:

Build 2024040105

### Environment

```markdown
- OS: Windows 11 Pro, macOS 13.1
- Linux Kernel: N/A
- Virtualization: None
- Clients: Web
- Browser: Firefox
- FFmpeg Version: 6.0.1-jellyfin, 6.1.1
- Playback Method: Direct
- Hardware Acceleration: QSV, VT
- GPU Model: Intel N100, Apple M1 Pro
- Plugins: Default
- Reverse Proxy: None
- Base URL: None
- Networking: NAT, Localhost
- Storage: SMB, Local SSD APFS
```

### Jellyfin logs

```shell
[2024-04-06 00:59:55.444 +08:00] [INF] [37] Emby.Server.Implementations.Session.SessionManager: Playback stopped reported by app "Jellyfin Web" "10.9.0" playing "cocoro*cosmetic". Stopped at "2698" ms
[...]

PR #5375, opened 2024-04-05
Remove getItemsForPlayback default limit

felix920506:
**Changes**
Remove the limit of getItemsForPlayback function if a limit is not given

**Issues**
Fixes https://github.com/jellyfin/jellyfin-web/issues/5374


felix920506:
I was told this is not a proper fix for the issue

PR #5378, opened 2024-04-08
fix: use correct hls containers for audio

### LINKED ISSUE #5221: Audio stuttering when transcode playing ogg vorbis in Safari
**Describe The Bug**
When transcode playing ogg vorbis files in safari, the audio stutters.
This doesn't seem to happen when transcoding from other formats (eg. opus)

**Media Information**
```
last
Container: ogg
Path: /Users/user/Music/test_music_lib/Arcaea All Songs/last.ogg
Size: 4 MB

Audio
Title: Last - VORBIS - Stereo
Codec: VORBIS
AVC: No
Layout: stereo
Channels: 2 ch
Bitrate: 192 kbps
Sample rate: 44100 Hz
Video range: Unknown
Video range type: Unknown
Default: No
Forced: No
External: No

Image
Codec: MJPEG
AVC: No
Profile: Baseline
Resolution: 768x768
Bit depth: 8 bit
Video range: Unknown
Video range type: Unknown
Color space: bt470bg
Pixel format: yuvj444p
Ref frames: 1
```

**Screenshots**
None

 - Platform: MacOS, iOS
 - Browser: Safari
 - Jellyfin Version: Master (Unstable)
 
**Additional Context**
Playing around in the source code and using hls.js instead of Safari's native HLS support seems to solve the issue on MacOS Safari

gnattu:
**Changes**

The current container setting was set 4 years ago and does not reflect current browsers. The value `aac` is not a valid container for hls either, as ffmpeg only supports hls in `ts` or `mp4` container.
We should prioritize using fmp4 when possible and fallback to mpegts if necessary.

**Issues**

Fixes #5221

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -718,9 +718,13 @@ export default function (options) {
     const hlsBreakOnNonKeyFrames = browser.iOS || browser.osx || browser.edge || !canPlayNativeHls();
 
     if (canPlayHls() && browser.enableHlsAudio !== false) {
+        let enableFmp4Hls = userSettings.preferFmp4HlsContainer();
+        if ((browser.safari || browser.tizen || browser.web0s) && !canPlayNativeHlsInFmp4()) {
+            enableFmp4Hls = false;
+        }
         profile.TranscodingProfiles.push({
             // hlsjs, edge, and android all seem to require ts container
-            Container: !canPlayNativeHls() || browser.edge || browser.android ? 'ts' : 'aac',
+            Container: enableFmp4Hls ? 'mp4' : 'ts',
             Type: 'Audio',
             AudioCodec: 'aac',
             Context: 'Streaming',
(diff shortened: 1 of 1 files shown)


src/scripts/browserDeviceProfile.js
```diff
@@ -718,9 +718,12 @@ export default function (options) {
     const hlsBreakOnNonKeyFrames = browser.iOS || browser.osx || browser.edge || !canPlayNativeHls();
 
     if (canPlayHls() && browser.enableHlsAudio !== false) {
+        let enableFmp4Hls = userSettings.preferFmp4HlsContainer();
+        if ((browser.safari || browser.tizen || browser.web0s) && !canPlayNativeHlsInFmp4()) {
+            enableFmp4Hls = false;
+        }
```
dmitrylyzo, line 724: Maybe extract this to reuse below (in the video block)?

felix920506:
I can confirm that this fixes the issue in Safari on both macOS and iOS

dmitrylyzo (approved):

PR #5380, opened 2024-04-09
Fix cards in emby-scroller being off-center from section titles

### LINKED ISSUE #5379: [Issue]: 'Cast & Crew' and 'More Like This' are not aligned with text in 10.9
### Please describe your bug

image

In 10.8.X the 'Cast & Crew' and 'More Like This' are aligned with the text. But in the new unstable release they are shifted to the right which seems like a bug.

### Reproduction Steps

No modification of the release, it is there by default. 

### Jellyfin Version

Unstable (master branch)

### if other:

_No response_

### Environment

```markdown
- OS:Docker
```

### Jellyfin logs

```shell
Not required
```

### FFmpeg logs

```shell
Not required
```

Not required

image
image

### Code of Conduct

ConnorS1110:
**Changes**
A change was made to `emby-scroller` to always have a 3.3% padding-left value (at a minimum). However, it needs to be 0 on the `itemDetails` controller to properly align. I created a new class in the `emby-scroller` that sets the padding-left attribute to 0 and added the class to the `emby-scroller`s on the page, so as to not break the other usages of this component in other places.

**Issues**
Fixes #5379

### DIFF
--- a/src/elements/emby-scroller/emby-scroller.scss
+++ b/src/elements/emby-scroller/emby-scroller.scss
@@ -26,3 +26,13 @@
         margin-left: 1.2em;
     }
 }
+
+.no-padding {
+    [dir="ltr"] & {
+        padding-left: 0;
+    }
+
+    [dir="rtl"] & {
+        padding-right: 0;
+    }
+}
--- a/src/controllers/itemDetails/index.html
+++ b/src/controllers/itemDetails/index.html
@@ -175,35 +175,35 @@
 
                 <div id="additionalPartsCollapsible" class="verticalSection detailVerticalSection hide">
                     <h2 class="sectionTitle sectionTitle-cards padded-right">${HeaderAdditionalParts}</h2>
-                    <div is="emby-scroller" class="padded-top-focusscale padded-bottom-focusscale" data-centerfocus="true">
+                    <div is="emby-scroller" class="padded-top-focusscale padded-bottom-focusscale no-padding" data-centerfocus="true">
                         <div id="additionalPartsContent" is="emby-itemscontainer" class="scrollSlider focuscontainer-x itemsContainer"></div>
                     </div>
                 </div>
 
                 <div class="verticalSection detailVerticalSection moreFromSeasonSection hide">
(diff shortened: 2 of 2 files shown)


aaemon (approved):
no issue

src/elements/emby-scroller/emby-scroller.scss
```diff
@@ -26,3 +26,7 @@
         margin-left: 1.2em;
     }
 }
+
+.no-padding-left {
+    padding-left: 0;
```
dmitrylyzo, line 31: We need `padding-right: 0;` for Right-To-Left text direction.

dmitrylyzo (commented):
The ideal would be to remove the original padding:
https://github.com/jellyfin/jellyfin-web/blob/6cc6474ba95d0597368ce99e02051c20215d7a8d/src/elements/emby-scroller/emby-scroller.scss#L5-L10
and add `padding-left padding-right` where needed.

src/elements/emby-scroller/emby-scroller.scss
ConnorS1110: I believe I have updated this now. It is more generic and looks for the direction and sets padding to be 0 on the correct side accordingly

src/elements/emby-scroller/emby-scroller.scss
```diff
@@ -3,9 +3,7 @@
 }
 
 .emby-scroller {
-    padding-left: 3.3%;
     padding-left: max(env(safe-area-inset-left), 3.3%);
-    padding-right: 3.3%;
     padding-right: max(env(safe-area-inset-right), 3.3%);
```
dmitrylyzo, line 7: These are for backward compatibility. Revert please.

ConnorS1110: Sorry thought your original comment was suggesting to remove that. Reverted.

dmitrylyzo (approved):

1hitsong (approved):
✅ Recreated the issue and confirmed this PR fixes it.

PR #5391, opened 2024-04-16
User configurable codecs (DTS, TrueHD)

dmitrylyzo:
**Changes**
- Extract `canPlayDts` function.
- Add DTS enabling option.
- Add TrueHD enabling option.
- Add Video Advanced section.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-webos/issues/190
Fixes https://github.com/jellyfin/jellyfin-tizen/issues/226

**Screenshots**
video-advanced

### DIFF
--- a/src/components/playbackSettings/playbackSettings.js
+++ b/src/components/playbackSettings/playbackSettings.js
@@ -174,6 +174,7 @@ function loadForm(context, user, userSettings, systemInfo, apiClient) {
     context.querySelector('.chkPlayDefaultAudioTrack').checked = user.Configuration.PlayDefaultAudioTrack || false;
     context.querySelector('.chkPreferFmp4HlsContainer').checked = userSettings.preferFmp4HlsContainer();
     context.querySelector('.chkEnableDts').checked = appSettings.enableDts();
+    context.querySelector('.chkEnableTrueHd').checked = appSettings.enableTrueHd();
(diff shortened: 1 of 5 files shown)


src/scripts/settings/userSettings.js
```diff
@@ -145,6 +145,32 @@ export class UserSettings {
...
+     * @param {boolean|undefined} val - Flag to enable 'Enable TrueHD' or undefined.
+     * @return {boolean} 'Enable TrueHD' state.
+     */
+    enableTrueHd(val) {
+        if (val !== undefined) {
+            return this.set('enableTrueHd', val.toString(), false);
+        }
+
+        return toBoolean(this.get('enableTrueHd', false), false);
+    }
```
dmitrylyzo, line 172: These settings should be app wide, not per user, IMO. But I copied `fMP4`.

thornbill:
Should we try to get this in for 10.9?

dmitrylyzo:
> Should we try to get this in for 10.9?

I'd say yes.
Because this fixes a regression for users who upgraded webOS 22 to 23 but didn't get DTS support.

hexoverflow:
I have been running jellyfin 10.9 on webOS with this change for the past week. It works fine. Without this patch I don't get any audio for dts files on my upgraded 2022 TV

thornbill (approved):

PR #5392, opened 2024-04-16
Convert playlist editor to typescript

thornbill:
**Changes**
* Removes an unused playlist controller
* Converts the playlist editor to typescript
* Replaces usage of the apiclient with the sdk in the playlist editor

**Issues**
N/A


nielsvanvelzen (approved):
Looks like migrated TypeScript code

src/scripts/playlists.js
```diff
@@ -1,203 +0,0 @@
-import loading from '../components/loading/loading';
```
dmitrylyzo, line 1: Why did this file become unnecessary?
How do we list playlists?

thornbill: I'm honestly not entirely sure... this has always been around but I don't think we have ever used it. Playlists are listed using the normal list controller. It's possible it was already dead at the time of the fork. 🤷‍♂️

src/components/playlisteditor/playlisteditor.ts
```diff
@@ -127,8 +127,6 @@ function triggerChange(select: HTMLSelectElement) {
 function populatePlaylists(editorOptions: PlaylistEditorOptions, panel: DialogElement) {
     const select = panel.querySelector<HTMLSelectElement>('#selectPlaylistToAddTo');
 
-    loading.hide();
```
dmitrylyzo: I wonder if `show` should be here (before the change). :thinking: 
If yes, it must go after `if (!select)`.

I just don't see which `show` corresponds to the `hide` below.

src/components/playlisteditor/playlisteditor.ts
```diff
@@ -0,0 +1,319 @@
...
+    const itemIds = dlg.querySelector<HTMLInputElement>('.fldSelectedItemIds')?.value || '';
+
+    if (id === 'queue') {
+        playbackManager.queue({
+            serverId: currentServerId,
+            ids: itemIds.split(',')
+        });
+        dlg.submitted = true;
+        dialogHelper.close(dlg);
+        return;
```
dmitrylyzo, line 108: We probably won't hide `loading` in `finally` because we're not returning `Promise` here.

src/components/playlisteditor/playlisteditor.ts
```diff
@@ -0,0 +1,319 @@
...
+
+    if (panel) {
+        const playlistId = panel.querySelector<HTMLSelectElement>('#selectPlaylistToAddTo')?.value;
+
+        loading.show();
+
+        if (playlistId) {
+            userSettings.set('playlisteditor-lastplaylistid', playlistId);
+            addToPlaylist(panel, playlistId)
+                ?.catch(err => {
```
dmitrylyzo, line 53: Doesn't a "failed" optional chain make `finally` unreachable?
I think if we guarantee Promise will be returned, we can drop the optional chaining.

src/components/playlisteditor/playlisteditor.ts
thornbill: Yeah that seems likely... although in my testing it is hidden before I see it anyway

src/components/playlisteditor/playlisteditor.ts
thornbill: Good call. Should be fixed now.

dmitrylyzo (approved):
LGTM. Except maybe too many optional chaining when requesting HTML elements - imo, over-protection since we generate them in this controller.

thornbill:
It might be over cautious but we need to handle it for typescript in some way since the functions _can_ return null 🤷‍♂️

PR #5393, opened 2024-04-17
Update logo assets

nielsvanvelzen:
**Changes**

Update the banner-light and banner-dark assets to the new Jellyfin logo. The dark variant is only used in the appletv theme only. The light variant is used in the header, splash screen and logo screensaver. I've checked all of these places for correctness and everything looks fine, even though the sizes changed a bit.

Ideally we'd use SVG but I'm sure there is a webos/tizen related reason for using bitmaps. I've exported the bitmaps from our Figma and optimized them with optipng.

**Screenshots**

917000032b
Purple Haze

975000032c
Apple TV

**Issues**


erikbucik (commented):
Cool, I'm just thinking if maybe I made the text too small.

thornbill:
Unfortunately we have to support back to the equivalent of Chrome 27 to support older TV platforms.

nielsvanvelzen:
The logo is also not added as an `` element but as a background-image via CSS. So we'd need to change that first. This PR is aimed to be as simple as possible so it won't introduce regression potentially and hold up the 10.9 release.

thornbill (approved):

PR #5394, opened 2024-04-18
Use % for playback slider marker postions

enter-a-random-username:
**Changes**
Use correct position, even on large windows resize and all cases. The value exists. if the offset of width is the issue. Use CSS calc. `'calc(' + valueMarker + '% - ' + markerRect.width / 2 + 'px)'`

**Issues**
Currently the position is often off windows with `px` values on GT webkit/blink and so on.

### DIFF
--- a/src/elements/emby-slider/emby-slider.js
+++ b/src/elements/emby-slider/emby-slider.js
@@ -239,7 +239,7 @@ function setMarker(range, valueMarker, marker, valueProgress) {
             return;
         }
 
-        marker.style.left = 'calc(' + valueMarker + '% - ' + markerRect.width / 2 + 'px)';
+        marker.style.left = `calc(${valueMarker}% - ${markerRect.width / 2}px)`;
 
         if (valueProgress >= valueMarker) {
             marker.classList.remove('unwatched');


src/elements/emby-slider/emby-slider.js
```diff
@@ -190,10 +190,7 @@ function setMarker(range, valueMarker, marker, valueProgress) {
             return;
         }
 
-        let markerPos = (bubbleTrackRect.width * valueMarker / 100) - markerRect.width / 2;
-        markerPos = Math.min(Math.max(markerPos, - markerRect.width / 2), bubbleTrackRect.width - markerRect.width / 2);
-
-        marker.style.left = markerPos + 'px';
+        marker.style.left = 'calc(' + valueMarker + '% - ' + markerRect.width / 2 + 'px)';
```
grhallenbeck, line 193: nit: this would be easier to read, and probably a bit safer, with template literals
```suggestion
        marker.style.left = `calc(${valueMarker}% - ${markerRect.width / 2}px)`;
```

Template literals ensure the values are strings, and help avoid unintentional math. This is especially true in cases like this where the string itself contains arithmetic characters, so using `+` for concatenation is a little less clear.

src/elements/emby-slider/emby-slider.js
```diff
@@ -190,10 +190,7 @@ function setMarker(range, valueMarker, marker, valueProgress) {
             return;
         }
 
-        let markerPos = (bubbleTrackRect.width * valueMarker / 100) - markerRect.width / 2;
-        markerPos = Math.min(Math.max(markerPos, - markerRect.width / 2), bubbleTrackRect.width - markerRect.width / 2);
-
-        marker.style.left = markerPos + 'px';
+        marker.style.left = `calc(${valueMarker}% - ${markerRect.width / 2}px)`;
```
nielsvanvelzen, line 193: I don't think we can use the CSS calc function as this won't work on the older browsers we need to support

dmitrylyzo: We want to support webOS 1.2 and Tizen 2.3. Both are old WebKit which is currently approximated with Chrome 27. It's not perfect, and sometimes it's better to test in an emulator.

thornbill (approved):

PR #5399, opened 2024-04-20
Remove OS check for VPP option

gnattu:
**Changes**

This removed the Linux OS check for displaying VPP Tone Mapping options, as the OS info is no longer available, and we don't agree with adding them back.

This option currently does nothing on Windows, so it is safe to be displayed

**Issues**

Closes jellyfin/jellyfin#11388

Fixes jellyfin/jellyfin#11382

### DIFF
--- a/src/controllers/dashboard/encodingsettings.js
+++ b/src/controllers/dashboard/encodingsettings.js
@@ -229,7 +229,7 @@ $(document).on('pageinit', '#encodingSettingsPage', function () {
             page.querySelector('.videoToolboxTonemappingOptions').classList.add('hide');
         }
 
-        if (systemInfo.OperatingSystem.toLowerCase() === 'linux' && (this.value == 'qsv' || this.value == 'vaapi')) {
+        if (this.value == 'qsv' || this.value == 'vaapi') {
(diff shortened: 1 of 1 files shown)


src/controllers/dashboard/encodingsettings.js
```diff
@@ -19,6 +19,7 @@ function loadPage(page, config, systemInfo) {
     page.querySelector('#chkHardwareEncoding').checked = config.EnableHardwareEncoding;
     page.querySelector('#chkAllowHevcEncoding').checked = config.AllowHevcEncoding;
     page.querySelector('#chkAllowAv1Encoding').checked = config.AllowAv1Encoding;
+    page.querySelector('#chkAllowMjpegEncoding').checked = config.AllowMjpegEncoding;
```
nyanmisaka, line 22: I thought this and L131 was removed before?

gnattu: Rebase issue, that one is out of tree for a long time and only merged 9 hours ago.

nyanmisaka (approved):

thornbill (approved):

PR #5401, opened 2024-04-20
Prevent changing volume with mouse wheel when dialog is shown

gnattu:
**Changes**

The dialog on the player could present a large list, requiring scrolling to find the desired option, such as a subtitle or audio track in a huge language list. Changing the volume during scrolling in these cases will be very annoying.

a2f1e1b0b23af18fdca024da2b173ccf

**Issues**

### DIFF
--- a/src/controllers/playback/video/index.js
+++ b/src/controllers/playback/video/index.js
@@ -1351,6 +1351,7 @@ export default function (view) {
     }
 
     function onWheel(e) {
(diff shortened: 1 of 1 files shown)


dmitrylyzo (approved):
At first I thought we could stop event propagation in the dialog. But volume changing when moving the cursor out of the dialog would still be annoying. So a full blocking seems like a good choice.

PR #5415, opened 2024-04-24
Fix shuffling for shows

andrewrabert:
**Changes**
Fixes shuffle playback of shows.

### DIFF
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -1868,16 +1868,17 @@ class PlaybackManager {
                 }, queryOptions));
             } else if (firstItem.Type === 'Series' || firstItem.Type === 'Season') {
                 const apiClient = ServerConnections.getApiClient(firstItem.ServerId);
+                const isSeason = firstItem.Type === 'Season';
 
                 promise = apiClient.getEpisodes(firstItem.SeriesId || firstItem.Id, {
                     IsVirtualUnaired: false,
                     IsMissing: false,
+                    SeasonId: isSeason ? firstItem.Id : undefined,
                     SortBy: options.shuffle ? 'Random' : undefined,
                     UserId: apiClient.getCurrentUserId(),
                     Fields: ['Chapters', 'Trickplay']
                 }).then(function (episodesResult) {
                     const originalResults = episodesResult.Items;
-                    const isSeries = firstItem.Type === 'Series';
 
                     let foundItem = false;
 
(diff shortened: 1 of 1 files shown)


Bond-009 (approved):

joshuaboniface (approved):

dmitrylyzo:
I wonder how it will find the first unplayed episode in the randomized queue.

src/components/playback/playbackmanager.js
```diff
@@ -1872,6 +1872,7 @@ class PlaybackManager {
                 promise = apiClient.getEpisodes(firstItem.SeriesId || firstItem.Id, {
                     IsVirtualUnaired: false,
                     IsMissing: false,
+                    SortBy: options.shuffle ? 'Random' : 'SortName',
```
thornbill, line 1875: If we get episodes in the correct order by default, it may be better to do this:

```suggestion
                    SortBy: options.shuffle ? 'Random' : undefined,
```

nvllsvm: Changed.

This is reasonable and seems to work. The server should respond with an appropriate default sort order when it's unspecified.

src/components/playback/playbackmanager.js
```diff
@@ -1868,48 +1868,37 @@ class PlaybackManager {
                 }, queryOptions));
             } else if (firstItem.Type === 'Series' || firstItem.Type === 'Season') {
                 const apiClient = ServerConnections.getApiClient(firstItem.ServerId);
+                const isSeason = firstItem.Type === 'Season';
 
                 promise = apiClient.getEpisodes(firstItem.SeriesId || firstItem.Id, {
                     IsVirtualUnaired: false,
                     IsMissing: false,
+                    SeasonId: isSeason ? firstItem.Id : undefined,
```
dmitrylyzo, line 1876: Iirc, the idea was to continue playing the series (by starting a single season), including the next seasons.
Currently, this limits queue to the one season.

nvllsvm: I'd be in favor of adding an explicit option to start playing a series at a particular season. Something like a `Play From Here` in the drop down.
image

I do not think it should be the default however, maybe a library-wide option instead? Automatically starting the next season doesn't always make sense when, for example, someone clicks play on a `Specials` season. The expectation as of 10.8.13 is that only the `Specials` season would play. 
image

Also doesn't make sense for shows with self-contained seasons.
image

dmitrylyzo: Related PR #4827

nvllsvm: I didn't realize this, but there's already an option to `Play all from here` shown on individual episodes in individual seasons. That option could likely be added to show on the season's drop down as well - thus implementing the behavior you're talking about.

PR #5423, opened 2024-04-26
Fix playback sequence of episodes in a series/season

andrewrabert:
**Changes**
- Adds an option to a season's menu to `Play All From Here` (default behavior is to play only the specific season)
- Playback will start on the first unplayed episode, or at the beginning of the season
- Fixes the previous episode OSD button. (Ex. starting playback on episode 2 will now allow the previous button to navigate to episode 1).

**Issues**
Relates to https://github.com/jellyfin/jellyfin-web/pull/5415#pullrequestreview-2020486301

### DIFF
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -1858,7 +1858,7 @@ class PlaybackManager {
                     }, queryOptions));
                 case 'Series':
                 case 'Season':
-                    return getSeriesOrSeasonPlaybackPromise(firstItem, options);
+                    return getSeriesOrSeasonPlaybackPromise(firstItem, options, items);
                 case 'Episode':
                     return getEpisodePlaybackPromise(firstItem, options, items);
             }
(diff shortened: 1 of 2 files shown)


1hitsong:
> Adds an option to a season's menu to Play All From Here (default behavior is to play only the specific season)

Confirmed. I see the option directly under the Play option.

> Playback will start on the first unplayed episode, or at the beginning of the season

Setup:
EP1: Unwatched
EP2: Watched

Outcome:
Playback starts with EP1

---
Setup:
EP1: Watched
EP2: Unwatched

Outcome:
Playback starts with EP2

---
Setup:
EP1: Watched
EP2: Watched

Outcome:
Playback starts with EP1

> Fixes the previous episode OSD button. (Ex. starting playback on episode 2 will now allow the previous button to navigate to episode 1).

Setup:
EP1: Watched
EP2: Unwatched

If I use the Play from here option, playback begins with EP2 and I can click the back button in the OSD to play EP1.

If I go into the season and press play on EP1, I can click the forward button to play EP2. I can then click the back in the OSD to play EP1.

However, if I go into the season and press play on EP2, if I click the back in the OSD nothing happens.

Is this expected behavior keeping this PR in mind?

nvllsvm:
> If I go into the season and press play on EP1, I can click the forward button to play EP2. I can then click the back in the OSD to play EP1.
> 
> However, if I go into the season and press play on EP2, if I click the back in the OSD nothing happens.

Fixed with https://github.com/jellyfin/jellyfin-web/pull/5423/commits/1586db13069829c9193cb5640fbdf9b24b3aa807

1hitsong:
> > However, if I go into the season and press play on EP2, if I click the back in the OSD nothing happens.
> 
> Fixed with [1586db1](https://github.com/jellyfin/jellyfin-web/commit/1586db13069829c9193cb5640fbdf9b24b3aa807)

✅ Confirmed that given this scenario, pressing the back button from direct playing episode 2 now moves playback to episode 1.

src/components/playback/playbackmanager.js
```diff
@@ -1868,38 +1868,43 @@ class PlaybackManager {
...
-                                return true;
+                                episodesResult.StartIndex = index;
+                                break;
                             }
-
-                            return false;
-                        });
+                        }
+                        if (episodesResult.StartIndex === undefined) {
+                            episodesResult.StartIndex = startSeasonId ? seasonStartIndex : 0;
```
1hitsong, line 1902: Is this line testing the correct variable? Should it be testing `seasonStartIndex`?

nvllsvm: It should work as I had it, but I guess it's possible that `seasonStartIndex` could never get set for some reason. Changed to https://github.com/jellyfin/jellyfin-web/pull/5423/commits/8fcc38a7bcb8b6fd0751c6e6e43abf7027601eac

1hitsong: All my tests still work as expected and this was the only line of code that gave me pause.

1hitsong (approved):

PR #5424, opened 2024-04-26
Redirect legacy URLs

### LINKED ISSUE #5388: Previously bookmarked URL fails to redirect properly
**Describe The Bug**

I have a bookmark for my Jellyfin instance of which the path is `/web/index.html#!/home.html`. When I open this bookmark in 10.9 it redirects to `/web/index.html#/!/home.html` which is invalid (should be `/web/index.html#/home.html`). The page will not load.

**Steps To Reproduce**
1. Open a new tab in your browser that does not have jellyfin-web loaded
2. Set the URL to `/web/index.html#!/home.html`
3. ??
4. Profit

**Expected Behavior**

Either the redirect should be fixed or the "wrong" URL should be loaded correctly.

**Logs**

No relevant logs found

**Screenshots**

Untitled

(Guess which theme I use)

 - Platform: Windows
 - Browser: Firefox 124.0.2
 - Jellyfin Version: 10.9 (web f2a5163096d2ced007ce96b0c4e2f711a5a02a75)

**Additional Context**

Only thing keeping me from using Jellyfin

andrewrabert:
**Changes**
Fixes a blank screen when using bookmarks made with older versions of Jellyfin (pre v10.9)

**Issues**
Resolves https://github.com/jellyfin/jellyfin-web/issues/5388

### DIFF
--- a/src/hooks/useLegacyRouterSync.ts
+++ b/src/hooks/useLegacyRouterSync.ts
@@ -2,8 +2,6 @@ import { Update } from 'history';
 import { useLayoutEffect, useState } from 'react';
 import type { History, Router } from '@remix-run/router';
 
-const normalizePath = (pathname: string) => pathname.replace(/^!/, '');
-
 interface UseLegacyRouterSyncProps {
   router: Router;
   history: History;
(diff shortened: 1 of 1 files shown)


1hitsong:
Confirmed this resolves issue #5388 by properly redirecting to proper home URL.

1hitsong (approved):

nielsvanvelzen:
My issue only mentions `home.html` but this issue exists for all routes. I think a fix that works for all of them would be better. If you follow my reproduction steps in the comment but open a different page before changing the URL to unstable demo the issue would still exist.

thornbill:
This _should_ be handled here: https://github.com/jellyfin/jellyfin-web/blob/master/src/hooks/useLegacyRouterSync.ts#L23

I'm not sure why that is no longer working... perhaps an update to react router broke it.

nvllsvm:
I think I fixed the problem in the existing redirect logic to now properly handle all legacy URLS (https://github.com/jellyfin/jellyfin-web/pull/5424/commits/3a0329fa90b21cb09acbe461985b8b169149246a)

nielsvanvelzen:
This appears to work right now when loading the page. Changing the fragment on an existing page (adding the `!` before the first `/`) still breaks.

nvllsvm:
@nielsvanvelzen Confirmed and fixed.

nielsvanvelzen (approved):

PR #5427, opened 2024-04-28
Use NormalizationGain instead of LUFS

Bond-009:
Web side for https://github.com/jellyfin/jellyfin/pull/11457

PR #5428, opened 2024-04-28
 Fall back to other normalization gain if preferred one is null 

Bond-009:
Depends on #5427

### DIFF
--- a/src/plugins/htmlAudioPlayer/plugin.js
+++ b/src/plugins/htmlAudioPlayer/plugin.js
@@ -113,12 +113,19 @@ class HtmlAudioPlayer {
             let val = options.url;
             console.debug('playing url: ' + val);
             import('../../scripts/settings/userSettings').then((userSettings) => {
-                if (userSettings.selectAudioNormalization() == 'TrackGain'
-                    && (options.item.NormalizationGain != null || options.mediaSource.albumNormalizationGain != null)) {
-                    self.gainNode.gain.value = Math.pow(10, ((options.item.NormalizationGain ?? options.mediaSource.albumNormalizationGain) / 20));
-                } else if (userSettings.selectAudioNormalization() == 'AlbumGain'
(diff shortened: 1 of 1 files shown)


src/plugins/htmlAudioPlayer/plugin.js
```diff
@@ -113,16 +113,16 @@ class HtmlAudioPlayer {
...
-                    console.debug('[HtmlAudioPlayer] Using album gain');
+                if (userSettings.selectAudioNormalization() == 'TrackGain'
+                    && (options.item.NormalizationGain != null || options.mediaSource.albumNormalizationGain != null)) {
+                    self.gainNode.gain.value = Math.pow(10, ((options.item.NormalizationGain ?? options.mediaSource.albumNormalizationGain) / 20));
+                } else if (userSettings.selectAudioNormalization() == 'AlbumGain'
+                    && (options.mediaSource.albumNormalizationGain != null || options.item.NormalizationGain != null)) {
+                    self.gainNode.gain.value = Math.pow(10, ((options.mediaSource.albumNormalizationGain ?? options.item.NormalizationGain) / 20));
                 } else {
                     self.gainNode.gain.value = 1;
                 }
```
nielsvanvelzen, line 124: This should achieve the same but is cleaner to read.

```suggestion
let normalizationGain;
if (userSettings.selectAudioNormalization() == 'TrackGain') {
	normalizationGain =
		options.item.NormalizationGain ??
		options.mediaSource.albumNormalizationGain;
} else if (userSettings.selectAudioNormalization() == 'AlbumGain') {
	normalizationGain =
		options.item.albumNormalizationGain ??
		options.mediaSource.NormalizationGain;
}

if (normalizationGain) {
	self.gainNode.gain.value = Math.pow(10, normalizationGain / 20);
} else {
	self.gainNode.gain.value = 1;
}
```

(not indented)

nielsvanvelzen (approved):

Bond-009:
Confirmed working :+1:

thornbill (approved):

PR #5432, opened 2024-04-29
Change "setup" to "set up" in source language

systemcrash:


src/strings/en-gb.json
```diff
@@ -1196,7 +1196,7 @@
     "HeaderSpecialEpisodeInfo": "Special Episode Info",
     "HeaderSortOrder": "Sort Order",
     "HeaderSortBy": "Sort By",
-    "HeaderSetupLibrary": "Setup your media libraries",
+    "HeaderSetupLibrary": "Set up your media libraries",
```
nielsvanvelzen, line 1199: Changes to translations should be made directly in https://translate.jellyfin.org/

systemcrash: Absolutely right. Should. I tried and it was uppgefuckad, so here you have it instead.

nielsvanvelzen: Not sure what you mean, translations should not be updated via pull requests as this often breaks our translation tool. Only edits in `en-us.json` (the source language) can be done in a pull request.

systemcrash: Ah well. 50% of the way there 😅

nielsvanvelzen (approved):

PR #5434, opened 2024-04-30
Add setting to prioritise what audio codec to transcode to

scampower3:
Add a setting to prioritise what audio codec to transcode to.

**Changes**

- Adds a dropdown option to select what audio codec is preferred. (Options avaliable are abit arbitrary)
- Shift the codec to the start of the string at getDeviceProfile function, will do nothing if the codec is not part of the transcodingProfile.AudioCodec string.

**Issues**

Helps resolve https://github.com/jellyfin/jellyfin-tizen/issues/204


src/scripts/settings/appSettings.js
```diff
@@ -132,6 +132,13 @@ class AppSettings {
         return toBoolean(this.get('limitSupportedVideoResolution'), false);
     }
 
+    selectPreferredTranscodeAudio(val) {
```
dmitrylyzo, line 140: Maybe just `preferredTranscodeAudio`?
The same for the key in the storage.

src/components/playbackSettings/playbackSettings.template.html
```diff
@@ -157,6 +157,17 @@ <h2 class="sectionTitle">
         <div class="selectContainer">
             <select is="emby-select" class="selectSkipBackLength" label="${LabelSkipBackLength}"></select>
         </div>
+        <div class="selectContainer">
+            <select is="emby-select" id="selectPreferredTranscodeAudio" label="${LabelSelectPreferredTranscodeAudio}">
+                <option value="auto">${Auto}</option>
```
dmitrylyzo, line 162: Can we use empty string as key for `Auto`?

dmitrylyzo (commented):
I wonder if audio codecs can simply be ordered by quality from best to worst?
The server will then select the first (best) one supported by the client and allowed by the server for transcoding.

src/components/playbackSettings/playbackSettings.template.html
scampower3: as in <option value="auto">${Auto}</option> to become <option value="auto"></option>?

dmitrylyzo: No
`<option value="">${Auto}</option>`

src/components/apphost.js
```diff
@@ -42,6 +42,7 @@ function getDeviceProfile(item) {
         }
 
         const maxVideoWidth = appSettings.maxVideoWidth();
+        const preferredTranscodeAudio = appSettings.preferredTranscodeAudio();
```
dmitrylyzo, line 45:
```suggestion
```

src/components/apphost.js
```diff
@@ -70,6 +71,20 @@ function getDeviceProfile(item) {
             });
         }
 
+        if (preferredTranscodeAudio !== '') {
```
dmitrylyzo, line 74: I would move the variable closer to where it is used and remove the strict comparison.
```suggestion
        const preferredTranscodeAudio = appSettings.preferredTranscodeAudio();

        if (preferredTranscodeAudio) {
```

src/scripts/settings/appSettings.js
```diff
@@ -132,6 +132,18 @@ class AppSettings {
         return toBoolean(this.get('limitSupportedVideoResolution'), false);
     }
 
+    /**
+     * Get or set 'Preferred Transcode Audio' state.
+     * @param {string|undefined} val - Flag to set 'Preferred Transcode Audio' or undefined.
+     * @return {string} 'Preferred Transcode Audio' state.
```
dmitrylyzo, line 138: "state" is more for flags (checkboxes).
```suggestion
     * Get or set preferred transcode audio.
     * @param {string|undefined} val - Preferred transcode audio or undefined.
     * @return {string} Preferred transcode audio.
```

dmitrylyzo (commented):
Should we add "codec" to the name? 🤔

src/strings/en-us.json
```diff
@@ -553,6 +553,7 @@
     "LabelAudioCodec": "Audio codec",
     "LabelAudioLanguagePreference": "Preferred audio language",
     "LabelSelectAudioNormalization": "Audio Normalization",
+    "LabelSelectPreferredTranscodeAudioCodec": "Preferred transcode audio codec",
```
gnattu, line 556: Please be explicit that this only affects Audio track in Videos. Because the Audio transcoding will use a very different server-side logic.

src/components/playbackSettings/playbackSettings.template.html
```diff
@@ -157,6 +157,17 @@ <h2 class="sectionTitle">
         <div class="selectContainer">
             <select is="emby-select" class="selectSkipBackLength" label="${LabelSkipBackLength}"></select>
         </div>
+        <div class="selectContainer">
+            <select is="emby-select" id="selectPreferredTranscodeAudioCodec" label="${LabelSelectPreferredTranscodeAudioCodec}">
+                <option value="">${Auto}</option>
+                <option value="aac">AAC</option>
+                <option value="ac3">AC3</option>
+                <option value="dts">DTS</option>
+                <option value="eac3">EAC3</option>
```
gnattu, line 166: Just a note: EAC3 encoding is not enabled on server side yet, if this is ever merged before the server-side change, then this option has to be removed.

src/strings/en-us.json
dmitrylyzo: What if we add a header?
```html
    <div class="verticalSection verticalSection-extrabottompadding">
        <h2 class="sectionTitle">
            ${HeaderVideoAdvanced}
        </h2>

        <!-- select goes here -->

    <div>
```
Not sure about wording.

src/components/playbackSettings/playbackSettings.template.html
```diff
@@ -157,6 +157,17 @@ <h2 class="sectionTitle">
...
+                <option value="flac">FLAC</option>
+                <option value="opus">Opus</option>
```
gnattu, line 168: ALAC is also a supported codec in 10.9:
```suggestion
                <option value="opus">Opus</option>
                <option value="alac">ALAC</option>
```

src/components/playbackSettings/playbackSettings.template.html
dmitrylyzo: This is just a hint. It has no effect if the codec is missing/disallowed on the  server. Probably should add a description to the field about this.

src/strings/en-us.json
gnattu: If we are not rushing this with 10.9 release, then we can do a re-organize of the transcoding panel to make it easier to follow. It currently has a lot of options inside and a lot of options are actually very hard to find.

dmitrylyzo: Yes, ideally there should be a dedicated view to customize the device profile.

The problem with the 10.9 release is that we can't add translations to the release branch - Weblate can only push to the master branch, iirc.

dmitrylyzo (approved):
Seems to work.

`*AudioCodecInVideo` could be renamed to `*VideoAudioCodec` (_CodecProfiles for audio in video are marked with `VideoAudio`_), but that's just nitpicking.

SplitMilky:
Where is this option?  

Is it possible to encode all audio to ac3 regardless if the player says it can play?  I have an issue where android 9 doesn't pass the info correctly on aac and a few of the higher end audio codecs.  This issue combines with the TV not having e-arc . 

So many of the audios are recognized as playable, and they really are and should be but android 9 messes something up and much of it comes out as TV Stereo.

PR #5437, opened 2024-04-30
Fix usages of item in location state

### LINKED ISSUE #5357: Display mirroring and theme media broken
In 10.9 all routes are served through the React router. This introduces regression in the themeMediaPlayer and displayMirrorManager plugins. Those rely on item data via the `viewshow` event, which is no longer set (always undefined).

While debugging this issue I could trace it back to the React router, in ViewManagerPage we set the state for the event from `location.state`, which is undefined. It is set in the show function (via showItem) in the appRouter though.

My suspicion is that the `createHashHistory` function does not allow state or there is some bug in it. No idea. Not familiar with React to debug that part. One possible solution is to update the two plugins to not rely on this data anymore.

## Trace

Only place that sets the "state" is the showItem function, which adds the complete BaseItemDto.

https://github.com/jellyfin/jellyfin-web/blob/f6bc10206eac4c905fdf864aa0182114b00d33ec/src/components/router/appRouter.js#L197

It is then send to the history.push function via the options parameter in the show function.

https://github.com/jellyfin/jellyfin-web/blob/f6bc10206eac4c905fdf864aa0182114b00d33ec/src/components/router/appRouter.js#106

Then, the useLocation() function is a reactive property that should contain it as state in ViewManagerPage, and the state is set in the event data (viewOptions) where the value is always undefined.

https://github.com/jellyfin/jellyfin-web/blob/f6bc10206eac4c905fdf864aa0182114b00d33ec/src/components/viewManager/ViewManagerPage.tsx#L38

thornbill:
**Changes**
* Moves the `useItem` hook to a separate file and add support for calling from outside react
* Updates appRouter, displayMirrorManager, and themeMediaPlayer to use query function

**Issues**
Fixes #5357

### DIFF
--- a/src/components/playback/displayMirrorManager.ts
+++ b/src/components/playback/displayMirrorManager.ts
@@ -1,38 +1,41 @@
-import type { BaseItemDto } from '@jellyfin/sdk/lib/generated-client';
+import ServerConnections from 'components/ServerConnections';
+import { getItemQuery } from 'hooks/useItem';
+import { toApi } from 'utils/jellyfin-apiclient/compat';
+import { queryClient } from 'utils/query/queryClient';
 
 import { playbackManager } from './playbackmanager';
 
-interface PlaybackInfo {
-    item: BaseItemDto;
-    context?: string;
-}
-
-function mirrorItem(info: PlaybackInfo, player?: unknown) {
-    const { item } = info;
-
-    playbackManager.displayContent({
-        ItemName: item.Name,
-        ItemId: item.Id,
-        ItemType: item.Type,
-        Context: info.context
-    }, player);
-}
-
-function mirrorIfEnabled(info: PlaybackInfo) {
-    if (info && playbackManager.enableDisplayMirroring()) {
+async function mirrorIfEnabled(serverId: string, itemId: string) {
+    if (playbackManager.enableDisplayMirroring()) {
         const playerInfo = playbackManager.getPlayerInfo();
 
         if (playerInfo && !playerInfo.isLocalPlayer && playerInfo.supportedCommands.indexOf('DisplayContent') !== -1) {
-            mirrorItem(info, playbackManager.getCurrentPlayer());
+            const apiClient = ServerConnections.getApiClient(serverId);
+            const api = toApi(apiClient);
+            const userId = apiClient.getCurrentUserId();
+
+            try {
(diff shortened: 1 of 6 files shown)


nielsvanvelzen:
Testing results:

- ✅ display mirroring
- ❌ themes, tried on a tv series with theme videos (and the preference for it enabled)
	`[ThemeMediaPlayer] failed to load theme media TypeError: undefined has no properties`
	appears to be at this line (compiled code): ` u = _.currentSettings.enableThemeVideos() &&`

src/components/playback/displayMirrorManager.ts
```diff
@@ -1,38 +1,41 @@
...
+            }
         }
     }
 }
 
 document.addEventListener('viewshow', e => {
-    const state = e.detail.state || {};
-    const { item } = state;
+    if (e.detail?.params?.id && e.detail?.params?.serverId) {
+        const { serverId, id } = e.detail.params;
```
dmitrylyzo, line 37: Can we do this instead of chaining that much?
```suggestion
    const { serverId, id } = e.detail?.params || {};
    if (id && serverId) {
```

src/components/themeMediaPlayer.js
```diff
@@ -50,44 +58,61 @@ function stopIfPlaying() {
...
+        if (item.CollectionType) {
+            stopIfPlaying();
+            return;
+        }
 
-    if (excludeTypes.indexOf(item.Type) !== -1) {
-        stopIfPlaying();
-        return;
-    }
+        if (excludeTypes.indexOf(item.Type) !== -1) {
```
dmitrylyzo, line 94:
```suggestion
        if (excludeTypes.includes(item.Type)) {
```

src/components/themeMediaPlayer.js
```diff
@@ -50,44 +58,61 @@ function stopIfPlaying() {
...
 }
 
-document.addEventListener('viewshow', function (e) {
-    const state = e.detail.state || {};
-    const item = state.item;
-
-    if (item?.ServerId) {
-        loadThemeMedia(item);
+document.addEventListener('viewshow', e => {
+    if (e.detail?.params?.id && e.detail?.params?.serverId) {
```
dmitrylyzo, line 113: As above.

nielsvanvelzen (approved):

PR #5441, opened 2024-05-01
Fix unplayed count styling when navigating back

andrewrabert:
Fixes the style of the unplayed count indicator in certain situations.

**Repoducing**
1. Have a season that is marked as played.
3. Navigate to the season's page
4. Mark an episode as unwatched
5. Navigate back to the previous page to see the styling issue.

### DIFF
--- a/src/components/cardbuilder/cardBuilder.js
+++ b/src/components/cardbuilder/cardBuilder.js
@@ -1285,8 +1285,7 @@ function updateUserData(card, userData) {
 
         if (!playedIndicator) {
             playedIndicator = document.createElement('div');
-            playedIndicator.classList.add('playedIndicator');
-            playedIndicator.classList.add('indicator');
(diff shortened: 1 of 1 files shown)


nielsvanvelzen (approved):

src/components/cardbuilder/cardBuilder.js
```diff
@@ -1303,6 +1303,7 @@ function updateUserData(card, userData) {
         if (!countIndicator) {
             countIndicator = document.createElement('div');
             countIndicator.classList.add('countIndicator');
+            countIndicator.classList.add('indicator');
```
thornbill, line 1306: I believe you could pass both classes in a single add call

nvllsvm: Yup - that works. Updated.

thornbill (approved):

PR #5442, opened 2024-05-01
Show "99+" in unplayed indicator when more 100 or more new items

andrewrabert:
The unplayed count does not fully fit inside the indicator when there are 1000 or more items. This PR changes the indicator to show "+" for any count greater or equal to 1000.

**Before**
before

**After**
after

### DIFF
--- a/src/components/indicators/indicators.js
+++ b/src/components/indicators/indicators.js
@@ -78,7 +78,7 @@ export function getPlayedIndicatorHtml(item) {
     if (enablePlayedIndicator(item)) {
         const userData = item.UserData || {};
         if (userData.UnplayedItemCount) {
-            return '<div class="countIndicator indicator">' + datetime.toLocaleString(userData.UnplayedItemCount) + '</div>';
+            return '<div class="countIndicator indicator">' + formatCountIndicator(userData.UnplayedItemCount) + '</div>';
         }
 
         if (userData.PlayedPercentage && userData.PlayedPercentage >= 100 || (userData.Played)) {
@@ -93,10 +93,14 @@ export function getChildCountIndicatorHtml(item, options) {
     const minCount = options?.minCount ? options.minCount : 0;
 
     if (item.ChildCount && item.ChildCount > minCount) {
-        return '<div class="countIndicator indicator">' + datetime.toLocaleString(item.ChildCount) + '</div>';
+        return '<div class="countIndicator indicator">' + formatCountIndicator(item.ChildCount) + '</div>';
     }
 
     return '';
+}
+
+function formatCountIndicator(count) {
+    return count >= 100 ? '99+' : count.toString();
 }
 
 export function getTimerIndicator(item) {
--- a/src/components/indicators/useIndicator.tsx
+++ b/src/components/indicators/useIndicator.tsx
@@ -60,6 +60,10 @@ const enablePlayedIndicator = (item: ItemDto) => {
     return itemHelper.canMarkPlayed(item);
 };
 
+const formatCountIndicator = (count: number) => {
+    return count >= 100 ? '99+' : count.toString();
+};
+
 const useIndicator = (item: ItemDto) => {
     const getMediaSourceIndicator = () => {
         const mediaSourceCount = item.MediaSourceCount ?? 0;
@@ -135,7 +139,7 @@ const useIndicator = (item: ItemDto) => {
         if (childCount > 1) {
             return (
                 <Box className='countIndicator indicator childCountIndicator'>
-                    {datetime.toLocaleString(item.ChildCount)}
+                    {formatCountIndicator(childCount)}
                 </Box>
             );
         }
@@ -149,7 +153,7 @@ const useIndicator = (item: ItemDto) => {
             if (userData.UnplayedItemCount) {
                 return (
                     <Box className='countIndicator indicator unplayedItemCount'>
-                        {datetime.toLocaleString(userData.UnplayedItemCount)}
+                        {formatCountIndicator(userData.UnplayedItemCount)}
                     </Box>
                 );
             }


src/components/indicators/indicators.js
```diff
@@ -93,12 +93,16 @@ export function getChildCountIndicatorHtml(item, options) {
...
     if (item.ChildCount && item.ChildCount > minCount) {
-        return '<div class="countIndicator indicator">' + datetime.toLocaleString(item.ChildCount) + '</div>';
+        return '<div class="countIndicator indicator">' + formatCountIndicator(item.ChildCount) + '</div>';
     }
 
     return '';
 }
 
+function formatCountIndicator(count) {
+    return count >= 1000 ? '1k+' : count.toString();
```
dmitrylyzo, line 103: Should `1k` be translatable?

dmitrylyzo:
There is another count indicator (for React): https://github.com/jellyfin/jellyfin-web/blob/850ac3837a2b470cc4dbe034bb34c7886faae063/src/components/indicators/useIndicator.tsx#L138

src/components/indicators/indicators.js
nvllsvm: Nope - the `k` suffix is an international standard.

dmitrylyzo: In Russian is `к` for "kilo".

nvllsvm:
@dmitrylyzo Updated.

src/components/indicators/indicators.js
nvllsvm: Ahh, good to know. I changed the indicator to be just `+`.

dmitrylyzo (approved):

dmitrylyzo:
We could show `999` when `count` is 999+ though. 🤷

src/components/indicators/indicators.js
scampower3: > In Russian is `к` for "kilo".

I mean tbf k is used to represent a 1000 cuz it stands for kilo and in SI standards kilo is 1000 :/

dmitrylyzo: > I mean tbf 1k is used to represent a 1000 cuz the k stands for kilo and in SI standards kilo is 1000. IMO, I think 1k is a perfectly fine way to represent 1000 cuz of that.

But in Russian texts, the Latin alphabet is not used - in Russian standards `к` is for `кило` ("kilo").

I may be wrong in the current case, as SI doesn't completely fit to it. For example, there is no one-character notation for billion (`1b`). At least I'm not aware of it. "Giga" (`Г` in Russian standards) doesn't work well.

I checked the "views" on YouTube, they use `тыс.` for thousands as the `Intl.NumberFormat` component does.
```
new Intl.NumberFormat('ru', { notation: 'compact' }).format(1000);       // "1 тыс."
new Intl.NumberFormat('ru', { notation: 'compact' }).format(1000000);    // "1 млн"
new Intl.NumberFormat('ru', { notation: 'compact' }).format(1000000000); // "1 млрд"
```

We could use `999+` (I believe I saw that somewhere), but that's a bit long for indicator.
`999` could work as if the 3-digit counter has reached its limit.

As a joke, we could use `∞` (infinity sign). :upside_down_face:

scampower3: That's fair enough.

thornbill:
Personally I think I prefer just the "+" out of the given options... I don't like losing 100-999 if we use 99+ and just using 999 for >999 seems misleading.

dmitrylyzo:
> Personally I think I prefer just the "+" out of the given options... I don't like losing 100-999 if we use 99+ and just using 999 for >999 seems misleading.

`+` just looks like you can add something. That's why I started thinking about what else we could use.

Shadowghost:
I agree with @dmitrylyzo, the `+` looks more like a clickable element. I'd argue that 99+ is used by a lot of other services

dmitrylyzo:
I prefer the infinity sign (`∞`) - it allows 1..999 and has no meaning other than "too many".

erikbucik:
The `+` sign means "add" and makes the indicator look like a small button, which it isn't.

The problem is that visually the label should have a two character maximum; it needs some padding otherwise it looks like a mistake (as seen with both `99+` and `999` in the example above).
Keep in mind that after the redesign we'll change the indicator's shape from a circle to a triangle, where we'll encounter the same problem.
new indicator examples



Like @thornbill said, using `999` for `>=999` is misleading.

The most common across UIs is `99+` because it's easy to understand; it means more than 99 items. (But again, it's perhaps too wide.)

And I agree that the infinity sign is the best looking one, even though it's too ambiguous and not the best choice in terms of UX accessibility.

erikbucik:
Then again, the amount of **unplayed** episodes isn't really that important, so we could get away with just showing a solid triangle. It means the show/series hasn't been watched completely, and that' it.

nvllsvm:
Changed to "99+"

dmitrylyzo (approved):

nielsvanvelzen (approved):

berdugorn:
how can this be disabled or adjusted?
It is a bad idea to show 99+ gor any 3 digit number. the 1K+ or infinity should be better for the shows with more than a thousand unwatched episodes...

kjdix13:
I want to be able to revert this as well.

jonsextra:
I'd also like this reverted. There's currently no quick way to get an accurate count of the number of episodes in a series.

zollo991:
Yeah, that would be really sweet if it could be changed. Would make things a lot easier.

gerbenvandijk:
Would like this to be optional at least, the UX is bad now. Been trying to sync my watch status with my old Plex install and its currently impossible to easily see how many episodes are unwatched on a per-show basis.

OdinVex:
> Would like this to be optional at least, the UX is bad now. Been trying to sync my watch status with my old Plex install and its currently impossible to easily see how many episodes are unwatched on a per-show basis.

Countdown to comment "minification" like everyone else: 3...2...1...

PR #5444, opened 2024-05-02
[en / i18n] grammar: less -> fewer

systemcrash:


src/strings/en-us.json
```diff
@@ -131,7 +131,7 @@
     "ButtonUninstall": "Uninstall",
     "ButtonUseQuickConnect": "Use Quick Connect",
     "ButtonWebsite": "Website",
-    "Bwdif": "BWDIF",
+    "Bwdif": "BWDIF - Bob Weaver DeInterlacing Filter",
```
nielsvanvelzen, line 134: The hardware acceleration options are formatted differently, better to keep it consistent.

```suggestion
    "Bwdif": "Bob Weaver DeInterlacing Filter (BWDIF)",
```

systemcrash: OK. Understood. Done.

nielsvanvelzen (approved):

PR #5452, opened 2024-05-05
Limit maximum FLAC audio channels in video to 2 on webOS

dmitrylyzo:
webOS doesn't seem to support FLAC with more than 2 channels in video.

**Changes**
- Limit maximum FLAC channels to 2 on webOS.
- Split each video transcoding profile with FLAC so that the containing FLAC is only applied to 2 channels audio.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-webos/issues/146

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -899,6 +899,30 @@ export default function (options) {
             Codec: 'flac',
             Conditions: flacConditions
         });
+
+        const flacTranscodingProfiles = [];
+
+        // Split each video transcoding profile with FLAC so that the containing FLAC is only applied to 2 channels audio
+        profile.TranscodingProfiles.forEach(transcodingProfile => {
+            if (transcodingProfile.Type !== 'Video') return;
+
+            const audioCodecs = transcodingProfile.AudioCodec.split(',');
+
+            if (!audioCodecs.includes('flac')) return;
+
+            const flacTranscodingProfile = { ...transcodingProfile };
+            flacTranscodingProfile.AudioCodec = 'flac';
+            flacTranscodingProfile.ApplyConditions = [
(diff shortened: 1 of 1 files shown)


ilmc888:
Great, thanks. Now that 10.9 is out, I can test this later this week if needed.

ilmc888:
A quick test seems to show that FLAC 7.1 is transcoded to FLAC. I think it is properly downmixed to 2.0 now, but the expected behavior is to transcode to 5.1/7.1 AAC (channels having priority over the format).

dmitrylyzo:
> A quick test seems to show that FLAC 7.1 is transcoded to FLAC. I think it is properly downmixed to 2.0 now, but the expected behavior is to transcode to 5.1/7.1 AAC (channels having priority over the format).

Yeah, that would be preferable.
I guess, the server is selecting FLAC to FLAC, just ignoring the max channels. And then it applies the limit.

We need to make changes to the server so that it tries not to change the number of channels (maybe not only).

dmitrylyzo:
Now I'm using `ApplyConditions` so that the video transcoding profile with FLAC is only applied to 2-channel audio.

ilmc888:
I can confirm that video with FLAC 7.1 is now transcoded to multichannel AAC and FLAC 2.0 is played directly (at least for music, I don't have video with FLAC 2.0).

The FLAC 5.1 test file (surround5.1.flac) doesn't play correctly as a music file. Only left and right front channels are heard.

In any case, this PR solves the problem I was having with video + FLAC.

dmitrylyzo:
Since there are [some uncertainties](https://github.com/jellyfin/jellyfin-webos/issues/146#issuecomment-2121242017) with FLAC as music, let's leave it as is and at least fix the video.

thornbill (approved):

PR #5460, opened 2024-05-07
Dynamic generate selectPersonType Options

scampower3:
**Changes**

Dynamically generate selectPersonType options from the SDK.
See: https://github.com/jellyfin/jellyfin-web/pull/5251#pullrequestreview-2041595431

**Issues**

### DIFF
--- a/src/components/metadataEditor/personEditor.template.html
+++ b/src/components/metadataEditor/personEditor.template.html
@@ -12,33 +12,7 @@
         </div>
 
         <div class="selectContainer">
-            <select is="emby-select" id="selectPersonType" class="selectPersonType" label="${LabelType}">
-                <option value=""></option>
-                <option value="Actor">${Actor}</option>
-                <option value="Composer">${Composer}</option>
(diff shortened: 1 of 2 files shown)


src/components/metadataEditor/personEditor.template.html
```diff
@@ -12,33 +12,7 @@ <h3 class="formDialogHeaderTitle">
         </div>
 
         <div class="selectContainer">
-            <select is="emby-select" id="selectPersonType" class="selectPersonType" label="${LabelType}">
```
thornbill, line 15: This should not have been removed right?

scampower3: Isn't removed. <select is="emby-select" id="selectPersonType" class="selectPersonType" label="${LabelType}"></select> is still there.

thornbill: Hah whoops missed that last line. 😂

thornbill (approved):

PR #5471, opened 2024-05-09
Fix publish workflow for non-prs

thornbill:
**Changes**
Fix the publish workflow not running on non-PR events

**Issues**
N/A

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -8,12 +8,37 @@ on:
       - completed
 
 jobs:
+  pr-context:
+    name: PR context
+    if: ${{ github.event.workflow_run.event == 'pull_request' }}
+    runs-on: ubuntu-latest
+    outputs:
+      branch: ${{ env.pr_branch }}
+      commit: ${{ env.pr_sha }}
+      pr_number: ${{ env.pr_number }}
+
+    steps:
+      - name: Get PR context
+        uses: dawidd6/action-download-artifact@09f2f74827fd3a8607589e5ad7f9398816f540fe # v3.1.4
+        id: pr_context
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -8,12 +8,37 @@ on:
       - completed
 
 jobs:
+  pr-context:
```
thornbill, line 11: This was re-ordered just to match the order they now run in

.github/workflows/publish.yml
```diff
@@ -8,12 +8,37 @@ on:
...
+          echo "pr_number=$(cat PR_number)" >> $GITHUB_ENV
+          echo "pr_sha=$(cat PR_sha)" >> $GITHUB_ENV
+
   publish:
     permissions:
       contents: read
       deployments: write
 
     name: Deploy to Cloudflare Pages
+    if: ${{ always() }}
```
thornbill, line 41: Don't skip this event if the `pr-context` job is skipped

.github/workflows/publish.yml
```diff
@@ -35,34 +60,10 @@ jobs:
           apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
           accountId: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
           projectName: jellyfin-web
-          branch: ${{ needs.pr-context.outputs.branch }}
+          branch: ${{ needs.pr-context.outputs.branch || github.ref_name }}
```
thornbill, line 63: Fallback to the current `ref_name` if not a PR event.

crobibero (approved):

PR #5473, opened 2024-05-10
Don’t set save lyrics locally as default

gnattu:
This option requires write permission to the media folder which should not be enabled by default.

**Changes**

**Issues**

See jellyfin/jellyfin#11523


dmitrylyzo:
Shouldn't we do the same with subtitles?

gnattu:
I think we should, but niels does not agree as we are saving subtitles with media as default for years.

Shadowghost:
I'd still argue that for consistency's sake we should set both to the same value

thornbill:
If the defaults need changed that discussion should happen on the server side. This PR only updates web to align with the server values.

PR #5476, opened 2024-05-10
Add configuration for scan fanout concurrency

thornbill:
**Changes**
* Adds configuration of the scan fanout concurrency  to the general settings page in the dashboard

Screenshot 2024-05-10 at 11-45-41 Jellyfin

**Issues**
N/A


nielsvanvelzen (approved):

nielsvanvelzen:
Perhaps we could move the warning to a paragraph below the "Performance" heading.

thornbill:
Yeah we could... I expect the network file systems part is more exclusive to scans though :man_shrugging:

thornbill:
I'm going to merge this as is so we have more time for translations... we can improve later if needed.

PR #5492, opened 2024-05-13
Fix chapter markings not displayed properly

Schoggi0815:
**Changes**
Chapter markers weren't always displayed. rarely they were there, most of the times they weren't.
This was due to a race condition, which also caused them to be only properly set for the first video to be shown.
This fixes the markers and makes sure they're always properly displayed.

**Issues**


thornbill:
Would you be comfortable rebasing this and changing the target branch to `release-10.9.z`? If not, we can just land this in master first and go from there. Thanks!

Schoggi0815:
Created a new pull-request on the `release-10.9.z` branch
#5519

PR #5495, opened 2024-05-13
Add finer playback speed between 0.75x and 1.25x (in 0.05 increments)

Maatss:
**Changes**
Adds additional 8 playback speeds for finer playback control. Speeds are between 0.75x and 1.25x in 0.05 increments.

**Feature request**
https://features.jellyfin.org/posts/2662/finer-playback-speeds-0-95x-1x-1-05x-1-1x

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -87,6 +87,7 @@
 - [JPUC1143](https://github.com/Jpuc1143)
 - [David Angel](https://github.com/davidangel)
 - [Pithaya](https://github.com/Pithaya)
+- [Maatss](https://github.com/Maatss)
 
 ## Emby Contributors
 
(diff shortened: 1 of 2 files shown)


Maatss:
Given the increased number of playback speed options, it would look nicer if the selection dialog was given a max-height attribute, and thus introducing a scroll bar. However, I could not find an easy way to do this..

ferferga:
I'd say that doing it as a combobox (like in jellyfin-vue) would provide a far better UX.

Maatss:
@ferferga I have not seen that client so not entirely sure what you mean. This is the difference of just introducing a `max-height: 30%` attribute to the dialog for reference.

Current:
Current_layout

w/ max-height: 30%
suggested_layout

PR #5500, opened 2024-05-13
Fix stored credentials not updating on logout

thornbill:
**Changes**
Fixes the updated saved server credentials not being persisted to local storage when a user logs out

**Issues**
Fixes https://github.com/jellyfin/jellyfin/issues/11598


nielsvanvelzen:
Shouldn't this target the release branch?

thornbill:
It doesn't _really_ matter since we can go either direction, but yes. I didn't think about it until after I opened the PR and I have not had the chance to update it yet. :sweat_smile:

nielsvanvelzen (approved):

Issue #5485, opened 2024-05-09
Syncplay sessions close after playing video for 30 seconds


soultaco83:
### Please describe your bug

Moving from https://github.com/jellyfin/jellyfin-web/issues/5446
When playing video from a syncplay session. The Sessions will close and stop playing

### Reproduction Steps

1: Start a syncplay session
2: Play any video file
3: Wait 30 seconds
4: Stream crashes and backs to previous page

### Jellyfin Version

Unstable (master branch)

### Environment

```markdown
- OS: Unraid
- Linux Kernel: Linux 6.7.12-Unraid x86_64
- Virtualization: Docker
- Clients: Browser
- Browser: Chrome, Firefox, Edge
- FFmpeg Version: 6.0.1-6
- Playback Method: Doesn't seem to be related
- Hardware Acceleration: QSV
- GPU Model: Arc A770
- Reverse Proxy: n/a
- Base URL: n/a
```

### Jellyfin logs

```shell
Attached below are server logs and snippet of console
Debug enabled



-05-09 03:18:31.074 -04:00] [DBG] [21] Jellyfin.Api.Auth.CustomAuthenticationHandler: AuthenticationScheme: "CustomAuthentication" was successfully authenticated.
[2024-05-09 03:18:31.083 -04:00] [INF] [21] Emby.Server.Implementations.SyncPlay.Group: Session "167f9f73419f326929282b1271a2ff32" requested Play in group "0210dd87-f79b-4b5a-9f5e-697e852764cc" that is Idle.
[2024-05-09 03:18:31.083 -04:00] [INF] [21] Emby.Server.Implementations.SyncPlay.Group: Group "0210dd87-f79b-4b5a-9f5e-697e852764cc" switching from Idle to Waiting.
[2024-05-09 03:18:31.085 -04:00] [DBG] [21] MediaBrowser.Controller.Entities.BaseItem: "test" has no parental rating set.
```
...
```
[2024-05-09 03:18:31.102 -04:00] [DBG] [21] MediaBrowser.Controller.SyncPlay.GroupStates.WaitingGroupState: Session "167f9f73419f326929282b1271a2ff32" set a new play queue in group "0210dd87-f79b-4b5a-9f5e-697e852764cc".
```
...
```

[...]

PR #5501, opened 2024-05-14
Fixed Sync Play

PeachesMLG:
Fixed issue with SyncPlay crashing after 30 seconds

**Changes**
I honestly have no idea why this broke anything, but this commit https://github.com/jellyfin/jellyfin-web/commit/1b2c99a4eca58c7f468ea9454a5a7c1127bcafd9 changed the timings of the events, so we were waiting for the `playbackstart` event after it had already been called. I tried looking into why it was happening in the first place, but after a couple hours i gave up and decided it was just magic. If anyone wants to look into why this happened feel free to close this and raise your own pr.

**Issues**
https://github.com/jellyfin/jellyfin-web/issues/5485


PeachesMLG:
Looks like this didnt resolve the issues, I suspect there are some timing issues going on with two async events being run concurrently where they need to be awaiting on each other, ill investigate this a bit further

Issue #5487, opened 2024-05-13
Synced lyrics doesn't autoscroll


Extarys:
### Please describe your bug

When using Jellyfin Web UI (same on desktop), the synched lyrics doesn't auto scroll.

The highlighted line should always be in the centre of the viewport IMO, or at least be visible at the bottom. See spotify (or Finamp)

(Thanks for this oh-so-awesome update, btw! <3)

### Reproduction Steps

1. Play a song with synched lyrics
2. Click the new mega awesome button to display lyrics
3. skip mid-way or wait until the highlighted line is below viewpoint

### Jellyfin Version

10.9.0

### Environment

```markdown
Client: Web, Desktop Linux
Browser: Firefox
```

### Jellyfin logs

```shell
Not needed
```

### FFmpeg logs

```shell
Not needed
```

### Please attach any browser or client logs here

Not needed

### Please attach any screenshots here

![image](https://github.com/jellyfin/jellyfin/assets/11581433/b9c86509-2d48-481f-a6c8-0f53aa2416d5)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

thornbill:
This was not included as part of the initial implementation so I’m tagging it as an enhancement.

Reapor-Yurnero:
I see this is implemented recently. Is it incorporated in the latest releast i.e. 10.9.7? On my end for songs with synced lyrics they are still displayed in plain text as follows without any scrolling:
```
[00:01.900]Lyrics by：Robert Thiele/George David Weiss
[00:02.860]Composed by：Robert Thiele/George David Weiss
[00:09.820]I see trees of green red roses too
```

thornbill:
New features go in major releases so the upcoming 10.10.0 release in this case.

Reapor-Yurnero:
Okay then this might be a bug?

Does my above lrc format look the same to yours?

PR #5502, opened 2024-05-14
Add scrolling of lyrics during music playback

RemainingStart:
**Changes**
Adds scrolling of lyrics during music playback.

**Enhancement**
#5487


Shadowghost:
I'd consider this a bugfix which should be PRed against 10.9.z release branch

dmitrylyzo:
This can be fun in TV display mode - we also scroll into view when navigating through the buttons.

Issue #5068, opened 2024-01-01
Player halts when playing any transcoded media on webOS 23


lawadr:
**Describe The Bug**

On a new (2023) LG 65" C3 (webOS 23) I am unable to play anything that is transcoded. No matter what the video or audio codec is, if it is coming from an HLS stream via transcoding it won't play correctly.

If I play from the start of the stream, everything works fine. If I skip ahead to an already transcoded time (within the red bar on the server's dashboard), it works fine. But if I skip ahead too far (past the red bar), or just start playing part way through, it bugs out.

When it bugs out, the screen updates to the correct frame, but playback doesn't carry on from that point. If I skip back or forward from then on, the same thing occurs. I can see the correct still frames, but no playback. It's like it's permanently paused. Sometimes I'll hear a couple of seconds of audio before it stops.

Oddly, I see no error in the Jellyfin server logs. The transcoding works fine and continues as normal while the playback has halted in the client. The client's Playback Info shows no dropped or corrupted frames, and the transcoding progress ticks upwards, but only if I repeatedly pause and unpause the stream (while it remains halted on screen).

This behaviour also occurs using the web interface in the webOS browser, so it isn't specific to the webOS client. I can't reproduce this anywhere outside of webOS though.

This issue should be reproducible every time by forcing any sort of transcoding. As I can't seem to find any related bug report for such a common use case, I'm thinking that this is either a regression or a webOS 23 specific issue. 

**Steps To Reproduce**

1. Find a movie with an image based subtitle track (e.g. PGS or DVDSUB)
2. Disable the subtitle tracks and play it
3. Skip ahead half way through the movie and stop (go back to Jellyfin)
4. Turn on the image based subtitle track and continue playing from where you left off

or...

1. Play any movie
2. Skip ahead half way through the movie
3. Change the quality to something like 720p
4. Stop (go back to Jellyfin)
5. Continue playing from where you left off

**Expected Behavior**

Plays the stream correctly

**System**

 - Platform: webOS 23
 - Jellyfin Version: 10.8.13

lawadr:
This still occurs. I've found quite a few mentions of this issue, especially in relation to DoVi. It seems to be a problem when the "prefer fMP4" option is enabled. Transcoding seems to work fine with this option turned off.

Here are some related links:
https://github.com/jellyfin/jellyfin-webos/issues/208
https://github.com/jellyfin/jellyfin-webos/issues/207
[...]

PR #5553, opened 2024-05-18
Fix Dolby Vision playback on webOS

GeorgeH005:
Force DoVi profile 8 support for LG TVs (CX etc.) that do support it but report otherwise.

**Changes**
browserDeviceProfile to force profile 8 support for TVs that support DoVi and run web0s version 5 and up.

**Issues**

Fixes
jellyfin/jellyfin#11705

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -247,9 +247,13 @@ function supportedDolbyVisionProfilesHevc(videoTestElement) {
             .replace(/no/, '')) {
             supportedProfiles.push(5);
         }
-        if (videoTestElement
-            .canPlayType('video/mp4; codecs="dvh1.08.09"')
-            .replace(/no/, '')) {
+        if (
+            videoTestElement
+                .canPlayType('video/mp4; codecs="dvh1.08.09"')
+                .replace(/no/, '')
+            // LG TVs from at least 2020 onwards should support profile 8, but they don't report it.
+            || (browser.web0sVersion >= 4)
+        ) {
             supportedProfiles.push(8);
         }
     }
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -1263,17 +1263,16 @@ export default function (options) {
     });
 
     if (browser.web0s && supportsDolbyVision(options)) {
-        // Disallow direct playing of DOVI media in containers not mp4.
-        // This paired with the "Prefer fMP4-HLS Container" client playback setting enables DOVI playback on webOS.
+        // Disallow direct playing of DOVI media in containers not ts or mp4.
         profile.CodecProfiles.push({
             Type: 'Video',
-            Container: '-mp4',
+            Container: '-mp4,ts',
             Codec: 'hevc',
             Conditions: [
                 {
                     Condition: 'EqualsAny',
                     Property: 'VideoRangeType',
-                    Value: 'SDR|HDR10|HLG',
+                    Value: hevcVideoRangeTypes.split('|').filter((v) => !v.startsWith('DOVI')).join('|'),
                     IsRequired: false
                 }
             ]
--- a/src/scripts/settings/userSettings.js
+++ b/src/scripts/settings/userSettings.js
@@ -1,7 +1,7 @@
-import appSettings from './appSettings';
-import browser from '../browser';
 import Events from '../../utils/events.ts';
 import { toBoolean } from '../../utils/string.ts';
+import browser from '../browser';
+import appSettings from './appSettings';
 
 function onSaveTimeout() {
     const self = this;
@@ -142,7 +142,7 @@ export class UserSettings {
         }
 
         // Enable it by default only for the platforms that play fMP4 for sure.
-        return toBoolean(this.get('preferFmp4HlsContainer', false), browser.safari || browser.firefox || browser.chrome || browser.edgeChromium);
+        return toBoolean(this.get('preferFmp4HlsContainer', false), browser.safari || browser.firefox || (browser.chrome && !browser.web0s && !browser.tizen) || browser.edgeChromium);
     }
 
     /**


src/scripts/browserDeviceProfile.js
```diff
@@ -247,9 +247,13 @@ function supportedDolbyVisionProfilesHevc(videoTestElement) {
...
         }
-        if (videoTestElement
-            .canPlayType('video/mp4; codecs="dvh1.08.09"')
-            .replace(/no/, '')) {
+        if (
+            videoTestElement
+                .canPlayType('video/mp4; codecs="dvh1.08.09"')
+                .replace(/no/, '')
+            // LG TVs from at least 2020 onwards should support profile 8, but they don't report it.
+            || (browser.web0s && browser.web0sVersion > 5)
```
dmitrylyzo, line 255: Testing for `web0sVersion` is enough (if not webOS, it is undefined and the check will be false).
```suggestion
            || browser.web0sVersion > 5
```

So webOS 5 doesn't support DoVi, but its updates (5.1, 5.2; _I don't know the exact versions) do, right?

gnattu:
We might need additional workarounds for webOS. A lot of users reported that webOS has problem seeking with fMP4 container and webOS can playback hevc in TS container as well. My question now would be that does webOS playback dolby vision in hls with ts container as well? If that is the case we may override the default container to prefer ts on webOS and push hevc support for ts containers.

854562:
This PR aims at WebOS >=5, but my C9 (2019) running WebOS 4.9.7 seems to play DoVi profile 8 content just fine (using LG's internal media player). In mp4 that is, as with remuxing to TS using ffmpeg I'm hitting the same roadblock as @GeorgeH005.

854562:
> Just making sure, it doesn't currently work with jellyfin v10.9.2, correct?

That's correct. A remux is triggered for mkv files but the player falls back to HDR10 anyway.

lgndvd:
> @gnattu Just occured to me, is it possible that the problematic models have an issue with `dvh1.08.09` and could be ok with something like `dvh1.08.06`? I know that it isn't unreasonable to assume that there are going to be 60fps media somewhere, and 4K media surpass the 25mbps "Main Tier" bitrate (Dolby Spec) all the time, but couldn't we have that as a fallback if it does actually report correctly? Just to avoid any hacky overrides.

Hi @GeorgeH005, some further info:

- just checked my DV8 files and Mediainfo reports all of them as `Dolby Vision, Version 1.0, Profile 8.1, dvhe.08.06, BL+RPU`, with minor differences about HDR10 / SMPTE compatibility;
- re-checked DV8 support on my LG CX playing some files locally via USB: can confirm native support for TS and MP4 files.

I also found some [discussions on GitHub](https://github.com/gpac/gpac/issues/1991#issuecomment-1006878183) (mentioned also on jellyfin-web repo), about different DV8 profiles and LG OLEDs behaviour (C8, CX).

Hope any of these helps.

gnattu:
> @lgndvd I will take a look at that discussion. My problem at the moment is the fact that I can't seem to get ffmpeg to remux an mkv containing Dolby Vision to mpegts without losing the DoVi metadata. If you remuxed those files yourself I would love to read how you did it.

If you use jellyfin-ffmpeg you should not lose anything?

lgndvd:
> @lgndvd I will take a look at that discussion. My problem at the moment is the fact that I can't seem to get ffmpeg to remux an mkv containing Dolby Vision to mpegts without losing the DoVi metadata. If you remuxed those files yourself I would love to read how you did it.

Yes, I do them with [tsMuxerGUI nightly](https://github.com/justdan96/tsMuxer/releases): shortly, I drag and drop my MKV DV sources, select video+audio tracks to be included, choose "TS muxing" as output and, usually in a couple of minutes, I get my working TS file. This output file direct plays flawlessly both locally and with Jellyfin 10.8.x.

I do some additional work outside of tsMuxer only when muxing from multiple sources and/or dealing with DTS audio (not supported by CX). I also don't include subtitles, because afaik TS supports only bitmap, so I pass them as external SRTs.

I also experimented with some mp4 workflows (using ffmpeg GUIs), and at some point had success...but the process was usually slower and generally more cumbersome when dealing with multiple tracks metadat...so I went back to my good old TS workflow (quicker, idiot-proof).

gnattu:
@GeorgeH005 Well you should not use a static file for testing because we applied the dolby vision data copy for hlsenc:  https://github.com/jellyfin/jellyfin-ffmpeg/blob/jellyfin/debian/patches/0033-add-hevc-dovi-sidedata-to-hlsenc.patch

Did you try jellyfin-ffmpeg on jellyfin with remuxing to ts? Did that work?

lgndvd:
> @gnattu Using jellyfin-ffmpeg I still can't get an mpegts file with DoVi metadata. Here is what I get using ``-vbsf hevc_mp4toannexb -c copy -sn -strict -2 <filename>.ts``:
> 
> 
> 
> ```
> 
> Video
> 
> ID                                       : 256 (0x100)
> 
> Menu ID                                  : 1 (0x1)
> 
> Format                                   : HEVC
> 
> Format/Info                              : High Efficiency Video Coding
> 
> Format profile                           : Main 10@L5.1@Main
> 
> Codec ID                                 : 36
> 
> Duration                                 : 47 min 22 s
> 
> Bit rate                                 : 19.2 Mb/s
> 
> Width                                    : 3 840 pixels
> 
> Height                                   : 2 160 pixels
> 
> Display aspect ratio                     : 16:9
> 
> Frame rate                               : 50.000 FPS
> 
> Color space                              : YUV
> 
> Chroma subsampling                       : 4:2:0
> 
> Bit depth                                : 10 bits
> 
> Bits/(Pixel*Frame)                       : 0.046
> 
> Stream size                              : 6.36 GiB (94%)
> 
> Color range                              : Full
> 
> ```
> 
> 
> 
> Converting with the same exact arguments to mp4 gives me this:
> 
> 
> 
> ```
> 
> Video
> 
> ID                                       : 1
> 
> Format                                   : HEVC
> 
> Format/Info                              : High Efficiency Video Coding
> 
> Format profile                           : Main 10@L5.1@Main
> 
> HDR format                               : Dolby Vision, Version 1.0, Profile 5, dvhe.05.09, BL+RPU, no metadata compression
> 
> Codec ID                                 : hev1
> 
> Codec ID/Info                            : High Efficiency Video Coding
> 
> Duration                                 : 47 min 22 s
> 
> Source duration                          : 47 min 22 s
> 
> Bit rate                                 : 19.2 Mb/s
> 
> Width                                    : 3 840 pixels
> 
> Height                                   : 2 160 pixels
> 
> Display aspect ratio                     : 16:9
> 
> Frame rate mode                          : Variable
> 
> Frame rate                               : 50.000 FPS
> 
> Minimum frame rate                       : 50.000 FPS
> 
> Maximum frame rate                       : 16 000.000 FPS
> 
> Color space                              : YUV
> 
> Chroma subsampling                       : 4:2:0
> 
> Bit depth                                : 10 bits
> 
> Bits/(Pixel*Frame)                       : 0.046
> 
> Stream size                              : 6.36 GiB (96%)
> 
> Source stream size                       : 6.36 GiB (96%)
> 
> Color range                              : Full
> 
> Menus                                    : 3
> 
> mdhd_Duration                            : 2842280
> 
> Codec configuration box                  : hvcC+dvcC
> 
> ```
> 
> 
> 
> If you have any suggestions as to which arguments I should use, I would be very happy to hear them. Didn't get a chance to play around with the codec tags by the way. Will try to see if safari can play back anything tagged hev1 or dvhe first and test at the Tv tomorrow.

@GeorgeH005 when dealing with ffmpeg, I learned that `-strict unofficial` works best when dealing with Dolby Vision Metadata. From what I'm reading, `-strict -2` should be the equivalent of `-strict experimental` value.

If I remember correctly, I used `-strict unofficial` to get some results in FFMPEG Batch Tool Converter (even if regarded mp4 testing), otherwise I wouldn't get any DV metadata.

Tomorrow I could take a look to my past FFMPEG arguments, if needed. Please consider that my knowledge is very limited, just sharing what I learned with reading, trial and error during last year.

gnattu:
@GeorgeH005 @lgndvd I looked back at the FFmpeg code and found that `mpegtsenc.c` does not handle Dolby Vision data. So, remuxing using TS on the fly is currently out of the question. However, we still need to enable support for videos in TS containers because many TVs support them. Additionally, a lot of TS files are actually in M2TS mode, which is used by Blu-ray discs. Blocking these files will, force a remux that could be avoided since they can be directly played.

nyanmisaka:
@GeorgeH005 Can you try this new jellyfin-ffmpeg? https://github.com/jellyfin/jellyfin-ffmpeg/pull/393

lgndvd:
> @GeorgeH005 @lgndvd I looked back at the FFmpeg code and found that `mpegtsenc.c` does not handle Dolby Vision data. So, remuxing using TS on the fly is currently out of the question. However, we still need to enable support for videos in TS containers because many TVs support them, and currently mp4 is the only allowed container. Additionally, a lot of TS files are actually in M2TS mode, which is used by Blu-ray discs. Blocking these files will, force a remux that could be avoided since they can be directly played.

Hi, just finished doing some additional testing:

**FFmpeg:** unfortunately, can confirm that TS and M2TS containers do not retain Dolby Vision metadata when using FFmpeg. Tested using version `6.1.1 Essentials x64` (within FFmpeg Batch AV Converter 3.0.5 on Windows) and different arguments, but per MediaInfo it's always a no-go.

**tsMuxer:** in addition to TS remuxing, I tested also M2TS. Dolby Vision metadata still gets injected (both Profile 5 and 8), but playback on LG CX had some weird behaviour. I remuxed three .m2ts files, one with DV5 metadata and two with DV8 metadata:

1. The DV5 one played correctly via USB, with Dolby Vision triggering.
2. One of the DV8 files didn't start, giving me infinite spinning wheel (after some time, "file format not recognized"). For this one, MediaInfo reports: `Dolby Vision, Version 1.0, Profile 8.1, dvhe.08.06, BL+RPU, extended metadata compression, HDR10 compatible / SMPTE ST 2094 App 4, Version HDR10+ Profile B, HDR10+ Profile B compatible`
3. The other one with DV8 metadata (but slightly different MediaInfo: `Dolby Vision, Version 1.0, Profile 8.1, dvhe.08.06, BL+RPU, extended metadata compression, HDR10 compatible / SMPTE ST 2086, Version HDR10, HDR10 compatible`) played correctly via USB.
- Direct playing via DLNA worked instead flawlessly with _all_ .m2ts files, triggering Dolby Vision. I tested DLNA from two different sources, Jellyfin and router NAS sharing.

**Recap and conclusions:** LG CX supports DV playback natively via MP4 (confirmed), TS (confirmed), maybe also M2TS (mixed results, needs further testing)...but FFmpeg seems unable to remux into `mpegts` retaining DoVi metadata.

To sum it up, as ideal behaviour Jellyfin should be able to:
- remux DoVi .mkv files into something that webOS / LG OLED clients can digest (probably MP4, due to the above issues...maybe one day also TS?);
- do this for both DV5 and DV8 metadata, fixing current 10.9.x DV8 regression (aka getting HDR fallback for _any_ container, even previously direct playing ones in 10.8.x);
- possibly, skip remuxing for DoVi .mp4/.ts files _(potentially also .m2ts)_ due to native support.

I don't know if any of these are viable, but I hope so. Keep up the good work!

nyanmisaka:
> Hi, just finished doing some additional testing:
> 
> **FFmpeg:** unfortunately, can confirm that TS and M2TS containers do not retain Dolby Vision metadata when using FFmpeg. Tested using version `6.1.1 Essentials x64` (within FFmpeg Batch AV Converter 3.0.5 on Windows) and different arguments, but per MediaInfo it's always a no-go.

@lgndvd Two of our custom patches are not available in vanilla FFmpeg to support DOVI in MPEG-TS & HLS.

- [0033-add-hevc-dovi-sidedata-to-hlsenc.patch](https://github.com/jellyfin/jellyfin-ffmpeg/blob/8492f082199a26e43191606e165c9e7cadb5b4d4/debian/patches/0033-add-hevc-dovi-sidedata-to-hlsenc.patch)
- [0074-add-dovi-descriptor-support-in-mpegtsenc.patch](https://github.com/jellyfin/jellyfin-ffmpeg/blob/8492f082199a26e43191606e165c9e7cadb5b4d4/debian/patches/0074-add-dovi-descriptor-support-in-mpegtsenc.patch)

You can try the latest test build here (https://github.com/jellyfin/jellyfin-ffmpeg/pull/393/checks -> Artifacts).

gnattu:
It is an LG problem. Samsung's Tizen also have similar compatibility issues with fmp4 container. So if we want to support smart TVs better we will want to use ts as default container I think.

nyanmisaka:
> @gnattu Codec tags make no difference. What I did notice though is that the problem occurs only if you try to seek to a region of the video that has not been remuxed yet. So for example (at around x640 speed) if exactly after starting the video I try to jump to the 3min mark it will work as expected. However if I try to seek to the end immediately it will hang, even if I try to seek to the (for example) 2 min mark afterwards. I am not familiar enough with our streaming code to diagnose this further, and as far as I can tell it does not occur on safari. If you could point me in the right direction I will be happy to test some more.
> 
> PS. If the file is already an mp4 (so no remuxing happening) there is no problem with seeking

It may be a problem with the fMP4 player itself. HLS.js had similar issues in the past, getting stuck after seeking.

nyanmisaka:
> > It may be a problem with the fMP4 player itself. HLS.js had similar issues in the past, getting stuck after seeking.
> 
> WebOS has native hls support though, doesn't it? If that is really the case is there something we can do?
> 
> PS. Yes it does and we already use it... I am going to skim over the code but I doubt I will find something given the amount of time I have available to invest into this.

I mean you probably need to use TS container instead. Because the native player on TV is not open source. It can only be *fixed* on the vendor side.

nyanmisaka:
> > @gnattu @nyanmisaka LG seem to have their own api for this: `mediaOption` which seems easy enough to use. I will play around with this if we are willing to add a "Hack" just for webos, and if it actually is advantageous.
> 
> Couldn't get it to make a difference. TS it is I guess. Should I change the default in this PR?

Just do it. Also since fMP4 seeking is partially broken on LG, you can also disable it.

Does the server side need to change to accommodate this behavior?

nyanmisaka:
> @nyanmisaka Your ffmpeg patches should be enough on their own. I don't think there is anything to be done server side.

I will tag a new jf-ffmpeg version for the next point release.

gnattu:
So basically, we need to:

- default to not remux to fmp4 on smart TVs (Tizen and webOS), both can do HEVC in ts
- Let webOS request remux to ts container for dolby vision videos

nyanmisaka:
> No it is not... Is there any way to force certain range types to remux to ts, even if prefer fmp4 is enabled? Or should we just have it disabled by default and use the changes above?

Doesn't the fMP4 issue happen with SDR videos? IMO if TS is good enough for SDR/HDR/DV + H264/H265, there is absolutely no need to enable it for now.

gnattu:
> No it is not... Is there any way to force certain range types to remux to ts, even if prefer fmp4 is enabled? Or should we just have it disabled by default and use the changes above?

You have to block all playback that is not in ts container to force remux, otherwise it will direct play videos in not compatible containers like mkv and use the HDR10 fallback.

Using the old way change the -mp4 to -ts can do this but it also blocks mp4 direct play, so we need to find a better way handling this.

dmitrylyzo:
> Using the old way change the -mp4 to -ts can do this but it also blocks mp4 direct play, so we need to find a better way handling this.

Why not use `-mp4,ts` (!mp4 && !ts)? With disabled fMP4, it should direct play MP4 and TS and remux to TS.

megosugit:
Hello all,
I went through this thread but I'm not sure if it is directly related to the seek/resume issue on LG TVs with DV movies.
It is mentioned though but I wonder if it clearly related or if it needs a separate issue.

Would this issue fix the following ?
- can't resume/seek some DV movies on LG TV when fMP4 is enabled
- disabling fMP4 fixes the seeking/resuming issue but then DV is not recognized and TV shows washed up colors in movies files not having HDR as fallback. (for movies having both DV and HDR, then TV switches to HDR and colors are OK).
Thanks

854562:
@megosugit This PR combined with the recent jellyfin-ffmpeg update fixes your second issue, since it allows DoVi content to be remuxed to TS and forces certain webOS versions to play it. That way you can safely disable fmp4 without losing out on DoVi.

The first issue is an LG problem and if I'm not mistaken not really feasible to be fixed in Jellyfin.

megosugit:
> @megosugit This PR combined with the recent jellyfin-ffmpeg update fixes your second issue, since it allows DoVi content to be remuxed to TS and forces certain webOS versions to play it. That way you can safely disable fmp4 without losing out on DoVi.
> 
> The first issue is an LG problem and if I'm not mistaken not really feasible to be fixed in Jellyfin.

Thanks, good to know I'm not the only one with this issue :)
Is it supposed to be available in next release (10.9.4) ?

megosugit:
> Sorry for not being active these past few days. It is exam season over here. I will try to implement and test what @dmitrylyzo said during the weekend.

Good luck with your exams :)

megosugit:
Hi guys,

I have seen 10.9.4 being released. from what I can see, this fix has not been implemented, do you confirm ?
I'll try my luck with this new version anyway 🤞

tomislav:
Fun fact, Dolby Vision over TS works on 10.9.7 for me (LG C2), but only if I start the Jellyfin app fresh. If I resume it (wake TV from sleep) it falls back to plain HDR. Then I exit the app, start it again and DV over TS works again (same file).

megosugit:
> Fun fact, Dolby Vision over TS works on 10.9.7 for me (LG C2), but only if I start the Jellyfin app fresh. If I resume it (wake TV from sleep) it falls back to plain HDR. Then I exit the app, start it again and DV over TS works again (same file).

I believe it works for me too (with Prefer fMP4-HLS Media Container disabled).
I haven't noticed the sleep issue (tested on G3).
Would I be able to reproduce your issue If I enter Jellyfin, play a movie, then go back to the TV Home menu without closing the app, then back to Jellyfin ?

tomislav:
> I believe it works for me too (with Prefer fMP4-HLS Media Container disabled). I haven't noticed the sleep issue (tested on G3). Would I be able to reproduce your issue If I enter Jellyfin, play a movie, then go back to the TV Home menu without closing the app, then back to Jellyfin ?

I noticed it happens when I start/resume Jellyfin the next day.

megosugit:
> > I believe it works for me too (with Prefer fMP4-HLS Media Container disabled). I haven't noticed the sleep issue (tested on G3). Would I be able to reproduce your issue If I enter Jellyfin, play a movie, then go back to the TV Home menu without closing the app, then back to Jellyfin ?
> 
> I noticed it happens when I start/resume Jellyfin the next day. Maybe try turning the TV off and back on again :)

on my LG G3, it is weird, sometimes resuming a DV movie works, sometimes it doesn't.
Exiting the app doesn't seem to help in my case.
But it is definitely an improvement compared to previous versions !

viktory36:
Hi @GeorgeH005 , hope you are well

I have a solution locally that works as outlined by Gnattu on May 25 and is implemented as suggested by dmitrylyzo on Jun 3 (using nyanmisaka's ffmpeg fixes).

I am finding that mp4 is direct played, and mkv is remuxed into dovi playback in a TS container (LG C2)

If you can allow me to write to this branch, I'd be happy to update the PR

Cheers

src/scripts/browserDeviceProfile.js
```diff
@@ -1263,17 +1267,18 @@ export default function (options) {
...
                 {
                     Condition: 'EqualsAny',
                     Property: 'VideoRangeType',
-                    Value: 'SDR|HDR10|HLG',
+                    Value: hevcVideoRangeTypes,
                     IsRequired: false
                 }
             ]
         });
+        hevcCodecProfileConditions[2].Value = hevcVideoRangeTypes.split('|').filter((v) => !v.startsWith('DOVI')).join('|');
```
dmitrylyzo, line 1281: I think accessing the condition via an index is unreliable.
Maybe something like:
```js
hevcCodecProfileConditions.find(x => x.Property === 'VideoRangeType')?.Value = ...
```
_Not tested._

nyanmisaka:
> The seeking issue could be related to #3416 if the same problem that occurred in hls.js occurs in smart tv platforms. Just wanted this mentioned here for future reference.

This is no longer an issue with HLS.js. Please report it to LG/webOS.

viktory36:
Hey, tried to post this as an in-line code comment but not sure if that popped up for all. Gonna delete that and add this comment instead:

I couldn't get MKV+DV remux or MP4 direct playback (on the C2) without also specifying the negative container conditions like so:
```
... L1821
if (webos) {
       profile.CodecProfiles.push({
            Type: 'Video',
            Container: '-ts,mp4',
            Codec: 'hevc',
            Conditions: hevcCodecProfileConditions with DOVI removed
        });

    } else {
// the default hevc Codec Profile pushed in all cases
        profile.CodecProfiles.push({
...
```
 Without going into the nitty-gritty of the conditions evaluator, the idea was to build a complete picture of what my CodecProfile should support and found that this works.

Please lmk if DV+MKV/MP4 Direct play works on your base (without any more changes), or consider this suggestion if you experience the same issue.

dmitrylyzo:
> Hey, tried to post this as an in-line code comment but not sure if that popped up for all.

Everyone who is subscribed will be notified.

> I couldn't get MKV+DV remux or MP4 direct playback (on the C2) without also specifying the negative container conditions like so:

Ah yes. 🤦‍♂️ In my variant the second codec profile will block DoVi for any container.

> Without going into the nitty-gritty of the conditions evaluator, the idea was to build a complete picture of what my CodecProfile should support and found that this works.

Without the default HEVC profile, no conditions apply to HEVC in TS/MP4, no?

New plan (adding TS and DoVi filtering to the original @viktory36 code):
```
# Disallow DoVi in non-TS and non-MP4 containers 
CodecProfile {
    Codec: 'hevc',
    Container: '-mp4,ts'
    VideoRangeType: hevcVideoRangeTypes.split(',').filter(r => !r.startsWith('DOVI')).join(',')
}

# default HEVC profile (allow DoVi if supported)
CodecProfile {
    Codec: 'hevc',
    VideoRangeType: hevcVideoRangeTypes
}
```
Iirc, we want to force DoVi without fallback stream, right?

viktory36:
Hi @dmitrylyzo ! hope you are well

Sorry - I clarified the code snippet in my previous comment. I had found that without that exact setup (pass both ts,mp4 and -ts,mp4 && dont apply the default hevc profile (without any container filters) for webos) I would either start remuxing mp4, or transcoding mp4, or direct play mkv. I tried a bunch of combinations and don't remember exactly what led to what outcomes, sorry =]

Please let me know if there are any specific codecprofile setups you want to know the outcome for, and I also really want to know if the problem I'm describing was specific to me or @GeorgeH005 would run into them too (on attempt to play DV MKV/direct play DV MP4)

dmitrylyzo:
@viktory36
I think it can be simplified.
Could you please try [this](https://github.com/jellyfin/jellyfin-web/pull/5553#issuecomment-2211859882)? _I added condition._

viktory36:
Okay, that definitely seems to work fine! Not sure what ghosts I saw on my screen earlier =]

LG C2
SDR MKV: direct play
HDR MVK and MP4: direct play
DOVI MKV: Re-mux
DOVI MP4: Direct play
DOVIWithHDR10 MKV: Re-mux
DOVIWithHDR10 MP4: Direct play

All OK.

src/scripts/browserDeviceProfile.js
```diff
@@ -1259,17 +1263,16 @@ export default function (options) {
     });
 
     if (browser.web0s && supportsDolbyVision(options)) {
-        // Disallow direct playing of DOVI media in containers not mp4.
-        // This paired with the "Prefer fMP4-HLS Container" client playback setting enables DOVI playback on webOS.
+        // Disallow direct playing of DOVI media in containers not ts or mp4.
         profile.CodecProfiles.push({
             Type: 'Video',
-            Container: '-mp4',
+            Container: '-ts,mp4',
```
dmitrylyzo, line 1269: The order doesn't matter, but I think we should stick to alphabetical order.
```suggestion
            Container: '-mp4,ts',
```

dmitrylyzo (commented):
I think it is almost good (just nitpicking). I don't have LG devices, so I can't do any testing of DoVi.

Ideally, there should be 3 commits: `disable fMP4 on webOS and Tizen by default`, `force DoVi8 on webOS >= 4` and `allow DoVi in TS`. But they will be squashed during backporting, so I dunno. 🤷‍♂️

dmitrylyzo:
> Should I try to force push the branch to just those three commits?

That would be nice.

dmitrylyzo (commented):
Remove ` and remux DoVi to TS` in the message of the second commit - it has been moved to the third one.
I would also rephrase the third commit: `Allow Dolby Vision in TS for webOS`.

viktory36:
Yep, everything works fine on the current solution for me.

src/scripts/settings/userSettings.js
```diff
@@ -142,7 +142,7 @@ export class UserSettings {
         }
 
         // Enable it by default only for the platforms that play fMP4 for sure.
-        return toBoolean(this.get('preferFmp4HlsContainer', false), browser.safari || browser.firefox || browser.chrome || browser.edgeChromium);
+        return toBoolean(this.get('preferFmp4HlsContainer', false), browser.safari || browser.firefox || (browser.chrome && !browser.web0s && !browser.tizen) || browser.edgeChromium);
```
nyanmisaka, line 145: fMP4 is enabled by default only for the following browsers. It is unchecked on others.
`(browser.safari || browser.firefox || browser.chrome || browser.edgeChromium)`

Perhaps you want to disable it completely to avoid confusing users?
https://github.com/jellyfin/jellyfin-web/blob/d60ddeb9edeb15d638d281923c06a573a74698e5/src/scripts/browserDeviceProfile.js#L747

dmitrylyzo (approved):
LGTM. We probably need to change the PR title.

src/scripts/settings/userSettings.js
dmitrylyzo: On my Tizen 4 TV, fMP4 requires the transcoding job to complete. But I have no idea if it works on modern models.

This checkbox is part (sort of) of the device profile settings, like `Enable DTS` and `Enable TrueHD`. So I agree with @GeorgeH005, we should leave it up to the user and add a warning.

Btw, Tizen isn't detected as Chrome, so this option wasn't turned on by default.

RohanKrishna500:
hello,i have faced the same issue on my LG A2 OLED and looking for fix or a workaround

MrPeteypoo:
Hey, I think I've been experiencing the same issues with my BDMV backups of Dune and MI Dead Reckoning. Is there any easy way for me to test these fixes in Docker? I'd love to try it on my LG G4

RohanKrishna500:
hi,the issue still persists on my LG A2 TV even after updating to 10.9.8 @GeorgeH005

RohanKrishna500:
> > hi,the issue still persists on my LG A2 TV even after updating to 10.9.8 @GeorgeH005
> 
> The seeking/resume issue or the Profile 8 issue?

profile 8 issue,playback info shows it as its remuxing but the end result is the fallback layer(HDR10)
tested on both jellyfin webos official app and the web page

RohanKrishna500:
> > > > hi,the issue still persists on my LG A2 TV even after updating to 10.9.8 @GeorgeH005
> > > 
> > > 
> > > The seeking/resume issue or the Profile 8 issue?
> > 
> > 
> > profile 8 issue,playback info shows it as its remuxing but the end result is the fallback layer(HDR10)
> > tested on both jellyfin webos official app and the web page
> 
> Test only on the official app. LG's web browser doesn't support Dolby Vision at least when it comes to Jellyfin.
> 
> Can you make sure that in the client's playback settings " Prefer fmp4-hls media container" is UNticked?

i tried on both and yes i have tried the "prefer fmp4" both unticked and ticked
ig if someone has the same model as mine could test and give feedback about that so that i could be sure its dependent on model

RohanKrishna500:
> Contributor

here you go
[MediaFileExample.txt](https://github.com/user-attachments/files/16325046/MediaFileExample.txt)

tomislav:
Are you sure this was included in 10.9.8? The changelog makes no mention of it.

nielsvanvelzen:
It's in the changelog 🤷 

https://github.com/jellyfin/jellyfin-web/releases/tag/v10.9.8

RohanKrishna500:
> > > Contributor
> > 
> > 
> > here you go [MediaFileExample.txt](https://github.com/user-attachments/files/16325046/MediaFileExample.txt)
> 
> Just to disprove an idea, do you have any other Dolby Vision profile 8 files that are NOT 60fps?

yep I have another one and tried it but it was dv profile 7 and not 60fps and the result is the same fallback layer

RohanKrishna500:
> > > > > Contributor
> > > > 
> > > > 
> > > > here you go [MediaFileExample.txt](https://github.com/user-attachments/files/16325046/MediaFileExample.txt)
> > > 
> > > 
> > > Just to disprove an idea, do you have any other Dolby Vision profile 8 files that are NOT 60fps?
> > 
> > 
> > yep I have another one and tried it but it was dv profile 7 and not 60fps and the result is the same fallback layer
> 
> Profile 7 should play as HDR10, because we (and most client devices as far as I am aware) don't handle the extra layers. So that is intended behaviour. Any Profile **8** files?

haven't tried but will do,should i try any specific DV profile(other than 8.1) like profile 5 or something because most of media are in profile 8.1 only

RohanKrishna500:
> PS. I think there are tools that can convert Profile 7 to 8 (maybe by discarding the Enhancement Layer? Not sure) but I wouldn't want to further complicate the testing.

im currently using ddvt tools if i want to convert mkv to mp4 but it becomes quite a hassle if i have to do it for every episode for an webseries or something.

RohanKrishna500:
> We haven't had any reports for Profile 5 issues so far. Test it if you want but the main focus is Profile 8. Fallback shouldn't matter (So 8.(1,2,4) are ok).

i tried another profile 8 dv file and still it fallbacks to hdr layer,i would like to test the files tested by the jellyfin team
here is the mediainfo for the tested file
[mediainfoexample2.txt](https://github.com/user-attachments/files/16325156/mediainfoexample2.txt)

sundermann:
Same on my C2. Remuxing is triggered but still plays as HDR:

[dovi.txt](https://github.com/user-attachments/files/16325162/dovi.txt)

sundermann:
```
[2024-07-21 17:48:18.594 +02:00] [INF] [8] MediaBrowser.MediaEncoding.Transcoding.TranscodeManager: "/usr/lib/jellyfin-ffmpeg/ffmpeg" "-analyzeduration 200M -probesize 1G -fflags +genpts  -i file:\"/media/Movies/Bullet Train (2022) [tmdbid-718930]/Bullet Train (2022).mkv\" -map_metadata -1 -map_chapters -1 -threads 0 -map 0:0 -map 0:1 -map -0:s -codec:v:0 copy -tag:v:0 hvc1 -bsf:v hevc_mp4toannexb -start_at_zero -codec:a:0 copy -copyts -avoid_negative_ts disabled -max_muxing_queue_size 2048 -f hls -max_delay 5000000 -hls_time 6 -hls_segment_type mpegts -start_number 0 -hls_segment_filename \"/config/transcodes/90a013db796e39890b745c16f0de8c75%d.ts\" -hls_playlist_type vod -hls_list_size 0 -y \"/config/transcodes/90a013db796e39890b745c16f0de8c75.m3u8\""
[2024-07-21 17:48:21.135 +02:00] [INF] [25] Emby.Server.Implementations.ScheduledTasks.TaskManager: "Webhook Item Added Notifier" Completed after 0 minute(s) and 0 seconds
[2024-07-21 17:48:24.397 +02:00] [INF] [10] Emby.Server.Implementations.HttpServer.WebSocketManager: WS "192.168.179.31" closed
[2024-07-21 17:48:24.652 +02:00] [INF] [15] Emby.Server.Implementations.HttpServer.WebSocketManager: WS "192.168.179.31" request
[2024-07-21 17:48:25.799 +02:00] [INF] [14] Emby.Server.Implementations.HttpServer.WebSocketManager: WS "192.168.179.31" closed
[2024-07-21 17:48:25.995 +02:00] [INF] [15] Emby.Server.Implementations.HttpServer.WebSocketManager: WS "192.168.179.31" request
[2024-07-21 17:48:27.825 +02:00] [INF] [15] Emby.Server.Implementations.HttpServer.WebSocketManager: WS "192.168.179.31" closed
[2024-07-21 17:48:28.040 +02:00] [INF] [14] Emby.Server.Implementations.HttpServer.WebSocketManager: WS "192.168.179.31" request
[2024-07-21 17:48:28.889 +02:00] [INF] [14] Emby.Server.Implementations.HttpServer.WebSocketManager: WS "192.168.179.31" closed
[2024-07-21 17:48:29.075 +02:00] [INF] [24] Emby.Server.Implementations.HttpServer.WebSocketManager: WS "192.168.179.31" request
[2024-07-21 17:48:29.223 +02:00] [INF] [24] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: Adding playback tracker : TW96aWxsYS81LjAgKFdlYjBTOyBMaW51eC9TbWFydFRWKSBBcHBsZVdlYktpdC81MzcuMzYgKEtIVE1MLCBsaWtlIEdlY2tvKSBDaHJvbWUvODcuMC40MjgwLjg4IFNhZmFyaS81MzcuMzYgV2ViQXBwTWFuYWdlcnwxNzA2OTgxMzU5Mzcz-05a9be82d69e40b884dc0ca550a1ff9f-4d94e5bb1057511e7d5eaf001b602173
[2024-07-21 17:48:29.223 +02:00] [INF] [24] Jellyfin.Plugin.PlaybackReporting.Data.PlaybackTracker: PlaybackTracker : Adding Start Event : 07/21/2024 17:48:29
[2024-07-21 17:48:29.223 +02:00] [INF] [24] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: Creating StartPlaybackTimer Task
[2024-07-21 17:48:29.223 +02:00] [INF] [8] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : Entered
[2024-07-21 17:48:29.687 +02:00] [INF] [24] Emby.Server.Implementations.HttpServer.WebSocketManager: WS "192.168.179.31" closed
[2024-07-21 17:48:29.773 +02:00] [INF] [10] Emby.Server.Implementations.HttpServer.WebSocketManager: WS "192.168.179.31" request
[2024-07-21 17:48:29.819 +02:00] [INF] [26] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: Processing playback tracker : "TW96aWxsYS81LjAgKFdlYjBTOyBMaW51eC9TbWFydFRWKSBBcHBsZVdlYktpdC81MzcuMzYgKEtIVE1MLCBsaWtlIEdlY2tvKSBDaHJvbWUvODcuMC40MjgwLjg4IFNhZmFyaS81MzcuMzYgV2ViQXBwTWFuYWdlcnwxNzA2OTgxMzU5Mzcz-05a9be82d69e40b884dc0ca550a1ff9f-4d94e5bb1057511e7d5eaf001b602173"
[2024-07-21 17:48:38.692 +02:00] [INF] [10] MediaBrowser.Controller.MediaEncoding.TranscodingJob: Stopping ffmpeg process with q command for "/config/transcodes/90a013db796e39890b745c16f0de8c75.m3u8"
[2024-07-21 17:48:38.782 +02:00] [INF] [10] MediaBrowser.MediaEncoding.Transcoding.TranscodeManager: FFmpeg exited with code 0
[2024-07-21 17:48:38.782 +02:00] [INF] [10] MediaBrowser.MediaEncoding.Transcoding.TranscodeManager: Deleting partial stream file(s) "/config/transcodes/90a013db796e39890b745c16f0de8c75.m3u8"
[2024-07-21 17:48:40.902 +02:00] [INF] [8] Emby.Server.Implementations.Session.SessionManager: Playback stopped reported by app "Jellyfin Web" "10.9.8" playing "Bullet Train". Stopped at "8678" ms
[2024-07-21 17:48:40.952 +02:00] [INF] [10] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: Playback stop tracker found, processing stop : "TW96aWxsYS81LjAgKFdlYjBTOyBMaW51eC9TbWFydFRWKSBBcHBsZVdlYktpdC81MzcuMzYgKEtIVE1MLCBsaWtlIEdlY2tvKSBDaHJvbWUvODcuMC40MjgwLjg4IFNhZmFyaS81MzcuMzYgV2ViQXBwTWFuYWdlcnwxNzA2OTgxMzU5Mzcz-05a9be82d69e40b884dc0ca550a1ff9f-4d94e5bb1057511e7d5eaf001b602173"
[2024-07-21 17:48:40.952 +02:00] [INF] [10] Jellyfin.Plugin.PlaybackReporting.Data.PlaybackTracker: PlaybackTracker : Adding Stop Event : 07/21/2024 17:48:40
[2024-07-21 17:48:40.952 +02:00] [INF] [10] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: Playback stop but TrackedPlaybackInfo not found! not storing activity in DB
[2024-07-21 17:48:49.223 +02:00] [INF] [26] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: session.RemoteEndPoint : "192.168.179.68"
[2024-07-21 17:48:49.223 +02:00] [INF] [26] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : event_playing_id     = "4d94e5bb1057511e7d5eaf001b602173"
[2024-07-21 17:48:49.223 +02:00] [INF] [26] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : event_user_id        = "05a9be82d69e40b884dc0ca550a1ff9f"
[2024-07-21 17:48:49.223 +02:00] [INF] [26] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : event_user_id_int    = 1
[2024-07-21 17:48:49.223 +02:00] [INF] [26] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : session_playing_id   = ""
[2024-07-21 17:48:49.223 +02:00] [INF] [26] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : session_user_id      = "05a9be82d69e40b884dc0ca550a1ff9f"
[2024-07-21 17:48:49.223 +02:00] [INF] [26] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : play_method          = "na"
[2024-07-21 17:48:49.223 +02:00] [INF] [26] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : e.ClientName         = "Jellyfin Web"
[2024-07-21 17:48:49.224 +02:00] [INF] [26] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : e.DeviceName         = "LG Smart TV"
[2024-07-21 17:48:49.224 +02:00] [INF] [26] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : ItemName             = "Bullet Train"
[2024-07-21 17:48:49.224 +02:00] [INF] [26] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : ItemId               = "4d94e5bb1057511e7d5eaf001b602173"
[2024-07-21 17:48:49.224 +02:00] [INF] [26] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : ItemType             = "Movie"
[2024-07-21 17:48:49.224 +02:00] [INF] [26] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : Details do not match for play item
[2024-07-21 17:48:49.224 +02:00] [INF] [26] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : Exited
[2024-07-21 17:48:51.135 +02:00] [INF] [25] Emby.Server.Implementations.ScheduledTasks.TaskManager: "Webhook Item Added Notifier" Completed after 0 minute(s) and 0 seconds
```

dmitrylyzo:
`-tag:v:0 hvc1`
Shouldn't this be something like `dvh1`?

sundermann:
@GeorgeH005 I'm running the docker amd64 but I can compile a build myself. I did also try using `jellyfin-ffmpeg` master with the same result

dmitrylyzo:
@sundermann Could you please enable debug logging and post the device profile of TV from the log?

RohanKrishna500:
> With a "regular" video file like the above, the only thing that should result in the stream being tagged as hvc1 is the client misreporting support again.

well im not a developer and still learning to code but I know someone who fixed this issue but on kodi app,maybe it will help here
the link to that post
https://github.com/xbmc/xbmc/issues/24390

RohanKrishna500:
@sundermann u had commented on the above kodi github issue,can u tell me your tv model

RohanKrishna500:
> That seems to reference HDR10+ metadata, not Dolby Vision. They are not encoded nor handled the same as far as I know. Shouldn't be an issue here as LG doesn't support it and ,I think, it should have an HDR10 fallback layer by spec.
yes im aware of that,is there any way to ignore the hdr10 metadata itself for profile 8 dv files so that maybe it will play it(it will act as profile 5 because of single dv layer)
this is not a fix but rater a workaround to figure out the problem occurring as logs arent clear

sundermann:
Ignore the kodi stuff. That is completely different as it's using the native media APIs.

@dmitrylyzo Not sure what to look for in the logs

dmitrylyzo:
> This?

No, it is JSON that will be labeled `[DBG]` and must contain `DeviceProfile`.

To enable debug logging: https://jellyfin.org/docs/general/administration/troubleshooting#debug-logging

I see `hevc-rangetype=SDR,HDR10,HLG` but this could be caused by the wrong profile.

Also, you have posted `api_key`. To invalidate it, just log out on the TV or remove its device in the Dashboard.

sundermann:
Don't worry about that `api_key`. My instance is not exposed to the internet. I finally found out how to properly enable debug logging. Here's the device profile json:

```json
{
   "Name":null,
   "Id":null,
   "MaxStreamingBitrate":120000000,
   "MaxStaticBitrate":100000000,
   "MusicStreamingTranscodingBitrate":384000,
   "MaxStaticMusicBitrate":8000000,
   "DirectPlayProfiles":[
      "DirectPlayProfile"{
         "Container":"webm",
         "AudioCodec":"vorbis,opus",
         "VideoCodec":"vp8,vp9,av1",
         "Type":"Video"
      },
      "DirectPlayProfile"{
         "Container":"mp4,m4v",
         "AudioCodec":"aac,mp3,ac3,eac3,mp2,pcm_s16le,pcm_s24le,opus,flac,vorbis",
         "VideoCodec":"h264,hevc,mpeg2video,vc1,vp9,av1",
         "Type":"Video"
      },
      "DirectPlayProfile"{
         "Container":"mkv",
         "AudioCodec":"aac,mp3,ac3,eac3,mp2,pcm_s16le,pcm_s24le,opus,flac,vorbis",
         "VideoCodec":"h264,hevc,mpeg2video,vc1,vp9,av1",
         "Type":"Video"
      },
      "DirectPlayProfile"{
         "Container":"m2ts",
         "AudioCodec":"aac,mp3,ac3,eac3,mp2,pcm_s16le,pcm_s24le,opus,flac,vorbis",
         "VideoCodec":"h264,vc1,mpeg2video",
         "Type":"Video"
      },
      "DirectPlayProfile"{
         "Container":"wmv",
         "AudioCodec":"",
         "VideoCodec":"",
         "Type":"Video"
      },
      "DirectPlayProfile"{
         "Container":"ts,mpegts",
         "AudioCodec":"aac,mp3,ac3,eac3,mp2,pcm_s16le,pcm_s24le,opus,flac,vorbis",
         "VideoCodec":"h264,hevc,vc1,mpeg2video",
         "Type":"Video"
      },
      "DirectPlayProfile"{
         "Container":"asf",
         "AudioCodec":"",
         "VideoCodec":"",
         "Type":"Video"
      },
      "DirectPlayProfile"{
         "Container":"avi",
         "AudioCodec":"aac,mp3,ac3,eac3,mp2,pcm_s16le,pcm_s24le,opus,flac,vorbis",
         "VideoCodec":"",
         "Type":"Video"
      },
      "DirectPlayProfile"{
         "Container":"mpg",
         "AudioCodec":"aac,mp3,ac3,eac3,mp2,pcm_s16le,pcm_s24le,opus,flac,vorbis",
         "VideoCodec":"",
         "Type":"Video"
      },
      "DirectPlayProfile"{
         "Container":"mpeg",
         "AudioCodec":"aac,mp3,ac3,eac3,mp2,pcm_s16le,pcm_s24le,opus,flac,vorbis",
         "VideoCodec":"",
         "Type":"Video"
      },
      "DirectPlayProfile"{
         "Container":"mov",
         "AudioCodec":"aac,mp3,ac3,eac3,mp2,pcm_s16le,pcm_s24le,opus,flac,vorbis",
         "VideoCodec":"h264",
         "Type":"Video"
      },
      "DirectPlayProfile"{
         "Container":"opus",
         "AudioCodec":null,
         "VideoCodec":null,
         "Type":"Audio"
      },
      "DirectPlayProfile"{
         "Container":"webm",
         "AudioCodec":"opus",
         "VideoCodec":null,
         "Type":"Audio"
      },
      "DirectPlayProfile"{
         "Container":"mp3",
         "AudioCodec":null,
         "VideoCodec":null,
         "Type":"Audio"
      },
      "DirectPlayProfile"{
         "Container":"aac",
         "AudioCodec":null,
         "VideoCodec":null,
         "Type":"Audio"
      },
      "DirectPlayProfile"{
         "Container":"m4a",
         "AudioCodec":"aac",
         "VideoCodec":null,
         "Type":"Audio"
      },
      "DirectPlayProfile"{
         "Container":"m4b",
         "AudioCodec":"aac",
         "VideoCodec":null,
         "Type":"Audio"
      },
      "DirectPlayProfile"{
         "Container":"flac",
         "AudioCodec":null,
         "VideoCodec":null,
         "Type":"Audio"
      },
      "DirectPlayProfile"{
         "Container":"webma",
         "AudioCodec":null,
         "VideoCodec":null,
         "Type":"Audio"
      },
      "DirectPlayProfile"{
         "Container":"webm",
         "AudioCodec":"webma",
         "VideoCodec":null,
         "Type":"Audio"
      },
      "DirectPlayProfile"{
         "Container":"wav",
         "AudioCodec":null,
         "VideoCodec":null,
         "Type":"Audio"
      },
      "DirectPlayProfile"{
         "Container":"ogg",
         "AudioCodec":null,
         "VideoCodec":null,
         "Type":"Audio"
      },
      "DirectPlayProfile"{
         "Container":"hls",
         "AudioCodec":"aac,mp3,ac3,eac3,mp2",
         "VideoCodec":"hevc,h264",
         "Type":"Video"
      }
   ],
   "TranscodingProfiles":[
      "TranscodingProfile"{
         "Container":"ts",
         "Type":"Audio",
         "VideoCodec":"",
         "AudioCodec":"aac",
         "Protocol":"hls",
         "EstimateContentLength":false,
         "EnableMpegtsM2TsMode":false,
         "TranscodeSeekInfo":"Auto",
         "CopyTimestamps":false,
         "Context":"Streaming",
         "EnableSubtitlesInManifest":false,
         "MaxAudioChannels":"6",
         "MinSegments":1,
         "SegmentLength":0,
         "BreakOnNonKeyFrames":false,
         "Conditions":[
            
         ]
      },
      "TranscodingProfile"{
         "Container":"aac",
         "Type":"Audio",
         "VideoCodec":"",
         "AudioCodec":"aac",
         "Protocol":"http",
         "EstimateContentLength":false,
         "EnableMpegtsM2TsMode":false,
         "TranscodeSeekInfo":"Auto",
         "CopyTimestamps":false,
         "Context":"Streaming",
         "EnableSubtitlesInManifest":false,
         "MaxAudioChannels":"6",
         "MinSegments":0,
         "SegmentLength":0,
         "BreakOnNonKeyFrames":false,
         "Conditions":[
            
         ]
      },
      "TranscodingProfile"{
         "Container":"mp3",
         "Type":"Audio",
         "VideoCodec":"",
         "AudioCodec":"mp3",
         "Protocol":"http",
         "EstimateContentLength":false,
         "EnableMpegtsM2TsMode":false,
         "TranscodeSeekInfo":"Auto",
         "CopyTimestamps":false,
         "Context":"Streaming",
         "EnableSubtitlesInManifest":false,
         "MaxAudioChannels":"6",
         "MinSegments":0,
         "SegmentLength":0,
         "BreakOnNonKeyFrames":false,
         "Conditions":[
            
         ]
      },
      "TranscodingProfile"{
         "Container":"opus",
         "Type":"Audio",
         "VideoCodec":"",
         "AudioCodec":"opus",
         "Protocol":"http",
         "EstimateContentLength":false,
         "EnableMpegtsM2TsMode":false,
         "TranscodeSeekInfo":"Auto",
         "CopyTimestamps":false,
         "Context":"Streaming",
         "EnableSubtitlesInManifest":false,
         "MaxAudioChannels":"6",
         "MinSegments":0,
         "SegmentLength":0,
         "BreakOnNonKeyFrames":false,
         "Conditions":[
            
         ]
      },
      "TranscodingProfile"{
         "Container":"wav",
         "Type":"Audio",
         "VideoCodec":"",
         "AudioCodec":"wav",
         "Protocol":"http",
         "EstimateContentLength":false,
         "EnableMpegtsM2TsMode":false,
         "TranscodeSeekInfo":"Auto",
         "CopyTimestamps":false,
         "Context":"Streaming",
         "EnableSubtitlesInManifest":false,
         "MaxAudioChannels":"6",
         "MinSegments":0,
         "SegmentLength":0,
         "BreakOnNonKeyFrames":false,
         "Conditions":[
            
         ]
      },
      "TranscodingProfile"{
         "Container":"opus",
         "Type":"Audio",
         "VideoCodec":"",
         "AudioCodec":"opus",
         "Protocol":"http",
         "EstimateContentLength":false,
         "EnableMpegtsM2TsMode":false,
         "TranscodeSeekInfo":"Auto",
         "CopyTimestamps":false,
         "Context":"Static",
         "EnableSubtitlesInManifest":false,
         "MaxAudioChannels":"6",
         "MinSegments":0,
         "SegmentLength":0,
         "BreakOnNonKeyFrames":false,
         "Conditions":[
            
         ]
      },
      "TranscodingProfile"{
         "Container":"mp3",
         "Type":"Audio",
         "VideoCodec":"",
         "AudioCodec":"mp3",
         "Protocol":"http",
         "EstimateContentLength":false,
         "EnableMpegtsM2TsMode":false,
         "TranscodeSeekInfo":"Auto",
         "CopyTimestamps":false,
         "Context":"Static",
         "EnableSubtitlesInManifest":false,
         "MaxAudioChannels":"6",
         "MinSegments":0,
         "SegmentLength":0,
         "BreakOnNonKeyFrames":false,
         "Conditions":[
            
         ]
      },
      "TranscodingProfile"{
         "Container":"aac",
         "Type":"Audio",
         "VideoCodec":"",
         "AudioCodec":"aac",
         "Protocol":"http",
         "EstimateContentLength":false,
         "EnableMpegtsM2TsMode":false,
         "TranscodeSeekInfo":"Auto",
         "CopyTimestamps":false,
         "Context":"Static",
         "EnableSubtitlesInManifest":false,
         "MaxAudioChannels":"6",
         "MinSegments":0,
         "SegmentLength":0,
         "BreakOnNonKeyFrames":false,
         "Conditions":[
            
         ]
      },
      "TranscodingProfile"{
         "Container":"wav",
         "Type":"Audio",
         "VideoCodec":"",
         "AudioCodec":"wav",
         "Protocol":"http",
         "EstimateContentLength":false,
         "EnableMpegtsM2TsMode":false,
         "TranscodeSeekInfo":"Auto",
         "CopyTimestamps":false,
         "Context":"Static",
         "EnableSubtitlesInManifest":false,
         "MaxAudioChannels":"6",
         "MinSegments":0,
         "SegmentLength":0,
         "BreakOnNonKeyFrames":false,
         "Conditions":[
            
         ]
      },
      "TranscodingProfile"{
         "Container":"ts",
         "Type":"Video",
         "VideoCodec":"hevc,h264",
         "AudioCodec":"aac,mp3,ac3,eac3,mp2",
         "Protocol":"hls",
         "EstimateContentLength":false,
         "EnableMpegtsM2TsMode":false,
         "TranscodeSeekInfo":"Auto",
         "CopyTimestamps":false,
         "Context":"Streaming",
         "EnableSubtitlesInManifest":false,
         "MaxAudioChannels":"6",
         "MinSegments":1,
         "SegmentLength":0,
         "BreakOnNonKeyFrames":false,
         "Conditions":[
            
         ]
      }
   ],
   "ContainerProfiles":[
      
   ],
   "CodecProfiles":[
      "CodecProfile"{
         "Type":"VideoAudio",
         "Conditions":[
            "ProfileCondition"{
               "Condition":"LessThanEqual",
               "Property":"AudioChannels",
               "Value":"2",
               "IsRequired":false
            }
         ],
         "ApplyConditions":[
            
         ],
         "Codec":"flac",
         "Container":null
      },
      "CodecProfile"{
         "Type":"Video",
         "Conditions":[
            "ProfileCondition"{
               "Condition":"NotEquals",
               "Property":"IsAnamorphic",
               "Value":"true",
               "IsRequired":false
            },
            "ProfileCondition"{
               "Condition":"EqualsAny",
               "Property":"VideoProfile",
               "Value":"high|main|baseline|constrained baseline",
               "IsRequired":false
            },
            "ProfileCondition"{
               "Condition":"EqualsAny",
               "Property":"VideoRangeType",
               "Value":"SDR",
               "IsRequired":false
            },
            "ProfileCondition"{
               "Condition":"LessThanEqual",
               "Property":"VideoLevel",
               "Value":"52",
               "IsRequired":false
            }
         ],
         "ApplyConditions":[
            
         ],
         "Codec":"h264",
         "Container":null
      },
      "CodecProfile"{
         "Type":"Video",
         "Conditions":[
            "ProfileCondition"{
               "Condition":"NotEquals",
               "Property":"IsAnamorphic",
               "Value":"true",
               "IsRequired":false
            },
            "ProfileCondition"{
               "Condition":"EqualsAny",
               "Property":"VideoProfile",
               "Value":"main|main 10",
               "IsRequired":false
            },
            "ProfileCondition"{
               "Condition":"EqualsAny",
               "Property":"VideoRangeType",
               "Value":"SDR|HDR10|HLG",
               "IsRequired":false
            },
            "ProfileCondition"{
               "Condition":"LessThanEqual",
               "Property":"VideoLevel",
               "Value":"183",
               "IsRequired":false
            }
         ],
         "ApplyConditions":[
            
         ],
         "Codec":"hevc",
         "Container":null
      },
      "CodecProfile"{
         "Type":"Video",
         "Conditions":[
            "ProfileCondition"{
               "Condition":"EqualsAny",
               "Property":"VideoRangeType",
               "Value":"SDR|HDR10|HLG",
               "IsRequired":false
            }
         ],
         "ApplyConditions":[
            
         ],
         "Codec":"vp9",
         "Container":null
      },
      "CodecProfile"{
         "Type":"Video",
         "Conditions":[
            "ProfileCondition"{
               "Condition":"NotEquals",
               "Property":"IsAnamorphic",
               "Value":"true",
               "IsRequired":false
            },
            "ProfileCondition"{
               "Condition":"EqualsAny",
               "Property":"VideoProfile",
               "Value":"main",
               "IsRequired":false
            },
            "ProfileCondition"{
               "Condition":"EqualsAny",
               "Property":"VideoRangeType",
               "Value":"SDR|HDR10|HLG",
               "IsRequired":false
            },
            "ProfileCondition"{
               "Condition":"LessThanEqual",
               "Property":"VideoLevel",
               "Value":"19",
               "IsRequired":false
            }
         ],
         "ApplyConditions":[
            
         ],
         "Codec":"av1",
         "Container":null
      }
   ],
   "SubtitleProfiles":[
      "SubtitleProfile"{
         "Format":"vtt",
         "Method":"External",
         "DidlMode":null,
         "Language":null,
         "Container":null
      },
      "SubtitleProfile"{
         "Format":"ass",
         "Method":"External",
         "DidlMode":null,
         "Language":null,
         "Container":null
      },
      "SubtitleProfile"{
         "Format":"ssa",
         "Method":"External",
         "DidlMode":null,
         "Language":null,
         "Container":null
      }
   ]
}
```

lgndvd:
Hi, everything seems good on LG CX with 10.9.8.

I tested the same files used when debugging 10.9.x behaviour [here](https://github.com/jellyfin/jellyfin/issues/11705#issuecomment-2118988363), Dolby Vision now triggers correctly even for MKV files (remuxing, as expected)...and for that, I sincerely love you guys ❤️

Nitpicking: I noticed that TS DoVi files are not direct playing (both DV5 and DV8), they trigger Dolby Vision but show as "remuxing". MP4 DoVi files are direct playing, as expected.

sundermann:
@GeorgeH005 Yes, this is from the app

sundermann:
Inspecting the app it tells me that `web0sVersion: 23` and `web0s: true`

sundermann:
It seems undefined but it's a bit hard to debug using only the minified sources.

lgndvd:
> Thanks!
> 
> 
> 
> As for the nitpicking, is "Prefer fmp4-hls media container" unticked in the playback settings?

Yep, disabled as you suggested during early testing.

gnattu:
Hello, could one of the LG users test this change for profile 8 support checking? After reading a few specification docs I found out the our profile testing is probably using the wrong codec string. Basically, the `dvh1` should present as a supplemental codec rather than the main codec for profile 8 dolby vision videos. Our current string is `video/mp4; codecs="dvh1.08.09`, but it should be  `video/mp4; codecs="hvc1.2.4.L184 dvh1.08.09"`. On Apple devices, the former will return false and on latter will return probably, is that also the case on LG TVs?

gnattu:
I think the `options.supportsDolbyVision` property is set by the LG client and for some reason that option is not set. We can add the same assumption to the `supportsDolbyVision` to force enable this on webOS though, if we are confident that all LG TVs really supports Dolby Vision

gnattu:
> That should affect all users though shouldn't it?

I have no idea why that option is not set for a specific user because it really should. But we are already doing the same thing for all Apple devices even though some of them does not really support dolby vision, but Apple has its own fallback mechanisms in Safari and the playback won't crash if the device does not support that, not sure what will happen on webOS though.

https://github.com/jellyfin/jellyfin-web/blob/0090ccd893b131ff100d98fe32d7e3c3cfea12d9/src/scripts/browserDeviceProfile.js#L235-L239

gnattu:
> I can attest to the fact that if Dolby Vision isn't triggered, Profile 5 content will look purple and green.

Yes, but we will be checking if the profile is supported later, and we *should* be safe unless webOS lies about this fact and reports that it plays profile 5 when it does not. 

> Unless we edit our HLS manifests and make them follow the same principle as above, listing dvh1.08 as an optional codec and not tagging the stream at all (if the metadata passthrough is not dependent on that). 

Apple requires all HDR streams in HLS to have an SDR fallback and dolby vision contents should label its profile in the SUPPLEMENTAL-CODECS attribute. We already did the former but not doing the latter yet. I plan to add that in 10.10 some time so that the client will have a better fallback if it really implements the fallback.

sundermann:
Reinstalling the app fixed it apparently. cc @RohanKrishna500

Not sure, why this was happening. It just seems weird.

gnattu:
That's good to hear, but I still want to check if `video/mp4; codecs="hvc1.2.4.L184 dvh1.08.09"` works on LG webOS because that looks better (and should be more accurate) than hardcoding a webOS version. I will push a change with the string anyway because the current one does not really work with Apple devices, and if it also works for LG it would be even better.

sundermann:
It does not seem to trigger reliably though. It seems sometimes it falls back to transcoding to SDR (saying the dynamic range is not supported), then trying to play the same file again triggers dolby vision correctly.

gnattu:
Just keep the bypass checking for now instead of playing with the codec string.

RohanKrishna500:
> It doesn't work that way... You have a "base stream" that is a normal HDR10 (or HLG or SDR, depends on the "fallback") stream that is then augmented by the extra metadata. There aren't two separate video streams. Even the profile 7 Enhancement Layers that are actually video streams, are just augmentations (overlays I think) , not copies of the same video stream with augmentations applied. That's what I think is true.

ah...i get it i thought it as seperate video streams.

RohanKrishna500:
> -2

is this the webos app u meant?....also i have this issue where even profile 5 seems to not play in proper colors in the previous jellyfin build whereas other lg users reported it to be fine when remuxed.
also @sundermann do u have the same model as mine(LG A2)?

megosugit:
I have a strange issue even with 10.9.8 or 10.9.9:
if I start a DoVi movie from the beginning, it seems to be working fine (watched 30min of a movie yesterday). but if I resume the same movie, it works for a few seconds, or sometimes minutes, and then starts to freeze. Sometimes, it continues by itself after the first freeze, then after the second freeze, it is completely stuck.
If I exit the movie and resume again, it starts with the same issue (works for a few seconds or minutes, then freeze again).

Am I alone with this issue ?

dmitrylyzo:
@megosugit Try disabling fMP4 on the TV. In 10.9.8 (iirc), it was enabled for webOS by default, but it doesn't work well. In 10.9.9 (this PR), it has been disabled, but you need to disable it manually if you saved settings in 10.9.8.

megosugit:
> @megosugit Try disabling fMP4 on the TV. In 10.9.8 (iirc), it was enabled for webOS by default, but it doesn't work well. In 10.9.9 (this PR), it has been disabled, but you need to disable it manually if you saved settings in 10.9.8.

It is disabled, when enabled it stays black

mfizz1:
> > @megosugit Try disabling fMP4 on the TV. In 10.9.8 (iirc), it was enabled for webOS by default, but it doesn't work well. In 10.9.9 (this PR), it has been disabled, but you need to disable it manually if you saved settings in 10.9.8.
> 
> It is disabled, when enabled it stays black

I raised a similar issue back in the day. I raised it in #5068. I don't believe there is a solution yet.

RohanKrishna500:
> Am I alone with this issue ?

hey  are u able to play dolby vision correctly though?

megosugit:
> [RohanKrishna500](/RohanKrishna500)

It triggers the Dolby Vision mode on my TV, but then it freezes randomly, so : no I'm not able play DV files correctly. It is a nightmare to watch a movie like this. I have to exit and resume several times the movie, it is bad

megosugit:
> From my experience: if you start to play the DV movie from scratch it works normally, but then if you try to skip forward then the screen goes black and it is not possible to move further or resume from where you were... you need to start it again from zero.. That's my experience. And when i looked at the server logs from my Jellyfin server it was trying to remux it but somehow it is not working.

that's what I used to have too, but after a Jellyfin update, I was able to resume so I thought it was fixed. But when I really used it to watch a movie, and not only test a few seconds, I discovered it can resume and play for a few minutes but then it freezes.

So in summary, I thought it was fixed but it is still broken. It is better than before though, but still useless to watch a movie without becoming mad.

RohanKrishna500:
> > [RohanKrishna500](/RohanKrishna500)
> 
> It triggers the Dolby Vision mode on my TV, but then it freezes randomly, so : no I'm not able play DV files correctly. It is a nightmare to watch a movie like this. I have to exit and resume several times the movie, it is bad

your tv model?

megosugit:
> > > [RohanKrishna500](/RohanKrishna500)
> > 
> > 
> > It triggers the Dolby Vision mode on my TV, but then it freezes randomly, so : no I'm not able play DV files correctly. It is a nightmare to watch a movie like this. I have to exit and resume several times the movie, it is bad
> 
> your tv model?

LG 77 G3

RohanKrishna500:
> > > > [RohanKrishna500](/RohanKrishna500)
> > > 
> > > 
> > > It triggers the Dolby Vision mode on my TV, but then it freezes randomly, so : no I'm not able play DV files correctly. It is a nightmare to watch a movie like this. I have to exit and resume several times the movie, it is bad
> > 
> > 
> > your tv model?
> 
> LG 77 G3

ig kodi will work fine for you bro try it

megosugit:
> > > > > [RohanKrishna500](/RohanKrishna500)
> > > > 
> > > > 
> > > > It triggers the Dolby Vision mode on my TV, but then it freezes randomly, so : no I'm not able play DV files correctly. It is a nightmare to watch a movie like this. I have to exit and resume several times the movie, it is bad
> > > 
> > > 
> > > your tv model?
> > 
> > 
> > LG 77 G3
> 
> ig kodi will work fine for you bro try it

Do I need to go in developer mode to install it? Looks a bit dodgy. 

I had kodi running on a PC previously, I thought jellyfin would be a smarter choice.

RohanKrishna500:
> > > > > > [RohanKrishna500](/RohanKrishna500)
> > > > > 
> > > > > 
> > > > > It triggers the Dolby Vision mode on my TV, but then it freezes randomly, so : no I'm not able play DV files correctly. It is a nightmare to watch a movie like this. I have to exit and resume several times the movie, it is bad
> > > > 
> > > > 
> > > > your tv model?
> > > 
> > > 
> > > LG 77 G3
> > 
> > 
> > ig kodi will work fine for you bro try it
> 
> Do I need to go in developer mode to install it? Looks a bit dodgy.
> 
> I had kodi running on a PC previously, I thought jellyfin would be a smarter choice.

It didn't work for me but might work for,and yes its completely safe and easy to install,i would prefer jellyfin over kodi because my media are stored in cloud and I mount it to my jellyfin server,for kodi only offline files

mtwo:
You need to go into developer mode, but it's pretty straightforward and it works well once it's installed.

FYI I _think_ that all of these solutions (Kodi nightlies with Dolby Vision compatibility mode, Jellyfin with the fixes discussed here) only work for a subset of all Dolby Vision content. Profile 7 FEL content (some UHD rips) requires special hardware to view the full enhancement layer. I ended up purchasing a Ugoos and installing CoreElec (Kodi) onto it, and then connecting it to Jellyfin.

RohanKrishna500:
Hi, checking if there has been any fix to this

gnattu:
I *think* we have to make some compatibility trade of options if some models do not even play dolby vision in hls properly. We already tried both fmp4 and ts and some TV model does not like that either. For such TV models we might need an extra option so that let the user to agree to only play the fallback layer in the original mkv and give up on dolby vision, or the user needs to remux to mp4 himself.

RohanKrishna500:
> I _think_ we have to make some compatibility trade of options if some models do not even play dolby vision in hls properly. We already tried both fmp4 and ts and some TV model does not like that either. For such TV models we might need an extra option so that let the user to agree to only play the fallback layer in the original mkv and give up on dolby vision, or the user needs to remux to mp4 himself.

yeah but i dont get it why is it like for some models it doesn't work but I have seen some users who have C series can easily play dolby vision files like by a simple remuxing.I guess it might be due to the processor used in some tv models that restricts it or something.

RohanKrishna500:
hi can someone tell me why dv p5 in mp4 are not direct played on jellyfin

PR #5566, opened 2024-05-20
Only connect to manual addresses specified by user

gnattu:
This should never be enabled on the web because users are not expected to connect to an address they don't specify. For specific use cases like auto endpoint switching between networks, this should be managed by DNS or the router itself, not by the application. 

Having an address that is not always connectable causes our Android users to be unable to reliably connect to the server. It also breaks many reverse-proxy setups, as this address exposed by the server usually bypasses the proxy unless explicitly configured by the user. This has far more negative impact than the benifit it brings.

**Changes**

- apiClient.manualAddressOnly = true

**Issues**

Reverts #3655

Fixes jellyfin/jellyfin-android#1382
Fixes jellyfin/jellyfin#11658

Maybe related: jellyfin/jellyfin-android#1385 jellyfin/jellyfin#11620

### DIFF
--- a/src/components/ServerConnections.js
+++ b/src/components/ServerConnections.js
@@ -65,7 +65,7 @@ class ServerConnections extends ConnectionManager {
         );
 
         apiClient.enableAutomaticNetworking = false;
-        apiClient.manualAddressOnly = false;
+        apiClient.manualAddressOnly = true;
 
         this.addApiClient(apiClient);
 


nielsvanvelzen (approved):

Bond-009 (approved):

ivan-danilov:
@thornbill, hopefully, 10.9.3 with this change will be released soon, as it is a deal-breaker on Android at the moment.

PR #5571, opened 2024-05-20
Fix search for tab library in experimental layout

grafixeyehero:
**Changes**
- Extract Search Button to component
- Fix Search for tab library based on url params `parentId` and `collectionType`
- Disable the `SearchButton` in the AppToolbar if the current location is the Search page.


src/apps/experimental/components/AppToolbar/index.tsx
```diff
@@ -38,25 +33,15 @@ const ExperimentalAppToolbar: FC<AppToolbarProps> = ({
 
     const isTabsAvailable = isTabPath(location.pathname);
     const isPublicPath = PUBLIC_PATHS.includes(location.pathname);
+    const isSearchPath = location.pathname === '/search.html';
```
thornbill, line 36: Maybe we should just move this into the component?

PR #5585, opened 2024-05-22
Add lyrics auto scroll

scampower3:
**Changes**

Adds autoscrolling to lyrics. Will stop autoscrolling if wheel or up key or down key is used. Can continue autoscrolling by selecting the lyrics line.
**Issues**

Fixes https://github.com/jellyfin/jellyfin-web/issues/5487


Shadowghost:
Maybe worth a PR against 10.9? IMHO it's a bit weird behaviour to not scroll on timed lyrics

scampower3:
Yeah, I agree. Changed the base to 10.9.z

src/controllers/lyrics.js
```diff
@@ -236,6 +241,12 @@ export default function (view) {
 
     view.addEventListener('viewshow', function () {
         Events.on(playbackManager, 'playerchange', onPlayerChange);
+        document.addEventListener('wheel', () => (autoScroll = false));
+        document.addEventListener('keydown', (e) => {
+            if (e.keyCode == 38 || e.keyCode == 40) {
```
dmitrylyzo, line 246: You can use the functions from `keyboardNavigation` to check the navigation keys:
https://github.com/jellyfin/jellyfin-web/blob/b71aaa05a655242f3cffdc3b46685c9a224f39b7/src/controllers/playback/video/index.js#L1202
https://github.com/jellyfin/jellyfin-web/blob/b71aaa05a655242f3cffdc3b46685c9a224f39b7/src/controllers/playback/video/index.js#L1236

src/controllers/lyrics.js
```diff
@@ -245,6 +256,12 @@ export default function (view) {
 
     view.addEventListener('viewbeforehide', function () {
         Events.off(playbackManager, 'playerchange', onPlayerChange);
+        document.removeEventListener('wheel', () => (autoScroll = false));
+        document.removeEventListener('keydown', (e) => {
```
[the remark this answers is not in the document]
scampower3: Seems to work fine unsubscribing without extracting the handlers. But I'll extract anyways.

src/controllers/lyrics.js
```diff
@@ -17,6 +19,7 @@ let currentItem;
 
 let savedLyrics;
 let isDynamicLyric = false;
+let autoScroll = true;
```
dmitrylyzo, line 22: I think we should set `autoScroll` to `true` in the `viewshow` handler so that we autoscroll when switching between views.

dmitrylyzo (commented):
It seems to work.

There's a conflict between focus scroll and lyrics scroll: when focused element leaves the view when some distant lyrics is active. Tbf, I don't think we should fix it right now (_because I don't know yet how_ :sweat_smile: ).
:point_up_2: in TV mode

src/controllers/lyrics.js
scampower3: Whoops. Forgot to consider that scenario.

src/controllers/lyrics.js
```diff
@@ -69,6 +72,9 @@ export default function (view) {
         if (lyric) {
             lyric.classList.remove('pastLyric');
             lyric.classList.remove('futureLyric');
+            if (autoScroll) {
+                scrollManager.scrollToElement(lyric, true);
```
dmitrylyzo, line 76: It now works when switching between views, but we need to use instant scrolling on `viewshow`.
It currently scrolls smoothly from top do bottom.

We could use a boolean (say `smoothScroll`) (set `false` in `viewshow`, set `true` here).
Or "enum" (state) for `autoScroll`: `noscroll = 0`, `instant = 1`, `smooth = 2`.

scampower3: I am partial to using Typscript enums for this. Where should I place the typescript file? In types or in the same folder as the lyrics.js

src/controllers/lyrics.js
```diff
@@ -5,18 +5,22 @@ import { appRouter } from '../components/router/appRouter';
...
 
 import '../styles/lyrics.scss';
+import { AutoScrollType } from './lyrics.types';
 
 let currentPlayer;
 let currentItem;
 
 let savedLyrics;
 let isDynamicLyric = false;
+let autoScroll = AutoScrollType.Smooth;
```
dmitrylyzo, line 23:
```suggestion
let autoScroll = AutoScrollType.Instant;
```

src/controllers/lyrics.js
```diff
@@ -69,6 +73,14 @@ export default function (view) {
...
             lyric.classList.remove('pastLyric');
             lyric.classList.remove('futureLyric');
+            if (autoScroll === AutoScrollType.Smooth) {
+                scrollManager.scrollToElement(lyric, true);
+            }
+            if (autoScroll === AutoScrollType.Instant) {
+                // instant scroll is used when the view is first loaded
+                scrollManager.scrollToElement(lyric, false);
+                autoScroll = AutoScrollType.Smooth;
+            }
```
dmitrylyzo, line 83:
```suggestion
            if (autoScroll !== AutoScrollType.NoScroll) {
                // instant scroll is used when the view is first loaded
                scrollManager.scrollToElement(lyric, autoScroll === AutoScrollType.Smooth);
                autoScroll = AutoScrollType.Smooth;
            }
```

dmitrylyzo (commented):
Desktop and Mobile display modes work fine.

In TV display mode, this `autoFocuser` call breaks the instant scroll to the current lyric.
https://github.com/jellyfin/jellyfin-web/blob/15e35fa45ffc928a2832277923a538d8b235fc9e/src/controllers/lyrics.js#L142
Called after `scrollToElement`, selecting the first lyric.

We could try replacing it with focusing the current lyrics (using `focusManager`).

src/controllers/lyrics.types.ts
```diff
@@ -0,0 +1,5 @@
+export enum AutoScrollType {
```
thornbill, line 1:
```suggestion
export enum AutoScroll {
```

thornbill (commented):
The code changes look fine to me but this could probably use another round of testing. 👍

Issue #5560, opened 2024-05-19
Since v10.9, a lot of my cover art for media is displayed in a square which crops out a lot of the image.


maru801:
**Describe The Bug**
Before v10.9, all the cover art for my shows were properly displayed in a rectangle, with only some image resolutions causing the cover art display to actually become a square (500px x 500px cover art for example).

Now, a lot of my cover art is for some reason becoming a square. Not all of my covers have this happen, but it appears to be random. I don't know what image resolutions will cause this.

Here's pictures describing what's happening:
[This is a show cover that's a square](https://i.imgur.com/OTxFAG8.jpeg)
[This is the actual image that's being displayed](https://i.imgur.com/hGnlGpX.jpeg)

[Example of how they're supposed to be](https://i.imgur.com/AGzvjWT.jpeg)

**Steps To Reproduce**
1. Look around your libraries and look for cover art that became squares when displayed.

**Expected Behavior**
Media cover art should be rectangles instead of being shrunk to a square.

**System (please complete the following information):**
 - Platform: Linux
 - Browser: Chrome, Firefox
 - Jellyfin Version: 10.9.2

Debouelesmorts:
I'm experiencing the same problem since upgrading from version 10.8.13 to 19.2.

kandykarter:
I have this issue as well, so far only noticed on one film, corrected by switching to a similar poster with a more "standard" size, 1500x1000.


![Screenshot from 2024-05-20 10-37-25](https://github.com/jellyfin/jellyfin-web/assets/107814899/99e92b62-80f8-4caf-b755-21f0a85911cc)

Debouelesmorts:
I have several thousand films and unlike @kandykarter this bug affects my covers by the hundreds.

scampower3:
What are the aspect ratios of the problematic images?

1hitsong:
Fixed by #5587

PR #5587, opened 2024-05-23
Fix square posters

scampower3:
According to TMDB, posters can also come in 1:1.33 (≈0.75) aspect ratio. So set aspect ratio less than 0.75 + 0.5 (for buffer) to be Portrait.
**Changes**

Change aspect ratio threshold to 0.8
**Issues**

Fixes https://github.com/jellyfin/jellyfin-web/issues/5560


thornbill:
There are a few places where this comparison happens: https://github.com/search?q=repo%3Ajellyfin%2Fjellyfin-web+71+NOT+language%3ASVG+NOT+language%3ASCSS&type=code

thornbill (approved):

PR #5563, opened 2024-05-20
Fix chapter name XSS injection in progress bar

nielsvanvelzen:
The chapter markers in the video player seekbar, introduced in 10.9, have some issues. First they add unsanitized user values to the seekbar (chapter name). And secondly, those names can be anything which could cause serious issues when they clash with existing CSS classes. I have no idea why we add these names as a class as this does not provide any benefit. For now, just add the specific className (which is currently always `chapterMarker`).

**Changes**
- Fix chapter name XSS injection in progress bar

**Issues**
Fixes #5561

PR #5593, opened 2024-05-23
Fix chapter marker re-rendering

thornbill:
**Changes**
* Prevents the marker html from being constantly re-rendered if `markerInfo` is already populated
* Fixes a small style regression from #5563 due to the size being dependent on the font-size set by the `material-icons` class

**Issues**
I can't find one, but I saw the chapter markers flashing being reported somewhere...

### DIFF
--- a/src/elements/emby-slider/emby-slider.js
+++ b/src/elements/emby-slider/emby-slider.js
@@ -203,14 +203,14 @@ function setMarker(range, valueMarker, marker, valueProgress) {
 }
 
 function updateMarkers(range, currentValue) {
-    if (range.getMarkerInfo) {
+    if (range.getMarkerInfo && !range.markerInfo) {
         range.markerInfo = range.getMarkerInfo();
 
-        range.markerContainerElement.innerHTML = '';
-
+        let markersHtml = '';
         range.markerInfo.forEach(() => {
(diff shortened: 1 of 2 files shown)


nielsvanvelzen (approved):

src/elements/emby-slider/emby-slider.js
```diff
@@ -203,14 +203,14 @@ function setMarker(range, valueMarker, marker, valueProgress) {
 }
 
 function updateMarkers(range, currentValue) {
-    if (range.getMarkerInfo) {
+    if (range.getMarkerInfo && !range.markerInfo) {
```
dmitrylyzo, line 206: Isn't `markerInfo` (as well as the slider itself) preserved when start playing the next video, so the markers won't be updated?

thornbill: In my testing it was not.

dmitrylyzo: I only get to the breakpoint once per refresh. So I assume they are preserved (because videoosd doesn't reload for the next item).

thornbill: hmm yeah you are right... I'm not sure how it worked when I originally tested... https://github.com/jellyfin/jellyfin-web/pull/5601

PR #5595, opened 2024-05-24
Fix Up upNextDialog blocked by transparent video controls

anthonym01:
On high display scaling, the controls sit over the up next dialog, i know its a sloppy change but i hope you wont mind :)

**Changes**
src\components\upnextdialog\upnextdialog.scss

**Issues**
Fix Up upNextDialog blocked by transparent video controls

### DIFF
--- a/src/components/upnextdialog/upnextdialog.scss
+++ b/src/components/upnextdialog/upnextdialog.scss
@@ -24,6 +24,7 @@
 
 .upNextDialog-hidden {
     opacity: 0;
+    z-index: -9999!important;
 }
 
 .upNextDialog-countdownText {
@@ -52,5 +53,6 @@
 @media all and (orientation: landscape) {
     .upNextDialog {
         flex-direction: row;
+        z-index: 9999;
     }
 }
--- a/src/index.html
+++ b/src/index.html
@@ -102,6 +102,7 @@
         .mouseIdle .hide-mouse-idle,
         .mouseIdle-tv .hide-mouse-idle-tv {
             display: none !important;
+            z-index: -9999;
         }
 
         .mainDrawerHandle {


dmitrylyzo:
UpNextDialog should only be displayed when the bottom OSD is hidden.
Could you post a screenshot of which element is overlapping?

vivekkairi:
> UpNextDialog should only be displayed when the bottom OSD is hidden. Could you post a screenshot of which element is overlapping?

The OSD is displayed above the UpNext component.
image

dmitrylyzo:
> The OSD is displayed above the UpNext component.

Then the bug is displaying OSD at the same time as UpNextDialog.

Hm, why does `videoOsdBottom-hidden` have opacity 0.8?

vivekkairi:
> > The OSD is displayed above the UpNext component.
> 
> Then the bug is displaying OSD at the same time as UpNextDialog.
> 
> Hm, why does `videoOsdBottom-hidden` have opacity 0.8?

I added that 0.8 for displaying the overlap. The actual code shows opacity of 0 and has lower z-index to hide the OSD.

dmitrylyzo:
> The actual code shows opacity of 0 and has lower z-index to hide the OSD.

The `hide` class (`display: none`) is actually used to hide the OSD. With this class, the element becomes completely transparent to the user's actions.
It should be added after the hiding animation is complete:
https://github.com/jellyfin/jellyfin-web/blob/003bc94e024a6c84190ab7f4cd5bb8d0ba22ff9d/src/controllers/playback/video/index.js#L334

There was a bug where the wrong element was hidden due to duplicate views. But this was fixed in 10.8: https://github.com/jellyfin/jellyfin-web/pull/4561

Could you check if the `hide` class is actually not added to the OSD?

dmitrylyzo:
I was able to reproduce the "missing hide class" in webOS 5 emulator with Ultrachromic skin. Disabling it fixes the problem.

thornbill:
Underlying issue is fixed in https://github.com/jellyfin/jellyfin-web/pull/5653

Issue #5524, opened 2024-05-16
Jellyfin Web - Activity Dashboard, not resizeable columns


h1ght:
**Describe The Bug**

It is not Possible to resize the columns in the table in the dashboard activity.
**Steps To Reproduce**

open dashboard activity and try to resize

**Expected Behavior**
expected to resizes the columns to expand the infos withtin.
**Logs**
doesnt get needed?
**Screenshots**
![Uploading user.PNG…]()

**System (please complete the following information):**
 - Platform: Windows
 - Browser: Firefox and Edge, tested with 2 different PCs
Server-Version
10.9.1
Web-Version
10.9.1
Build-Version
10.9.1

**Additional Context**
official docker image jellyfin/jellyfin

regards

Bedz01:
Yep, same issue for me.
![JELLY](https://github.com/jellyfin/jellyfin-web/assets/10061448/d741db71-bd0d-4b49-b53c-82bda7c3ff42)

LittleBigOwI:
This isn't a bug. Jellyfin is currently using @mui/x-data-grid v.6 to render the activity page, which does not come with resizeable columns by default. 

See:
- [mui v6 docs](https://v6.mui.com/x/react-data-grid) (no resizeable columns for free version)
- [mui v7](https://mui.com/x/react-data-grid/) (resizeable columns on free version)

thornbill:
This has been added for the upcoming 10.10 release.

PR #5597, opened 2024-05-24
Migrate to x-data-grid v7

LittleBigOwI:
**Changes**

**Issues**

Migrate to v7.5.1, from 6.19.11

**Changes**
- Changed `valueGetter` and `valueFormatter` signatures as said in the MUI migration docs.

Fixes. #5524

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -42,8 +42,7 @@ const config = {
             '@mui/private-theming': '@mui/private-theming/legacy',
             '@mui/styled-engine': '@mui/styled-engine/legacy',
             '@mui/system': '@mui/system/legacy',
-            '@mui/utils': '@mui/utils/legacy',
-            '@mui/x-data-grid': '@mui/x-data-grid/legacy'
+            '@mui/utils': '@mui/utils/legacy'
         }
     },
     plugins: [
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,7 @@
         "@loadable/component": "5.16.4",
         "@mui/icons-material": "5.15.17",
         "@mui/material": "5.15.17",
-        "@mui/x-data-grid": "6.19.11",
+        "@mui/x-data-grid": "7.5.1",
         "@react-hook/resize-observer": "1.2.6",
         "@tanstack/react-query": "4.36.1",
         "@tanstack/react-query-devtools": "4.36.1",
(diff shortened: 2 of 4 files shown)


LittleBigOwI:
From my understanding, the es-check error comes from the `@mui/x-data-grid` library and not my code? Not sure.

dmitrylyzo:
> From my understanding, the es-check error comes from the `@mui/x-data-grid` library and not my code? Not sure.

In general, this means you need to add that library to the [list](https://github.com/jellyfin/jellyfin-web/blob/2d4621104ba97356ebc1efe30a18f3b31291340d/webpack.common.js#L177-L208) so that it is compatible with the old devices we are targeting (mostly TVs).

But since the Dashboard isn't accessible from TV, perhaps we need to add an exception to ES-check (allow ES6 for the Dashboard)? :thinking:

LittleBigOwI:
How exactly would adding the library to the list make a difference in compatibility with TVs? Do older TVs not support ES6? Would putting the library in the list whitelist it from checks?

dmitrylyzo:
> How exactly would adding the library to the list make a difference in compatibility with TVs?

The listed modules will be transpiled to ES2015.

> Do older TVs not support ES6?

Yes. The lowest targets are Tizen 2.3 and webOS 1.2. They are from 2015 (old WebKit ~ Chrome 27).

> Would putting the library in the list whitelist it from checks?

The list I mentioned is for babel-loader, i.e. transpilation.

For ES-check, it is https://github.com/jellyfin/jellyfin-web/blob/2d4621104ba97356ebc1efe30a18f3b31291340d/.escheckrc#L5-L10
I think it is safe to exclude Dashboard from ES-check. cc @thornbill

webpack.common.js
```diff
@@ -205,7 +205,7 @@ const config = {
                     path.resolve(__dirname, 'node_modules/ssr-window'),
                     path.resolve(__dirname, 'node_modules/swiper'),
                     path.resolve(__dirname, 'node_modules/usehooks-ts'),
-                    path.resolve(__dirname, 'src')
+                    path.resolve(__dirname, 'node_modules/')
```
dmitrylyzo, line 208:
```suggestion
                    path.resolve(__dirname, 'src')
```
and keep this list alphabetical.

But I would wait to see what Bill has to say.

LittleBigOwI: Woops sorr about that

LittleBigOwI:
Guess I'll wait for Bill before committing anything then. But is the only option to fix `es-check` to ignore the file completely like this? It seems that adding `x-data-grid` to the list of transpiled libraries doesn't fix the check. And if it does come to ignoring the library completely, I'm guessing that adding it to the list of transpiled libraries doesn't make much sense.

```json
"not": [
    "./dist/libraries/pdf.worker.js",
    "./dist/libraries/worker-bundle.js",
    "./dist/libraries/wasm-gen/libarchive.js",
    "./dist/node_modules.@mui.x-data-grid.*.js",
    "./dist/serviceworker.js"
]
```

thornbill:
> I think it is safe to exclude Dashboard from ES-check.

Yeah I agree. We will probably need a separate build of the dashboard to fully realize this, but for this PR we should be good to just ignore the data grid since it isn't used elsewhere. 👍

LittleBigOwI:
All done! Thanks guys for being helpful 👍

thornbill:
@LittleBigOwI Thanks for the PR! Please don't keep adding merge commits. Unless there is a conflict these just add noise. Thanks again!

webpack.common.js
```diff
@@ -176,6 +175,7 @@ const config = {
                 include: [
                     path.resolve(__dirname, 'node_modules/@jellyfin/libass-wasm'),
                     path.resolve(__dirname, 'node_modules/@jellyfin/sdk'),
+                    path.resolve(__dirname, 'node_modules/@mui/x-data-drid'),
```
thornbill, line 178:
```suggestion
                    path.resolve(__dirname, 'node_modules/@mui/x-data-grid'),
```

.escheckrc
```diff
@@ -6,6 +6,7 @@
         "./dist/libraries/pdf.worker.js",
         "./dist/libraries/worker-bundle.js",
         "./dist/libraries/wasm-gen/libarchive.js",
+        "./dist/node_modules.@mui.x-data-grid.*.js",
```
thornbill, line 9: It looks like this is probably not needed with the typo corrected.

thornbill (approved):

PR #5604, opened 2024-05-25
Add button to reset metadata in metadata editor

scampower3:
**Changes**

Adds a button to reset metadata inputs in the metadata editor.
https://features.jellyfin.org/posts/2235/an-option-to-reset-mis-populated-metadata
**Issues**

### DIFF
--- a/src/components/metadataEditor/metadataEditor.js
+++ b/src/components/metadataEditor/metadataEditor.js
@@ -298,6 +298,42 @@ function bindAll(elems, eventName, fn) {
     }
 }
 
+function onResetClick(e) {
+    const resetElementId = ['#txtName', '#txtOriginalName', '#txtSortName', '#txtCommunityRating', '#txtCriticRating', '#txtIndexNumber',
+        '#txtAirsBeforeSeason', '#txtAirsAfterSeason', '#txtAirsBeforeEpisode', '#txtParentIndexNumber', '#txtAlbum',
+        '#txtAlbumArtist', '#txtArtist', '#txtOverview', '#selectStatus', '#txtAirTime', '#txtPremiereDate', '#txtDateAdded', '#txtEndDate',
+        '#txtProductionYear', '#selectHeight', '#txtOriginalAspectRatio', '#select3dFormat', '#selectOfficialRating', '#selectCustomRating',
+        '#txtSeriesRuntime', '#txtTagline'];
+    const form = e.target.parentElement.parentElement.parentElement;
+    resetElementId.forEach(function (id) {
+        form.querySelector(id).value = null;
+    });
+    form.querySelector('#selectDisplayOrder').value = '';
+    form.querySelector('#listGenres').innerHTML = '';
+    form.querySelector('#listTags').innerHTML = '';
+    form.querySelector('#listStudios').innerHTML = '';
+    form.querySelector('#peopleList').innerHTML = '';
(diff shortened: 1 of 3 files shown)


src/components/metadataEditor/metadataEditor.template.html
```diff
@@ -264,6 +264,9 @@ <h2>${HeaderMetadataSettings}</h2>
                 <button is="emby-button" type="button" class="raised button-cancel block btnCancel formDialogFooterItem">
                     <span>${ButtonCancel}</span>
                 </button>
+                <button is="emby-button" type="button" class="raised button-reset block btnReset formDialogFooterItem">
```
scampower3, line 267: not sure if should be in a diff colour than the other 2 btns.

TimGels:
I am a bit conflicted with this. Such button is rarely used, if I were to guess. And showing the button prominent alongside Save and Cancel could be a UX downside. Most people don't have a need to clear all metadata fields at once, and this could cause for irritation when they accidentally click on "Reset" instead of "Save" after manually editing fields. At the same time, this is also true for "Cancel", so I am not sure how much I overthink this. But I can see the usefulness when wrong metadata is filled (although a new re-scan will just fill the fields with new metadata).

scampower3:
> Such button is rarely used, if I were to guess. 

I agree that this button is probably going to be infrequently used. Jellyfin does a pretty good job at identifying correctly.

> showing the button prominent alongside Save and Cancel could be a UX downside

Probably. Could move it onto a less prominent location too.

> But I can see the usefulness when wrong metadata is filled 

Right. The main case for this button is when wrong metadata is filled. Sometimes, it can be easier to just refill the entire metadata form, instead of checking and editing each field individually. I think the metadata editor form has enough inputs to justify a dedicated button to reset the form.

thornbill:
Let's post a screenshot of this in the ui/ux matrix room and request some feedback from Erik

thornbill:
@jellyfin-bot rebase

src/components/metadataEditor/metadataEditor.js
```diff
@@ -298,6 +298,42 @@ function bindAll(elems, eventName, fn) {
     }
 }
 
+function onResetClick(e) {
+    const resetElementId = ['#txtName', '#txtOriginalName', '#txtSortName', '#txtCommunityRating', '#txtCriticRating', '#txtIndexNumber',
+        '#txtAirsBeforeSeason', '#txtAirsAfterSeason', '#txtAirsBeforeEpisode', '#txtParentIndexNumber', '#txtAlbum',
+        '#txtAlbumArtist', '#txtArtist', '#txtOverview', '#selectStatus', '#txtAirTime', '#txtPremiereDate', '#txtDateAdded', '#txtEndDate',
+        '#txtProductionYear', '#selectHeight', '#txtOriginalAspectRatio', '#select3dFormat', '#selectOfficialRating', '#selectCustomRating',
+        '#txtSeriesRuntime', '#txtTagline'];
+    const form = e.target.parentElement.parentElement.parentElement;
```
thornbill, line 307: Untested but I think we can do this instead of chaining `parentElement`.

```suggestion
    const form = currentContext?.querySelector('form');
```

thornbill (commented):
I think this is fine for now. The metadata editor needs a UI/UX overhaul at some point anyway.

thornbill (approved):

PR #5614, opened 2024-05-27
Music Player: Control volume slider with mouse wheel & arrow keys

### LINKED ISSUE #5613: Music Player: Contains rol volume slider with mouse wheel & arrow keys.
Hello, I am interested to implement volume slider control with mass wheel and up/down keyboard events. 

I am using the player, mostly for music collection, is there anything I should know about this feature for context or as it relates to video playback? Also has this already been implemented yet in the the rewrite?

nwhitmont:
**Changes**

Add event handlers to now playing volume slider to handle scroll wheel and keyboard up/down.

**Issues**

Fixes #5613


nwhitmont:
@gnattu Is there some mode state variable that would provide contextual enabling of arrow keys for volume only in music mode?

dmitrylyzo:
NowPlayingBar is not used in TV display mode, so it is quite safe. But other volume sliders will remain without keyboard/wheel control.

FYI, slider has keyboard dragging (`enableKeyboardDragging`). But we only use it in TV display mode (`layoutManager.tv`).

src/components/nowPlayingBar/nowPlayingBar.js
```diff
@@ -822,3 +850,5 @@ document.addEventListener('viewbeforeshow', function (e) {
         }
     }
 });
+
+document.addEventListener('keydown', handleVolumeKeyInput);
```
thornbill, line 854: Event listeners should be added/removed in the same place as all the others.

nwhitmont: @thornbill I don't understand your ask here.  

The `addEventListener` is literally on the next line after the existing event listener for this component.  See line with event listener already in this file: https://github.com/jellyfin/jellyfin-web/pull/5614/files/6b568edca381c5d4ddc6cc59df1a7fcd80161482#diff-d43a0a5c9a2354bd74a607e4a1c94f73a2415d3b7ec6b3802a34f6e14926b0c0R836

If there is another location, eg. "the same place as all the others" - it is unclear which file your are referring to.

dmitrylyzo:
Oops, nowPlayingBar is not being used on TV, but the module can still be [loaded in TV display mode](https://github.com/jellyfin/jellyfin-web/blob/5d18bdf191ba0ac1d84085141fdf8caa1d7da83b/src/index.jsx#L170). Here I agree with gnattu (https://github.com/jellyfin/jellyfin-web/pull/5614#issuecomment-2135229680) - listening `keydown` from document will interfere with navigation. Could you please test your changes in the TV display mode?

nwhitmont:
> Oops, nowPlayingBar is not being used on TV, but the module can still be [loaded in TV display mode](https://github.com/jellyfin/jellyfin-web/blob/5d18bdf191ba0ac1d84085141fdf8caa1d7da83b/src/index.jsx#L170). Here I agree with gnattu ([#5614 (comment)](https://github.com/jellyfin/jellyfin-web/pull/5614#issuecomment-2135229680)) - listening `keydown` from document will interfere with navigation. Could you please test your changes in the TV display mode?
> 
> UPD: It may be better not to load the module for the TV layout rather than the TV browser.

@gnattu I'm a music-only user, & I have not used the tv/video mode at all. Can you describe the steps you had in mind to QA the TV mode? Thanks!

dmitrylyzo:
> Can you describe the steps you had in mind to QA the TV mode?

In the browser on PC:
1. Switch to TV display mode (User settings -> Display -> Display mode).
2. Save & Refresh.
3. Using arrow keys navigate to the music album.
4. Start some song.
5. Try using up/down arrow keys to see if they change the volume and simultaneously change the focus.

Lastent:
I tested the feature of using the mouse wheel to change the volume while listening
to music, and I also tested it with the keyboard arrow keys.
I used Firefox 155 on Windows 11.

**What works**

Without the PR, turning the mouse wheel over the volume control scrolls the page content.
With the PR, the volume goes down when the wheel is turned up, and goes up when the
wheel is turned down.

**Wheel direction**

I expected the wheel to work the other way around: I expected the volume to go up when
turning the wheel up, and to go down when turning it down.
When using the mouse wheel with the video player, the behavior is the expected one.

**Crash**

When pressing the arrow keys (up or down), JavaScript errors appeared.
This behavior seems to happen as soon as I load the page, but not when music is
already playing.

**Search**

The search feature does not work either, it does not let me type anything. I made a
change so that using the arrow keys no longer crashes, but the search still does
not work.

**TV mode**

To get to TV mode I went to the user settings, and under Display I chose TV mode.
Once there, the behavior was similar to the one in Auto mode: the errors appeared
when pressing the arrow keys.

**Summary**

Volume control with the mouse wheel works, but other parts stop working.

PR #5617, opened 2024-05-27
Allow VideoStreamCopy for remote source fallback

gnattu:
**Changes**

During LiveTV playback, a fallback is usually needed because the first attempt would be try to direct play the remote url of that channel. If that failed we should still allow stream copy because the playback would still success in this case. The server side will enforce the most compatible format (h264+ts) and still do transcoding if that condition is not met.

**Issues**

Depends on jellyfin/jellyfin#11851 to guarantee the stream compatibility and actually doing remux for certain input videos

### DIFF
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -3309,7 +3309,7 @@ class PlaybackManager {
 
                     changeStream(player, startTime, {
                         EnableDirectPlay: false,
-                        EnableDirectStream: false,
+                        EnableDirectStream: tryVideoStreamCopy,
                         AllowVideoStreamCopy: tryVideoStreamCopy,
                         AllowAudioStreamCopy: currentlyPreventsAudioStreamCopy || currentlyPreventsVideoStreamCopy ? false : null
                     });
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -3296,18 +3296,21 @@ class PlaybackManager {
             const streamInfo = error.streamInfo || getPlayerData(player).streamInfo;
 
             if (streamInfo?.url) {
+                const isAlreadyFallbacking = streamInfo.url.toLowerCase().includes('transcodereasons=directplayerror');
                 const currentlyPreventsVideoStreamCopy = streamInfo.url.toLowerCase().indexOf('allowvideostreamcopy=false') !== -1;
                 const currentlyPreventsAudioStreamCopy = streamInfo.url.toLowerCase().indexOf('allowaudiostreamcopy=false') !== -1;
 
                 // Auto switch to transcoding
                 if (enablePlaybackRetryWithTranscoding(streamInfo, errorType, currentlyPreventsVideoStreamCopy, currentlyPreventsAudioStreamCopy)) {
                     const startTime = getCurrentTicks(player) || streamInfo.playerStartPositionTicks;
+                    const isRemoteSource = streamInfo.item.LocationType === 'Remote';
+                    // force transcoding and only allow remuxing for remote source like liveTV, but only for initial trial
+                    const tryVideoStreamCopy = isRemoteSource && !isAlreadyFallbacking;
 
                     changeStream(player, startTime, {
-                        // force transcoding and only allow remuxing for remote source like liveTV
                         EnableDirectPlay: false,
                         EnableDirectStream: false,
-                        AllowVideoStreamCopy: streamInfo.item.LocationType === 'Remote',
+                        AllowVideoStreamCopy: tryVideoStreamCopy,
                         AllowAudioStreamCopy: currentlyPreventsAudioStreamCopy || currentlyPreventsVideoStreamCopy ? false : null
                     });
 
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -3304,10 +3304,10 @@ class PlaybackManager {
                     const startTime = getCurrentTicks(player) || streamInfo.playerStartPositionTicks;
 
                     changeStream(player, startTime, {
-                        // force transcoding
+                        // force transcoding and only allow remuxing for remote source like liveTV
                         EnableDirectPlay: false,
                         EnableDirectStream: false,
-                        AllowVideoStreamCopy: false,
+                        AllowVideoStreamCopy: streamInfo.item.LocationType === 'Remote' ? true : false,
                         AllowAudioStreamCopy: currentlyPreventsAudioStreamCopy || currentlyPreventsVideoStreamCopy ? false : null
                     });
 
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -3307,7 +3307,7 @@ class PlaybackManager {
                         // force transcoding and only allow remuxing for remote source like liveTV
                         EnableDirectPlay: false,
                         EnableDirectStream: false,
-                        AllowVideoStreamCopy: streamInfo.item.LocationType === 'Remote' ? true : false,
+                        AllowVideoStreamCopy: streamInfo.item.LocationType === 'Remote',
                         AllowAudioStreamCopy: currentlyPreventsAudioStreamCopy || currentlyPreventsVideoStreamCopy ? false : null
                     });
 
--- a/src/utils/mediaSource.ts
+++ b/src/utils/mediaSource.ts
@@ -6,9 +6,6 @@ import type { MediaSourceInfo } from '@jellyfin/sdk/lib/generated-client';
  * @returns _true_ if the media source is an HLS stream, _false_ otherwise.
  */
 export function isHls(mediaSource: MediaSourceInfo|null|undefined): boolean {
-    if (mediaSource?.TranscodingSubProtocol?.toUpperCase() === 'HLS') {
-        return true;
-    }
-
-    return mediaSource?.Container?.toUpperCase() === 'HLS';
+    return mediaSource?.TranscodingSubProtocol?.toUpperCase() === 'HLS'
+        || mediaSource?.Container?.toUpperCase() === 'HLS';
 }
--- a/src/utils/mediaSource.ts
+++ b/src/utils/mediaSource.ts
@@ -6,6 +6,9 @@ import type { MediaSourceInfo } from '@jellyfin/sdk/lib/generated-client';
  * @returns _true_ if the media source is an HLS stream, _false_ otherwise.
  */
 export function isHls(mediaSource: MediaSourceInfo|null|undefined): boolean {
-    const protocol = mediaSource?.TranscodingSubProtocol || mediaSource?.Container;
-    return protocol?.toUpperCase() === 'HLS';
+    if (mediaSource?.TranscodingSubProtocol?.toUpperCase() === 'HLS') {
+        return true;
+    }
+
+    return mediaSource?.Container?.toUpperCase() === 'HLS';
 }


src/components/playback/playbackmanager.js
```diff
@@ -3304,10 +3304,10 @@ class PlaybackManager {
                     const startTime = getCurrentTicks(player) || streamInfo.playerStartPositionTicks;
 
                     changeStream(player, startTime, {
-                        // force transcoding
+                        // force transcoding and only allow remuxing for remote source like liveTV
                         EnableDirectPlay: false,
                         EnableDirectStream: false,
-                        AllowVideoStreamCopy: false,
+                        AllowVideoStreamCopy: streamInfo.item.LocationType === 'Remote',
```
dmitrylyzo, line 3310: What happens when it fails with `AllowAudioStreamCopy: false`?
Will it try `AllowVideoStreamCopy: false`?

Like:
1. Direct play.
2. Allow copy video, allow copy audio. 👈 LiveTV retry
3. Allow copy video, deny copy audio.
4. Deny copy video, deny copy audio.

Or will it skip 4 from the above?

gnattu: It does not need to do anymore retry. This will be the last one because server-side profile will be limited to ts+h264, and the server will do transcode if the original source is not compatible with this profile. AudioStream is not relevant and the server can transcode it at will.

If there is a client that does not even play h264+ts then it basically cannot play anything.

dmitrylyzo: With prefer fMP4, there will also be `h264-mp4`.

AFAIK, the last retry is for when we can't reliably generate a device profile. There may be some quirks. 🤷

gnattu: > With prefer fMP4, there will also be h264-mp4.

No, I intentionally override it on server. The client preference does nothing. The global setting should not affect LiveTV at all due to complicated compatibility reasons. 

> AFAIK, the last retry is for when we can't reliably generate a device profile. 

For LiveTV this is almost always the case, because the fallback is super common as the direct play of remote url almost always fail for hls.js, and Safari can play it a little bit, but still not all, which the fallback to server will be very, very common. 

> There may be some quirks. 🤷

LiveTV itself is a quirk.

dmitrylyzo: > No, I intentionally override it on server. The client preference does nothing. The global setting should not affect LiveTV at all due to complicated compatibility reasons.

And then they provide you with LiveTV with HEVC. 🙃

> For LiveTV this is almost always the case, because the fallback is super common as the direct play of remote url almost always fail for hls.js, and Safari can play it a little bit, but still not all, which the fallback to server will be very, very common.

I mean, it's gonna be like:
1. Direct play.
2. Direct play, remux or transcode (only if codec not supported).
3. Direct stream or transcode (only if codec not supported).

But no full (forced) transcoding as a last chance. In case of some unexpected media error/player quirk/solar storm.

When I tested a direct HLS profile for LiveTV in Chrome, it played directly (no retires) or didn't play due to inaccessibility. Probably because the server just returned the URL without any validation.

gnattu: > And then they provide you with LiveTV with HEVC. 🙃

This will trigger transcode, what's the problem? The server is not as dumb as you think. You enable copy does not mean server will always copy. It will check if it really meet the request. 

Enabling HEVC LiveTV is way beyond the 10.9 goal. We cannot do it right without change of the API.

> But no full (forced) transcoding as a last chance. In case of some unexpected media error/player quirk/solar storm.

The users are complaining because we are doing forced transcode and skipping remux. Do you have a better way that allows testing for remux for LiveTV that actually works? Yes forced transcoding is safe, but user hates transcode very very very much.

> When I tested a direct HLS profile for LiveTV in Chrome, it played directly (no retires) or didn't play due to inaccessibility. Probably because the server just returned the URL without any validation.

Please, it did not play for **a lot** of streams. The fallback is **very very common**. Please don't pretend all livetv streamers provide good hls streams that can be played by browser that easy. Just look at the server repo and search for live tv issue and you will know that. Please don't make naive assumption about how good the direct play will work.

dmitrylyzo: > This will trigger transcode, what's the problem? The server is not as dumb as you think. You enable copy does not mean server will always copy. It will check if it really meet the request.

I know how StreamBuilder works and what is a mess. 🙃
I meant, users will complain "why HEVC -> H264 ?" and you will have to remove that "not use fMP4" block.

> The users are complaining because we are doing forced transcode and skipping remux. Do you have a better way that allows testing for remux for LiveTV that actually works? Yes forced transcoding is safe, but user hates transcode very very very much.

Just add this 4th retry in addition to the one you added in this PR so it will play with full transcode instead of playback failure.

> Please, it did not play for **a lot** of streams. The fallback is **very very common**. Please don't pretend all livetv streamers provide good hls streams that can be played by browser that easy. Just look at the server repo and search for live tv issue and you will know that. Please don't make naive assumption about how good the direct play will work.

I don't make any assumptions, I am saying how it worked when I added direct HLS profile.
AFAIK, the server does probing of LiveTV and stores media info in JSON-file somewhere (_I don't know how to refresh it, btw_).
StreamBuilder was getting actual media info to make right decision. The problem here is that the probing may fail - in my case, due to unavailability.
In 10.8, I once had to reorder this probing, placing it before the stream building. In unstable, it worked without reordering.

gnattu: > I meant, users will complain "why HEVC -> H264 ?" and you will have to remove that "not use fMP4" block.

I won't. I will tell them it is not supported. Wait for 10.10. You misunderstood the whole point here because it seems like you are totally unfamiliar with what the users are facing:

```
// Remove all fmp4 transcoding profiles, because it causes playback error and/or A/V sync issues
// Notably: Some channels won't play on FireFox and LG webOS even after transcoding
// Some channels from HDHomerun will experience A/V sync issues due to remux into fmp4
```

Plus, HEVC LiveTV never really worked before, even in 10.8 due to fmp4 support is really flawed at that time. I'm solving "but it works in 10.8" issue, not adding fancy new functions. For ATSC 3.0 TV stream the server has to support basic AC-4 audio which is also missing. So HEVC stream can definitely wait. 

> StreamBuilder was getting actual media info to make right decision.

If you are getting here the direct play attempt is already failed.

The most cursed part of this is, the actual media info can also be wrong. The bitrate could be null and causing the client to request extremely high bitrate, and if the transcode is triggered it will try to encode at 1000mbps. That's why LiveTV needs a separate encoding upper limit to overcome this situation. When the source bitrate is unavailable we need a sane value that will not fail encoders and players. I currently set the hard limit at 30mbps.

dmitrylyzo: > I won't. I will tell them it is not supported. Wait for 10.10.

Where did I say 10.9?

> If you are getting here the direct play attempt is already failed.

Ok. So it was working as it should, just for me.

> The most cursed part of this is, the actual media info can also be wrong. The bitrate could be null and causing the client to request extremely high bitrate, and if the transcode is triggered it will try to encode at 1000mbps. That's why LiveTV needs a separate encoding upper limit to overcome this situation. When the source bitrate is unavailable we need a sane value that will not fail encoders and players. I currently set the hard limit at 30mbps.

And retries are for these cases (abnormal media and hacks).
I am not against this PR, I am against removing forced transcoding (the last retry).

gnattu: > Ok. So it was working as it should, just for me.

This is normal for LiveTV. It is impossible to test what the providers are doing because most of them will do it wrong. The "supposed to work" way almost always don't work.

> I am not against this PR, I am against removing forced transcoding (the last retry).

Added.

src/components/playback/playbackmanager.js
```diff
@@ -3296,18 +3296,21 @@ class PlaybackManager {
             const streamInfo = error.streamInfo || getPlayerData(player).streamInfo;
 
             if (streamInfo?.url) {
+                const isAlreadyFallbacking = streamInfo.url.toLowerCase().includes('transcodereasons=directplayerror');
```
dmitrylyzo, line 3299: `transcodereasons=directplayerror` doesn't work because it happens when StreamBuilder didn't find the DirectProfile.

Tested by hacking `htmlVideoPlayer` and it is looping (4 tries are expected):
```diff
diff --git a/src/plugins/htmlVideoPlayer/plugin.js b/src/plugins/htmlVideoPlayer/plugin.js
index f26e89cc9..adbd7675c 100644
--- a/src/plugins/htmlVideoPlayer/plugin.js
+++ b/src/plugins/htmlVideoPlayer/plugin.js
@@ -391,7 +391,12 @@ export class HtmlVideoPlayer {
 
         return this.createMediaElement(options).then(elem => {
             return this.updateVideoUrl(options).then(() => {
-                return this.setCurrentSrc(elem, options);
+                return this.setCurrentSrc(elem, options).then((r) => {
+                    setTimeout(() => {
+                        if (this.#mediaElement) onErrorInternal(this, MediaError.MEDIA_DECODE_ERROR);
+                    }, 3000);
+                    return r;
+                });
             });
         });
     }
```

I think we should test for `transcodereasons` because `transcodereasons` is an indication of performing "transcoding" (remux, direct stream).
```suggestion
                const isAlreadyFallbacking = streamInfo.url.toLowerCase().includes('transcodereasons');
```

dmitrylyzo (commented):
Having an extra try (for remux) doesn't hurt, but I have a strong feeling that we are trying to fix it from the wrong side.

[This](https://github.com/jellyfin/jellyfin/commit/949a587a882b83c550f096afd5432e0da6727447) is what I tried for direct HLS (but it seemed to work without these changes).
My opinion, we need to probe the source before passing MediaSources to the StreamBuilder. Otherwise, it processes the raw source (until the next playback).
Not tested with STRM.

src/utils/mediaSource.ts
```diff
@@ -6,6 +6,9 @@ import type { MediaSourceInfo } from '@jellyfin/sdk/lib/generated-client';
  * @returns _true_ if the media source is an HLS stream, _false_ otherwise.
  */
 export function isHls(mediaSource: MediaSourceInfo|null|undefined): boolean {
-    const protocol = mediaSource?.TranscodingSubProtocol || mediaSource?.Container;
-    return protocol?.toUpperCase() === 'HLS';
+    if (mediaSource?.TranscodingSubProtocol?.toUpperCase() === 'HLS') {
+        return true;
+    }
+
+    return mediaSource?.Container?.toUpperCase() === 'HLS';
```
dmitrylyzo, line 13: _It is much more convenient to view the code from a PC rather than a mobile._ :sweat_smile: 
```suggestion
    return mediaSource?.TranscodingSubProtocol?.toUpperCase() === 'HLS'
        || mediaSource?.Container?.toUpperCase() === 'HLS';
```

src/components/playback/playbackmanager.js
```diff
@@ -3296,18 +3296,21 @@ class PlaybackManager {
...
                 if (enablePlaybackRetryWithTranscoding(streamInfo, errorType, currentlyPreventsVideoStreamCopy, currentlyPreventsAudioStreamCopy)) {
                     const startTime = getCurrentTicks(player) || streamInfo.playerStartPositionTicks;
+                    const isRemoteSource = streamInfo.item.LocationType === 'Remote';
+                    // force transcoding and only allow remuxing for remote source like liveTV, but only for initial trial
+                    const tryVideoStreamCopy = isRemoteSource && !isAlreadyFallbacking;
 
                     changeStream(player, startTime, {
-                        // force transcoding
                         EnableDirectPlay: false,
                         EnableDirectStream: false,
```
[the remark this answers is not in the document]
gnattu: The definition of direct stream seems to be changed multiple times in the server side and has a different meaning now... but enable it does not hurt.

dmitrylyzo (approved):

dmitrylyzo:
This PR doesn't require any changes in server. I tested it on the `releaze-10.9.z`. :point_left: _`backend` label_
Or has something changed?

gnattu:
> This PR doesn't require any changes in server. I tested it on the `releaze-10.9.z`. 👈 _`backend` label_ Or has something changed?

Not all HLS streams need the server-side change. Actually, streams strictly obeys standard should not need any server side change. The server-side change is to add compatibility for TV Tuners that is lying about the stream info like bitrate so that the client requests a wrong bitrate and triggered transcoding.

dmitrylyzo:
I mean, we can merge it without waiting for the server - these web changes are pretty universal.
It is also a bugfix (!): HLS transcoding subprotocol.

gnattu:
> I mean, we can merge it without waiting for the server - these web changes are pretty universal. It is also a bugfix (!): HLS transcoding subprotocol.

Yes you can. Nothing should break.

thornbill:
I labelled it as "backend" due to this comment in the description:

> Depends on https://github.com/jellyfin/jellyfin/pull/11851 to guarantee the stream compatibility and actually doing remux for certain input videos

I will remove that and get this merged since that isn't required. :+1:

PR #5619, opened 2024-05-28
Clear the cache view on user logout

grafixeyehero:
**Changes**
* Reset cached views on user logout  

**Issues**
https://github.com/jellyfin/jellyfin-web/issues/5615

### DIFF
--- a/src/components/ServerConnections.js
+++ b/src/components/ServerConnections.js
@@ -6,6 +6,7 @@ import Dashboard from '../utils/dashboard';
 import Events from '../utils/events.ts';
 import { setUserInfo } from '../scripts/settings/userSettings';
 import appSettings from '../scripts/settings/appSettings';
+import viewContainer from './viewContainer';
 import { queryClient } from 'utils/query/queryClient';
 
 const normalizeImageOptions = options => {
@@ -44,6 +45,8 @@ class ServerConnections extends ConnectionManager {
             credentialProvider.credentials(credentialProvider.credentials());
             // Reset the query cache
             queryClient.resetQueries();
+            // Reset cached views
+            viewContainer.reset();
 
             if (window.NativeShell && typeof window.NativeShell.onLocalUserSignedOut === 'function') {
                 window.NativeShell.onLocalUserSignedOut(logoutInfo);


src/components/ServerConnections.js
```diff
@@ -40,10 +39,10 @@ class ServerConnections extends ConnectionManager {
 
         Events.on(this, 'localusersignedout', (_e, logoutInfo) => {
             setUserInfo(null, null);
+            // Reset cached views
+            document.dispatchEvent(new CustomEvent('skinunload', {}));
```
thornbill, line 43: I would remove the code related to that event and just call the `viewContainer` function directly.

```suggestion
            viewContainer.reset();
```

src/components/ServerConnections.js
```diff
@@ -40,10 +39,10 @@ class ServerConnections extends ConnectionManager {
...
             // Ensure the updated credentials are persisted to storage
             credentialProvider.credentials(credentialProvider.credentials());
-            // Reset the query cache
-            queryClient.resetQueries();
```
thornbill, line 46: Let's move this back since it wasn't the root of the issue and we can keep this reset logic in one place.

thornbill (changes requested):

scampower3 (approved):

src/components/ServerConnections.js
grafixeyehero: the `queryClient.resetQueries()` doesn't rest the `tanstack cache`. moving the reset logic to dashboard page > logout it work as expected

grafixeyehero: Before

https://github.com/jellyfin/jellyfin-web/assets/32230989/b27b902a-5cc9-4b29-aeea-fe78b58000f2

After

https://github.com/jellyfin/jellyfin-web/assets/32230989/c6723364-f4aa-471b-958a-6a7e294fc94a

thornbill (approved):

src/components/ServerConnections.js
grafixeyehero: https://github.com/jellyfin/jellyfin-web/pull/5636

PR #5620, opened 2024-05-28
Allow to display detailLogo on TV with CSS customization

vdubuisson:
**Changes**

Add the possibility to display the detailLogo on TV with CSS customization.
If you don't customize your CSS, you won't see any difference and the detailLogo will still be hidden.

### DIFF
--- a/src/styles/librarybrowser.scss
+++ b/src/styles/librarybrowser.scss
@@ -921,8 +921,11 @@
     position: relative;
 }
 
-.layout-mobile .detailLogo {
-    display: none;
+.layout-mobile,
+.layout-tv {
+    .detailLogo {
+        display: none;
+    }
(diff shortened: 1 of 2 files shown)


thornbill (approved):

src/styles/librarybrowser.scss
```diff
@@ -921,8 +921,10 @@
     position: relative;
 }
 
-.layout-mobile .detailLogo {
-    display: none;
+.layout-mobile, .layout-tv {
```
thornbill, line 924:
```suggestion
.layout-mobile,
.layout-tv {
```

thornbill (changes requested):

tedhinklater:
woo!

PR #5621, opened 2024-05-28
Prioritise HEVC over H264 in HLS TS streams on webOS

FintasticMan:
**Changes**
Moved the HEVC codec higher in the priority ranking over H264 when transcoding to an HLS stream in a TS container.

**Issues**
None


scampower3:
Perhaps it might be better if it is user configurable setting like https://github.com/jellyfin/jellyfin-web/pull/5434

dmitrylyzo:
I don't see any problem with prioritizing HEVC for Tizen as well. Anyway, it is up to the server to make the final decision.
For audio, I tried a [different approach](https://github.com/dmitrylyzo/jellyfin-web/commit/56f234bc18ebc56b95753524171f81fef9a82eac) to not making big changes (_codecs are in reversed order_).

FintasticMan:
I think that it might be good to simply prioritise HEVC for 10.9, and then add user-configuration for 10.10. What do you think?

scampower3 (approved):

PR #5625, opened 2024-05-29
Add setting for preferred video codec

FintasticMan:
**Changes**
This is the equivalent of #5434 for the video codec. It was suggested in https://github.com/jellyfin/jellyfin-web/pull/5621#issuecomment-2136768438.

**Issues**
None


src/components/playbackSettings/playbackSettings.template.html
```diff
@@ -180,6 +180,17 @@ <h2 class="sectionTitle">
             <div class="fieldDescription checkboxFieldDescription">${EnableTrueHdHelp}</div>
         </div>
 
+        <div class="selectContainer">
+            <select is="emby-select" id="selectPreferredTranscodeVideoCodec" label="${LabelSelectPreferredTranscodeVideoCodec}">
+                <option value="">${Auto}</option>
+                <option value="h264">H264</option>
+                <option value="hevc">HEVC</option>
+                <option value="av1">AV1</option>
+                <option value="vp9">VP9</option>
```
nyanmisaka, line 189: Transcoding to VP9 is not well supported and tested. Also it's not widely available in hardware encoder. IMO there's no need to show it.

src/components/playbackSettings/playbackSettings.template.html
```diff
@@ -180,6 +180,16 @@ <h2 class="sectionTitle">
             <div class="fieldDescription checkboxFieldDescription">${EnableTrueHdHelp}</div>
         </div>
 
+        <div class="selectContainer">
+            <select is="emby-select" id="selectPreferredTranscodeVideoCodec" label="${LabelSelectPreferredTranscodeVideoCodec}">
+                <option value="">${Auto}</option>
+                <option value="h264">H264</option>
```
TimGels, line 186:
```suggestion
                <option value="h264">AVC</option>
```
I think it is more consistent if AVC was used. Given that HEVC is also used for h265.

TimGels: Fair reason. I do not know why we actually have a difference between the naming of HEVC and AVC. I had a look and our media info even shows `AVC: Yes`. But to my understanding h264 is always AVC as it is the same but with a different name.

TimGels: I agree with using h264 as we use it on other places as well.

scampower3 (approved):

src/strings/en-us.json
```diff
@@ -1391,6 +1392,7 @@
     "Season": "Season",
     "SecondarySubtitles": "Secondary Subtitles",
     "SelectAdminUsername": "Please select a username for the admin account.",
+    "SelectPreferredTranscodeVideoCodecHelp": "Select the preferred video codec to transcode to. If the preferred codec is not supported, the server will use the next best available codec.",
```
thornbill, line 1395:
```suggestion
    "SelectPreferredTranscodeVideoCodecHelp": "Select the preferred video codec for transcoding. If the preferred codec is not supported, the server will use the next best available codec.",
```

FintasticMan: I personally think that it is more obvious what it means the way I stated it, but I'm happy to change it. My way is also the same as the string for the preferred audio codec.

Issue #5280, opened 2023-12-01
[Issue]: Xbox App is completely broken using latest Jellyfin server


pokkst:
### Please describe your bug

Using the latest commits from jellyfin-git on AUR, the Xbox app is completely broken. It completely fails to load anything and always goes to the screen shown in the image.

I can confirm this issue occurs on the Xbox app on Xbox Series X, and Xbox One (shouldn't matter since the app is just the Edge browser), but also occurs across multiple networks (local and remote).

### Jellyfin Version

Other

### if other:

10.9.0

### Environment

```markdown
- OS:
- Linux Kernel:
- Virtualization:
- Clients: Jellyfin for Xbox
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Please attach any screenshots here

![issue](https://github.com/jellyfin/jellyfin/assets/47200567/2b60f8cb-a858-4b84-add6-e9b134be37c4)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

thornbill:
> They just completely deleted two of my other bug reports after proving them wrong.

Which issues do you have concerns about @pokkst?

pokkst:
I had two other bug reports that were fixed for a while, then were reintroduced in a recent commit. I was then told that that was unlikely, and after showing proof that the bugs were back (Recently Added movies/shows/music thumbnails were incorrectly sized and metadata was missing, and another where the top navbar was blank and had no "Suggestions", "Episodes", etc. buttons) the issues were deleted. I'm tired of dealing with you immature developers.

thornbill:
Marking as confirmed since there is an open PR to fix this.

RaafatAkkad:
Hi @thornbill, what's the open PR to fix this?

The https://github.com/jellyfin/jellyfin-web/tree/release-10.9.z branch works fine but the `master` branch errors after the Jellyfin splash screen. This can be reproduced by running https://github.com/jellyfin/jellyfin-uwp on a Windows PC as that uses Edge WebView (non chromium).

Server version: 10.9.9
Web version: 10.10.0 (236870e12)

```
Unexpected Application Error!
Invalid argument
TypeError: Invalid argument
   at se (http://192.168.0.5:8096/web/node_modules.%40remix-run.router.bundle.js?283bdc1b8417442b6403:2:22936)
[...]

PR #5628, opened 2024-05-29
Enter FullScreen before playback starts to facilitate DoVi support on Xbox One/Series

RaafatAkkad:
**Changes**

In order for DoVi videos to play on Xbox, the client must switch the resolution to DolbyVisionLowLatency. Currently `enableFullscreen()` is called after the video begins playing but the Xbox expects the display to be in DoVi mode before. This PR moves `enableFullscreen()` & `disableFullscreen()` to playbackmanager.js. This PR also extends the `window.NativeShell.enableFullscreen()` to take in video information so that NativeShell can switch the display resolution, hdr mode. This also enables matching the resolution & refresh rate.

Note: The Xbox is an always `full screen` device, but I have used `enableFullscreen()` to switch the resolution to match the video (typically 24hz) and then `disableFullscreen()` to switch it back to the default mode (typically 60hz)

The changes can be tested on Xbox One/Series by compiling `jellyfin-uwp` using my fork https://github.com/jellyfin/jellyfin-uwp/pull/44 . For Dolby Vision videos to play [`supportsDolbyVision() in browserDeviceProfile.js`](https://github.com/jellyfin/jellyfin-web/blob/release-10.9.z/src/scripts/browserDeviceProfile.js#L235). This will be done in ano to return `true` but that will be done in another pull request.

**Testing:**
I have tested `jellyfin-web` on
-Xbox Series X (The 1080p profile 5 video from https://developer.dolby.com/tools-media/sample-media/video-streams/dolby-vision-streams plays. Xbox Series X also plays the 4k version but jellyfin-web currently limits the Xbox to 1080p)
-Windows 11 (to check enableFullScreen/disableFullScreen works but I don't have a HDR capable monitor so playback doesn't work)

**Backwards Compatibility Testing:**
-I have also tested on Android with https://play.google.com/store/apps/details?id=org.jellyfin.mobile to ensure that enableFullScreen()/disableFullScreen() still works
-I have not tested backwards compatibility with the currently published Xbox Jellyfin app as it's not compatible with newer Jellyfin versions #5280. 

This is a new feature so not super urgent but I would like to get it into 10.9 eventually.

**Issues**


RaafatAkkad:
> A quick look.
> 
> Have you tested Theme video (video backdrops)?

Good Spot. Thanks. I didn't know this was a feature. I pushed a fix so it now works with https://emby.media/support/articles/Theme-Songs-Videos.html

dmitrylyzo:
> * Xbox Series X also plays the 4k version but jellyfin-web currently limits the Xbox to 1080p

You can pass `maxVideoWidth` with an appropriate value in the options to the profile builder to override that limit.

RaafatAkkad:
> > * Xbox Series X also plays the 4k version but jellyfin-web currently limits the Xbox to 1080p
> 
> You can pass `maxVideoWidth` with an appropriate value in the options to the profile builder to override that limit.

Thanks. Done in https://github.com/jellyfin/jellyfin-uwp/pull/46/files#diff-a79f550a8266caae0f42f8b32b0afbf4450f95a096dac613d35d39b7b6eebd9a

thornbill:
Is #5669 dependent on this?

RaafatAkkad:
> Is #5669 dependent on this?

Yes, DoVi videos fail with an error if the display is not in Dolby Vision mode and the jellyfin server will transcode the video.

RaafatAkkad:
Hi @dmitrylyzo, Currently `master` does not work with the Xbox app which uses Edge WebView (Non Chromium) https://github.com/jellyfin/jellyfin-web/issues/5280#issuecomment-2276627333. Once a workaround is found or jellyfin-uwp is moved to Edge WebView2 (Chromium based) then I will rebase this.

PR #5629, opened 2024-05-29
Update supportsDolbyVision to return true on Xbox WebView2

RaafatAkkad:
**Changes**
Depends on
https://github.com/jellyfin/jellyfin-web/pull/5628
https://github.com/jellyfin/jellyfin-uwp/pull/44

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -235,6 +235,7 @@ function supportsHlg(options) {
 function supportsDolbyVision(options) {
     return options.supportsDolbyVision ?? (false // eslint-disable-line sonarjs/no-redundant-boolean
             || browser.safari && ((browser.iOS && browser.iOSVersion >= 13) || browser.osx)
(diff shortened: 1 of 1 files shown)


src/scripts/browserDeviceProfile.js
```diff
@@ -235,6 +235,7 @@ function supportsHlg(options) {
 function supportsDolbyVision(options) {
     return options.supportsDolbyVision ?? (false // eslint-disable-line sonarjs/no-redundant-boolean
             || browser.safari && ((browser.iOS && browser.iOSVersion >= 13) || browser.osx)
+            || browser.edgeChromium && browser.xboxOne && window.NativeShell
```
dmitrylyzo, line 238: For apps, pass `supportsDolbyVision: true` in the options to the profile builder from the NativeShell implementation.

RaafatAkkad: Good point. Thanks. Added in https://github.com/jellyfin/jellyfin-uwp/pull/44

PR #5640, opened 2024-05-31
Fix background being invisible with theme videos

mihawk90:
**Changes**
I noticed that since I updated to 10.9 the skin's background would disappear when a theme video was playing and therefore some text would be unreadable depending on the video in the background.

Before:
image

After:
image

This basically restores pre-10.9 behaviour.

**Issues**
Couldn't find any open issues for this.

---

Calling in @grhallenbeck since the cause was introduced in ce4c7aed5e2f9315e62b7e9aa364700d112aebc3 / #5076 and there might have been a reason I'm not seeing.

The `videoPlayerContainer` and `backgroundContainer` elements are the first 2 elements in the DOM structure though. They are already at the lowest layer of elements, making `z-index` seem redundant here. The issue is that the `videoPlayerContainer` specifies no `z-index` at all so the `-1` made the `backgroundContainer` be rendered behind the video (or culled entirely, not sure how the actual rendering pipeline works :P ).

### DIFF
--- a/src/styles/site.scss
+++ b/src/styles/site.scss
@@ -39,7 +39,6 @@ body {
     right: 0;
     bottom: 0;
     contain: strict;
-    z-index: -1;
 }
 
 .layout-mobile,


dmitrylyzo:
I think this should be rebased on the `release-10.9.z` branch.
Then edit the target branch using `Edit` next to the title.

mihawk90:
Sorry my bad, I didn't think of targeting the release branch.

Welp, something went wrong with that rebase :rofl:

mihawk90:
Ah there we go, I should have updated the base before pushing I guess...

dmitrylyzo (approved):
I also don't know what the original purpose of this `z-index` is. Perhaps it is a leftover.
Seems to work as intended without it.

thornbill:
(rebased to trigger CI)

PR #5645, opened 2024-06-02
Migrating home to react with reactquery and mui

bu3alwa:
I am looking to contribute to jellyfin since I use the application a lot. I looked the projects plans and it seems like there is a plan for experimental to be on react using react-query and MUI. I think I can help in that effort as I familiar in that section. 

I started experimenting and looking at the code and looking at the effort it would take to migrate the Home page. I see a lot of hybrid components that seem to rely on the old way of doing things. I assume that should be slowly removed moving forward. 

MUI is not my favorite library to deal with but I can work with it but I am not sure how style management should be done because I read somewhere there is a plan to move to sass as well. But there are other components using sx and I used it for this draft.

What I want to know is is this the proper path to continue working on this section of the code? It is currently not done and there is more I need to understand with the api and all the features to be done. But before I put more effort into it I wanted to get thoughts and feedback if this is what I should be doing. And if its not then what should I a be doing to make it follow the goals of the project.


bu3alwa:
@thornbill sorry for the ping but don't know if a draft notifies maintainers and I don't want to clutter the issues with opening a ticket.

thornbill:
No worries, I just hadn't got back to responding to this. :sweat_smile:

At a high level you are definitely on the right path. For Sass specifically, that was prior to us starting to use MUI components (actually before any React was used). I need to sit down and write some documentation about some of these topics... that is pretty high on my to-do list now that 10.9 has shipped.

I only looked over the code changes briefly, but I did notice a couple things:

* There is no need to support the tab stuff for the home screen in these new implementations. That was just there for compatibility with the stable layout. Just add dedicated routes for `/home` and `/favorites` and add redirects.
* I wouldn't rewrite the Card component as part of this effort. I would recommend using the existing [SectionContainer](https://github.com/jellyfin/jellyfin-web/blob/master/src/apps/experimental/components/library/SectionContainer.tsx) component for the home screen sections also. I'm sure at some point we will need to do a full port of these components to MUI at some point, but that would majorly increase the size of this PR and the cards are likely to be different in the future redesign effort.

bu3alwa:
Ok thanks for the feedback. Ill give it another go this weekend. Yea I was thinking of tackling the Section container to MUI as well but that would make the pr much larger. I can work on those later possibly.

There was some stuff that I removed on the home that had the resume stuff. I think if thats needed it would be better to move to the /video page and maybe pass an id or the item with a playback position index in search params. That is something that might be good for refreshing the page and still staying on the video you are on.

bu3alwa:
@thornbill So I am a bit confused with the api calls and a bit lost. And the api documentation page is very slow to search around with not sure why. I have a few questions 

1. We have the continue watching but it seems to be separated by Movies and Shows but the current iteration combines both. Can I query the api to get both instead of each separately. 
2. When I want to iterate through latestmedia is there a way to get the list of libraries with their types. I see ordered views under the current user but that includes collections and playlists and I don't think I should be iterating those.
3. We have `SectionApiMethods` but `ResumeItems` seems to be called `resume` from the api and the casing as well doesn't match here. Was there changes to the api should I update these as well?
4. I have multiple components for `NextUp`, `RecentlyAdded`, etc. Should those all be combined into one or is it fine for it to be separated like that. I think I just don't understand how to get all that I need with api calls yet and separating them was the easiest but I dunno if it's possible to have them all combined.

I pushed the latest of what I have right now and will update more once I get a better understanding of the api.

bu3alwa:
I think I am getting somewhere with it and understanding the API

bu3alwa:
Should I create a ticket for this or can I use a ticket that exists somewhere? @thornbill

bu3alwa:
Ready for a review

src/utils/sections.ts
```diff
@@ -172,7 +172,7 @@ export const getProgramSections = (): Section[] => {
     return [
         {
             name: 'HeaderOnNow',
-            itemTypes: 'Programs',
+            itemTypes: 'Program',
```
grafixeyehero, line 175: here to the other

src/utils/sections.ts
```diff
@@ -89,7 +89,7 @@ export const getSuggestionSections = (): Section[] => {
         {
             name: 'NextUp',
             apiMethod: SectionApiMethod.NextUp,
-            itemTypes: 'nextup',
+            itemTypes: 'Episode',
```
grafixeyehero, line 92: the previous value was intentionally kept for compatibility for list page

src/utils/sections.ts
bu3alwa: ok ill just make a new getHomeSections with a different type to match whats sent from the api. Or I could combine both types in Section. itemTypes was typed string which I think should just be the string literals instead.

grafixeyehero: you can use 'parametersOptions'

src/utils/sections.ts
bu3alwa: reverted

src/utils/sections.ts
bu3alwa: oh yea that makes sense. Let me fix it.

src/hooks/useFetchItems.ts
```diff
@@ -897,3 +932,169 @@ export const useGetProgramsSectionsWithItems = (
...
+    const currentApi = useApi();
+    return useQuery({
+        queryKey: ['CurrentUser'],
+        queryFn: ({ signal }) =>
+            fetchGetCurrentUser(currentApi, { signal })
+    });
+};
+
+// not provided in the SDK, delete once added
+function getDisplayPreferencesAPI(api: Api): DisplayPreferencesApi {
```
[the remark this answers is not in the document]
bu3alwa: resolved

src/utils/sections.ts
bu3alwa: ok pushed an update. What I was doing was redundant. Thanks

bu3alwa:
anything else I can work on while this is in review?

src/apps/experimental/components/library/HomeSectionsView.tsx
```diff
@@ -0,0 +1,127 @@
...
+            return [SectionType.NextUp];
+        }
+
+        if (sectionType === HomeSectionType.LatestMedia && view?.CollectionType === 'movies') {
+            return [SectionType.LatestMovies];
+        }
+
+        if (sectionType === HomeSectionType.LatestMedia && view?.CollectionType === 'tvshows') {
+            return [SectionType.LatestEpisode];
+        }
```
grafixeyehero, line 35: you are missing collection type like books, photos, musics...

src/apps/experimental/components/library/HomeSectionsView.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+        if (sectionType === HomeSectionType.Resume) {
```
grafixeyehero, line 37: missing HomeSectionType ResumeAudio and ResumeBook

src/utils/sections.ts
```diff
@@ -368,3 +368,136 @@ export const getProgramSections = (): Section[] => {
...
+                shape: CardShape.SquareOverflow,
+                showParentTitle: true,
+                overlayPlayButton: true,
+                coverImage: true
+            }
+        },
+        {
+            name: 'HeaderRecentlyPlayed',
+            itemTypes: 'Audio',
+            type: SectionType.RecentlyPlayedMusic,
```
grafixeyehero, line 467: this feature not used on home section type

src/utils/sections.ts
```diff
@@ -368,3 +368,136 @@ export const getProgramSections = (): Section[] => {
...
+                showParentTitle: true,
+                action: 'instantmix',
+                overlayMoreButton: true,
+                coverImage: true
+            }
+        },
+        {
+            name: 'HeaderFrequentlyPlayed',
+            itemTypes: 'Audio',
+            type: SectionType.FrequentlyPlayedMusic,
```
grafixeyehero, line 486: this too maybe we can add in the future.

grafixeyehero (commented):
> There was some stuff that I removed on the home that had the resume stuff. I think if thats needed it would be better to move to the /video page and maybe pass an id or the item with a playback position index in search params. That is something that might be good for refreshing the page and still staying on the video you are on.

I have a concern with removing stuff.
additionally, @thornbill will provide you feedback

src/apps/experimental/routes/asyncRoutes/user.ts
```diff
@@ -9,5 +9,7 @@ export const ASYNC_USER_ROUTES: AsyncRoute[] = [
     { path: 'tv.html', page: 'shows', type: AsyncRouteType.Experimental },
     { path: 'music.html', page: 'music', type: AsyncRouteType.Experimental },
     { path: 'livetv.html', page: 'livetv', type: AsyncRouteType.Experimental },
-    { path: 'mypreferencesdisplay.html', page: 'user/display', type: AsyncRouteType.Experimental }
+    { path: 'mypreferencesdisplay.html', page: 'user/display', type: AsyncRouteType.Experimental },
+    { path: 'favorites.html', page: 'favorites', type: AsyncRouteType.Experimental },
+    { path: 'video.html', page: 'video' }
```
grafixeyehero, line 14: this can be removed
```
    { path: 'video.html', page: 'video' }
```

src/apps/experimental/components/library/HomeSectionsView.tsx
bu3alwa: ok, I don't have those in my library I need to add some and figure that out.

grafixeyehero: @bu3alwa if you allow me, I made some changes I will push the change or I will create a PR

bu3alwa: added you to the repo

grafixeyehero: @bu3alwa here is the change I made  https://github.com/grafixeyehero/jellyfin-web/commits/convert-home-page-to-react/

bu3alwa: I added some comments on them. Let me know what you think?

grafixeyehero: @bu3alwa updated

bu3alwa: are you pushing it to this branch? or should I be doing something to it?

grafixeyehero: here https://github.com/bu3alwa/jellyfin-web/pull/1

grafixeyehero (approved):

bu3alwa:
Should I create an issue for this or is this fine?

bu3alwa:
Is this not wanted anymore @grafixeyehero @thornbill?

PR #5651, opened 2024-06-03
move up next dialog to front

### LINKED ISSUE #5650: Up Next container buttons not clickable
**Describe The Bug**
Up Next Video container buttons not clickable. Instead they fast forward./backwards.

**Steps To Reproduce**
1. Watch any media in a TV series until the last minute of the episode. Up next container appears.
2. Click any button in the Up Next container.

**Expected Behavior**
Buttons should perform their regular operation.

 - Platform: Windows
 - Browser: Chrome
 - Jellyfin Version: 10.9.0

pvico12:
**Changes**
Up next dialog was not at the upmost front of the page, so the buttons were rendered unclickable and instead triggered the click event on the fast forward/backward button. Z-index added to the dialog so that it as at the top of the screen when it appears and the buttons register the click event properly to operate as intended.

**Issues**
Fixes #5650

### DIFF
--- a/src/components/upnextdialog/upnextdialog.scss
+++ b/src/components/upnextdialog/upnextdialog.scss
@@ -3,6 +3,7 @@
     right: 0;
     bottom: 0;
     width: 30em;
+    z-index: 1000;
     padding: 1em;
     display: flex;
     flex-direction: column;
--- a/src/components/upnextdialog/upnextdialog.scss
+++ b/src/components/upnextdialog/upnextdialog.scss
@@ -3,7 +3,6 @@
     right: 0;
     bottom: 0;
     width: 30em;
-    z-index: 1000;
     padding: 1em;
     display: flex;
     flex-direction: column;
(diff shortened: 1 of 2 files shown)


dmitrylyzo:
OSD should have `hide` class after animationg is complete.
The problem is probably that the custom CSS breaks animation: https://github.com/jellyfin/jellyfin-web/pull/5595#issuecomment-2141960724

I think we just need to swap these (adding class and event subscription):
https://github.com/jellyfin/jellyfin-web/blob/d311dec1db973e5dbc0359ebd1dea6fb8311a760/src/controllers/playback/video/index.js#L365-L369

pvico12:
Swapping these two does not do the trick. The videoOsdBottom-hidden class just sets the opacity of the OsdBottom to 0, which does not prevent it from being clicked.

dmitrylyzo:
#5653

pvico12:
Noted, closing this PR.

PR #5659, opened 2024-06-05
Fixed being unable to properly long press on cards to multiselect on Firefox

ConnorS1110:
**Changes**
In `showSelected` we were not adding the `checkedInitial` class for Firefox when selecting an element. This did not allow for proper selection of the cards. There was a comment that claimed this was because Firefox did not handle click events the same as other browsers, but it seems this was either not true, or was maybe misunderstood. I simply removed the condition and it has restored proper multiselect functionality on Firefox.

**Issues**
None that I know of, but this has been bothering me for many months now


thornbill:
That has been around [forever](https://github.com/jellyfin/jellyfin-web/blob/de6ac33ec186d63eaf2e77800a61d2d743357edd/src/bower_components/emby-webcomponents/multiselect/multiselect.js#L104) so it's likely just outdated. If you are able, please update this to use `release-10.9.z` as the target branch.

ConnorS1110:
I've really messed something up in a rebase I am just going to branch off the release branch and make my change there

oddstr13:
I'd like to note that it *is* possible to use the multiselect in Firefox right now, but a long-click does not work, you have to long-press into a drag :smile:

ConnorS1110:
Right it still just feels really bad. It should work like all the other browsers now

oddstr13:
Yep, but it's a workaround for people using old versions, should work on pretty much any Jellyfin version, it's been quite a while since I discovered it :wink:

PR #5660, opened 2024-06-05
Fixed being unable to properly long press on cards to multiselect on Firefox

ConnorS1110:
**Changes**
In `showSelected` we were not adding the `checkedInitial` class for Firefox when selecting an element. This did not allow for proper selection of the cards. There was a comment that claimed this was because Firefox did not handle click events the same as other browsers, but it seems this was either not true, or was maybe misunderstood. I simply removed the condition and it has restored proper multiselect functionality on Firefox.

**Issues**
None that I know of, but this has been bothering me for many months now


ConnorS1110:
@thornbill here is my fixed PR

nielsvanvelzen (approved):

PR #5666, opened 2024-06-06
Update up-next text when auto play next episode is disabled

### LINKED ISSUE #5643: "Next episode" box language implies the next episode will autoplay even when autoplay is off
This one is fairly straight forward. I have "play next episode automatically" off, but I have "show next video info during playback" on.

The box that appears with about 30 seconds to go says "Next episode playing in [countdown]" with a "Start now" button. It doesn't actually auto-play.

This is all well and good, but the verbiage should not suggest that the next episode will play if the autoplay feature is off. The verbiage of that first line should be changed to something like simply "Next episode:" 

I noted this bug that was resolved relatively recently that _might_ be related, so I thought I'd link it: https://github.com/jellyfin/jellyfin-web/pull/5398

**Steps To Reproduce**
1. Settings > playback
2. Play next episode automatically [off]
3. Show next video info during playback [on]
4. Play a TV episode that isn't the last episode of a show. Skip to the last minute and wait for the popup to appear

NextEp1

I don't think logs are relevant unless someone tells me that the wording actually is supposed to be different if auto-play is off - I'm assuming this was just an oversight that the wording is currently coded to be the same whether autoplay is on or off. If it should already be using different language, let me know and I can produce logs.

 - Platform: Windows 10
 - Browser: Chrome (10.9.3) and Jellyfin Media Player (10.9.2)
 - Jellyfin Version: 10.9.3

**Additional Context**

ConnorS1110:
**Changes**
When `Play next episode automatically` is disabled, but the up-next popup is enabled, it still shows a countdown timer, which is misleading because it will not auto play the next episode. It now checks user configuration settings to see if you have the popup enabled and have `EnableNextEpisodeAutoPlay` set to false, which will now just show `Next Episode/Video` instead of also including the countdown timer.

**Issues**
Fixes #5643


src/strings/en-us.json
```diff
@@ -441,7 +441,9 @@
     "HeaderNewDevices": "New Devices",
     "HeaderNewRepository": "New Repository",
     "HeaderNextEpisodePlayingInValue": "Next Episode Playing in {0}",
+    "HeaderNextEpisodePlayingInValueNoAutoPlay": "Next Episode",
     "HeaderNextVideoPlayingInValue": "Next Video Playing in {0}",
+    "HeaderNextVideoPlayingInValueNoAutoPlay": "Next Video",
```
thornbill, line 446: We could consolidate these a bit...
```suggestion
    "HeaderNextEpisodePlayingInValue": "Next Episode Playing in {0}",
    "HeaderNextEpisodePlayingInValueNoAutoPlay": "Next Episode",
    "HeaderNextVideoPlayingInValue": "Next Video Playing in {0}",
    "HeaderNextVideoPlayingInValueNoAutoPlay": "Next Video",
```
```suggestion
    "HeaderNextEpisodePlayingInValueNoAutoPlay": "Next Episode",
    "HeaderNextItemPlayingInValue": "{0} Playing in {1}",
    "HeaderNextVideoPlayingInValueNoAutoPlay": "Next Video",
```

Then we can pass the item type text ("Next Episode"/"Next Video") to the translate function in addition to the time.

ConnorS1110: There was actually no reason not to consolidate further, so I did. That header text can go back to being a ternary now with these changes instead of an if/else with nested ternaries inside of it

src/components/upnextdialog/upnextdialog.js
```diff
@@ -54,9 +55,9 @@ function setNextVideoText() {
 
     const timeText = '<span class="upNextDialog-countdownText">' + globalize.translate('HeaderSecondsValue', secondsRemaining) + '</span>';
 
-    const nextVideoText = instance.itemType === 'Episode' ?
-        globalize.translate('HeaderNextEpisodePlayingInValue', timeText) :
-        globalize.translate('HeaderNextVideoPlayingInValue', timeText);
+    const nextVideoText = instance.showStaticNextText ?
+        globalize.translate('HeaderNextItem', instance.itemType) :
```
thornbill, line 59: itemType should be localized

ConnorS1110: But the rest of the string isn't? What are you suggesting I should be doing instead

thornbill: When calling the `translate` function the first parameter is the key used to lookup a translation and additional parameters replace the placeholders in the translation string, but they are not treated as translation keys themselves. You will need to do something like:

```js
globalize.translate('Key', globalize.translate(itemType));
```

ConnorS1110: Oh my bad. I thought you were saying itemType is ALREADY localized and that I should be doing something else so I can not localizing an already localized string. I'll update this then

ConnorS1110: Updated

thornbill: Ah yeah my bad. I wasn't very clear in my first comment. 😅

thornbill (approved):

854562:
I saw the new strings from this PR pop up on Weblate and have noticed that the separate strings for "next episode" and "next video" have been combined into one string with a variable: `Next {0} Playing in {1}`, in which `{0}` is either `Episode` or `Video`. I believe this creates potential localisation issues.

1. In many languages, the form of the adjective (`next`) depends on the grammatical gender of the noun it pertains to (`episode`, `video`). For languages in which the translations for "episode" and "video" are of different grammatical genders, this creates a problem.

2. In English, Title Case is very common for dialogue texts like this, but in many languages (including Dutch, which I am most familiar with), Title Case is Never a Thing and sentence case is always used. Therefore, "episode" and "video" should not start with a capital letter in these languages, which this string change makes impossible.

I realise these points, particularly the second, might come across as nitpicky, but I did want to point them out regardless. In summary, I would suggest reverting the string changes, since as far as I understand, the previous situation worked just fine.

PR #5670, opened 2024-06-06
Disable PiP button on Xbox as it doesn't support it

RaafatAkkad:


src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1719,7 +1719,7 @@ export class HtmlVideoPlayer {
                     .isViewModeSupported(Windows.UI.ViewManagement.ApplicationViewMode.compactOverlay))
             // Check standard PiP support
             || document.pictureInPictureEnabled
-        ) {
+        ) && !browser.xboxOne) {
```
thornbill, line 1722: I’m guessing this is currently falling under the check for windows? Maybe this church could be moved there?

thornbill:

PR #5671, opened 2024-06-06
Update pdfjs due to xss vulnerability

thornbill:
**Changes**
Updates pdfjs to prevent a XSS vulnerability via malicious PDF files

**NOTE:** I will backport this manually because the lockfile change will likely cause issues

**Issues**
Closes #5462


Bond-009 (approved):

thornbill:
Will redo this for master since it became more complicated, but the XSS vulnerability is mitigated by #5694

PR #5685, opened 2024-06-09
play/pause when Enter is pressed on non-dragging position slider

shoffing:
Hey there, first time contributor here! Open to any and all feedback.

# Changes

Background:  This fixes a minor annoyance while using Jellyfin on my LG TV that does not have a play/pause media button on the remote (only an "OK" button between the navigation arrows):

image

## **Current behavior**

Pressing "Enter" on the now playing position slider when it is not being dragged does nothing. *This means resuming playback after scrubbing the position slider requires 4 inputs every time to manually click the OSD Play button: `Down -> Right -> Right -> OK`*

## **New behavior**

Pressing "Enter" on the position slider when it is not being dragged will play/pause the current media. *Now resuming playback after scrubbing only requires one remote input: `OK`*

To achieve this, I needed to dispatch a new event "playpause" from the emby-slider component, and listen for that event in the playback/video controller.

## Manual testing

I tested these changes manually using LG's simulator.

### DIFF
--- a/src/controllers/playback/video/index.js
+++ b/src/controllers/playback/video/index.js
@@ -1775,8 +1775,13 @@ export default function (view) {
         }
     });
 
-    nowPlayingPositionSlider.addEventListener('playpause', function () {
-        playbackManager.playPause(currentPlayer);
+    nowPlayingPositionSlider.addEventListener('keydown', function (e) {
+        if (e.defaultPrevented) return;
+
+        const key = keyboardnavigation.getKeyName(e);
+        if (key === 'Enter') {
+            playbackManager.playPause(currentPlayer);
+        }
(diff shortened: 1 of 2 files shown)


dmitrylyzo:
Play/pause should not be in the slider - it is a generic element that is not tied to playback. You probably need to catch `click` event from the slider when it was not processed by the slider.

dmitrylyzo:
I suggest you not to use your `master` branch for PRs. Create a separate branch from your current `master` (_it will contain current changes_) and open a new PR with the new branch (_AFAIK, there's no way to change the source branch of PR_).

shoffing:
Closed in favor of #5687

PR #5687, opened 2024-06-09
play/pause when Enter is pressed on non-dragging position slider

shoffing:
(new PR created because #5685 was wrong branch)

Hey there, first time contributor here! Open to any and all feedback.

# Changes

Background:  This fixes a minor annoyance while using Jellyfin on my LG TV that does not have a play/pause media button on the remote (only an "OK" button between the navigation arrows):

image

## **Current behavior**

Pressing "Enter" on the now playing position slider when it is not being dragged does nothing. *This means resuming playback after scrubbing the position slider requires 4 inputs every time to manually click the OSD Play button: `Down -> Right -> Right -> OK`*

## **New behavior**

Pressing "Enter" on the position slider when it is not being dragged will play/pause the current media. *Now resuming playback after scrubbing only requires one remote input: `OK`*

To achieve this, I play/pause the video in a new "keydown" listener attached to the position slider in the video controller. It does not play/pause if default was prevented (which happens when the slider is dragged (left/right arrows) and Enter is pressed).

## Manual testing

I tested these changes manually using LG's simulator.

### DIFF
--- a/src/controllers/playback/video/index.js
+++ b/src/controllers/playback/video/index.js
@@ -1775,8 +1775,13 @@ export default function (view) {
         }
     });
 
-    nowPlayingPositionSlider.addEventListener('playpause', function () {
-        playbackManager.playPause(currentPlayer);
+    nowPlayingPositionSlider.addEventListener('keydown', function (e) {
+        if (e.defaultPrevented) return;
+
+        const key = keyboardnavigation.getKeyName(e);
+        if (key === 'Enter') {
+            playbackManager.playPause(currentPlayer);
+        }
     });
 
     nowPlayingPositionSlider.updateBubbleHtml = function(bubble, value) {
--- a/src/controllers/playback/video/index.js
+++ b/src/controllers/playback/video/index.js
@@ -1775,6 +1775,10 @@ export default function (view) {
         }
     });
 
+    nowPlayingPositionSlider.addEventListener('playpause', function () {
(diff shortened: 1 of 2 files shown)


src/controllers/playback/video/index.js
```diff
@@ -1775,6 +1775,15 @@ export default function (view) {
         }
     });
 
+    nowPlayingPositionSlider.addEventListener('keydown', function (e) {
+        if (e.defaultPrevented) return;
```
shoffing, line 1779: an alternative here would be to call `stopImmediatePropagation()` in emby-slider.js, but I'm guessing we'd prefer not to updated a core component's source if possible?

dmitrylyzo: Technically, keyboard dragging should probably do `stopImmediatePropagation()`, since it is an internal behaviour that overrides the standard one. But I would use your approach because it is currently safer. Moreover, one day emby-slider will be replaced by another one from the MUI.

shoffing: Alright, sounds good!

dmitrylyzo (approved):
LGTM. Commits can be squashed.

We probably need to fix `Space` when the focus is on the slider (not in this PR). Currently it moves the focus to `btnPause`: https://github.com/jellyfin/jellyfin-web/blob/2f4d6effe0c57d00e9b33a9194e96aec26f64073/src/controllers/playback/video/index.js#L1204

PR #5688, opened 2024-06-09
Add support for native PGS subtitle rendering without transcoding

Arcus92:
**Changes**
This PR allows the client to request and render native PGS subtitle streams without transcoding. PGS is a graphical subtitle format. The subtitles are rendered locally by the client on-top of the video element.

This adds a new dependency [`libpgs-js`](https://github.com/Arcus92/libpgs-js). I just created this library for the Jellyfin project. If desired, you can fork it into the @jellyfin domain.

**Tasks**
- [x] Disable native PGS support if 'Subtitle burn-in' is not set to 'auto'.
- [x] Improve performance by using WebWorkers to render subtitles in a separate thread.

**Related Pull Request (backend):** https://github.com/jellyfin/jellyfin/pull/12056
**Related Pull Request (vue):** https://github.com/jellyfin/jellyfin-vue/pull/2404

### DIFF
--- a/src/components/subtitlesettings/subtitlesettings.js
+++ b/src/components/subtitlesettings/subtitlesettings.js
@@ -117,15 +117,10 @@ function onSubtitleModeChange(e) {
 
 function onSubtitleBurnInChange(e) {
     const view = dom.parentWithClass(e.target, 'subtitlesettings');
-
     const fieldRenderPgs = view.querySelector('.fldRenderPgs');
 
     // Pgs option is only available if burn-in mode is set to 'auto' (empty string)
-    if (this.value) {
-        fieldRenderPgs.classList.add('hide');
-    } else {
-        fieldRenderPgs.classList.remove('hide');
-    }
+    fieldRenderPgs.classList.toggle('hide', !!this.value);
 }
 
 function onAppearanceFieldChange(e) {
--- a/src/components/subtitlesettings/subtitlesettings.js
+++ b/src/components/subtitlesettings/subtitlesettings.js
@@ -61,6 +61,9 @@ function loadForm(context, user, userSettings, appearanceSettings, apiClient) {
         context.querySelector('#sliderVerticalPosition').value = appearanceSettings.verticalPosition;
 
         context.querySelector('#selectSubtitleBurnIn').value = appSettings.get('subtitleburnin') || '';
+        context.querySelector('#chkSubtitleRenderPgs').checked = appSettings.get('subtitlerenderpgs') === 'true';
+
+        context.querySelector('#selectSubtitleBurnIn').dispatchEvent(new CustomEvent('change', {}));
 
         onAppearanceFieldChange({
             target: context.querySelector('#selectTextSize')
@@ -86,6 +89,7 @@ function save(instance, context, userId, userSettings, apiClient, enableSaveConf
     loading.show();
 
     appSettings.set('subtitleburnin', context.querySelector('#selectSubtitleBurnIn').value);
+    appSettings.set('subtitlerenderpgs', context.querySelector('#chkSubtitleRenderPgs').checked);
 
     apiClient.getUser(userId).then(function (user) {
         saveUser(context, user, userSettings, instance.appearanceKey, apiClient).then(function () {
@@ -109,6 +113,19 @@ function onSubtitleModeChange(e) {
         subtitlesHelp[i].classList.add('hide');
     }
     view.querySelector('.subtitles' + this.value + 'Help').classList.remove('hide');
+}
+
+function onSubtitleBurnInChange(e) {
+    const view = dom.parentWithClass(e.target, 'subtitlesettings');
+
+    const fieldRenderPgs = view.querySelector('.fldRenderPgs');
+
+    // Pgs option is only available if burn-in mode is set to 'auto' (empty string)
+    if (this.value) {
+        fieldRenderPgs.classList.add('hide');
+    } else {
+        fieldRenderPgs.classList.remove('hide');
+    }
 }
 
 function onAppearanceFieldChange(e) {
@@ -166,6 +183,7 @@ function embed(options, self) {
     options.element.querySelector('form').addEventListener('submit', self.onSubmit.bind(self));
 
     options.element.querySelector('#selectSubtitlePlaybackMode').addEventListener('change', onSubtitleModeChange);
+    options.element.querySelector('#selectSubtitleBurnIn').addEventListener('change', onSubtitleBurnInChange);
     options.element.querySelector('#selectTextSize').addEventListener('change', onAppearanceFieldChange);
     options.element.querySelector('#selectTextWeight').addEventListener('change', onAppearanceFieldChange);
     options.element.querySelector('#selectDropShadow').addEventListener('change', onAppearanceFieldChange);
--- a/src/components/subtitlesettings/subtitlesettings.template.html
+++ b/src/components/subtitlesettings/subtitlesettings.template.html
@@ -37,7 +37,6 @@
             <label>
                 <input is="emby-checkbox" type="checkbox" id="chkSubtitleRenderPgs" />
                 <span>${RenderPgsSubtitle}</span>
-
             </label>
             <div class="fieldDescription checkboxFieldDescription">${RenderPgsSubtitleHelp}</div>
         </div>
--- a/src/components/subtitlesettings/subtitlesettings.template.html
+++ b/src/components/subtitlesettings/subtitlesettings.template.html
@@ -32,6 +32,15 @@
             </select>
             <div class="fieldDescription">${BurnSubtitlesHelp}</div>
         </div>
+
+        <div class="checkboxContainer checkboxContainer-withDescription fldRenderPgs hide">
+            <label>
+                <input is="emby-checkbox" type="checkbox" id="chkSubtitleRenderPgs" />
+                <span>${RenderPgsSubtitle}</span>
+
+            </label>
+            <div class="fieldDescription checkboxFieldDescription">${RenderPgsSubtitleHelp}</div>
+        </div>
     </div>
 
     <div class="verticalSection subtitleAppearanceSection hide">
--- a/package-lock.json
+++ b/package-lock.json
@@ -45,6 +45,7 @@
         "jquery": "3.7.1",
         "jstree": "3.3.16",
         "libarchive.js": "2.0.2",
+        "libpgs": "0.2.0",
         "lodash-es": "4.17.21",
         "markdown-it": "14.1.0",
         "material-design-icons-iconfont": "6.7.0",
@@ -136,6 +137,22 @@
       },
       "optionalDependencies": {
         "sass-embedded": "1.77.8"
+      }
+    },
+    "../libpgs-js": {
+      "name": "libpgs",
+      "version": "0.2.0",
+      "license": "MIT",
+      "devDependencies": {
+        "@tsconfig/recommended": "^1.0.6",
+        "@types/jest": "^29.5.12",
+        "jest": "^29.7.0",
+        "ts-jest": "^29.1.4",
+        "ts-loader": "^9.5.1",
+        "ts-node": "^10.9.2",
+        "typescript": "^5.4.5",
+        "webpack": "^5.91.0",
+        "webpack-cli": "^5.1.4"
       }
     },
     "node_modules/@aashutoshrathi/word-wrap": {
@@ -14970,6 +14987,10 @@
       "dependencies": {
         "comlink": "^4.4.1"
       }
+    },
+    "node_modules/libpgs": {
+      "resolved": "../libpgs-js",
+      "link": true
     },
     "node_modules/lie": {
       "version": "3.1.1",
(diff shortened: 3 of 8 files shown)


Arcus92:
I might need a little help over at [libpgs-js](https://github.com/Arcus92/libpgs-js) on how to get the web-worker output ES compatible.

ferferga:
@Arcus92 May I ask you why webpack over rollup or Vite?

Arcus92:
@ferferga To be honest ... I haven't worked on web projects in a long time and it is the only bundler I kinda know. I'll look into other tools over the weekend.

ferferga:
I know the basics of rollup, but I know that Vite has [this option](https://vitejs.dev/config/worker-options#worker-format), which I use in Vue. I'm not sure how good Vite is for libraries though, and that's why I suggested rollup (since Vite is based on it). Your library is rather simple in terms of building, so I believe it shouldn't be too problematic with either bundler, but you'll likely have much more luck with both instead of webpack.

Arcus92:
As suggested in the backend PR, there is now an opt-in setting to enable experimental Pgs rendering. I'm working on a few performance and load improvements.

src/components/subtitlesettings/subtitlesettings.js
```diff
@@ -111,6 +115,19 @@ function onSubtitleModeChange(e) {
...
+    const view = dom.parentWithClass(e.target, 'subtitlesettings');
+
+    const fieldRenderPgs = view.querySelector('.fldRenderPgs');
+
+    // Pgs option is only available if burn-in mode is set to 'auto' (empty string)
+    if (this.value) {
+        fieldRenderPgs.classList.add('hide');
+    } else {
+        fieldRenderPgs.classList.remove('hide');
+    }
```
thornbill, line 128:
```suggestion
    fieldRenderPgs.classList.toggle('hide', !!this.value);
```

src/components/subtitlesettings/subtitlesettings.js
```diff
@@ -111,6 +115,19 @@ function onSubtitleModeChange(e) {
     view.querySelector('.subtitles' + this.value + 'Help').classList.remove('hide');
 }
 
+function onSubtitleBurnInChange(e) {
+    const view = dom.parentWithClass(e.target, 'subtitlesettings');
+
```
thornbill, line 120:
```suggestion
```

src/components/subtitlesettings/subtitlesettings.template.html
```diff
@@ -32,6 +32,17 @@ <h2 class="sectionTitle">
             </select>
             <div class="fieldDescription">${BurnSubtitlesHelp}</div>
         </div>
+
+        <div class="checkboxContainer checkboxContainer-withDescription fldRenderPgs hide">
+            <label>
+                <input is="emby-checkbox" type="checkbox" id="chkSubtitleRenderPgs" />
+                <span>${RenderPgsSubtitle}</span>
+
```
thornbill, line 40:
```suggestion
```

thornbill (changes requested):

Sab44:
I just found this PR, thanks for putting in the effort for this!
One question, once this is merged will we also be able to adjust vertical subtitle position for PGS subtitles (similar to SRT subtitles)?
If not, what would need to be done to make that possible?
Thanks!

dmitrylyzo:
@Sab44 PGS subtitles are image-based. Changing their position will result in markers/labels being displayed in the wrong place. The same goes for SSA/ASS.

dmitrylyzo:
Regarding Tizen and webOS:
`transferControlToOffscreen` is [supported](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/transferControlToOffscreen#browser_compatibility) since Chrome 69. So `libpgs-js` will only work on Tizen 5.5+ and webOS 6+.

webOS 5 emulator:
```
Uncaught (in promise) TypeError: this.canvas.transferControlToOffscreen is not a function
```

Arcus92:
> Regarding Tizen and webOS: `transferControlToOffscreen` is [supported](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/transferControlToOffscreen#browser_compatibility) since Chrome 69. So `libpgs-js` will only work on Tizen 5.5+ and webOS 6+.

@dmitrylyzo Thanks for testing this on webOS! I could add a fallback in `libpgs-js` and use single thread rendering if web worker transfer is not supported. But I don't think rendering on the main thread on these devices is a good idea. For now I'll exclude pgs format support if `transferControlToOffscreen` isn't available.

dmitrylyzo:
> @dmitrylyzo Thanks for testing this on webOS! I could add a fallback in `libpgs-js` and use single thread rendering if web worker transfer is not supported. But I don't think rendering on the main thread on these devices is a good idea. It could lead to playback stuttering. For now I'll exclude pgs format support if `transferControlToOffscreen` isn't available.

JSO renders in the worker, preparing images to be put on the canvas in the main thread. So the main thread only copies the image buffer from the worker and overlays it onto the canvas.

You probably can do the same: parse PGS in the worker and produce a set of images.

Arcus92:
I should be able to refactor the lib to support both render paths: Rendering in worker by default and rendering in main thread as a fallback. I also try to setup a webOS emulator to verify the changes. Sounds like a fun weekend project :smile:

Arcus92:
I haven't tested it in Tizen yet but it's now working in webOS 5.
grafik

Arcus92:
Hi @Copper-Bot,
It should run on Tizen now but I can't get Tizen Studio and the emulator to work. I someone can test it please let me know.

VobSub sounds like a cool follow up idea. The format is simpler than PGS but consists of two files _(.idx and .sub)_. That makes it tricky to serve from the backend in a single request. Also `ffmpeg` doesn't support extracting VobSub files. You need to export it to a .mkv and then use `mkvextract` adding a new required binary to the mix. 
Maybe the backend could _simply_ convert it to a PGS file. Let us continue that discussion in a separate issue once this is closed.

dmitrylyzo:
> It should run on Tizen now but I can't get Tizen Studio and the emulator to work. If someone can test it please let me know.

Testing webOS is almost enough:
- Tizen 2.3 ~ webOS 1.2
- Tizen 2.4 ~ webOS 2
- Tizen 4 ~ webOS 4
- Tizen 5.5 ~ webOS 5
- Tizen 6 ~ webOS 6

Not an exact match, but still.
https://developer.samsung.com/smarttv/develop/specifications/web-engine-specifications.html
https://webostv.developer.lge.com/develop/specifications/web-api-and-web-engin

I usually only test webOS 1 and webOS 5. Sometimes webOS 3, but it has glitchy emulator.

Tizen 2.x and webOS 1/2 have the oldest web engines we are trying to support.

Arcus92:
Had no luck today. I can't get past the server selection in webOS 1.2. The app and web-site just locks up. Even with a clean `release-10.9.z` instance. I'll try another version next week.

<details>
  <summary><strong>Here is the output of the Web Inspector</strong></summary>
  
</details>

dmitrylyzo:
> Here is the output of the Web Inspector

Oh. I usually change https://github.com/jellyfin/jellyfin-web/blob/8296d5ac105ab37d257d45897078e47f5e2aaa42/webpack.dev.js#L9
to just `source-map`. And, iirc, development build stopped working after some `@tanstack/*-devtools` were updated (arrow functions). Only production build is currently working on such old web engines.
Adding `@tanstack/*-devtools` to Babel doesn't help. 🙁

Arcus92:
I'm sorry, I lied. The previous commit didn't work on webOS. I've reset my VM and forgot to check the PGS setting, so I just saw the transcoded version...
I was able to make it work in webOS 1.2 (WebKit 537.41) and webOS 5 but I had to disable any worker support. It should be supported but the thread never starts and it doesn't give any errors.

I'll spent a few more days trying to make it work more efficient and discovering the quirks of early day web worker implementation.

dmitrylyzo:
> It should be supported but the thread never starts and it doesn't give any errors.

I guess this is because the libpgs worker uses an unsupported syntax (`let`, `const`, arrow function, optional chaining) or API.
Btw, `Promise` in webOS 2 is glitchy and should be polyfilled anyway.

In JSO, legacy worker is ES5 with a bunch of [polyfills](https://github.com/jellyfin/JavascriptSubtitlesOctopus/blob/master/src/polyfill.js). It is generated using Emscripten, which also adds a lot of polyfills.

Iirc, we tried to babel workers, but it didn't work well. I think it will be fine to have 2 versions of workers: legacy and modern.

Arcus92:
I had no luck with the web worker so far. Event the simplest worker *(no const, nor let)* wouldn't load in the webOS 5 emulator. I see the worker js downloaded in the network tab but the code is not executed and no error is logged in the console *(I know `console` isn't available in that Chromium version)*. I thought this may be an security think but HTTPS doesn't work either. I used this example: https://dev.david-schulte.de/worker/ 

The only solution that works on webOS is to skip the worker and render everything on single thread.

dmitrylyzo:
> I know `console` isn't available in that Chromium version

Have you tried using Chrome 79 (or older)? AFAIK, Chrome 80+ doesn't work.

Arcus92:
> Have you tried using Chrome 79 (or older)? AFAIK, Chrome 80+ doesn't work.

@dmitrylyzo WebOS 5 is using Chromium 68. Workers _should_ work since version 4.

dmitrylyzo:
> @dmitrylyzo WebOS 5 is using Chromium 68. Workers _should_ work since version 4.

I mean, you need to use Chrome 79 (and older) for debugging webOS 5 because Google removed some API.

Arcus92:
> I mean, you need to use Chrome 79 (and older) for debugging webOS 5 because Google removed some API.

Oh, now I got you. Thanks for the hint. The oldest working build I could find is 75 but debugging doesn't change. However, the web worker is working on Desktop. 68 won't boot up on my desktop.

dmitrylyzo:
@Arcus92 I got it working on webOS 5. We need to polyfill `Array.flatMap`:
```
this.displaySets[a].windowDefinitions.flatMap is not a function
```

Arcus92:
@dmitrylyzo Already fixed this in the [`webos`](https://github.com/Arcus92/libpgs-js/tree/webos). Haven't merged it yet because the worker is not ... working.

dmitrylyzo:
> @dmitrylyzo Already fixed this in the [`webos`](https://github.com/Arcus92/libpgs-js/tree/webos) branch. Haven't merged it yet because the worker is not ... working.

I'm testing with libpgs-0.5.0, and I managed to load the worker even on webOS 1.2.
At a minimum, it requires polyfilling of `Promise` and `fetch`. I did't get it to parse PGS (_it's too late currently :sleeping_bed:_).

FIY, you can get `console` in the worker on webOS 1.2:
- add to worker:
https://github.com/jellyfin/JavascriptSubtitlesOctopus/blob/ec9f7e91a9b082dae18380898b2cd10413cc7e20/src/pre-worker.js#L1-L33
https://github.com/jellyfin/JavascriptSubtitlesOctopus/blob/ec9f7e91a9b082dae18380898b2cd10413cc7e20/src/post-worker.js#L622-L625
- add to main:
https://github.com/jellyfin/JavascriptSubtitlesOctopus/blob/ec9f7e91a9b082dae18380898b2cd10413cc7e20/src/subtitles-octopus.js#L623-L642

dmitrylyzo:
@Arcus92 My test [branch](https://github.com/dmitrylyzo/libpgs-js/tree/legacy) with your `webos` branch:
- No-offscreen worker starts on webOS 1.2, but fails on [`ImageData is undefined`](https://github.com/dmitrylyzo/libpgs-js/blob/d1c43be1768de4b9719e9e4ac689fdaedb78ce19/src/pgs.ts#L252) (no `ImageData` in the worker scope).
- Render in the main thread works on webOS 1.2.

Arcus92:
@dmitrylyzo Sorry I was busy for a few days. Thanks for your contribution. I'll have a look at your commits right now.

Arcus92:
I finally fixed the issues by adding a browser detection in `libpgs`. WebOS with a Chrome version below 68 will now use the main-thread renderer. Firefox and Chrome version without ImageData-in-worker support will also fallback to the main-thread.

Normally I wouldn't check feature support via the *userAgent*, but in this case some of these legacy browser *(like WebOS 5.0)* have a broken worker implementation. I may need to add exceptions for other platforms if future issues are reported.

dmitrylyzo (approved):
PR works on:
- webOS 1.2 (main thread)
- webOS 5 (main thread)
- Chrome 128 (worker)
- Firefox 130 (worker)

I think we can use worker even on webOS 1.2, but it can be implemented separately.

src/components/subtitlesettings/subtitlesettings.template.html
```diff
@@ -32,6 +32,16 @@ <h2 class="sectionTitle">
...
+        <div class="checkboxContainer checkboxContainer-withDescription fldRenderPgs hide">
+            <label>
+                <input is="emby-checkbox" type="checkbox" id="chkSubtitleRenderPgs" />
+                <span>${RenderPgsSubtitle}</span>
+            </label>
+            <div class="fieldDescription checkboxFieldDescription">${RenderPgsSubtitleHelp}</div>
+        </div>
+
+
     </div>
```
dmitrylyzo, line 45:
```suggestion
        </div>
    </div>
```

src/components/subtitlesettings/subtitlesettings.js
```diff
@@ -61,6 +61,9 @@ function loadForm(context, user, userSettings, appearanceSettings, apiClient) {
         context.querySelector('#sliderVerticalPosition').value = appearanceSettings.verticalPosition;
 
         context.querySelector('#selectSubtitleBurnIn').value = appSettings.get('subtitleburnin') || '';
+        context.querySelector('#chkSubtitleRenderPgs').checked = appSettings.get('subtitlerenderpgs') === 'true' || false;
```
dmitrylyzo, line 64: Oh, just noticed - fallback is unnecessary here.
```suggestion
        context.querySelector('#chkSubtitleRenderPgs').checked = appSettings.get('subtitlerenderpgs') === 'true';
```

thornbill (approved):

Issue #5647, opened 2024-06-02
Autoplay play now hide buttons are not clickable because of playbar conttrols still active 10.9.3


luky92:
**Describe The Bug**

**Steps To Reproduce**

1. Start playing an episode of any series make sure autoplay next episode is on 
2. wait for the next episode in x s popup 
3. try to click play now button
4. button will not work but the peisode might rewind depending on cursor possition ( looks like palybar controls are still active but invisable)

**Expected Behavior**
button skips to next ep 

**Logs**

**Screenshots**

**System (please complete the following information):**
 - Platform: docker linuxserver image
 - Browser: Chrome
 - Jellyfin Version: 10.9.3

**Additional Context**

dmitrylyzo:
https://github.com/jellyfin/jellyfin-web/issues/5650#issuecomment-2145113706

dmitrylyzo:
Should be fixed in the next point release (by #5653).

PR #5708, opened 2024-06-14
Change "Videos" collection name to "Movies"

collinlowing:
I simply changed the Collection Item's title name from 'Videos' to 'Movies' as requested by https://github.com/jellyfin/jellyfin-web/issues/5647. This should not break any translations.

**Changes**

Changes the collectionItemTypes name from 'Videos' to 'Movies'

**Issues**

Fixes https://github.com/jellyfin/jellyfin-web/issues/5647


thornbill:
This is not correct since not all videos are movies… https://github.com/jellyfin/jellyfin-web/pull/5760 seems to be taking a more correct approach.

thornbill:
Closing in favor of #5760

PR #5709, opened 2024-06-14
Fixes .devcontainer to resolve error "workspace is not found" when opening Dev Container

collinlowing:
**Changes**

Adds the workspace folder path mount. Executes command `npm install` and opens bash terminal in dev container after creation for convience.

**Issues**

No issue is reported. When attempting to open the workspace in the dev container, the error "Workspace is not found" appears. Additionally, the postCreateCommands do not execute because `source` and `nvm` are not found. These steps are removed because they are redundant.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -5,15 +5,19 @@
 	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
 	"image": "mcr.microsoft.com/devcontainers/javascript-node:1-20-bullseye",
 
+	// Set workspace path for the dev container. This is the root of the project in the container.
+  	"workspaceFolder": "/workspace",
+	"workspaceMount": "source=${localWorkspaceFolder},target=/workspace,type=bind",
+
 	// Features to add to the dev container. More info: https://containers.dev/features.
 	// "features": {},
 
 	// Use 'forwardPorts' to make a list of ports inside the container available locally.
-	// "forwardPorts": [],
+	"forwardPorts": [ 3000 ],
 
 	// Use 'postCreateCommand' to run commands after the container is created.
 	//https://github.com/microsoft/vscode-dev-containers/issues/559
-	"postCreateCommand": "source $NVM_DIR/nvm.sh && nvm install 20"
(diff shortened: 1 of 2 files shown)


JPVenson:
i cannot currently test this as my gh workspace time is depleted for this month but i will test this next month

JPVenson (changes requested):
postCreateCommand breaks codespaces

collinlowing:
> postCreateCommand breaks codespaces

Does codespaces throw an error message?

ferferga:
I don't understand why this is necessary? By default, Codespaces shell is located at the root of the repository, not at the home folder of the user.

.devcontainer/devcontainer.json
```diff
@@ -5,6 +5,10 @@
 	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
 	"image": "mcr.microsoft.com/devcontainers/javascript-node:1-20-bullseye",
 
+	// Set workspace path for the dev container. This is the root of the project in the container.
+  	"workspaceFolder": "/workspace",
+	"workspaceMount": "source=${localWorkspaceFolder},target=/workspace,type=bind",
+
```
oddstr13, line 11: This shouldn't be needed, what software are you using for launching devcontainers?

.devcontainer/devcontainer.json
```diff
@@ -13,7 +17,7 @@
 
 	// Use 'postCreateCommand' to run commands after the container is created.
 	//https://github.com/microsoft/vscode-dev-containers/issues/559
-	"postCreateCommand": "source $NVM_DIR/nvm.sh && nvm install 20"
+	"postCreateCommand": "npm install && bash"
```
oddstr13, line 20:
```suggestion
	"postCreateCommand": "npm install"
```

`postCreateCommand` is for setting up the environment, not for launching a shell.

oddstr13 (changes requested):

.devcontainer/devcontainer.json
collinlowing: I was using VS Code on Manjaro Linux

thornbill:
Closing based on feedback above

Issue #5468, opened 2024-05-09
Support hotkey chapters like mpv


enter-a-random-username:
Easy way to change chapters, go back or forwards.

```patch
diff --git a/src/scripts/keyboardNavigation.js b/src/scripts/keyboardNavigation.js
index dedea77d7..dedea77d5 100644
--- a/src/scripts/keyboardNavigation.js
+++ b/src/scripts/keyboardNavigation.js
@@ -43,7 +43,7 @@ const KeyNames = {
 /**
  * Keys used for keyboard navigation.
  */
-const NavigationKeys = ['ArrowLeft', 'ArrowRight', 'ArrowUp', 'ArrowDown'];
+const NavigationKeys = ['ArrowLeft', 'ArrowRight', 'ArrowUp', 'ArrowDown', 'PageUp', 'PageDown'];
 
 /**
  * Elements for which navigation should be constrained.
@@ -140,6 +140,13 @@ export function enable() {
                 inputManager.handleCommand('down');
                 break;
 
+            case 'PageUp':
+                inputManager.handleCommand('nextchapter');
+                break;
+            case 'PageDown':
+                inputManager.handleCommand('previouschapter');
+                break;
+
             case 'Back':
                 inputManager.handleCommand('back');
                 break;

```

In general more hotkeys would be nice. Maybe even a mapping option for common controls from the player.

PR #5711, opened 2024-06-16
Support chapters via keyboard navigation

enter-a-random-username:
**Changes**
Use 'page up' and 'page down' based on MPV key mapping to jump forwards and backwards

**Issues**
#5468

### DIFF
--- a/src/scripts/keyboardNavigation.js
+++ b/src/scripts/keyboardNavigation.js
@@ -43,7 +43,7 @@ const KeyNames = {
 /**
  * Keys used for keyboard navigation.
  */
-const NavigationKeys = ['ArrowLeft', 'ArrowRight', 'ArrowUp', 'ArrowDown'];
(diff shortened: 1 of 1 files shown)


src/scripts/keyboardNavigation.js
```diff
@@ -43,7 +43,7 @@ const KeyNames = {
 /**
  * Keys used for keyboard navigation.
  */
-const NavigationKeys = ['ArrowLeft', 'ArrowRight', 'ArrowUp', 'ArrowDown'];
+const NavigationKeys = ['ArrowLeft', 'ArrowRight', 'ArrowUp', 'ArrowDown', 'PageUp', 'PageDown'];
```
dmitrylyzo, line 46: This way they won't work in the non-TV display mode. Are you trying to add them for TV display mode only?

dmitrylyzo:
These key bindings already exist for the video player:
https://github.com/jellyfin/jellyfin-web/blob/e5e68062f5da3ec1cc15fa4c3e0fb6c10eb4efb2/src/controllers/playback/video/index.js#L1337-L1342

PR #5713, opened 2024-06-16
Fix typo when setting up Live TV tuner

bzf:
CleanShot 2024-06-16 at 21 18 18@2x


thornbill:
Translations for languages other than the base language (en-us) should be updated in [our weblate instance](https://translate.jellyfin.org/projects/jellyfin/jellyfin-web/). Thanks!

PR #5717, opened 2024-06-18
Add filter status indicator

grafixeyehero:
**Changes**
- Add Filter Status Indicator
- Fix save query settings

**Issues**
https://github.com/jellyfin/jellyfin-web/issues/40

### DIFF
--- a/src/components/filterdialog/filterIndicator.js
+++ b/src/components/filterdialog/filterIndicator.js
@@ -3,6 +3,7 @@ import './filterIndicator.scss';
 export function getFilterStatus(query) {
     return Boolean(
         query.Filters
+            || query.IsFavorite
             || query.VideoTypes
             || query.SeriesStatus
             || query.Is4K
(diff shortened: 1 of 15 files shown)


thornbill:
Screenshot 2024-08-15 at 03-25-35 Movies

The hover styling is a little off, otherwise this works well!

src/controllers/livetv/livetvchannels.js
```diff
@@ -117,6 +118,9 @@ export default function (view, params, tabContent) {
         loading.show();
         isLoading = true;
         const query = getQuery();
+        const hasFilters = getFilterStatus(query);
+        setFilterStatus(context, hasFilters);
```
thornbill, line 122: Maybe we should change the function to just accept the query directly instead of calling `getFilterStatus` immediately followed by `setFilterStatus` every time.

src/components/filterdialog/filterIndicator.js
```diff
@@ -0,0 +1,49 @@
+import './filterIndicator.scss';
+
+export function getFilterStatus(query) {
```
thornbill, line 3:
```suggestion
function getFilterStatus(query) {
```

thornbill (approved):

PR #5719, opened 2024-06-18
Fix Trickplay thumbnail in older web engines

dmitrylyzo:
**Changes**
Replace image with CSS `background`.

**Issues**
In webOS 1.2, the image is scaled to fit in the `chapterThumpWrapper` and ignores position. Since it doesn't seem to support `resize` nor `object-fit` to prevent scaling, use background instead.

I avoided reusing `chapterThumb` with necessary CSS changes since it is used by chapters.


thornbill (approved):
Works for me in (current) Firefox.

PR #5721, opened 2024-06-18
Use `aspectRatio` `cover` for theme video backdrop

mihawk90:
## Changes
This has bothered me for a long time. Currently, [theme video backdrops](https://jellyfin.org/docs/general/server/media/shows/#theme-videos) use the default aspect ratio of the video player, resulting in black areas around the video, which just looks awkward depending on the viewport.
This change essentially mimics what static backdrops do and uses the `cover` aspect ratio for this purpose.

**Before:**
image

**After:**
image

## Issues
None I'm aware of or could find

edit:
Force-pushed because CI was being a Diva :sweat_smile:

### DIFF
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -629,7 +629,8 @@ function truncatePlayOptions(playOptions) {
         mediaSourceId: playOptions.mediaSourceId,
         audioStreamIndex: playOptions.audioStreamIndex,
         subtitleStreamIndex: playOptions.subtitleStreamIndex,
-        startPositionTicks: playOptions.startPositionTicks
+        startPositionTicks: playOptions.startPositionTicks,
+        aspectRatio: playOptions.aspectRatio
     };
 }
 
@@ -2277,6 +2278,8 @@ class PlaybackManager {
                 playOptions.isFirstItem = false;
             } else {
                 playOptions.isFirstItem = true;
+                // resetting the aspectRatio from a previous theme video playback
+                if (!playOptions.aspectRatio) playOptions.aspectRatio = 'auto';
             }
 
             const apiClient = ServerConnections.getApiClient(item.ServerId);
@@ -2604,6 +2607,7 @@ class PlaybackManager {
                     const streamInfo = createStreamInfo(apiClient, item.MediaType, item, mediaSource, startPosition, player);
 
                     streamInfo.fullscreen = playOptions.fullscreen;
+                    streamInfo.aspectRatio = playOptions.aspectRatio;
 
                     const playerData = getPlayerData(player);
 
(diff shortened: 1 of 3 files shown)


src/components/playback/playbackmanager.js
```diff
@@ -2277,6 +2278,8 @@ class PlaybackManager {
                 playOptions.isFirstItem = false;
             } else {
                 playOptions.isFirstItem = true;
+                // resetting the aspectRatio from a previous theme video playback
+                if (!playOptions.aspectRatio) playOptions.aspectRatio = 'auto';
             }
```
dmitrylyzo, line 2283: This code (`isFirstItem`) is broken (never worked?), imo.

Tested on a playlist. When you click next, `playOptions.isFirstItem` is always `undefined` because `playOptions` goes from `getDefaultPlayOptions`:
https://github.com/jellyfin/jellyfin-web/blob/c9d47f0a0b40f96116db9cfb8274881a0fae0e5c/src/components/playback/playbackmanager.js#L3027
When you click previous, `isFirstItem` is set (`true`) and changes from `true` to `false`. It looks like `playOptions` is only stored in the first item. :confused: 
When you click next again, it is `undefined` again.
When you click previous again, `isFirstItem` is set (`false`) and changes from `false` to `true`.
:spaghetti: :man_facepalming:

We need to fix `isFirstItem` first.

dmitrylyzo (commented):
Well, it should work when we fix `isFirstItem`. :sweat_smile:

src/components/playback/playbackmanager.js
mihawk90: Heh, I tested skipping to the next item, but didn't think of going back to previous :D

thornbill:
Tagging as blocked since it looks like `isFirstItem` seems to need fixed before this could land.

mihawk90:
It seems like something similar to this was merged somewhere along the way.
In 10.11 the theme media is already covering on the first item, however it breaks on subsequent items when there are multiple.

thornbill:
:man_facepalming: oh man I completely forgot that this PR existed... so in 10.11 it is breaking when there are multiple video backdrops?

mihawk90:
> 🤦‍♂️ oh man I completely forgot that this PR existed... so in 10.11 it is breaking when there are multiple video backdrops?

Well depends how you define "break" :P 

Since it didn't work before at all, it's definitely a step better.
The first background that plays is covering, the following are not.

I think that's mainly what was holding this PR up, because I got to the same point. But as @dmitrylyzo pointed out there were issues deeper in the system that would break video playback. I didn't look too deep into it but I was way out of my depth on that one :D 

FWIW I don't see video playback breaking either with the current state.

mihawk90:
Just going through my open PR list and I'm wondering if this is worth keeping open.

As noted above the state that this PR achieved is already in a release, what's missing is maintaining the aspect ratio across multiple items. But that is something that I am not in the weeds enough to figure out, and even if I was it should probably be its own PR.

I'll leave it open for maintainers to decide, but wanted to bring it up.

Issue #5729, opened 2024-06-24
Jellyfin web client referes to cache for /system/info/public API call and gets the wrong server version after upgrade.


Doxterpepper:
**Describe The Bug**
Upgrading the Jellyfin version in docker from 1.8 to 1.9 the jellyfin web application does not recognize the upgraded server and gives `minServerVersion requirement not met. Server version: 10.8.13`. This appears to be due to a cached response in the browser from `/system/info/public`.

**Steps To Reproduce**

1. Have a production docker container running 10.8
2. Upgrade the docker container with `docker service update --image jellyfin/jellyfin:1.9 jellyfin_service`

_Docker container is updated to version 1.9, logs report 1.9, but calls to `/system/info/public` return 1.8.13. Server setup fails due to too old of a server version._

**Expected Behavior**
After updating the server version, the web client should connect to the new server without issue. Both are running the correct version.

**Logs**
Jellyfin server logs:
> [20:06:10] [INF] [1] Main: Jellyfin version: 10.9.6

Jellyfin web client logs:
> minServerVersion requirement not met. Server version: 10.8.13

**Screenshots**
On connecting to the server, a call to the browser cache can be seen.
![Cached call](https://github.com/jellyfin/jellyfin-web/assets/5078395/e3a58272-cb76-4eca-8122-f021a5760696)

On re-sending the request, the request goes to the server instead, bypassing the cache and returning the correct version.
![Non-Cached Call](https://github.com/jellyfin/jellyfin-web/assets/5078395/ac98f07d-5d0d-49d9-80d7-08699aa64cb2)

**System (please complete the following information):**
 - Platform: Linux 
 - Browser: Firefox
 - Jellyfin Version (docker images): jellyfin/jellyfin:1.8.13, jellyfin/jellyfin:1.9

PR #5680, opened 2024-06-07
Fix episode overview markdown render

Chaitanya-Shahare:
**Changes**

- used `markdownIt` to render the markdown in the episode's overview, as expected (and done in detail view of the episode)

**Issues**
Fixes #5679 

**Screenshots**

### Before

image

### After

image

PR #5730, opened 2024-06-24
Add no-cache attribute for fetch requests to /system/info/public to prevent stale server info

Doxterpepper:
**Changes**
- Changing caching policy for fetch calls to /System/Info/Public to avoid retrieving stale cached server info.

**Issues**

#5729

### DIFF
--- a/src/components/ConnectionRequired.tsx
+++ b/src/components/ConnectionRequired.tsx
@@ -83,7 +83,7 @@ const ConnectionRequired: FunctionComponent<ConnectionRequiredProps> = ({
         if (firstConnection.State === ConnectionState.ServerSignIn) {
             // Verify the wizard is complete
             try {
-                const infoResponse = await fetch(`${firstConnection.ApiClient.serverAddress()}/System/Info/Public`);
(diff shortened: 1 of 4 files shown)


dmitrylyzo:
I think this should go to 10.9.
Could you please rebase and re-target the PR? For reference https://github.com/jellyfin/jellyfin-web/pull/5680#issuecomment-2156018605

Doxterpepper:
Re targeted 1.9.z, but maybe it should go in both 1.9.z and master so it covers future releases?

dmitrylyzo:
> Re targeted 1.9.z, but maybe it should go in both 1.9.z and master so it covers future releases?

It will be automatically backported to the `master` on release.

thornbill:
It should be added as a header here also: https://github.com/jellyfin/jellyfin-web/blob/release-10.9.z/src/hooks/useSystemInfo.ts#L27

fatexs:
I think this should be a blocker for any further release at this point. The amount if issues tracing back to this is crazy...

thornbill:
1. We don't block other bug fixes being released because another bug has not been fixed yet.
2. I seriously doubt this will actually fix the majority of the "related" issues.

thornbill (approved):

PR #5744, opened 2024-06-30
Fix download all for Safari

### LINKED ISSUE #5672: Pressing "Download All" on a show/season downloads only the last episode
**Describe The Bug**
When I press the "Download All" button of a show/season, only the last episode is downloaded and without any subtitles.

**Expected Behavior**
I expected the full show to be downloaded

 - Platform: macOS and iPhone
 - Browser: Safari (on Firefox it downloads the entire show)
 - Jellyfin Version: 10.9.6

jchuong:
**Changes**

Added check to use fallback downloader for iOS
Added check for safari to use the delayed download function
Remove old comment about firefox, as a.click() supported since Firefox 75 (2020) [ref](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/click#browser_compatibility)

**Issues**

Fixes #5672


jchuong:
I've done basic test with 

* Windows Firefox 128.0b9
* MacOS Safari 17.4.1
* iPad Safari 

I have *not* tested:
* Other browsers in iOS (Chrome or Firefox for example)
* iPhone
* Shows that are not on same domain see below
https://github.com/jellyfin/jellyfin-web/blob/c40bf002df05c527126a16e5dde13fe09a83ddb7/src/scripts/multiDownload.js#L30-L35

src/scripts/multiDownload.js
```diff
@@ -38,24 +38,26 @@ function download(url) {
...
         return fallback(urls);
     }
 
     let delay = 0;
 
     urls.forEach(function (url) {
         // the download init has to be sequential for firefox if the urls are not on the same domain
-        if (browser.firefox && !sameDomain(url)) {
+        // desktop safari also requires the delay
+        if ((browser.firefox && !sameDomain(url))
```
gnattu, line 59: I'd like to add this delay to all browsers. On browsers that seem to be happy with this will still missing some episodes without the delay.

gnattu (changes requested):
Can you also rebase this to 10.9 branch as a hotfix?

We may rewrite this to be less hacky in 10.10

jchuong:
Closing to rebase on 10.9

PR #5753, opened 2024-07-07
Add 'how many times' is the speed of transcoding in Playback Info

iFraan:
Added a display number to show 'how many times' is the speed of the transcoding. 

**Changes**
Just that. Here's an image:
image


src/components/playerstats/playerstats.js
```diff
@@ -191,6 +191,16 @@ function getDisplayBitrate(bitrate) {
     }
 }
 
+function getDisplayTranscodeFps(session, player) {
+    const mediaSource = playbackManager.currentMediaSource(player) ?? {};
+    const videoStream = (mediaSource.MediaStreams ?? []).find((s) => s.Type === 'Video') ?? {};
```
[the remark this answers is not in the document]
iFraan: Thank you for the review!
I think it is best to check `videoStream.AverageFrameRate` and return the original string instead of just returning `'-'`

Changing `??` to `||`.

src/components/playerstats/playerstats.js
```diff
@@ -191,6 +191,20 @@ function getDisplayBitrate(bitrate) {
...
 
+function getDisplayTranscodeFps(session, player) {
+    const mediaSource = playbackManager.currentMediaSource(player) || {};
+    const videoStream = (mediaSource.MediaStreams || []).find((s) => s.Type === 'Video') || {};
+
+    const originalFramerate = videoStream.AverageFrameRate;
+    const transcodeFramerate = session.TranscodingInfo.Framerate;
+
+    if (!originalFramerate) {
+        return `${session.TranscodingInfo.Framerate} fps`;
```
dmitrylyzo, line 202: You can reuse variable.
```suggestion
        return `${transcodeFramerate} fps`;
```

src/components/playerstats/playerstats.js
```diff
@@ -191,6 +191,20 @@ function getDisplayBitrate(bitrate) {
...
+    }
+
+    return `${(transcodeFramerate / originalFramerate).toFixed(2)}x (${session.TranscodingInfo.Framerate} fps)`;
```
dmitrylyzo, line 205: Same.
```suggestion
    return `${(transcodeFramerate / originalFramerate).toFixed(2)}x (${transcodeFramerate} fps)`;
```

thornbill (approved):

PR #5760, opened 2024-07-08
Improve item list titles for collections

### LINKED ISSUE #5403: If a collection only contains movies, Jellyfin should display it as 'Movies' instead of 'Videos'
Since 99% of all collections include only movies, Jellyfin should display 'Movies' instead of 'Videos' if it's determined that a collection only includes movies.

venkat-karasani:
**Changes**
For the content inside collections, we will display correct title based on the type and media type.

**Issues**
Fixes #5403

### DIFF
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -1467,7 +1467,7 @@ function renderChildren(page, item) {
                 name: globalize.translate('Episodes'),
                 type: 'Episode'
             }, {
-                name: globalize.translate('Videos'),
+                name: globalize.translate('HeaderVideos'),
                 mediaType: 'Video'
             }, {
                 name: globalize.translate('Albums'),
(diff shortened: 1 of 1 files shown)


src/controllers/itemDetails/index.js
```diff
@@ -1458,11 +1458,17 @@ function renderChildren(page, item) {
...
+                name: globalize.translate('Movies'),
+                type: 'Movie'
             }, {
                 name: globalize.translate('Series'),
                 type: 'Series'
+            }, {
+                name: globalize.translate('Episodes'),
+                type: 'Episode'
+            }, {
+                name: globalize.translate('Videos'),
```
thornbill, line 1470: It doesn't look like "Videos" is a valid translation key:

```suggestion
                name: globalize.translate('HeaderVideos'),
```

thornbill (commented):
Looks good! Just one minor suggestion. :rocket:

thornbill (approved):

PR #5765, opened 2024-07-09
Use production year for movies in video player

### LINKED ISSUE #5736: Title in video player displays release year for movies rather than production year
**Describe The Bug**
Since 10.9 (PR https://github.com/jellyfin/jellyfin-web/pull/3621), the year is appended to the video title in the video player. However, this is the release year (metadata editor: `Release date`)  and not the production year (metadata editor: `Year`) . Usually these do not differ, but sometimes they do. In some cases/locales, movies are not released until years after their production year.

I think this would only need to be changed for movies, not for tv episodes.

**Steps To Reproduce**
1. Play movie for which production year and release year metadata are different.
2. OSD video title will show "Movie (`Release year`)" rather than production year.

**Expected Behavior**
In most (all?) contexts, when a year is displayed in parentheses next to a movie title, this is the production year. I would argue this would be expected here too.

The movie listings display the production year as well, so it would be expected for them to correspond.

**Logs**
N/A

**Screenshots**
image
image

 - Platform: Linux
 - Browser: Firefox
 - Jellyfin Version: 10.9.7

venkat-karasani:
**Changes**
we will display production year in title for movies if it is present and else premiere date.

**Issues**
Fixes #5736

### DIFF
--- a/src/controllers/playback/video/index.js
+++ b/src/controllers/playback/video/index.js
@@ -261,7 +261,7 @@ export default function (view) {
         // Display the item with its premiere date if it has one
         let title = itemName;
         if (item.Type == 'Movie' && item.ProductionYear) {
-            title += ` (${item.ProductionYear})`;
+            title += ` (${datetime.toLocaleString(item.ProductionYear, { useGrouping: false })})`;
         } else if (item.PremiereDate) {
             try {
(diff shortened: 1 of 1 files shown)


thornbill:
Does this match the logic we use for cards and the item details page?

venkat-karasani:
> Does this match the logic we use for cards and the item details page?

Both cards and details page seem to be using production Year if present.

card builder:
image

item details page:
image

thornbill (approved):

PR #5774, opened 2024-07-11
Add user setting for whether to display ratings in Video Player and Up Next Dialog Window

flebel9:
**Changes**
Adds user setting under Playback section to control whether ratings should be displayed in the Video Player and Up Next Dialog Window.

Image of setting:
msedge_l3vXxguNka

Player bar with setting disabled and enabled:
msedge_I4pBRBPSsk
msedge_4s7VKNrZ4n

Next up window with setting disabled and enabled:

msedge_FqFuvfhY0C
msedge_2KZLxc9q9b

### DIFF
--- a/src/components/playbackSettings/playbackSettings.template.html
+++ b/src/components/playbackSettings/playbackSettings.template.html
@@ -146,6 +146,14 @@
             </div>
         </div>
 
+        <div class="checkboxContainer checkboxContainer-withDescription">
+            <label>
+                <input type="checkbox" is="emby-checkbox" class="chkDisplayRatingsInPlayer" />
+                <span>${EnableDisplayRatingsInPlayer}</span>
+            </label>
+            <div class="fieldDescription checkboxFieldDescription">${EnableDisplayRatingsInPlayerHelp}</div>
+        </div>
+
         <div class="selectContainer">
             <select is="emby-select" class="selectChromecastVersion" label="${LabelChromecastVersion}"></select>
         </div>
--- a/src/components/playbackSettings/playbackSettings.template.html
+++ b/src/components/playbackSettings/playbackSettings.template.html
@@ -136,6 +136,14 @@
             <div class="fieldDescription checkboxFieldDescription">${EnableNextVideoInfoOverlayHelp}</div>
         </div>
 
+        <div class="checkboxContainer checkboxContainer-withDescription">
+            <label>
+                <input type="checkbox" is="emby-checkbox" class="chkDisplayRatingsInPlayer" />
+                <span>${EnableDisplayRatingsInPlayer}</span>
+            </label>
+            <div class="fieldDescription checkboxFieldDescription">${EnableDisplayRatingsInPlayerHelp}</div>
+        </div>
+
         <div class="checkboxContainer fldExternalPlayer checkboxContainer-withDescription hide">
             <label>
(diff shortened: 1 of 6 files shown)


src/strings/en-gb.json
```diff
@@ -207,6 +207,8 @@
     "EnableBackdropsHelp": "Display the backdrops in the background of some pages while browsing the library.",
     "EnableCinemaMode": "Cinema mode",
     "EnableDisplayMirroring": "Display mirroring",
+    "EnableDisplayRatingsInPlayer": "Display ratings during playback",
```
dmitrylyzo, line 210: Don't manually edit languages other than `en-us` - this may cause conflicts with Weblate. Use Weblate for translation. Also, I don't think we need to translate GB if it is the same in US.

src/controllers/playback/video/index.js
```diff
@@ -95,18 +95,20 @@ export default function (view) {
...
-            endsAt: false,
-            year: false,
-            programIndicator: false,
-            runtime: false,
-            subtitles: false,
-            originalAirDate: false,
-            episodeTitle: false
-        });
+        if (userSettings.enableDisplayRatingsInPlayer()) {
+            ratingsText.innerHTML = mediaInfo.getPrimaryMediaInfoHtml(displayItem, {
```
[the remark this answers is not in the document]
dmitrylyzo: Currently, this won't cause a problem because we can't change `enableDisplayRatingsInPlayer` during playback. In general we should fill everything for the current item, as if the settings can change on the fly, so that we don't have problems in the future.

src/scripts/settings/userSettings.js
```diff
@@ -184,6 +184,19 @@ export class UserSettings {
...
+     * Get or set 'Enable Display Ratings in Player' state.
+     * @param {boolean|undefined} val - Flag to enable 'Enable Display Ratings in Player' or undefined.
+     * @return {boolean} 'Enable Display Ratings in Player' state.
+     */
+    enableDisplayRatingsInPlayer(val) {
+        if (val !== undefined) {
+            return this.set('enableDisplayRatingsInPlayer', val.toString());
+        }
+
+        return toBoolean(this.get('enableDisplayRatingsInPlayer', false));
```
dmitrylyzo, line 197: Shouldn't we enable it by default to keep the current behavior?

src/components/playbackSettings/playbackSettings.template.html
```diff
@@ -146,6 +146,14 @@ <h2 class="sectionTitle">
             </div>
         </div>
 
+        <div class="checkboxContainer checkboxContainer-withDescription">
```
dmitrylyzo, line 149: I would place it before `fldExternalPlayer`. When the latter is displayed, the order is odd: NextVideoOverlay, ExternalPlayer, DisplayRatings (_for next video overlay_).

nielsvanvelzen:
My 2 cents on this; I'm not a fan of yet another setting for an extremely small detail. I do think those ratings are quite useless during playback so personally I'd remove them without a preference.

dmitrylyzo:
Ratings have been added in #4491 to close the feature request.

We don't visit item details to see rating for the next item, so it looks more useful in the NextUp dialog compared to the main OSD.

Since we don't have a clear opinion, I would make it customisable. But I'm not good at UI design. 😅

thornbill:
I think my vote would be for no options. Remove it from the OSD and keep it in the next up dialog.

PR #5778, opened 2024-07-13
Include The Name Of The TV Show while displaying Episodes In Playlists.

### LINKED ISSUE #5750: TV Show Episodes In Playlists Don't Include The Name Of The Show
I have a playlist of tv shows I might like to watch next or soon (say 20 or 30).  The playlist includes S01E01 of the desired shows.  When I display the contents of the playlist, the text describing each item only includes the S01E01 and the name of the episode.  This is not terribly useful information if I'm browsing the playlist -- The episode names mean nothing to me since I have not watched the show yet (and the episode name might be something generic like "Pilot" anyway).  Would be so much more useful if the TV Show name and S##E## was shown and then, maybe, the episode name if there is room.  This is how it appears to work in the AndroidTV client and, given that, I'm not sure if I'm describing a bug here or if I'm suggesting an enhancement.  For sure, the existing behaviour is not optimal for TV shows.

**Steps To Reproduce**
1) Add a TV Show episode to a playlist using the web client
2) See that the TV Show is not displayed in the contents of the updated playlist.

**Expected Behavior**
1) The playlist displays something that is meaningful to the user to identify the playlist item (i.e. the name of the TV Show)

 - Platform: Windows
 - Browser: Firefox, Chrome, Edge (probably others)
 - Jellyfin Version: 10.9.7

venkat-karasani:
**Changes**
While getting the name for the episode to display in playlist, adding the series name if it is getting displayed inside a playlist
**Note:** Also did the change where the similar naming convention is followed for the episodes when they are displayed in a collection. I will revert this if it is not needed.

**Issues**
Fixes #5750


thornbill:
Could you post a screenshot of what changed (current appearance vs with this change applied)? Thanks!

venkat-karasani:
> Could you post a screenshot of what changed (current appearance vs with this change applied)? Thanks!

Here you go,

Before:
before

After:
after

From the images you can see that the before image does not have the series name and it is not quite useful but the after image has the series name at the beginning of the title and is useful.

thornbill:
It looks like this is already supported if we pass `showParentTitle: true` here: https://github.com/jellyfin/jellyfin-web/blob/master/src/scripts/playlistViewer.js#L22

venkat-karasani:
> It looks like this is already supported if we pass `showParentTitle: true` here: https://github.com/jellyfin/jellyfin-web/blob/master/src/scripts/playlistViewer.js#L22

Yes, it seems to be working. This is how it looks with these changes:
image

thornbill (approved):

MovieWiz:
Should I be seeing this fix in version 10.9.11 which was released about a week after this change?

thornbill:
No

image

PR #5782, opened 2024-07-14
Adding husky, lint-staged to run linting on staged files pre commit

venkat-karasani:
**Changes**
Adding **husky**, **lint-staged** to enforce running eslint and stylelint on the staged files when the dev tries to commit a change on his local. Pass **--no-verify** to skip running these checks on your commit if needed.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,1 @@
+npx lint-staged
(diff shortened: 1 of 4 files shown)


thornbill:
I don't personally have strong feelings either way, but the team discussed adding linting on commit previously and decided against adopting it as it can interfere with developer workflows.

PR #5800, opened 2024-07-16
Add trickplay key frame only extraction option

gnattu:
**Changes**

- Adds key frame only extraction option to trickplay
- Mentions VideoToolbox for hardware mjpeg encoder per https://github.com/jellyfin/jellyfin/pull/11510

**Issues**

Depends on https://github.com/jellyfin/jellyfin/pull/11511


src/apps/dashboard/routes/playback/trickplay.tsx
```diff
@@ -33,6 +33,7 @@ const PlaybackTrickplay: FC = () => {
 
         (page.querySelector('.chkEnableHwAcceleration') as HTMLInputElement).checked = options?.EnableHwAcceleration || false;
         (page.querySelector('.chkEnableHwEncoding') as HTMLInputElement).checked = options?.EnableHwEncoding || false;
+        (page.querySelector('.chkEnableKeyFrameOnlyExtraction') as HTMLInputElement).checked = options?.EnableKeyFrameOnlyExtraction || false;
```
thornbill, line 36: Should we enable this by default if the performance difference is significant?

thornbill (approved):

PR #5817, opened 2024-07-22
Fix undefined serverId in Person card

viown:
**Changes**
Fixes undefined 'serverId' for the person cards under 'Cast & Crew' and possibly in other places as well.

**Issues**

### DIFF
--- a/src/components/cardbuilder/cardBuilder.js
+++ b/src/components/cardbuilder/cardBuilder.js
@@ -1139,7 +1139,9 @@ function getHoverMenuHtml(item, action) {
     let html = '';
 
     html += '<div class="cardOverlayContainer itemAction" data-action="' + action + '">';
-    const url = appRouter.getRouteUrl(item);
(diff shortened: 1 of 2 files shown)


thornbill:
We should avoid using browser globals… you might be able to use a function in `ServerConnections` instead.

src/components/cardbuilder/cardBuilder.js
```diff
@@ -1139,7 +1139,9 @@ function getHoverMenuHtml(item, action) {
     let html = '';
 
     html += '<div class="cardOverlayContainer itemAction" data-action="' + action + '">';
-    const url = appRouter.getRouteUrl(item);
+    const url = appRouter.getRouteUrl(item, {
+        serverId: item.ServerId ?? ServerConnections.currentApiClient().serverId()
```
thornbill, line 1143:
```suggestion
        serverId: item.ServerId || ServerConnections.currentApiClient().serverId()
```

thornbill (approved):

PR #5818, opened 2024-07-22
Update issue templates

spirillen:
## Changes
Updated the issue templates to support the new yaml format.

This is done to help submitter to type in tedious information such as OS, Browser, Versions as these as easily can be picked from a dropdown optional list.

All fields should have unique identifies following the simple camelCase syntax to make it easier to extract the information via the GH API.

I've kept the old issue styled `.md` files for reference, and they should be deleted before any merge actually happens.

Please enjoy my small giveback to an awesome project.

## Issues
Replacing the tedious of typing the standard values of OS, Browser, Versions

## Important
Before actual merging the old issue templates in markdown format need to be deleted or renamed to eg. `.old`

These have solely been kept as reference of contents in the new yaml styled files

This is done to help submitter to type in tedious information such as OS, Browser, Versions as these as easily can be picked from a dropdown optional list.                                                                          
                                                                                                                                                                                                                                     
All fields should have uniq identifies following the simple camellCase syntax to make it easier to extract the information via the GH API.                                                                                           
                                                                                                                                                                                                                                     
I've kept the old issue styled `.md` files for reference, and they should be deleted before any merge actually happens.                                                                                                              
                                                                                                                                                                                                                                     
Please enjoy my small giveback to an awsome project.

----------

## TODO before merge

- [ ] Remove old markdown issue templates
  - [ ] Delete the META template?
- [ ] Adapt current versions, at time of merge


thornbill:
Thanks for the contribution!

I haven't reviewed this in depth yet, but we recently updated the server issue template so we should probably make sure this is generally similar. Refs: https://github.com/jellyfin/jellyfin/pull/11135

* I would also remove versions that do not exist currently. (We will probably automate updating the versions in the template at some point.)
* I think we can just remove the meta issue template... I think it pretty much only gets (ab)used to not follow the normal issue format.

spirillen:
Hi @thornbill 

I have added a todo list to have these comments in mind before merging. This also means I did not change anything yet, as you mentioned your self, your comments are at first look.

.github/ISSUE_TEMPLATE/1-bug-report.yml
```diff
@@ -0,0 +1,128 @@
...
+        - Navigator
+        - Other (Write which above)
+
+  - type: dropdown
+    id: version
+    validations:
+      required: true
+    attributes:
+      label: Jellyfin Version
+      description: Please tell us which version of Jellyfin you are using when this happen
```
oddstr13, line 97:
```suggestion
      description: Please tell us which version of Jellyfin you are using when the issue happens
```

.github/ISSUE_TEMPLATE/1-bug-report.yml
```diff
@@ -0,0 +1,128 @@
+name: Bug Report
+description: 'You have noticed a general issue or regression, and would like to report it'
```
oddstr13, line 2:
```suggestion
description: 'You have noticed a general issue or regression, that you would like to report'
```

.github/ISSUE_TEMPLATE/1-bug-report.yml
```diff
@@ -0,0 +1,128 @@
...
+        - 10.10.0
+        - 10.10.1
+        - 10.10.2
+        - 10.10.3
+        - 10.10.4
+        - 10.10.5
+        - 10.10.6
+        - 10.10.7
+        - 10.10.8
+        - 10.10.9
```
oddstr13, line 119:
```suggestion
```

.github/ISSUE_TEMPLATE/2-playback-issue.yml
```diff
@@ -0,0 +1,104 @@
...
+      description: A clear and concise description of what the bug is.
+
+  - type: textarea
+    id: media
+    validations:
+      required: true
+    attributes:
+      label: Media Information
+      description: Please paste any ffprobe or MediaInfo logs.
+      render: bash
```
oddstr13, line 22:
```suggestion
      description: Please include any ffprobe or MediaInfo logs.
```

.github/ISSUE_TEMPLATE/2-playback-issue.yml
```diff
@@ -0,0 +1,104 @@
...
+        - 10.10.0
+        - 10.10.1
+        - 10.10.2
+        - 10.10.3
+        - 10.10.4
+        - 10.10.5
+        - 10.10.6
+        - 10.10.7
+        - 10.10.8
+        - 10.10.9
```
oddstr13, line 95:
```suggestion
        - 10.10.3
```

.github/ISSUE_TEMPLATE/1-bug-report.yml
```diff
@@ -0,0 +1,128 @@
...
+    validations:
+      required: false
+    attributes:
+      label: Logs
+      description: Please paste any log errors.
+      placeholder: |
+        You can find the logs at...
+
+      render: bash
+
```
oddstr13, line 48:
```suggestion
```

.github/ISSUE_TEMPLATE/1-bug-report.yml
```diff
@@ -0,0 +1,128 @@
+name: Bug Report
+description: 'You have noticed a general issue or regression, and would like to report it'
+projects:
+  - jellyfin/jellyfin-web
```
oddstr13, line 4:
```suggestion
```
won't work for users without write access to the project in question, at least according to the docs

.github/ISSUE_TEMPLATE/2-playback-issue.yml
```diff
@@ -0,0 +1,104 @@
...
+      label: Screenshots
+      description: Add screenshots from the Playback Data and Media Info.
+
+  - type: dropdown
+    id: playSystemOS
+    validations:
+      required: true
+    attributes:
+      label: Your OS
+      description: please complete the following information
```
oddstr13, line 38:
```suggestion
      description: Please complete the following information
```

.github/ISSUE_TEMPLATE/1-bug-report.yml
BotBlake: Additionally, I would suggest using an `input` field instead of a `dropdown`. I personally dislike needing to specify "Other" in a separate field. While I understand it's a GitHub limitation, it's not ideal.

I'd pass this to the Jellyfin-Web crew to decide. Template wise this is totaly okay though!

.github/ISSUE_TEMPLATE/2-playback-issue.yml
BotBlake: Afaik, when adding the `render` option, GitHub will no longer allow a File Upload as a submission in that Text Field.
In the end its a matter of choice, if the team rather wants a bash Codeblock, or a bare Textfield where Users Potentialy Upload logfiles to.

In my personal Oppinion, rendering to bash might the better choice.

.github/ISSUE_TEMPLATE/1-bug-report.yml
spirillen: Leaving for now, as this might change in future

.github/ISSUE_TEMPLATE/1-bug-report.yml
spirillen: Can change the render to logs... I've added the rendering to encapsulate any text that might convert into undesirably links

.github/ISSUE_TEMPLATE/1-bug-report.yml
spirillen: > I would suggest using an input field instead of a dropdown

The choice in the dropdown, is to help avoiding against typo's and tedious typing, that just as well "pre typed" in a dropdown. 

The dropdown of versions should of course be maintained by the release script

.github/ISSUE_TEMPLATE/2-playback-issue.yml
spirillen: The third option is to add a `upload log file(s) here`

.github/ISSUE_TEMPLATE/2-playback-issue.yml
spirillen: Same comment as https://github.com/jellyfin/jellyfin-web/pull/5818/files#r1894567432

.github/ISSUE_TEMPLATE/1-bug-report.yml
```diff
@@ -0,0 +1,128 @@
...
+    validations:
+      required: true
+    attributes:
+      label: Browser
+      description: Which browser(s) did you use
+      multiple: true
+      options:
+        - Firefox-ESR
+        - Firefox
+        - Chrome based (Chromium, Edge)
```
BotBlake, line 87: Chromium based.
Chrome is based on Chromium. Edge is based on Chromium.
https://en.wikipedia.org/wiki/Chromium_(web_browser)

Words mixed up.

.github/ISSUE_TEMPLATE/1-bug-report.yml
BotBlake: In my opinion, the lack of a dedicated "other" field will break automations, as it won't be possible to reliably extract the contents of this value.

All custom builds from source will be categorized as "other," and all custom browsers will also be labeled as "other." Furthermore, every release will require an update to the dropdown, even if this becomes automated in the future.

If the dropdown included a writable "other" field, the situation would be different, but unfortunately, this is a technical limitation from GitHub.

While the user can specify the value in another field, automations won’t be able to find it. Additionally, this approach adds extra effort, as the user would need to select "other," scroll back up, and manually enter the version information.

.github/ISSUE_TEMPLATE/1-bug-report.yml
spirillen: Super, will fix the wording.

.github/ISSUE_TEMPLATE/1-bug-report.yml
spirillen: @BotBlake I see your point and can follow your POV as wellm Yet not fully convinced it will be a problem.

Chosing to let the question open for other comments, before choosing to fiddle more with this part

spirillen: 🤷🏻‍♂️ You do realize, this is the point, for which I made this template??

Why not make a option for major `version.x.x`, that opens for code errors?

thornbill:
Superseded by https://github.com/jellyfin/jellyfin-web/pull/6464

PR #5821, opened 2024-07-22
fix: use navigate instead of resolver

ilteoood:
**Changes**
In the routes, use the `Navigate` element instead of redirect resolver

**Issues**
Fixes https://github.com/jellyfin/jellyfin-tizen/issues/279

### DIFF
--- a/src/apps/experimental/routes/routes.tsx
+++ b/src/apps/experimental/routes/routes.tsx
@@ -34,7 +34,7 @@ export const EXPERIMENTAL_APP_ROUTES: RouteObject[] = [
             },
 
             /* Public routes */
-            { index: true, element: <Navigate to='/home.html' /> },
+            { index: true, element: <Navigate replace to='/home.html' /> },
             ...LEGACY_PUBLIC_ROUTES.map(toViewManagerPageRoute)
         ]
     },
--- a/src/apps/experimental/routes/routes.tsx
+++ b/src/apps/experimental/routes/routes.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { RouteObject, redirect } from 'react-router-dom';
+import { Navigate, RouteObject } from 'react-router-dom';
 
 import { REDIRECTS } from 'apps/dashboard/routes/_redirects';
 import ConnectionRequired from 'components/ConnectionRequired';
@@ -34,7 +34,7 @@ export const EXPERIMENTAL_APP_ROUTES: RouteObject[] = [
             },
 
             /* Public routes */
-            { index: true, loader: () => redirect('/home.html') },
+            { index: true, element: <Navigate to='/home.html' /> },
             ...LEGACY_PUBLIC_ROUTES.map(toViewManagerPageRoute)
         ]
     },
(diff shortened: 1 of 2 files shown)


ilteoood:
IMO would be awesome to have it also into the `release-10.9.z` branch.
As of now I can't configure the application on my tv

src/apps/experimental/routes/routes.tsx
```diff
@@ -34,7 +34,7 @@ export const EXPERIMENTAL_APP_ROUTES: RouteObject[] = [
             },
 
             /* Public routes */
-            { index: true, loader: () => redirect('/home.html') },
+            { index: true, element: <Navigate to='/home.html' /> },
```
dmitrylyzo, line 37: Don't we need `replace` here?

src/apps/stable/routes/routes.tsx
```diff
@@ -29,7 +29,7 @@ export const STABLE_APP_ROUTES: RouteObject[] = [
             },
 
             /* Public routes */
-            { index: true, loader: () => redirect('/home.html') },
+            { index: true, element: <Navigate to='/home.html' /> },
```
dmitrylyzo, line 32: Don't we need `replace` here?

dmitrylyzo (commented):
> IMO would be awesome to have it also into the release-10.9.z branch.

Since you used your `master` branch for PR, I suggest creating a new branch from `release-10.9.z`, cherry-pick commit and open a new PR targeted `release-10.9.z`.

It is always recommended to make changes in a separate branch - less pain for making further changes.

src/apps/experimental/routes/routes.tsx
ilteoood: By default the `redirect` function doesn't replace. Do you want me to still replace it?

src/apps/stable/routes/routes.tsx
ilteoood: By default the `redirect` function doesn't replace. Do you want me to still replace it?

src/apps/experimental/routes/routes.tsx
dmitrylyzo: Doesn't `redirect` do this via `302` code?

ilteoood: Nope, and to confirm it I've also tried it in one of my react projects

dmitrylyzo: Hmm, looks like you're right.
Anyway, I would add `replace` to remove redundant history entry (of course, if it will work on Tizen):
| no `replace` | `replace` |
|---|---|
|history1|history2|

ilteoood:
```suggestion
            { index: true, element: <Navigate replace to='/home.html' /> },
```

src/apps/stable/routes/routes.tsx
ilteoood:
```suggestion
            { index: true, element: <Navigate replace to='/home.html' /> },
```

src/apps/experimental/routes/routes.tsx
ilteoood: Done on both the files, and also on the PR for release

dmitrylyzo:
Closed in favor of #5823.

PR #5823, opened 2024-07-23
fix: use navigate instead of resolver

ilteoood:
**Changes**
In the routes, use the `Navigate` element instead of redirect resolver

**Issues**
Fixes https://github.com/jellyfin/jellyfin-tizen/issues/279

### DIFF
--- a/src/apps/experimental/routes/routes.tsx
+++ b/src/apps/experimental/routes/routes.tsx
@@ -36,7 +36,7 @@ export const EXPERIMENTAL_APP_ROUTES: RouteObject[] = [
             },
 
             /* Public routes */
-            { index: true, element: <Navigate to='/home.html' /> },
+            { index: true, element: <Navigate replace to='/home.html' /> },
             ...LEGACY_PUBLIC_ROUTES.map(toViewManagerPageRoute)
         ]
     },
--- a/src/apps/experimental/routes/routes.tsx
+++ b/src/apps/experimental/routes/routes.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { RouteObject, redirect } from 'react-router-dom';
+import { Navigate, RouteObject } from 'react-router-dom';
 
 import { REDIRECTS } from 'apps/dashboard/routes/_redirects';
 import ConnectionRequired from 'components/ConnectionRequired';
@@ -36,7 +36,7 @@ export const EXPERIMENTAL_APP_ROUTES: RouteObject[] = [
             },
 
             /* Public routes */
-            { index: true, loader: () => redirect('/home.html') },
+            { index: true, element: <Navigate to='/home.html' /> },
             ...LEGACY_PUBLIC_ROUTES.map(toViewManagerPageRoute)
         ]
     },
--- a/src/apps/stable/routes/routes.tsx
+++ b/src/apps/stable/routes/routes.tsx
@@ -28,7 +28,7 @@ export const STABLE_APP_ROUTES: RouteObject[] = [
             },
 
             /* Public routes */
-            { index: true, element: <Navigate to='/home.html' /> },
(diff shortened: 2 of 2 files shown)


dmitrylyzo:
FYI, we backport from the release branch to `master`, so other PR may be closed.

Does this fix the Tizen issue? I mean, have you tested this?

ilteoood:
Yes, yesterday before opening the pr. My patched version is the one I'm running on the tv now.

thornbill:
I'm curious what is the underlying issue with using `redirect`? Is it using something internally that we fail to polyfill?

ilteoood:
> I'm curious what is the underlying issue with using `redirect`? Is it using something internally that we fail to polyfill?

It seems to be an odd internal behavior of the router: on tizen, instead of looking for the resolver, it checks just for the `element` property and then throws the error.
Additionally, looking at the [official documentation](https://reactrouter.com/en/main/fetch/redirect) seems that redirect should be used when handling http calls. In other scenarios, the `Navigate` is the way to go

ilteoood:
I don't think so, this is more a library specific implementation rather than issue with the js engine itself

ilteoood:
Here is the piece of code that is triggering it: https://github.com/remix-run/react-router/blob/0f2b167e9b862d5e6b3425438787c8e7b39197f4/packages/react-router/lib/hooks.tsx#L450

ilteoood:
Hi @thornbill, is this ready to he merged or do I need to do something more?

ilteoood:
Ping @thornbill

thornbill:
You really haven't explained why this is an issue for Tizen. The code you link to appears to only print a developer warning and looking at the message

> This means it will render an <Outlet /> with a null value by default resulting in an "empty" page.

it describes the behavior we _want_ to happen. Also if you look at the react-router examples, you can see they have an [authentication example](https://github.com/remix-run/react-router/blob/dev/examples/auth-router-provider/src/App.tsx) that our code is based on (a `loader` returning a `redirect`).

ilteoood:
The issue is explained in the original issue https://github.com/jellyfin/jellyfin-tizen/issues/279.
It is an issue because it doesn't load the login screen.
Additionally, on the exaple you pasted I don't see a redirect but just an empty object instead of an http call. In fact, according to the official doc, it should be used to provide data to the route: https://reactrouter.com/en/main/route/loader, which is not our scenario.

thornbill:
My question is why is this only an issue for Tizen. If the problem was that react router doesn't support what we are doing for with the redirect then it wouldn't work anywhere.

ilteoood:
Honestly speaking I don't know. The Tizen version I'm using is very old and the peculiarity of its engine implementation could have some bugs that can be hardly reproduced elsewhere. This would not be a new thing, since even in this repo there are some patches just for this platform.

ilteoood:
BTW, are we seriously spending weeks discussing on a PR of 2 lines of code (repeated 2 times), that changes stuff following the library's best practice?
Version 10.9.9 has been released today and I'll need to manually patch it again to make it work on my TV.

ilteoood:
Looking it a bit deeper it could be that my tizen version doesn't fully support the header or the response classes to do redirects.
Here is the implementation I checked: https://github.com/remix-run/react-router/blob/55c18bfc8e20417204cce5ecda88fe03520ad257/packages/router/utils.ts#L1615

I'm supposing it because jellyfin already has a patch for the headers constructor for tizen. Unfortunately I'm not 100 sure about it because for my tizen version there isn't any emulator and test each time on the TV is extremely painful.

ilteoood:
Another user has the same issue: https://github.com/jellyfin/jellyfin-tizen/issues/279#issuecomment-2282197307

Can we consider merging this?

thornbill (approved):

PR #5824, opened 2024-07-23
Don't sort media streams for media info

Bond-009:
This way it keeps a logical order i.e. video info first, followed by audio and subtitles

### DIFF
--- a/src/components/itemMediaInfo/itemMediaInfo.js
+++ b/src/components/itemMediaInfo/itemMediaInfo.js
@@ -71,7 +71,7 @@ function getMediaSourceHtml(user, item, version) {
         const size = `${(version.Size / (1024 * 1024)).toFixed(0)} MB`;
         html += `${createAttribute(globalize.translate('MediaInfoSize'), size)}<br/>`;
     }
-    version.MediaStreams.sort(itemHelper.sortTracks);
+
     for (const stream of version.MediaStreams) {
         if (stream.Type === 'Data') {
             continue;


thornbill (changes requested):
You need to remove the unused import of `itemHelper` now.

thornbill:
This only changes the display order in the media info modal, so I doubt it has any impact on strm support.

thornbill:
With this change I am seeing external subtitles listed first... is that intentional @Bond-009?

dmitrylyzo:
> is that intentional

This is exactly what is done for STRM: external streams go first to support unprobed STRM.

https://github.com/jellyfin/jellyfin/pull/7529
> add external streams before adding internal streams

We just need to sort without:
https://github.com/jellyfin/jellyfin-web/blob/12ba71781ef96e900cd63e35186dff98d9e46027/src/components/itemHelper.js#L318-L321
_`sortTracks` is also used by the item details and playback manager._

Bond-009:
> With this change I am seeing external subtitles listed first... is that intentional @Bond-009?

No that wasn't my intention, it's still better than the current behavior, but not ideal.

dmitrylyzo:
> No that wasn't my intention, it's still better than the current behavior, but not ideal.

I think the best we can do is to add another `sortTracks` function without [default and forced sorting](https://github.com/jellyfin/jellyfin-web/blob/12ba71781ef96e900cd63e35186dff98d9e46027/src/components/itemHelper.js#L318-L321).

Issue #5656, opened 2024-06-04
Caching of pages and no automatic refresh


TheHYPO35:
I'm not 100% sure this would be considered a bug, or if it was an intended feature. It also feels like this may be related to this post:

https://github.com/jellyfin/jellyfin-web/issues/5583

When I go into a TV show, if I change the primary image or metadata for the series itself, this will automatically trigger a "refresh" of the page. However, if I change the primary image of a _season_ that appears on that TV show's page, it does not trigger an automatic refresh. [In steps, I select "edit images" on a season, the popup appears and I can change the primary image for a season which refreshes the "edit images" popup. However, when I back out of the "edit images" screen to the TV series, the old image still appears on the series' page.

That's a bit of an issue, but it's not the real "bug" I came to point out. What I noticed has changed is what I used to do to see changes on a page that hasn't automatically refreshed. I used to be able to click on one of the show's seasons, and then click on the show's title at the top of the season page to return to the show's homepage, or click on "home" and click on the series again. Those clicks would load fresh copies of those pages.

However, in the latest versions, it seems that this is now loading a cached version of these pages. In older versions, going "back" or "forward" with browser buttons or alt+arrows would return to cached versions of the page, but clicking on links again would load a fresh copy.

This isn't a massive issue in a browser where the browser "refresh" button is available, but it's having trickle-down effects on things like the ios version, and jellyfin media player where there is no other way to refresh pages. Perhaps the response will be that those applications need to implement a refresh button, but I thought I'd point this out in case it wasn't intended behaviour in the first place. 

So to summarize, there are two possible issues here - 1) that the show page does not automatically refresh when an image on it (such as a primary season image) is changed like it automatically refreshes with other metadata changes; and 2) that the pages are caching and not doing a fresh load when you click on links to them.

**System (please complete the following information):**
The issue seems to occur on all web clients, but I've tested in Chrome, iOS app and JMP
Jellyfin v. 10.9.3

PR #5825, opened 2024-07-24
Replace history syncing with RouterHistory

thornbill:
**Changes**
* Unifies the experimental and stable app routers (similar to the changes in #5802)
* Adds `RouterHistory` as an implementation of the `History` interface backed by the underlying react-router instance. This replaces the history syncing implementation that causes history actions to be invalid and/or duplicated (everything registers as a `REPLACE`).
* Adds redirects for legacy `#!` url format

**What needs testing?**
* General navigation throughout the app
* Opening/closing dialogs
* Redirects for legacy urls starting with `#!`

**Issues**
Related to #5656 (fixes the incorrect and duplicate history events that prevents us from reverting the caching behavior) (will require additional changes to fully fix)


thornbill:
Preview URL: https://44f001e6.jellyfin-web.pages.dev/

nielsvanvelzen:
Did a bunch of navigation, playing items, opening menus/overlays, going to dashboard and back. Both on the unstable demo and my stable server. Overall everything feels fine, found two issues that I could also reproduce in 10.9.8, so most likely not related to this PR.

- When opening a dialog (e.g. "cast to" menu) and then close it, the browser things it can go forward. (also in 10.9.8)
- Opening the "favorite" tab on the homepage does not allow to go back to the "home" tab with a back press. (also in 10.9.8)

grafixeyehero (approved):
LGTM

PR #5826, opened 2024-07-25
Fix overly aggressive view caching

### LINKED ISSUE #5656: Caching of pages and no automatic refresh

thornbill:
**Changes**
* Fixes `ViewManagerPage` always using cached version of views when available. This effectively reverts to the pre-10.9 behavior of only using the cached view on `POP` events (navigating back/forward).

**Depends on #5825**

**Issues**
Fixes #5656

### DIFF
--- a/src/components/viewManager/ViewManagerPage.tsx
+++ b/src/components/viewManager/ViewManagerPage.tsx
@@ -1,5 +1,6 @@
+import { Action } from 'history';
 import { FunctionComponent, useEffect } from 'react';
-import { useLocation } from 'react-router-dom';
+import { useLocation, useNavigationType } from 'react-router-dom';
 
 import globalize from '../../scripts/globalize';
 import type { RestoreViewFailResponse } from '../../types/viewManager';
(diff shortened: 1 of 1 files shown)


joshuaboniface (approved):

thornbill:
It looks like this breaks modals during playback. We probably need to remove navigation type from the  effect dependency array to prevent page re-renders when state changes are pushed.

thornbill:
Should be fixed now. This can be tested using this preview url: https://ded825b2.jellyfin-web.pages.dev/

Issue #5395, opened 2024-04-18
Trickplay and chapter names


enter-a-random-username:
I never understand the Playback buddle vs chapters

**3 states**
- Time only - no chapter names
- Chapter - image + name + time on hover in multiple lines
- Trickplay - Image + time  with huge `h2` time and no chapter name

Changes to with master branch aka 10.9
![Screenshot_2024070710707Aprild0707](https://github.com/jellyfin/jellyfin-web/assets/112093608/687790be-a8b0-42d9-b8b3-5d8a120bd7c2)

Add a `span` for the name align it. Decrease the font size to same as chapter images mode. Should be fine with 320 and ellipsis Chapter name

PR #5832, opened 2024-07-26
Add chapter titles to trickplay popup

enter-a-random-username:
**Changes**
Add chapter name to Trickplay similar to chapter only

image

My CSS, but i think maybe a generic solution wants something else.

```css
.chapterThumbText {
    font-size: 1.2rem;
}

.chapterThumbName {
    position: absolute;
    right: 0;
    top: 0;
    font-size: 1.2rem;
    bottom: 0;
    margin: 0.5rem;
}
```

**Reason for CSS**
 - At the moment the font size for the time is huge
 - Thumb size is configurable (at very small images this could be a problem - i use 480px and HDPI 240px on pixel density >= 2.0 )
 - Maybe some overflow hidden for very long chapter names
 - Maybe a uniform solution with normal chapter and trickplay would be nice

Closes https://github.com/jellyfin/jellyfin-web/issues/5395


gnattu:
> Thumb size is configurable (at very small images this could be a problem - i use 480px and HDPI 240px on pixel density >= 2.0 )

I think we should just hide this if the thumb size is small.

> At the moment the font size for the time is huge

And I think the PR should include a change set the font size for this and this should not be part of the custom css if we are shipping this a part of the standard behavior.

src/controllers/playback/video/index.js
```diff
@@ -1384,11 +1384,13 @@ export default function (view) {
         let chapterThumbContainer = bubble.querySelector('.chapterThumbContainer');
         let chapterThumb;
         let chapterThumbText;
+        let chapterThumbName;
 
         // Create bubble elements if they don't already exist
         if (chapterThumbContainer) {
             chapterThumb = chapterThumbContainer.querySelector('.chapterThumbWrapper');
             chapterThumbText = chapterThumbContainer.querySelector('.chapterThumbText');
+            chapterThumbName = chapterThumbContainer.querySelector('.chapterThumbName');
```
dmitrylyzo, line 1393:
```suggestion
            chapterThumbText = chapterThumbContainer.querySelector('h2.chapterThumbText');
            chapterThumbName = chapterThumbContainer.querySelector('div.chapterThumbText');
```
Here we probably need to introduce `chapterThumbTime` and (_as you did_) `chapterThumbName`.

src/controllers/playback/video/index.js
```diff
@@ -1435,6 +1452,7 @@ export default function (view) {
         chapterThumb.style.backgroundPositionY = offsetY + 'px';
 
         chapterThumbText.textContent = datetime.getDisplayRunningTime(positionTicks);
+        chapterThumbName.textContent = (chapter ? chapter?.Name : '');
```
dmitrylyzo, line 1455:
```suggestion
        chapterThumbName.textContent = chapter?.Name || '';
```

src/controllers/playback/video/index.js
```diff
@@ -1407,11 +1409,26 @@ export default function (view) {
...
+        let chapter;
+        if (item?.Chapters?.length) {
+            for (let i = 0, length = item.Chapters.length; i < length; i++) {
+                const currentChapter = item.Chapters[i];
+
+                if (positionTicks >= currentChapter.StartPositionTicks) {
+                    chapter = currentChapter;
+                }
+            }
+        }
```
dmitrylyzo, line 1430: Not tested.
```suggestion
        for (const currentChapter of item.Chapters || []) {
            if (positionTicks < currentChapter.StartPositionTicks) {
                break;
            }

            chapter = currentChapter;
        }
```
`item.Id` is accessed below, so I think we can skip testing it (`?.`).
I also think we can stop the loop in the `else` block, assuming the chapters are sorted by time. I have rearranged the loop body. We can use `findLast` instead of loop though.

src/controllers/playback/video/index.js
```diff
@@ -1410,6 +1412,20 @@ export default function (view) {
             chapterThumbText = document.createElement('h2');
             chapterThumbText.classList.add('chapterThumbText');
             chapterThumbTextContainer.appendChild(chapterThumbText);
+            chapterThumbName = document.createElement('span');
+            chapterThumbName.classList.add('chapterThumbName');
+            chapterThumbTextContainer.appendChild(chapterThumbName);
```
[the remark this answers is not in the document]
dmitrylyzo: I think this will be redesigned at some point. We just should keep them consistent.
As an option, we could use ellipsis.
Also, we could add a footer below the thumbnail.

dmitrylyzo (approved):

Issue #5841, opened 2024-07-27
NPM Install Never Finishes on Node v22


Drew-Daniels:
**Describe The Bug**
This is _not_ a bug with the application, just an issue I ran into when configuring this application to run locally.

I noticed that when trying to run `npm install` in `jellyfin-web` while using `node v22`, it would never finish. When I downgraded to `v20`, it worked.

**Steps To Reproduce**
1. Navigate to root of `jellyfin-web`
2. `rm -rf node_modules`
3. Switch to `node v22`
4. `npm i`
5. See that `npm install` never finishes
6. Switch to `node v22`
7. `rm -rf node_modules`
8. `npm i`
9. See that `npm install` is able to complete successfully

**Expected Behavior**
I would expect that we could use `node v22` to install all dependencies for this project if the [node engine](https://github.com/jellyfin/jellyfin-web/blob/master/package.json#L159) is listed as `>=20`.

**Logs**

**Screenshots**

**System (please complete the following information):**
 - Platform: MacOS
 - Browser: N/A
 - Jellyfin Version: N/A

**Additional Context**
I'd like to suggest adding an `.nvmrc` file so that any developers who contribute to this repo can automatically have their runtime manager (`nvm`, `mise`, etc.) select to use `v20`

satmandu:
Have you tried using node 22.5.1? Node v22.5 had a problem that broke building jellyfin-web: https://github.com/npm/cli/issues/7657 
I was able to resolve by updating to 22.5.1.

bu3alwa:
v22.5 had a problem with node and it broke a lot of packages https://github.com/nodejs/node/pull/53934/files
The project is compiled on v20 on the github workflow so if you developing v20 is ideal. In addition you should probably use `npm ci` if you want to install it locally.

satmandu:
> v22.5 had a problem with node and it broke a lot of packages https://github.com/nodejs/node/pull/53934/files
> The project is compiled on v20 on the github workflow so if you developing v20 is ideal. In addition you should probably use `npm ci` if you want to install it locally.

Oh yeah, v22.5 was really broken. I've been using v22.5.1 with `npm ci` inside my docker container since it came out and it's been working just fine.

PR #5842, opened 2024-07-27
Add .nvmrc

Drew-Daniels:
**Changes**
Adds an `.nvmrc` file that can be used by `node` runtime managers like `nvm` and `mise` to automatically use a specified version of `node`.

Ensures that all developers are using the same major version of `node`, and would hopefully lessen friction to newer developers getting started with the project.

**Issues**
[jellyfin-web/5841](https://github.com/jellyfin/jellyfin-web/issues/5841)


satmandu:
FYI I've been able to use Node 22.5.1 with jellyfin-web...

thornbill:
I think this is fine to add! It might be a good idea to add a note in the README that we stick to the current Node LTS release.

It would also be nice to update the GitHub actions to reference this to make it easier to update versions. That could be done in a separate PR though.

thornbill (approved):

PR #5843, opened 2024-07-27
Fix swipe gestures on android for book reader

viown:
Fixes swipe gestures not working on android.

**Changes**
Changes the swipe gestures to use the element loaded by epubjs which seems to fix this issue.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-web/issues/5830


thornbill (approved):
Still works on iOS in my testing

PR #5175, opened 2024-01-30
Conditionally show 'Trailers' tab (addresses #4375)

vreelb:
Per the discussion on the linked issues there is no "trailers channel" and there seemed to be consensus that until broader support for trailers was implemented it should be removed to avoid confusing users.

**Changes**
Removes 'Trailers' tab from being visible the 'Movie' UI by default; if at least one item of type 'Trailer' is returned by the server the tab will be re-enabled.

No trailers (default):
image
Trailer detected:
image

**Issues**
Addresses #4375 (and by extension https://github.com/jellyfin/jellyfin/issues/2963).

Issue #4375, opened 2023-02-24
install the trailers channel


tillhanke:
I just stumbled upon the Trailer section and got confused, how to find the "trailer channel". Looking into the jellyfin repo I found a old Issue (https://github.com/jellyfin/jellyfin/issues/2963) in which this is discussed at length. There is also a plugin mentioned:
https://github.com/crobibero/jellyfin-plugin-tmdb-trailers
which will include the trailers into jellyfin.

Now I think it would be better for future users, if this way is not necessary to find the plugin. Couldn't it just be mentioned in the error text:
https://github.com/jellyfin/jellyfin-web/blob/6eb760b03ca438bd7a334e0e930bb777a6671856/src/strings/en-us.json#L1090
Either mention the closed issue, or the plugin directly?

Thanks for the hard work! I really love Jellyfin.

Maltavius:
IMO the Trailer-tab should just list all media that is in a Folder named Trailer in all libraries or ends with -trailer.
Use the Poster and title from the parent item to make it look good.

Maltavius:
> > not everyone is a developer or has the time to spend.
> 
> This also applies to the very few active jellyfin-web contributors. The linked issue was locked because it went offtopic. If someone opens a pull request to modify the message (by either removing it or rephrasing or whatever) I'm confident it will be merged without problem.

Thank you for taking the time to explain why the problem won't be fixed. 

If it was just changing the text in line 1090 of en-us.json I would. But it involves learning how to make a branch, probably downloading the entire source code for Jellyfin, then learning how to make a pull request

"MessageNoTrailersFound": "This tab won't show anything unless you install the trailers plugin found here: https://github.com/crobibero/jellyfin-plugin-tmdb-trailers",

Maltavius:
Just hide the trailer channel until this is fixed.
It's still an issue and has been an issue.

kebogaer:
> Just hide the trailer channel until this is fixed. It's still an issue and has been an issue.

I agree, this is buggy.
Ok admitted when you google you will eventually find the tmdb-trailer plugin, but that;s besides this issue.

This tab is confusing for new users and should either be removed or updated with working funcitonality

thornbill:
PRs are still welcome. Although any direct mentions of the trailers plugin would not be accepted since it may violate terms of service.

vreelb:
[...]

PR #5846, opened 2024-07-28
Only conditionally show 'Trailers' tab.

vreelb:
Recreated from PR https://github.com/jellyfin/jellyfin-web/pull/5175 due to force-pushing rules.

**Changes**
Removes 'Trailers' tab from being visible the 'Movie' UI by default; if at least one item of type 'Trailer' is returned by the server the tab will be re-enabled.

Also some minor refactoring on the page to aid in tab-handling. Previously many magic numbers were used but now it actually uses a mapping to those numbers.

No trailers (default):
image
Trailer detected:
image

**Issues**
Addresses #4375 (and by extension https://github.com/jellyfin/jellyfin/issues/2963).

### DIFF
--- a/src/controllers/movies/moviesrecommended.js
+++ b/src/controllers/movies/moviesrecommended.js
@@ -20,6 +20,15 @@ import 'elements/emby-itemscontainer/emby-itemscontainer';
 import 'elements/emby-tabs/emby-tabs';
 import 'elements/emby-button/emby-button';
 
+const TAB_INDEX_MAPPING = {
+    [LibraryTab.Movies]: 0,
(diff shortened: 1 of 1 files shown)


hthetiot:
Changes look great. Love the TAB_INDEX_MAPPING addition. 
One day you going to get it merge I hope.

soultaco83:
New interface appears to work as intended. I do not have any trailers. Upon install of the PR Trailers is now gone from my Movies tab. 
I am unable to test if the tab comes back when trailers are installed. As i thought there was a trailer plugin but it no longer appears viable to use: https://github.com/crobibero/jellyfin-plugin-tmdb-trailers

PR #5871, opened 2024-08-04
Better codec profile for Safari  with 10.10 features

gnattu:
This uses the new VP9 remuxing and audio remuxing features to reduce transcoding on Safari, also removed some problematic direct play profiles.

**Changes**

- Add opus profile for Safari
- Add VP9 remuxing profile for Safari
- Remove Vorbis profile on non-webm container for Safari
- Remove direct VP9 playback in mp4 container for iOS Safari

**Issues**

Depends on https://github.com/jellyfin/jellyfin/pull/11399
Depends on https://github.com/jellyfin/jellyfin/pull/11489

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -584,12 +584,10 @@ export default function (options) {
         hlsInFmp4VideoAudioCodecs.push('opus');
     }
 
-    if (browser.safari) {
-        if (safariSupportsOpus) {
-            videoAudioCodecs.push('opus');
-            webmAudioCodecs.push('opus');
-            hlsInFmp4VideoAudioCodecs.push('opus');
-        }
+    if (safariSupportsOpus) {
+        videoAudioCodecs.push('opus');
+        webmAudioCodecs.push('opus');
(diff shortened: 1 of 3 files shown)


src/scripts/browserDeviceProfile.js
```diff
@@ -580,9 +581,13 @@ export default function (options) {
         if (browser.tizen) {
             hlsInTsVideoAudioCodecs.push('opus');
         }
-        if (!browser.safari) {
-            hlsInFmp4VideoAudioCodecs.push('opus');
-        }
+        hlsInFmp4VideoAudioCodecs.push('opus');
+    }
+
+    if (safariSupportsOpus) {
```
thornbill, line 587: Does `canPlayAudioFormat` need updates to exclude Safari?

gnattu: Safari does not report opus as supported with the way we are checking there so no need to exclude there, at least for now.

thornbill: Maybe this should be an `else if` in case that behavior changes for Safari in the future?

thornbill (approved):

PR #5872, opened 2024-08-04
Fix remux detection for pure audio session

gnattu:
**Changes**

- Correctly check remuxing for audio only playback session for 10.10 audio remux.

**Issues**

### DIFF
--- a/src/components/playback/playmethodhelper.js
+++ b/src/components/playback/playmethodhelper.js
@@ -2,7 +2,8 @@ export function getDisplayPlayMethod(session) {
     if (!session.NowPlayingItem) {
         return null;
     }
-    if ((!session.TranscodingInfo?.VideoCodec && session.TranscodingInfo.IsAudioDirect) || (session.TranscodingInfo?.IsVideoDirect && session.TranscodingInfo.IsAudioDirect)) {
+
+    if ((session.TranscodingInfo?.IsVideoDirect || !session.TranscodingInfo?.VideoCodec) && session.TranscodingInfo?.IsAudioDirect) {
         return 'Remux';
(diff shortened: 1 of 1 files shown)


src/components/playback/playmethodhelper.js
```diff
@@ -2,8 +2,7 @@ export function getDisplayPlayMethod(session) {
     if (!session.NowPlayingItem) {
         return null;
     }
-
-    if (session.TranscodingInfo?.IsVideoDirect && session.TranscodingInfo.IsAudioDirect) {
+    if ((!session.TranscodingInfo?.VideoCodec && session.TranscodingInfo.IsAudioDirect) || (session.TranscodingInfo?.IsVideoDirect && session.TranscodingInfo.IsAudioDirect)) {
```
dmitrylyzo, line 5: Alternatively (not tested):
```suggestion
    if ((session.TranscodingInfo?.IsVideoDirect || !session.TranscodingInfo?.VideoCodec) && session.TranscodingInfo.IsAudioDirect) {
```

Shadowghost:
Out of interest: Why is this 10.10 specific?

gnattu:
> Out of interest: Why is this 10.10 specific?

10.9 does not support audio remuxing at all.

src/components/playback/playmethodhelper.js
gnattu: When audio remuxing the transcodingInfo will look like this:

image

But I do agree that we probably want to use `session.TranscodingInfo?` just in case.

dmitrylyzo (approved):

thornbill (approved):

PR #5876, opened 2024-08-05
Add Firefox 134+ to list of HEVC-supporting browsers

### LINKED ISSUE #5706: HEVC support for Firefox 120/121+ (Windows)
It looks like Mozilla has started rolling out support for hardware HEVC decoding on Firefox (Windows) from version 120 Nightly/121+.

It's opt-in for now, in nightly HEVC is already enabled by default.
I ran the test on my Jellyfin 10.9.6 instance by modifying already compiled equivalent (main.jellyfin.bundle.js) of the code from src/scripts/browserDeviceProfile.js (lines 618-622):

from:
```
    if (canPlayHevc(videoTestElement, options)
        && (browser.edgeChromium || browser.safari || browser.tizen || browser.web0s || (browser.chrome && (!browser.android || browser.versionMajor >= 105)) || (browser.opera && !browser.mobile))) {
        // Chromium used to support HEVC on Android but not via MSE
        hlsInFmp4VideoCodecs.push('hevc');
    }
```

to:
```
    if (canPlayHevc(videoTestElement, options)
        && (browser.edgeChromium || browser.safari || browser.tizen || browser.web0s || (browser.chrome && (!browser.android || browser.versionMajor >= 105)) || (browser.opera && !browser.mobile) || (browser.firefox && browser.versionMajor >= 121))) {
        // Chromium used to support HEVC on Android but not via MSE
        hlsInFmp4VideoCodecs.push('hevc');
    }
```

Everything seems to be working properly, HEVC decoding works, Jellyfin responds correctly to configuration changes on the Firefox side:
- media.wmf.hevc.enabled - after turning off, Jellyfin falls back to H264 transcoding
- media.wmf.media-engine.enabled - disabling/enabling affects the number of supported HEVC DirectPlay formats, but HEVC transcoding still works properly.

I think it is worth considering adding HEVC support for Firefox (even if it's partial).
This change, of course, would require checking compatibility and whether it would cause problems on Firefox on other operating systems.

fanonwue:
Mozilla implemented support for HEVC decoding using WMF on Windows. Support for HEVC playback is being tested anyway using the canPlayHevc() function, so this should be a safe change for other operating systems (or systems without hardware HEVC decoding) as well.

**Changes**
Firefox (121 and later) supports HEVC decode in fMP4 containers when it's using the new WMF-based media backend, so it should be added to the conditional check for that. The new backend is only available on Windows and not enabled by default (iirc), but when it's present, decoding works perfectly. 
The `canPlayHevc()` function correctly reports decoding capabilities, so a fallback to H264 happens when `media.wmf.hevc.enabled` is disabled, WMF is not present or no hardware decoding is available (I was only able to test this using WSL2 so far, but I doubt that this is going to be different for "real" systems).

**Issues**
Fixes #5706

EDIT: Changed version in title to 132 as there have been bugs regarding 10-bit HEVC in earlier versions.

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -651,7 +651,7 @@ export default function (options) {
     }
 
     if (canPlayHevc(videoTestElement, options)
-        && (browser.edgeChromium || browser.safari || browser.tizen || browser.web0s || (browser.chrome && (!browser.android || browser.versionMajor >= 105)) || (browser.opera && !browser.mobile) || (browser.firefox && browser.versionMajor >= 121))) {
+        && (browser.edgeChromium || browser.safari || browser.tizen || browser.web0s || (browser.chrome && (!browser.android || browser.versionMajor >= 105)) || (browser.opera && !browser.mobile) || (browser.firefox && browser.versionMajor >= 132))) {
         // Chromium used to support HEVC on Android but not via MSE
         hlsInFmp4VideoCodecs.push('hevc');
     }
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -651,7 +651,7 @@ export default function (options) {
     }
 
     if (canPlayHevc(videoTestElement, options)
-        && (browser.edgeChromium || browser.safari || browser.tizen || browser.web0s || (browser.chrome && (!browser.android || browser.versionMajor >= 105)) || (browser.opera && !browser.mobile))) {
+        && (browser.edgeChromium || browser.safari || browser.tizen || browser.web0s || (browser.chrome && (!browser.android || browser.versionMajor >= 105)) || (browser.opera && !browser.mobile) || (browser.firefox && browser.versionMajor >= 121))) {
         // Chromium used to support HEVC on Android but not via MSE
         hlsInFmp4VideoCodecs.push('hevc');
     }
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -651,7 +651,7 @@ export default function (options) {
     }
 
     if (canPlayHevc(videoTestElement, options)
-        && (browser.edgeChromium || browser.safari || browser.tizen || browser.web0s || (browser.chrome && (!browser.android || browser.versionMajor >= 105)) || (browser.opera && !browser.mobile) || (browser.firefox && browser.versionMajor >= 132))) {
+        && (browser.edgeChromium || browser.safari || browser.tizen || browser.web0s || (browser.chrome && (!browser.android || browser.versionMajor >= 105)) || (browser.opera && !browser.mobile) || (browser.firefox && browser.versionMajor >= 134))) {
         // Chromium used to support HEVC on Android but not via MSE
         hlsInFmp4VideoCodecs.push('hevc');
     }


gnattu:
Firefox does not support 10bit HEVC and is lying with the feature tests. The playback will automatically fallback to transcoding with our fail and retry mechanism, but I think it is better to add check here:

https://github.com/jellyfin/jellyfin-web/blob/9fe96669370e4616dec6b4420f0aa7816cec8c8f/src/scripts/browserDeviceProfile.js#L964-L983

Basically we should skip these main10 profile tests for Firefox so that they are not mistakenly marked as supported to avoid unnecessary retry and reduce video load time.

fanonwue:
Nothing can ever be easy, huh?
Sorry about that, I don't have any HDR capable display so I can't really test it I think.

So basically if the browser is Firefox, we should only add `main` and not `main 10` to the profiles, right?

fanonwue:
I have added an override for Firefox's real capabilities at the end of the block of checks that determine HEVC profile compatibility.
https://github.com/jellyfin/jellyfin-web/blob/dac35a1d4b1d7a87a9bb0231742be035fa4d8a33/src/scripts/browserDeviceProfile.js#L983-L987

fanonwue:
Hi,

I understand that you are probably quite busy. But I wanted to ask whether anyone from the JF team thinks we can proceed with adding this? Is there something else I could do?

gnattu:
Please be patient as the web team is super busy. It is normal that a PR is not merged for weeks as lots of mine are in this state as well. I'm good with your current solution but we still need the web team members to look at it.

kam821:
1) It may be a good idea to restrict HEVC to Windows only by adding check:
(browser.firefox && browser.versionMajor >= 121 && browser.windows)
2) Bugs I've noticed so far - on one HEVC-encoded video, a horizontal, green line is displayed at the bottom of the video content.
So far, this problem only affects one video (series). Once, randomly I managed to play that title without that line appearing, but generally this line is almost always displayed.

Green line:
obraz

Encoding parameters of problematic video:
obraz

gnattu:
> It may be a good idea to further restrict HEVC to Windows only by adding appropriate check:

Firefox on other platforms do not report HEVC as supported so this is currently redundant

> Bugs I've noticed so far - on one HEVC-encoded video, a horizontal, green line is displayed at the bottom of the video content.

We can do little if it is a firefox bug

nyanmisaka:
> > We can do little if it is a firefox bug
> 
> This may not necessarily be a problem with Firefox itself. I found some links that describe same/similar problem that has been fixed in the past. [jellyfin/jellyfin#2292](https://github.com/jellyfin/jellyfin/issues/2292) [jellyfin/jellyfin#3021](https://github.com/jellyfin/jellyfin/issues/3021) [jellyfin/jellyfin#5535](https://github.com/jellyfin/jellyfin/issues/5535)

One is an AMD driver issue and the other two are a QSV filter issue I handled in the past. They are not related to this one because there is no hardware transcoding involved. If Chrome doesn't have this problem on the same file, then you need to wait for Firefox to fix their frame cropping.

GodTamIt:
> Firefox does not support 10bit HEVC and is lying with the feature tests.

@gnattu Which feature tests/under which conditions is Firefox lying? I recently filed [a related bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1915307) and may look into upstreaming a fix for it so might be useful to collect all issues into a fix.

This link might be helpful: https://cconcolato.github.io/media-mime-support

gnattu:
> Which feature tests/under which conditions is Firefox lying? 

Firefox reports HEVC main10 has supported for [`canPlayType()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/canPlayType) method but it actually does not, see code quoted [here](https://github.com/jellyfin/jellyfin-web/pull/5876#issuecomment-2282850106) for the real test.

nyanmisaka:
@fanonwue Just tested Firefox 132 stable with `media.wmf.hevc.enabled == 1` and found that HEVC Main 10 playback has been fixed in this version. Client-side HDR tonemapping is still not supported. But I think this PR can go ahead.

firefox132

fanonwue:
> @fanonwue Just tested Firefox 132 stable with `media.wmf.hevc.enabled == 1` and found that HEVC Main 10 playback has been fixed in this version. Client-side HDR tonemapping is still not supported. But I think this PR can go ahead.
> 
> firefox132
> 

In that case, do you want me to adjust the workaround to only work below 132?

nyanmisaka:
> In that case, do you want me to adjust the workaround to only work below 132?

Since HEVC is experimental in older versions, I don't think it's necessary to add a workaround specifically for those broken versions, as browser updates are usually frequent.

nyanmisaka:
> Sorry, this is my first time contributing to a larger project: Do you want me to revert the last commit that introduced the override then? I think it would be good to update the minimum FF version to 132 then as well, I've set it to 121 in the original commit.

It should be enough to bump the checked version up to 132. 
All that's left is for the user to manually enable the experimental flag in Firefox.

fanonwue:
> It should be enough to bump the checked version up to 132.

I've removed the previous override (disallowing HEVC main10) and raised the minimum version to 132.

fanonwue:
> We have more problems regarding this because Firefox 132 on Windows is lying again where it will report HEVC as supported even that flag is disabled as long as the media codec extension is installed which is super annoying

Just tested it with Firefox Nightly 134 and this does not happen. My test video is HEVC main10, it's streaming directly when setting `media.wmf.hevc.enabled` to 1. But when you set it back to 0 (and restart the browser), it gets transcoded to H264 again.

src/scripts/browserDeviceProfile.js
```diff
@@ -620,7 +620,7 @@ export default function (options) {
     }
 
     if (canPlayHevc(videoTestElement, options)
-        && (browser.edgeChromium || browser.safari || browser.tizen || browser.web0s || (browser.chrome && (!browser.android || browser.versionMajor >= 105)) || (browser.opera && !browser.mobile))) {
+        && (browser.edgeChromium || browser.safari || browser.tizen || browser.web0s || (browser.chrome && (!browser.android || browser.versionMajor >= 105)) || (browser.opera && !browser.mobile) || (browser.firefox && browser.versionMajor >= 132))) {
```
gnattu, line 623:
```suggestion
        && (browser.edgeChromium || browser.safari || browser.tizen || browser.web0s || (browser.chrome && (!browser.android || browser.versionMajor >= 105)) || (browser.opera && !browser.mobile) || (browser.firefox && browser.versionMajor >= 134))) {
```

gnattu (changes requested):
Bump the version to 134 where the HEVC support is official and do not require a feature flag. We should be ready to go with this PR.

gnattu (approved):

wbsdty331:
Firefox 136 brings [hardware-accelerated HEVC playback](https://www.mozilla.org/en-US/firefox/136.0beta/releasenotes/)  to macOS.

should we need to raise minimum version to 136+?

fanonwue:
I would assume Firefox reports it's playback capabilities correctly, meaning it wouldn't report being compatible with any HEVC stuff on Mac. But I don't use Apple devices so I can't test that.

thornbill (approved):

PR #5884, opened 2024-08-06
add filters by "NoSubtitles"

PabloTorres1975:
As my first contribution I added a new filter option to view only items without subtitles 
(the inverse of HasSubtitles, currently you can only filter items with subtitles present.)

Changes to filterfeatures, filtertemplates, filterdialogs etc.

The string "NoSubtitiles" should be added to the translation files ( in spanish = "Sin Subtítulos" )


thornbill:
What is the use case for this?

thornbill:
This seems like more of an administrative feature and not something that is generally useful when playing media. iirc the Reports plugin can already find media without subtitles.

thornbill:
Closing based on the feedback above.

Issue #1012, opened 2020-04-01
Errors in editing library should produce an error message


jahmad:
**Describe the bug**
Trying to change library settings, only results loading symbol forever, no changes happen.

**System (please complete the following information):**
 - OS: Debian
 - Virtualization: Docker (linuxserver image)
 - Clients: macOS
 - Browser: Chrome 80
 - Jellyfin Version: 10.5.0

**To Reproduce**

1. Go to Dashboard -> Library
2. Pick a library
3. Click on three dots
4. Select manage library
5. Change some setting, Click OK
6. Loading symbol keep spinning, no changes happened on settings.

**Expected behavior**
Setting changes

**Logs**
Browser console game me this error:

```
 Uncaught Error: null name
    at ApiClient.updateVirtualFolderOptions (apiclientcore.js?v=10.5.0:883)
    at HTMLButtonElement.onEditLibrary (medialibraryeditor.js?v=10.5.0:14)
ApiClient.updateVirtualFolderOptions @ apiclientcore.js?v=10.5.0:883
onEditLibrary @ medialibraryeditor.js?v=10.5.0:14
```

JustAMan:
@jahmad I've edited your issue title (to request an error message) and transferred to the repo that I think is responsible for the issue you're seeing.

joshuaboniface:
This is definitely still relevant, as seen in #4945. Any errors should generate a valid error popup to indicate what's (possibly) going wrong rather than silently dying with a spinner.

theevilapplepie:
@jahmad  -- PR #5893 merged and should ensure an error is displayed, Can you confirm and close? ( Will need to be JellyFin >= v10.10 )

PR #5893, opened 2024-08-08
Show error when library ids are invalid

Narfinger:
When the library has some invalid items, the api requests can return an undefined ItemId. In the current version after editing the library it spins endlessly instead of changing or erroring out.
This PR gives at least a helpful error message when this happens. This relates to issues such as this:
https://github.com/jellyfin/jellyfin/issues/12100#issue-2354924002
https://github.com/jellyfin/jellyfin-web/issues/1012

Two things that might need to change: 
- At the moment the error message is hardcoded in english because I do not know how to get the translation involved.
- Perhaps it would be better to catch the undefined ItemId when loading the dialog instead of closing the dialog but I am not sure when this is done.

### DIFF
--- a/src/components/mediaLibraryEditor/mediaLibraryEditor.js
+++ b/src/components/mediaLibraryEditor/mediaLibraryEditor.js
@@ -33,7 +33,7 @@ function onEditLibrary() {
     const dlg = dom.parentWithClass(this, 'dlg-libraryeditor');
     // when the library has moved or symlinked, the ItemId is not correct anymore
     // this can lead to a forever spinning value on edit the library parameters
-    if (currentOptions.library.ItemId == undefined) {
+    if (!currentOptions.library.ItemId) {
         loading.hide();
         dialogHelper.close(dlg);
         alert({
--- a/src/components/mediaLibraryEditor/mediaLibraryEditor.js
+++ b/src/components/mediaLibraryEditor/mediaLibraryEditor.js
@@ -18,6 +18,7 @@ import '../formdialog.scss';
 import '../../elements/emby-toggle/emby-toggle';
 import '../../styles/flexstyles.scss';
 import './style.scss';
+import alert from '../alert';
 import toast from '../toast/toast';
 import confirm from '../confirm/confirm';
 import template from './mediaLibraryEditor.template.html';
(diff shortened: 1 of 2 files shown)


src/components/mediaLibraryEditor/mediaLibraryEditor.js
```diff
@@ -31,6 +31,14 @@ function onEditLibrary() {
     loading.show();
     const dlg = dom.parentWithClass(this, 'dlg-libraryeditor');
     let libraryOptions = libraryoptionseditor.getLibraryOptions(dlg.querySelector('.libraryOptions'));
+    // when the library has moved or symlinked, the ItemId is not correct anymore
+    // this can lead to a forever spinning value on edit the library parameters
+    if (currentOptions.library.ItemId == undefined) {
+        toast('The library setting is in an invalid state, cannot edit. You are most likely suffering from a bug where the path in the database is not the absolute path in the filesystem');
```
dmitrylyzo, line 37: 1. Make this text translatable: move this text to `strings/en-us.json` and use `globalize.translate`.
2. I think we need to use an alert so that the user has enough time to read. Similar to this:
https://github.com/jellyfin/jellyfin-web/blob/1a9f51c9b0d589f81e09c0614dbe360ee8e78307/src/plugins/experimentalWarnings/plugin.js#L32
You will probably have to close this dialog first and then open alert.

src/components/mediaLibraryEditor/mediaLibraryEditor.js
```diff
@@ -31,6 +31,14 @@ function onEditLibrary() {
     loading.show();
     const dlg = dom.parentWithClass(this, 'dlg-libraryeditor');
     let libraryOptions = libraryoptionseditor.getLibraryOptions(dlg.querySelector('.libraryOptions'));
```
dmitrylyzo, line 33: Move this line (`libraryOptions`) after the new block, because otherwise this variable will be set but not used.

Narfinger:
I incorporated the comments.
By the way, I wonder if we can check for this condition already in the overview of the libraries in the admin view. I noticed that it does not show the posters, so there is probably already something going wrong there.

src/components/mediaLibraryEditor/mediaLibraryEditor.js
```diff
@@ -30,6 +31,16 @@ function onEditLibrary() {
     isCreating = true;
     loading.show();
     const dlg = dom.parentWithClass(this, 'dlg-libraryeditor');
+    // when the library has moved or symlinked, the ItemId is not correct anymore
+    // this can lead to a forever spinning value on edit the library parameters
+    if (currentOptions.library.ItemId == undefined) {
```
thornbill, line 36:
```suggestion
    if (!currentOptions.library.ItemId) {
```

src/strings/en-us.json
```diff
@@ -968,6 +968,7 @@
     "LibraryAccessHelp": "Select the libraries to share with this user. Administrators will be able to edit all folders using the metadata manager.",
     "LibraryScanFanoutConcurrency": "Parallel library scan tasks limit",
     "LibraryScanFanoutConcurrencyHelp": "Maximum number of parallel tasks during library scans. Setting this to 0 will choose a limit based on your systems core count. WARNING: Setting this number too high may cause issues with network file systems; if you encounter problems lower this number.",
+    "LibraryInvalidItemIdError": "The library setting is in an invalid state and cannot be edited. You are most likely suffering from a bug where the path in the database is not the absolute path in the filesystem",
```
thornbill, line 971:
```suggestion
    "LibraryInvalidItemIdError": "The library is in an invalid state and cannot be edited. You are possibly encountering a bug: the path in the database is not the correct path on the filesystem.",
```

thornbill (approved):

PR #5901, opened 2024-08-10
Add support for H264 High 10 Profile on Safari

gnattu:
Both macOS and iOS Safari have enabled H.264 Hi10P decoding support with some restrictions. For our use case, we need to force all H.264 Hi10P video files to trigger remuxing, as this profile is only supported in fMP4 within HLS. Attempts to direct play the original video will fail.

To support this functionality, the server API needs to be modified to accept HLS-specific codec profiles.

This also added a user-configurable option to let the user to opt-in this feature, as the precise feature detection is impossible.

**Changes**

**Issues**

Depends on Server change https://github.com/jellyfin/jellyfin/pull/12420

### DIFF
--- a/src/components/playbackSettings/playbackSettings.template.html
+++ b/src/components/playbackSettings/playbackSettings.template.html
@@ -180,7 +180,7 @@
             <div class="fieldDescription checkboxFieldDescription">${EnableTrueHdHelp}</div>
         </div>
 
-        <div class="enableHi10pSection checkboxContainer checkboxContainer-withDescription fldEnableDts hide">
+        <div class="checkboxContainer checkboxContainer-withDescription fldEnableHi10p hide">
              <label>
                  <input type="checkbox" is="emby-checkbox" class="chkEnableHi10p" />
                  <span>${EnableHi10p}</span>
--- a/src/components/playbackSettings/playbackSettings.template.html
+++ b/src/components/playbackSettings/playbackSettings.template.html
@@ -180,6 +180,14 @@
             <div class="fieldDescription checkboxFieldDescription">${EnableTrueHdHelp}</div>
         </div>
 
+        <div class="enableHi10pSection checkboxContainer checkboxContainer-withDescription fldEnableDts hide">
+             <label>
+                 <input type="checkbox" is="emby-checkbox" class="chkEnableHi10p" />
+                 <span>${EnableHi10p}</span>
+             </label>
+             <div class="fieldDescription checkboxFieldDescription">${EnableHi10pHelp}</div>
+        </div>
+
         <div class="selectContainer">
             <select is="emby-select" id="selectPreferredTranscodeVideoCodec" label="${LabelSelectPreferredTranscodeVideoCodec}">
                 <option value="">${Auto}</option>
--- a/src/components/playbackSettings/playbackSettings.js
+++ b/src/components/playbackSettings/playbackSettings.js
@@ -1,5 +1,6 @@
 import appSettings from '../../scripts/settings/appSettings';
 import { appHost } from '../apphost';
+import browser from '../../scripts/browser';
 import focusManager from '../focusManager';
 import qualityoptions from '../qualityOptions';
 import globalize from '../../scripts/globalize';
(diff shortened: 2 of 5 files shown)


nyanmisaka:
Apple hardware has always been able to handle High10P, and on older systems you can just use a fake mimetype to trick Safari into playing it in fMP4 as well.

Is there a changelog that mentions this?

gnattu:
> Apple hardware has always been able to handle High10P, and on older systems you can just use a fake mimetype to trick Safari into playing it in fMP4 as well.

The hi10p hardware decoding support should be present back to at least iOS 12 era so we can assume this will work for all modern iOS devices. What I'm not confident with is the macOS support status. There are still Intel Macs that never has hi10p hardware, and Apple's release note documented that hardware Hi10P was enabled for videotoolbox on macOS >= 14.6 which makes me not quite confident because I don't know if I would:

- break Intel Macs
- break Apple Silicon Macs on old OS

nyanmisaka:
> > Apple hardware has always been able to handle High10P, and on older systems you can just use a fake mimetype to trick Safari into playing it in fMP4 as well.
> 
> The hi10p hardware decoding support should be present back to at least iOS 12 era so we can assume this will work for all modern iOS devices. What I'm not confident with is the macOS support status. There are still Intel Macs that never has hi10p hardware (probably the model with T2 chip could use the decoder in that chip but I'm not sure), and Apple's release note documented that hardware Hi10P was enabled for videotoolbox on macOS >= 14.6 which makes me not quite confident because I don't know if I would:
> 
> * break Intel Macs
> * break Apple Silicon Macs on old OS

I'm not encouraging this on older systems xD, especially when they have no software fallbacks.

gnattu:
Just tested with macOS 12 + Safari 15 on Apple M1, the player does not fail but the video is just all black. I think the Safari version check is necessary here. Not sure about Intel Macs as I don't have one right now.

nyanmisaka:
> Just tested with macOS 12 + Safari 15 on Apple M1, the player does not fail but the video is just all black. I think the Safari version check is necessary here. Not sure about Intel Macs as I don't have one right now.

Good to know. Is HEVC Rext (4:4:4 & 4:2:2) content possible in Safari?

gnattu:
> Is HEVC Rext (4:4:4 & 4:2:2) content possible in Safari?

Yes. Although we have not added support check for that yet. But such videos are still rare.

After playing with it a little bit more I think we probably need to add a switch to the playback options because:

- There is no easy way to check macOS version. Unlike iOS where we can assume Safari version == iOS version, macOS Safari can be updated as a separate component which means old macOS can still have latest Safari. The macOS version in UA is capped at 10.15.7 because too many websites think macOS major version cannot be larger than 10
- There is no easy way to check if current Mac is an Intel Mac. UA and navigator will always report Intel because the web assumes non-Intel Macs are old ppc Macs.
- There is no easy way to check if the platform can really play Hi10P. The capability tests are always passing, but the video could still display all black frames.

nyanmisaka:
> After playing with it a little bit more I think we probably need to add a switch to the playback options because:
> 
> * There is no easy way to check macOS version. Unlike iOS where we can assume Safari version == iOS version, macOS Safari can be updated as a separate component which means old macOS can still have latest Safari. The macOS version in UA is capped at 10.15.7 because too many websites think macOS major version cannot be larger than 10
> * There is no easy way to check if current Mac is an Intel Mac. UA and navigator will always report Intel because the web assumes non-Intel Macs are old ppc Macs.
> * There is no easy way to check if the platform can really play Hi10P. The capability tests are always passing, but the video could still display all black frames.

So let users opt in on Apple devices to avoid such hassle.

gnattu:
> So let users opt in on Apple devices to avoid such hassle.

Added a playback setting defaulted to false. This option will only show if the browser is Safari and the user will need to explicitly enable it.

src/components/playbackSettings/playbackSettings.template.html
```diff
@@ -180,6 +180,14 @@ <h2 class="sectionTitle">
             <div class="fieldDescription checkboxFieldDescription">${EnableTrueHdHelp}</div>
         </div>
 
+        <div class="enableHi10pSection checkboxContainer checkboxContainer-withDescription fldEnableDts hide">
```
dmitrylyzo, line 183: `fldEnableDts` -> `fldEnableHi10p` and use it instead of `enableHi10pSection`.

dmitrylyzo (approved):

PR #5904, opened 2024-08-11
Add empty password save error message

GabrielGavrilov:
Since administrators aren't allowed to unset their passwords, there should be some sort of toast message.

Currently if an administrator were to try to unset their password, they would get a "Invalid username or password. Please try again." message. 

**Changes**
Whenever an administrator will try to unset their own password, they would get a "New password cannot be empty." message. 

**Issues**
Fixes https://github.com/jellyfin/jellyfin-web/issues/5752


src/components/dashboard/users/UserPasswordForm.tsx
```diff
@@ -7,13 +7,15 @@ import loading from '../../loading/loading';
 import toast from '../../toast/toast';
 import ButtonElement from '../../../elements/ButtonElement';
 import InputElement from '../../../elements/InputElement';
+import { UserDto } from '@jellyfin/sdk/lib/generated-client';
```
thornbill, line 10:
```suggestion
import type { UserDto } from '@jellyfin/sdk/lib/generated-client';
```

This should be moved to the top of the list of imports also.

src/components/dashboard/users/UserPasswordForm.tsx
```diff
@@ -28,17 +30,17 @@ const UserPasswordForm: FunctionComponent<IProps> = ({ userId }: IProps) => {
             return;
         }
 
-        const user = await window.ApiClient.getUser(userId);
+        user.current = await window.ApiClient.getUser(userId);
```
thornbill, line 33: What is the advantage to using a ref for this value?

GabrielGavrilov: Sorry I'm still new to React 😅

I was unable to use a state within the onSubmit() function. It always threw an error. So I thought that a ref would be the next best thing.

thornbill: Using `state` _might_ be better, but that would require a larger refactor of this form... so I think this is fine.

PR #5906, opened 2024-08-12
Add option to always burn in subtitles if transcoding is triggered

gnattu:
**Changes**

Subtitles often have synchronization issues after video transcoding. This option allows the client to opt-in to always burn in all subtitles, preventing out-of-sync subtitles at the cost of transcoding performance.

This also fixed the helper text for subtitle burn in as that option itself will trigger transcoding.

**Issues**

Depends on https://github.com/jellyfin/jellyfin/pull/12430


src/strings/en-us.json
```diff
@@ -1623,6 +1625,7 @@
     "ContainerNotSupported": "The container is not supported",
     "SubtitleCodecNotSupported": "The subtitle codec is not supported",
     "VideoCodecNotSupported": "The video codec is not supported",
+    "VideoCodecTagNotSupported": "The video codec tag is not supported",
```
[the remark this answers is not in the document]
gnattu: Well this is related because this reason is only available after the linked PR on server.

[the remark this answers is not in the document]
gnattu: This part is now in #5950

src/components/subtitlesettings/subtitlesettings.template.html
```diff
@@ -34,6 +34,14 @@ <h2 class="sectionTitle">
         </div>
     </div>
 
+    <div class="checkboxContainer checkboxContainer-withDescription">
+        <label>
+            <input is="emby-checkbox" type="checkbox" id="chkAlwaysBurnInSubtitleWhenTranscoding" />
+            <span>${AlwaysBurnInSubtitleWhenTranscoding}</span>
```
dmitrylyzo, line 40: The localisation for `AlwaysBurnInSubtitleWhenTranscoding` and `AlwaysBurnInSubtitleWhenTranscodingHelp` was accidentally removed.

dmitrylyzo (approved):
The changes requested below are not necessary - it works as is.

gnattu:
The logic is now reworked to reflect the transcoding behavior more accurately. Now depends on server PR #12676

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1566,15 +1567,50 @@ export class HtmlVideoPlayer {
...
+            if (enableNativeTrackSupport(this._currentPlayOptions?.mediaSource, track)) {
+                if (streamIndex !== -1) {
+                    this.setCueAppearance();
+                }
+            } else {
+                // null these out to disable the player's native display (handled below)
+                streamIndex = -1;
+                track = null;
+            }
         }
```
dmitrylyzo, line 1614: It would be better to rewrite it in a way that removes the duplication:
```js
let sessionPromise;
if (!isDirectPlay && burnInWhenTranscoding) {
    sessionPromise = apiClient.getSessions({
                deviceId: apiClient.deviceId()
            }).then(function (sessions) {
                ...
            }) ...
} else {
    sessionPromise = Promise.resolve({});
}

sessionPromise.then((s) => {
    if (!s.TranscodingInfo || s.TranscodingInfo?.IsVideoDirect) {
        ...
    } else {
        ...
    }
}
```

dmitrylyzo (commented):
A quick look.

I was also thinking that if we could get the current (sub)stream URL, we could get the "delivery mode" from there. But I'm not sure we can get such information from a native player.

gnattu:
> I was also thinking that if we could get the current (sub)stream URL, we could get the "delivery mode" from there.

Not going to work because there will be no change to that because the transcoding pipeline checking for copy codec happens way after that

dmitrylyzo:
> Not going to work because there will be no change to that because the transcoding pipeline checking for copy codec happens way after that

Doing this in the `master.m3u8` endpoint would be sufficient. What if streams (`main.m3u8`) will have an unambiguous set of parameters (1 video codec, 1 audio codec, subtitle delivery method, etc.)?
We are currently passing all possible options, which leads to ambiguity. Btw, transcoding jobs can be distinguished by these (strict) parameters.

gnattu:
> What if streams (main.m3u8) will have an unambiguous set of parameters (1 video codec, 1 audio codec, subtitle delivery method, etc.)?

The transcoding pipeline needs to perform some sort of fallback due to reasons like encoders being unavailable iirc so we always have ambiguity because the actual codec also depends on the server configuration inside and outside jellyfin.

dmitrylyzo:
> The transcoding pipeline needs to perform some sort of fallback due to reasons like encoders being unavailable iirc so we always have ambiguity because the actual codec also depends on the server configuration inside and outside jellyfin.

If we are not going to support changing the server configuration on the fly (during playback), we can "perform some sort of fallback due to reasons like encoders being unavailable" in the `master.m3u8` endpoint and generate streams with concrete set of parameters.

Moreover, we probably can (should) account for "encoders being unavailable" in the PlaybackInfo endpoint.

gnattu:
> If we are not going to support changing the server configuration on the fly (during playback),

We will, because hardware encoders are less reliable than you think and we may implement safe fallback later to handle ffmpeg crashing due to hardware encoder failures.

dmitrylyzo:
> We will, because hardware encoders are less reliable than you think and we may implement safe fallback later to handle ffmpeg crashing due to hardware encoder failures.

Doesn't this that after a crash mean you will start generating a stream with different parameters on the same URL?
HLS on my TV doesn't like that (changing codecs on the fly).

gnattu:
> Doesn't this mean that after a crash you will start generating a stream with different parameters on the same URL?

It is still to be discussed because we don't have a concrete plan on how this should be implemented, but either way we need to refactor the transcoder part to even make this api change possible.

dmitrylyzo:
In addition to Session promise refactoring:
> https://github.com/jellyfin/jellyfin-web/pull/5906#discussion_r1769259172
> here and below

https://github.com/jellyfin/jellyfin-web/blob/e4c20df5aebd4b5564577274d7b9f9396356f8fd/src/components/playback/playbackmanager.js#L420-L422

dmitrylyzo (changes requested):
I tried changing subtitles during playback (SRT -> ASS (burn) -> SRT). On the second change, it disables subtitles. _Burn when transcoding is disabled._

It requests a session too early because PlaybackManager sets the subtitles right away.
https://github.com/jellyfin/jellyfin-web/blob/23ee5e62a7aa52dc8d0b7f3e6fb65d9ab4416eda/src/components/playback/playbackmanager.js#L1552
`changeStream` (video element/hls.js) doesn't have enough time to update the session on the server.

We could delay changing subtitles in the player and call `setSubtitleStreamIndex` in `onStartedAndNavigatedToOsd` (it's already there).

gnattu:
> Burn when transcoding is disabled.

Then nothing should be requested, why it behaves differently?

gnattu:
I really cannot reproduce this behavior with current implementation. Nothing is requested when you turned that option off so even if you set it right away it should not disable the srt subtitle.

dmitrylyzo:
> I really cannot reproduce this behavior with current implementation. Nothing is requested when you turned that option off so even if you set it right away it should not disable the srt subtitle.

At first I tested with this feature enabled, and the session was requested before the player changed the video.
I then disabled it, but the issue remained. Perhaps it is not related.

dmitrylyzo:
> Then nothing should be requested, why it behaves differently?

Btw, we hide it, so it can be enabled but hidden.

gnattu:
> Btw, we hide it, so it can be enabled but hidden.

No, if you select the subtitle again you can show it. You disabled that option, nothing will be hidden

gnattu:
I can confirm what you found is not related to this change. You can even reproduce the index reverted to -1 behavior on current master

dmitrylyzo:
> No, if you select the subtitle again you can show it. You disabled that option, nothing will be hidden

I mean that `alwaysBurnInSubtitleWhenTranscoding` remains enabled (but hidden) when you change burn-in option.

gnattu:
> I mean that alwaysBurnInSubtitleWhenTranscoding remains enabled (but hidden) when you change burn-in option.

WTF, that is so wrong. We need to always show that.

gnattu:
Wait, it does not hide? the one becomes hidden is the PGS one not the burn in one

dmitrylyzo:
> WTF, that is so wrong. We need to always show that.

Oh, sorry - false alarm. I looked at the PGS one. :sweat_smile:

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1566,16 +1567,53 @@ export class HtmlVideoPlayer {
...
-            // null these out to disable the player's native display (handled below)
-            streamIndex = -1;
-            track = null;
+            sessionPromise = Promise.resolve({});
         }
+
+        const player = this;
+
+        sessionPromise.then((s) => {
+            if (!s.TranscodingInfo || s.TranscodingInfo?.IsVideoDirect) {
```
dmitrylyzo, line 1591: Unnecessary optional chaining.
```suggestion
            if (!s.TranscodingInfo || s.TranscodingInfo.IsVideoDirect) {
```

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1566,16 +1567,53 @@ export class HtmlVideoPlayer {
...
+                return Promise.resolve({});
+            });
         } else {
-            // null these out to disable the player's native display (handled below)
-            streamIndex = -1;
-            track = null;
+            sessionPromise = Promise.resolve({});
         }
+
+        const player = this;
```
dmitrylyzo, line 1588: Technically, arrow function saves the original `this` by itself.
But using a helper variable with a meaningful name is clearer. :man_shrugging:

dmitrylyzo (approved):

PR #5910, opened 2024-08-13
Fix "Download All" for Safari

jchuong:
Changes

Added check to use fallback downloader for iOS
Changed download to always use `setTimeout` instead of conditionally on browser.
Remove old comment about firefox, as `a.click()` supported since Firefox 75 (2020) [ref](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/click#browser_compatibility)

Issues

Fixes https://github.com/jellyfin/jellyfin-web/issues/5672

Related #5744 (reopened against 10.9)

### DIFF
--- a/src/scripts/multiDownload.js
+++ b/src/scripts/multiDownload.js
@@ -27,13 +27,6 @@ function fallback(urls) {
     })();
 }
 
-function sameDomain(url) {
-    const a = document.createElement('a');
-    a.href = url;
-
-    return window.location.hostname === a.hostname && window.location.protocol === a.protocol;
-}
-
 function download(url) {
     const a = document.createElement('a');
     a.download = '';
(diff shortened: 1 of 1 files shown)


src/scripts/multiDownload.js
```diff
@@ -27,39 +27,26 @@ function fallback(urls) {
...
 }
 
 export default function (urls) {
     if (!urls) {
         throw new Error('`urls` required');
     }
 
-    if (typeof document.createElement('a').download === 'undefined') {
+    if (typeof document.createElement('a').download === 'undefined'
+        || browser.iOS) {
```
thornbill, line 43:
```suggestion
    if (typeof document.createElement('a').download === 'undefined' || browser.iOS) {
```

Issue #5874, opened 2024-08-04
Remote controls in dashboard "active devices" always defaults to pause button


nielsvanvelzen:
**Describe The Bug**

The remote controls in the "active devices" section of the dashboard always default to pause button. Even when the media is already paused. Pressing the pause buttons will still trigger correct behavior (playing again) and afterwards the visuals are corrected.

**Steps To Reproduce**

1. In another window/client play media, then pause it
2. Open dashboard
3. See bug
4. Press the "pause" button to actually play again

**Expected Behavior**

It should display the play icon instead of the pause icon.

**Logs**

**Screenshots**

![93200003a7](https://github.com/user-attachments/assets/644e02cf-0e08-4089-a461-8b42b376d0b7)

**System (please complete the following information):**
 - Platform: [e.g. Linux, Windows, iPhone, Tizen]
 - Browser: [e.g. Firefox, Chrome, Safari]
 - Jellyfin Version: [e.g. 10.6.0]

**Additional Context**

Literally unusable

kuligs2:
i wonder why there is no next/prev buttons here? Is there an API endpoint where i could trigger remotely to skip to the next movie if a playlist is playing? Plan is to set up this in homeassistant

PR #5915, opened 2024-08-15
Fix incorrect initial play icon in remote control section

viown:
The play icon buttons in the remote control section would incorrectly display the pause icon when the media was paused until the session gets updated.

**Changes**
Swap out the play arrow and pause icons in the initial state.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-web/issues/5874


dmitrylyzo (approved):
For reference: https://github.com/jellyfin/jellyfin-web/blob/76a28e125ef5f61d458e60f59da24ffa80187f6a/src/controllers/dashboard/dashboard.js#L580

PR #5923, opened 2024-08-17
Fix safari volume being reset when track changed

gnattu:
**Changes**

This is a follow-up of #5920. That PR missed the case that the gain value will be reset during track change. This PR always set the gain value relative to current volume on Safari

**Issues**

### DIFF
--- a/src/plugins/htmlAudioPlayer/plugin.js
+++ b/src/plugins/htmlAudioPlayer/plugin.js
@@ -126,7 +126,6 @@ class HtmlAudioPlayer {
                 if (normalizationGain) {
                     self.normalizationGain = Math.pow(10, normalizationGain / 20);
                     self.gainNode.gain.value = self.normalizationGain;
-
                 } else {
                     self.gainNode.gain.value = 1;
                     self.normalizationGain = 1;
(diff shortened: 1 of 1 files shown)


src/plugins/htmlAudioPlayer/plugin.js
```diff
@@ -124,13 +124,13 @@ class HtmlAudioPlayer {
                 }
 
                 if (normalizationGain) {
-                    self.gainNode.gain.value = Math.pow(10, normalizationGain / 20);
-                    self.normalizationGain = self.gainNode.gain.value;
+                    self.normalizationGain = Math.pow(10, normalizationGain / 20);
+                    self.gainNode.gain.value = browser.safari ? elem.volume * self.normalizationGain : self.normalizationGain;
```
dmitrylyzo, line 128: You can probably extract volume multiplication, placing following after else block:
```js
if (browser.safari) {
    // Gain value is absolute in Safari. Add volume from the slider
    self.gainNode.gain.value *= elem.volume;
}
```
_Comment can be enhanced._

thornbill (approved):

PR #5927, opened 2024-08-18
Use /Items endpoint to populate Music and Tvshows Recently Added

scampower3:
**Changes**

Uses the /Items endpoint to populate Music and Tvshows Recently Added home section instead of /Items/Latest endpoint
**Issues**

Helps resolve https://github.com/jellyfin/jellyfin/issues/10993 and https://github.com/jellyfin/jellyfin/issues/1880


thornbill:
The issue is that the latest endpoint has some different functionality when it comes to grouping. (If there is a single new episode then the episode is returned vs. multiple new episodes will return the show.)

scampower3:
Right. I was following [this](https://github.com/jellyfin/jellyfin/pull/11512#issuecomment-2103074797) when writing this pr.

scampower3:
Seems like ItemSortBy.DateLastContentAdded is not actually sorting by date last episode added.

PR #5931, opened 2024-08-19
Extend Next Up function to Collections

tedhinklater:
**Changes**
Extended renderNextUp to include boxsets, API call differentiates between series and boxsets

**Issues**
None

### DIFF
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -764,16 +764,29 @@ function setPeopleHeader(page, item) {
 function renderNextUp(page, item, user) {
     const section = page.querySelector('.nextUpSection');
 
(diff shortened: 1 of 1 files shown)


thornbill:
This doesn't work? What is the goal with this change?

tedhinklater:
I was hoping to be able to add Next Up to collections, so it shows the Next Up (next unwatched) on the Collection's page like it does with a Series' page. 

I did think it wouldnt be that simple. Shall I just put it in the feature requests?

thornbill:
Please test that changes are working as you expect before opening a pull request (or alternatively mark it as a draft until you have it in a working state).

Sounds like this existing feature request: https://features.jellyfin.org/posts/1606/suggest-next-from-collection-in-up-next

Issue #5909, opened 2024-08-12
Playback Error when using shuffle


putridpete:
Whenever I shuffle media that I've filtered by genre, such as horror for instance, I get a Playback error that says "No player found for the requested media" every time. Happens on any web browser, JMP and on the WebOS app.

**Steps To Reproduce**

1. Go to 'Movies' (TV shows have this issue too)
2. Click on 'Genres'
3. Select any genre
4. Click 'Shuffle'
5. Error occurs

**Expected Behavior**
Media should play shuffled.

**Logs**
Because it refuses to play, I see no relevant logs, but after opening the browser dev console (Firefox) I saw this message that could be related: "No player found for the requested media: undefined main.jellyfin.bundle.js:2:142706".

**System:**
 - Platform: Linux, WebOS
 - Browser: Firefox
 - Jellyfin Version: 10.9.9, 1.11.1 (JMP)

**Additional Context**
Curiously enough, no problem occurs when clicking on "Play all" instead. This issue [was first reported on the Jellyfin forums](https://forum.jellyfin.org/t-playback-error-when-using-shuffle?pid=30753#pid30753), some additional info could be posted here.

dmitrylyzo:
I don't know how it worked before, but now we don't generate a play queue for `Genre` item: https://github.com/jellyfin/jellyfin-web/blob/7c651d0a5dae957843c1d2f061038b17c17fab41/src/components/playback/playbackmanager.js#L1827
I tried adding a similar thing to `MusicGenre`: https://github.com/jellyfin/jellyfin-web/blob/7c651d0a5dae957843c1d2f061038b17c17fab41/src/components/playback/playbackmanager.js#L1857-L1864
But it fetches items across different libraries because `Genre` item has no parent (and the query as well).

WhateverTheyCallMe:
This is still happening to me in browser and iOS apps

Server version
10.10.6
Web version
10.10.6 (c335a3024)
Build version
10.10.6

SylvainBigonneau:
Same on Android (v2.6.2, server and web on v10.10.6)

Itschotsch:
Seems to still be happening on the latest version of Jellyfin on the Firefox web UI and the Android app.

Fabeson:
@thornbill @dmitrylyzo Could you please have a look into this? 

[...]

PR #5949, opened 2024-08-23
Fix play all & shuffle not working on genres

viown:
**Changes**
Fix shuffle/play all not working on movies & TV show genres.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-web/issues/5909

### DIFF
--- a/src/controllers/list.js
+++ b/src/controllers/list.js
@@ -851,7 +851,11 @@ class ItemsView {
             setTitle(null);
             getItem(params).then(function (item) {
                 setTitle(item);
-                item.ParentId = params.parentId;
+                
+                if (item.Type == 'Genre') {
+                    item.ParentId = params.parentId;
+                }
+
                 self.currentItem = item;
                 const refresh = !isRestored;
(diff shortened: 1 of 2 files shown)


src/controllers/list.js
```diff
@@ -851,6 +851,10 @@ class ItemsView {
             setTitle(null);
             getItem(params).then(function (item) {
                 setTitle(item);
+                if (item.Type == 'Genre') {
```
thornbill, line 854: Just for my own understanding, why is this change needed?

viown: Without the ParentId, media from other libraries matching the genre will be queued as well.

thornbill (approved):

WhateverTheyCallMe:
This is still happening to me in browser and iOS apps

Server version
10.10.6
Web version
10.10.6 (c335a3024)
Build version
10.10.6

viown:
@WhateverTheyCallMe Please file an issue with your reproduction steps as it works on my end.

PR #5955, opened 2024-08-25
Support for Home and Search media keys in TV mode

mprasil:
The remote I have has some extra "media" buttons besides the already supported playback controls.

This adds support for "_Find_" button to navigate to search page and "_BrowserHome_" button to navigate back to main screen.

Adding these to the `NavigationKeys` only enables functionality for TV mode which I think is pretty reasonable constraint - on actual desktop people might prefer to use these keys to control the browser rather than Jellyfin interface.

### DIFF
--- a/src/scripts/keyboardNavigation.js
+++ b/src/scripts/keyboardNavigation.js
@@ -43,7 +43,7 @@ const KeyNames = {
 /**
  * Keys used for keyboard navigation.
  */
-const NavigationKeys = ['ArrowLeft', 'ArrowRight', 'ArrowUp', 'ArrowDown'];
+const NavigationKeys = ['ArrowLeft', 'ArrowRight', 'ArrowUp', 'ArrowDown', 'BrowserHome', 'Find'];
 
 /**
  * Keys used for media playback control.
@@ -180,6 +180,13 @@ export function enable() {
                 } else {
                     capture = false;
                 }
+                break;
+
+            case 'Find':
+                inputManager.handleCommand('search');
+                break;
+            case 'BrowserHome':
+                inputManager.handleCommand('home');
                 break;
 
             case 'MediaPlay':


thornbill (commented):
The code changes look good to me. @dmitrylyzo is this something you would like to test?

dmitrylyzo:
> The code changes look good to me. @dmitrylyzo is this something you would like to test?

Yeah, the code looks fine, but I don't know which buttons in the emulator the new codes correspond to.

mprasil:
@dmitrylyzo would this help? https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values

dmitrylyzo:
> @dmitrylyzo would this help? https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values

I mean, I tried all buttons in the webOS 5 emulator and didn't get these codes. Or do you mean you have some sort of universal remote connected to the TV (mini keyboard)?

My main concern is the possibility of unnecessary overriding of default actions. All I can say is that `Home` (🏠, opens LG launcher) on the LG remote generates `Meta` code, so it doesn't seem to be overridden.

Btw, there's `BrowserSearch` code. I suppose your remote generates a more generic code (`Find`), right?

mprasil:
> Or do you mean you have some sort of universal remote connected to the TV (mini keyboard)?

Yes this is exactly my situation. I do think some "multimedia" keyboards also have some of these buttons. 

> My main concern is the possibility of unnecessary overriding of default actions. All I can say is that Home (🏠, opens LG launcher) on the LG remote generates Meta code, so it doesn't seem to be overridden.

Yeah some remotes have Home mapped to Meta as that is what various systems use to go to OS home screen, show start menu, etc.. I would be careful about overriding that one as it would be either captured by OS or it might prevent user from returning to OS home easily. My remote specifically has `BrowserHome` button which is not `Meta`.

> Btw, there's BrowserSearch code. I suppose your remote generates a more generic code (Find), right?

Yes, although I think one of the remotes I had did generate `BrowserSearch` instead. As far as browser/OS is concerned these are two totally different buttons. IIRC `BrowserSearch` is actually `Search` button which some systems intercept on OS level. (For example Gnome desktop will show overview screen with search focused)

> My main concern is the possibility of unnecessary overriding of default actions.

That's why the override only applies to TV interface where both `BrowserHome` and `Find` aren't likely to be used for other purpose. FWIW Emby does intercept these in TV mode to go to home screen or show search interface. At the end of the day I'd be actually fine with any shortcut, worst case I'd just remap these keys to anything, but right now there is no keyboard shortcut for either of these actions, so I submitted PR with keys I'd personally use and that used to work in Emby. I do think using these keys makes sense, but _any_ key would work for me.

mprasil:
@dmitrylyzo if you're using X11 on Linux to run your emulator, I think you can simulate the keys with `xdotool key Find` and `xdotool key XF86HomePage`. I'm not sure if there's similar tool for Wayland or other systems. Also not 100% sure about the key codes for xdotool - got them from [here](https://xkbcommon.org/doc/current/xkbcommon-keysyms_8h.html).

dmitrylyzo (approved):
Using the wireless mini-keyboard, I was able to use `Home` in the browser (in TV mode).
`Find` did not work (_KRunner menu appeared_).
In Tizen and webOS, the buttons don't work. But they are unlikely to conflict.
I think in the future we will probably have a key mapping controller - we will be able to reassign them.

Since it works for you. LGTM

PR #5959, opened 2024-08-25
attempt to fix getChangedEvents function

dredstone1:
Fix: Optimize Player State Updates to Reduce UI Refreshes
This PR optimizes how player state updates are handled to reduce unnecessary UI refreshes. Instead of triggering 'statechange', 'timeupdate', and 'pause' events on every update, the code now only triggers these events when the relevant player state properties change.
This results in a smoother user experience and improved performance.

this is the first version of the fix, so it may contain mistakes


dredstone1 (commented):
add a comment

thornbill:
This seems like it is attempting the same thing as https://github.com/jellyfin/jellyfin-web/pull/5179

dredstone1:
I dont know, but there was an TODO comment there, the function always ended up just updating everything, now it updates depending on the needs

PR #5966, opened 2024-08-26
Fix hidden scheduled task progress updates causing errors

p0358:
**Changes**

An `IConfigurableScheduledTask` can be defined with `IsHidden => true`, which as expected causes it to not be drawn in the UI, nor are its elements created. However, task progress updates would still arrive and cause an error in code that assumed the elements would exist.

**Issues**

N/A, but merging jellyfin/jellyfin#12511 has a chance of exposing this bug if someone opens admin panel on scheduled tasks page quickly enough after server startup (overall a rather minor bug probably as such)

### DIFF
--- a/src/controllers/dashboard/scheduledtasks/scheduledtasks.js
+++ b/src/controllers/dashboard/scheduledtasks/scheduledtasks.js
@@ -132,8 +132,11 @@ function updateTaskButton(elem, state) {
 export default function(view) {
     function updateTasks(tasks) {
         for (const task of tasks) {
-            view.querySelector('#taskProgress' + task.Id).innerHTML = getTaskProgressHtml(task);
-            updateTaskButton(view.querySelector('#btnTask' + task.Id), task.State);
+            const taskProgress = view.querySelector(`#taskProgress${task.Id}`);
+            if (taskProgress) taskProgress.innerHTML = getTaskProgressHtml(task);
+
+            const taskButton = view.querySelector(`#btnTask${task.Id}`);
+            if (taskButton) updateTaskButton(taskButton, task.State);
         }
     }
 


src/controllers/dashboard/scheduledtasks/scheduledtasks.js
```diff
@@ -132,8 +132,10 @@ function updateTaskButton(elem, state) {
 export default function(view) {
     function updateTasks(tasks) {
         for (const task of tasks) {
-            view.querySelector('#taskProgress' + task.Id).innerHTML = getTaskProgressHtml(task);
-            updateTaskButton(view.querySelector('#btnTask' + task.Id), task.State);
+            const taskProgress = view.querySelector('#taskProgress' + task.Id);
+            if (taskProgress) taskProgress.innerHTML = getTaskProgressHtml(task);
+            const btnTask = view.querySelector('#btnTask' + task.Id);
+            if (btnTask) updateTaskButton(btnTask, task.State);
```
thornbill, line 138:
```suggestion
            const taskProgress = view.querySelector(`#taskProgress${task.Id}`);
            if (taskProgress) taskProgress.innerHTML = getTaskProgressHtml(task);

            const taskButton = view.querySelector(`#btnTask${task.Id}`);
            if (taskButton) updateTaskButton(taskButton, task.State);
```

thornbill (changes requested):
Looks good! I just have a couple minor suggestions.

src/controllers/dashboard/scheduledtasks/scheduledtasks.js
p0358: Fair enough, I originally just used the html id for variable name

p0358:
>  Merging can be performed automatically once the requested changes are addressed. 

idk if doing a force-push instead of second commit didn't break github there

thornbill (approved):

PR #5969, opened 2024-08-26
Update HDR and WebM device checks

GodTamIt:
This updates some of the device checks for browser compatibility.

**Changes**

* WebM support is partial/unreliable from Safari versions 17.0-17.4. It was fixed in 17.4:
  * https://developer.apple.com/documentation/safari-release-notes/safari-17_4-release-notes
* Be more precise about Dolby Vision support on Apple devices. Specifically, Safari 13.1 added support:
  * https://developer.apple.com/documentation/safari-release-notes/safari-13_1-release_notes
* Enable HDR support for all browsers on iOS since all browsers are required to use WebKit underneath. HDR support in browsers was added in iOS 13.4 when Safari 13.1 was released.
  * https://developer.apple.com/documentation/safari-release-notes/safari-13_1-release_notes
* Firefox 100+ has support for HDR on macOS/OS X. It requires OS support, which was added in macOS 10.15 Catalina.
  * https://www.mozilla.org/en-US/firefox/100.0/releasenotes/
* Chrome on mobile now has client side tone-mapping. Tested on Chrome Android 128 but first support version is unknown.

### DIFF
--- a/src/scripts/browser.js
+++ b/src/scripts/browser.js
@@ -99,6 +99,19 @@ function iOSversion() {
     return [];
 }
 
+function osxVersion() {
+    const matches = /Mac OS X (\d+)[._](\d+)[._]?(\d+)?/i.test(navigator.userAgent);
+    if (matches) {
+        return [
+            parseInt(matches[1], 10),
+            parseInt(matches[2], 10),
+            parseInt(matches[3] || 0, 10)
+        ];
+    }
+
+    return [];
+}
+
 function web0sVersion(browser) {
     // Detect webOS version by web engine version
 
@@ -228,16 +241,21 @@ const uaMatch = function (ua) {
     version = version || match[2] || '0';
 
     let versionMajor = parseInt(version.split('.')[0], 10);
-
     if (isNaN(versionMajor)) {
         versionMajor = 0;
+    }
+
+    let versionMinor = parseInt(version.split('.')[1], 10);
+    if (isNaN(versionMinor)) {
+        versionMinor = 0;
     }
 
     return {
         browser: browser,
         version: version,
         platform: platform_match[0] || '',
-        versionMajor: versionMajor
+        versionMajor: versionMajor,
+        versionMinor: versionMinor
     };
 };
 
@@ -250,6 +268,7 @@ if (matched.browser) {
     browser[matched.browser] = true;
     browser.version = matched.version;
     browser.versionMajor = matched.versionMajor;
+    browser.versionMinor = matched.versionMinor;
 }
 
 if (matched.platform) {
@@ -269,6 +288,18 @@ browser.osx = userAgent.toLowerCase().indexOf('mac os x') !== -1;
 // https://forums.developer.apple.com/thread/119186
 if (browser.osx && !browser.iphone && !browser.ipod && !browser.ipad && navigator.maxTouchPoints > 1) {
     browser.ipad = true;
+}
+
+if (browser.osx && !browser.iphone && !browser.ipod && !browser.ipad) {
+    browser.osxVersion = osxVersion();
+
+    if (browser.osxVersion) {
+        if (browser.osxVersion.length >= 2) {
+            browser.osxVersion = browser.osxVersion[0] + (browser.osxVersion[1] / 10);
+        } else {
+            browser.osxVersion = browser.osxVersion[0];
+        }
+    }
 }
 
 if (userAgent.toLowerCase().indexOf('playstation 4') !== -1) {
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -220,11 +220,17 @@ function supportsHdr10(options) {
     return options.supportsHdr10 ?? (false // eslint-disable-line sonarjs/no-redundant-boolean
             || browser.tizen
             || browser.web0s
-            || browser.safari && ((browser.iOS && browser.iOSVersion >= 11) || browser.osx)
-            // Chrome mobile and Firefox have no client side tone-mapping
+            // All browsers on iOS are required to use WebKit underneath. HDR support in browsers was added in iOS 13.4 when Safari 13.1 was released.
+            //  * https://developer.apple.com/documentation/safari-release-notes/safari-13_1-release_notes
+            || browser.iOS && browser.iOSVersion >= 13.4
+            // Chrome mobile have no client side tone-mapping
             // Edge Chromium 121+ fixed the tone-mapping color issue on Nvidia
             || browser.edgeChromium && (browser.versionMajor >= 121)
-            || browser.chrome && !browser.mobile
+            // Chrome on mobile now has client side tone-mapping. Tested on Chrome Android 128 but first support version is unknown.
+            || browser.chrome && !browser.iOS && browser.mobile && (browser.versionMajor >= 128)
+            // Firefox 100+ has support for HDR on macOS/OS X. It requires OS support, which was added in macOS 10.15 Catalina.
+            //  * https://www.mozilla.org/en-US/firefox/100.0/releasenotes/
+            || browser.firefox && browser.osx && (!browser.iphone && !browser.ipod && !browser.ipad) && (browser.osxVersion >= 10.15) && (browser.versionMajor >= 100)
     );
 }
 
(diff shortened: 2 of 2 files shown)


src/scripts/browserDeviceProfile.js
```diff
@@ -659,6 +677,14 @@ export default function (options) {
         webmVideoCodecs.push('vp8');
     }
 
+    // WebM support is partial/unreliable from Safari versions 17.0-17.4. It was fixed in 17.4:
+    //  * https://developer.apple.com/documentation/safari-release-notes/safari-17_4-release-notes
+    const safariSupportsWebm = browser.safari && (false // eslint-disable-line sonarjs/no-redundant-boolean
+        || browser.versionMajor >= 15 && browser.versionMajor < 17
+        || browser.versionMajor == 17 && browser.versionMinor[1] >= 4
```
dmitrylyzo, line 684:
```suggestion
        || browser.versionMajor == 17 && browser.versionMinor >= 4
```

src/scripts/browserDeviceProfile.js
```diff
@@ -659,6 +677,14 @@ export default function (options) {
         webmVideoCodecs.push('vp8');
     }
 
+    // WebM support is partial/unreliable from Safari versions 17.0-17.4. It was fixed in 17.4:
+    //  * https://developer.apple.com/documentation/safari-release-notes/safari-17_4-release-notes
+    const safariSupportsWebm = browser.safari && (false // eslint-disable-line sonarjs/no-redundant-boolean
```
GodTamIt, line 682: Unclear why Sonarlint doesn't pick up this lint suppression. This is following the same code-style as the rest of the file.

src/scripts/browserDeviceProfile.js
```diff
@@ -233,8 +239,20 @@ function supportsHlg(options) {
 }
 
 function supportsDolbyVision(options) {
+    // Safari 13.1 added support for Dolby Vision:
+    //  * https://developer.apple.com/documentation/safari-release-notes/safari-13_1-release_notes
+    const safariSupportsDoVi = browser.safari && (false // eslint-disable-line sonarjs/no-redundant-boolean
```
GodTamIt, line 244: Unclear why Sonarlint doesn't pick up this lint suppression. This is following the same code-style as the rest of the file.

src/scripts/browserDeviceProfile.js
```diff
@@ -670,18 +696,16 @@ export default function (options) {
         if (browser.safari) {
             hlsInFmp4VideoCodecs.push('vp9');
         }
-        // webm support is unreliable on safari 17
-        if (!browser.safari
-             || (browser.safari && browser.versionMajor >= 15 && browser.versionMajor < 17)) {
+        // WebM support is unreliable in some versions of Safari.
+        if (!browser.safari || (browser.safari && safariSupportsWebm)) {
```
dmitrylyzo, line 700: You can omit the second `browser.safari` because `!X || (X && Y) = (!X || X) && (!X || Y)` and `!X || X` is always `true`.
```suggestion
        if (!browser.safari || safariSupportsWebm) {
```

src/scripts/browserDeviceProfile.js
```diff
@@ -670,18 +696,16 @@ export default function (options) {
...
         }
     }
 
     if (canPlayAv1(videoTestElement)) {
         mp4VideoCodecs.push('av1');
-        // webm support is unreliable on safari 17
-        if (!browser.safari
-             || (browser.safari && browser.versionMajor >= 15 && browser.versionMajor < 17)) {
+        // WebM support is unreliable in some versions of Safari.
+        if (!browser.safari || (browser.safari && safariSupportsWebm)) {
```
dmitrylyzo, line 708: Same.
```suggestion
        if (!browser.safari || safariSupportsWebm) {
```

src/scripts/browserDeviceProfile.js
dmitrylyzo: Iirc, it was manually rejected.

src/scripts/browserDeviceProfile.js
```diff
@@ -233,8 +239,20 @@ function supportsHlg(options) {
...
+    );
+
     return options.supportsDolbyVision ?? (false // eslint-disable-line sonarjs/no-redundant-boolean
-            || browser.safari && ((browser.iOS && browser.iOSVersion >= 13) || browser.osx)
+            || safariSupportsDoVi
+            // Chrome/Chromium-based browsers added support for Dolby Vision fallback on version 122+.
+            // This does not mean its supported, just that media extensions now report support (and more importantly, no support) properly.
+            //   * https://chromium-review.googlesource.com/c/chromium/src/+/5058372
+            || browser.chrome && (browser.versionMajor >= 122)
+            || browser.edgeChromium && (browser.versionMajor >= 122)
```
nyanmisaka, line 255: I'm not sure about Chromium on Apple, maybe @gnattu can give more insights about that. But [HLS.js](https://github.com/video-dev/hls.js) used on Windows/Linux does not yet support `dvh1/dvhe` fourcc. Even edgeChromium requires an additional plugin and an experimental flag to enable DoVi support.

So claiming Chromium supports DoVi does no good, we already use their fallback layer to play HDR/HLG/SDR.

GodTamIt: The goal here wasn't to "enable" DoVi for these browsers but to indicate that these browsers now support accurate reporting of their capabilities. See the linked Chromium CL comment for info and the bug number.

It looks like the Chromium team is staging changes to support Dolby Vision on different platforms ([macOS CL](https://chromium-review.googlesource.com/c/chromium/src/+/5667325)), so enabling this should automatically enable DoVi once that lands, no?

HLS.js not supporting might be the bigger issue here for other platforms.

src/scripts/browser.js
```diff
@@ -99,6 +99,19 @@ function iOSversion() {
     return [];
 }
 
+function osxVersion() {
+    const matches = /Mac OS X (\d+)[._](\d+)[._]?(\d+)?/i.test(navigator.userAgent);
```
gnattu, line 103: Please don't. You should not try to get macOS version from user agent because you cannot. All browsers only reports 10.15 on macOS now because a lot of websites have stupid assumptions that major version of macOS cannot be larger than 10.

src/scripts/browserDeviceProfile.js
```diff
@@ -220,11 +220,17 @@ function supportsHdr10(options) {
...
+            || browser.iOS && browser.iOSVersion >= 13.4
+            // Chrome mobile have no client side tone-mapping
             // Edge Chromium 121+ fixed the tone-mapping color issue on Nvidia
             || browser.edgeChromium && (browser.versionMajor >= 121)
-            || browser.chrome && !browser.mobile
+            // Chrome on mobile now has client side tone-mapping. Tested on Chrome Android 128 but first support version is unknown.
+            || browser.chrome && !browser.iOS && browser.mobile && (browser.versionMajor >= 128)
+            // Firefox 100+ has support for HDR on macOS/OS X. It requires OS support, which was added in macOS 10.15 Catalina.
+            //  * https://www.mozilla.org/en-US/firefox/100.0/releasenotes/
+            || browser.firefox && browser.osx && (!browser.iphone && !browser.ipod && !browser.ipad) && (browser.osxVersion >= 10.15) && (browser.versionMajor >= 100)
```
gnattu, line 233: Please don't. The support is very bad and even direct playing an mp4 file will result in washed out color as firefox does not do client-side tone mapping and does not trigger HDR colors in all cases. It seems like such support is never completed on platforms other than YouTube.

src/scripts/browserDeviceProfile.js
```diff
@@ -659,6 +677,14 @@ export default function (options) {
         webmVideoCodecs.push('vp8');
     }
 
+    // WebM support is partial/unreliable from Safari versions 17.0-17.3. It was fixed in 17.4:
+    //  * https://developer.apple.com/documentation/safari-release-notes/safari-17_4-release-notes
+    const safariSupportsWebm = browser.safari && (false // eslint-disable-line sonarjs/no-redundant-boolean
```
gnattu, line 682: Please don't. It is still having problem for very long videos and I'm still experiencing player crashing on Safari TP tagged at 18.0. Please revert that.

src/scripts/browserDeviceProfile.js
gnattu: Please don't. It is still lying about that support and will report it 'maybe' plays dovi profile 5 videos when it will do client side tone mapping with wrong colorspace

gnattu (changes requested):
Basically all of your changes are not reliable and should be reverted. Sorry. Next time please try verify the media capabilities yourself before trusting the release notes from vendors that much. From my experiences, what they claim and what they do is usually different.

src/scripts/browserDeviceProfile.js
GodTamIt: @gnattu I tested this on my macOS 14.6 and Firefox 129.0.1. Is there a particular setup you're running into that's breaking for you?

All files here seem to work for me: https://drive.google.com/drive/folders/1dSPUmJ_Y99-VL6TxfndPbISyUfRJ3SAY

gnattu: > Is there a particular setup you're running into that's breaking for you?

It is having problems identifying HDR files so you are lucky that all your files are working because some of them are not. For example, this sample video from LG will trigger HDR playback in both Safari and Chrome but not in Firefox:

https://mega.nz/file/5lsGUYpT#B_KNF251rODrq5lWwGYHxSuBsiqEZQsGCRCWO83iAdY

GodTamIt: This is because the video provided above is in an unusual format of VP9 + AAC in an mp4 container. Firefox's parsing logic is stricter about the container <-> codec mappings. This behavior isn't unique to Firefox -- iOS's browsers also have a history of not playing nice with VP9 in mp4 containers either.

I will create a separate change to fix this specifically for Firefox and break up this change into individual pieces.

PR #5970, opened 2024-08-26
Migrate to material react table

thornbill:
**Changes**
* Replaces the `@mui/x-data-grid` library with `material-react-table`. This library is still based on mui (and tanstack table), but without any paywalled features (unlike the current library).
* Fixes some issues with the `useLogEntries` and `useUsers` hooks
* Updates components and hooks to use new project structure

In a future PR we could improve the filters over the default text based filtering. (The datetime filtering requires some additional work to ensure our current locale is available in the React context.)

**Issues**
N/A


src/apps/dashboard/features/activity/components/LogLevelCell.tsx
```diff
@@ -0,0 +1,13 @@
...
+import React, { type FC } from 'react';
+
+import { ActivityLogEntryCell } from '../types/ActivityLogEntryCell';
+import LogLevelChip from './LogLevelChip';
+
+const LogLevelCell: FC<ActivityLogEntryCell> = ({ cell }) => (
+    cell.getValue<LogLevel | undefined>() ? (
+        <LogLevelChip level={cell.getValue<LogLevel>()} />
+    ) : undefined
+);
```
dmitrylyzo, line 11: `getValue` once. Worth?
```suggestion
const LogLevelCell: FC<ActivityLogEntryCell> = ({ cell }) => {
    const v = cell.getValue<LogLevel | undefined>();
    return v ? (
        <LogLevelChip level={v} />
    ) : undefined;
};
```

dmitrylyzo (commented):
Seems to work, in general.
There are checkboxes to hide columns, but for some reason they don't work.

dmitrylyzo (approved):

PR #5978, opened 2024-08-28
Make Live TV compatibility profiles customizable

gnattu:
**Changes**

This adds the GUI to set compatibility profiles for LiveTV tuners.

**Issues**

Depends on https://github.com/jellyfin/jellyfin/pull/12529


src/controllers/livetvtuner.js
```diff
@@ -165,6 +174,24 @@ function onTypeChange() {
...
+        view.querySelector('.fldStreamSharing').classList.remove('hide');
+    } else {
+        view.querySelector('.fldStreamSharing').classList.add('hide');
+    }
+
+    if (supportsFallbackBitrate) {
+        view.querySelector('.fldFallbackMaxStreamingBitrate').classList.remove('hide');
+    } else {
+        view.querySelector('.fldFallbackMaxStreamingBitrate').classList.add('hide');
+    }
```
dmitrylyzo, line 193: Use `classList.toggle`.

src/strings/en-us.json
```diff
@@ -28,8 +28,10 @@
     "AllowSubtitleManagement": "Allow this user to edit subtitles",
     "AllowFfmpegThrottling": "Throttle Transcodes",
     "AllowFfmpegThrottlingHelp": "When a transcode or remux gets far enough ahead from the current playback position, pause the process so it will consume fewer resources. This is most useful when watching without seeking often. Turn this off if you experience playback issues.",
+    "AllowFmp4TranscodingContainerHelp": "Allow fmp4 transcoding container for this tuner to enable HEVC and HDR contents. Not all tuners are compatible with this container. Disable this if you experience playback issues.",
```
dmitrylyzo, line 31: We use `fMP4` for the "prefer" option. Shouldn't we do the same here?
```suggestion
    "AllowFmp4TranscodingContainerHelp": "Allow fMP4 transcoding container for this tuner to enable HEVC and HDR contents. Not all tuners are compatible with this container. Disable this if you experience playback issues.",
```

src/strings/en-us.json
```diff
@@ -561,7 +564,9 @@
     "LabelAllowContentWithTags": "Allow items with tags",
     "LabelAllowedRemoteAddresses": "Remote IP address filter",
     "LabelAllowedRemoteAddressesMode": "Remote IP address filter mode",
+    "LabelAllowFmp4TranscodingContainer": "Allow fmp4 transcoding container",
```
dmitrylyzo, line 567:
```suggestion
    "LabelAllowFmp4TranscodingContainer": "Allow fMP4 transcoding container",
```

dmitrylyzo:
You reverted `fmp4 -> fMP4`. By accident?

gnattu:
> You reverted `fmp4 -> fMP4`. By accident?

Rebasing hates me. Fixed

src/controllers/livetvtuner.html
```diff
@@ -38,6 +38,11 @@ <h1 class="sectionTitle">${HeaderLiveTvTunerSetup}</h1>
                     <div class="fieldDescription">${SimultaneousConnectionLimitHelp}</div>
                 </div>
 
+                <div class="inputContainer fldFallbackMaxStreamingBitrate hide">
+                    <input is="emby-input" type="number" pattern="[0-9]*" required="required" min="0" step="1" class="txtFallbackMaxStreamingBitrate" label="${LabelFallbackMaxStreamingBitrate}" autocomplete="off" value="30" />
```
dmitrylyzo, line 42: Currently, we can pass `0` and it is saved correctly. But it's actually an invalid value - [Min(..., 0)](https://github.com/jellyfin/jellyfin/blob/62712aa12cdc88c524923cb318783b23a8a7a2c5/Jellyfin.Api/Helpers/StreamingHelpers.cs#L152). Moreover, we [use `30` in case of `0`](https://github.com/jellyfin/jellyfin-web/blob/10a8dead0caae9b99f8a08e9b2862b9861dfc29d/src/controllers/livetvtuner.js#L67) in web.

Maybe use `1` as a minimum?
```suggestion
                    <input is="emby-input" type="number" pattern="[0-9]*" required="required" min="1" step="1" class="txtFallbackMaxStreamingBitrate" label="${LabelFallbackMaxStreamingBitrate}" autocomplete="off" value="30" />
```

gnattu: I think it might be more reasonable to do such thing on the server side because not only this field might pass incorrect value, and any client/plugins can also set some unreasonable values. But it does not hurt to set the lower limit to 1 on web as well.

dmitrylyzo (approved):

theguymadmax:
Are these profiles exclusive to M3U tuners? They don’t appear when HD HomeRun is selected as the tuner type.

gnattu:
> Are these profiles exclusive to M3U tuners? They don’t appear when HD HomeRun is selected as the tuner type.

HDHR always requires most compatible transcoding profile and will always return a shared HttpStream with current implementation because that was what verified to work and not enforcing most compatible profile caused lots of problems. To make it worse, I cannot test this particular tuner at all.

PR #5983, opened 2024-08-28
Fix overly strict dovi level testing

nyanmisaka:
4k@60fps is usually only seen in demos so it's a bit overkill for testing dovi caps. Lower it to the more common 4k@24fps.

Ref: https://professional.dolby.com/siteassets/content-creation/dolby-vision-for-content-creators/dolby-vision-streams-within-the-http-live-streaming-format-v2.0-13-november-2018.pdf

**Changes**
- Fix overly strict dovi level testing

**Issues**
- iPad Pro 2018 only reports up to 4k30 dovi support after upgrading to iPadOS 17, but can actually play 4k60 normally and it used to report 4k60 on older systems.

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -242,16 +242,16 @@ function supportedDolbyVisionProfilesHevc(videoTestElement) {
     if (browser.xboxOne) return [5, 8];
 
     const supportedProfiles = [];
-    // Profiles 5/8 4k@60fps
+    // Profiles 5/8 4k@24fps
     if (videoTestElement.canPlayType) {
         if (videoTestElement
-            .canPlayType('video/mp4; codecs="dvh1.05.09"')
+            .canPlayType('video/mp4; codecs="dvh1.05.06"')
             .replace(/no/, '')) {
             supportedProfiles.push(5);
         }
         if (
             videoTestElement
-                .canPlayType('video/mp4; codecs="dvh1.08.09"')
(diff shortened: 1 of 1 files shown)


gnattu (approved):
Actually I was facing some bugs that in rare case the DOVI mode is not triggered and Profile 5 60fps videos are rendered in wrong color, but I cannot reliably reproduce it now. Probably this is the quirk but let's only test for 24fps for now.

BTW, if someone has iPad Pro 2024 with M4 chip I'd like to get the report of Dolby Vision Profile 10 playback testing results. It is rumored to support `dav1.10.09` but I'm not verified that yet.

nyanmisaka:
> Actually I was facing some bugs that in rare case the DOVI mode is not triggered and Profile 5 60fps videos are rendered in wrong color, but I cannot reliably reproduce it now. Probably this is the quirk but let's only test for 24fps for now.

I can't imagine a device that supports Dolby Vision but doesn't support 4k60 playback.

> BTW, if someone has iPad Pro 2024 with M4 chip I'd like to get the report of Dolby Vision Profile 10 playback testing results. It is rumored to support `dav1.10.09` but I'm not verified that yet.

`dav1` is already in the documentation. https://developer.apple.com/documentation/http-live-streaming/hls-authoring-specification-for-apple-devices-appendixes#The-SUPPLEMENTAL-CODECS-attribute

Tested that jellyfin-ffmpeg7 can tonemap and remux `dav1`. It seems that no player yet is capable of this fourcc, software decoders always expect `av01`.

thornbill (approved):

PR #5986, opened 2024-08-29
Add audio options to workaround compatability problems

gnattu:
**Changes**

This adds some audio compatibility options to workaround certain issues:

- Allows the client to opt-out the VBR audio encoding enabled on server to prevent buffering issue
- Allows to client to force remuxing flac files when the file has metadata that the browser rejects or will calculate wrong timestamp
- Allows the client to force remuxing mp3 files when some files will lead to inaccurate timestamps

**Issues**

Fixes https://github.com/jellyfin/jellyfin/issues/3731 (incompatible metadata)
Fixes https://github.com/jellyfin/jellyfin/issues/11113 (inaccurate timestamp)


felix920506 (commented):
I think it would be better to have the "force remux flac" option on by default for chromium and Firefox, as these browsers are known to have problems with flac metadata. This would ensure maximum compatibility for most users. 

We can then tell users to disable it if they are that allergic to remuxing.

gnattu:
I still think something like this should be the last resort and not a default setting unless this issue is very widespread, but IMO it is not the case yet.

src/components/playbackSettings/playbackSettings.template.html
```diff
@@ -204,6 +212,28 @@ <h2 class="sectionTitle">
         </div>
     </div>
 
+    <div class="verticalSection verticalSection-extrabottompadding">
+        <h2 class="sectionTitle">
+            ${HeaderAudioAdvanced}
+        </h2>
+
```
thornbill, line 219: The Audio Normalization option should probably be moved to this new section.

dmitrylyzo (approved):

SirWobbyTheFirst:
Is there a way to force this on for all users on a server? I'm having issues with Firefox users being unable to stream FLAC and I'd rather just default that option to on for all users rather than guide them through it manually.

PR #5990, opened 2024-08-29
Allow editing of original title for Season and Episode items

scampower3:
**Changes**

Allows editing of original title for Season and Episode items. 

**Issues**

### DIFF
--- a/src/components/metadataEditor/metadataEditor.js
+++ b/src/components/metadataEditor/metadataEditor.js
@@ -540,6 +540,7 @@ function setFieldVisibilities(context, item) {
     } else {
         hideElement('#fldPath', context);
     }
+    
     if (item.Type === 'Series' || item.Type === 'Season' || item.Type === 'Episode' || item.Type === 'Movie' || item.Type === 'Trailer' || item.Type === 'Person') {
         showElement('#fldOriginalName', context);
     } else {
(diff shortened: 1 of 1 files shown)


src/components/metadataEditor/metadataEditor.js
```diff
@@ -541,7 +541,7 @@ function setFieldVisibilities(context, item) {
         hideElement('#fldPath', context);
     }
 
-    if (item.Type === 'Series' || item.Type === 'Movie' || item.Type === 'Trailer' || item.Type === 'Person') {
+    if (item.Type === 'Series' || item.Type === 'Season' || item.Type === 'Episode' || item.Type === 'Movie' || item.Type === 'Trailer' || item.Type === 'Person') {
```
thornbill, line 544: We should do something like:

```js
    if ([ 'Series', 'Season', 'Episode', ... ].includes(item.Type)) {
```

Bonus points if you also use the `BaseItemKind` enum values instead of hardcoded strings. :wink:

thornbill (changes requested):

src/components/metadataEditor/metadataEditor.js
```diff
@@ -21,7 +21,7 @@ import ServerConnections from '../ServerConnections';
 import toast from '../toast/toast';
 import { appRouter } from '../router/appRouter';
 import template from './metadataEditor.template.html';
-import { SeriesStatus } from '@jellyfin/sdk/lib/generated-client';
+import { BaseItemKind, SeriesStatus } from '@jellyfin/sdk/lib/generated-client';
```
thornbill, line 24: To ensure tree-shaking can happen during the build, we should use the full path for these imports `@jellyfin/sdk/lib/generated-client/models/base-item-kind` (I think).

thornbill (approved):

PR #5997, opened 2024-08-30
Fix @tanstack/query-devtools on legacy browsers

dmitrylyzo:
**Changes**
- Add `@tanstack/query-devtools` to Babel.
- Don't load `ReactQueryDevtools` on unsupported devices.

**Issues**
N/A

**Remarks**
- webOS 1.2 - ReactQueryDevtools is skipped.
- webOS 5 - ReactQueryDevtools seems to work.
We probably don't need it on TVs though - weird navigation.


src/RootApp.tsx
```diff
@@ -18,7 +18,9 @@ const RootApp = () => (
                 </WebConfigProvider>
             </UserSettingsProvider>
         </ApiProvider>
-        <ReactQueryDevtools initialIsOpen={false} />
+        {window.Proxy && ( // '@tanstack/query-devtools' requires 'Proxy', which cannot be polyfilled for legacy browsers
```
thornbill, line 21: Should we also check `browser.tv`?

dmitrylyzo: To hide it on TVs?

thornbill: Yes since it makes navigation weird... although I have been thinking about adding a custom menu entry or something for it because I hate the floating button 😅

thornbill (approved):

Issue #3994, opened 2022-10-04
[Issue]: Audio track reset on next episode


Plinz:
Please describe your bug

When I'm watching a TV show with an audio track that is not the default one, and I finish an episode or use button next episode, the next episode doesn't keep the previous audio track on. But, on UI, it shows that it is selected, so I need to set another one first and then set back the one I want.

Step to reproduce:

    watch a TV show with multiple audio track
    choose a different audio track than the default one
    go to next episode

Expected:

    The audio track is kept in the next episode.

What happen instead:

    The audio track from previous episode is selected, but default audio track is playing.

Jellyfin Version
10.8.4

    OS: Linux
    Virtualization: no
    Clients: Browser
    Browser: Firefox 105.0.1
    FFmpeg Version: 5.1.1
    Playback Method: Streaming direct
    Hardware Acceleration: none
    Plugins: none
    Reverse Proxy: nginx
    Base URL: /jellyfin
    Networking: Host
    Storage: local

kiawizard:
This is a very important issue, please do not close it. Imagine you have a series with 2 different language audio tracks. You are watching the secondary language now. If you have "Play default audio track regardless of language" unchecked in your profile (default), at the beginning of each episode you have to switch audio track twice - to 1st, then to 2nd. With the setting checked, you have to change audio once per episode. Normally you'd want to select 2nd language once and then watch the whole season :-)

kiawizard:
this issue still affects you in the latest release

jmerdich:
Still present on 10.8.11 (tested on Firefox, Chrome, and Android-app-webplayer).

This wouldn't be so bad if the default audio track was playing **and** the audio dropdown indicated it, but because they're mismatched, you have to change audio settings twice to get to the language that was playing on the previous episode.

heschlie:
Seeing the same issue at least on the web client, I haven't seen it on the Android client. Also on Firefox, haven't tested on other browsers.

vladkar:
I’m experiencing the same issue where I need to reload the audio track for each episode. Although it appears correct in the user interface, I have to switch it back to a different audio track and then select the desired one. If I use subtitles, I also need to turn them off and then back on again because they freeze after the first adjustment. As a result, I have to make four changes in total for each episode if I need both an audio track and subtitles. This issue has persisted for years across various versions.

filipfigzalski:
I have the same issue, after each episode language changes to the default one, even though the GUI shows otherwise.

[...]

PR #4852, opened 2023-10-07
Next episode has mismatched language between server and ui

jmerdich:
**Root cause**

The autoset options only seem to take effect on non-transcoded media-- the transcode URL is never updated! This means that if the selected audio does not match the default, the audio/subs are in the default but the UI reflects the auto-set variant (showing a language that isn't playing).

**Changes**

When setting the track in the the auto-set logic, also trivially change the audio index in the transcode URL to match.

This fixes it in most cases, but there may be some edge cases that need to go through the full validation to recreate the URL. I don't know the code well enough for that, though.

**Issues**

Fixes #3994

PR #5998, opened 2024-08-30
Next episode has mismatched language between server and ui (rebased)

Kinsteen:
Simple rebase on new master to remove merge conflicts of #4852. Didn't test yet, would be happy if people could!

> **Root cause**
> 
> The autoset options only seem to take effect on non-transcoded media-- the transcode URL is never updated! This means that if the selected audio does not match the default, the audio/subs are in the default but the UI reflects the auto-set variant (showing a language that isn't playing).
> 
> **Changes**
> 
> When setting the track in the the auto-set logic, also trivially change the audio index in the transcode URL to match.
> 
> This fixes it in most cases, but there may be some edge cases that need to go through the full validation to recreate the URL. I don't know the code well enough for that, though.
> 
> **Issues**
> 
> Fixes https://github.com/jellyfin/jellyfin-web/issues/3994


viown:
Was able to reproduce the mentioned issue, and after testing this, it does seem to fix it.

dmitrylyzo:
_In addition to https://github.com/jellyfin/jellyfin-web/pull/4852#issuecomment-2118867286_

Changing streams after receiving PlaybackInfo may result in unwanted remuxing/transcoding:
- The "best" subtitles may not require transcoding.
- The "best" audio may not require remuxing. _<- less possible_

It's better to do this on the server side so that every client gets this auto-select feature.

But speaking of web, we should do ranking and update corresponding fields before getting PlaybackInfo:
https://github.com/jellyfin/jellyfin-web/blob/f4a872d78073823ba6102d9fc5245804d18e26d6/src/components/playback/playbackmanager.js#L2577-L2578

viown:
> _In addition to [#4852 (comment)](https://github.com/jellyfin/jellyfin-web/pull/4852#issuecomment-2118867286)_
> 
> Changing streams after receiving PlaybackInfo may result in unwanted remuxing/transcoding:
> 
>     * The "best" subtitles may not require transcoding.
> 
>     * The "best" audio may not require remuxing. _<- less possible_
> 
> 
> It's better to do this on the server side so that every client gets this auto-select feature.
> 
> But speaking of web, we should do ranking and update corresponding fields before getting PlaybackInfo:
> 
> https://github.com/jellyfin/jellyfin-web/blob/f4a872d78073823ba6102d9fc5245804d18e26d6/src/components/playback/playbackmanager.js#L2577-L2578

Setting the audioStreamIndex/subtitleStreamIndex in options does not seem to affect the stream URL in some scenarios (only when remuxing/direct playing). Any idea why that is?

dmitrylyzo:
> Setting the audioStreamIndex/subtitleStreamIndex in options does not seem to affect the stream URL in some scenarios (only when remuxing/direct playing). Any idea why that is?

Because they are processed on the client side when DirectPlay. It should use `DefaultAudioStreamIndex` and `DefaultSubtitleStreamIndex`.

Kinsteen:
Seems so, closing

PR #6002, opened 2024-08-30
Only show IsAVC for video streams

Bond-009:


src/components/itemMediaInfo/itemMediaInfo.js
```diff
@@ -107,7 +107,7 @@ function getMediaSourceHtml(user, item, version) {
         if (stream.CodecTag) {
             attributes.push(createAttribute(globalize.translate('MediaInfoCodecTag'), stream.CodecTag));
         }
-        if (stream.IsAVC != null) {
+        if (stream.Type === 'Video' && stream.IsAVC != null) {
```
dmitrylyzo, line 110: Although this change looks fine, I see that `IsAVC` is supposed to be nullable. Did we accidentally set it for non-video streams?

Bond-009: At least on my install, yes

dmitrylyzo: It probably becomes `false` here https://github.com/jellyfin/jellyfin/blob/72077490447d3ff588563cc7f8aa705e1a7e9ecc/Emby.Server.Implementations/Data/SqliteItemRepository.cs#L5627
`TryGetBoolean` doesn't seem to return `null` if the value in the DB is `NULL`.
Wouldn't `result` be assigned the value `false` instead of `null`?
https://github.com/jellyfin/jellyfin/blob/72077490447d3ff588563cc7f8aa705e1a7e9ecc/Emby.Server.Implementations/Data/SqliteExtensions.cs#L151

Bond-009: > TryGetBoolean doesn't seem to return null if the value in the DB is NULL.
Wouldn't result be assigned the value false instead of null?

No, it doesn't assign any value as the Try method returns false bc it's null inside of the DB.
We don't have to assign anything as the default value is `null`

dmitrylyzo: This means that we saved `false` in the database. I checked my database - not a single `NULL`.

thornbill: @Bond-009 any thoughts on this?

Bond-009:
https://github.com/jellyfin/jellyfin/pull/16507 should fix it server side
(does fix anything that's already stored in the DB)

thornbill:
So this should be closed?

PR #6004, opened 2024-08-31
Add ability to select QSV device in the UI

evanjarrett:
**Changes**

This provides an input field to specify the device used for QSV rendering.

**Issues**
This depends on the backend PR here: https://github.com/jellyfin/jellyfin/pull/12384

### DIFF
--- a/src/strings/en-us.json
+++ b/src/strings/en-us.json
@@ -840,6 +840,8 @@
     "LabelPublicHttpsPortHelp": "The public port number that should be mapped to the local HTTPS port.",
     "LabelPublishedServerUri": "Published Server URIs",
     "LabelPublishedServerUriHelp": "Override the URI used by Jellyfin, based on the interface, or client IP address. For example: internal=http://jellyfin.example.com, external=https://jellyfin.example.com, or all=https://jellyfin.example.com",
+    "LabelQsvDevice": "QSV Device",
+    "LabelQsvDeviceHelp": "This is the render node that is used for hardware acceleration.",
     "LabelQuickConnectCode": "Quick Connect code",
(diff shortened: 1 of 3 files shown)


src/strings/en-us.json
```diff
@@ -789,6 +789,8 @@
     "LabelPublicHttpsPortHelp": "The public port number that should be mapped to the local HTTPS port.",
     "LabelPublishedServerUri": "Published Server URIs",
     "LabelPublishedServerUriHelp": "Override the URI used by Jellyfin, based on the interface, or client IP address.",
+    "LabelQsvDevice": "QSV Device",
+    "LabelQsvDeviceHelp": "This is the render node that is used for hardware acceleration.",
```
nyanmisaka, line 793: Please add a slightly more detailed description of what the user should fill in. And emphasize that this should be left blank by default.

evanjarrett: Hows something like this?

"LabelQsvDeviceHelp": "Specify the device for Intel QSV. On Linux, this is the render node, e.g., /dev/dri/renderD128. On Windows, this is the device index starting from 0. Leave blank to use the default device."

nyanmisaka: `"LabelQsvDeviceHelp": "Specify the device for Intel QSV on a multi-GPU system. On Linux, this is the render node, e.g., /dev/dri/renderD128. On Windows, this is the device index starting from 0. Leave blank unless you know what you are doing."`

Made some changes to keep away newbie users from touching this.

Bond-009:
Server PR is now merged

nyanmisaka:
@thornbill I think this is no longer blocked by the server and can be merged to 10.10.

thornbill (approved):

PR #6005, opened 2024-08-31
Update slideshow.js with new default from 3 to 7 seconds per slide

markster1700:
Changed default 3 seconds slide hide function to 7 seconds. 3 seconds per slide as a default is very short and unreasonable value. Maybe in the future this can be implemented as a parameter.

**Changes**

**Issues**

### DIFF
--- a/src/components/slideshow/slideshow.js
+++ b/src/components/slideshow/slideshow.js
@@ -617,7 +617,7 @@ export default function (options) {
      */
     function startHideTimer() {
         stopHideTimer();
-        hideTimeout = setTimeout(hideOsd, 3000);
+        hideTimeout = setTimeout(hideOsd, 7000);
     }
 
     /**


dmitrylyzo:
You are currently changing the OSD hide timeout, but you probably wanted to change the timeout between slides.
There's a PR for that: https://github.com/jellyfin/jellyfin-web/pull/3744

markster1700:
I wanted to change how long each slide is visible to user. At present its 3 seconds, which is a very fast slideshow. 7 seconds its a value that is more reasonable to accept as a default.

thornbill:
https://github.com/jellyfin/jellyfin-web/pull/5675#issuecomment-2219729402

markster1700:
The option you mentioned was 2 years ago and it did not get implemented as far as I can tell. So, what is going to happen. Simple change default is faster. 
Or please merge the suggested changes from 2 years ago. What are we waiting for?

markster1700:
Great news. I just updated to lates 10.11.5 but cannot find any setting to configure this delay.

PR #6016, opened 2024-09-04
Show slideshow controls when touched

tcely:
It appears that a previous commit changed the blocks so that only mouse events were recognized.

**Changes**

Moved the `showOsd` call outside of the "mouse" block.

**Issues**

### DIFF
--- a/src/components/slideshow/slideshow.js
+++ b/src/components/slideshow/slideshow.js
@@ -732,9 +732,8 @@ export default function (options) {
 
             obj.x = eventX;
             obj.y = eventY;
-
-            showOsd();
         }
+        showOsd();
     }
 
     /**


thornbill (approved):
This seems to have always worked this way. It shouldn't be an issue on mobile since the controls are always visible there, but it likely wouldn't have worked properly on computers with touchscreen monitors.

tcely:
> It shouldn't be an issue on mobile since the controls are always visible there, but it likely wouldn't have worked properly on computers with touchscreen monitors.

I can confirm that on mobile, both the app and browser, on my phone the controls disappear when you pause and they do not return.

tcely:
Controls now appear in `10.9.11` as expected. Thanks!

Issue #6007, opened 2024-09-02
Devices in User Device Access List not correctly named 


HBA-Home:
**Describe The Bug**
After naming devices in the Devices tab, (for example "HBA iPad Air" or "HBA iPhone 15") and then going to user HBA to set the device access list, it doesn't show the devices as I named them, it shows them as ("iPad - Swiftfin iPadOS", or "iPhone - Jellyfin Mobile (iOS)")

**Steps To Reproduce**

1. Go to the Devices tab
2. Name a device
3. Go to the Users tab and manage a user's Device Access list
4. Devices are listed in their original default name and not the custom name you've set

**Expected Behavior**
Devices in the User's Device Access list should have the correct name as set by the user in the Device Tab.

**Screenshots**

Devices Properly Named in the Devices Tab:
![image](https://github.com/user-attachments/assets/ff4f9d75-11fe-4be8-8344-8995f1a9ffed)

Devices in the User's Device Access List not having the proper name:
![image](https://github.com/user-attachments/assets/9a2b0f2f-d0f1-4319-a293-b62ea2cc93b4)

PR #6018, opened 2024-09-04
Prefer custom device name in device access list

viown:
**Changes**
Use custom device name when available in device access list.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-web/issues/6007

### DIFF
--- a/src/apps/dashboard/routes/users/access.tsx
+++ b/src/apps/dashboard/routes/users/access.tsx
@@ -17,6 +17,7 @@ type ItemsArr = {
     Name?: string | null;
     Id?: string | null;
     AppName?: string | null;
+    CustomName?: string | null;
     checkedAttribute?: string
 };
 
@@ -112,6 +113,7 @@ const UserLibraryAccess = () => {
                 Id: device.Id,
                 Name: device.Name,
                 AppName: device.AppName,
+                CustomName: device.CustomName,
                 checkedAttribute: checkedAttribute
             });
         }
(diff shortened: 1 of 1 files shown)


src/apps/dashboard/routes/users/access.tsx
```diff
@@ -307,7 +309,7 @@ const UserLibraryAccess: FunctionComponent = () => {
                                 key={Item.Id}
                                 className='chkDevice'
                                 itemId={Item.Id}
-                                itemName={Item.Name}
+                                itemName={Item.CustomName || Item.Name}
```
danditomaso, line 312: This may cause a problem if a user choose to use an empty string as a custom name (as silly as that might be). Wondering if this should be flipped to use nullish coalescing (??) instead of | |

dmitrylyzo: I think we don't need empty custom name. `||` will work as needed - name will be used.
The user can still use a space (`' '`) though. 😅

danditomaso: Fair point, but if we dont need custom name the author should remove it?

viown: When a user wants to remove a custom name from a device, they'll clear the textbox and hit save. This does result in an empty custom name, but it's most likely not what the user wants to see.

thornbill (approved):

PR #6020, opened 2024-09-04
Add tag support to list view

thornbill:
**Changes**
* Adds tag support to the list view
* Replaces the search link for tags with the list view link

**Issues**
N/A

### DIFF
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -1080,7 +1080,7 @@ function renderDetails(page, item, apiClient, context) {
         renderLinks(page, item);
     }
 
-    renderTags(page, item, context);
+    renderTags(page, item);
     renderSeriesAirTime(page, item);
 }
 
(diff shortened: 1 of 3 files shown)


viown:
https://github.com/jellyfin/jellyfin-web/blob/d01cfa498b5f3d8b8fb3996aa1cf4949b7cd0bd6/src/controllers/list.js#L1211

Maybe we can add `params.tag` here so the title and year are displayed.

thornbill:
Thanks for the feedback @viown! I made some adjustments to add `tag` as the `type` as I noticed the default sort order was also weird without specifying a `type`.

src/controllers/itemDetails/index.js
```diff
@@ -1309,8 +1309,13 @@ function renderTags(page, item) {
     }
 
     tags.forEach(tag => {
+        const href = appRouter.getRouteUrl('tag', {
+            context,
```
grafixeyehero, line 1313: removed unused context

grafixeyehero (commented):
LGTM 🚀

grafixeyehero (approved):

Issue #6025, opened 2024-09-05
When an item is selected via the overflow menu it requires two clicks to deselect it using the checkbox.


danditomaso:
**Describe The Bug**
When hovering on any main card (tv shows, movies, music, audio books, etc), using the overflow menu and pressing the select option, this will result in the item being selected, unfortuntely it appears there is a bug as it now requires two clicks in the checkbox to deselect the selected item via the checkbox.. 

**Steps To Reproduce**
1. Go to the home screen
2. Hover on TV Shows, click on the three dot menu (overflow menu)
3. Click "Select" option from menu
4. That item should now be selected, and the title ribbon shows one item selected
5. Deselecting that item using the checkbox requires two clicks

**Expected Behavior**
Whenever a single item is selected, it should require a single click to deselect it using the checkbox, similar to to how the X in the title bar operates.

**Screenshots**
<img width="1904" alt="Screenshot 2024-09-05 at 2 33 34 PM" src="https://github.com/user-attachments/assets/1e682349-5fbc-447c-a5d2-1c065888f758">
<img width="1884" alt="Screenshot 2024-09-05 at 2 33 40 PM" src="https://github.com/user-attachments/assets/d3dcb08c-7725-4a49-91ac-9466e7c43e0c">
<img width="1908" alt="Screenshot 2024-09-05 at 2 33 47 PM" src="https://github.com/user-attachments/assets/46ad2415-8bca-4054-99a8-5e662e8f6f26">
<img width="1905" alt="Screenshot 2024-09-05 at 2 33 56 PM" src="https://github.com/user-attachments/assets/ba2956b9-5122-43aa-85f6-2dfd8cda87d9">

**System (please complete the following information):**
 - Platform: Mac OSX 14.6.1
 - Browser: Firefox
 - Jellyfin Version: 10.9.0

PR #6027, opened 2024-09-06
Fix double click to uncheck in multi-select menu

viown:
Selection panels have a "checkedInitial" class which is used to prevent an immediate check and uncheck when long pressing a card element. This class will still be there when clicking on 3 dots -> select, resulting in the user having to click twice to uncheck the initial element.

**Changes**
Add an argument to only add this class when long pressing.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-web/issues/6025


prsantos-com:
Confirmed the bug was present and tested the fix branch to verify that the changes fixed the bug. Deselecting the initial item now only requires one click as expected.

thornbill (approved):

PR #6028, opened 2024-09-07
Update behavior of expanding drawer sections

thornbill:
**Changes**
* Changes the behavior of the expanding drawer sections to expand/collapse and not navigate on click

**Issues**
N/A


crobibero (approved):
Works for me when tested locally :+1:

nielsvanvelzen (approved):
Much better behavior!

Issue #5878, opened 2024-08-05
Connection with MPV Shim gets interrupted every time I refresh the page.


mkrsticch:
### This issue respects the following points:

- [x] This is a **bug**, not a question or a configuration issue; Please visit our forum or chat rooms first to troubleshoot with volunteers, before creating a report. The links can be found [here](https://jellyfin.org/contact/).
- [X] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [X] I'm using an up to date version of Jellyfin Server stable, unstable or master; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [X] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [X] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Description of the bug

After upgrading to version 10.9.9 of Jellyfin, the connection with MPV Shim v2.8.0 gets interrupted every time I refresh the page. 

### Reproduction steps

1. Sign in on the Jellyfin web client MPV Shim
2. Connect the web client MPV Shim
3. Refresh the page
4. MPV Shim is disconnected now

### What is the current _bug_ behavior?

Connection with MPV Shim gets interrupted every time I refresh the page.

### What is the expected _correct_ behavior?

MPV Shim should stay connected

### Jellyfin Server version

10.9.8+

### Specify version number

10.9.9

### Specify the build version

10.9.9

### Environment

```markdown
- OS: ADM 4.3.1.R752
- Linux Kernel:
- Virtualization: Docker
- Clients: MPV Shim 2.8.0
- Browser: Chrome
- FFmpeg Version: 
- Playback Method: Direct
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

```shell
[2024-08-05 14:30:16.900 +02:00] [INF] [1] Main: Jellyfin version: "10.9.9"
[...]

PR #6036, opened 2024-09-08
Fix autocast listener

viown:
When debugging the release version, I noticed that the 'websocketopen' event isn't assigned to the api client when it's triggered:
image
But it's still triggered after the "initializing connection listener" log message, so I can only assume it's being assigned to the wrong api client.

**Changes**
Add the websocket event to all api clients and listen for new ones

**Issues**
May fix https://github.com/jellyfin/jellyfin-web/issues/5878

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -74,7 +74,7 @@ build: ${__JF_BUILD_VERSION__}`);
     }
 
     // Initialize automatic (default) cast target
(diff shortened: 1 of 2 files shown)


viown:
Rebased to master. Confirmed changes still work as expected.

thornbill (approved):

PR #6039, opened 2024-09-08
Add detail view buttons

grafixeyehero:
**Changes**
- Add detail view buttons
- Add global Api hooks


src/apps/experimental/features/details/components/buttons/CancelSeriesTimerButton.tsx
```diff
@@ -0,0 +1,66 @@
...
+                        timerId: itemId
+                    },
+                    {
+                        onSuccess: async () => {
+                            toast(globalize.translate('SeriesCancelled'));
+                            loading.hide();
+                            navigate('/livetv.html');
+                        },
+                        onError: (err: unknown) => {
+                            console.error(
```
dmitrylyzo, line 42: Shouldn't we hide loading when an error occurs?
Maybe also a `toast` with an error message so the user can notice it?

src/apps/experimental/features/details/components/buttons/CancelTimerButton.tsx
```diff
@@ -0,0 +1,58 @@
...
+                onSuccess: async () => {
+                    toast(globalize.translate('RecordingCancelled'));
+                    loading.hide();
+                    await queryClient.invalidateQueries({
+                        queryKey
+                    });
+                },
+
+                onError: (err: unknown) => {
+                    console.error(
```
dmitrylyzo, line 38: Shouldn't we hide loading when an error occurs?
Maybe also a `toast` with an error message so the user can notice it?

src/apps/experimental/features/details/components/buttons/ShuffleButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+const ShuffleButton: FC<ShuffleButtonProps> = ({ item }) => {
+    const shuffle = useCallback(() => {
+        playbackManager.shuffle(item);
+    }, [item]);
+
+    return (
+        <IconButton
+            title={globalize.translate('Shuffle')}
+            className='paper-icon-button-light btnShuffle autoSize'
```
dmitrylyzo, line 21: Do `paper-icon-button-light` and `autoSize` work on MUI elements?

src/apps/experimental/features/details/components/buttons/SplitVersionsButton.tsx
```diff
@@ -0,0 +1,65 @@
...
+                    },
+                    {
+                        onSuccess: async () => {
+                            loading.hide();
+                            await queryClient.invalidateQueries({
+                                queryKey
+                            });
+                        },
+                        onError: (err: unknown) => {
+                            console.error(
```
dmitrylyzo, line 41: Shouldn't we hide loading when an error occurs?
Maybe also a `toast` with an error message so the user can notice it?

src/apps/experimental/features/details/hooks/api/useGetItemByType.ts
```diff
@@ -0,0 +1,63 @@
...
+import { useQuery } from '@tanstack/react-query';
+import { type JellyfinApiContext, useApi } from 'hooks/useApi';
+import type { ItemDto } from 'types/base/models/item-dto';
+import type{ NullableString } from 'types/base/common/shared/types';
+import { ItemKind } from 'types/base/models/item-kind';
+
+const getItemByType = async (
+    apiContext: JellyfinApiContext,
+    itemType: ItemKind,
+    itemId: NullableString,
```
dmitrylyzo, line 13: IMO, `itemId` shouldn't be nullable.

src/apps/experimental/features/details/components/buttons/ShuffleButton.tsx
grafixeyehero: updated wrong class name

src/apps/experimental/features/details/components/buttons/CancelSeriesTimerButton.tsx
```diff
@@ -0,0 +1,64 @@
...
+                    },
+                    {
+                        onSuccess: async () => {
+                            toast(globalize.translate('SeriesCancelled'));
+                            loading.hide();
+                            navigate('/livetv.html');
+                        },
+                        onError: (err: unknown) => {
+                            loading.hide();
+                            toast(globalize.translate('MessageCancelSeriesTimerError', err));
```
dmitrylyzo, line 43: No, no. The output to the console was fine. Show the toast in addition to it. Also, I'm not sure if `err` should be added to toast, as it can be complex (object).

src/apps/experimental/features/details/components/buttons/CancelTimerButton.tsx
```diff
@@ -0,0 +1,56 @@
...
+                    toast(globalize.translate('RecordingCancelled'));
+                    loading.hide();
+                    await queryClient.invalidateQueries({
+                        queryKey
+                    });
+                },
+
+                onError: (err: unknown) => {
+                    loading.hide();
+                    toast(globalize.translate('MessageCancelTimerError', err));
```
dmitrylyzo, line 39: Same.

src/apps/experimental/features/details/components/buttons/SplitVersionsButton.tsx
```diff
@@ -0,0 +1,64 @@
...
+                    {
+                        onSuccess: async () => {
+                            loading.hide();
+                            await queryClient.invalidateQueries({
+                                queryKey
+                            });
+                        },
+                        onError: (err: unknown) => {
+                            loading.hide();
+                            toast(globalize.translate('MessageSplitVersionsError', err));
```
dmitrylyzo, line 43: Same.

src/apps/experimental/features/details/hooks/api/useGetItemByType.ts
```diff
@@ -0,0 +1,61 @@
...
+
+const getItemByType = async (
+    apiContext: JellyfinApiContext,
+    itemType: ItemKind,
+    itemId: string,
+    options?: AxiosRequestConfig
+) => {
+    const { api, user } = apiContext;
+    if (!api) throw new Error('No API instance available');
+    if (!user?.Id) throw new Error('No User ID provided');
```
dmitrylyzo, line 17: I think the "not null" test (removed) could work for an empty string as well.

grafixeyehero: react query will handle it 
https://github.com/jellyfin/jellyfin-web/blob/b378c46ab56788d83751a8ac6200fa4e4686a132/src/apps/experimental/features/details/hooks/api/useGetItemByType.ts#L59

src/strings/en-us.json
```diff
@@ -1045,6 +1045,9 @@
     "MessageAreYouSureDeleteSubtitles": "Are you sure you wish to delete this subtitle file?",
     "MessageAreYouSureYouWishToRemoveMediaFolder": "Are you sure you wish to remove this media folder?",
     "MessageBrowsePluginCatalog": "Browse our plugin catalog to view available plugins.",
+    "MessageCancelSeriesTimerError": "An error occurred while cancel series timer",
+    "MessageCancelTimerError": "An error occurred while cancel timer",
+    "MessageSplitVersionsError": "An error occurred while split versions",
```
dmitrylyzo, line 1050: Move `MessageSplitVersionsError` down - in alphabetical order.

src/strings/en-us.json
```diff
@@ -1045,6 +1045,8 @@
     "MessageAreYouSureDeleteSubtitles": "Are you sure you wish to delete this subtitle file?",
     "MessageAreYouSureYouWishToRemoveMediaFolder": "Are you sure you wish to remove this media folder?",
     "MessageBrowsePluginCatalog": "Browse our plugin catalog to view available plugins.",
+    "MessageCancelSeriesTimerError": "An error occurred while cancel series timer",
```
thornbill, line 1048:
```suggestion
    "MessageCancelSeriesTimerError": "An error occurred while canceling the series timer",
```

src/strings/en-us.json
```diff
@@ -1045,6 +1045,8 @@
...
+    "MessageCancelTimerError": "An error occurred while cancel timer",
```
thornbill, line 1049:
```suggestion
    "MessageCancelTimerError": "An error occurred while canceling the timer",
```

src/strings/en-us.json
```diff
@@ -1097,6 +1099,7 @@
     "MessageRenameMediaFolder": "Renaming a media library will cause all metadata to be lost, proceed with caution.",
     "MessageRepositoryInstallDisclaimer": "WARNING: Installing a third party plugin repository carries risks. It may contain unstable or malicious code, and may change at any time. Only install repositories from authors that you trust.",
     "MessageSent": "Message sent.",
+    "MessageSplitVersionsError": "An error occurred while split versions",
```
thornbill, line 1102:
```suggestion
    "MessageSplitVersionsError": "An error occurred while splitting versions",
```

src/apps/experimental/features/details/hooks/api/useGetItemByType.ts
```diff
@@ -0,0 +1,61 @@
...
+import { ItemKind } from 'types/base/models/item-kind';
+
+const getItemByType = async (
+    apiContext: JellyfinApiContext,
+    itemType: ItemKind,
+    itemId: string,
+    options?: AxiosRequestConfig
+) => {
+    const { api, user } = apiContext;
+    if (!api) throw new Error('No API instance available');
```
thornbill, line 16: Even though it _shouldn't_ be possible with the enabled check, I have found I sometimes still encounter this while running webpack dev server. I have started just logging a warning instead of throwing, but maybe we should throw so we are forced to handle the error state properly. I'm curious what you think about this.

thornbill (changes requested):

src/apps/experimental/features/details/hooks/api/useGetItemByType.ts
grafixeyehero: can you give me more info 😟

thornbill: I've never been able to replicate it reliably, but here is an example of where I just logged a warning instead of threw an error: https://github.com/jellyfin/jellyfin-web/blob/master/src/apps/dashboard/features/activity/api/useLogEntries.ts#L14-L17

grafixeyehero: @thornbill we can use this approach too to display error on the page
```js
if (!api) return Promise.reject(new Error('[getItemByType] No API instance available'));
```

thornbill: That looks like a good option!

thornbill:
@grafixeyehero now that I am looking at this not on mobile wouldn't `throw` and `return Promise.reject` behave the same way for `async` functions? :sweat_smile:

grafixeyehero:
> @grafixeyehero now that I am looking at this not on mobile wouldn't `throw` and `return Promise.reject` behave the same way for `async` functions? 😅

I need to know more about the specific issue you’re facing

thornbill:
Sorry I think we should just revert that last change. I have seen issues with live reloading when running the webpack dev server where queries seem to run when they shouldn't be enabled, but after thinking about it more, if that becomes a recurring issue we should investigate a proper fix at that time instead of just silencing the error.

thornbill (approved):

PR #6043, opened 2024-09-09
Change encoding preset and play stats for server change

gnattu:
**Changes**

- Use correct "auto" string value for encoding presets because server is now using enum for it
- Hide the hardware acceleration type info in play stats because can be misleading. 

The hardware acceleration type currently only reflects the configuration in the dashboard, without accounting for potential software fallback and/or situations like QSV interop with VAAPI. This can result in displaying that a hardware encoder of certain type is in use when it's actually not. In 10.9 and earlier, we did not display this field due to an enum type mismatch. However, this issue has been fixed on the server side, causing the field to now appear, though with inaccurate data. We need to hide that until the server refactored the transcoding pipeline so that the actual encoder can be reported.

**Issues**

Changes made for jellyfin/jellyfin#12561

Fixes jellyfin/jellyfin#12630

### DIFF
--- a/src/components/playerstats/playerstats.js
+++ b/src/components/playerstats/playerstats.js
@@ -172,12 +172,17 @@ function getTranscodingStats(session, player, displayPlayMethod) {
                 value: session.TranscodingInfo.TranscodeReasons.map(translateReason).join('<br/>')
             });
         }
-        if (session.TranscodingInfo.HardwareAccelerationType) {
-            sessionStats.push({
-                label: globalize.translate('LabelHardwareEncoding'),
-                value: session.TranscodingInfo.HardwareAccelerationType
-            });
-        }
+        // Hide this for now because it is not useful in its current state.
+        // This only reflects the configuration in the dashboard, but the actual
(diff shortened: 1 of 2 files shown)


src/components/playerstats/playerstats.js
```diff
@@ -172,12 +172,17 @@ function getTranscodingStats(session, player, displayPlayMethod) {
...
+        // This only reflects the configuration in the dashboard, but the actual
+        // decoder/encoder selection is more complex. As a result, the hardware
+        // encoder may not be used even if hardware acceleration is configured,
+        // making the display of hardware acceleration misleading.
+        // if (session.TranscodingInfo.HardwareAccelerationType) {
+        //     sessionStats.push({
+        //         label: globalize.translate('LabelHardwareEncoding'),
+        //         value: session.TranscodingInfo.HardwareAccelerationType
+        //     });
+        // }
```
nyanmisaka, line 185: Actually you can just delete it. This field has little value to the user.

nyanmisaka (approved):

thornbill (approved):

PR #6060, opened 2024-09-11
Add support for cbt and cb7 books

thornbill:
**Changes**
Adds support for cbt and cb7 book files. libarchive.js supports both tar and 7zip formats so this should work the same as rar/zip currently do.

**Issues**
N/A


oddstr13 (approved):
Tested and working.
Server also needs support for cb7 and cbt files in the «Comic Book Zip Archive Cover Extractor». (Is that from the Bookshelf plugin?)

PR #6063, opened 2024-09-11
Securize CI, reusable workflows, simplifications

ferferga:
* Removed the bot's rebase command, since that feature is now integrated in GitHub.
* Use reusable workflows to kepp `pull_request_target`'s secrets safe and reduce duplication.
* Added dependency review action to ensure no vulnerable dependencies are introduced in the codebase
* Merge all scheduled tasks into a single file.


.github/workflows/schedule.yml
```diff
@@ -1,10 +1,10 @@
-name: Stale Check
+name: Scheduled tasks 🕑
 
 on:
   schedule:
     - cron: '30 1 * * *'
```
thornbill, line 5: The SDK update task should run later so that it runs after the SDK is built

.github/workflows/schedule.yml
```diff
@@ -1,10 +1,10 @@
...
   workflow_dispatch:
-
+
```
thornbill, line 7:
```suggestion

```

.github/workflows/schedule.yml
ferferga: Okay, so that's the reason for the timing, I didn't know about that.

I believe this is really fragile then (and also a future maintenance problem willing to happen). For these kind of things, I usually use this: https://github.com/marketplace/actions/trigger-workflow-and-wait (sso in SDK repo we trigger an update here).

Either way (and I believe you told me so at some point, but I forgot sorry! 😅) why isn't this with Renovate? This can be a good chance to review that.

thornbill: We were using the `unstable` tag in the package file (which is [not supported](https://github.com/renovatebot/renovate/issues/29303)  by renovate) but that was causing issues. It looks like we could switch to renovate with the follow tag option and overriding stability days.

thornbill: The switch to renovate is done now: https://github.com/jellyfin/jellyfin-web/pull/6156

thornbill (approved):

PR #6071, opened 2024-09-14
Separate bitrate control from resolution

gnattu:
**Changes**

- Remove all resolution related text from bitrate control as the server does not set resolution based on that. The actual field is kept as I'm unsure if that is used anywhere
- Only keep on entry of bitrate that is higher than original bitrate to allow using source bitrate, hide all other entries as the extremely high ones like 120Mbps is meaningless for 4Mbps source, and such option usually confuses our user.

**Issues**

A follow up of the server change: jellyfin/jellyfin#12644, but that not depends on that PR, this just works better with that merged.

### DIFF
--- a/src/components/qualityOptions.js
+++ b/src/components/qualityOptions.js
@@ -46,7 +46,7 @@ export function getVideoQualityOptions(options) {
         if (videoBitRate < 0 || (videoBitRate > 0 && c.bitrate < videoBitRate)) {
             qualityOptions.push(c);
         }
-    })
+    });
 
     if (maxStreamingBitrate) {
         let selectedIndex = qualityOptions.length - 1;
--- a/src/components/qualityOptions.js
+++ b/src/components/qualityOptions.js
@@ -1,3 +1,4 @@
+import minBy from 'lodash-es/minBy';
 import { appHost } from '../components/apphost';
 import globalize from '../lib/globalize';
 import appSettings from '../scripts/settings/appSettings';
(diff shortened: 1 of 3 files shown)


src/components/qualityOptions.js
```diff
@@ -30,31 +36,17 @@ export function getVideoQualityOptions(options) {
...
-    if (maxAllowedWidth >= 1260 && hostScreenWidth >= 650) {
-        qualityOptions.push({ name: '720p - 8 Mbps', maxHeight: 720, bitrate: 8000000 });
-        qualityOptions.push({ name: '720p - 6 Mbps', maxHeight: 720, bitrate: 6000000 });
-        qualityOptions.push({ name: '720p - 4 Mbps', maxHeight: 720, bitrate: 4000000 });
-    }
-    if (maxAllowedWidth >= 620) {
-        qualityOptions.push({ name: '480p - 3 Mbps', maxHeight: 480, bitrate: 3000000 });
-        qualityOptions.push({ name: '480p - 1.5 Mbps', maxHeight: 480, bitrate: 1500000 });
-        qualityOptions.push({ name: '480p - 720 kbps', maxHeight: 480, bitrate: 720000 });
+    if (videoBitRate > 0 && videoBitRate < 120000000) {
```
dmitrylyzo, line 39: Assuming the bitrates are sorted. This way you can change them without touching here.
```suggestion
    if (videoBitRate > 0 && videoBitRate < bitrateConfigurations[0].bitrate) {
```

src/components/qualityOptions.js
```diff
@@ -30,31 +36,17 @@ export function getVideoQualityOptions(options) {
...
+        // Push one entry that has higher limit than video bitrate to allow using source bitrate when Auto is also limited
+        const sourceOptions = minBy(bitrateConfigurations.filter((c) => c.bitrate > videoBitRate), (x) => x.bitrate);
+        qualityOptions.push(sourceOptions);
     }
 
-    qualityOptions.push({ name: '360p - 420 kbps', maxHeight: 360, bitrate: 420000 });
+    bitrateConfigurations.forEach((c) => {
+        if (videoBitRate < 0 || (videoBitRate > 0 && c.bitrate < videoBitRate)) {
```
dmitrylyzo, line 46: What about `videoBitRate = 0`?
Also, equal bitrate seems to be discarded.
```suggestion
        if (videoBitRate <= 0 || c.bitrate <= videoBitRate) {
```

src/components/qualityOptions.js
```diff
@@ -30,31 +36,17 @@ export function getVideoQualityOptions(options) {
...
-        qualityOptions.push({ name: '720p - 6 Mbps', maxHeight: 720, bitrate: 6000000 });
-        qualityOptions.push({ name: '720p - 4 Mbps', maxHeight: 720, bitrate: 4000000 });
-    }
-    if (maxAllowedWidth >= 620) {
-        qualityOptions.push({ name: '480p - 3 Mbps', maxHeight: 480, bitrate: 3000000 });
-        qualityOptions.push({ name: '480p - 1.5 Mbps', maxHeight: 480, bitrate: 1500000 });
-        qualityOptions.push({ name: '480p - 720 kbps', maxHeight: 480, bitrate: 720000 });
+    if (videoBitRate > 0 && videoBitRate < bitrateConfigurations[0].bitrate) {
+        // Push one entry that has higher limit than video bitrate to allow using source bitrate when Auto is also limited
+        const sourceOptions = minBy(bitrateConfigurations.filter((c) => c.bitrate > videoBitRate), (x) => x.bitrate);
```
dmitrylyzo, line 41: Btw, since they are sorted, we can just take the last one from the filtered array instead of using `minBy`.

dmitrylyzo (commented):
Seems to work.

1. Assume we have HEVC/AV1 and have to transcode to H264. The next higher bitrate may not be enough for comparable quality.

2. When we select some bitrate (say 15Mbps) and then start another video with a lower bitrate (say 4Mbps), it puts a checkmark on `Auto`. It worked the same way before, though. :man_shrugging:

gnattu:
> When we select some bitrate (say 15Mbps) and then start another video with a lower bitrate (say 4Mbps), it puts a checkmark on Auto. It worked the same way before, though. 🤷‍♂️

I noticed this as well but I decided to leave it the same way as before because it always has been like this for a few years. If we want to change this we need to think about what way is better and change that accordingly which is out of scope of this PR.

> Assume we have HEVC/AV1 and have to transcode to H264. The next higher bitrate may not be enough for comparable quality.

Or do we port the scaling function into the web as well and also scale based on that?

gnattu:
> Hmm, no - we limit it with the currently selected resolution. But using a second higher bitrate is still an option.

As a simplified option we can just display the option that is first higher than 1.5x of the source bitrate when we are doing hevc/av1/vp9 to h264. This is a rough estimate but should work good enough for most of them because it is the estimated compression efficiency in common bitrate range (8-10mbit/s)

gnattu:
I made the tradeoff option: increase the bitrate used to filter the control options by a factor of 1.5 when it is not too high (below the 20Mbps option). Ideally we only have to do this when transcoding to h264, but unfortunately we need extra api call to get this kind of info and it is not ideal to me. Maybe we can improve the API in future releases but I think this is okay for now.

dmitrylyzo (approved):

PR #6075, opened 2024-09-16
Redesign library navigation in experimental layout

thornbill:
**Changes**
Still in progress, just pushing up so people can try it out

**Issues**
N/A


sjorge:
Having the back button appear also interferes with muscle memory, you're used to hitting spot X and then picking library Y from the drop down. Now you want the other library, spot X might be X+1 which is annoying.

PR #6080, opened 2024-09-17
Discard `chrome` and `safari` on Tizen and webOS

dmitrylyzo:
We consider `safari` to be an actual Safari browser and `chrome` to be an actual Chrome browser.
Tizen and webOS may be "Safari" (WebKit) or "Chrome", but their capabilities are different from Safari and Chrome.

When testing https://github.com/jellyfin/jellyfin-web/pull/6073 webOS 1.2 (WebKit) tried to play TS+H264 using HLS.js.
Here we should check the transcoding codec, but we check the codec of the original media source:
https://github.com/jellyfin/jellyfin-web/blob/e5df4dd56bc180dfa24a52a99c718459a4074d56/src/components/htmlMediaHelper.js#L35

Although the use of HLS.js is caused by another bug, the mentioned inconsistency can cause problems when generating a device profile.

**Changes**
Discard `chrome` and `safari` on Tizen and webOS.

**Issues**
N/A

Alternatively, we could add `&& !browser.web0s && !browser.tizen` everywhere, but that's too wordy.

**Notes**
Chromecast will be disabled on Tizen and webOS.
https://github.com/jellyfin/jellyfin-web/blob/0ae080a15065cba2ab134e5d9d3faf6213b6c2ef/src/index.jsx#L150-L152

### DIFF
--- a/src/scripts/browser.js
+++ b/src/scripts/browser.js
@@ -293,12 +293,17 @@ browser.edgeUwp = browser.edge && (userAgent.toLowerCase().indexOf('msapphost') 
 
 if (browser.web0s) {
     browser.web0sVersion = web0sVersion(browser);
-} else if (browser.tizen) {
-    // UserAgent string contains 'Safari' and 'safari' is set by matched browser, but we only want 'tizen' to be true
-    delete browser.safari;
 
+    // UserAgent string contains 'Chrome' and 'Safari', but we only want 'web0s' to be true
+    delete browser.chrome;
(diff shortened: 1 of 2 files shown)


gnattu:
> Here we should check the transcoding codec, but we check the codec of the original media source:

The problem is that we don't know the actual transcoding codec until the client loads the HLS stream and play it and we need a player to load HLS first.

thornbill (approved):

PR #6082, opened 2024-09-17
Update mixed and collection library icons

thornbill:
**Changes**
* Updates the icons for mixed libraries and collections

| New | Old |
|---|---|
| Screenshot 2024-09-17 at 12-39-00 Jellyfin | Screenshot 2024-09-17 at 12-51-30 Jellyfin |

**Issues**
N/A

### DIFF
--- a/src/components/common/DefaultIconText.tsx
+++ b/src/components/common/DefaultIconText.tsx
@@ -1,7 +1,7 @@
 import { BaseItemKind } from '@jellyfin/sdk/lib/generated-client/models/base-item-kind';
 import React, { type FC } from 'react';
 import Icon from '@mui/material/Icon';
-import imageHelper from 'utils/image';
+import { getItemTypeIcon, getLibraryIcon } from 'utils/image';
 import DefaultName from './DefaultName';
 import type { ItemDto } from 'types/base/models/item-dto';
 
@@ -14,38 +14,24 @@ const DefaultIconText: FC<DefaultIconTextProps> = ({
     item,
     defaultCardImageIcon
 }) => {
-    if (item.CollectionType) {
-        return (
-            <Icon
-                className='cardImageIcon'
-                sx={{ color: 'inherit', fontSize: '5em' }}
-                aria-hidden='true'
-            >
-                {imageHelper.getLibraryIcon(item.CollectionType)}
-            </Icon>
-        );
+    let icon;
+
+    if (item.Type === BaseItemKind.CollectionFolder || item.CollectionType) {
+        icon = getLibraryIcon(item.CollectionType);
     }
 
-    if (item.Type && !(item.Type === BaseItemKind.TvChannel || item.Type === BaseItemKind.Studio )) {
-        return (
-            <Icon
-                className='cardImageIcon'
-                sx={{ color: 'inherit', fontSize: '5em' }}
-                aria-hidden='true'
-            >
-                {imageHelper.getItemTypeIcon(item.Type)}
-            </Icon>
-        );
+    if (!icon) {
+        icon = getItemTypeIcon(item.Type, defaultCardImageIcon);
     }
 
-    if (defaultCardImageIcon) {
+    if (icon) {
         return (
             <Icon
                 className='cardImageIcon'
                 sx={{ color: 'inherit', fontSize: '5em' }}
                 aria-hidden='true'
             >
-                {defaultCardImageIcon}
+                {icon}
             </Icon>
         );
     }
--- a/src/utils/image.ts
+++ b/src/utils/image.ts
@@ -1,3 +1,4 @@
+import { CollectionType } from '@jellyfin/sdk/lib/generated-client/models/collection-type';
 import { BaseItemKind } from '@jellyfin/sdk/lib/generated-client/models/base-item-kind';
 import type { DeviceInfo } from '@jellyfin/sdk/lib/generated-client/models/device-info';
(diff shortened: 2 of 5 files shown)


dmitrylyzo:
Experimental only?
https://github.com/jellyfin/jellyfin-web/blob/1564c93fbc2ccdab06c77b18114f274e7eff5f1b/src/utils/image.ts#L78
https://github.com/jellyfin/jellyfin-web/blob/1564c93fbc2ccdab06c77b18114f274e7eff5f1b/src/utils/image.ts#L107
https://github.com/jellyfin/jellyfin-web/blob/1564c93fbc2ccdab06c77b18114f274e7eff5f1b/src/components/cardbuilder/cardBuilder.js#L1182

thornbill:
Good call @dmitrylyzo! They should be consistent now.

src/utils/image.ts
```diff
@@ -75,36 +76,39 @@ export function getDeviceIcon(info: DeviceInfo | SessionInfo) {
...
+        case CollectionType.Photos:
+            return 'photo';
+        case CollectionType.Livetv:
             return 'live_tv';
-        case 'tvshows':
+        case CollectionType.Tvshows:
             return 'tv';
-        case 'trailers':
+        case CollectionType.Trailers:
             return 'local_movies';
```
dmitrylyzo, line 93: They look the same, but for consistency.
```suggestion
            return 'theaters';
```

src/utils/image.ts
```diff
@@ -75,36 +76,39 @@ export function getDeviceIcon(info: DeviceInfo | SessionInfo) {
...
+        case CollectionType.Musicvideos:
             return 'music_video';
-        case 'books':
-            return 'library_books';
+        case CollectionType.Books:
+            return 'book';
+        case CollectionType.Boxsets:
+            return 'video_library';
+        case CollectionType.Playlists:
+            return 'library_add';
```
dmitrylyzo, line 101: They look the same, but for consistency.
```suggestion
            return 'queue';
```

src/utils/image.ts
```diff
@@ -125,15 +129,15 @@ export function getItemTypeIcon(itemType: BaseItemKind | string) {
         case BaseItemKind.Folder:
             return 'folder';
         case BaseItemKind.BoxSet:
-            return 'collections';
+            return 'video_library';
         case BaseItemKind.Playlist:
-            return 'view_list';
+            return 'library_add';
```
dmitrylyzo, line 134:
```suggestion
            return 'queue';
```

src/utils/image.ts
```diff
@@ -125,15 +129,15 @@ export function getItemTypeIcon(itemType: BaseItemKind | string) {
...
         case BaseItemKind.Photo:
             return 'photo';
         case BaseItemKind.PhotoAlbum:
             return 'photo_album';
         default:
-            return 'folder';
+            return defaultIcon || 'folder';
```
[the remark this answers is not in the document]
thornbill: I should just update that to follow the same logic as my changes in cardbuilder :upside_down_face:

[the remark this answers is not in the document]
thornbill: Should be fixed now and added tests

src/components/common/DefaultIconText.tsx
```diff
@@ -14,38 +14,24 @@ const DefaultIconText: FC<DefaultIconTextProps> = ({
...
-                {imageHelper.getLibraryIcon(item.CollectionType)}
-            </Icon>
-        );
+    let icon;
+
+    if (item.Type === BaseItemKind.CollectionFolder || item.CollectionType) {
+        icon = getLibraryIcon(item.CollectionType);
     }
 
-    if (item.Type && !(item.Type === BaseItemKind.TvChannel || item.Type === BaseItemKind.Studio )) {
```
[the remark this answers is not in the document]
thornbill: There should be no impact. There isn't an explicit icon provided for these types so it will fallback to the default icon (if present) which should match the previous behavior.

src/components/common/DefaultIconText.tsx
```diff
@@ -14,38 +14,24 @@ const DefaultIconText: FC<DefaultIconTextProps> = ({
...
-                className='cardImageIcon'
-                sx={{ color: 'inherit', fontSize: '5em' }}
-                aria-hidden='true'
-            >
-                {imageHelper.getLibraryIcon(item.CollectionType)}
-            </Icon>
-        );
+    let icon;
+
+    if (item.Type === BaseItemKind.CollectionFolder || item.CollectionType) {
```
grafixeyehero, line 19: Recordings item should be a folder instead of quiz. The Recordings item type value is `CollectionFolder` but the `CollectionType` is undefined

src/utils/image.ts
```diff
@@ -125,15 +129,17 @@ export function getItemTypeIcon(itemType: BaseItemKind | string) {
...
-            return 'collections';
+            return 'video_library';
         case BaseItemKind.Playlist:
-            return 'view_list';
+            return 'queue';
         case BaseItemKind.Photo:
             return 'photo';
         case BaseItemKind.PhotoAlbum:
             return 'photo_album';
+        case undefined:
```
grafixeyehero, line 139: Chapters use the default icon that pass form the card opts but chapters does not have item type so it considered undefined.

src/components/common/DefaultIconText.tsx
thornbill: It is also a type of mixed library so I think using the same icon is fine :man_shrugging:

grafixeyehero (approved):

PR #6090, opened 2024-09-20
Allow Dolby Vision fallback layer on Tizen 3+

dmitrylyzo:
**Changes**
Allow Dolby Vision fallback layer on Tizen 3+.

**Issues**
N/A

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -1107,16 +1107,28 @@ export default function (options) {
     let vp9VideoRangeTypes = 'SDR';
     let av1VideoRangeTypes = 'SDR';
 
+    if (browser.tizenVersion >= 3) {
+        hevcVideoRangeTypes += '|DOVIWithSDR';
+    }
+
     if (supportsHdr10(options)) {
         hevcVideoRangeTypes += '|HDR10';
         vp9VideoRangeTypes += '|HDR10';
         av1VideoRangeTypes += '|HDR10';
+
+        if (browser.tizenVersion >= 3) {
+            hevcVideoRangeTypes += '|DOVIWithHDR10';
+        }
     }
 
     if (supportsHlg(options)) {
         hevcVideoRangeTypes += '|HLG';
         vp9VideoRangeTypes += '|HLG';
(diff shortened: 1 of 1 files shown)


gnattu:
Does directstream/remux still work for dolby vision in this case? Just adding profile 8 will cause ffmpeg to tag the video with `dvh1` fourcc instead of `dvh1` fourcc. If Tizen can handle fallback even in this case then this would be safe enough, otherwise we have to find a way to only add those for direct play, but remove those for remux and direct stream.

dmitrylyzo:
> Does directstream/remux still work for dolby vision in this case? Just adding profile 8 will cause ffmpeg to tag the video with `dvh1` fourcc instead of `dvh1` fourcc. If Tizen can handle fallback even in this case then this would be safe enough, otherwise we have to find a way to only add those for direct play, but remove those for remux and direct stream.

Retested with "Stereo only": Direct streaming, playing fine.
Also, remuxed to MKV - also playing fine.

The only thing: on LT27H390 (Tizen 3) the colors are more "yellowish", on UE65NU7400 (Tizen 4) they are "reddish". I was testing `P81_GlassBlowing2_1920x1080%4059.94fps_15200kbps_fmp4` sample. "Reddish" are looking better/natural, imo.

:man_facepalming: LT27H390 (Tizen 3) isn't HDR Ready. This can be fixed on the app side.

Question. Can we just do this?
```js
    if (supportsHdr10(options)) {
        hevcVideoRangeTypes += '|HDR10|DOVIWithHDR10';
        vp9VideoRangeTypes += '|HDR10';
        av1VideoRangeTypes += '|HDR10';
    }

    if (supportsHlg(options)) {
        hevcVideoRangeTypes += '|HLG|DOVIWithHLG';
        vp9VideoRangeTypes += '|HLG';
        av1VideoRangeTypes += '|HLG';
    }
```

gnattu:
> Question. Can we just do this?

If only for Tizen then yes, for other devices no because not all devices accepts the `dvh1` fourcc tagged by ffmpeg and will not play the fallback layer at all.

thornbill (approved):

Issue #3732, opened 2022-06-29
Shuffle button functionality currently shuffles the whole collection/library, ignoring filters


davidmurdoch:
https://github.com/jellyfin/jellyfin-web/pull/3627 added a shuffle button to the Movies libraries (yay!), but it doesn't take the filters into account.

**Steps To Reproduce**

1. Go to Movies
2. Add some filters.
3. Click on the Shuffle button
4. You might get a movie that is not included in the filter
5. Be sad

**Expected Behavior**
Shuffle should use the filters. It shouldn't make the user sad (unless the movie is a sad movie).

PR #6111, opened 2024-09-24
Make movie's shuffle button use the current filters

davidmurdoch:
**Changes**

Updates the random movies button so it chooses a random movie from the _filtered_ list, instead of choosing a random movie from all movies.

Note: I do think this was the original intent of the movie's Shuffle button, as the button is removed from the UI when the applied filters return 0 movies.

**Issues**
Fixes [#3732](https://github.com/jellyfin/jellyfin-web/issues/3732)

### DIFF
--- a/src/controllers/movies/movies.js
+++ b/src/controllers/movies/movies.js
@@ -36,6 +36,8 @@ export default function (view, params, tabContent, options) {
         const newQuery = { ...query, SortBy: 'Random', Limit: 1 };
         return ApiClient.getItems(ApiClient.getCurrentUserId(), newQuery).then(({ Items }) => {
             playbackManager.shuffle(Items[0]);
+        }).finally(() => {
+            isLoading = false;
         });
     }
 
--- a/src/controllers/movies/movies.js
+++ b/src/controllers/movies/movies.js
@@ -33,7 +33,7 @@ export default function (view, params, tabContent, options) {
     function shuffle() {
         isLoading = true;
         loading.show();
-        const newQuery = { ...query, SortBy: 'Random' };
+        const newQuery = { ...query, SortBy: 'Random', Limit: 1 };
         return ApiClient.getItems(ApiClient.getCurrentUserId(), newQuery).then(({ Items }) => {
             playbackManager.shuffle(Items[0]);
         });
--- a/src/controllers/movies/movies.js
+++ b/src/controllers/movies/movies.js
@@ -33,9 +33,12 @@ export default function (view, params, tabContent, options) {
     function shuffle() {
         isLoading = true;
         loading.show();
-        const newQuery = { ...query, SortBy: 'Random', Limit: 1 };
+        const newQuery = { ...query, SortBy: 'Random', StartIndex: 0, Limit: 300 };
         return ApiClient.getItems(ApiClient.getCurrentUserId(), newQuery).then(({ Items }) => {
-            playbackManager.shuffle(Items[0]);
+            playbackManager.play({
+                items: Items,
+                autoplay: true
+            });
         }).finally(() => {
             isLoading = false;
         });
--- a/src/controllers/movies/movies.js
+++ b/src/controllers/movies/movies.js
@@ -31,11 +31,11 @@ export default function (view, params, tabContent, options) {
     }
 
     function shuffle() {
-        ApiClient.getItem(
-            ApiClient.getCurrentUserId(),
-            params.topParentId
-        ).then((item) => {
-            playbackManager.shuffle(item);
+        isLoading = true;
+        loading.show();
+        const newQuery = { ...query, SortBy: 'Random' };
+        return ApiClient.getItems(ApiClient.getCurrentUserId(), newQuery).then(({ Items }) => {
+            playbackManager.shuffle(Items[0]);
         });
     }
 


src/controllers/movies/movies.js
```diff
@@ -30,11 +30,11 @@ export default function (view, params, tabContent, options) {
...
-            ApiClient.getCurrentUserId(),
-            params.topParentId
-        ).then((item) => {
-            playbackManager.shuffle(item);
+        isLoading = true;
+        loading.show();
+        const newQuery = { ...query, SortBy: 'Random', Limit: 1 };
+        return ApiClient.getItems(ApiClient.getCurrentUserId(), newQuery).then(({ Items }) => {
+            playbackManager.shuffle(Items[0]);
         });
```
dmitrylyzo, line 38: _I haven't tested - reviewing from mobile._
Shouldn't we reset `isLoading`? Otherwise, I believe the prev/next buttons will be blocked.
```suggestion
        }).finally(() => {
            isLoading = false;
        });
```
Logically, we should call `loading.hide()` as well, but this might hide the one from PlaybackManager.

davidmurdoch: I copied the pattern from the function right above this one. 

```
function fetchData() {
        isLoading = true;
        loading.show();
        return ApiClient.getItems(ApiClient.getCurrentUserId(), query);
    }
```


The changes I made here to set the loading flag and show the loading element aren't necessary to fix the filtered shuffle feature. It was just a "drive by" improvement, as it is nice to see some immediate indication that the shuffle button was clicked.

I can remove these unrelated changes if the PR is better that way.

dmitrylyzo: `fetchData` is used to refresh the item list, and `isLoading = true` - to block intermediate clicking on prev/next buttons.
Then `afterRefresh` is called and resets `isLoading`: https://github.com/jellyfin/jellyfin-web/blob/a419aa9e89b6991fe4fc652abe2fafba166998bf/src/controllers/movies/movies.js#L91

Just try:
1. Shuffle movies.
2. Back from the playback.
3. Click prev/next button.

dmitrylyzo:
Could you also rebase your branch onto recent `master` and force-push?

src/controllers/movies/movies.js
davidmurdoch: Confirmed that this would introduce a bug with the prev/next buttons.  I'll make your suggested changes soon.

davidmurdoch:
> Could you also rebase your branch onto recent `master` and force-push?

Done. Let me know if you need the `loading.hide()` added in the finally block, too (or feel free to commit and merge to this branch yourself!)

dmitrylyzo (commented):
Tested it a bit more.
Previously, it randomly played the entire Movie library (not just one random movie).
Shuffle button on the list page seems to be implemented this way:
https://github.com/jellyfin/jellyfin-web/blob/97d4f94d6f2fdf44306e1563dde510873d9dccea/src/controllers/list.js#L775-L788

src/controllers/movies/movies.js
```diff
@@ -31,11 +31,13 @@ export default function (view, params, tabContent, options) {
...
-        ApiClient.getItem(
-            ApiClient.getCurrentUserId(),
-            params.topParentId
-        ).then((item) => {
-            playbackManager.shuffle(item);
+        isLoading = true;
+        loading.show();
+        const newQuery = { ...query, SortBy: 'Random', Limit: 1 };
+        return ApiClient.getItems(ApiClient.getCurrentUserId(), newQuery).then(({ Items }) => {
+            playbackManager.shuffle(Items[0]);
```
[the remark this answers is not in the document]
davidmurdoch: Oh interesting. I had thought or noticed that it plays _all_ movies. This makes sense for songs, and maybe tv, but I don't think it makes sense for movies.

I was thinking it should be more of an "I'm feeling lucky" button than a "play everything in a random order" button.

I actually like "play everything in a random order" functionality _more_ (as long as it applies filters, of course).

I have no idea how to determine if `300` is a reasonable `Limit`, so I'm just going to merge your suggestion and hope it is. :-)

davidmurdoch:
@dmitrylyzo  I merged your latest suggested changes. Do we still need the `finally` block added in a previous commit?

dmitrylyzo:
> Do we still need the `finally` block added in a previous commit?

Yes, because `isLoading = true` is still there and we need to reset it.

src/controllers/movies/movies.js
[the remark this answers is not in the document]
dmitrylyzo: > I have no idea how to determine if `300` is a reasonable `Limit`

`300` is some arbitrary, compromise value that is used in the playback manager itself and in the list.
https://github.com/jellyfin/jellyfin-web/blob/97d4f94d6f2fdf44306e1563dde510873d9dccea/src/components/playback/playbackmanager.js#L136
https://github.com/jellyfin/jellyfin-web/blob/97d4f94d6f2fdf44306e1563dde510873d9dccea/src/controllers/list.js#L781

dmitrylyzo (approved):

src/controllers/movies/movies.js
```diff
@@ -31,11 +31,16 @@ export default function (view, params, tabContent, options) {
...
     function shuffle() {
-        ApiClient.getItem(
-            ApiClient.getCurrentUserId(),
-            params.topParentId
-        ).then((item) => {
-            playbackManager.shuffle(item);
+        isLoading = true;
+        loading.show();
+        const newQuery = { ...query, SortBy: 'Random', StartIndex: 0, Limit: 300 };
+        return ApiClient.getItems(ApiClient.getCurrentUserId(), newQuery).then(({ Items }) => {
```
thornbill, line 37: We should probably use the `getItems` utility function: https://github.com/jellyfin/jellyfin-web/blob/master/src/utils/jellyfin-apiclient/getItems.ts#L67

dmitrylyzo: It is mostly for long `Ids`. Maybe only for consistency. 🤷‍♂️

thornbill: Ah I missed the limit being set so this is fine. 👍👍

PR #6121, opened 2024-09-26
Keybind C to auto select subtitle track

BakerAO:
Keybind C to auto select subtitle track

**Changes**
Add `c` case to `onKeyDown`. The function selects the first available subtitle track, or if subtitles are already selected, disables them. 

**Issues**
Feature Request: https://features.jellyfin.org/posts/1245/enable-disable-subtitles-using-a-keyboard-shortcut

### DIFF
--- a/src/controllers/playback/video/index.js
+++ b/src/controllers/playback/video/index.js
@@ -1431,7 +1431,7 @@ export default function (view) {
             case 'C':
                 if (!e.shiftKey) {
                     e.preventDefault();
-                    autoSelectSubtitleTrack();
+                    toggleDefaultSubtitleTrack();
                 }
                 break;
             case 'g':
--- a/src/controllers/playback/video/index.js
+++ b/src/controllers/playback/video/index.js
@@ -1204,8 +1204,8 @@ export default function (view) {
         const currentIndex = playbackManager.getSubtitleStreamIndex(player) ?? -1;
         const defaultIndex = player?.streamInfo?.mediaSource?.DefaultSubtitleStreamIndex ?? -1;
 
-        if (currentIndex === -1 && streams.length > 0) {
-            const subtitleIndex = defaultIndex != -1 ? defaultIndex : streams[0].Index;
+        if (currentIndex === -1 && subtitleTracks.length > 0) {
+            const subtitleIndex = defaultIndex != -1 ? defaultIndex : subtitleTracks[0].Index;
             playbackManager.setSubtitleStreamIndex(subtitleIndex, player);
         } else {
             playbackManager.setSubtitleStreamIndex(-1, player);
--- a/src/controllers/playback/video/index.js
+++ b/src/controllers/playback/video/index.js
@@ -1200,17 +1200,20 @@ export default function (view) {
 
     function autoSelectSubtitleTrack() {
         const player = currentPlayer;
-        const streams = playbackManager.subtitleTracks(player);
+        const streams = playbackManager.subtitleTracks(player) ?? [];
         const currentIndex = playbackManager.getSubtitleStreamIndex(player) ?? -1;
+        const defaultIndex = player?.streamInfo?.mediaSource?.DefaultSubtitleStreamIndex ?? -1;
 
         if (currentIndex === -1 && streams.length > 0) {
-            const firstSubtitleIndex = streams[0].Index;
-            playbackManager.setSubtitleStreamIndex(firstSubtitleIndex, player);
+            if (defaultIndex > -1) {
+                playbackManager.setSubtitleStreamIndex(defaultIndex, player);
+            } else {
+                const firstSubtitleIndex = streams[0].Index;
+                playbackManager.setSubtitleStreamIndex(firstSubtitleIndex, player);
+            }
         } else {
             playbackManager.setSubtitleStreamIndex(-1, player);
         }
-
-        toggleSubtitleSync();
     }
 
     /**
--- a/src/controllers/playback/video/index.js
+++ b/src/controllers/playback/video/index.js
@@ -1200,7 +1200,7 @@ export default function (view) {
 
     function autoSelectSubtitleTrack() {
         const player = currentPlayer;
-        const streams = playbackManager.subtitleTracks(player) ?? [];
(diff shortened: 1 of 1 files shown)


viown:
I'm not sure if the user always expects the first subtitle track to be selected as they may have different subtitle preferences.

Maybe you can use `currentPlayer.streamInfo.mediaSource.DefaultSubtitleStreamIndex` and fallback to `streams[0].Index` if it's -1. Or, alternatively, something like `DefaultSubtitleStreamIndex` -> `Subtitle Track with 'default' flag on` -> `streams[0].Index`.

You could also take advantage of the user's subtitle language preference, but that may be overcomplicating things as `DefaultSubtitleStreamIndex` should already point to the 'best' track unless it was manually changed or turned off.

BakerAO:
Thank you for the feedback, that makes sense. I like the idea of using `DefaultSubtitleStreamIndex`. I imagine most users don't want to cycle through the options (like English to French to Italian, etc), typically they just want them on or off in their language, but that would work as well.

viown:
Good point, I didn't consider that.

BakerAO:
@viown and @Shadowghost, updated to use the default index if its set

src/controllers/playback/video/index.js
```diff
@@ -1176,6 +1176,24 @@ export default function (view) {
...
+        const currentIndex = playbackManager.getSubtitleStreamIndex(player) ?? -1;
+        const defaultIndex = player?.streamInfo?.mediaSource?.DefaultSubtitleStreamIndex ?? -1;
+
+        if (currentIndex === -1 && streams.length > 0) {
+            if (defaultIndex > -1) {
+                playbackManager.setSubtitleStreamIndex(defaultIndex, player);
+            } else {
+                const firstSubtitleIndex = streams[0].Index;
+                playbackManager.setSubtitleStreamIndex(firstSubtitleIndex, player);
+            }
```
viown, line 1191:
```suggestion
            const subtitleIndex = defaultIndex != -1 ? defaultIndex : streams[0].Index;
            playbackManager.setSubtitleStreamIndex(subtitleIndex, player);
```

BakerAO:
Commited your change @viown, that made sense

src/controllers/playback/video/index.js
```diff
@@ -1176,6 +1176,20 @@ export default function (view) {
         }
     }
 
+    function autoSelectSubtitleTrack() {
+        const player = currentPlayer;
+        const streams = playbackManager.subtitleTracks(player) ?? [];
+        const currentIndex = playbackManager.getSubtitleStreamIndex(player) ?? -1;
```
viown, line 1182:
```suggestion
        const currentIndex = playbackManager.getSubtitleStreamIndex(player);
```

BakerAO: i found that `getSubtitleStreamIndex` can be `undefined` on first load with no subtitle selected

BakerAO: i could change the `if` statement to include undefined if the nullish coalescence generates excess code

viown: > i found that `getSubtitleStreamIndex` can be `undefined` on first load with no subtitle selected

Hmm, you're right. This does sound like a bug. Looks like it's coming from here: https://github.com/jellyfin/jellyfin-web/blob/master/src/components/playback/playbackmanager.js#L2624

You can probably just leave it as-is for now. `getSubtitleStreamIndex` can be null, like for e.g, when playing a trailer, but it wouldn't matter in your case since it'll lead to a no-op.

BakerAO: @viown Sounds good. Im not familiar with the PR culture here, do you want me to `Resolve conversation`?

BakerAO:
@viown does this PR still look good to you? let me know if you want any changes

src/controllers/playback/video/index.js
```diff
@@ -1184,6 +1184,20 @@ export default function (view) {
         }
     }
 
+    function autoSelectSubtitleTrack() {
+        const player = currentPlayer;
```
thornbill, line 1188: We should pass the player as a function parameter to be consistent with the other functions.

src/controllers/playback/video/index.js
```diff
@@ -1184,6 +1184,20 @@ export default function (view) {
...
+        const streams = playbackManager.subtitleTracks(player);
```
thornbill, line 1189:
```suggestion
        const subtitleTracks = playbackManager.subtitleTracks(player);
```

src/controllers/playback/video/index.js
```diff
@@ -1184,6 +1184,20 @@ export default function (view) {
         }
     }
 
+    function autoSelectSubtitleTrack() {
```
thornbill, line 1187:
```suggestion
    function toggleDefaultSubtitleTrack() {
```

thornbill (commented):
Looks good overall! I just had some minor feedback.

src/controllers/playback/video/index.js
BakerAO: hello @thornbill , thanks for reviewing, i merged your other suggestions, but wasn't sure what you meant for this one. Excuse my lack of knowledge here, but to me it looked like the other functions are also referencing the `currentPlayer` like here:
https://github.com/jellyfin/jellyfin-web/blob/9d7475e4e8fc809fde9afe68f1abb94a1745b314/src/controllers/playback/video/index.js#L1186
https://github.com/jellyfin/jellyfin-web/blob/9d7475e4e8fc809fde9afe68f1abb94a1745b314/src/controllers/playback/video/index.js#L960
https://github.com/jellyfin/jellyfin-web/blob/9d7475e4e8fc809fde9afe68f1abb94a1745b314/src/controllers/playback/video/index.js#L990
Do I need to use the `bindToPlayer` function?
Do you want me to change the argument to be `toggleDefaultSubtitleTrack(player)`?
Thank you!

BakerAO: Disregard the previous comment, I looked again at the other key bound functions and passed the `currentPlayer` as an argument

PR #6137, opened 2024-09-29
remove dependency on `@uupaa/dynamic-import-polyfill`

revam:
**Changes**

Removed the dependency on `@uupaa/dynamic-import-polyfill` and switched to using raw `import()`, because the polyfill were having issues with `async`/`await` in plugin modules which were using `await` at the top-level in the module to e.g. do dynamic imports because plugins don't have a base-point to do static imports of other modules.

**Issues**

-

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -82,7 +82,6 @@
         "@types/sortablejs": "1.15.8",
         "@typescript-eslint/eslint-plugin": "5.62.0",
         "@typescript-eslint/parser": "5.62.0",
-        "@uupaa/dynamic-import-polyfill": "1.0.2",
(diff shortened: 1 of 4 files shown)


revam:
The `es-check` error seems to not want the raw `import` in the bundle, even though it is used intentionally as a keyword in this PR. No idea what to do to suppress it for that single use case.

Gauvino:
Please, change your base branch to master, I think no more PR will be merged to 10.9 since it's freeze now

revam:
Moved to `master`. Still no idea how to fix es-check to allow an exception for this change.

PR #6142, opened 2024-10-01
Only add gain node and AudioContext when normalization enabled

gnattu:
Currently, we create a Web Audio session and connect to it regardless of the normalization status, hoping it will work. However, it turns out that some devices exhibit buggy behavior when the Web Audio session is connected to a gain node, causing distorted audio playback. This PR makes that the gain element is only created when normalization is not off. The downside is that users must manually stop the currently playing song and start a new one to apply the state change between on and off. By doing this, users experiencing bugged playback can turn off the normalization so that the playback behavior would be the same as 10.8 which *should* be fine.

We do need this workaround because:

- Device and/or OS vendors are less concerned about this issue than we are, meaning it might take a long time to be fixed upstream.
- Not all devices have the option to update to the latest software. We need to provide users with such devices a workaround so that they can at least have a normal playback, even if it means losing a feature.

**Changes**

- Only create gain element when normalization is not off.

**Issues**

See jellyfin/jellyfin#11614, hopefully will workaround that.

### DIFF
--- a/src/plugins/htmlAudioPlayer/plugin.js
+++ b/src/plugins/htmlAudioPlayer/plugin.js
@@ -128,6 +128,7 @@ class HtmlAudioPlayer {
 
                 if (!self.gainNode) {
                     addGainElement(elem);
+                    if (!self.gainNode) return;
                 }
 
                 if (normalizationGain) {
--- a/src/plugins/htmlAudioPlayer/plugin.js
+++ b/src/plugins/htmlAudioPlayer/plugin.js
@@ -121,6 +121,13 @@ class HtmlAudioPlayer {
                     normalizationGain =
                         options.mediaSource.albumNormalizationGain
                         ?? options.item.NormalizationGain;
+                } else {
+                    console.debug('normalization disabled')
+                    return;
+                }
+
+                if (!self.gainNode) {
+                    addGainElement(elem);
                 }
 
                 if (normalizationGain) {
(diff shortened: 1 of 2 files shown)


src/plugins/htmlAudioPlayer/plugin.js
```diff
@@ -121,6 +121,13 @@ class HtmlAudioPlayer {
                     normalizationGain =
                         options.mediaSource.albumNormalizationGain
                         ?? options.item.NormalizationGain;
+                } else {
+                    console.debug('normalization disabled');
+                    return;
+                }
+
+                if (!self.gainNode) {
+                    addGainElement(elem);
```
dmitrylyzo, line 130: Probably missing from the beginning.
```suggestion
                    addGainElement(elem);
                    if (!self.gainNode) return;
```
`addGainElement` already has an error message.

dmitrylyzo (approved):

ferferga (changes requested):
I wanted to comment this yesterday, but wanted to actually verify it, and I can confirm now that this doesn't fix the issue.

The problem start as soon as the AudioContext is connected to the media element, no other nodes are necessary. In Vue for instance, we use WebAudio for analyzing the audio waveform, but we don't apply any gain (or any other node that modifies the output), yet the error still happens.

In order to fix the underlying issue, WebAudio must be completely suspended.

gnattu:
But what this PR actually does is not to crate the `AudioContext` at all when normalization is off? So it makes me quite confused why that did not fix your problem.

thornbill (approved):

PR #6149, opened 2024-10-01
Add playback subscriber abstraction

thornbill:
**Changes**
Adds an abstraction for subscribing to playback events

NOTE: Nothing is actually using this currently, but it provides the basis for segment skipping (and a lot of existing code could be refactored to use it for simplicity). I did write a `LoggingSubscriber` for testing locally, but I'm not sure that would provide any real value to actually add.

**Issues**
N/A


nielsvanvelzen:
I could give some critique about some of the events but those are already part of the playback manager so not related to this PR.

Issue #6143, opened 2024-10-01
Changing dashboard page applies non-dashboard theme


nielsvanvelzen:
**Describe The Bug**
Opening the dashboard works fine, but changing the dashboard page once opened it will apply the non-dashboard theme.

**Steps To Reproduce**

1. Make sure to have a separate theme configured for the dashboard. In my case I have "Purple Haze" as default and "Dark" for the dashboard.
2. Open the dashboard
3. Note that the theme is correct
4. Open a different dashboard page from the sidebar
5. Notice the wrong theme is activated

**Expected Behavior**

The non-dashboard theme should not be applied when changing dashboard pages.

**Logs**

**Screenshots**

**System (please complete the following information):**
 - Platform: Windows
 - Browser: Firefox
 - Jellyfin Version: 10.10 2024093005

**Additional Context**

PR #6150, opened 2024-10-01
Only connect to ServerConnections once

viown:
Currently, the client will send a request to `ServerConnections.connect` every time we switch pages.  With this, the behaviour should now be the same as prior to app routing being removed.

**Changes**
Change it so that we only connect to ServerConnections once with each session, reducing unnecessary calls to `/System/Info/Public` with each navigation.
Fix issue with logging in after refreshing login/select server page.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-web/issues/6143

### DIFF
--- a/src/components/ServerConnections.js
+++ b/src/components/ServerConnections.js
@@ -33,6 +33,7 @@ class ServerConnections extends ConnectionManager {
     constructor() {
         super(...arguments);
         this.localApiClient = null;
+        this.firstConnection = null;
 
         // Set the apiclient minimum version to match the SDK
         this._minServerVersion = MINIMUM_VERSION;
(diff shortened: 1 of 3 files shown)


viown:
https://github.com/jellyfin/jellyfin-web/pull/6150/files#diff-c84866e0254b477bbe1e330af1825378d7c0e3dc5cf838f08cb4fb768eb70ba5R156

There is an issue where if you try to login after refreshing the login page it won't redirect you (this existed prior to this PR, even in 10.9), so might as well fix it here.

src/components/ServerConnections.js
```diff
@@ -33,6 +33,7 @@ class ServerConnections extends ConnectionManager {
     constructor() {
         super(...arguments);
         this.localApiClient = null;
+        this.firstConnection = null;
```
thornbill, line 36: Maybe we should use a react context??? :man_shrugging:

viown: This value is pretty much single-use (it's set to null after fetching), so it might not benefit much from being turned into a react context?

thornbill: :+1: I was just trying to think of a more "react" way of handling this, but it's fine.

thornbill (approved):

PR #6162, opened 2024-10-04
Fix missing navigational buttons in the Book Player top bar

### LINKED ISSUE #6161: Book Player top bar is missing navigation buttons
**Describe The Bug**

As suggested by line `214` and `215` of `src/plugins/bookPlayer.plugin.js`, the template should contain navigation buttons but does not.

**Steps To Reproduce**

Note: Reproduction requires the server to have a book library and for the viewer to be on a mobile device.
1. Go to Books
2. Select an individual Book
3. Hit the Play button
4. Observe the top bar does not contain a forward/next or backward/previous button

**Expected Behavior**

The top bar of the book player should contain forward and backward buttons so users can navigate the content they're viewing.

**Screenshots**

Untitled

**System**
 - Platform: Android
 - Browser: Jellyfin App (Chromium)
 - Jellyfin Version: 10.9.9

XeClutch:
**Changes**
Added in navigational buttons to the top bar of the Book Player

**Issues**
Fixes #6161

### DIFF
--- a/src/plugins/bookPlayer/template.html
+++ b/src/plugins/bookPlayer/template.html
@@ -2,12 +2,12 @@
     <button is="paper-icon-button-light" id="btnBookplayerToc" class="autoSize bookplayerButton hide-mouse-idle-tv" tabindex="-1">
         <span class="material-icons bookplayerButtonIcon toc" aria-hidden="true"></span>
     </button>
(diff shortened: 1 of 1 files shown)


thornbill (approved):
It does cause the header to wrap on smaller screens, but we can address that later.

PR #6165, opened 2024-10-05
Added option to control how long the screensaver takes to start

ethanarns:
**Changes**
- Added setting in Display Settings for time until screensaver, defaulting to 3 minutes
- Load the setting from Display Settings to change the time it takes for the screensaver to activate
- Reduced screensaver check time from 10 seconds to 5 seconds
- Added English translations for new settings

**Issues**


src/scripts/screensavermanager.js
```diff
@@ -11,7 +11,13 @@ import './screensavermanager.scss';
...
     // Returns the minimum amount of idle time required before the screen saver can be displayed
     //time units used Millisecond
-    return 180000;
+    const timeSeconds = userSettings.get('screensaverTime', false);
+    // If it is zero or undefined
+    if (!timeSeconds) {
+        // Defaults to 3 minutes
+        return 180000;
+    }
+    return timeSeconds * 1000;
```
dmitrylyzo, line 20: `screensaverTime` performs the necessary checks.
```suggestion
    return userSettings.screensaverTime() * 1000;
```

src/strings/en-gb.json
```diff
@@ -1769,6 +1769,8 @@
     "ListView": "List View",
     "LabelBackdropScreensaverInterval": "Backdrop Screensaver Interval",
     "LabelBackdropScreensaverIntervalHelp": "The time in seconds between different backdrops when using the backdrop screensaver.",
+    "LabelScreensaverTimeHelp": "The amount of time in seconds of inactivity required to start the screensaver.",
```
dmitrylyzo, line 1772: Don't add strings to the languages other than `en-us`. Only Weblate should be used to work with them.

src/strings/en-us.json
```diff
@@ -599,6 +599,8 @@
     "LabelBaseUrlHelp": "Add a custom subdirectory to the server URL. For example: <code>http://example.com/<b>&lt;baseurl&gt;</b></code>",
     "LabelBackdropScreensaverInterval": "Backdrop Screensaver Interval",
     "LabelBackdropScreensaverIntervalHelp": "The time in seconds between different backdrops when using the backdrop screensaver.",
+    "LabelScreensaverTimeHelp": "The amount of time in seconds of inactivity required to start the screensaver.",
+    "LabelScreensaverTime": "Screensaver Time",
```
dmitrylyzo, line 603: Add new strings in alphabetical order (move below and swap).

ethanarns: Changes applied, thank you

dmitrylyzo: Also move them below. They should be after: https://github.com/jellyfin/jellyfin-web/blob/b7995eee4a9b13bb67d4adce03da726f22621f6f/src/strings/en-us.json#L865

ethanarns: Fixed that as well

src/scripts/screensavermanager.js
```diff
@@ -11,7 +11,7 @@ import './screensavermanager.scss';
 function getMinIdleTime() {
     // Returns the minimum amount of idle time required before the screen saver can be displayed
     //time units used Millisecond
-    return 180000;
+    return userSettings.get('screensaverTime') * 1000;
```
dmitrylyzo, line 14: The suggestion from https://github.com/jellyfin/jellyfin-web/pull/6165#discussion_r1788527870 was not fully applied.
```suggestion
    return userSettings.screensaverTime() * 1000;
```

ethanarns: My bad, change made

PR #6168, opened 2024-10-06
Fix UI freeze when opening same path

viown:
When clicking on a link that leads to the same path (such as for example the same episode under the 'More from' section in the details page), the UI will freeze because it's waiting on a promise that will never resolve. Regression from https://github.com/jellyfin/jellyfin-web/pull/6024

**Changes**
Listen for path changes and call `this.onViewShow()` if the path did not change.

**Issues**

### DIFF
--- a/src/components/router/appRouter.js
+++ b/src/components/router/appRouter.js
@@ -39,6 +39,9 @@ class AppRouter {
     constructor() {
         document.addEventListener('viewshow', () => this.onViewShow());
 
+        this.lastPath = history.location.pathname + history.location.search;
+        this.listen();
+
         // TODO: Can this baseRoute logic be simplified?
         this.baseRoute = window.location.href.split('?')[0].replace(this.#getRequestFile(), '');
         // support hashbang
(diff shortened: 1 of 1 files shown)


src/components/router/appRouter.js
```diff
@@ -100,6 +100,22 @@ class AppRouter {
         return this.promiseShow;
     }
 
+    listen() {
+        let lastPath = history.location.pathname + history.location.search;
```
thornbill, line 104: Could this just be moved to the constructor?

src/index.jsx
```diff
@@ -110,6 +110,9 @@ build: ${__JF_BUILD_VERSION__}`);
         Events.on(apiClient, 'requestfail', appRouter.onRequestFail);
     });
 
+    // Listen for route changes
+    appRouter.listen();
```
thornbill, line 114: Could the call to `listen()` be moved into the constructor also? Eventually I would like to remove app router entirely, so keeping it more self-contained should help with that.

thornbill (approved):

PR #6172, opened 2024-10-08
Fix playback subscriber callback binding

thornbill:
**Changes**
* Fixes playback subscriber callbacks not being bound to the right context
* Changes player access to `protected` so implementations can access the player instance

**Issues**
N/A

### DIFF
--- a/src/apps/stable/features/playback/utils/playbackSubscriber.ts
+++ b/src/apps/stable/features/playback/utils/playbackSubscriber.ts
@@ -39,35 +39,35 @@ export interface PlaybackSubscriber {
 }
 
 export abstract class PlaybackSubscriber {
-    private player: Plugin | undefined;
+    protected player: Plugin | undefined;
 
     private readonly playbackManagerEvents = {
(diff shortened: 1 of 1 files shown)


nielsvanvelzen (approved):

src/apps/stable/features/playback/utils/playbackSubscriber.ts
```diff
@@ -87,15 +87,15 @@ export abstract class PlaybackSubscriber {
 
         if (this.player) {
             Object.entries(this.playerEvents).forEach(([event, handler]) => {
-                if (handler) Events.off(this.player, event, handler);
+                if (handler) Events.off(this.player, event, handler.bind(this));
```
dmitrylyzo, line 90: Doesn't `bind` return a new function and `handler` won't be disconnected?

thornbill: Good call. I think this is better...

dmitrylyzo (approved):

Issue #4698, opened 2023-06-27
Implement playlist ACL Interface


Shadowghost:
Our playlists always included the possibility to define shared users, up until now if a user created a playlist his username was added to that list and that was it.

Since https://github.com/jellyfin/jellyfin/pull/9466 the creator gets set as `OwnerUserId` which always has full access and any other user can be added to `Shares` with readonly or write permissions.
Additionally the playlist can be marked as `OpenAccess` which gives all users read permissions.

The items within a playlist are filtered for each user, so even if the playlist includes restricted items, they are hidden for users without access to those items.

**Expected Behavior**
* Interface should provide means to add and remove users to `Shares` with the ability to give and revoke edit permissions
* Interface should provide means to set a playlist as `OpenAccess`

thornbill:
The following functionality is present for 10.9.0:

* A playlist can be marked as public (or not) when created
* When adding an item to a playlist only playlists that the user has edit perms for are listed
* The controls for removing and reordering items in a playlist are hidden if a user does not have edit perms

For 10.10.0 we still need:

* The ability to change the visibility of an existing playlist
* UI for managing permissions for a playlist at the individual user level

viown:
After thinking about this feature some more, I don't think we should enable individual user level management for playlists for all users as that may introduce some privacy concerns.

A server change may be needed to introduce a per-user configuration option for enabling playlist user-level management and possibly also a global option for enabling the feature.

PR #6184, opened 2024-10-09
Add playlist editing

thornbill:
**Changes**
* Updates the playlist editor to support editing playlists :scream: 

Currently this allows changing the playlist name and public access but **not** user level access controls

**Issues**
Part of  #4698


solidsnake1298:
I was able to toggle the public access setting of an existing playlist, which worked as expected.

crobibero (approved):
Works for me, only note is that is uses the same icon as "Edit metadata"

src/components/itemContextMenu.js
```diff
@@ -466,6 +477,15 @@ function executeCommand(item, id, options) {
...
                 editItem(apiClient, item).then(getResolveFunction(resolve, id, true), getResolveFunction(resolve, id));
                 break;
+            case 'editplaylist':
+                import('./playlisteditor/playlisteditor').then(({ default: PlaylistEditor }) => {
+                    const playlistEditor = new PlaylistEditor();
+                    playlistEditor.show({
+                        id: itemId,
+                        serverId
+                    }).then(getResolveFunction(resolve, id, true), getResolveFunction(resolve, id));
+                });
```
ferferga, line 487: Isn't this more readable without chaining thens?

thornbill: I'm not sure I follow what you are suggesting 😅

nielsvanvelzen:
Seems to work fine in my testing with a non-admin user. Only UX issue I could find is when you use a space as name it allows you to save that, not a major issue but might be good to check for that (e.g. `name.trim().length`).

thornbill:
The context menu needs a total overhaul so I'm not touching that here beyond adding the new entry 😅

src/utils/string.test.ts
```diff
@@ -1,6 +1,24 @@
 import { describe, expect, it } from 'vitest';
 
-import { toBoolean, toFloat } from './string';
+import { isBlank, toBoolean, toFloat } from './string';
+
+describe('isBlamk', () => {
```
dmitrylyzo, line 5: `blaMk` 😄

thornbill: this is what happens when you type while holding a kid :rofl:

src/components/playlisteditor/playlisteditor.ts
```diff
@@ -99,6 +113,29 @@ function redirectToPlaylist(id: string | undefined) {
     appRouter.showItem(id, currentServerId);
 }
 
+function updatePlaylist(dlg: DialogElement) {
+    const apiClient = ServerConnections.getApiClient(currentServerId);
+    const api = toApi(apiClient);
+
+    if (!dlg.playlistId) return Promise.reject(new Error('Missing playlist ID'));
```
dmitrylyzo, line 120: I'd move this to the beginning of the function to avoid unnecessary calls in case of an error.

dmitrylyzo (approved):
LGTM. Just nitpicking.

Dyredhead:
Any update on whether this is finished and if it is, what Jellyfin release this will be in? Struggling to find info on this very important feature.

PR #6185, opened 2024-10-09
Add details view main section

grafixeyehero:
**Changes**
- Add detail view main section


thornbill:
This looks pretty good so far. I will probably wait to merge it post-10.10 since the page is missing some sections.

Also I would not necessarily be too strict in replicating the current UI exactly since we will be overhauling it. (I would love to eliminate the different colored banner there... :sweat_smile:)

PR #6187, opened 2024-10-10
Fix padding issues in the metadata manager

thornbill:
**Changes**
* Fixes some padding issues in the metadata manager... it's not perfect, but much better imo

<table>
<tr>
<th>Before</th>
<td>

Screenshot 2024-10-10 at 11-01-08 Metadata Manager

</td>
</tr>
<th>After</th>
<td>

Screenshot 2024-10-10 at 11-00-00 Metadata Manager

</td>
<tr>
</tr>
</table>

**Issues**
N/A


nielsvanvelzen (approved):
Still looks like 💩 but slightly better

PR #6189, opened 2024-10-13
Migrate webpack scripts to TypeScript

joeykilpatrick:
**Changes**
Migrate webpack build scripts to TypeScript (webpack.\*.js --> webpack.\*.ts)
Adds ts-node as loader for webpack CLI


joeykilpatrick:
1. I'm not sure how to get Github Actions to re-review now that eslint issue is fixed.
2. The Sonarcloud failure is not related to this change, but I would be happy to add `webpack.*.ts` to the list of excluded files for that scan. Just looking for some feedback before I do that. I don't see a platform-independent way to address the issue.

thornbill:
No worries. I'll take care of both of those. The review just has to be dismissed since the action doesn't update the status once issues are fixed and I can accept the issue in sonarcloud to clear that.

PR #6196, opened 2024-10-14
Add 'ask to skip' to media segments

viown:
(Does not support TV OSes just yet)

The button appears for 6 seconds, and can be shown again by moving the mouse to unhide the OSD panel.

Testing appreciated.

**Changes**
- Add a "Prompt To Skip" option to media segments
- Display "Next Episode" if it's an outro segment, and the segment exceeds or is equal to the runtime, and if there is a next track.

**Screenshots**
Untitled
Screenshot_20241014_105324

**Issues**


src/components/playback/skipsegment.ts
```diff
@@ -0,0 +1,176 @@
...
+
+    currentPlayer = player;
+
+    Events.on(player, 'promptskip', onPromptSkip);
+    Events.on(document, EventType.SHOW_VIDEO_OSD, onOsdChanged);
+    Events.on(player, 'playbackstop', onPlaybackStopped);
+    Events.on(player, 'timeupdate', onTimeUpdate);
+}
+
+Events.on(playbackManager, 'playerchange', (_e, newPlayer) => {
```
thornbill, line 173: I would make this a `PlaybackSubscriber` so the event handling is managed for you. 😉

viown: I was actually considering this but didn't look into it enough to see if that could work.

I'll change this to use PlaybackSubscriber, thanks!

nielsvanvelzen (commented):
Would be nice if we kept this implementation consistent with ATV (https://github.com/jellyfin/jellyfin-androidtv/pull/4068):

- Name the action ÄskToSkip"
- Only show for segments with duration of 3 seconds or more
- Auto-hide after 8 seconds
- Skipping does nothing if there is only 1 second left in the segment (just like the normal skip action)

viown:
Good point. And it already checks for segments <3s.

dmitrylyzo:
We have `UpNextDialog`: https://github.com/jellyfin/jellyfin-web/blob/50dd1fb855a00633659a92430b8e97e397f06222/src/controllers/playback/video/index.js#L669-L684
I thought we could extract it as `PlaybackSubscriber`, but it could be merged to `MediaSegmentManager`.

viown:
> We have `UpNextDialog`:
> 
> https://github.com/jellyfin/jellyfin-web/blob/50dd1fb855a00633659a92430b8e97e397f06222/src/controllers/playback/video/index.js#L669-L684
> 
> I thought we could extract it as `PlaybackSubscriber`, but it could be merged to `MediaSegmentManager`.

Do you suggest we use the `UpNextDialog` for all segments?

dmitrylyzo:
> Do you suggest we use the `UpNextDialog` for all segments or just the "Next Episode" part?

:thinking: Not for all segments, for sure. I like simple buttons for general case. `UpNextDialog` should be able to work without segments (_I'm leaning more towards a dedicated `PlaybackSubscriber`_), and we need to make sure it doesn't conflict with the segment action buttons.

viown:
I do like the idea of using the up next dialog over the current "Next Episode" button for outro segments as there is a possibility these could clash with each other currently.

Since I believe this dialog shows up 30s before the runtime, we can keep that behaviour as is, but also show the dialog on outro segments that result in the next track. This won't eliminate the clashing possibility completely, but it's acceptable.

Shadowghost:
If skipping a segment leads to playing the next episode, I think it's sensible to re-use the dialogue - but with a different text to properly indicate the skipping

pimw1:
> ## Cloudflare Pages deployment
> **Latest commit**	`6ca8311`
> **Status**	✅ Deployed!
> **Preview URL**	https://3e8577ef.jellyfin-web.pages.dev
> **Type**	🔀 Preview
> **[View build logs](https://github.com/jellyfin/jellyfin-web/actions/runs/11345908756)**

The into skipper works as expected. I don't have media with outro segments, so i have not tested that.

viown:
Updated to use `UpNextDialog` for outro segments leading to next track.

src/apps/stable/features/playback/utils/playbackSubscriber.ts
```diff
@@ -11,13 +11,15 @@ import Events, { type Event } from 'utils/events';
 import { PlaybackManagerEvent } from '../constants/playbackManagerEvent';
 import { PlayerEvent } from '../constants/playerEvent';
 import type { ManagedPlayerStopInfo, MovedItem, PlayerError, PlayerErrorCode, PlayerStopInfo, RemovedItems } from '../types/callbacks';
+import { MediaSegmentDto } from '@jellyfin/sdk/lib/generated-client';
```
thornbill, line 14: We should import from `...generated-client/models/media-segment-dto`

src/components/playback/playbackmanager.js
```diff
@@ -3680,6 +3693,10 @@ export class PlaybackManager {
         return this.getCurrentTicks(player) / 10000;
     }
 
+    getNextItem() {
+        return this._playQueueManager.getNextItemInfo();
```
thornbill, line 3697: Do we need to do anything for player managed queues? Or does the `playQueueManager` handle that logic?

src/controllers/playback/video/index.js
```diff
@@ -631,6 +632,17 @@ export default function (view) {
         }
     }
 
+    function onPromptSkip(e, mediaSegment) {
+        const player = this;
+        if (mediaSegment && player && mediaSegment.EndTicks != null
+            && mediaSegment.Type === MediaSegmentType.Outro
```
thornbill, line 638: Should we hardcode the check for `Outro` segments? What if it is something like a recording with a commercial at the end?

src/components/playback/playbackmanager.js
```diff
@@ -933,6 +935,14 @@ export class PlaybackManager {
             return Promise.resolve(self._playQueueManager.getPlaylist());
         };
 
+        self.promptToSkip = function (mediaSegment, player) {
+            player = player || self._currentPlayer;
+
+            if (mediaSegment && this._skipSegment) {
+                Events.trigger(player, 'promptskip', [mediaSegment]);
```
thornbill, line 942: Should probably use the constant value for the event type.

src/controllers/playback/video/index.js
```diff
@@ -579,6 +578,7 @@ export default function (view) {
         }, state);
         Events.on(player, 'playbackstart', onPlaybackStart);
         Events.on(player, 'playbackstop', onPlaybackStopped);
+        Events.on(player, 'promptskip', onPromptSkip);
```
thornbill, line 581: Same here regarding using the constant values. (The other events should probably be updated in a future cleanup PR)

src/apps/stable/features/playback/utils/playbackSubscriber.ts
viown: Fixed in `skipsegment` as well.

src/controllers/playback/video/index.js
viown: Sorry, I'm not sure I understand.

thornbill: I'm wondering if we need to check the segment type at all here. An outro might not be the last segment in a file. 🤷‍♂️

viown: It does make sense to remove this check as it can apply to other segment types also.

solidsnake1298:
Skip prompts work for all segment types.  As does the prompt to start the next episode for outro segments.

thornbill (approved):

PR #6200, opened 2024-10-14
Add User-Configurable Playback Duration for Media Segment Start and End

rlauuzo:
Introduces new user-configurable settings for adjusting the playback duration of media segment start and end times. The MediaSegmentManager is updated to pre-adjust segments based on these settings, giving users greater flexibility in playback control.

### DIFF
--- a/src/apps/stable/features/playback/utils/mediaSegmentManager.ts
+++ b/src/apps/stable/features/playback/utils/mediaSegmentManager.ts
@@ -30,11 +30,22 @@ class MediaSegmentManager extends PlaybackSubscriber {
 
         try {
             const { data: mediaSegments } = await mediaSegmentsApi.getItemSegments({ itemId, includeSegmentTypes });
-            this.mediaSegments = mediaSegments.Items || [];
+            this.mediaSegments = this.adjustMediaSegments(mediaSegments.Items || []);
         } catch (err) {
             console.error('[MediaSegmentManager] failed to fetch segments', err);
             this.mediaSegments = [];
         }
+    }
+
+    private adjustMediaSegments(segments: MediaSegmentDto[]): MediaSegmentDto[] {
+        const startPlayTime = userSettings.mediaSegmentStartToPlay(undefined) * TICKS_PER_SECOND;
+        const endPlayTime = userSettings.mediaSegmentEndToPlay(undefined) * TICKS_PER_SECOND;
+
+        return segments.map(segment => ({
+            ...segment,
+            StartTicks: segment.StartTicks !== undefined ? segment.StartTicks + startPlayTime : undefined,
+            EndTicks: segment.EndTicks !== undefined ? segment.EndTicks - endPlayTime : undefined
+        }));
     }
 
     private performAction(mediaSegment: MediaSegmentDto) {
--- a/src/apps/stable/features/playback/utils/mediaSegmentManager.ts
+++ b/src/apps/stable/features/playback/utils/mediaSegmentManager.ts
@@ -38,19 +38,17 @@ class MediaSegmentManager extends PlaybackSubscriber {
     }
 
     private adjustMediaSegments(segments: MediaSegmentDto[]): MediaSegmentDto[] {
-        const startPlayTime = userSettings.mediaSegmentStartToPlay() * TICKS_PER_SECOND;
         const endPlayTime = userSettings.mediaSegmentEndToPlay() * TICKS_PER_SECOND;
 
         return segments.flatMap(segment => {
-            const adjustedStart = segment.StartTicks !== undefined ? segment.StartTicks + startPlayTime : undefined;
             const adjustedEnd = segment.EndTicks !== undefined ? segment.EndTicks - endPlayTime : undefined;
 
-            if (adjustedStart !== undefined && adjustedEnd !== undefined && adjustedStart >= adjustedEnd) {
-                console.debug(`Skipping segment with StartTicks (${adjustedStart}) >= EndTicks (${adjustedEnd})`);
+            if (segment.StartTicks !== undefined && adjustedEnd !== undefined && segment.StartTicks >= adjustedEnd) {
+                console.debug(`Skipping segment with StartTicks (${segment.StartTicks}) >= EndTicks (${adjustedEnd})`);
                 return [];
             }
 
-            return [{ ...segment, StartTicks: adjustedStart, EndTicks: adjustedEnd }];
+            return [{ ...segment, EndTicks: adjustedEnd }];
         });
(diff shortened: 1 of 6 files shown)


src/apps/stable/features/playback/utils/mediaSegmentManager.ts
```diff
@@ -30,13 +30,24 @@ class MediaSegmentManager extends PlaybackSubscriber {
...
+            this.mediaSegments = this.adjustMediaSegments(mediaSegments.Items || []);
         } catch (err) {
             console.error('[MediaSegmentManager] failed to fetch segments', err);
             this.mediaSegments = [];
         }
     }
 
+    private adjustMediaSegments(segments: MediaSegmentDto[]): MediaSegmentDto[] {
+        const startPlayTime = userSettings.mediaSegmentStartToPlay(undefined) * TICKS_PER_SECOND;
+        const endPlayTime = userSettings.mediaSegmentEndToPlay(undefined) * TICKS_PER_SECOND;
```
viown, line 42:
```suggestion
        const endPlayTime = userSettings.mediaSegmentEndToPlay() * TICKS_PER_SECOND;
```

dmitrylyzo:
> giving users greater flexibility in playback control.

Could you explain your use case?

It can be an adjustment to seek before an exact start/end time. But one setting is enough, imo. It can also be used for prev/next chapter buttons.

rlauuzo:
My intention is to allow users to adjust playback slightly before the segment ends, similar to how Netflix jumps a couple of seconds before the intro ends so that you still see the show's logo and avoid cutting off any actual content.

Regarding the settings—did I overlook an existing option that already handles this, or would this be a useful new feature?

dmitrylyzo:
> My intention is to allow users to adjust playback slightly before the segment ends, similar to how Netflix jumps a couple of seconds before the intro ends so that you still see the show's logo and avoid cutting off any actual content.

That's exactly what I meant. But why do you need two separate settings, one for the beginning and one for the end?

> Regarding the settings—did I overlook an existing option that already handles this, or would this be a useful new feature?

There is currently no such adjustment. I just mean that the new setting can also be used to navigate through chapters.

rlauuzo:
Applies the user's segment end setting to the next chapter skip feature. Also removes the previous adjustment for start position.

thornbill:
I really just don't see the point in this. If there is an issue with the segment timing that seems like an issue with either the media or the plugin providing the segment data. :man_shrugging:

jumoog:
It's not a segment issue, it's more a user preference. At the moment we have that in our plugin and we have users who adjust the timings. Since there can be multiple plugins, it makes sense to move this to jellyfin now. I think that was @rlauuzo intention.

AbandonedCart:
> I really just don't see the point in this. If there is an issue with the segment timing that seems like an issue with either the media or the plugin providing the segment data. 🤷‍♂️

Allow me to chime in with some use cases that may have been overlooked.

1. Demon Slayer and a number of other anime use the same intro and then the same sound for the title splash screen. This title is indistinguishable from the intro itself. A lot of people want to see this title. Chapter markers often discard this.

2. User often want to retain a portion of the credits because action is still happening as they begin to roll. This is, once again, somewhat indistinguishable from the actual credits. Despite the lack of dialog, it is a part of the show. Chapter markers often overlook this.

3. User preference. It’s something that a lot of users have requested for a number of other reasons. The other two points cover most of the logical reasons, but there are still a number of others with other reasons that specifically ask for instructions on how to edit.

I completely understand the assumption that the detection is the problem. Ironically, it’s that the detection is too precise, but it’s far easier to provide flexibility than try to force every file and user to come to an agreement.

nielsvanvelzen:
Shifting the start and end times of segments is something the segment provider should do. We could add options for that to our chapter segment provider if there is demand. It's definitely **not** up to the client to do. A client does not know the source of a segment and thus what the timings are based on.

AbandonedCart:
> Shifting the start and end times of segments is something the segment provider should do.

That's fine. Keep in mind that the more you defer to us as the provider, the more heavily you are relying on us to fuel the feature. While the UI matters, the button and an endpoint is essentially the only part of the feature Jellyfin has actually integrated.

rlauuzo:
> Shifting the start and end times of segments is something the segment provider should do. We could add options for that to our chapter segment provider if there is demand. It's definitely **not** up to the client to do. A client does not know the source of a segment and thus what the timings are based on.

I strongly disagree with that position. It's akin to suggesting that subtitle offset adjustments shouldn't be handled by the client. Instead, subtitle providers should rewrite all subtitle files with modified timings whenever a simple setting like offset is changed.

joshjryan:
> Shifting the start and end times of segments is something the segment provider should do. We could add options for that to our chapter segment provider if there is demand. It's definitely **not** up to the client to do. A client does not know the source of a segment and thus what the timings are based on.

I think the key part of this improvement is that it's *not* shifting the segment. It's giving an option to resume playback a few seconds before thr segment ends.

I think it's great that segments are exactly accurate in start/end ticks. But, a user should have the option to skip the segment, but resume a few seconds early.

ethaldeman:
> I really just don't see the point in this. If there is an issue with the segment timing that seems like an issue with either the media or the plugin providing the segment data. 🤷‍♂️

> Shifting the start and end times of segments is something the segment provider should do. 

It isn't an issue with the segment data. The segment data is 100% accurate. Is is just the user preference to show a few seconds of the skipped intro or other media segment. Skipping right to end end of an intro can be jarring, and an option to view the last few seconds of the intro  helps with this. 

This is a capability of the Intro Skipper plugin I heavily use and would appreciate being formally added to Jellyfin.

src/apps/stable/features/playback/utils/mediaSegmentManager.ts
```diff
@@ -30,13 +30,28 @@ class MediaSegmentManager extends PlaybackSubscriber {
...
             const { data: mediaSegments } = await mediaSegmentsApi.getItemSegments({ itemId, includeSegmentTypes });
-            this.mediaSegments = mediaSegments.Items || [];
+            this.mediaSegments = this.adjustMediaSegments(mediaSegments.Items || []);
         } catch (err) {
             console.error('[MediaSegmentManager] failed to fetch segments', err);
             this.mediaSegments = [];
         }
     }
 
+    private adjustMediaSegments(segments: MediaSegmentDto[]): MediaSegmentDto[] {
```
joshjryan, line 40: In order to decrease confusion, I wonder if this method should be renamed to `applySegmentSkipPreferences()` (or something along those lines, so that it's clear that you aren't adjusting the segment itself).

PR #6201, opened 2024-10-14
Rollback to dompurify v2 for legacy support

thornbill:
**Changes**
dompurify v3 removed legacy browser support. v2 is still receiving security updates so we should use it instead of an outdated version of v3.

**Issues**
Fixes known CVEs
Closes https://github.com/jellyfin/jellyfin-web/pull/6077
Closes https://github.com/jellyfin/jellyfin-web/pull/6076

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -10,6 +10,11 @@
       "followTag": "unstable",
       "minimumReleaseAge": null,
       "schedule": [ "after 7:00 am" ]
+    },
+    {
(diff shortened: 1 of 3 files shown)


ferferga (changes requested):
Shouldn't it be also excluded from renovate config?

thornbill:
Yeah good call @ferferga... better to handle that properly than deal with closing the PR all the time :sweat_smile:

ferferga (approved):

PR #6205, opened 2024-10-15
Add hack to prevent rerendering legacy views in development

thornbill:
**Changes**
This fixes an issue where legacy views will sometimes break when running in development mode due to React intentionally re-rendering as a debugging feature. Most notably the video player would crash when this happened.

**Issues**
N/A


nielsvanvelzen (approved):

dmitrylyzo:
It looks like we should check `location` and if it has been changed, load the view. At the moment only the first view is loaded in non-experimental mode.

abdulmueid:
This commit breaks navigation for me in Desktop and TV mode during development and production. I can't navigate to different screens past this commit, everything works if I rollback to one commit before this one. 
Just reporting here.

Issue #6211, opened 2024-10-16
[bug] Unintentional wrap of button label for "SaveChanges" (multiple languages)


bluemoehre:
**Describe The Bug**

The label of the save button breaks into multiple lines. This unnecessarily affects the layout and wastes space on mobile screens. 
It happens with several languages if the label has multiple words and exceeds ~16 chars.

[All related project files (59)](https://github.com/search?q=repo%3Ajellyfin%2Fjellyfin-web+SaveChanges&type=code)

(This is an UI/UX issue only - no functional problems detected yet)

**Steps To Reproduce**

1. Change UI language to eg. `DE` |  `FR` | `HU` | `RO` …
2. Go to Overview
3. Open any item's menu
4. Select Edit Metadata

**Expected Behavior**

Button label _should not break_ into multiple lines. This can be achieved by using shorter text, enlarging the buttons or allow them to grow.

At least for `DE`, `EN`, `ES`, `FR` I can say that **a single word** like "Save" or "Apply" **is absolutely common**.
I am sure it's the same with more languages, so **I would recommend updating the translations**.

**EDIT:** I found there is already a short version `Save` in most languages. I am not sure why this is mixed.

**Screenshots**

![Screenshot](https://github.com/user-attachments/assets/b83613ab-79b9-41d1-8fd7-c78967a525c2)

**System (please complete the following information):**
 - Platform: all
 - Browser: all
 - Jellyfin Version: 10.9.11

**Additional Context**

I've already checked the CSS, and if we prefer changes here, it's also possible. Just a slightly higher risk of breaking something elsewhere.

tylers-username:
MR: https://github.com/jellyfin/jellyfin-web/pull/6250

bluemoehre:
Bump

thornbill:
Is this still an issue since your PR #6212 changed the text to use the shorter "Save" text?

bluemoehre:
Yes it's fixed for DE. I did not check the other languages yet and did not want to rely on a guess.

bluemoehre:
Looks good to me. Thanks everyone.

PR #6212, opened 2024-10-16
Use "Save" in Metadata Editor and remove "SaveChanges".

bluemoehre:
Unified/merged `Save` & `SaveChanges` translations.

**Changes**

- Merged both into `SaveChanges`
  - because this key was more recent in commit history
  - and is more accurate about the context
- Use of shortest value to prevent unnecessary line-breaks in UI & save space
- Double checked all translations with ChatGPT 4o 
  - should be common for a software context
  - and for saving or applying user input

**Issues**

#6211


bluemoehre:
Technically, I have no problems with changing it back to `Save`, but the reasoning (above) still makes sense to me, especially if I compare all the rest of the keys.

> FYI, other languages should only be edited with Weblate to avoid conflicts.

It's a one-click in my IDE to convert everything. Will it cause problems that way?

thornbill:
Yes it does cause issues, and we do not accept llm generated translations.

bluemoehre:
> we do not accept llm generated translations.

It's not LLM generated translations! It's just checked by LLM because I didn't know if the older or newer version was the one to take. So it's just for everyone's assurance to not only rely on me. (Obviously fixing a long time bug which affects all language files is nothing easy without automation)

> Yes it does cause issues

How can I change this PR so we can get rid of the duplicates, without doing everything manually?

dmitrylyzo:
> How can I change this PR so we can get rid of the duplicates, without doing everything manually?

Since we only need to narrow the text:
1. Revert all.
2. Change `SaveChanges` to `Save` in `src/components/metadataEditor/metadataEditor.template.html`.
3. Remove `SaveChanges` from `src/strings/en-us.json` (since it's no longer in use). Weblate will remove it from other languages.

bluemoehre:
@dmitrylyzo done as you suggested. For DE it's fine. Native speakers of the respective languages (FR, PT, BU, etc.) must adapt their buttons themselves, then.

thornbill:
Sorry for misunderstanding your GPT comment before.

Our weblate system will remove the translations in other languages once it is removed from the primary language. If the "Save" translation is problematic that would be concerning since it is already used in similar edit dialogs/screens.

dmitrylyzo (approved):
LGTM. Merge with squash.
I have changed the PR title, but it can be improved if necessary.

bluemoehre:
Thanks everyone! 🙂👍

PR #6231, opened 2024-10-20
Fix tags and schedule edit for parental control

gnattu:
Our callback implementation for the parental control page was quite hacky and fragile. It stopped working properly in the current master branch, likely due to updates in React. This isn't the only instance where we manipulate DOM objects directly instead of using React, so other pages may also be affected, but this one has been reported.

In the parental control page, the popup callback was added twice since `useEffect` was executed twice in each render cycle. Also, the delete tag/schedule button had no callbacks assigned because the DOM object wasn't available when the data loading hook was executing.

**Changes**

- Migrate onClick to pure react callback for delete operations.
- Unbind showing popup callbacks when unmounting

**Issues**

### DIFF
--- a/src/apps/dashboard/routes/users/parentalcontrol.tsx
+++ b/src/apps/dashboard/routes/users/parentalcontrol.tsx
@@ -70,13 +70,6 @@ const UserParentalControl = () => {
     const [ blockedTags, setBlockedTags ] = useState<string[]>([]);
     const libraryMenu = useMemo(async () => ((await import('../../../../scripts/libraryMenu')).default), []);
 
-    // The following are meant to be reset on each render.
-    // These are to prevent multiple callbacks to be added to a single element in one render as useEffect may be executed multiple times in each render.
-    let allowedTagsPopupCallback: (() => void) | null = null;
-    let blockedTagsPopupCallback: (() => void) | null = null;
-    let accessSchedulesPopupCallback: (() => void) | null = null;
-    let formSubmissionCallback: ((e: Event) => void) | null = null;
-
     const element = useRef<HTMLDivElement>(null);
 
     const populateRatings = useCallback((allParentalRatings: ParentalRating[]) => {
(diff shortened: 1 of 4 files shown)


src/apps/dashboard/routes/users/parentalcontrol.tsx
```diff
@@ -70,6 +70,13 @@ const UserParentalControl = () => {
     const [ blockedTags, setBlockedTags ] = useState<string[]>([]);
     const libraryMenu = useMemo(async () => ((await import('../../../../scripts/libraryMenu')).default), []);
 
+    // The following are meant to be reset on each render.
+    // These are to prevent multiple callbacks to be added to a single element in one render as useEffect may be executed multiple times in each render.
```
viown, line 74: If I'm correct, this is only specific to strict mode and won't happen in production?

src/apps/dashboard/routes/users/parentalcontrol.tsx
```diff
@@ -503,11 +479,12 @@
                         <div className='accessScheduleList paperList'>
                             {accessSchedules.map((accessSchedule, index) => {
                                 return <AccessScheduleList
-                                    key={accessSchedule.Id}
+                                    key={index}
```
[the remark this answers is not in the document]
gnattu: This ID is never populated in the client side and will cause problems if you add multiple schedules at once.

src/apps/dashboard/routes/users/parentalcontrol.tsx
gnattu: We are using Strict Mode, and in React 18 this will be executed twice.

src/apps/dashboard/routes/users/parentalcontrol.tsx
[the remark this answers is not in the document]
gnattu: Now workaround that by using the day of week + hours string representation as the key, which should be okay for sane users as they will not create the exact same schedule multiple times.

src/apps/dashboard/routes/users/parentalcontrol.tsx
viown: I think it would be cleaner to unbind the event listeners when useEffect is unmounted? Unless there is something barring this from happening that I missed.

gnattu: Makes sense. Now unbind all event listeners on unmounting

src/apps/dashboard/routes/users/parentalcontrol.tsx
```diff
@@ -348,45 +306,31 @@ const UserParentalControl = () => {
...
-            btnDelete.addEventListener('click', function () {
-                const index = parseInt(btnDelete.getAttribute('data-index') ?? '0', 10);
-                const newindex = accessSchedules.filter((_e, i) => i != index);
-                setAccessSchedules(newindex);
-            });
-        }
-    }, [accessSchedules]);
+        };
+        (page.querySelector('#btnAddSchedule') as HTMLButtonElement).addEventListener('click', accessSchedulesPopupCallback);
+        const allowedTagsPopupCallback = showAllowedTagPopup;
```
viown, line 320: These shouldn't be needed anymore?

gnattu: The schedule one should still be needed, the other can be removed

viown (approved):

PR #6257, opened 2024-10-25
Smart font streaming for external ASS subtitles

p0358:
!! requires jellyfin/jellyfin#12877 -- read that PR for more information and discussion on this feature, to avoid duplicating explanations here

**Changes**

- support grabbing the font list and fonts for external ASS subtitles
  - currently it modifies the DeliveryUrl of media stream of subtitles to make it into the fonts endpoint, discuss on the main PR if instead of that, something like `FontsUrl` should be added next to `DeliveryUrl` or not
  - don't try to grab fonts if it's not a normal .ass file from the API, and one that's not inside of .mkv (the latter comes from assumption from main PR that for now we don't scan to missing fonts in MKV, since there's no code right now to filter out existing font attachments, and we don't want to duplicate fonts, assuming that usually fonts are embedded in MKV properly)
- refactor the function to make it more readable and better by making all promises parallel at once, to avoid waiting a few round-trip times (and wasm loading + both font endpoints can all take some time on cold requests, so it's only logical to do them all at once to decrease any potential delay in displaying subtitles)
  - turns 3 round-trips into just one, yay

**Issues**
Closes #5960 old PR by superseding it with this

(also since this is a draft, I didn't test the code yet, will remove this line after I do)

### DIFF
--- a/src/plugins/htmlVideoPlayer/plugin.js
+++ b/src/plugins/htmlVideoPlayer/plugin.js
@@ -164,6 +164,27 @@ function getTextTrackUrl(track, item, format) {
     return url;
 }
 
+function getSubtitlesFontsUrl(track, item) {
+    if (itemHelper.isLocalItem(item) && track.Path) {
+        return false;
+    }
+
+    if (!track.IsExternal || track.IsExternalUrl) {
+        return false;
+    }
+
+    const apiClient = ServerConnections.getApiClient(item.ServerId);
+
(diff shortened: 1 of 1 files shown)


p0358:
Ah no, I actually messed up with the `encodingOptions`, I guess one more round trip might be unavoidable... I think I'll refactor it in a way where `encodingOptions -> grab fallback fonts` will be an independent promise, since those combined could take faster than waiting for wasm loading and other fonts list loading before making the second request (radical performance thinking xd)

Issue #6263, opened 2024-10-28
Persistent filter issue in shows and music libraries


theguymadmax:
**Describe The Bug**

In a shows or music library, selecting a letter to filter by on the vertical index persists when leaving the library. In contrast, the filter resets upon exiting the following library types: Movies, Mixed Movies & Shows, and Collections.

**Steps To Reproduce**

1. Navigate to your Jellyfin homepage
2. Select a Shows/Music library
3. Click on a letter to filter the library
4. Click the Home icon to return to the homepage
5. Return to the Shows/Music library 
6. Observe that the library remains filtered by the previous selection.

**Expected Behavior**

The filter should be cleared when leaving the library, returning to an unfiltered view.

**System (please complete the following information):**
 - Platform: Linux
 - Browser: Firefox
 - Jellyfin Version: 10.10.0

aaemon:
btnShuffle, btnSelectView, btnSort, btnFilter, and alphaPicker are all persistent across different devices. So, trying to use a filter on one device will mess up all other devices connected to the same account.

PR #6264, opened 2024-10-28
Don't persist letter filter when querying

viown:
Currently, the letter filter will only be persisted in some media types if you first select a letter, then use 'sort' or 'filter' (which calls save). In other media types like shows it's saved when items are loaded.

**Changes**
To stay consistent, prevent the letter filter from being saved anywhere.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-web/issues/6263

### DIFF
--- a/src/scripts/settings/userSettings.js
+++ b/src/scripts/settings/userSettings.js
@@ -536,7 +536,10 @@ export class UserSettings {
      * @param {Object} query - Query.
      */
     saveQuerySettings(key, query) {
-        return this.set(key, JSON.stringify(query));
+        const newQuery = { ...query };
+        delete newQuery.NameStartsWith;
+        delete newQuery.NameLessThan;
+        return this.set(key, JSON.stringify(newQuery));
     }
 
     /**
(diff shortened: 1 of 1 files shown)


viown:
I just noticed that If someone has an active letter filter and they used this, they wouldn't be able to get rid of the filter. To circumvent this, I'll remove these fields in `loadQuerySettings` as well.

viown:
Hmm, wondering if it's better to 'whitelist' certain settings instead.

thornbill:
That was the previous approach... https://github.com/jellyfin/jellyfin-web/pull/5717/commits/7074d41a56f237418747f094ea9f01779389bd2e

viown:
> btnShuffle, btnSelectView, btnSort, btnFilter, and alphaPicker are all persistent across different devices. So, trying to use a filter on one device messes up all other devices connected to the same account.

The settings are indeed server-side. To match previous behavior, we can make sorting work across devices but keep filtering local to each device. It's better that we do this in another PR.

PR #6266, opened 2024-10-28
Make ask to skip the default for 'Intro' and 'Outro' segments

viown:
To stay consistent with https://github.com/jellyfin/jellyfin-androidtv/pull/4111

**Changes**
- Make 'Ask to Skip' the default for intro and outro segments

**Issues**


thornbill (approved):

joshrandall8478:
can you please turn this into a server admin setting, or make it somewhat apparent to change the default? my server uses another plugin to skip intros and outros, and i want the server button to be the only button there by default.

PR #6271, opened 2024-10-30
Refactor media session to playback subscriber

thornbill:
**Changes**
* Migrates MediaSession handling to TypeScript and a PlaybackSubscriber
* Removes duplicate functions for getting the "now playing" image

**Issues**
N/A


src/apps/stable/features/playback/utils/mediaSessionSubscriber.ts
```diff
@@ -0,0 +1,178 @@
...
+/** The default image resolutions to provide to the media session */
+const DEFAULT_IMAGE_SIZES = [96, 128, 192, 256, 384, 512];
+
+const hasNavigatorSession = 'mediaSession' in navigator;
+const hasNativeShell = !!window.NativeShell;
+
+const getArtwork = (item: ItemDto): MediaImage[] => {
+    const artwork: MediaImage[] = [];
+
+    DEFAULT_IMAGE_SIZES.forEach(height => {
```
ferferga, line 23: This can be a map

src/apps/stable/features/playback/utils/mediaSessionSubscriber.ts
```diff
@@ -0,0 +1,178 @@
...
+        const isVideo = item.MediaType === MediaType.Video;
+        const isLocalPlayer = !!this.player?.isLocalPlayer;
+
+        // Local players do their own notifications
+        if (isLocalPlayer && isVideo) {
+            console.debug('[MediaSessionSubscriber] ignoring local player update');
+            return;
+        }
+
+        const album = item.Album || undefined;
```
ferferga, line 112: Better to use ??

At some point we should tackle the annoying `null` introduced by the SDK 😅.

src/apps/stable/features/playback/utils/mediaSessionSubscriber.ts
```diff
@@ -0,0 +1,178 @@
...
+        const names = nowplayinghelper.getNowPlayingNames(item);
+        const artist = names[names.length - 1].text;
+        const title = names.length === 1 ? undefined : names[0].text;
```
ferferga, line 115: This condition is unnecessary? If you use nullish coalescing, text will be undefined if the index doesn't exist.

src/apps/stable/features/playback/utils/mediaSessionSubscriber.ts
thornbill: I didn't use 'map' since it's not 1:1 (`getImageUrl` can return `null` in which case we skip that entry)

src/apps/stable/features/playback/utils/mediaSessionSubscriber.ts
thornbill: Dmitry doesn't like `??` because the code babel produces for it is really verbose.

Agreed it would be really nice if the SDK only used `undefined` though. :upside_down_face:

src/apps/stable/features/playback/utils/mediaSessionSubscriber.ts
thornbill: Good call... I tend to forget about using nullish coalescing when accessing by index. :sweat_smile:

src/apps/stable/features/playback/utils/image.ts
```diff
@@ -0,0 +1,59 @@
...
+        return ServerConnections.getApiClient(item.ServerId).getScaledImageUrl(item.SeriesId, options);
+    }
+
+    if (options.type === ImageType.Thumb) {
+        if (item.SeriesId && item.SeriesThumbImageTag) {
+            options.tag = item.SeriesThumbImageTag;
+
+            return ServerConnections.getApiClient(item.ServerId).getScaledImageUrl(item.SeriesId, options);
+        }
+        if (item.ParentThumbItemId && item.ParentThumbImageTag) {
```
1hitsong, line 29: Add blank link between these two if blocks. All other conditionals have a blank line between blocks.

src/apps/stable/features/playback/utils/mediaSessionSubscriber.ts
ferferga: I also see that you guys don't have https://www.typescriptlang.org/tsconfig/#noUncheckedIndexedAccess enabled, which in these cases it's a life saver

thornbill: Oh nice! I will enable that in a separate PR because it introduced 70 errors. :upside_down_face:

ferferga: Lucky you, I had a bunch of errors in Vue when I enabled it!

ferferga (approved):

Issue #6268, opened 2024-10-29
Web Interface not Honoring Library Page Size


JRGonz:
When setting the Library Page Size to 0 (continuous scrolling) my library will still stop at 100 and not advance when the next (right arrow) is clicked. If I change the setting to 500 (as an example) then it will still only display 100 and advance to the next page but it is not 500, it is 100. 

Another library is stuck on continuous scrolling. If I change the Library Page Size setting to 500 then it will still show all shows in the library. 

I have tried to clear the browser cache, used private browsing, and tried 3 different browsers. Even more strange is if I use a different browser (with no Jellyfin cache / cookie data / etc) then it will be broken as well. I do not see this behavior for movie libraries. They seem to follow the setting just fine. I haven't checked music libraries, but show libraries are definitely affected. 

1. Go to Settings, Display
2. Change Library Page Size to 0 (zero)
3. Go to a library with > 100 items. 
4. The page will show 100 and not advance to the next 100 items (even though it should be showing all items instead of only 100)

The expected behavior is that the web client will honor the Library Page Size setting

I did not see any logs applicable to an issue. 

 - Platform: Debian Linux
 - Browser: LibreWolf, Brave, Epiphany, Jellyfin Media Player
 - Jellyfin Version: 10.10.0

Client is Ubuntu running above browsers. 
![Pasted image](https://github.com/user-attachments/assets/4fcf010b-8c5d-4855-8820-7159b2b76f84)
![Pasted image (2)](https://github.com/user-attachments/assets/aabd4974-e3e9-4668-8303-73c4be67e09c)
![Pasted image (3)](https://github.com/user-attachments/assets/c786fa43-3bb1-47e0-bdcc-b9ab9640a568)
![Pasted image (4)](https://github.com/user-attachments/assets/640c62fe-c6e0-4740-ad40-bef985da8dfc)
![Pasted image (5)](https://github.com/user-attachments/assets/3b89c1df-4387-40e4-b89e-4db1dbbcbfa4)
![Pasted image (6)](https://github.com/user-attachments/assets/68a31947-9506-4815-b619-1a6b749b4e15)

JRGonz:
No. This is with the standard web interface (I tried the browsers listed above)  and the official Jellyfin Media Player flatpak. I can't seem to sort out why it is doing it. I even went so far as to remove one of the shows libraries and re-add it and it still persists.

JRGonz:
[...]

PR #6274, opened 2024-10-30
Make filtering local to each device

viown:
Supersedes https://github.com/jellyfin/jellyfin-web/pull/6264

This makes 'filtering' local to each device while keeping sorting settings stored on the server. Additionally, this fixes the issue with storing query options which don't make sense to save (current page, letter filter, etc)

Due to a key change, this will reset any saved filtering settings but will keep sorting intact.

**Changes**
- Keep filtering local to each device, while only storing sorting settings on the server.
- Make it so that we only save filtering and sort options.
- Filter out any unneeded configuration options saved from previous versions.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-web/issues/6263
Fixes https://github.com/jellyfin/jellyfin-web/issues/6268
https://github.com/jellyfin/jellyfin-web/pull/6264#issuecomment-2443685093

### DIFF
--- a/src/scripts/settings/userSettings.js
+++ b/src/scripts/settings/userSettings.js
@@ -23,7 +23,9 @@ const filterSettingsPostfix = '-filter';
 const allowedFilterSettings = [
     'Filters', 'HasSubtitles', 'HasTrailer', 'HasSpecialFeature',
     'HasThemeSong', 'HasThemeVideo', 'Genres', 'OfficialRatings',
-    'Tags', 'VideoTypes', 'IsHD', 'Is4K', 'Is3D', 'Years'
+    'Tags', 'VideoTypes', 'IsSD', 'IsHD', 'Is4K', 'Is3D',
+    'IsFavorite', 'IsMissing', 'IsUnaired', 'ParentIndexNumber',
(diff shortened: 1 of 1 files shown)


src/scripts/settings/userSettings.js
```diff
@@ -521,13 +539,17 @@ export class UserSettings {
...
+
+        if (sortSettings) {
+            sortSettings = filterQuerySettings(JSON.parse(sortSettings), allowedSortSettings);
+        }
+        if (filterSettings) {
+            filterSettings = filterQuerySettings(JSON.parse(filterSettings), allowedFilterSettings);
         }
 
-        return query;
+        return Object.assign(query, sortSettings, filterSettings);
```
thornbill, line 552: Sonarcloud marks `Object.assign` as a code smell. We should use the spread operator instead.

thornbill: :+1: good call

thornbill (approved):

aaemon:
Is it fixed in the new 10.10.1 release? I could not find this in the changelog. Thanks.

PR #6276, opened 2024-10-30
Update plugin category labels

thornbill:
**Changes**
Prepares for new plugin categories

**Issues**
N/A


thornbill:
Maybe this should be deferred to 10.11 since we are lacking good translations? We just can't really update the plugin categories in advance unless we are okay with them not having translations :man_shrugging:

nielsvanvelzen:
We already need to backport translations for the "ask to skip" media segment action, so if we merge this "in time" we could include the translations updates for this PR as well. Although I understand it might be better to defer that to 10.11.

nielsvanvelzen (approved):

Issue #6280, opened 2024-10-31
Device Icons for WebOS & Updated Infuse Naming


JPKribs:
**Describe The Bug**
Infuse shows up as a generic icon in the Dashboard and webOS does as well. Infuse should use the Apple Icon and webOS should have a webOS icon.

**Steps To Reproduce**
1. Go to Devices
2. Observe LG TVs and Infuse clients

**Expected Behavior**
Jellyfin-Web is looking for Infuse which they decided to change the name. So, it should use Infuse-Direct or Infuse-Library. WebOS should be parsed as it's own type instead of using generic icon.

**Logs**
N/A

**Screenshots**
![Image](https://github.com/user-attachments/assets/ea2872bb-6c52-4085-aa3b-919b71151837)
![Image](https://github.com/user-attachments/assets/d63169f4-ab3f-4077-aacc-35dbe9691c32)

**System (please complete the following information):**
 - Platform: Jellyfin-Web
 - Browser: Firefox, Chrome, Safari
 - Jellyfin Version: 10.10.0

**Additional Context**
I have a PR for this!

JPKribs:
PR for this can be found here: https://github.com/jellyfin/jellyfin-web/pull/6281

PR #6281, opened 2024-10-31
Add Device Icons for WebOS & Updated Infuse Naming

JPKribs:
**Changes**
First time doing anything on Web so please let me know if this isn't right! 

[Over in Swiftfin](https://github.com/jellyfin/Swiftfin/pull/1275), I added Device Logos and I threw in a webOS (LG TVs) icon as well. I just wanted to add that logo and the logic so Jellyfin-Web had it as well. In my testing, 'LG Smart TV' covers all of the TVs (all 2 I had to test) but I'm unsure if there is a different name for projectors?

I am using webOS's icon from here: https://en.wikipedia.org/wiki/WebOS#/media/File:LG_WebOS_New.svg.

Additionally, during my Swiftfin changes, I found that Infuse doesn't use 'Infuse' as a device name after 7.7(?) and now uses 'Infuse-Direct' or 'Infuse-Library' on later versions. While I was adding WebOS, I added this change as well. I left the original just plain 'Infuse' for anyone running previous versions as well.

**Issues**
https://github.com/jellyfin/jellyfin-web/issues/6280


viown:
I get this when I try the icon:

image

JPKribs:
> I get this when I try the icon:
> 
> image

I'm sorry about that! I think I have this resolved. Additionally, while I was looking at my devices I realized Swiftfin wasn't in there on web for Apple either! So I've added both variants of Swiftfin. It looks like 1.0.1 and below is just Jellyfin tvOS/iOS/iPadOS and 1.1+ is Swiftfin tvOS/iOS/iPadOS. Additionally, I put the client name in for LG TV since it looks like we're looking at both client and device.

Let me know if this scaling looks right! I added some items next to it for comparison:

Untitled

JPKribs:
Oh! Lastly, the Samsung.svg should be Samsungtv.svg. I updated that too.

thornbill:
I don't think we should add padding to the logo image to fix what is really a bug with the css. (I also don't think that fixing that bug is a blocker for this PR. :slightly_smiling_face:)

JPKribs:
> I don't think we should add padding to the logo image to fix what is really a bug with the css. (I also don't think that fixing that bug is a blocker for this PR. 🙂)

Sounds good! I think on my last iteration the WebOS logo should match the same size as the other logos. Would you rather I trim that to just be the size of the actual logo?

Let me know what works best! Aside from CSS changes, that might be a little beyond me haha.

thornbill:
Yeah I would just make it the size of the logo itself. I had a fix for the css the other day but I think I deleted it at some point. I'll open a PR for it when I get it fixed again. 🙃

JPKribs:
There we go! I kept updating the image but GitHub was showing it with huge vertical margins. I think that might be just how GitHub displays svgs? Either way, I've updated the image so it should just be exactly their [Wikipedia SVG](https://upload.wikimedia.org/wikipedia/commons/5/59/LG_WebOS_New.svg) but white to mirror our existing logos.

I'll also directly attach the svg to this comment just in case something weird is going on.

thornbill (approved):

PR #6282, opened 2024-11-01
Add content hash to CSS files

viown:
**Changes**
Add a content hash to CSS files to prevent caching issues

**Issues**


nielsvanvelzen (approved):
This seems to fix the issue, network inspector shows the loaded CSS files now have hashes in the name.

Can you change the target branch to release-10.10.z and add the "stable backport" label?

PR #6295, opened 2024-11-04
Enable TV Support for 'Ask to Skip'

viown:
**Changes**
Enables TV Support for the skip button

**Issues**

### DIFF
--- a/src/components/playback/skipsegment.ts
+++ b/src/components/playback/skipsegment.ts
@@ -38,25 +38,25 @@ class SkipSegment extends PlaybackSubscriber {
 
     createSkipElement() {
         if (!this.skipElement && this.currentSegment) {
-            let buttonHtml = '';
+            const elem = document.createElement('button');
+            elem.classList.add('skip-button');
+            elem.classList.add('hide');
+            elem.classList.add('skip-button-hidden');
+            elem.setAttribute('is', 'emby-button');
 
-            buttonHtml += '<button is="emby-button" class="skip-button hide skip-button-hidden"></button>';
-
-            document.body.insertAdjacentHTML('beforeend', buttonHtml);
-
-            this.skipElement = document.body.querySelector('.skip-button');
-            if (this.skipElement) {
-                this.skipElement.addEventListener('click', () => {
-                    const time = this.playbackManager.currentTime() * TICKS_PER_MILLISECOND;
-                    if (this.currentSegment?.EndTicks) {
-                        if (time < this.currentSegment.EndTicks - TICKS_PER_SECOND) {
-                            this.playbackManager.seek(this.currentSegment.EndTicks);
-                        } else {
-                            this.hideSkipButton();
-                        }
+            elem.addEventListener('click', () => {
+                const time = this.playbackManager.currentTime() * TICKS_PER_MILLISECOND;
+                if (this.currentSegment?.EndTicks) {
+                    if (time < this.currentSegment.EndTicks - TICKS_PER_SECOND) {
+                        this.playbackManager.seek(this.currentSegment.EndTicks);
+                    } else {
+                        this.hideSkipButton();
                     }
-                });
-            }
+                }
+            });
+
+            document.body.appendChild(elem);
+            this.skipElement = elem;
         }
     }
 
@@ -83,8 +83,8 @@ class SkipSegment extends PlaybackSubscriber {
 
             void elem.offsetWidth;
 
-            const osdVisible = document.activeElement && focusManager.isCurrentlyFocusable(document.activeElement);
-            if (options.focus && !osdVisible) {
+            const hasFocus = document.activeElement && focusManager.isCurrentlyFocusable(document.activeElement);
+            if (options.focus && !hasFocus) {
                 focusManager.focus(elem);
             }
 
--- a/src/components/playback/skipsegment.ts
+++ b/src/components/playback/skipsegment.ts
@@ -9,14 +9,16 @@ import './skipbutton.scss';
 import dom from 'scripts/dom';
 import globalize from 'lib/globalize';
 import * as userSettings from 'scripts/settings/userSettings';
+import focusManager from 'components/focusManager';
 
 interface ShowOptions {
     animate?: boolean;
     keep?: boolean;
+    focus?: boolean;
 }
 
 class SkipSegment extends PlaybackSubscriber {
-    private skipElement: HTMLButtonElement | undefined;
+    private skipElement: HTMLButtonElement | null | undefined;
     private currentSegment: MediaSegmentDto | null | undefined;
     private hideTimeout: ReturnType<typeof setTimeout> | null | undefined;
 
@@ -78,6 +80,10 @@ class SkipSegment extends PlaybackSubscriber {
 
             void elem.offsetWidth;
 
+            if (options.focus) {
+                focusManager.focus(elem);
+            }
+
             requestAnimationFrame(() => {
                 elem.classList.remove('skip-button-hidden');
 
@@ -116,7 +122,8 @@ class SkipSegment extends PlaybackSubscriber {
             if (isOpen) {
                 this.showSkipButton({
                     animate: false,
-                    keep: true
+                    keep: true,
+                    focus: false
                 });
             } else if (!this.hideTimeout) {
                 this.hideSkipButton();
@@ -140,7 +147,7 @@ class SkipSegment extends PlaybackSubscriber {
 
             this.setButtonText();
 
-            this.showSkipButton({ animate: true });
+            this.showSkipButton({ animate: true, focus: true });
         }
     }
 
@@ -165,6 +172,7 @@ class SkipSegment extends PlaybackSubscriber {
     onPlaybackStop() {
         this.currentSegment = null;
         this.hideSkipButton();
+        this.skipElement = null;
         Events.off(document, EventType.SHOW_VIDEO_OSD, this.onOsdChanged);
     }
 }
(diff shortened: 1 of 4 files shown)


dmitrylyzo:
1. I think we should focus the skip button if OSD is hidden.
2. Found bug: [`_skipElement`](https://github.com/jellyfin/jellyfin-web/blob/ab1cf1e435c1a8e09aef8a16398ea6d82ff685e7/src/components/playback/skipsegment.ts#L36) is not reattached on the second video playback. _Separate PR._
    1. Start video playback.
    2. Wait for the Skip button to appear.
    3. Stop playback (return to the library/item).
    4. Resume video playback. Button is not attached to the DOM.

viown:
> I think we should focus the skip button (on show) if OSD is hidden.

I agree, it's not intuitive to have to click up two times when the button appears.

> Found bug: [_skipElement](https://github.com/jellyfin/jellyfin-web/blob/ab1cf1e435c1a8e09aef8a16398ea6d82ff685e7/src/components/playback/skipsegment.ts#L36) is not reattached on the second video playback. Separate PR.

Thanks! This is an issue with just this PR because `videoOsdPage` is unmounted when playback stops and it originally only created the button once in the body and re-used it on second playback.

src/components/playback/skipsegment.ts
```diff
@@ -140,7 +148,10 @@ class SkipSegment extends PlaybackSubscriber {
 
             this.setButtonText();
 
-            this.showSkipButton({ animate: true });
+            this.showSkipButton({
+                animate: true,
+                focus: browser.tv || browser.xboxOne || browser.ps4
```
dmitrylyzo, line 153: Use `layoutManager.tv` instead. For example, you can use TV display mode on PC.

dmitrylyzo:
@viown Could you try changing the skip button to `emby-button`?
> We should use emby-button to highlight the focus state. There is no (native) focus indication in webOS 1.2.

Alternatively, we can add `focus` and `hover` (https://github.com/jellyfin/jellyfin-web/pull/6364) styles. But `emby-button` is already styled.

Also, don't focus the skip button when the OSD is visible.

viown:
Yes, to do this I will have to move the button to the OSD template, which is probably what I should've done to begin with.

Also, the button should not focus when the OSD is visible. Do you see different results in your testing?

dmitrylyzo:
> Do you see different results in your testing?

Yes, it grabs the focus when I navigate the playback control buttons (to keep the OSD visible).

With this clip: https://github.com/user-attachments/assets/95ac625f-a41d-46d1-9125-1f6c4356552a
1. Switch to the TV display mode.
2. Enable `Ask to Skip` for `Intro` and `Outro` only.
3. Start that clip.
4. Seek to the middle.
5. Use the Left/Right buttons to move between the playback control buttons (to keep OSD visible).
6. When `Outro' begins, the skip button grabs the focus.

viown:
Updated to use emby-button. Your mentioned issue should be fixed. I also noticed that the button would cause the video to pause on click (just this PR), so that was fixed as well.

Given how complex this PR became, I wonder if it still makes sense for a 10.10.z release?

dmitrylyzo:
1. `this.skipElement` seems to become invalid when the next episode starts playing.
Start playlist with 2 video items with chapters.
Wait for the skip button to appear.
Click `Next track`.
The skip button doesn't appear unless you refresh the web page.

2. [This code](https://github.com/jellyfin/jellyfin-web/blob/2b2d41adce675aea3c51f575710d087b7607e0ea/src/controllers/playback/video/index.js#L1240) pauses playback when using `Enter` on the skip button (in the TV mode).
As a quick fix:
```js
if (e.target.tagName !== 'BUTTON') {
}
```
We are losing the ability to pause using the `OK' button on the TV remote control, so we need to rethink the skip button focus logic and/or UI.

> Given how complex this PR became, I wonder if it still makes sense for a 10.10.z release?

Alternatively, we could try moving the skip button back from the video OSD and block the focus change by checking the current `document.activeElement`. Don't change the focus if `activeElement` is a valid focusable element.
The problem is that `activeElement` can be `undefined`, `body` or the last focused element when out of focus when hiding an element. So it should be checked with care, but [isCurrentlyFocusable](https://github.com/jellyfin/jellyfin-web/blob/2b2d41adce675aea3c51f575710d087b7607e0ea/src/components/focusManager.js#L95) is probably sufficient.

src/components/playback/skipsegment.ts
```diff
@@ -9,49 +9,54 @@ import './skipbutton.scss';
...
-            elem.addEventListener('click', () => {
-                const time = this.playbackManager.currentTime() * TICKS_PER_MILLISECOND;
-                if (this.currentSegment?.EndTicks) {
-                    if (time < this.currentSegment.EndTicks - TICKS_PER_SECOND) {
-                        this.playbackManager.seek(this.currentSegment.EndTicks);
-                    } else {
-                        this.hideSkipButton();
+            let buttonHtml = '';
+
+            buttonHtml += '<button is="emby-button" class="skip-button hide skip-button-hidden"></button>';
```
dmitrylyzo, line 43: You can use `createElement` approach and add `elem.setAttribute('is', 'emby-button');`:
https://github.com/jellyfin/jellyfin-web/blob/9d7532ceabb47bca34e535f96cfc68efad0c52c1/src/controllers/session/login/index.js#L301

src/components/playback/skipsegment.ts
```diff
@@ -78,6 +83,11 @@ class SkipSegment extends PlaybackSubscriber {
 
             void elem.offsetWidth;
 
+            const osdVisible = document.activeElement && focusManager.isCurrentlyFocusable(document.activeElement);
+            if (options.focus && !osdVisible) {
```
dmitrylyzo, line 87: A more appropriate name, imo. 
```suggestion
            const hasFocus = document.activeElement && focusManager.isCurrentlyFocusable(document.activeElement);
            if (options.focus && !hasFocus) {
```

dmitrylyzo (commented):
A few minor notes. Not tested yet.

dmitrylyzo (commented):
Almost fine.
1. There's a problem with CustomElements (see below). _Major issue._
2. The focus is lost after we hide the skip button.
_This is a global problem. If we don't find a good concept for changing focus after hiding the current element, I think we can leave it to the next version - the user can use `Back` to hide the OSD and show it again using the navigation keys._

src/components/playback/skipsegment.ts
```diff
@@ -9,35 +9,40 @@ import './skipbutton.scss';
...
-        }
-    }
-
     createSkipElement() {
         if (!this.skipElement && this.currentSegment) {
             const elem = document.createElement('button');
             elem.classList.add('skip-button');
             elem.classList.add('hide');
             elem.classList.add('skip-button-hidden');
+            elem.setAttribute('is', 'emby-button');
```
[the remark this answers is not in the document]
viown: Yeah, I noticed that `document.createElement('button', { is: 'emby-button' });` didn't work and went with the string approach. I've reverted that commit for now.

viown:
Yeah, I noticed that focus was lost after clicking the button, but wasn't sure what the best way to go about it was.

Since skipping will summon the OSD panel, we might be able to change the focus to the play/pause button. Adding `focusManager.focus(document.body.querySelector('.btnPause'));` after [this line](https://github.com/jellyfin/jellyfin-web/blob/dc12d8c6bc5d11ba69f98eb071586a3c22461254/src/components/playback/skipsegment.ts#L53) seems to work.

src/controllers/playback/video/index.js
```diff
@@ -1237,7 +1237,9 @@ export default function (view) {
                     }
                     return;
                 case 'Enter':
-                    playbackManager.playPause(currentPlayer);
+                    if (e.target.tagName !== 'BUTTON') {
+                        playbackManager.playPause(currentPlayer);
+                    }
                     showOsd(btnPlayPause);
```
dmitrylyzo, line 1243: We can also keep the OSD hidden. Some users are even against displaying the OSD on pause, but that's a question for UI designers.
```suggestion
                        showOsd(btnPlayPause);
                    }
```

dmitrylyzo (approved):
Not sure what to do with focus (focus loss can happen not only on the video OSD), but we can probably fix that later. As workaround: users can hide the OSD with `Back`/`Escape` buttons and regain focus when it reappears.

rlauuzo:
To address the focus issue, you can make the following changes:  

1. **Update lines 1254-1256 in `index.js`** to handle the `Enter` key differently based on the element triggering the event:  
   ```javascript
   case 'Enter':
       showOsd(e.target.classList.contains('skip-button') ? btnPlayPause : undefined);
       break;
   ```  

2. **Prevent the skip button from losing focus** by modifying lines 357-359:  
   ```javascript
   if (document.activeElement && !document.activeElement.classList.contains('skip-button')) {
       document.activeElement.blur();
   }
   ```

dmitrylyzo:
> 1. **Update lines 1254-1256 in `index.js`** to handle the `Enter` key differently based on the element triggering the event:

I think it would be better to restore it here: https://github.com/jellyfin/jellyfin-web/blob/5f105696b4ca7a3492d97ee828393ba89b1842c2/src/controllers/playback/video/index.js#L341-L343
We need to remove `focusElement` from the condition. If `!focusElement` and `document.activeElement` is not focusable, use `btnPause` as the focus element.
```js
 } else if (currentVisibleMenu === 'osd' && !layoutManager.mobile) {
    if (!focusElement && document.activeElement && !focusManager.isCurrentlyFocusable(document.activeElement) {
        focusElement = osdBottomElement.querySelector('.btnPause');
    }
    _focus(focusElement); 
 }
```
_Not tested._

Iirc, we also need to add range input (playback progress bar) as focusable element to `isCurrentlyFocusable`. The skip button grabs focus from it.

> 2\. **Prevent the skip button from losing focus** by modifying lines 357-359:

We should check that `document.activeElement` is a child of `osdBottomElement`.

PR #6317, opened 2024-11-13
Fix secondary audio on Tizen 8

dmitrylyzo:
There are reports that additional audio track (AudioTrack API) doesn't work on Tizen 8.
_Did Samsung break the AudioTrack API?_ 🤔 

**Changes**
Don't use AudioTrack API on Tizen 8+.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-tizen/issues/293


dmitrylyzo:
This may become unnecessary if Samsung fixes the AudioTrack API.

kimahry:
Tested on TV and it fixed my issue.
Will this be merged ?

thornbill (approved):

Snuggerdog:
Still experiencing issues changing audio track after using the 10.10.z build. I select the alternate track plays in the original audio. 

TV model Samsung Q70AA

PR #6320, opened 2024-11-14
Resolve card hover state appearing between items

LittleBigOwI:
When navigating the UI, a bug caused cards to display as "hovered" even when the cursor was positioned between two cards.

#### Changes
- Adjusted the margin between cards and set gap in css.

#### Issues
- Couldn't find matching issue for this.


dmitrylyzo:
The problem is that `gap` doesn't work on older web engines (Tizen, webOS).

LittleBigOwI:
Okay - then what do you think of adding a spacer div to the card layout? This should work on all web engines.

dmitrylyzo:
> Okay - then what do you think of adding a spacer div to the card layout? This should work on all web engines.

Yeah, I was thinking of wrapping the cards with a "card cell" (div with margins), but that would require revising every "card generator" (unfortunately, there are several places where cards are constructed), and I'm not sure there's no logic that strictly requires the `card` class to be the topmost.

"Card generators": [1](https://github.com/jellyfin/jellyfin-web/blob/1fd7bf0c5192ffa316e256a41f6a6616b547db31/src/components/cardbuilder/cardBuilder.js#L878), [2](https://github.com/jellyfin/jellyfin-web/blob/1fd7bf0c5192ffa316e256a41f6a6616b547db31/src/components/cardbuilder/chaptercardbuilder.js#L84), [3](https://github.com/jellyfin/jellyfin-web/blob/1fd7bf0c5192ffa316e256a41f6a6616b547db31/src/controllers/session/login/index.js#L150-L178), [4](https://github.com/jellyfin/jellyfin-web/blob/1fd7bf0c5192ffa316e256a41f6a6616b547db31/src/components/tunerPicker.js#L49-L65), [5](https://github.com/jellyfin/jellyfin-web/blob/1fd7bf0c5192ffa316e256a41f6a6616b547db31/src/components/cardbuilder/Card/Card.tsx#L14)

Issue #6322, opened 2024-11-16
Preview "Ask To Skip" - No Button


AbandonedCart:
**Describe The Bug**
Enabling "Ask To Skip" for "Preview Segments" doesn't enable a skip button

**Steps To Reproduce**

1. Generate a preview segment (dealer's choice)
2. Open the client settings
3. Select "Playback"
4. Scroll to "Preview Segments"
5. Enable "Ask To Skip"
6. Save
7. Play media file with know preview segment
8. Observe nothing happening during preview

**Expected Behavior**
A button similar to the intro and outro button to appear

**Logs**
No errors encountered.

**Screenshots**
![Image](https://github.com/user-attachments/assets/832bf1f0-b728-4270-849e-0e394981bc5d)

**System (please complete the following information):**
 - Platform: macOS / Windows / Android
 - Browser: Firefox / Jellyfin Media Player
 - Jellyfin Version: 10.10.1

**Additional Context**
Tested in Firefox on Mac, Windows, and Android. Tested in Jellyfin Media Player on Windows.
Enabling the "Skip" option appears to function normally.

viown:
This works as expected on my end in 10.10.1:

![Image](https://github.com/user-attachments/assets/ce47674d-1eaf-4947-9896-d315ec4ea1c9)

Have you confirmed the segment is within the expected range and is of type "Preview" (2)?

AbandonedCart:
> Have you confirmed the segment is within the expected range and is of type "Preview" (2)?

Yes

> Do you also see a "found Preview segment" log message in console?

Server? Browser? Can you be more specific?

AbandonedCart:
I did some messing around and realized the problem is that the labeling is convoluted.

In television, a recap typically recaps the season so far and a preview previews the upcoming episode.

In jellyfin, a preview is at the beginning (probably attributed to movie previews) and a recap is at the end (I don't have good logic for why anyone thought a show or movie recaps itself)

PR #6323, opened 2024-11-16
Indicate how overlapping prompt is handled

AbandonedCart:
Adds a note to inform the user that an overlapping segment prompt will be discarded in favor of the "Show next video info during playback" prompt.

**Changes**
Clarifies the action already taken to avoid any confusion, especially for anyone that stops or manually skips before seeing the delayed prompt and doesn't come to the realization themselves.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-web/issues/6322 (by preventing similar confusion)


viown:
Maybe we should update the text in the up next dialog to properly indicate the segment.

AbandonedCart:
> Maybe we should update the text in the up next dialog to properly indicate the segment.

I'm not sure I understand the purpose of that.

AbandonedCart:
The wording may be confusing for some.

Other possible ways to word it may be:

- This dialog will replace ask to skip when displayed during a segment.
- Segments at the end of a video will only display this dialog.
- Some other version?

PR #6324, opened 2024-11-17
Fix spinner direction (should be clockwise) in @keyframes spin

carsso:
**Changes**
Fixing spinner direction (should be clockwise) and removing duplicate animation `@keyframes spin`

Previous behaviour (spinning in wrong direction, anti-clockwise):

### DIFF
--- a/src/styles/videoosd.scss
+++ b/src/styles/videoosd.scss
@@ -172,14 +172,8 @@
     margin-left: auto;
 }
 
-@keyframes spin {
-    100% {
-        transform: rotate(360deg);
-    }
-}
-
 .osdMediaStatus .animate {
-    animation: spin 4s linear infinite;
+    animation: spin-clockwise 4s linear infinite;
 }
 
 @media all and (max-width: 30em) {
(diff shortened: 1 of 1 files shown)


src/styles/videoosd.scss
```diff
@@ -172,12 +172,6 @@
     margin-left: auto;
 }
 
-@keyframes spin {
```
dmitrylyzo, line 175: I would keep this one and remove the one below - less changes and the animation is defined before use.

dmitrylyzo:
Since this is a bugfix, could you please retarget it to the `release-10.10.z` branch? (https://github.com/jellyfin/jellyfin-web/pull/5680#issuecomment-2156018605)

src/styles/videoosd.scss
carsso: I see all `@keyframes` are defined at the end of the file.
I'll create two explicite spin animations, one for clockwise, one for counterclockwise, and use each animation in the right class.

carsso:
Target changed to `release-10.10.z `.
Fix updated to handle both spinning directions (one for SyncPlay icon that is counterclockwise, one for player data fetching icon that is clockwise).

carsso:
Works beatifuly for both on preview.

Player data fetching icon :

SyncPlay icon :

thornbill (approved):

PR #6325, opened 2024-11-18
Add show password toggle button

FrancescoLanthaler:
Add a "show password" button on login screen

**Changes**
Add show password toggle button to login screen. Changes input field type to text when clicked. On submit, field type is turned back to password.

**Issues**
[Feature](https://features.jellyfin.org/posts/2924/add-a-show-password-button-on-login-screen)


tylers-username:
Considering security, as a rule, we probably generally want to avoid the use of `.innerHtml =`. Since it is being set to static text, it's fine.

Considering best practices, the `span` that toggles preview text should probably be a `button` since it is a call-to-action, clickable element.

Other than that, this MR should be good once checks pass.

src/controllers/session/login/login.scss
```diff
@@ -3,6 +3,27 @@
     margin-top: 2em;
 }
 
+.inputInner {
+    position: relative;
+    height: fit-content;
+
+    label {
```
dfederm, line 10: This doesn't match anything since you have the label as a sibling, not a child

src/controllers/session/login/index.html
```diff
@@ -5,11 +5,16 @@
...
             <div class="inputContainer">
-                <input is="emby-input" type="text" id="txtManualName" required="required" label="${LabelUser}" autocomplete="username" autocapitalize="off" />
+                <input is="emby-input" type="text" id="txtManualName" required="required" label="${LabelUser}"
+                    autocomplete="username" autocapitalize="off" />
             </div>
 
-            <div class="inputContainer">
-                <input is="emby-input" id="txtManualPassword" type="password" label="${LabelPassword}" autocomplete="current-password" />
+            <div class="inputContainer passwordInput">
+                <label for="txtManualPassword">${LabelPassword}</label>
```
dfederm, line 13: The changes made to the password label make it inconsistent with the username label. Notably the color.

image
image

I believe this is because you removed `label="${LabelPassword}"`, which the `empy-input` uses to set the classes `inputLabelFocused` and `inputLabelUnfocused`.

src/controllers/session/login/index.html
```diff
@@ -5,11 +5,16 @@
...
+                <div class="inputInner">
+                    <input is="emby-input" id="txtManualPassword" type="password" autocomplete="current-password" />
+                    <span class="togglePasswordVisibility material-icons">visibility</span>
```
dfederm, line 16: Consider using a `<button type="button">` instead of `<span>` for the toggle to provide proper keyboard accessibility (focusable, responds to Enter/Space).

PR #6327, opened 2024-11-18
Move strict mode to page component

thornbill:
**Changes**
* Moves `StrictMode` to the `Page` component instead of the `toAsyncPageRoute` function
* Adds support for routes returning `RouteObject`s directly (similar to the examples in the [official documentation for lazy](https://reactrouter.com/en/main/route/lazy))

**Issues**
N/A


1hitsong (approved):
Tested in Waterfox on Windows 11. Noticed no changes.

PR #6329, opened 2024-11-19
Fix changing lyric fetcher order not changeable in library options

nielsvanvelzen:
**Changes**
- Fix changing lyric fetcher order not changeable in library options (event listener was missing)
- Fix help text about lyrics

**Issues**


crobibero (approved):
disney-tarzan-tantor

thornbill (approved):

PR #6330, opened 2024-11-19
Add media segment providers to library options

nielsvanvelzen:
**Changes**

- Add a new "media segment providers" section to the library options
 - ~~Requires an API change that fixes a typo in jellyfin/jellyfin#13097~~
 - Relies on the server to send the correct available providers for each library type
   - This means our chapter provider shows up for music libraries right now (because audio can have chapters)

**Screenshots**

91900004a0

**Issues**

~~Note: Will cause merge-conflict with #6329 when backporting~~


jumoog:
It think you forgot `mixed` libraries

nielsvanvelzen:
> It think you forgot `mixed` libraries

Good catch, I've updated it to not check the library type locally but instead rely on the server to send the accurate providers via the "available options" endpoint.

thornbill (approved):

PR #6332, opened 2024-11-20
Add detail page link in context menu

othmar52:
**Changes**
Currently it is impossible to navigate to the detail page of a playlist item.  
this PR adds a detail page link to items context menu:  

image

**Issues**
no known iussues

### DIFF
--- a/src/components/itemContextMenu.js
+++ b/src/components/itemContextMenu.js
@@ -294,6 +294,15 @@ export async function getCommands(options) {
         });
     }
 
+    if (canPlay) {
+        commands.push({
+            name: globalize.translate('ShowDetails'),
+            id: 'navigatetodetailpage',
+            icon: 'remove_red_eye'
+        });
+    }
+
+
     if (item.PlaylistItemId && options.playlistId && options.canEditPlaylist) {
         commands.push({
             name: globalize.translate('RemoveFromPlaylist'),
@@ -630,6 +639,10 @@ function executeCommand(item, id, options) {
                 break;
             case 'cancelseriestimer':
                 deleteSeriesTimer(apiClient, item, resolve, id);
+                break;
+            case 'navigatetodetailpage':
+                appRouter.showItem(item, item.ServerId);
+                getResolveFunction(resolve, id)();
                 break;
             default:
                 reject();
--- a/src/components/itemContextMenu.js
+++ b/src/components/itemContextMenu.js
@@ -294,7 +294,7 @@ export async function getCommands(options) {
         });
     }
 
-    if (canPlay) {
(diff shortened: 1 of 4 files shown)


src/strings/de.json
```diff
@@ -991,6 +991,7 @@
     "SettingsWarning": "Das Verändern dieser Werte kann Instabilität und Verbindungsprobleme hervorrufen. Wenn Probleme auftreten sollten, empfehlen wir, diese Einstellungen auf die Standardwerte zurückzustellen.",
     "Share": "Teilen",
     "ShowAdvancedSettings": "Erweiterte Einstellungen anzeigen",
+    "ShowDetails": "Details anzeigen",
```
dmitrylyzo, line 994: Languages other than `en-us` should only be updated with Weblate (after PR merge).
```suggestion
```

src/strings/en-gb.json
```diff
@@ -432,6 +432,7 @@
     "ShowYear": "Show the year",
     "ShowIndicatorsFor": "Show indicators for",
     "ShowAdvancedSettings": "Show advanced settings",
+    "ShowDetails": "Show details",
```
dmitrylyzo, line 435: Languages other than `en-us` should only be updated with Weblate (after PR merge).
```suggestion
```

src/components/itemContextMenu.js
```diff
@@ -294,6 +294,14 @@ export async function getCommands(options) {
         });
     }
 
+    if (canPlay) {
+        commands.push({
+            name: globalize.translate('ShowDetails'),
+            id: 'navigatetodetailpage',
```
dmitrylyzo, line 300: Maybe call id `itemdetails` like `artist`, `album` and `lyrics`?

othmar52: @dmitrylyzo thanks for your feedback!  
I've realized I forgot to hide the "Details" link when the user is already on the details page. I tried searching the source for similar `appRouter` conditions but couldn't find anything relevant.

Could you point me in the right direction?

othmar52:
@dmitrylyzo thanks for your feedback!

I've realized I forgot to hide the "Details" link when the user is already on the details page. I tried searching the source for similar appRouter conditions but couldn't find anything relevant.

Could you point me in the right direction?

thornbill:
I don't think adding more options to the context menu is a great solution for this. The menu already holds an unmanageable number of options that often overflow the screen on smaller devices.

In other places, you can already navigate to the details page by clicking on text links or an "info" icon in the list view. I would prefer we do something similar here.

thornbill:
Closing as stale, but feel free to reopen if you address the feedback above!

PR #6334, opened 2024-11-20
Add branding settings page

thornbill:
**Changes**
* Moves branding settings to a separate dashboard page
* Fixes some styling issues with the `emby-textarea` component

Screenshot 2024-11-20 at 18-10-42 Jellyfin

**Issues**
N/A

### DIFF
--- a/src/apps/dashboard/components/drawer/sections/ServerDrawerSection.tsx
+++ b/src/apps/dashboard/components/drawer/sections/ServerDrawerSection.tsx
@@ -1,4 +1,5 @@
 import { Dashboard, ExpandLess, ExpandMore, LibraryAdd, People, PlayCircle, Settings } from '@mui/icons-material';
+import Palette from '@mui/icons-material/Palette';
 import Collapse from '@mui/material/Collapse';
 import List from '@mui/material/List';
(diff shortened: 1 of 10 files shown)


nielsvanvelzen:
We do have an endpoint to upload a custom splash screen to, perhaps we can add support for that in this new branding page?

thornbill:
Yeah I was thinking about adding that in a separate PR to keep this one smaller. 😉

nielsvanvelzen (approved):
Everything seems to work as expected. Some feedback for future improvements:

- Allow uploading the splash image(mentioned in earlier comment)
- Displaying the current splash image
- Maybe rename "splash screen" to "server background image" or something similar to make it a bit more descriptive
- Previewing the login disclaimer for when markdown/html is used
  - Use a WYSIWYG editor, may be a bit overkill though
- Use a larger minimum height for both text areas so it's obvious they're not a single line input

thornbill:
> Might be worth adding this feedback here [jellyfin/jellyfin-meta#81](https://github.com/jellyfin/jellyfin-meta/discussions/81) also

These items are either already supported or do not require support by the server. I did create #6335 to track them here though.

PR #6348, opened 2024-11-26
Stop multi select on ESC key press

othmar52:
**Changes**
Pressing the ESC key on the keyboard will exit multi select mode  

**Issues**
N/A

***Personal note***

A tip for anyone who finds themselves in accidently activated multi-select mode:  
Make sure not to hold down the mouse button for more than one second when interacting with the horizontal scroll items.

### DIFF
--- a/src/scripts/keyboardNavigation.js
+++ b/src/scripts/keyboardNavigation.js
@@ -7,6 +7,7 @@ import browser from './browser';
 import inputManager from './inputManager';
 import layoutManager from '../components/layoutManager';
 import appSettings from './settings/appSettings';
+import { stopMultiSelect } from '../components/multiSelect/multiSelect';
 
 /**
  * Key name mapping.
@@ -178,6 +179,7 @@ export function enable() {
                 break;
 
             case 'Escape':
+                stopMultiSelect();
                 if (layoutManager.tv) {
                     inputManager.handleCommand('back');
                 } else {


gompa:
this works fine on preview, user is now able to exit multi select mode with esc
(to test go to the home view or a collection, press and hold a item (*optional: select multiple items) press esc to exit multiselect mode)

src/scripts/keyboardNavigation.js
```diff
@@ -178,6 +179,7 @@ export function enable() {
                 break;
 
             case 'Escape':
+                stopMultiSelect();
```
dmitrylyzo, line 182: IMO, this should be handled by the multiselect controller and prevent the default action so that the generic keyboard handler (this) [ignores the event](https://github.com/jellyfin/jellyfin-web/blob/0327f3317ef9a9c95f85c6d40fd0878cfd268995/src/scripts/keyboardNavigation.js#L128).

xanderphillips (approved):
Change appears to be correct and tight in scope.  stopMultiSelect was already exported from multiSelect.js and hideSelections() is guarded so that the call's a no-op if multi select's not active, so no cost for other ESC's.  Having it before the TV layout check lets it stop multi-select and go back, which seems right.   The tip info might belong in the wiki instead of the PR, but that's neither here nor there.  

LGTM

PR #6351, opened 2024-11-27
Experimental feature: show paths mode

othmar52:
### **Experimental Feature: Display File Paths in the Jellyfin Frontend**

**Background:**  
Currently, verifying whether media identification is accurate or problematic can be quite tedious.  
In my case, the Metadata Manager often proves ineffective, especially for users with large media collections.  
To address this, I implemented an optional feature to display file paths directly in the frontend, making it much easier to validate media identification.

---

### **Screenshots**

- **New Icon for Toggling the Mode**  
   A new icon has been added to enable or disable this mode.

   image

- **Changes to the List View of movies (when mode is active):**  
  - The year is displayed next to the title.  
  - Each word/chunk of the title is highlighted in green if it matches a string in the file or directory name.  
    - This makes it much easier to identify items that have been incorrectly matched.  
  - The last part of the full media path is displayed below the title.  

<p align="center">
list item with colored words/chunks
</p>

<p align="center">
  Screenshot: List View with deliberately mismatched items  <br />
Screenshot: List View with deliberately mismatched items
</p>

<p align="center">
  Screenshot: List View with my actual collection  <br />
  (when everything is green there should be no need to take a closer look)<br />
Screenshot: List View with my actual collection
</p>

- **Changes to the Detail View of movies (when mode is active):**  
  - File size and media path are displayed.  
  - The media's directory path can be copied to the clipboard, allowing quick navigation to the filesystem for any required changes.  

image

---

### **TODOs**

- [ ] Find a better name for "Show Paths Mode" (suggestions welcome).  
- [ ] Decide if this feature should be controlled by user policies or be admin-only.  
- [ ] Determine whether the "Copy Path" button should include the filename or only the parent directory.  
- [ ] Is controlling visibility via a body class acceptable, or is it a no-go?  
- [ ] Currently this new mode only affects movies. should this be extended to other media types?  

---

**Any suggestions or feedback are very welcome!**  
Do you think that other users like this feature as well or i am the only one that needs something like this?


thornbill:
I can this type of functionality being potentially useful as part of the metadata manager, but I would not accept it as part of the main user facing UI.

Issue #6350, opened 2024-11-27
MaxParentalRating does not retrieve current state


solidsnake1298:
**Describe The Bug**
After setting a parental rating, I can confirm that it is actually set based on the contents displayed in the library.  But when I return to the Parental Control panel it does not display the current value.

**Steps To Reproduce**

1. Go to Dashboard > Users > Select User > Parental Control tab
2. Set a value for the user
3. Save and close or click on another tab for the user profile
4. Return to the Parental Control tab.

**Expected Behavior**
The expected behavior is that the current value was be displayed after saving a setting

**Logs**
There are no server logs, but here is what is return for the user when clicking on the Parental Control tab.

```
{
    "Name": "Test",
    "ServerId": "bfcc7726262845b39bf7fd006bd56f20",
    "Id": "b9a5d83f14c5489baf93f063862e5b8a",
    "HasPassword": true,
    "HasConfiguredPassword": true,
    "HasConfiguredEasyPassword": false,
    "EnableAutoLogin": false,
    "LastLoginDate": "2024-11-26T16:12:42.4064669Z",
    "LastActivityDate": "2024-11-26T16:12:42.4064669Z",
    "Configuration": {
        "AudioLanguagePreference": "",
        "PlayDefaultAudioTrack": true,
        "SubtitleLanguagePreference": "",
        "DisplayMissingEpisodes": false,
        "GroupedFolders": [],
        "SubtitleMode": "Default",
        "DisplayCollectionsView": false,
        "EnableLocalPassword": false,
        "OrderedViews": [],
        "LatestItemsExcludes": [],
        "MyMediaExcludes": [],
        "HidePlayedInLatest": true,
        "RememberAudioSelections": true,
        "RememberSubtitleSelections": true,
        "EnableNextEpisodeAutoPlay": true,
        "CastReceiverId": "F007D354"
    },
    "Policy": {
        "IsAdministrator": false,
        "IsHidden": true,
        "EnableCollectionManagement": false,
        "EnableSubtitleManagement": false,
        "EnableLyricManagement": false,
        "IsDisabled": false,
        "MaxParentalRating": 0,
        "BlockedTags": [],
        "AllowedTags": [],
        "EnableUserPreferenceAccess": true,
        "AccessSchedules": [],
        "BlockUnratedItems": [],
        "EnableRemoteControlOfOtherUsers": false,
        "EnableSharedDeviceControl": true,
        "EnableRemoteAccess": true,
[...]

PR #6352, opened 2024-11-27
Fix maximum allowed parental rating not showing up

viown:
Fixes the maximum allowed parental rating not showing its current state.

Additional note: The 'Unrated' parental rating currently does not seem to have a value and cannot be set. Setting to Unrated is the same as setting no rating. This PR will also hide it from the options.

**Changes**
Move parental rating to a separate hook in order to load it when the parental ratings load.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-web/issues/6350
Fixes https://github.com/jellyfin/jellyfin/issues/12418

### DIFF
--- a/src/apps/dashboard/routes/users/parentalcontrol.tsx
+++ b/src/apps/dashboard/routes/users/parentalcontrol.tsx
@@ -164,13 +164,11 @@ const UserParentalControl = () => {
         populateRatings(allParentalRatings);
 
         let ratingValue = '';
-        if (user.Policy?.MaxParentalRating != null) {
-            allParentalRatings.forEach(rating => {
-                if (rating.Value != null && user.Policy?.MaxParentalRating != null && user.Policy.MaxParentalRating >= rating.Value) {
-                    ratingValue = `${rating.Value}`;
-                }
-            });
-        }
+        allParentalRatings.forEach(rating => {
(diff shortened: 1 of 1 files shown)


src/apps/dashboard/routes/users/parentalcontrol.tsx
```diff
@@ -163,15 +164,15 @@ const UserParentalControl = () => {
         populateRatings(allParentalRatings);
 
         let ratingValue = '';
-        if (user.Policy?.MaxParentalRating) {
+        if (user.Policy?.MaxParentalRating != null) {
             allParentalRatings.forEach(rating => {
-                if (rating.Value && user.Policy?.MaxParentalRating && user.Policy.MaxParentalRating >= rating.Value) {
+                if (rating.Value != null && user.Policy?.MaxParentalRating != null && user.Policy.MaxParentalRating >= rating.Value) {
```
dmitrylyzo, line 169: Isn't the second `user.Policy?.MaxParentalRating != null` unnecessary? We have checked it above.

viown: Yes, but eslint still requires it. Not sure why.

thornbill (approved):

PR #6354, opened 2024-11-27
make tonemapping helptext make sense

felix920506:
per discussion on matrix yesterday, remove unnecessary clutter from the helptext


Shadowghost (approved):

src/strings/en-us.json
```diff
@@ -955,14 +955,14 @@
     "LabelTitle": "Title",
     "LabelTonemappingAlgorithm": "Select the Tone mapping algorithm to use",
     "LabelTonemappingDesat": "Tone mapping desat",
-    "LabelTonemappingDesatHelp": "Apply desaturation for highlights that exceed this level of brightness. The higher the parameter, the more color information will be preserved. This setting helps prevent unnaturally blown-out colors for super-highlights, by (smoothly) turning into white instead. This makes images feel more natural, at the cost of reducing information about out-of-range colors. The recommended and default values are 0 and 0.5.",
+    "LabelTonemappingDesatHelp": "Apply desaturation for highlights that exceed this level of brightness. The higher the parameter, the more color information will be preserved. This setting helps prevent unnaturally blown-out colors for super-highlights, by (smoothly) turning into white instead. This makes images feel more natural, at the cost of reducing information about out-of-range colors. The recommended value is 0.",
```
joshuaboniface, line 958:
```suggestion
    "LabelTonemappingDesatHelp": "Apply desaturation for highlights that exceed this level of brightness to prevent blown-out colors for super-highlights by smoothing to white. Higher values preserve more color information. The recommended value is 0.",
```

src/strings/en-us.json
```diff
@@ -955,14 +955,14 @@
...
     "LabelTonemappingParam": "Tone mapping param",
     "LabelTonemappingParamHelp": "Tune the tone mapping algorithm. The recommended and default values are NaN. Generally leave it blank.",
     "LabelTonemappingPeak": "Tone mapping peak",
-    "LabelTonemappingPeakHelp": "Override signal/nominal/reference peak with this value. Useful when the embedded peak information in display metadata is not reliable or when tone mapping from a lower range to a higher range. The recommended and default values are 100 and 0.",
+    "LabelTonemappingPeakHelp": "Override signal/nominal/reference peak with this value. Useful when the embedded peak information in display metadata is not reliable or when tone mapping from a lower range to a higher range. The recommended value is 100.",
```
joshuaboniface, line 962:
```suggestion
    "LabelTonemappingPeakHelp": "Override signal/nominal/reference peak with this value when the embedded peak information in display metadata is not reliable or when tone mapping from a lower to a higher range. The recommended value is 100.",
```

src/strings/en-us.json
```diff
@@ -955,14 +955,14 @@
...
     "LabelTonemappingRange": "Tone mapping range",
     "LabelTonemappingThreshold": "Tone mapping threshold",
-    "LabelTonemappingThresholdHelp": "The tone mapping algorithm parameters is fine-tuned per each scene. And a threshold is used to detect whether the scene has changed or not. If the distance between the current frame average brightness and the current running average exceeds a threshold value, we would re-calculate scene average and peak brightness. The recommended and default values are 0.8 and 0.2.",
+    "LabelTonemappingThresholdHelp": "The tone mapping algorithm parameters is fine-tuned per each scene. And a threshold is used to detect whether the scene has changed or not. If the distance between the current frame average brightness and the current running average exceeds a threshold value, we would re-calculate scene average and peak brightness. The recommended value is 0.8.",
```
joshuaboniface, line 965:
```suggestion
    "LabelTonemappingThresholdHelp": "Fine-tune the tone mapping algorithm parameters per scene to detect whether the scene has changed or not by re-calculating the scene average and peak brightness if the distance between the current frame average brightness and the current running average exceeds this threshold value. The recommended value is 0.8.",
```

joshuaboniface (approved):
I actually think they can be simplified and cleaned up a lot more but this is a good start. Feel free to ignore but I think these are slightly shorter and clearer.

felix920506:
I think we should have @nyanmisaka or @gnattu check if your suggestions are still accurate as I feel like they slightly change the meaning

src/strings/en-us.json
gnattu: This value is no longer needed as we are using static tone mappers now. Probably better to just remove it

src/strings/en-us.json
gnattu: 0 means no additional desaturation. We should document that

felix920506: @gnattu what would you suggest the string be changed to?

src/strings/en-us.json
felix920506: I think this changes the meaning because the old string implies it overrides the value on all media while the suggested change implies that it is only overridden when the metadata of source media is unreliable

src/strings/en-us.json
felix920506: String removed

src/strings/en-us.json
gnattu: Just 0(disable) should be enough

src/strings/en-us.json
gnattu:
```suggestion
    "LabelTonemappingPeakHelp": "Override the embedded metadata value for the input signal with this peak value instead. The default value is 100 (1000nit).",
```

gnattu: Don't explain too much because the user still won't get what value to set.

gnattu (approved):

TimGels (approved):

src/strings/en-us.json
```diff
@@ -955,14 +955,13 @@
...
     "LabelTonemappingParamHelp": "Tune the tone mapping algorithm. The recommended and default values are NaN. Generally leave it blank.",
     "LabelTonemappingPeak": "Tone mapping peak",
-    "LabelTonemappingPeakHelp": "Override signal/nominal/reference peak with this value. Useful when the embedded peak information in display metadata is not reliable or when tone mapping from a lower range to a higher range. The recommended and default values are 100 and 0.",
+    "LabelTonemappingPeakHelp": "Override the embedded metadata value for the input signal with this peak value instead. The default value is 100 (1000nit).",
     "LabelTonemappingRange": "Tone mapping range",
     "LabelTonemappingThreshold": "Tone mapping threshold",
-    "LabelTonemappingThresholdHelp": "The tone mapping algorithm parameters is fine-tuned per each scene. And a threshold is used to detect whether the scene has changed or not. If the distance between the current frame average brightness and the current running average exceeds a threshold value, we would re-calculate scene average and peak brightness. The recommended and default values are 0.8 and 0.2.",
```
thornbill, line 965: You should also remove the usage of this?

thornbill (changes requested):

src/strings/en-us.json
felix920506: This string is already not used anywhere in web

src/strings/en-us.json
```diff
@@ -955,14 +955,13 @@
...
+    "LabelTonemappingDesatHelp": "Apply desaturation for highlights that exceed this level of brightness. The recommended value is 0 (disable).",
     "LabelTonemappingParam": "Tone mapping param",
     "LabelTonemappingParamHelp": "Tune the tone mapping algorithm. The recommended and default values are NaN. Generally leave it blank.",
     "LabelTonemappingPeak": "Tone mapping peak",
-    "LabelTonemappingPeakHelp": "Override signal/nominal/reference peak with this value. Useful when the embedded peak information in display metadata is not reliable or when tone mapping from a lower range to a higher range. The recommended and default values are 100 and 0.",
+    "LabelTonemappingPeakHelp": "Override the embedded metadata value for the input signal with this peak value instead. The default value is 100 (1000nit).",
     "LabelTonemappingRange": "Tone mapping range",
     "LabelTonemappingThreshold": "Tone mapping threshold",
```
thornbill, line 964: This also appears unused...
```suggestion
```

felix920506: This string is used here https://github.com/jellyfin/jellyfin-web/blob/75ddb0acccee8fcc941312bd8e8a00257d848104/src/controllers/dashboard/encodingsettings.html#L232

thornbill (approved):

PR #6356, opened 2024-11-28
Suggest ImageExtractionTimeoutMs changes for Idle priority

tcely:
After seeing repeated logs about `ffmpeg` being started then killed for being unresponsive, I tracked down this configuration setting on GitHub.

So far, my Trickplay generation is much more reliable with a timeout of 30,000 instead. Having a pointer to this configuration setting would have saved me a lot of confusion when my logs filled with the creating/unresponsive cycle.

Even better would be to expose that configuration setting in the UI, unless I missed it somewhere.

**Changes**

Add text pointing at the timeout configuration value that may need to be increased.

**Issues**

Fixes jellyfin/jellyfin#13116

### DIFF
--- a/src/strings/en-us.json
+++ b/src/strings/en-us.json
@@ -1780,7 +1780,7 @@
     "PriorityBelowNormal": "Below Normal",
     "PriorityIdle": "Idle",
     "LabelProcessPriority": "Process Priority",
-    "LabelProcessPriorityHelp": "Setting this lower or higher will determine how the CPU prioritizes the ffmpeg trickplay generation process in relation to other processes. If you notice slowdown while generating trickplay images but don't want to fully stop their generation, try lowering this as well as the thread count.",
+    "LabelProcessPriorityHelp": "Setting this lower or higher will determine how the CPU prioritizes the ffmpeg trickplay generation process in relation to other processes. If you notice slowdown while generating trickplay images but don't want to fully stop their generation, try lowering this as well as the thread count. Increasing the 'ImageExtractionTimeoutMs' value above 20,000 may be needed for the 'Idle' priority.",
     "LabelImageInterval": "Image Interval",
     "LabelImageIntervalHelp": "Interval of time (ms) between each new trickplay image.",
     "LabelWidthResolutions": "Width Resolutions",


src/strings/en-us.json
```diff
@@ -1780,7 +1780,7 @@
     "PriorityBelowNormal": "Below Normal",
     "PriorityIdle": "Idle",
     "LabelProcessPriority": "Process Priority",
-    "LabelProcessPriorityHelp": "Setting this lower or higher will determine how the CPU prioritizes the ffmpeg trickplay generation process in relation to other processes. If you notice slowdown while generating trickplay images but don't want to fully stop their generation, try lowering this as well as the thread count.",
+    "LabelProcessPriorityHelp": "Setting this lower or higher will determine how the CPU prioritizes the ffmpeg trickplay generation process in relation to other processes. If you notice slowdown while generating trickplay images but don't want to fully stop their generation, try lowering this as well as the thread count. Increasing the 'ImageExtractionTimeoutMs' value above 20,000 may be needed for the 'Idle' priority.",
```
gnattu, line 1783: I don’t believe a specific numerical value accurately represents the current situation. The idle priority will halt all trickplay generation when the server experiences moderate load, which could potentially take significantly longer than 20000 milliseconds.

tcely: Yes, giving a hint on what value to change and the idea that it must be higher than the current value were the only goals for this message.

My own system has been happily using Idle with 30,000. An ultimate fix for this is to not impose an arbitrary wall clock limit on every image generation loop. There seems to be opposition to that for reasons that I don't understand.

In the current situation, we should at least let the users know which value to adjust.

gnattu: > My own system has been happily using Idle with 30,000. An ultimate fix for this is to not impose an arbitrary wall clock limit on every image generation loop. There seems to be opposition to that for reasons that I don't understand.

It is not that hard to understand. The server cannot tell if the extraction is really hanging or just took too long. Because ffmpeg is not bug-free and user's video files are not always technically correct, and in the most unideal case the ffmpeg will just sit there entering endless loop so that the image extraction is never finished. A timeout is a must to make the extraction process to "guaranteed to complete". 

My ultimate suggestion would be removing the idle level entirely. It is not designed to work with this priority and it should not be set with this priority.

tcely: A timeout is required. That part is understood. However, why is it not sufficient that any image was created? After the first image, `ffmpeg` is known to be working well enough. Is that really likely to end in a state that requires killing the process?

I don't understand why it is required that a new image was created every time the loop checks.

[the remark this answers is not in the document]
tcely: A suboptimal condition may arise. What's the user to do?

gnattu: > However, why is it not sufficient that any image was created?

Because hanging can occur at any frame, not just the initial frame. Your operating system does similar things: it performs periodic checks to determine if a window is not responsible instead of assuming that an application will remain in a good state indefinitely after startup. 

In jellyfin, not only image extractor does this. Our media transcoder does the similar things. The server's transcoding handler periodically pings the transcoder and if the ping timed out the transcoder process will get killed as well.

PR #6359, opened 2024-12-01
Add new checkbox on user setting to set the permission EnableNoPassword

mirillow:
I'm adding a new checkbox to set the new permission made in the server's PR github.com/jellyfin/jellyfin/pull/13135
I believe it's better to look at that PR first.

**Changes**

- Added a new checkbox
- Included the new policy EnableNoPassword


src/strings/pt-br.json
```diff
@@ -859,6 +859,7 @@
     "OptionDatePlayed": "Data de Reprodução",
     "OptionDisableUser": "Desativar este usuário",
     "OptionDisableUserHelp": "O servidor não permitirá nenhuma conexão deste usuário. Conexões existentes serão encerradas imediatamente.",
+    "OptionEnableNoPasswordHelp": "Permitirá que este usuário defina sua senha como vazia ao alterar senhas.",
```
viown, line 862: I believe you should only edit en-us.json as the other files are handled by weblate.

PR #6364, opened 2024-12-05
CSS redesign of the skip button

CryLegende:
Improved positioning, hover animation, and enhanced design to match the Jellyfin interface.


thornbill:
I think this is no longer necessary since https://github.com/jellyfin/jellyfin-web/pull/6295 changed the button to use our standard component...

PR #967, opened 2020-03-21
Refactor slideshow.js

MrTimscampi:
**Changes**

Refactors slideshow.js to use standard features of Swiper where applicable.

* Now uses virtual slides, only loading three items at a time (Current, previous and next), and preloads images (Better memory usage and performance for large libraries, as well as nicer experience).
* Handle previous/next keyboard navigation using Swiper itself
* Support play and pause keyboard inputs
* Use the full resolution image
* Disable autoplay by default except on Chromecast
* Use Swiper itself to bind events to the next/previous arrows
* Remove dead code and useless cruft

As a result of using virtual slides, looping had to be disabled.

**Issues**

Fixes #915 
Fixes #918

PR #6370, opened 2024-12-10
Remove classlist shim

3flex:
**Changes**
Remove the `classlist.js` dependency. Supported browsers natively support this now.

**Issues**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28,7 +28,6 @@
         "@types/react-lazy-load-image-component": "1.6.4",
         "abortcontroller-polyfill": "1.7.5",
         "blurhash": "2.0.5",
-        "classlist.js": "https://github.com/eligrey/classList.js/archive/1.2.20180112.tar.gz",
         "classnames": "2.5.1",
         "core-js": "3.38.1",
         "date-fns": "2.30.0",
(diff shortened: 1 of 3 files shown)


3flex:
Oh possibly. I did have a quick look at the Jellyfin docs which only mentioned browser support but not things like Tizen - but I didn't know what kind of support was offered there.

Is it documented anywhere? Or do they use old Chrome versions? Some old versions are listed in packages.json. Those all have support for this natively.

Maybe it's safer to leave it then, I understand embedded web browsers can be quite old, and without documentation it's hard to say if things would break.

3flex:
There's info on Tizen here: anything from Tizen 2 and up (2017 models and newer) will work because it uses Chromium 47+ and classlist support was added in 27: https://developer.samsung.com/smarttv/develop/specifications/web-engine-specifications.html

Again without docs on supported environments and versions it's hard to say definitely.

thornbill:
Supported browsers: https://github.com/jellyfin/jellyfin-web/blob/master/package.json#L139

The polyfill was added due to missing support for `classList.replace`: https://github.com/jellyfin/jellyfin-web/pull/967#discussion_r398918989

3flex:
Thanks, I'd found the browser list already but didn't realise that the specific method used would still need the polyfill. I'm guessing there are plenty of gaps in the test suite so I'll avoid this kind of PR in future.

Closing.

PR #6372, opened 2024-12-10
Update to typescript-eslint v8

3flex:
**Changes**
* Update typescript-eslint to v8
* Disable rule that now requires typed linting - enabling this will impact performance and the checks already take a little while
* Fix violations that are easy to address
* Disable rules on remaining lines that now throw an error when linting

**Issues**


src/components/AppHeader.tsx
```diff
@@ -9,6 +9,7 @@ const AppHeader: FC<AppHeaderParams> = ({
 }) => {
     useEffect(() => {
         // Initialize the UI components after first render
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
         import('../scripts/libraryMenu');
```
thornbill, line 13:
```suggestion
        void import('../scripts/libraryMenu');
```

src/components/Backdrop.tsx
```diff
@@ -13,6 +13,7 @@ const styles = layoutManager.experimental ? {
 const Backdrop = () => {
     useEffect(() => {
         // Initialize the UI components after first render
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
         import('../scripts/autoBackdrops');
```
thornbill, line 17:
```suggestion
        void import('../scripts/autoBackdrops');
```

thornbill (changes requested):

PR #6373, opened 2024-12-11
Prevent opus more than 2 channels being remuxed on Safari

gnattu:
Safari only supports stereo Opus, which requires a similar workaround as we are applying on WebOS for FLAC.

This would currently force downmix 5.1 opus to stereo. After jellyfin/jellyfin#13209 landed in server the server will fallback to transcode 5.1 aac when source is 5.1 opus. 

**Changes**

**Issues**


dmitrylyzo (approved):
LGTM. Just a small remark.

PR #6383, opened 2024-12-14
Add user-level access controls to playlists

### LINKED ISSUE #4698: Implement playlist ACL Interface

viown:
**Changes**
- Adds the ability to share private playlists with other JF users.

**Screenshots**

**Issues**
Closes https://github.com/jellyfin/jellyfin-web/issues/4698

### DIFF
--- a/src/components/playlisteditor/playlisteditor.ts
+++ b/src/components/playlisteditor/playlisteditor.ts
@@ -28,6 +28,9 @@ import 'elements/emby-select/emby-select';
 
 import 'material-design-icons-iconfont';
 import '../formdialog.scss';
+import { getUserApi } from '@jellyfin/sdk/lib/utils/api/user-api';
+import type { UserDto } from '@jellyfin/sdk/lib/generated-client/models/user-dto';
+import type { PlaylistUserPermissions } from '@jellyfin/sdk/lib/generated-client/models/playlist-user-permissions';
 
 interface DialogElement extends HTMLDivElement {
     playlistId?: string
@@ -97,6 +100,7 @@ function createPlaylist(dlg: DialogElement) {
             createPlaylistDto: {
                 Name: name,
                 IsPublic: dlg.querySelector<HTMLInputElement>('#chkPlaylistPublic')?.checked,
+                Users: getUsers(dlg),
                 Ids: itemIds?.split(','),
                 UserId: apiClient.getCurrentUserId()
             }
@@ -127,6 +131,7 @@ function updatePlaylist(dlg: DialogElement) {
             playlistId: dlg.playlistId,
             updatePlaylistDto: {
                 Name: name,
+                Users: getUsers(dlg),
                 IsPublic: dlg.querySelector<HTMLInputElement>('#chkPlaylistPublic')?.checked
             }
         })
@@ -279,6 +284,21 @@ function getEditorHtml(items: string[], options: PlaylistEditorOptions) {
         </div>
     </div>`;
 
+    html += `
+    <div>
+        <div class="sectionTitleContainer flex align-items-center">
+            <h2 className='sectionTitle'>
+                Users
+            </h2>
+            <button id="btnAddUser" is="emby-button" class="fab submit sectionTitleButton">
+                <span class="material-icons add" aria-hidden="true"></span>
+            </button>
+        </div>
+
+        <div class="sharesList paperList"></div>
+    </div>
+    `;
+
     // newPlaylistInfo
     html += '</div>';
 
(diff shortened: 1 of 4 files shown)


larrypozdeev:
This feature will expose the names of all users on the server instance through user select. Would it be better to require manual input of an existing username instead?

I'm new here, so not sure if this is a concern.

viown:
An authenticated user has always had the ability to list other users via the API, this just implements it.

For home users who share their server with friends and family, this isn't much of a concern. I could see it being a problem for larger servers, however that's never been an intended use case for Jellyfin.

larrypozdeev:
Got it, thanks for explaining!

I-G-1-1:
Why still not merged since Dec 14, 2024? 
I'm not a programmer so I cannot help, but I think Jellyfin needs this basic playlist user control feature!
"How can I share a playlist with another user and let him modify the playlist?" is a recurrent question on the forum.
Thanks

viown:
After thinking about it some more, I do agree with @larrypozdeev's comment that some server changes will likely be needed to add a per-user configuration option to allow them to share their playlist with others.

You can currently achieve playlist sharing by modifying the playlist's XML file and adding the user under shares.

I-G-1-1:
Thanks you for the answer.

Can you please share an example of the playlist with manually added shares?
I was using this manual edit on 10.8.x, but stopped working on 10.10.x. 
So if you can give an example of the playlist I can double check what stopped working for me. Thanks

Regarding the "user-level access controls to playlists" feature I think that for now a solution could be that only Admin users can assign playlist to a user and choice who can read or edit a playlist. As Admin they already have access to all username of a Jellyfin istance.

I-G-1-1:
> You can currently achieve playlist sharing by modifying the playlist's XML file and adding the user under shares.

@viown It seems that's not possible at the moment. Here is what I tried (Jellyfin 10.10.7):
```
...
  <OwnerUserId>407b5d79a41e4bcfa40b5a3d079c2081</OwnerUserId>
  <PlaylistItems>
... ... ...
  </PlaylistItems>
  <Shares>
    <User>
      <Name>User01</Name>
      <UserId>407b5d79a41e4bcfa40b5a3d079c2081</UserId>
    </User>
    <User>
      <Name>User02</Name>
      <UserId>d25c960a21f74c45988bd89f08a85bde</UserId>
    </User>
  </Shares>
 <PlaylistMediaType>Audio</PlaylistMediaType>
</Item>
```
I got the \<UserId\> from the jellyfin.db/users. User01 is the owner of the shared playlist (selected to share with all the other users when created)

At this point User02 could see the playlist generated by User01 but Jellyfin completely ignores the \<Shares\> section and the User02 added in the Shares section couldn't edit the playlist. Only User01 that has his UUID in the \<OwnerUserId\> section could edit the playlist.
If User01 add a new song to the playlist, the manually added \<Shares\> section is completely deleted.

So to me seems there is no solution at the moment for sharing a playlist between 2 users with the ability to edit for both.

viown:
I don't think you need to specify the owner of the playlist in the shares.

```xml
<Shares>
  <Share>
    <UserId>00000000-0000-0000-0000-000000000000</UserId>
    <CanEdit>true</CanEdit>
  </Share>
</Shares>
```
Setting the `CanEdit` property to true allows the user to modify the playlist.

If it's still not working out, you can try using the API method directly (https://api.jellyfin.org/#tag/Playlists/operation/UpdatePlaylist) to add a user to a playlist.

I-G-1-1:
> ```
> <Shares>
>   <Share>
>     <UserId>00000000-0000-0000-0000-000000000000</UserId>
>     <CanEdit>true</CanEdit>
>   </Share>
> </Shares>
> ```
this doesn't work in jellyfin 10.10.7.

Lost a lot of time trying to make (with AI help) a python script to grant access to a playlist created by another user using `http://localhost/Playlists/{playlistId}/Users/{userId}` API with payload 
```
{
"CanEdit": true
}
```

The script use an API key but I always get error 400 because the API call contains the userID and I found you cannot make API call impersonating a user while using a token in Jellyfin.

So, as it seems there is no other solution, can you please merge this PR. Thanks

viown:
@I-G-1-1 Open the [Cloudflare deployment for this PR](https://github.com/jellyfin/jellyfin-web/pull/6383#issuecomment-2543342427), click "Change Server" then add your server and login with your user. Then, go to your playlist -> click the 3 dots -> "Edit" and then you'll be able to use the features of this PR to add your user. Alternatively, pull this branch and run it locally.

I hope that helps. I'll be marking our comments as off-topic since it's unrelated to the PR.

JoelAzaria:
> @I-G-1-1 Open the [Cloudflare deployment for this PR](https://github.com/jellyfin/jellyfin-web/pull/6383#issuecomment-2543342427), click "Change Server" then add your server and login with your user. Then, go to your playlist -> click the 3 dots -> "Edit" and then you'll be able to use the features of this PR to add your user. Alternatively, pull this branch and run it locally.
> 
> I hope that helps. I'll be marking our comments as off-topic since it's unrelated to the PR.

As one who depends on precompiled docker containers for Jellyfin I cannot implement this but wanted to add one more vote for "please accept this PR".

Till then I'l be waiting patiently...

I-G-1-1:
I don't know if it's off-topic.. however to mitigate this issue for now I made a python script to grant/revoke write permission to users different from the owner of the playlist. You can find it [here](https://forum.jellyfin.org/t-playlist-sharing-with-write-permission) in the Jellyfin forum.

PR #6385, opened 2024-12-15
Fix lyrics styling

lxqueen:
The styling for the lyrics page needed a tiny bit of TLC for a nicer cross-device experience.

### Changes
- Lyrics container now has extra top padding and max-width
- Lyrics font size and bounding box now scales better on mobile & TV
  - Switched to `rem`
  - Shrunk slightly on non-large-desktop screens
  - Added horizontal padding for mobile
- Added slight glow on dynamic lyric hover, with fast transition
- Increased future lyric opacity slightly
- Empty lyric lines (used as gap between sections in some LRC files) now have a small forced height to separate said sections

### Comparison
|     | Before | After |
|:---:|:------:|:-----:|
| Desktop | desktop-before | desktop-after-hovered |
| Mobile | iphonese-before | iphonese-after |
| TV\* | tv-before | tv-after |

*\* Don't worry about the serif font on TV, I was alternating by user agent rather than going into settings every time.*

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -93,6 +93,7 @@
 - [Connor Smith](https://github.com/ConnorS1110)
 - [iFraan](https://github.com/iFraan)
 - [Ali](https://github.com/bu3alwa)
+- [Alexis Queen](https://github.com/lxqueen)
 
 ## Emby Contributors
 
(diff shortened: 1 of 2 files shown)


lxqueen:
_(Force pushed to fix linting errors.)_

PR #6393, opened 2024-12-17
Migrate logs to React

viown:
**Changes**
Migrates the dashboard logs page to React + TS SDK

**Issues**


src/apps/dashboard/routes/logs.tsx
```diff
@@ -0,0 +1,150 @@
...
+        if (!api) return;
+
+        return getSystemApi(api)
+            .getServerLogs()
+            .then(({ data }) => {
+                setLogs(data);
+            });
+    }, [api]);
+
+    const onSubmit = useCallback((e: FormEvent<HTMLFormElement>) => {
```
thornbill, line 40: We should use react router's actions and loaders to initialize the page data and handle submissions. You can see an example of this in https://github.com/jellyfin/jellyfin-web/pull/6334.

src/apps/dashboard/routes/logs.tsx
thornbill, line 1: I would prefer using `logs/index.tsx`.

src/apps/dashboard/routes/logs.tsx
viown: I've noticed that inside `loader`, the access token for `ServerConnections.getCurrentApi()` will be null when refreshing the page (going from another page to it works fine). This wasn't a problem for the branding page because the configuration request does not need an access token.

Also usage of `loader` will not show a loading indicator.

Have not looked into why that is, but came up with a workaround for now.

src/apps/dashboard/routes/logs.tsx
thornbill: Sure that has been on my list 😅

src/apps/dashboard/routes/logs/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+                        <Button
+                            type='submit'
+                            size='large'
+                        >
+                            {globalize.translate('Save')}
+                        </Button>
+                    </Stack>
+                </Form>
+                <div className='serverLogs readOnlyContent'>
+                    <div className='paperList'>
```
thornbill, line 126: Could we move this to a mui <List> also?

src/apps/dashboard/routes/logs/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+                                <Switch
+                                    checked={logOptions?.EnableSlowResponseWarning}
+                                    onChange={setLogWarningMessage}
+                                    name={'EnableWarningMessage'}
+                                />
+                            }
+                            label={globalize.translate('LabelSlowResponseEnabled')}
+                        />
+
+                        <TextField
```
thornbill, line 108: We should disable this if the option is disabled.

src/apps/dashboard/features/logs/api/useLogOptions.ts
```diff
@@ -0,0 +1,26 @@
...
+import { useApi } from 'hooks/useApi';
+import type { AxiosRequestConfig } from 'axios';
+
+export const fetchLogOptions = async (api?: Api, options?: AxiosRequestConfig) => {
+    if (!api) {
+        console.error('[useLogOptions] No API instance available');
+        return;
+    }
+
+    const response = await getConfigurationApi(api).getConfiguration(options);
```
thornbill, line 13: This seems pretty generic... maybe it should just be `useConfiguration`?

src/apps/dashboard/features/logs/api/useLogEntries.ts
```diff
@@ -0,0 +1,26 @@
...
+        console.error('[useLogEntries] No API instance available');
+        return;
+    }
+
+    const response = await getSystemApi(api).getServerLogs(options);
+
+    return response.data;
+};
+
+export const useLogEntries = () => {
```
thornbill, line 18: Maybe `useServerLogs` to match the api call?

src/apps/dashboard/features/logs/api/useLogOptions.ts
viown: Moved to **src/hooks** also.

src/apps/dashboard/routes/logs/index.tsx
viown: Updated look:
image
Let me know if I should change anything.

thornbill: Maybe just use a `<List>` and `<ListItem>` instead of cards?

src/apps/dashboard/features/logs/components/LogItemList.tsx
```diff
@@ -0,0 +1,50 @@
+import React, { FunctionComponent } from 'react';
+import type { LogFile } from '@jellyfin/sdk/lib/generated-client/models/log-file';
+import { List, ListItem, ListItemButton, ListItemText } from '@mui/material';
```
thornbill, line 3: We should use the direct imports for each of these to help with tree-shaking.

src/apps/dashboard/features/logs/components/LogItemList.tsx
```diff
@@ -0,0 +1,50 @@
...
+import datetime from 'scripts/datetime';
+
+type LogItemProps = {
+    logs: LogFile[];
+};
+
+const LogItemList: FunctionComponent<LogItemProps> = ({ logs }: LogItemProps) => {
+    const { api } = useApi();
+
+    const getLogFileUrl = (logFile: LogFile) => {
```
thornbill, line 14: Let's add a `TODO` comment to replace this with `getUri` in the SDK once it is available.

src/apps/dashboard/features/logs/components/LogItemList.tsx
```diff
@@ -0,0 +1,50 @@
...
+            {logs.map(log => {
+                return (
+                    <ListItem key={log.Name} disablePadding>
+                        <ListItemButton href={getLogFileUrl(log)} target='_blank'>
+                            <ListItemText
+                                primary={log.Name}
+                                primaryTypographyProps={{ variant: 'h3' }}
+                                secondary={getDate(log)}
+                                secondaryTypographyProps={{ variant: 'body1' }}
+                            />
```
thornbill, line 41: Maybe we should add the `OpenInNew` icon after the text :man_shrugging:

src/apps/dashboard/routes/logs/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+                        <Button
+                            type='submit'
+                            size='large'
+                        >
+                            {globalize.translate('Save')}
+                        </Button>
+                    </Stack>
+                </Form>
+                <Box className='serverLogs readOnlyContent' sx={{ mt: 1 }}>
```
thornbill, line 124:
```suggestion
                <Box className='serverLogs readOnlyContent' sx={{ mt: 3 }}>
```

We use 3 for the form spacing, so it probably makes sense to use the same here.

thornbill (changes requested):

src/apps/dashboard/features/logs/components/LogItemList.tsx
viown: Is this what you meant?
image

thornbill: Yeah... imo it gives a better indication of what clicking the log is going to do... but I'm not a ui/ux person :sweat_smile:

viown: I'm not either, but I agree that I think this is better.

thornbill (approved):

Hello-World-Traveler:
Could a button be added to the log page to bring you all the way to the bottom?

```
import React from 'react';

const ScrollToBottomButton = () => {
  const scrollToBottom = () => {
    window.scrollTo({
      top: document.body.scrollHeight,
      behavior: 'smooth' // This makes the scroll smooth
    });
  };

  return (
    <button onClick={scrollToBottom} style={buttonStyle}>
      Scroll to Bottom
    </button>
  );
};

// Optional: Add some basic styles for the button
const buttonStyle = {
  position: 'fixed',
  bottom: '20px',
  right: '20px',
  padding: '10px 20px',
  backgroundColor: '#007BFF',
  color: 'white',
  border: 'none',
  borderRadius: '5px',
  cursor: 'pointer',
};

const App = () => {
  return (
    <div>
      <h1>Scroll to Bottom Example</h1>
      <div style={{ height: '1500px', padding: '20px' }}>
        {/* Simulating a long page */}
        <p>Scroll down to see the button.</p>
      </div>
      <ScrollToBottomButton />
    </div>
  );
};

export default App;
```
```
import React from 'react';
import { createRoot } from 'react-dom/client'; // Import createRoot
import App from './App';

const container = document.getElementById('root');
const root = createRoot(container); // Create a root
root.render(<App />); // Render the App component
```
Button works but not styled to jellyfin.

viown:
@Hello-World-Traveler I added a watch feature to the log viewer here: https://github.com/jellyfin/jellyfin-web/pull/7268

It will scroll you to the bottom of the page and keeps the logs up to date.

PR #6396, opened 2024-12-17
Remove invalid item id usage as media source id

### LINKED ISSUE #6395: Item ID is used as MediaSourceId
**Describe The Bug**
I am debugging playback issues of my plugin implementing the `IChannel` interface and I noticed that the call sent to the `/Items/<uid>/PlaybackInfo` now contains a key `MediaSourceId` in the JSON, which equals the `uid` of the URL. With this key set, the JF server returns no compatible streams. When fetching from the 10.9 webclient (with a JF 10.10 server), this key is missing and the server does report streams and playback starts.

Looking at the JF source, there is some filtering if this key is set. It appears that this key should not be set to the `item.Id` value.

Looking at the change history, it looks like an accidental [` || item.Id`](https://github.com/viown/jellyfin-web/blob/4c31742cc50c19702f2c1610c2bcfc0d6c4d210b/src/components/playback/playbackmanager.js#L2576) was introduced in #6112 to query for the item, perhaps not realising it is later used in the query later on for server-side filtering. This was partially fixed by #6166 for Live TV by not setting this variable in that case.

**Media Information**
N/A

**Screenshots**
N/A

 - Platform: Linux
 - Browser: Firefox
 - Jellyfin Version: 10.10.3

**Additional Context**
N/A

Kevinjil:
Looking at the change history, it looks like an accidental ` || item.Id` was introduced in 4c31742cc50c19702f2c1610c2bcfc0d6c4d210b to query for the item, perhaps not realising it is later used in the query later on for server-side filtering. This was partially fixed by c54db604d9426f3b3d46222c9161b88aa210d4da and
ca4763512f4a4e2a73e37600a5dc520f917c2274 for Live TV by not setting this variable in that case.

**Changes**
Only use the fallback item id in the API call

**Issues**
Fixes #6395

### DIFF
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -2602,16 +2602,11 @@ export class PlaybackManager {
                 });
             }
 
+            let mediaSourceId = playOptions.mediaSourceId;
+
             const apiClient = ServerConnections.getApiClient(item.ServerId);
-            let mediaSourceId;
-
             const isLiveTv = [BaseItemKind.TvChannel, BaseItemKind.LiveTvChannel].includes(item.Type);
-
-            if (!isLiveTv) {
-                mediaSourceId = playOptions.mediaSourceId || item.Id;
-            }
-
-            const getMediaStreams = isLiveTv ? Promise.resolve([]) : apiClient.getItem(apiClient.getCurrentUserId(), mediaSourceId)
+            const getMediaStreams = isLiveTv ? Promise.resolve([]) : apiClient.getItem(apiClient.getCurrentUserId(), mediaSourceId || item.Id)
                 .then(fullItem => {
                     return fullItem.MediaStreams;
                 });
@@ -2644,13 +2639,20 @@ export class PlaybackManager {
                 playOptions.items = null;
 
                 const trackOptions = {};
+                let isIdFallbackNeeded = false;
 
                 autoSetNextTracks(prevSource, mediaStreams, trackOptions, user.Configuration.RememberAudioSelections, user.Configuration.RememberSubtitleSelections);
                 if (trackOptions.DefaultAudioStreamIndex != null) {
                     options.audioStreamIndex = trackOptions.DefaultAudioStreamIndex;
+                    isIdFallbackNeeded = true;
                 }
                 if (trackOptions.DefaultSubtitleStreamIndex != null) {
                     options.subtitleStreamIndex = trackOptions.DefaultSubtitleStreamIndex;
+                    isIdFallbackNeeded = true;
+                }
+
+                if (isIdFallbackNeeded) {
+                    mediaSourceId ||= item.Id;
                 }
 
                 return getPlaybackMediaSource(player, apiClient, deviceProfile, item, mediaSourceId, options).then(async (mediaSource) => {


Kevinjil:
Marked as draft as I want to test some more scenarios, and perhaps get some insight from the original commit authors as to why it was added.

viown:
https://github.com/jellyfin/jellyfin-web/pull/6166#discussion_r1788537914

Kevinjil:
> [#6166 (comment)](https://github.com/jellyfin/jellyfin-web/pull/6166#discussion_r1788537914)
> 
> The server returning no streams when mediaSourceId is defined is believed to be a server issue. See [#6166 (comment)](https://github.com/jellyfin/jellyfin-web/pull/6166#issuecomment-2394996916)

I believe this is not a server issue: we are querying for streams that match the given media source id, where the id we give is not a media source. An empty result is expected from an API point of view.

viown:
This should be a case specific to Live TV (which should have been fixed). Omitting item id might cause a regression of the original issue (https://github.com/jellyfin/jellyfin-web/issues/3994) due to PlaybackInfo not respecting the provided indices.

We might instead need to check for your scenario [here](https://github.com/jellyfin/jellyfin-web/blob/052eb6d5ee729d105f3867e5979a67f40432f071/src/components/playback/playbackmanager.js#L2586)

Kevinjil:
> This should be a case specific to Live TV (which should have been fixed). Omitting item id might cause a regression of the original issue (#3994) due to PlaybackInfo not respecting the provided indices.

I think that selectively applying the fix might be a better idea than trying to figure out all scenarios where is does not apply. I do not have the data to test the intended scenario but this seems like the solution.

> We might instead need to check for your scenario [here](https://github.com/jellyfin/jellyfin-web/blob/052eb6d5ee729d105f3867e5979a67f40432f071/src/components/playback/playbackmanager.js#L2586)

Unfortunately, the base kind in this case is `Movie`, for which the fix was intended if I understood the issue correctly. Hence I propose to only apply the fix if the tracks are selected.

src/components/playback/playbackmanager.js
```diff
@@ -2580,16 +2580,15 @@ export class PlaybackManager {
...
 
-            const apiClient = ServerConnections.getApiClient(item.ServerId);
-            let mediaSourceId;
-
-            const isLiveTv = [BaseItemKind.TvChannel, BaseItemKind.LiveTvChannel].includes(item.Type);
-
-            if (!isLiveTv) {
-                mediaSourceId = playOptions.mediaSourceId || item.Id;
+            let mediaSourceId = playOptions.mediaSourceId;
+            const needsTrackWorkaround = (playOptions.audioStreamIndex || playOptions.subtitleStreamIndex);
```
viown, line 2584: We should do it after `autoSetNextTracks` ([here](https://github.com/jellyfin/jellyfin-web/blob/6fd90b0c368bc8dc460fa7b404d47d9928b9be12/src/components/playback/playbackmanager.js#L2626)), like how it was done [before](https://github.com/viown/jellyfin-web/blob/95e23875f4e2761a8ed65d73dd5b94977e0d0d79/src/components/playback/playbackmanager.js#L2614).

Kevinjil:
@viown As I agree with https://github.com/jellyfin/jellyfin-web/pull/6166#discussion_r1788534366 that is might be a weird place to do this logic, I kept the `needsTrackWorkaround` boolean to indicate why we are doing this here.

viown:
@Kevinjil Could you re-target your branch to `release-10-10.z`?

Kevinjil:
> @Kevinjil Could you re-target your branch to `release-10-10.z`?

Done

dmitrylyzo:
> We might instead need to check for your scenario [here](https://github.com/jellyfin/jellyfin-web/blob/052eb6d5ee729d105f3867e5979a67f40432f071/src/components/playback/playbackmanager.js#L2586)

[There](https://github.com/jellyfin/jellyfin-web/blame/c2c5228c06a55a8b946de5741d4f9e1c30f1c837/src/components/playback/playbackmanager.js#L2579) was `BaseItemKind.Channel`, but Bill [suggested](https://github.com/jellyfin/jellyfin-web/pull/6166#discussion_r1792166790) removing it.

If `Channel` is supposed to work as a LiveTV channel, we should just add it to the list. I am not familiar with Channel plugins, so I could be wrong.

Kevinjil:
> > We might instead need to check for your scenario [here](https://github.com/jellyfin/jellyfin-web/blob/052eb6d5ee729d105f3867e5979a67f40432f071/src/components/playback/playbackmanager.js#L2586)
> 
> [There](https://github.com/jellyfin/jellyfin-web/blame/c2c5228c06a55a8b946de5741d4f9e1c30f1c837/src/components/playback/playbackmanager.js#L2579) was `BaseItemKind.Channel`, but Bill [suggested](https://github.com/jellyfin/jellyfin-web/pull/6166#discussion_r1792166790) removing it.
> 
> If `Channel` is supposed to work as a LiveTV channel, we should just add it to the list. I am not familiar with Channel plugins, so I could be wrong.

Adding `Channel` will not solve the problem, the items provided by the channel themselves are `Movie` items.

src/components/playback/playbackmanager.js
```diff
@@ -2645,13 +2640,20 @@ export class PlaybackManager {
                 playOptions.items = null;
 
                 const trackOptions = {};
+                let needsTrackWorkaround = false;
```
thornbill, line 2643: Just a minor note, I normally try to prefix boolean variables with `is` or `has`... maybe something like `isIdFallbackNeeded` could work here

viown: It could possibly make more logical sense to make setting `mediaSourceId` to be part of `autoSetNextTracks` (as that's what this is specific to). I'm not sure if I would consider it a workaround anyway since `playOptions.mediaSourceId` would be filled in when the user manually selects a subtitle or audio track.

Kevinjil: @viown How would you propose to set the `mediaSourceId` in `autoSetNextTracks`? Do you mean passing along the item id as parameter, and filling it in the `trackOptions` when needed?

viown: We could update `autoSetNextTracks` to make it mutate `playOptions` or `options` directly and removing `trackOptions` as it might be redundant in this case.

Kevinjil: > We could update `autoSetNextTracks` to make it mutate `playOptions` or `options` directly and removing `trackOptions` as it might be redundant in this case.

I've puzzled a bit, but I don't think that removing the `trackOptions` will result in cleaner code, especially since the track options are also used after the next asynchronous call to set the secondary subtitle index on the retrieved media source.

If we don't want to call it a workaround, [setting it in the 'weird' spot](https://github.com/jellyfin/jellyfin-web/pull/6166/files/95e23875f4e2761a8ed65d73dd5b94977e0d0d79#diff-b338dd22dfaa8db18edede53fe56aabb5d911c5e611cc33ad3c914410d021cbfR2614) might still be the best solution. If I'm missing something and you have a better suggestion, please enlighten me @viown ;).

thornbill (approved):

PR #6408, opened 2024-12-24
Fix focus loss on popups in Samsung Tizen 8 Smart TVs

dantesbr:
The web client loses focus when a popup is displayed on Samsung Smart TVs with Tizen 8 (OneUI/Tizen 8). This issue only occurs on the TV itself and not on the Tizen emulator.
Fixes the issue https://github.com/jellyfin/jellyfin-tizen/issues/311

**Changes**  
This PR updates `confirm.js` to ensure that the native `window.confirm` is not used on Samsung Smart TVs with Tizen 8 (OneUI). Instead, it uses a custom confirmation dialog to prevent focus-related issues specific to this platform.  

**Issues**  
Fixes https://github.com/jellyfin/jellyfin-tizen/issues/311

### DIFF
--- a/src/components/alert.js
+++ b/src/components/alert.js
@@ -6,7 +6,8 @@ import globalize from '../lib/globalize';
 export default async function (text, title) {
     // Modals seem to be blocked on Web OS and Tizen 2.x
     const canUseNativeAlert = !!(
-        !browser.web0s
+        !(browser.tizenVersion && browser.tizenVersion >= 8)
+        && !browser.web0s
         && !(browser.tizenVersion && browser.tizenVersion < 3)
         && browser.tv
         && window.alert
--- a/src/components/alert.js
+++ b/src/components/alert.js
@@ -6,9 +6,8 @@ import globalize from '../lib/globalize';
 export default async function (text, title) {
     // Modals seem to be blocked on Web OS and Tizen 2.x
     const canUseNativeAlert = !!(
-        !(browser.tizenVersion && browser.tizenVersion >= 8)
-        && !browser.web0s
-        && !(browser.tizenVersion && browser.tizenVersion < 3)
+        !browser.web0s
+        && !(browser.tizenVersion && (browser.tizenVersion < 3 || browser.tizenVersion >= 8))
         && browser.tv
         && window.alert
     );
--- a/src/components/confirm/confirm.js
+++ b/src/components/confirm/confirm.js
@@ -7,6 +7,7 @@ function useNativeConfirm() {
     // webOS seems to block modals
     // Tizen 2.x seems to block modals
     return !browser.web0s
+        && !(browser.tizenVersion && browser.tizenVersion >= 8)
         && !(browser.tizenVersion && browser.tizenVersion < 3)
         && browser.tv
         && window.confirm;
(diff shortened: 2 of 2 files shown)


src/components/confirm/confirm.js
```diff
@@ -4,6 +4,9 @@ import dialog from '../dialog/dialog';
 import globalize from '../../lib/globalize';
 
 function useNativeConfirm() {
+    if (browser.tizenVersion && browser.tizenVersion >= 8) {
```
thornbill, line 7: The should be integrated with the Tizen check below

thornbill (changes requested):

morpheus133:
I think same modification is needed in also here:
https://github.com/jellyfin/jellyfin-web/blob/5b4cfbf4103c939b59c316a262a7a60fc8d7759b/src/components/alert.js#L6-L14

src/components/alert.js
```diff
@@ -6,7 +6,8 @@ import globalize from '../lib/globalize';
 export default async function (text, title) {
     // Modals seem to be blocked on Web OS and Tizen 2.x
     const canUseNativeAlert = !!(
-        !browser.web0s
+        !(browser.tizenVersion && browser.tizenVersion >= 8)
+        && !browser.web0s
         && !(browser.tizenVersion && browser.tizenVersion < 3)
```
dmitrylyzo, line 11:
```suggestion
        !browser.web0s
        && !(browser.tizenVersion && (browser.tizenVersion < 3 || browser.tizenVersion >= 8))
```
Technically, we can omit `browser.tizenVersion` and just do `!(browser.tizenVersion < 3 || browser.tizenVersion >= 8)`.

src/components/confirm/confirm.js
```diff
@@ -7,6 +7,7 @@ function useNativeConfirm() {
     // webOS seems to block modals
     // Tizen 2.x seems to block modals
     return !browser.web0s
+        && !(browser.tizenVersion && browser.tizenVersion >= 8)
         && !(browser.tizenVersion && browser.tizenVersion < 3)
```
dmitrylyzo, line 11:
```suggestion
        && !(browser.tizenVersion && (browser.tizenVersion < 3 || browser.tizenVersion >= 8))
```

morpheus133: Can I have a question:
Why do we use NativeAlert at all?

With this modification we wont use it before 4 and after 8. 
And most probably nobody will check it  after 8.X that the function is corrected by Samsung so it will remain there forever.
Simple we can use this: `&& !(browser.tizenVersion)` and all version will be the same.

dmitrylyzo: > Why do we use NativeAlert at all?

Tbh, I ask the same question myself. 😅
It is inherited from Emby. 🤷‍♂️

> Simple we can use this: `&& !(browser.tizenVersion)` and all version will be the same.

`!browser.tizen` is shorter.

dmitrylyzo (approved):
LGTM.
But really, do we need native modals (https://github.com/jellyfin/jellyfin-web/pull/6408#discussion_r1904679775)?

Bill, don't forget to squash commits :sweat_smile:

thornbill:
I have no clue if they are needed... I wouldn't remove them in a backport but seems like a good thing to add to the queue for 10.11. 🙂

thornbill (approved):

PR #6410, opened 2024-12-27
Simple attempt at adding song listing to artist page

theevilapplepie:
This adds a song list within the artist detail.

It solves the probem where you can't remember a song name, so you have to jump album to album looking for it and if it's not part of a known album it's impossible to find without manally scanning the songs list ( impossible for a large library ) or being able to guess at the song name well enough in search.

I understand this may not be ideal for the following reasons:
* There is no pagination
* I'm not confident the "Special Feaure" isn't used somehow by MusicArtist but I couldn't find anywhere it is
* The presentation does not match the new Vue frontend

I welcome your feedback/commentary regardless of outcome.
This was a few hours getting setup/familiar with the codebase and was done as simply as reasonable to accomplish the goal.

**Changes**
* Updated listview to support adding album for audio types
* Updated itemDetails/index.js to reuse "Special Feature" for the "MusicArtist" item type for presenting songs by the artist

**Issues**
No issues resolved, feature


src/controllers/itemDetails/index.js
```diff
@@ -1778,6 +1779,7 @@ function renderMusicVideos(page, item, user) {
             page.querySelector('#musicVideosCollapsible').classList.add('hide');
         }
     });
+
```
theevilapplepie, line 1782:
```suggestion
```

src/controllers/itemDetails/index.js
```diff
@@ -1754,6 +1754,7 @@ function renderCollectionItemType(page, parentItem, type, items) {
 }
 
 function renderMusicVideos(page, item, user) {
+
```
theevilapplepie, line 1757:
```suggestion
```

theevilapplepie (commented):
Closing review

src/controllers/itemDetails/index.js
```diff
@@ -826,7 +826,7 @@ function setInitialCollapsibleState(page, item, apiClient, context, user) {
 
     renderScenes(page, item);
 
-    if (item.SpecialFeatureCount > 0) {
+    if (item.SpecialFeatureCount > 0 || item.Type == 'MusicArtist') {
```
theevilapplepie, line 829: This should probably be ===
```suggestion
    if (item.SpecialFeatureCount > 0 || item.Type === 'MusicArtist') {
```

src/controllers/itemDetails/index.js
```diff
@@ -1829,14 +1829,49 @@ function getVideosHtml(items) {
...
+            });
+            imageLoader.lazyChildren(specialsContent);
+        } else {
+            page.querySelector('#specialsCollapsible').classList.add('hide');
+        }
     });
 }
 
+function renderSpecials(page, item, user) {
+    if ( item.Type == 'MusicArtist' ) {
```
theevilapplepie, line 1864: Should probably be ===
```suggestion
    if ( item.Type === 'MusicArtist' ) {
```

thornbill:
I don't see the advantage in reusing the special features container for this? Why not add a new container element to the template or reuse the song listing container for albums?

theevilapplepie:
> I don't see the advantage in reusing the special features container for this?
I reused the specials container since it wasn't in use to avoid adding new elements as the template is already huge.
I can just add one though, no problem.

> Why not .. reuse the song listing container for albums?
I believe it was due to placement, but I'll get a better answer for when I return. :)

I'm going to close this out to review and come back with an updated version.

PR #6413, opened 2024-12-29
Prevent Focus Loss When Skip Button is Pressed

rlauuzo:
Addresses the focus-related issue when interacting with the skip button during playback.

1. Ensures that after the skip button is pressed and the OSD is visible, the focus automatically moves to the pause button.
2. Updates the logic to retain focus on the skip button when the OSD hides.

### DIFF
--- a/src/controllers/playback/video/index.js
+++ b/src/controllers/playback/video/index.js
@@ -354,7 +354,7 @@ export default function (view) {
             toggleSubtitleSync('hide');
 
             // Firefox does not blur by itself
-            if (document.activeElement) {
+            if (osdBottomElement.contains(document.activeElement)) {
                 document.activeElement.blur();
             }
         }
@@ -1252,7 +1252,7 @@ export default function (view) {
 
         switch (key) {
             case 'Enter':
-                showOsd();
+                showOsd(e.target.classList.contains('skip-button') ? btnPlayPause : undefined);
                 break;
             case 'Escape':
             case 'Back':


src/controllers/playback/video/index.js
```diff
@@ -1252,7 +1252,7 @@ export default function (view) {
 
         switch (key) {
             case 'Enter':
-                showOsd();
+                showOsd(e.target.classList.contains('skip-button') ? btnPlayPause : undefined);
```
dmitrylyzo, line 1255: I prefer this to be fixed for the general case (for example, if we add another button). Did you try what I [posted](https://github.com/jellyfin/jellyfin-web/pull/6295#issuecomment-2564754517)?

rlauuzo: Yes, I tried your suggestion, but it essentially prevented the skip button from gaining focus when the OSD was visible.

rlauuzo: Would it be better to change this to:

```javascript
showOsd(osdBottomElement.contains(e.target) ? undefined : btnPlayPause);
```

dmitrylyzo: > Yes, I tried your suggestion, but it essentially prevented the skip button from gaining focus when the OSD was visible.

It shouldn't gain focus when the OSD is visible. More precisely, it shouldn't grab focus from a valid element that the user interacts with.

We are trying to fix the loss of focus after clicking the skip button, no?
There may be a problem with my suggestion if `showOsd` is called before the focus is blurred.


> Would it be better to change this to:
> 
> ```js
> showOsd(osdBottomElement.contains(e.target) ? undefined : btnPlayPause);
> ```

I guess it will interfere with the subtitle offset dialog - it doesn't belong to `osdBottomElement`.

rlauuzo: > It shouldn't gain focus when the OSD is visible. More precisely, it shouldn't grab focus from a valid element that the user interacts with.

I mean it's impossible to use the navigation keys to reach the skip button. The focus keeps jumping back to the pause button every time.

src/controllers/playback/video/index.js
```diff
@@ -354,7 +365,7 @@ export default function (view) {
             toggleSubtitleSync('hide');
 
             // Firefox does not blur by itself
-            if (document.activeElement) {
+            if (document.activeElement && !focusManager.isCurrentlyFocusable(document.activeElement)) {
```
dmitrylyzo, line 368: This (blur) is probably no longer necessary in modern Firefox.

We cannot use `isCurrentlyFocusable` until the OSD is completely hidden. As an option (not to break something):
```suggestion
            if (osdBottomElement.contains(document.activeElement) || headerElement.contains(document.activeElement)) {
```

rlauuzo:
Hey, one thing I noticed, unrelated to the PR, is that on my TV (WebOS 6; this doesn't happen on PC browsers), when the button is pressed, it briefly reappears after the seek and then hides again.

src/controllers/playback/video/index.js
```diff
@@ -354,7 +367,8 @@ export default function (view) {
             toggleSubtitleSync('hide');
 
             // Firefox does not blur by itself
-            if (document.activeElement) {
+            if (osdBottomElement.contains(document.activeElement)
+                || headerElement.contains(document.activeElement)) {
```
dmitrylyzo, line 371: Just remembered about `SubtitleSync` - it was blurred by the OSD. Currently we only blur the elements that belong to the OSD. So we should make `SubtitleSync` to blur itself near this:
https://github.com/jellyfin/jellyfin-web/blob/89867d21dae1d2426e0fafe8610f0531d5b70b6f/src/components/subtitlesync/subtitlesync.js#L151

We could also just revert the blurring to `if (document.activeElement)` since we are hiding everything anyway. Then we probably need to fix the refocusing on the pause button when hiding the skip button.

But since, as I said, it (blur) is probably unnecessary now, we can leave it as is (for now) and remove it completely later.

For the record, this is reproducible in Firefox 72. We don't have to support such an old Firefox because it can/should be updated unlike the web engine in TV.

dmitrylyzo:
> Hey, one thing I noticed, unrelated to the PR, is that on my TV (WebOS 6; this doesn't happen on PC browsers), when the button is pressed, it briefly reappears after the seek and then hides again.

Maybe it seeks on the keyframe that is before the end of the segment?
Large intervals between keyframes make skipping unreliable.

dmitrylyzo (approved):

src/controllers/playback/video/index.js
rlauuzo: Another thing: `SubtitleSync` can't be accessed using navigation keys if the skip button is visible.

dmitrylyzo: > Another thing: `SubtitleSync` can't be accessed using navigation keys if the skip button is visible.

Yes, this is a problem with current navigation in general: we don't have a "navmap", the focus manager moves along horizontal and vertical "lanes".

We could probably search for the nearest element within the container crossed by the lane. Obviously not for this PR.

It would also be better to take into account how the navigation will be implemented in React.

PR #6416, opened 2024-12-30
Fix missing space in pt-BR translation

andersonmendess:
Add a missing space between words in the HeaderNextItemPlayingInValue string in the Portuguese (pt-BR) translation.

**Changes**
Added a missing space in the HeaderNextItemPlayingInValue string in the Portuguese (pt-BR) translation file. This ensures proper spacing between words, improving the readability of the message displayed to users.

**Issues**
N/A (This is a minor fix without an associated issue.)


thornbill:
Translation changes should be made on translate.jellyfin.org excluding updates to the source language (en-us).

PR #6421, opened 2025-01-02
Fix metadata dialog backdrop click to exit

viown:
**Changes**
- Fixes not being able to click the backdrop of the metadata dialog to exit
- Fixes dragging mouse from inner dialog to outside causing it to close (Firefox)

**Issues**
Fixes https://github.com/jellyfin/jellyfin-web/issues/5723

### DIFF
--- a/src/components/dialogHelper/dialogHelper.js
+++ b/src/components/dialogHelper/dialogHelper.js
@@ -43,7 +43,7 @@ function tryRemoveElement(elem) {
     }
 }
 
-function DialogHashHandler(dlg, hash, resolve, dlgOptions) {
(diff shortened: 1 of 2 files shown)


nielsvanvelzen:
When a dialog contains forms it should require an explicit close button to be pressed or an additional dialog that shows to confirm with the user that any changes made to the form will not be saved. It is extremely easy to close a dialog using a click on the scrim by accident.

thornbill:
I think we massively overuse modals, and I hate it, but this makes the modal function the same as all the others while fixing the original bug.

thornbill (approved):

PR #6427, opened 2025-01-06
Fix styling for wide device icons

thornbill:
**Changes**
Fixes wide device icons being cutoff on the sides. This affected the webOS icon added in https://github.com/jellyfin/jellyfin-web/pull/6281 in particular.

**Issues**
N/A


viown:
Noticed the Kodi logo can be tiny because of this:

image

viown (approved):
I think it overall looks better than before

PR #6439, opened 2025-01-11
Add CSS variable to override font for easy theming

JohnyTheCarrot:
**Changes**
This PR adds a couple CSS variables for the user to override in order to add a custom font.
Why not just overwrite it using a custom `html` rule with a `font-family` property? There are "typography" elements that manually re-specify the font family, and given that those class names seem to be hash based, they would be subject to change as the CSS would change, assuming I'm right.

Added vars to override:
`--jf-font-override`
`--jf-font-override--ja`
`--jf-font-override--ko`
`--jf-font-override--zh-CN`
`--jf-font-override--zh-TW`
`--jf-font-override--zh-HK`
Each of the per-lang overwrites will fall back to `--jf-font-override` if not overridden, and `--jf-font-override` will fall back to Noto Sans if not overriden.

When wanting to use `font-family` anywhere in the code base, one would reference the `--jf-current-font` variable to retrieve the current effective font, overridden or not.
A `jf` prefix has been added to the vars to avoid clashes with user themes.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -94,6 +94,7 @@
 - [iFraan](https://github.com/iFraan)
 - [Ali](https://github.com/bu3alwa)
 - [K. Kyle Puchkov](https://github.com/kepper104)
+- [JohnyTheCarrot](https://github.com/johnythecarrot)
 
 ## Emby Contributors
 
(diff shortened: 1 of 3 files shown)


JohnyTheCarrot:
Maybe the file wouldn't need to be called `fonts.noto.scss` anymore? Lmk

thornbill:
Support for CSS variables for theming is planned (it's been on the roadmap for a couple years), but there is some preliminary work that needs to be done in the themes and we need to maintain fallbacks since we support browsers (TVs) that do not support CSS variables.

PR #6441, opened 2025-01-12
Enable no unchecked index access rule

thornbill:
**Changes**
Enables the no unchecked index access rule for typescript as recommended [here](https://github.com/jellyfin/jellyfin-web/pull/6271#discussion_r1902931000) and fixes the related errors

**Issues**
N/A

### DIFF
--- a/src/components/actionSheet/actionSheet.ts
+++ b/src/components/actionSheet/actionSheet.ts
@@ -93,6 +93,8 @@ function getPosition(positionTo: Element, options: Options, dlg: HTMLElement) {
 
     const pos = getOffsets([positionTo])[0];
 
+    if (!pos) return;
+
     if (options.positionY !== 'top') {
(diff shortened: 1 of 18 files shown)


src/components/actionSheet/actionSheet.ts
```diff
@@ -250,6 +252,8 @@ export function show(options: Options) {
     for (let i = 0; i < options.items.length; i++) {
         const item = options.items[i];
 
+        if (!item) continue;
```
dmitrylyzo, line 255: Can happen?
If can - the error is shadowed.

thornbill:
I've moved this to a draft... I think this setting might have value, but need to reconsider how to apply it in practice based on the feedback above. Maybe using non-null assertions would be a better practice for the cases where a value should always be present. :man_shrugging:

PR #6453, opened 2025-01-16
Remove safe area padding on video container

thornbill:
**Changes**
Removes the safe area padding from the video container. The controls still have the correct padding, but this allows the video to cover the entire screen.

**Before**
IMG_7346

**After**
IMG_7344

**Issues**
Related to https://github.com/jellyfin/jellyfin-expo/issues/377

### DIFF
--- a/src/plugins/htmlVideoPlayer/style.scss
+++ b/src/plugins/htmlVideoPlayer/style.scss
@@ -7,10 +7,6 @@
     display: flex;
     align-items: center;
     background: #000 !important;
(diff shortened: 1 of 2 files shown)


dmitrylyzo:
I think these should be removed as well:
https://github.com/jellyfin/jellyfin-web/blob/d158c39fa447bbaaf5e58c374488f19f0a62c876/src/plugins/youtubePlayer/style.scss#L10-L13

dmitrylyzo (approved):

PR #6485, opened 2025-01-28
Append Box Set to TMDB external link for collections

theguymadmax:
**Changes**
 Append "(Box Set)" to TheMovieDb URL when it leads to a collection.

**Issues**

### DIFF
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -718,7 +718,11 @@ function renderLinks(page, item) {
 
     if (item.ExternalUrls) {
         for (const url of item.ExternalUrls) {
-            links.push(`<a is="emby-linkbutton" class="button-link" href="${url.Url}" target="_blank">${escapeHtml(url.Name)}</a>`);
+            let linkText = escapeHtml(url.Name);
+            if (url.Url.includes('themoviedb.org/collection/')) {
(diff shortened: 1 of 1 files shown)


thornbill:
I know we discussed this on matrix, but care to rehash the specifics here about why this should be done on the client and not server... I don't recall the details and doubt I could dig that far back in element :sweat_smile: thanks!

theguymadmax:
The server distinguishes the external provider using the Type property:
  • ProviderName = "TheMovieDB"
  • Type = "Movie" or "Box Set"


Changing the ProviderName on the server would cause issues in places where it is used correctly, such as in the metadata editor, where the format is: ProviderName + Type + "Id"

If the ProviderName were changed on the server, it would result in other duplication:
Box Set Metdata


In the item details, we use just the ProviderName, which is why the TheMovieDB is duplicated. Adding the Type would be redundant, and it would appear for all names, leading to it being displayed as:
  • TheMovieDB Movie, TheMovieDB Box Set

thornbill:
Thanks for refreshing my memory! Do you think the server could provide the type in the API? It would be nice to avoid needing to check the URL... :sweat_smile:

theguymadmax:
After PR [#13175](https://github.com/jellyfin/jellyfin/pull/13175) collections links no longer show on a movie details page, so i guess this PR is no longer needed. Also, the TheMovieDB will now be displayed as TMDB.

PR #6488, opened 2025-01-30
Implement horizontal scroll seeking in video player

FrostI7Alex:
**Changes**
Implements horizontal scroll seeking in the video player, allowing to navigate through the video by scrolling left or right. Scrolling left rewinds, while scrolling right fast-forwards.
This feature will be disabled by default, with an option to enable it in the controls menu.

**Issues**
N/A


gompa:
this works on Preview on debian using a logitech mouse and pushing the scroll wheel to the side

PR #6510, opened 2025-02-02
Fix re-focusing on pause button when displaying OSD

dmitrylyzo:
**Changes**
- Add container for skip button to make it focusable.
`FocusManager.isCurrentlyFocusableInternal` doesn't work with fixed elements.
- Remove focus debounce to immediately update `document.activeElement`.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-tizen/issues/331

### DIFF
--- a/src/controllers/playback/video/index.js
+++ b/src/controllers/playback/video/index.js
@@ -1,5 +1,4 @@
 import escapeHtml from 'escape-html';
-import debounce from 'lodash-es/debounce';
 import { playbackManager } from '../../../components/playback/playbackmanager';
 import browser from '../../../scripts/browser';
 import dom from '../../../scripts/dom';
@@ -326,7 +325,16 @@ export default function (view) {
         elem.removeEventListener(transitionEndEventName, onHideAnimationComplete);
     }
 
-    const _focus = debounce((focusElement) => focusManager.focus(focusElement), 50);
+    const _focus = function (focusElement) {
+        // If no focus element is provided, try to keep current focus if it's valid,
+        // otherwise default to pause button
+        const currentFocus = focusElement || document.activeElement;
+        if (!currentFocus || !focusManager.isCurrentlyFocusable(currentFocus)) {
+            focusElement = osdBottomElement.querySelector('.btnPause');
+        }
+
+        if (focusElement) focusManager.focus(focusElement);
+    };
 
     function showMainOsdControls(focusElement) {
         if (!currentVisibleMenu) {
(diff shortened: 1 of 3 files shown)


src/controllers/playback/video/index.js
```diff
@@ -326,7 +326,93 @@ export default function (view) {
...
+    const _focus2 = function (focusElement) {
+        // If no focus element is provided, try to keep current focus if it's valid,
+        // otherwise default to pause button
+        const currentFocus = focusElement || document.activeElement;
+        if (!currentFocus || !focusManager.isCurrentlyFocusable(currentFocus)) {
+            focusElement = osdBottomElement.querySelector('.btnPause');
+        }
+
+        if (focusElement) focusManager.focus(focusElement);
+    };
```
dmitrylyzo, line 415: Hmm, do we need to debounce/timeout focus at all? `_focus2` seems to work in webOS 1.2, webOS 5, Firefox 134, Chrome 132.

Maybe timeout was for Internet Explorer? :thinking:

startswithaj: Are these changes to the focus function necessary? 

Does wrapping the skip button fix this issue by itself?

src/controllers/playback/video/index.js
```diff
@@ -326,7 +326,93 @@ export default function (view) {
...
+        debounced.cancel = cancel;
+        debounced.flush = flush;
+
+        return debounced;
+    }();
+
+    const _focus2 = function (focusElement) {
+        // If no focus element is provided, try to keep current focus if it's valid,
+        // otherwise default to pause button
+        const currentFocus = focusElement || document.activeElement;
```
startswithaj, line 409: So the main fix here is always making sure this check runs right?

What does wrapping the skip button in a container do?

startswithaj:
Should this element below have the `.focusable` class?

```
<input type="range" step=".01" min="0" max="100" value="0" is="emby-slider" class="osdPositionSlider" data-slider-keep-progress="true" />
```

I see that isCurrentlyFocusable:

```
// Determines if a focusable element can be focused at a given point in time
function isCurrentlyFocusable(elem) {
    if (!elem.classList?.contains('focusable')) {
        if (elem.disabled) {
            return false;
        }

        if (elem.getAttribute('tabindex') === '-1') {
            return false;
        }

        if (elem.tagName === 'INPUT') {
            const type = elem.type;
            if (type === 'range') {
                return false;
            }
            if (type === 'file') {
                return false;
            }
        }
    }

    return isCurrentlyFocusableInternal(elem);
}
```

Will ignore `inputs` of type `range`

src/controllers/playback/video/index.js
dmitrylyzo: > So the main fix here is always making sure this check runs right?

More precisely, `currentFocus` should be a valid element for `focusManager.isCurrentlyFocusable(currentFocus)`.

> What does wrapping the skip button in a container do?

`FocusManager.isCurrentlyFocusableInternal` doesn't work with
fixed elements - they aren't considered focusable.

dmitrylyzo:
> Should this element below have the `.focusable` class?

It is only set for TV mode:
https://github.com/jellyfin/jellyfin-web/blob/54cde6c195096a7c5bf6e66e4c44003efd933e3a/src/controllers/playback/video/index.js#L1737-L1739

src/controllers/playback/video/index.js
dmitrylyzo: Technically, there's 2 fixes:
1. Making the Skip button "focusable".
2. Fixing focus on an element passed in `showOsd`.
There're 2 versions: delayed and non-delayed focus. I'm leaning towards the latter, which is currently in use.

thornbill (approved):

PR #6516, opened 2025-02-06
Migrate resume & streaming to React

viown:
Both of these in one PR since they are very simple pages.

**Changes**
Migrates the resume and streaming pages to React + TS SDK

**Issues**


src/apps/dashboard/routes/playback/resume.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+    await getConfigurationApi(api)
+        .updateConfiguration({ serverConfiguration: config });
+
+    return {
+        isSaved: true
+    };
+};
+
+const Resume = () => {
```
thornbill, line 44: We should prefer exporting a named `Component` since that is the react router way... using the default export as the component is something I added for backwards compatibility with our existing pages

```suggestion
export const Component = () => {
```

then set the display name `Component.displayName = 'ResumePage';`

PR #6518, opened 2025-02-07
Scale the lyrics page responsively

lavafroth:
**Changes**
- Changes the lyrics text size from 30px to 1.5em. Looks proportionate on all devices.
- Adds a small breathing room of 0.5em around the lyrics container to ensure
the text does not touch the screen edges on smaller devices.
- Smoothly transition the opacity of each `lyricsLine`.

**Issues**

### DIFF
--- a/src/styles/lyrics.scss
+++ b/src/styles/lyrics.scss
@@ -1,5 +1,7 @@
 .lyricPage {
     padding-top: 4.2em !important;
+    padding-left: 0.5em !important;
+    padding-right: 0.5em !important;
     display: flex;
     justify-content: center;
 }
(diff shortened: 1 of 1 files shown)


src/styles/lyrics.scss
```diff
@@ -13,9 +15,12 @@
     display: inline-block;
     width: fit-content;
     margin: 0.1em;
-    font-size: 30px;
+    font-size: 1.5em;
```
thornbill, line 18: This seems small on desktop and `rem` is probably a better unit for font sizes.

lavafroth: Would you require me to increase the font size in that case?

thornbill: Yes, the font seems pretty small on desktop compared to what it was previously.

PR #6524, opened 2025-02-14
Convert trickplay to mui

viown:
**Changes**
Converts the trickplay page to use material ui

image

**Issues**


thornbill:
Using uncontrolled inputs is producing some errors if you save changes, navigate away from the page and back, since the default values are changed.

viown:
:thinking: I would prefer using uncontrolled here since it's a simple form.

viown:
I was able to fix this by invalidating the query after save.

Issue #6358, opened 2024-11-30
Helper text under "Date added" section of metadata edit screen points to the wrong page.


Stylo101:
**Describe The Bug**
The helper text under the "Date added" section on the metadata edit page points to the wrong location.
![Image](https://github.com/user-attachments/assets/c1fd66a0-e5c8-4c41-97c9-f041597a4af8)
The text currently states "Set up how metadata for 'Date added' is determined in the Dashboard > Libraries > NFO Settings"

However the setting presently exists under Dashboard > Libraries > Display
![Image](https://github.com/user-attachments/assets/cab7b940-6ee2-4eb2-9b85-bf93600693d3)

This appears to be tied to the ConfigureDateAdded Key in Weblate
https://translate.jellyfin.org/translate/jellyfin/jellyfin-web/en/?q=ConfigureDateAdded&sort_by=-priority%2Cposition&checksum=#translations

**Steps To Reproduce**
1. Go to any individual piece of media.
2. Hover over it and click on the three dots, and "Edit metadata"
3. Scroll down to the section that says 'Date added', and observe the text below the entry box.
4. Close this window, click the "Burger Menu" on the top left of the page.
5. Under the "Administration" section select "Dashboard"
6. On the dashboard, select the "Libraries " > "NFO Settings" section, and note the absence of the setting .
7. Now select "Libraries " > "Display" and you will see the setting in question.

**Expected Behavior**
That the ConfigureDateAdded key correctly points the user to Dashboard > Libraries > Display

**Logs**
N/A

**Screenshots**
See Above

**System (please complete the following information):**
 - Platform: Windows, Android Client
 - Browser: Firefox
 - Jellyfin Version: 10.10.3

**Additional Context**
N/A

lx1922:
Are we able to close this issue now?

thornbill:
Yes thanks for checking! In the future if you put "Fixes #6358" in the PR description, GitHub will close the issue automatically when the PR is merged. 👍

PR #6525, opened 2025-02-14
Fix Date Added helper text

lx1922:
**Changes**
Updated the helper text string in all translation json files

**Issues**
https://github.com/jellyfin/jellyfin-web/issues/6358


thornbill:
Only changes to the en-us file should be committed directly. Other languages are managed in [weblate](https://translate.jellyfin.org).

lx1922:
Sorry about the multiple commits, its been a while.
If you can squash them, that'd be great. thanks

thornbill (approved):

PR #6528, opened 2025-02-14
Remove trailing whitespace from new users' usernames

Jxiced:
**Changes**
Add validation when creating a new user to ensure that the username does not have any trailing whitespaces.

**Issues**
This is alongside this PR: https://github.com/jellyfin/jellyfin/pull/13556


src/apps/dashboard/routes/users/add.tsx
```diff
@@ -113,6 +113,13 @@ const UserNew = () => {
             const userInput: UserInput = {};
             userInput.Name = (page.querySelector('#txtUsername') as HTMLInputElement).value;
             userInput.Password = (page.querySelector('#txtPassword') as HTMLInputElement).value;
+
+            if (/^(?:\s.*|.*\s)$/.test(userInput.Name)) {
```
grhallenbeck, line 117: Why not just trim the username?

Jxiced: I believe @gnattu implied a notification to the user [here](https://github.com/jellyfin/jellyfin/pull/13556#pullrequestreview-2618792089). I could change it to trim here instead if that's better?

grhallenbeck: I can see the reasoning for validating and returning an error from the server side, but client side I don't see a reason to show an error instead of just fixing something as trivial as leading or trailing whitespace.

Jxiced: > I can see the reasoning for validating and returning an error from the server side, but client side I don't see a reason to show an error instead of just fixing something as trivial as leading or trailing whitespace.

Yeah that makes sense, please see the update. Thanks!

Bond-009:
Server PR has been merged

viown (commented):
Should we trim the username in the login part as well?

EDIT: actually I don't think a migration was done for accounts with trailing white spaces already?

thornbill:
Yeah afaik we have made no effort to migrate existing usernames so we can't trim on login

thornbill (approved):

PR #6531, opened 2025-02-18
Add a basic nix flake for development

invertedEcho:
**Changes**

This PR adds a basic nix flake, providing reproducible development environments for [nix](https://nixos.org/) users.
It defines dependencies like Node.js with version pinning, ensuring developers have all dependencies installed on their system/current shell needed to run the project.


invertedEcho:
I will convert this to ready for review once I have ensured the nix flake covers everything needed to run this project.

invertedEcho:
`sass-embedded` is broken on NixOS. As I don't want to alter the project in any way just because it is broken on a Linux Distribution, I've decided to apply a small fix in the nix flake (which will only apply for nix users) which `rm`'s the sass-embedded directores from `node_modules`. It is btw also exactly what the official Nix Package does for jellyfin-web.

Also see:
https://github.com/sass/embedded-host-node/issues/354
and
https://github.com/NixOS/nixpkgs/pull/351966/files#diff-627fe354cb06fef65417ca768bd1a60e83edfa7b20551aed65a3a0d4facd6452R46-R49

thornbill:
I'm not sure nix is popular enough to warrant maintaining this in the repo :man_shrugging:

invertedEcho:
> I'm not sure nix is popular enough to warrant maintaining this in the repo 🤷‍♂️

I see your point yes. I wonder if having a file (especially one like this) in a repository automatically means it has to be maintained and always stay up to date? 

It's just every nix developer that would come across this repository would need (not totally necessary but definitely the optimal nix workflow) to write a flake like this one.

I totally understand your point so feel free to close this PR, but still consider the point that this is a nice to have for any future nix developer, especially because it required time to figure out why the dev server would not start with a very explanatory error message, which could be saved for the next person

thornbill:
Ok I'm fine with merging this. It basically falls under the same vein as the dev container stuff. I don't want the repo to become bloated with such configs, but as long as it only remains a couple, I think it's fine.

One minor note for future PRs, we [don't use Conventional Commits](https://jellyfin.org/docs/general/contributing/development#pull-request-guidelines).

Issue #248, opened 2019-04-16
Hiding Tags on Details screen


AFunnyMeme:
**Describe the feature you'd like**
I plan to have many users on my JellyFin server, so I use the tags to filter out certain titles from other people. This works fine, but it'd be nice to have the option to hide the "Tags" from being listed on the Item Details page, so nobody is aware of titles being hidden.

PR #6560, opened 2025-02-21
Prefer hls.js for webOS

aedalzotto:
This PR simply enables hls.js for webOS TVs.
The reason for this is the strange behavior of LG's own HLS implementation that causes several problems reported here and in jellyfin-webos repositories ([#275 in jellyfin-webos](https://github.com/jellyfin/jellyfin-webos/issues/275), [#248 in jellyfin-webos](https://github.com/jellyfin/jellyfin-webos/issues/248))

I need some feedback on a few points:
* fMP4 now works and seems needed for HEVC playback. Maybe enable by default?
* Some playback profiles are not recognized anymore. E.g. VC-1 trancodes when using HLS, and DV P8 always transcodes. DV P5 in mp4 container still direct plays but P5 in mkv fails to play as it should remux.

**Changes**
Enables hlsJs for all webOS clients


dmitrylyzo:
Does hls.js work on older webOS (1, 2, 3)?

Also, I believe you need to tune the device profile in `browserDeviceProfile`.

aedalzotto:
Maybe limiting hls.js to webOS 22+ is a good idea.
I will try some changes in browserDeviceProfile.

Thanks!

aedalzotto:
I am not being able to pull this one because it seems that anything I do in browserDeviceProfile is not being honored

sundermann:
You can make the webOS app inspectable by either building one yourself and installing it through dev mode or if you are rooted you can edit `appinfo.json` and set `inspectable: true` directly. That way you can connect chrome to your TV and debug the application directly

aedalzotto:
I've created more problems than solved.
Apparently, hls.js is having problems playing DV P5, while DV P8 refuses to play completely, either falling back to HDR or triggering a transcode.
Furthermore, transcoded videos are also transcoding audio, even when the audio track is supported (eac3).

I will not be able to take a look into this issue for a while, so if someone wants to dig further, another person already [asked LG for feedback](https://forum.webostv.developer.lge.com/t/stereo-only-for-dolby-vision-video-with-multi-channel-audio-in-hls-streams/23069).

PR #6561, opened 2025-02-21
Add log viewer to dashboard

nielsvanvelzen:
Add a new page to view logs in the dashboard instead of opening a new tab with the API key in the URL. I kept it simple as I'm sure there's enough React rules I've broken already!

**Changes**
- Add new dashboard page to view a log file
- Open log viewer from dashboard logs list

**Screenshots**

Screenshot 2025-02-22 at 22-26-56 upload_Jellyfin Android TV_0 0 0-dev 1_20250221200030_7064061cfa16413693e752c01a2f2545 log

Screenshot 2025-02-22 at 22-26-39 log_20250222 log

**Planned**

If someone else is willing to do it: be my guest, otherwise I'd like to add frontmatter parsing in a future PR. Some clients (ATV) add YAML frontmatter with some information about the log file which can be used for displaying the log in various ways:
  - Client link
    The `client`, `client_version` and `client_repository` fields in the frontmatter can be used to display basic client info & a button to the GH issues
  - Markdown preview
    When the `format` field in the fronmatter is set to `markdown` we should render the log as markdown (copy still copies raw file, download will set media type to markdown)

Additionally, it might be useful to add a "live reload" feature where the log file is refetched every few seconds. This would be useful for watching the server logs of the current day or an active ffmpeg transcode.

**Issues**

### DIFF
--- a/src/apps/dashboard/features/logs/api/useServerLog.ts
+++ b/src/apps/dashboard/features/logs/api/useServerLog.ts
@@ -1,0 +1,35 @@
+import { Api } from '@jellyfin/sdk';
+import { getSystemApi } from '@jellyfin/sdk/lib/utils/api/system-api';
+import { useQuery } from '@tanstack/react-query';
+import { useApi } from 'hooks/useApi';
+import type { AxiosRequestConfig } from 'axios';
+
+const fetchServerLog = async (
+    api?: Api,
+    name?: string,
+    options?: AxiosRequestConfig
+) => {
+    if (!api) {
+        console.error('[useServerLog] No API instance available');
+        return;
+    }
+
+    if (!name) {
+        console.error('[useServerLog] Name is required');
+        return;
+    }
+
+    const response = await getSystemApi(api).getLogFile({ name }, options);
+
+    // FIXME: TypeScript SDK thinks it is returning a File but in reality it is a string
+    return response.data as never as string;
+};
+export const useServerLog = (name: string) => {
+    const { api } = useApi();
+
+    return useQuery({
+        queryKey: ['ServerLog', name],
+        queryFn: ({ signal }) => fetchServerLog(api, name, { signal }),
+        enabled: !!api
+    });
+};
--- a/src/apps/dashboard/features/logs/api/useServerLog.ts
+++ b/src/apps/dashboard/features/logs/api/useServerLog.ts
@@ -5,20 +5,14 @@ import { useApi } from 'hooks/useApi';
 import type { AxiosRequestConfig } from 'axios';
 
 const fetchServerLog = async (
-    api?: Api,
-    name?: string,
+    api: Api | undefined,
+    name: string,
     options?: AxiosRequestConfig
 ) => {
     if (!api) {
         console.error('[useServerLog] No API instance available');
         return;
     }
-
-    if (!name) {
-        console.error('[useServerLog] Name is required');
-        return;
-    }
-
     const response = await getSystemApi(api).getLogFile({ name }, options);
 
     // FIXME: TypeScript SDK thinks it is returning a File but in reality it is a string
--- a/src/apps/dashboard/features/logs/api/useServerLog.ts
+++ b/src/apps/dashboard/features/logs/api/useServerLog.ts
@@ -5,7 +5,7 @@ import { useApi } from 'hooks/useApi';
 import type { AxiosRequestConfig } from 'axios';
 
 const fetchServerLog = async (
-    api: Api | undefined,
+    api: Api,
     name: string,
     options?: AxiosRequestConfig
 ) => {
@@ -23,7 +23,7 @@ export const useServerLog = (name: string) => {
 
     return useQuery({
         queryKey: ['ServerLog', name],
-        queryFn: ({ signal }) => fetchServerLog(api, name, { signal }),
+        queryFn: ({ signal }) => fetchServerLog(api!, name, { signal }),
         enabled: !!api
     });
 };
--- a/src/apps/dashboard/features/logs/api/useServerLog.ts
+++ b/src/apps/dashboard/features/logs/api/useServerLog.ts
@@ -9,10 +9,6 @@ const fetchServerLog = async (
     name: string,
     options?: AxiosRequestConfig
 ) => {
-    if (!api) {
-        console.error('[useServerLog] No API instance available');
-        return;
-    }
     const response = await getSystemApi(api).getLogFile({ name }, options);
 
     // FIXME: TypeScript SDK thinks it is returning a File but in reality it is a string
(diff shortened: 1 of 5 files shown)


src/apps/dashboard/routes/logs/file.tsx
```diff
@@ -0,0 +1,112 @@
...
+                    {!error && !loading && (
+                        <>
+                            <ButtonGroup variant='contained' sx={{ mt: 2 }}>
+                                <Button
+                                    disabled={!('clipboard' in navigator)}
+                                    startIcon={<ContentCopy />}
+                                    onClick={copyToClipboard}
+                                >
+                                    {globalize.translate('Copy')}
+                                </Button>
```
viown, line 88: Maybe we can show a success style effect when clicked? We have [copied](https://github.com/jellyfin/jellyfin-web/blob/7904bdd981cea1bfcd89dd6bbcf40bf90fe64369/src/strings/en-us.json#L186). A toast message would probably work fine too.

src/apps/dashboard/features/logs/api/useServerLog.ts
```diff
@@ -0,0 +1,35 @@
+import { Api } from '@jellyfin/sdk';
+import { getSystemApi } from '@jellyfin/sdk/lib/utils/api/system-api';
+import { useQuery } from '@tanstack/react-query';
+import { useApi } from 'hooks/useApi';
+import type { AxiosRequestConfig } from 'axios';
+
+const fetchServerLog = async (
+    api?: Api,
+    name?: string,
```
viown, line 9:
```suggestion
    api: Api,
    name: string,
```

I think we can just use a non-null assert on `api` and make `name` required.

viown (commented):
Looks good! Just some minor feedback and suggestions.

src/apps/dashboard/features/logs/api/useServerLog.ts
nielsvanvelzen: The useServerLog function uses `useApi()` to get the API instance, however that value can be undefined. I took this from the existing useServerLogs function.
The name parameter I can indeed make required.

src/apps/dashboard/routes/logs/file.tsx
```diff
@@ -0,0 +1,112 @@
...
+import {
+    Alert,
+    Box,
+    Button,
+    ButtonGroup,
+    Card,
+    CardContent,
+    Container,
+    Typography
+} from '@mui/material';
```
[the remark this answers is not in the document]
nielsvanvelzen: ESLint will immediatly complain if I do that

`'@mui/material' import is duplicated.eslint` [no-duplicate-imports](https://eslint.org/docs/latest/rules/no-duplicate-imports)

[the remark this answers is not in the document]
dmitrylyzo: > ESLint will immediatly complain if I do that
> 
> `'@mui/material' import is duplicated.eslint` [no-duplicate-imports](https://eslint.org/docs/latest/rules/no-duplicate-imports)

Did you use different modules?
```ts
import Alert from '@mui/material/Alert';
import Button from '@mui/material/Button';
...
```

src/apps/dashboard/features/logs/api/useServerLog.ts
viown: Since we are using `enabled: !!api` in `useQuery`, the query will never run unless `api` exists, and the check for it in `fetchServerLog` is pretty much only to make typescript happy.

So I think it's valid for a non-null assert like so:
```ts
queryFn: ({ signal }) => fetchServerLog(api!, name, { signal })
```

src/apps/dashboard/routes/logs/file.tsx
```diff
@@ -0,0 +1,112 @@
...
+                            key='error'
+                            severity='error'
+                            sx={{ mt: 2 }}
+                            action={
+                                <Button
+                                    color='inherit'
+                                    size='small'
+                                    onClick={retry}
+                                >
+                                    Retry
```
viown, line 69: should be localized.

src/apps/dashboard/features/logs/api/useServerLog.ts
```diff
@@ -0,0 +1,29 @@
...
+
+const fetchServerLog = async (
+    api: Api,
+    name: string,
+    options?: AxiosRequestConfig
+) => {
+    if (!api) {
+        console.error('[useServerLog] No API instance available');
+        return;
+    }
```
viown, line 15:
```suggestion

```
We can remove this now since it's no longer needed.

src/apps/dashboard/routes/logs/file.tsx
```diff
@@ -0,0 +1,112 @@
...
+        isError: error,
+        isPending: loading,
+        data: log,
+        refetch
+    } = useServerLog(fileName ?? '');
+
+    const retry = useCallback(() => refetch(), [refetch]);
+
+    const copyToClipboard = useCallback(async () => {
+        if ('clipboard' in navigator && log) {
```
thornbill, line 30: Use the `copy` function from `scripts/clipboard` for backwards compatibility.

src/apps/dashboard/routes/logs/file.tsx
```diff
@@ -0,0 +1,112 @@
...
+                                <Button
+                                    startIcon={<FileDownload />}
+                                    onClick={downloadFile}
+                                >
+                                    {globalize.translate('Download')}
+                                </Button>
+                            </ButtonGroup>
+
+                            <Card sx={{ mt: 2 }}>
+                                <CardContent>
```
thornbill, line 98: Maybe just use `<Paper>` instead of a card?

src/apps/dashboard/features/logs/components/LogItemList.tsx
```diff
@@ -34,15 +21,14 @@ const LogItemList: FunctionComponent<LogItemProps> = ({ logs }: LogItemProps) =>
             {logs.map(log => {
                 return (
                     <ListItem key={log.Name} disablePadding>
-                        <ListItemButton href={getLogFileUrl(log)} target='_blank'>
+                        <ListItemLink to={`/dashboard/logs/${log.Name}`}>
```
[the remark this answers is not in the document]
thornbill: I couldn't find a definitive answer for this, but it _seems_ like react router should handle it automatically.

src/apps/dashboard/routes/logs/file.tsx
nielsvanvelzen: I did switch it to Paper but had to manually add the padding to it. I also noticed I forgot to add `overflow:auto` for large lines so added that now as well.

src/apps/dashboard/routes/logs/file.tsx
```diff
@@ -0,0 +1,110 @@
...
+                        </Alert>
+                    )}
+
+                    {loading && <Loading />}
+
+                    {!error && !loading && (
+                        <>
+                            <ButtonGroup variant='contained' sx={{ mt: 2 }}>
+                                <Button
+                                    disabled={!('clipboard' in navigator)}
```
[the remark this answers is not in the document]
nielsvanvelzen: I should've seen that one, oops

thornbill (approved):

PR #6564, opened 2025-02-21
Redesign library navigation in experimental layout

thornbill:
**Changes**
Draft so people can test it out. The code is a mess... just wanting feedback on the UX before investing more time.

Third time is the charm?

**Issues**
N/A

### DIFF
--- a/src/RootAppRouter.tsx
+++ b/src/RootAppRouter.tsx
@@ -17,7 +17,7 @@ import { createRouterHistory } from 'components/router/routerHistory';
 import UserThemeProvider from 'themes/UserThemeProvider';
 
 const layoutMode = localStorage.getItem('layout');
-const isExperimentalLayout = layoutMode === 'experimental';
+const isExperimentalLayout = layoutMode === 'experimental' || !layoutMode;
 
 const router = createHashRouter([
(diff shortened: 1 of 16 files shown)


jhparker88:
I find a rather tedious to navigate via those different dropdowns [1] [2]. Having something at the top (or left which is made visible when one is using the mouse over it or a remote control like done in amazon prime video) to navigate to broad categories on Movies | Shows | Music | Collections should help. Moving the second [2] navigational entity to a dropdown seems rather counter productive for having easy access those categories and confusing when placed near to each other.

[1]
Screenshot 2025-02-24 21 42 01

[2]
Screenshot 2025-02-24 21 42 12

jhparker88:
While I'm add it, I noticed a glitch in the deployed cloudfare instance, the user branded icon [1] has a bluish background [1, 2] which is caused by `background-color: rgb(0, 114, 154);` applied to the `MuiAvatar-root MuiAvatar-circular css-kvlks5` element.


Screenshot 2025-02-24 22 07 17
Screenshot 2025-02-24 22 07 59

thornbill:
Closed in favor of #6582

PR #6570, opened 2025-02-22
Fix catalog plugin page not setting page title

nielsvanvelzen:
I'm a React expert now.

**Changes**
Fix catalog plugin page not setting page title, before this PR it would keep the title of the last page (most commonly "Catalog").

**Issues**


nielsvanvelzen:
Changed the base to release branch as I noticed the problem already exists in 10.10

thornbill (approved):

PR #6571, opened 2025-02-23
metadata update function catch case fixes #6559

### LINKED ISSUE #6559: Endlessly spin when adding people without selecting type

### Describe the bug

When adding people to an item, if a type is not selected, clicking the save button will cause an endlessly spinning loading icon

### Reproduction Steps

1. Edit metadata on a video
2. Add people without selecting type
3. Hit save

### Expected/Actual behaviour

Expected: The app should either display an error or save without issues. 
Actual: Endlessly spinning

### Logs

_No response_

### Server version

10.10.6

### Web version

10.10.6

### Build version

10.10.6

### Platform

macOS

### Browser

Firefox

### Additional information

The request to the server fails with error code 400

jagadam97:
Fix: catch block for metadataupdate function to fix infinite loader when update api fails fixes #6559

**Changes**
1. as described in #6559 there is a infinite loader in update metadata when the api fails. This can be avoided by using try catch method. as such the same has been implemented in this PR. 

**Issues**
fixes #6559 

**Proof of testing**
https://youtu.be/3YHXEthmzoY

### DIFF
--- a/src/components/metadataEditor/metadataEditor.js
+++ b/src/components/metadataEditor/metadataEditor.js
@@ -38,35 +38,32 @@ function closeDialog() {
     }
 }
 
-function submitUpdatedItem(form, item) {
-    function afterContentTypeUpdated() {
+async function submitUpdatedItem(form, item) {
+    const afterContentTypeUpdated = () => {
         toast(globalize.translate('MessageItemSaved'));
 
         loading.hide();
         closeDialog();
-    }
-
+    };
     const apiClient = getApiClient();
 
-    apiClient.updateItem(item).then(function () {
-        const newContentType = form.querySelector('#selectContentType').value || '';
+    try {
+        await apiClient.updateItem(item);
+        const newContentType = form.querySelector('#selectContentType')?.value || '';
 
-        if ((metadataEditorInfo.ContentType || '') !== newContentType) {
-            apiClient.ajax({
+        await apiClient.ajax({
+            url: apiClient.getUrl(`Items/${item.Id}/ContentType`, {
+                ContentType: newContentType
+            }),
 
-                url: apiClient.getUrl('Items/' + item.Id + '/ContentType', {
-                    ContentType: newContentType
-                }),
+            type: 'POST'
+        });
 
-                type: 'POST'
-
-            }).then(function () {
-                afterContentTypeUpdated();
-            });
-        } else {
-            afterContentTypeUpdated();
-        }
-    });
+        afterContentTypeUpdated();
+    } catch {
+        toast(globalize.translate('MessageItemSaveFailed'));
+        loading.hide();
+    }
 }
 
 function getSelectedAirDays(form) {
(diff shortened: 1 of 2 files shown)


src/components/metadataEditor/metadataEditor.js
```diff
@@ -38,35 +38,33 @@ function closeDialog() {
...
+    };
 
-        if ((metadataEditorInfo.ContentType || '') !== newContentType) {
-            apiClient.ajax({
+    try {
+        await apiClient.updateItem(item);
+        const newContentType = form.querySelector('#selectContentType')?.value || '';
 
-                url: apiClient.getUrl('Items/' + item.Id + '/ContentType', {
+        if ((metadataEditorInfo?.ContentType || '') !== newContentType) {
```
dmitrylyzo, line 54: We probably shouldn't even start doing anything if `metadataEditorInfo` is undefined/null. Maybe just throw an exception at the beginning of the function instead of optional chaining?

src/styles/videoosd.scss
```diff
@@ -85,7 +85,7 @@
     bottom: 0;
     left: 0;
     right: 0;
-    background: rgba(0, 0, 0, 0.7);
+    background: none;
```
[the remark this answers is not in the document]
viown: I would split this fix into a separate PR targeting the `release-10.10.z` branch.

src/components/metadataEditor/metadataEditor.js
jagadam97: seeing that this check is to just update the items content type when there is change in content type and ignore if otherwise. we can directly ignore the check and lets do the update content api even if its redundant.

src/components/metadataEditor/metadataEditor.js
```diff
@@ -38,35 +38,32 @@ function closeDialog() {
...
-    function afterContentTypeUpdated() {
+async function submitUpdatedItem(form, item) {
+    const afterContentTypeUpdated = () => {
         toast(globalize.translate('MessageItemSaved'));
 
         loading.hide();
         closeDialog();
-    }
-
+    };
```
thornbill, line 47: This is only called once now, so please remove the function and inline the contents.

src/components/metadataEditor/metadataEditor.js
```diff
@@ -38,35 +38,32 @@ function closeDialog() {
...
 
-    apiClient.updateItem(item).then(function () {
-        const newContentType = form.querySelector('#selectContentType').value || '';
+    try {
+        await apiClient.updateItem(item);
+        const newContentType = form.querySelector('#selectContentType')?.value || '';
 
-        if ((metadataEditorInfo.ContentType || '') !== newContentType) {
-            apiClient.ajax({
+        await apiClient.ajax({
```
thornbill, line 54: You removed the check to see if the content type changed?

thornbill (changes requested):

PR #6577, opened 2025-02-25
Fix 6576 by preferring Logo image on Guide

### LINKED ISSUE #6576: [bug] Web TV Guide Channel Image Mapping

### Describe the bug

The Guide view renders the channel images from the Primary image. However, Primary seems to be more for poster images (as seen in movies). The aspect ratio of the guide logo and Primary image on the details page are basically inverted. Having the guide render from a different image type may be better. Perhaps Logo? Thumb?

Here's an example of a channel's detail page:
Image

And an example of the same channel in the guide view:
Image

By using Logo as Primary the channel details page renders incorrectly (though admittedly the guide looks better):
Image

Image

### Reproduction Steps

1. Upload a poster sized image for Primary on a Channel
2. Look at Guide view on Live TV

### Expected/Actual behaviour

Channel logo should render a Logo or Thumb and the details page should render the poster (which is default behavior across collections)

### Logs

_No response_

### Server version

10.10.3

### Web version

10.10.3

### Build version

10.10.3

### Platform

Any app using the jellyfin-web app

### Browser

All

### Additional information

_No response_

robertdown:
**Changes**
If a Channel has a Logo image, use that image on the guide over a Primary image type. This change ensures current state does not break by preferring a Logo type but falling back to a Primary type if no Logo exists.

**Issues**
Fixes #6576

### DIFF
--- a/src/strings/vi.json
+++ b/src/strings/vi.json
@@ -1997,7 +1997,7 @@
     "CustomSubtitleStylingHelp": "Kiểu phụ đề sẽ hoạt động trên hầu hết các thiết bị nhưng đi kèm với chi phí hiệu suất bổ sung.",
     "LabelSubtitleStyling": "Kiểu phụ đề",
     "Native": "Gốc",
-    "LibraryNameInvalid": "Tên thư viện không thể trống hoặc có khoảng trống ở đầu/cuối.",
+    "LibraryNameInvalid": "Tên thư viện không để trống.",
     "LabelDevice": "Thiết bị",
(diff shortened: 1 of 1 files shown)


thornbill:
This change is interesting... I believe our live tv guide providers are written so that the logo is used for the primary image. I definitely think there is an argument for **not** doing that and updating the clients to use the primary image as a fallback as you have done here, but I think that would need to start with backend updates first... and coordination among other client apps.

PR #6583, opened 2025-02-26
Fix skip button not displaying correctly with OSD

rlauuzo:
Resolves an issue where the skip button for media segments fails to appear on the OSD after the first episode in a playlist.  The problem was caused by removing the OSD event listener on playback stop.

### DIFF
--- a/src/components/playback/skipsegment.ts
+++ b/src/components/playback/skipsegment.ts
@@ -188,7 +188,7 @@ class SkipSegment extends PlaybackSubscriber {
     onPlaybackStop() {
         this.currentSegment = null;
         this.hideSkipButton();
-        if (!this.playbackManager.getCurrentPlayer()) {
+        if (!this.playbackManager.getNextItem()) {
             Events.off(document, EventType.SHOW_VIDEO_OSD, this.onOsdChanged);
         }
     }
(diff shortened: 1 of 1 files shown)


src/components/playback/skipsegment.ts
```diff
@@ -188,7 +188,9 @@ class SkipSegment extends PlaybackSubscriber {
     onPlaybackStop() {
         this.currentSegment = null;
         this.hideSkipButton();
-        Events.off(document, EventType.SHOW_VIDEO_OSD, this.onOsdChanged);
+        if (!this.playbackManager.getCurrentPlayer()) {
```
thornbill, line 191: The `onPlaybackStop` callback is called with `PlaybackStopInfo` which includes the `nextItem` (when present). I think it would be better to check for the existence of a `nextItem` for this.

https://github.com/jellyfin/jellyfin-web/blob/master/src/apps/stable/features/playback/utils/playbackSubscriber.ts#L19

rlauuzo: Now checks for the next item using playbackManager.getNextItem().

rlauuzo: I’ve updated to check the parameter value now.

thornbill (approved):

rlauuzo:
@thornbill this isn’t in master yet. Do you want me to PR it again or will you handle it?

thornbill:
Hey @rlauuzo, sorry this one slipped my mind. I had tried running our backport script but due to the age it didn't get picked up and I forgot to handle it manually before the release. I have opened #7219 now so it will land in the next bugfix release (10.11.1) and get backported to master from there.

PR #6584, opened 2025-02-27
Fix text truncation bug

### LINKED ISSUE #5608: Dashboard/Users fields truncated.
**Describe The Bug**

Since the update from Jellyfin 10.8.13 to 10.9.2, the appearance of the dashboard has changed slightly. In the user window, the user fields are now truncated. I have uploaded an image from both 10.8.13 and 10.9.2 for clarification.

**Steps To Reproduce**

1. Go to '...../web/#/dashboard/users'
2. See error

 - Platform: Windows
 - Browser: Chrome
 - Jellyfin Version: 10.9.2
 - Server: Debian 10.9.2

**Screenshots** 

10.8.13 - OK

10.9.2 - NOT OK

MindfulMinun:
The fixed height can cause the text to be cut off in other languages

**Changes**
- Updated a CSS rule to use `min-height` instead of a fixed `height`

**Issues**
Fixes #5608

### DIFF
--- a/src/styles/dashboard.scss
+++ b/src/styles/dashboard.scss
@@ -216,7 +216,7 @@ div[data-role=controlgroup] a.ui-btn-active {
 
 .localUsers .cardText-secondary {
     white-space: pre-wrap;
-    height: 3em;
+    min-height: 3em;
 }
 
 .customCssContainer textarea {


MindfulMinun:
Note how the first user's subtitle is slightly longer than the rest of the cards

p0358:
Yeah that does not look good at all (the varying height in a single row)

MindfulMinun:
Any other suggestions?

p0358:
Yes, as I suggested in this comment: https://github.com/jellyfin/jellyfin-web/issues/5608#issuecomment-2347033113 (assuming it still works with current CSS)

If everything in one row can adapt to the tallest entry in a single row, then it should look pretty acceptable. Perhaps most idealistically everything would have the same height, but I feel like that'd then involve some unreasonable sorcery to code it...

MindfulMinun:
See https://github.com/jellyfin/jellyfin-web/pull/6589

PR #6589, opened 2025-02-28
Fix text truncation for user cards in dashboard

### LINKED ISSUE #5608: Dashboard/Users fields truncated.

MindfulMinun:
The fixed height can cause the text to be cut off in other languages

**Changes**
- Updated a CSS rule to use `min-height` instead of a fixed `height` for the text description
- Added a `height: 100%` to the card box so these share the same height, the height of the longest card.

**Issues**
Fixes #5608


MindfulMinun:
Thanks @p0358, that means a lot, especially since this is my first contribution to this project :)

PR #6595, opened 2025-03-05
Add support for HEVC codec in Hisense Vidaa OS

mvshao:
**Changes**
The Jellyfin has a problem playing HEVC videos on Hisense TV with Vidaa OS. That PR is meant to fix this issue. The idea was taken from [Jellyfin forum](https://forum.jellyfin.org/t-setting-up-jellyfin-on-a-hisense-vidaa-os-tv), and it solves the issue.

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -11,7 +11,7 @@ function canPlayHevc(videoTestElement, options) {
         return true;
     }
 
-    if (browser.ps4) {
+    if (browser.ps4 || browser.vidaa) {
         return false;
     }
 


mvshao:
@dmitrylyzo Sorry, my bad in naming the PR, I was just thinking about my next steps. Good catch thx!

viown:
Wasn't this [already disabled](https://github.com/jellyfin/jellyfin-web/pull/6135)?

mvshao:
@viow I got some hiccups/artifacts on the newest version of Jellyfin-Web on Hisesne, and that change fixed the problem. Maybe it is an issue with my TV. I will leave the decision if we need to merge it or not to you guys!

dmitrylyzo:
> Wasn't this [already disabled](https://github.com/jellyfin/jellyfin-web/pull/6135)?

No, we seem to have come to the conclusion that the problem is with MKV, not HEVC.

Yagisan:
I have a user that is reporting the same issue. I can see clearly in the logs that it is remuxing the HEVC content in MKV containers to MP4 containers, but the Hisense TV with Vidaa OS is unable to actually play the content.

It has no trouble playing back H264 content, so I would prefer to simply transcode any HEVC content to H264 sent to that device. Please do consider this patch for the next update.

dmitrylyzo:
> I have a user that is reporting the same issue. I can see clearly in the logs that it is remuxing the HEVC content in MKV containers to MP4 containers, but the Hisense TV with Vidaa OS is unable to actually play the content.

What TV model?
Ask them to try disabling `Prefer fMP4` (remux to TS).

> It has no trouble playing back H264 content, so I would prefer to simply transcode any HEVC content to H264 sent to that device. Please do consider this patch for the next update.

From the original [PR](https://github.com/jellyfin/jellyfin-web/pull/6135), HEVC worked on some models.

Yagisan:
The TV Model is a Hisense 55A7G
Software version: V0000.06.12X.N1212
Extended Version: N1213

It failed to playback HEVC content with `Prefer fMP4` both enabled and disabled.
My current workaround is to globally disable HEVC transcoding - but I would really like to turn that back on if possible.

dmitrylyzo:
> The TV Model is a Hisense 55A7G
> Software version: V0000.06.12X.N1212
> Extended Version: N1213

Maybe we can detect this model/OS version and disable HEVC only for it?

Looks like A7 is 2021, U7 (which seems to work) is 2024.

> It failed to playback HEVC content with `Prefer fMP4` both enabled and disabled.

Could you remux MKV to MP4 and check if it plays HEVC in DirectPlay mode?

Yagisan:
@dmitrylyzo Sorry for the delay in responding.

The user has provided the following feedback to me:

Content transcoded to HEVC always stops after 8 minutes of playback, regardless of container.
Content transcoded to H264 always stops after 16 minutes of playback, regardless of container.
Direct Playing content plays the entire file regardless of being HEVC, H264 or container.

I'm pretty sure something on that system is broken, but there are no updates for it so it's probably unfixable.

The user got so fed up with the VIDAA system that they went out and bought a Fire TV stick instead and is now using that.

thornbill:
So is this no longer needed?

PR #6601, opened 2025-03-12
Add basic 404 page and *.html redirect

thornbill:
**Changes**
* Adds a basic 404 page and redirect for `*.html` paths when a match is not found

Screenshot 2025-03-12 at 17-18-53 Jellyfin

(A follow up PR will remove `.html` from all route paths.)

**Issues**
N/A


src/apps/stable/routes/routes.tsx
```diff
@@ -28,7 +29,13 @@ export const STABLE_APP_ROUTES: RouteObject[] = [
 
             /* Public routes */
             { index: true, element: <Navigate replace to='/home.html' /> },
-            ...LEGACY_PUBLIC_ROUTES.map(toViewManagerPageRoute)
+            ...LEGACY_PUBLIC_ROUTES.map(toViewManagerPageRoute),
+
+            /* Fallback route for invalid paths */
+            {
+                path: '*',
+                Component: FallbackRoute
```
viown, line 37: Refreshing the page does cause the header not to load. This is similar to an issue I fixed [here](https://github.com/jellyfin/jellyfin-web/pull/6388), which in this case is likely caused by the page loading without the connection from `ConnectionRequired` being created.

On the experimental layout, the user profile gets hidden on refresh (like in your screenshot).

At a glance, I can't think of a good way to fix this other than possibly just removing the header since this is a non-user route.

viown (commented):
On the stable app, the text is all the way at the left of the screen (not centered).

src/apps/stable/routes/routes.tsx
thornbill: Good catch. We should *always* be wrapping routes with `<ConnectionRequired>`. I'm not sure why we weren't for public routes and how that hasn't already caused issues... :sweat_smile:

src/components/router/FallbackRoute.tsx
```diff
@@ -0,0 +1,54 @@
...
+                to={to}
+            />
+        );
+    }
+
+    return (
+        <Page
+            id='fallbackPage'
+            className='mainAnimatedPage libraryPage'
+        >
```
viown, line 39:
```suggestion
        <Page
            id='fallbackPage'
            title={globalize.translate('HeaderPageNotFound')}
            className='mainAnimatedPage libraryPage'
        >
```

Maybe we can add the title?

viown (approved):

PR #6610, opened 2025-03-15
Add space key functionality for playback control youtube like

Ale4224:
**Changes**

Implemented handling for the space key to toggle play/pause and adjust playback rate. Introduced a timeout mechanism to increase playback speed if the space key is held down for more than 500ms. Added corresponding keyup event listener to manage playback state effectively. Moved the play/pause current functionality to keyUp event.


Eyt-Lev:
maybe add a label at the top to indicate when in `2x speed` mode, and implement this functionality for a mouse (that should also do it for mobile)

PR #6614, opened 2025-03-16
Require password before setting as admin

viown:
https://github.com/jellyfin/jellyfin/pull/12935

**Changes**

**Issues**


thornbill:
We can start enforcing this in web before the server PR lands imo

viown:
Currently, this will prevent already-existing admins with no password from updating their policy. Waiting on [this review](https://github.com/jellyfin/jellyfin/pull/12935#discussion_r1997703254) to then adjust accordingly.

PR #6616, opened 2025-03-17
Add splash screen preview and upload/delete

thornbill:
**Changes**
* Adds a preview of the current splash screen image
* Adds support for uploading and deleting a custom splash screen image
* Refactors the plugin image component for general use

Screenshot 2025-03-17 at 13-57-20 Branding

**Issues**
N/A


src/apps/dashboard/routes/branding/index.tsx
```diff
@@ -107,16 +189,75 @@ export const Component = () => {
...
+                                        onChange={onSplashscreenUpload}
+                                    />
+                                    {globalize.translate('UploadCustomImage')}
+                                </Button>
+
+                                <Button
+                                    variant='outlined'
+                                    color='error'
+                                    startIcon={<Delete />}
+                                    disabled={!brandingOptions.SplashscreenEnabled}
```
nielsvanvelzen, line 254: Would it be possible to disable this button if there is no custom splash screen set? Or do we not have enough information to know that?

thornbill: As far as I know we have no way of knowing that.

viown (commented):
Updating the named configuration will reset the custom splash screen to its original state (happens when I hit save). Looks like a server issue? Maybe as a workaround, we can re-upload the splash screen whenever the user hits save.

thornbill:
Good catch @viown that's definitely a server bug.

src/apps/dashboard/routes/branding/index.tsx
```diff
@@ -60,22 +66,91 @@ export const loader = () => {
...
+
+    const setSplashscreenEnabled = useCallback((event: React.ChangeEvent<HTMLInputElement>, isEnabled: boolean) => {
         setBrandingOptions({
             ...brandingOptions,
             [BrandingOption.SplashscreenEnabled]: isEnabled
         });
-    }, [ brandingOptions ]);
+
+        submit(event.target.form);
+    }, [ brandingOptions, submit ]);
```
viown, line 153: Toggling the enable splash screen switch will save the entire form. This is not really a big deal, but it could lead to unintended changes being saved.

Quickfix:
```js
const setSplashscreenEnabled = useCallback(async (event: React.ChangeEvent<HTMLInputElement>, isEnabled: boolean) => {
    await getConfigurationApi(api!)
        .updateNamedConfiguration({
            key: BRANDING_CONFIG_KEY,
            body: JSON.stringify({
                ...defaultBrandingOptions,
                SplashscreenEnabled: isEnabled
            })
        });

    await queryClient.invalidateQueries({
        queryKey: [ QUERY_KEY ]
    });
}, [ api, defaultBrandingOptions ]);
```
Code above requires making the branding page uncontrolled. See below.

src/apps/dashboard/routes/branding/index.tsx
```diff
@@ -60,22 +66,91 @@ export const loader = () => {
...
     const navigation = useNavigation();
     const actionData = useActionData() as ActionData | undefined;
     const isSubmitting = navigation.state === 'submitting';
+    const submit = useSubmit();
 
     const {
         data: defaultBrandingOptions,
         isPending
     } = useBrandingOptions();
     const [ brandingOptions, setBrandingOptions ] = useState(defaultBrandingOptions || {});
```
viown, line 79: Suggestion: I think we can benefit by making the branding page uncontrolled and removing this `useState` and using `defaultBrandingOptions` directly. Reduces some complexity. What do you think?

src/apps/dashboard/routes/branding/index.tsx
```diff
@@ -107,16 +189,75 @@ export const Component = () => {
...
+
+                            <Stack
+                                spacing={{ xs: 3, sm: 2 }}
+                                sx={{ flex: '1 1 0' }}
+                            >
+                                <FormControlLabel
+                                    control={
+                                        <Switch
+                                            name={BrandingOption.SplashscreenEnabled}
+                                            checked={brandingOptions?.SplashscreenEnabled}
```
viown, line 224: We can remove `undefined` from the type union by checking for `isError`.

thornbill:
I disabled the upload/delete custom image options for now until the server issue can be addressed.

KGT1:
Hey, the server issue should be fixed. just an fyi: i think you can still upload and delete the splashscreen even when it is disabled. they are also not deleted when disabling, so it would probably make sense to make it viewable to admins even when disabled. ill take a look later.

viown (approved):

PR #6644, opened 2025-03-23
Fix active device cards from getting too large/overflowing on some resolutions

jessielw:
**Changes**
I noticed on my 125% scaled 1080p monitor that some of the **Active Device** cards were over flowing into the right hand side of the screen. 

**Before** 
fix1

**After**
fix2

I was looking for a more elegant solution, but `dashboardSection` is used in multiple places on the dashboard. This should suffice and still work great.

### DIFF
--- a/src/styles/dashboard.scss
+++ b/src/styles/dashboard.scss
@@ -268,7 +268,7 @@ div[data-role=controlgroup] a.ui-btn-active {
 }
 
 .activeSession {
-    min-width: 20rem;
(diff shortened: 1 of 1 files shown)


thornbill:
In the future please only merge master back into your branch if there are conflicts. Also, we [prefer human readable](https://jellyfin.org/docs/general/contributing/development/#pull-request-guidelines) commit messages to the "conventional commits" style. Thanks!

PR #6649, opened 2025-03-24
new styling to lyrics page

gm852:
**Changes**
Added album cover, album name, creator, and song name to the lyrics page for looks. Works for both small and large screens.

**Issues**
N/A

---
image

image


gm852:
sorry for so many commits, lint on my end kept saying no issues up until now.

PR #6656, opened 2025-03-24
Add opensearch search plugin

Kwonunn:
**Changes**
Add an [OpenSearch](https://developer.mozilla.org/en-US/docs/Web/XML/Guides/OpenSearch) search plugin description file, to let browsers search directly on Jellyfin instances.

This currently contains a hardcoded URL pointing at localhost, because I don't know how to use react to create a dynamic response. I imagine it is fairly easy.

### DIFF
--- a/src/opensearch.xml
+++ b/src/opensearch.xml
@@ -1,0 +1,9 @@
+<OpenSearchDescription
+  xmlns="http://a9.com/-/spec/opensearch/1.1/"
+  xmlns:moz="http://www.mozilla.org/2006/browser/search/">
+  <ShortName>Jellyfin</ShortName>
+  <Description>Jellyfin Media Server</Description>
+  <InputEncoding>UTF-8</InputEncoding>
+  <Image width="256" height="256" type="image/x-icon">favicon.png</Image>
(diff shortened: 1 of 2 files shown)


src/opensearch.xml
```diff
@@ -0,0 +1,9 @@
+<OpenSearchDescription
+  xmlns="http://a9.com/-/spec/opensearch/1.1/"
+  xmlns:moz="http://www.mozilla.org/2006/browser/search/">
+  <ShortName>Jellyfin</ShortName>
+  <Description>Jellyfin Media Server</Description>
+  <InputEncoding>UTF-8</InputEncoding>
+  <Image width="256" height="256" type="image/x-icon">favicon.png</Image>
+  <Url type="text/html" template="http://localhost:8080/#/search?query={searchTerms}" />
```
Kwonunn, line 8: The URL in this line needs to dynamically changed to the root URL of the instance.

thornbill:
This would likely need to be handled server side. The web client cannot generate an XML file with the correct server address.

Kwonunn:
That makes sense, in that case I'll see what I can do server-side. I will need to add the link tag to the main HTML, so I'll leave this PR open for now.

PR #6662, opened 2025-03-25
Move themes to sass

thornbill:
**Changes**
* Move themes to sass files
* Update webpack config to build sass themes

**Issues**
N/A


webpack.dev.js
```diff
@@ -1,11 +1,26 @@
...
     // In order for live reload to work we must use "web" as the target not "browserslist"
     target: process.env.WEBPACK_SERVE ? 'web' : 'browserslist',
     mode: 'development',
-    entry: { 'main.jellyfin': './index.jsx' },
+    entry: {
+        'main.jellyfin': './index.jsx',
+        ...THEMES.reduce((acc, theme) => {
+            acc[`themes/${theme}`] = `./themes/${theme}/theme.scss`;
+            return acc;
+        }, {})
```
viown, line 22: Would it be better to generate the themes from path?

thornbill: :+1: good call

webpack.common.js
```diff
@@ -60,13 +71,14 @@ const config = {
...
+            hash: true,
+            chunks: [
+                'main.jellyfin',
+                'serviceworker'
+            ]
         }),
         new CopyPlugin({
             patterns: [
-                {
-                    from: 'themes/**/*.{css,jpg}'
```
viown, line 68: Looks like the themes can't load the background images from themes nor the jellyfin logo:

```
GET
http://localhost:8080/themes/blueradiance/47410544679fce8d5411.jpg
[HTTP/1.1 404 Not Found
```

The theme images aren't being copied, but the Jellyfin logo is being copied correctly however the reference to it is incorrect.

PR #6664, opened 2025-03-26
infinitescroll

gompa:
feat: Add Infinite Scroll Feature

- Modified `musicalbums.js` ,`musicartists.js`, `movies.js` and `tvshows.js` to support infinite scrolling:
  - Implemented logic to load more items when scrolling near the bottom of the page.
  - Updated the HTML structure to append new items instead of replacing the existing content.
- Added a new method `enableInfiniteScroll` in `userSettings.js` to get or set the infinite scroll setting.
- Added a new checkbox in display settings to enable/disable infinite scroll.
- Updated `displaySettings.js` to handle the loading and saving of the infinite scroll setting.

Note: ESLint is failing on the use of `AbortController` with the following message:
  error: AbortController is not supported in iOS Safari 10.0-10.2, Chrome 27

I am not sure how to handle the ESLint error and am wondering if this could be handled by a polyfill.

there is an issue when switching between album and artist views where the scroll position gets transitioned incorrectly. I still need to debug this issue.

If you have any thoughts or suggestions, I'd really appreciate input.


thornbill:
Is this not the same as setting the "Library page size" to 0?

xtreemze:
I think this is a very thoughful and useful feature! Nice alternative to pagination. Setting page size to 0 tends to suffer quite a bit with performance.

Not sure why this should be exclusive to Music only. I think any media library would benefit from this option.

southofrubick:
I think the compatibility warnings from eslint should be pretty safe to disregard, as the versions are for safari and chrome are 8, and 12 years old. Other then that, I would love to see this get merged at some point. Great job!

viown:
> I am not sure how to handle the ESLint error and am wondering if this could be handled by a polyfill.

There should already be a polyfill for it iirc.

> No, I think that loads the entire database at once. This loads data in chunks based on the pagination value as you scroll near the bottom. It seems to work a lot faster on my slow media server.

Since we already state in the helper text for library size that setting to 0 disables paging, it might make sense to apply this special behaviour when it's 0 instead of adding a new setting -- would boost the performance for a lot of users who currently use 0.

gompa:
> Since we already state in the helper text for library size that setting to 0 disables paging, it might make sense to apply this special behaviour when it's 0 instead of adding a new setting -- would boost the performance for a lot of users who currently use 0.

that might be a good idea when it is tested some more, now its easy to find issues related to the infinite scrolling by turning the flag off
I can only test the functionality on some browsers with a 4k and 1080p screen and my phone so I'm afraid I missed cases where it doesn't work
but if you prefer i can change it to be enabled by default if  `userSettings.libraryPageSize() == 0 `

boatboatboatboat:
What should be done further to get this merged? The SQ duplicate code warning can probably be ignored due to the nature of the controllers, and there should already be a [polyfill for the AbortController](https://github.com/jellyfin/jellyfin-web/blob/bbced2590b334056ce97c8ef913f7efd8a9a6a55/src/lib/legacy/index.ts#L9).

maduranma:
This should be not only merged but also I think it should be the default option, like any other clients (Swiftfin i.e.)

Issue #6591, opened 2025-03-01
Skip Intro Button Formatting


PeachesMLG:
### This issue respects the following points:

- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-web/issues?q=is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

The formatting on the skip intro button could be optimised

![Image](https://github.com/user-attachments/assets/25f7d1a3-b6fc-4f61-ba85-94400d356b9c)

In my opinion it should be closer to the bottom right of the screen

![Image](https://github.com/user-attachments/assets/1bd83ef6-df4d-4048-bf21-70fc5f896a7c)

### Reproduction Steps

N/A

### Expected/Actual behaviour

N/A

### Server version

10.10.6

### Web version

10.10.6

### Build version

10.10.6

### Platform

Linux

### Browser

Firefox

### Additional information

I am currently using the IntroSkipper 1.10.10.15 plugin, 

https://manifest.intro-skipper.org/manifest.json

However I believe this only analysises media segments and leaves displaying the Skip Intro button to Jellyfin

854562:
While the button's design could be improved, I don't think it should be tucked away in a corner as drastically as you suggest. That would make it easier to miss or misclick. I feel like it should at least appear over the content rather than beside or below it. Since the button disappears after X seconds, it's not overly intrusive anyway.

PeachesMLG:
Probabbly could do with @erikbucik giving some suggestions

But if you look at most other streaming platforms, they have similar designs of it being in the bottom right

![Image](https://github.com/user-attachments/assets/8f529641-9cca-4e33-8e26-8eb9a8cb4fd8)

![Image](https://github.com/user-attachments/assets/b4600f7f-084f-4ea8-837a-21697d5d431b)

![Image](https://github.com/user-attachments/assets/786fc8ce-be47-4cb2-8487-82626e388c07)

viown:
The design you see there was actually a temporary placeholder that I couldn't get around to updating in time for the 10.10 release 😅 

I've updated the design and positioning to match JF's UI more:

![Image](https://github.com/user-attachments/assets/fb636962-3327-4288-adaa-585b0a6a1a5c)
[...]

PR #6670, opened 2025-03-27
Fix skip button positioning & design

viown:
[Screenshots here](https://github.com/jellyfin/jellyfin-web/issues/6591#issuecomment-2707079148)

**Changes**
Fixes the positioning and styling of the skip button

**Issues**
[](https://github.com/jellyfin/jellyfin-web/issues/6591)

### DIFF
--- a/src/components/playback/skipbutton.scss
+++ b/src/components/playback/skipbutton.scss
@@ -2,25 +2,19 @@
     display: flex;
     align-items: center;
     position: fixed;
-    bottom: 18%;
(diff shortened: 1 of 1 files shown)


dmitrylyzo:
I added `skip-button-container` in #6510 - will conflict. I think we can fix the design in the release branch as well.

thornbill:
I have backported the changes. There is a chance we will still do another 10.10 release but nothing for certain.

PR #6672, opened 2025-03-27
Add public user listing option to networking settings

mateuszdrab:
**Changes**
Add public user listing option to networking settings
image

This is to provide web client control of the feature added in https://github.com/jellyfin/jellyfin/pull/13776


src/apps/dashboard/controllers/networking.js
```diff
@@ -55,6 +55,8 @@ function onSubmit(e) {
                 config.RequireHttps = form.querySelector('#chkRequireHttps').checked;
                 config.BaseUrl = form.querySelector('#txtBaseUrl').value;
                 config.EnableRemoteAccess = form.querySelector('#chkRemoteAccess').checked;
+                config.PublicUserListing = form.querySelector('#selectPublicUserListing').value === 'always' || form.querySelector('#selectPublicUserListing').value === 'local';
+                config.PublicUserListingLocalOnly = form.querySelector('#selectPublicUserListing').value === 'local';
```
dmitrylyzo, line 59: Maybe take the value once?

mateuszdrab: As in retrieve `form.querySelector('#selectPublicUserListing').value` into a variable and compare against that?

dmitrylyzo: Yes

src/apps/dashboard/controllers/networking.js
```diff
@@ -55,6 +57,8 @@ function onSubmit(e) {
                 config.RequireHttps = form.querySelector('#chkRequireHttps').checked;
                 config.BaseUrl = form.querySelector('#txtBaseUrl').value;
                 config.EnableRemoteAccess = form.querySelector('#chkRemoteAccess').checked;
+                config.PublicUserListing = publicUserListing === 'always' || publicUserListing === 'local';
+                config.PublicUserListingLocalOnly = form.querySelector('#selectPublicUserListing').value === 'local';
```
dmitrylyzo, line 61:
```suggestion
                config.PublicUserListingLocalOnly = publicUserListing === 'local';
```

src/apps/dashboard/controllers/networking.js
```diff
@@ -142,6 +146,13 @@ export default function (view) {
         page.querySelector('#chkEnableIP6').checked = config.EnableIPv6;
         page.querySelector('#chkEnableIP4').checked = config.EnableIPv4;
         page.querySelector('#txtPublishedServer').value = (config.PublishedServerUriBySubnet || []).join(', ');
+        if (config.PublicUserListingLocalOnly && config.PublicUserListing) {
+            page.querySelector('#selectPublicUserListing').value = 'local';
+        } else if (config.PublicUserListing) {
+            page.querySelector('#selectPublicUserListing').value = 'always';
+        } else {
+            page.querySelector('#selectPublicUserListing').value = 'never';
+        }
```
dmitrylyzo, line 155: For maintainability: fewer places with the identifier.
```suggestion

        const selectPublicUserListing = page.querySelector('#selectPublicUserListing');

        if (config.PublicUserListingLocalOnly && config.PublicUserListing) {
            selectPublicUserListing.value = 'local';
        } else if (config.PublicUserListing) {
            selectPublicUserListing.value = 'always';
        } else {
            selectPublicUserListing.value = 'never';
        }
```

mateuszdrab: that should be done, thanks for the feedback!

src/apps/dashboard/controllers/networking.js
```diff
@@ -46,6 +46,8 @@ function onSubmit(e) {
                     return s.length > 0;
                 });
 
+                const publicUserListing = form.querySelector('#selectPublicUserListing');
```
dmitrylyzo, line 49: I think `.value` was fine.
Otherwise, it would be better to rename `publicUserListing` to `selectPublicUserListing` to match the other function. Or vice versa.

src/apps/dashboard/controllers/networking.js
```diff
@@ -142,6 +146,14 @@ export default function (view) {
         page.querySelector('#chkEnableIP6').checked = config.EnableIPv6;
         page.querySelector('#chkEnableIP4').checked = config.EnableIPv4;
         page.querySelector('#txtPublishedServer').value = (config.PublishedServerUriBySubnet || []).join(', ');
+        const selectPublicUserListing = page.querySelector('#selectPublicUserListing');
```
dmitrylyzo, line 149: I would separate the assignment with blank lines (as I did in my previous suggestion).
```suggestion

        const selectPublicUserListing = page.querySelector('#selectPublicUserListing');

```

src/apps/dashboard/controllers/networking.js
```diff
@@ -142,6 +146,14 @@ export default function (view) {
...
+        if (config.PublicUserListingLocalOnly && config.PublicUserListing) {
+            selectPublicUserListing.value = 'local';
+        } else if (config.PublicUserListing) {
+            selectPublicUserListing.value = 'always';
+        } else {
+            selectPublicUserListing.value = 'never';
+        }
```
dmitrylyzo, line 156: I would put a blank line here as well to separate the `if/else` block (activity 1) from hiding the loading (activity 2).
```suggestion
        }

```
Iirc, we have a rule to separate blocks in general, but it doesn't seem to be checked by linter. 🤷‍♂️

src/apps/dashboard/controllers/networking.js
mateuszdrab: Great, renamed the variable and extra spacing added

thatguy-za:
Hi @mateuszdrab was this ever added to Jellyfin?

mateuszdrab:
> Hi @mateuszdrab was this ever added to Jellyfin?

Unfortunately, it was not.

PR #6673, opened 2025-03-28
Implement Theme Song Volume Control

### LINKED ISSUE #3086: theme song / theme videos volume adjustments
Currently when a theme song is playing in the background you have no way to control the volume it is playing at. When using Syncplay a Playback Control Bar appears and you can pause, play, skip and adjust the volume as described in #2918 and #3059.

I think it would be worth having an option to always display the Playback Control Bar when the theme song / theme video is playing especially since https://github.com/jellyfin/jellyfin/issues/537 is not yet implemented and it is not uncommon for different theme songs to have completely different volumes. Alternatively a less noticeable / intrusive volume control such as the one during video playback when using the arrow keys (just display when adjustments are made) could be considered.

I am hoping to get some different opinions on this topic.

NathanGrenier:
**About**

Implement basic input slider in "Display" settings menu to set theme song volume level during playback.

**Changes**

- Add UI elements to change theme song volume level
- Add settings to store theme song volume level
- TODO: Set the theme song playback audio level using the settings value (without affecting the media playback volume in the process)

**Issues**

Fixes #3086

**Details**
Currently, I've only setup the ui and backend logic for setting and getting the theme song volume level.

I'm able to set the theme song's volume level quite easily in `components/themeMediaPlayer.js`, but doing that has the unintended side effect of also setting the volume level of all media playback. This is because there's only 1 shared media player for the application.

I tried a workaround involving storing the previous audio level then restoring it once the theme song playback ended, but couldn't  get it to work perfectly without any issues / minor bugs.

I also wasn't able to test the changes made to `src/apps/experimental/features/preferences/components/LibraryPreferences.tsx` because I'm not sure where to access that UI.


AfterAllTheseYearsStillKnowNothing:
Was this ever fully implemented.  It says it passed QC in deployment, but I don't see it in the latest build

NathanGrenier:
> Was this ever fully implemented. It says it passed QC in deployment, but I don't see it in the latest build

I'm pretty sure it never was.

PR #6674, opened 2025-03-28
Implement Theme Song Volume Control

### LINKED ISSUE #3086: theme song / theme videos volume adjustments

NathanGrenier:
**About**

Implement basic input slider in "Display" settings menu to set theme song volume level during playback.

**Changes**

- Add UI elements to change theme song volume level
- Add settings to store theme song volume level
- TODO: Set the theme song playback audio level using the settings value (without affecting the media playback volume in the process)

**Issues**

Fixes #3086

**Details**
Currently, I've only setup the ui and backend logic for setting and getting the theme song volume level.

I'm able to set the theme song's volume level quite easily in `components/themeMediaPlayer.js`, but doing that has the unintended side effect of also setting the volume level of all media playback. This is because there's only 1 shared media player for the application.

I tried a workaround involving storing the previous audio level then restoring it once the theme song playback ended, but couldn't  get it to work perfectly without any issues / minor bugs.

I also wasn't able to test the changes made to `src/apps/experimental/features/preferences/components/LibraryPreferences.tsx` because I'm not sure where to access that UI.


src/scripts/settings/userSettings.js
```diff
@@ -230,6 +230,19 @@ export class UserSettings {
...
+     * Get or set 'Theme Songs' volume level.
+     * @param {number|undefined} [val] - Volume level to set.
+     * @return {number} Current volume level.
+     */
+    themeSongsVolumeLevel(val) {
+        if (val !== undefined) {
+            return this.set('themeSongsVolumeLevel', val.toString());
+        }
+
+        return parseInt(this.get('themeSongsVolumeLevel') || '100', 10);
```
dmitrylyzo, line 243: IMO, it would be better to make the default volume lower.
```suggestion
        return parseInt(this.get('themeSongsVolumeLevel') || '50', 10);
```

src/apps/experimental/features/preferences/types/displaySettingsValues.ts
```diff
@@ -9,6 +9,7 @@ export interface DisplaySettingsValues {
     enableItemDetailsBanner: boolean;
     enableLibraryBackdrops: boolean;
     enableLibraryThemeSongs: boolean;
+    libraryThemeSongsVolumeLevel: number;
```
dmitrylyzo, line 12: I think it was supposed to be in alphabetical order.

src/apps/experimental/features/preferences/hooks/useDisplaySettings.ts
```diff
@@ -87,6 +87,7 @@ async function loadDisplaySettings({
         enableItemDetailsBanner: Boolean(settings.detailsBanner()),
         enableLibraryBackdrops: Boolean(settings.enableBackdrops()),
         enableLibraryThemeSongs: Boolean(settings.enableThemeSongs()),
+        libraryThemeSongsVolumeLevel: settings.themeSongsVolumeLevel(),
```
dmitrylyzo, line 90: I think it was supposed to be in alphabetical order.

src/apps/experimental/routes/user/display/index.tsx
```diff
@@ -40,6 +40,19 @@ export default function UserDisplayPreferences() {
...
+        const value = Array.isArray(newValue) ? newValue[0] : newValue;
+
+        if (values?.[fieldName] !== value) {
+            updateField({
+                name: fieldName,
+                value: value.toString()
+            });
+        }
+    }, [updateField, values]);
+
```
dmitrylyzo, line 55: Isn't `handleFieldChange` (`onChange` callback) enough?

dmitrylyzo:
> This is because there's only 1 shared media player for the application.

The media element is [not destroyed](https://github.com/jellyfin/jellyfin-web/blob/3df2a1fe53f1c5ed3dc058f4c56ce2bd76b3d868/src/plugins/htmlAudioPlayer/plugin.js#L258-L261). If this is not on purpose, we should destroy it completely, as [in the video player](https://github.com/jellyfin/jellyfin-web/blob/3df2a1fe53f1c5ed3dc058f4c56ce2bd76b3d868/src/plugins/htmlVideoPlayer/plugin.js#L855-L873). But we can leave it as is for now.

> I tried a workaround involving storing the previous audio level then restoring it once the theme song playback ended, but couldn't get it to work perfectly without any issues / minor bugs.

I think we can introduce a "volume mode" (primary and secondary) and make players use it to get/set the appropriate value.

https://github.com/jellyfin/jellyfin-web/blob/3df2a1fe53f1c5ed3dc058f4c56ce2bd76b3d868/src/plugins/htmlAudioPlayer/plugin.js#L280-L283
https://github.com/jellyfin/jellyfin-web/blob/3df2a1fe53f1c5ed3dc058f4c56ce2bd76b3d868/src/plugins/htmlAudioPlayer/plugin.js#L325
https://github.com/jellyfin/jellyfin-web/blob/3df2a1fe53f1c5ed3dc058f4c56ce2bd76b3d868/src/components/htmlMediaHelper.js#L6-L14

We can use `fullscreen` from the `options` passed to `play` - for "theme players" is `false`.

We also need to move [this](https://github.com/jellyfin/jellyfin-web/blob/3df2a1fe53f1c5ed3dc058f4c56ce2bd76b3d868/src/plugins/htmlAudioPlayer/plugin.js#L280-L283) to [`play`](https://github.com/jellyfin/jellyfin-web/blob/3df2a1fe53f1c5ed3dc058f4c56ce2bd76b3d868/src/plugins/htmlAudioPlayer/plugin.js#L104-L106).

> I also wasn't able to test the changes made to `src/apps/experimental/features/preferences/components/LibraryPreferences.tsx` because I'm not sure where to access that UI.

`User settings / Display / Display mode` -> `Experimental`.

Issue #6522, opened 2025-02-13
LG C2 audio transcoding always plays as stereo


aedalzotto:
**Describe The Bug**
Unsupported audio tracks correctly trigger audio transcode. 
Although multichannel audio is transcoded correctly to multichannel AAC (or AC3, when selected), the TV passes through stereo PCM through eARC.
This behavior occurs both using the TV browser and the jellyfin-webos application.
This behavior occurs regardless of the maximum channels option being set to 5.1/7.1

**Steps To Reproduce**
1. Play an unsupported multichannel audio track (tested with DTS[-HD MA] and FLAC 5.1) or lower the playback quality to force a transcode due to bitrate
2. Verify using Playback Info that the client receives multichannel AAC (or multichannel AC3, when opted in client playback options)
3. Check the eARC receiver to state that it is, in fact, receiving stereo PCM

**Expected Behavior**
* If transcoding to multichannel AAC, the TV should play multichannel PCM through eARC. I have verified that HE-AAC and AAC-LC correctly play as multichannel PCM when not transcoding.
* If transcoding to multichannel AC3, the TV should pass through the codec through eARC. I have verified that AC3 correctly passes through eARC when not transcoding.

**Logs**
jellyfin-server log with DTS-HD MA to AAC 6-channel
```
[INF] /usr/lib/jellyfin-ffmpeg/ffmpeg -analyzeduration 200M -probesize 1G -ss 00:10:12.875 -readrate 10 -fflags +genpts -f matroska  -i file:"_.mkv" -map_metadata -1 -map_chapters -1 -threads 0 -map 0:0 -map 0:1 -map -0:s -codec:v:0 copy -bsf:v h264_mp4toannexb -start_at_zero -codec:a:0 aac -ac 6 -ab 640000 -copyts -avoid_negative_ts disabled -max_muxing_queue_size 2048 -f hls -max_delay 5000000 -hls_time 6 -hls_segment_type mpegts -start_number 102 -hls_segment_filename "_%d.ts" -hls_playlist_type vod -hls_list_size 0 -y "_.m3u8"
```

**System (please complete the following information):**
 - Platform: LG OLED C2 v23.20.56 (recent webOS 24 update). Both with Jellyfin v1.2.2 and using the TV browser.
 - Jellyfin Version: 10.10.5

**Additional Context**
Additional bug: jellyfin-webos supports Opus playback, but when opting to transcode unsupported tracks to Opus, it requests the server for AAC.

[...]

Issue #6533, opened 2025-02-18
LG webOS - Playback hangs when audio is transcoded to stereo


0xBADEAFFE:
### This issue respects the following points:

- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-web/issues?q=is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

If I set the maximum permitted audio channels to stereo in the playback settings on the TV and the sound has to be transcoded to stereo, playback hangs after a while. In this case, it is always at the same points.

Version info:
- Jellyfin 10.10.6ubu2404-ls53 docker image on unraid
- LG C2 with webOS TV-Version webOS24 / 9.2.1-2607, Dolby Vision 32.22
- Jellyfin for WebOS 1.2.2 from LG app store

### Reproduction Steps

1. go to Settings in the app and then Playback
2. set the maximum permitted audio channels to stereo and save the change
3. play a video
4. if the audio is transcoded to stereo, the entire playback hangs after a while

### Expected/Actual behaviour

Despite transcoding, playback should not hang. 

### Media info of the file

```shell
lib/jellyfin-ffmpeg/ffprobe "data/tvshows/Silo/Season 02/Silo.S02E01.GERMAN.1080p.h264.mkv"
ffprobe version 7.0.2-Jellyfin Copyright (c) 2007-2024 the FFmpeg developers
  built with gcc 13 (Ubuntu 13.3.0-6ubuntu2~24.04)
[...]

Issue #6484, opened 2025-01-27
LG G4 transcoding (DoVi, DTS, TrueHD)


yorah:
**Describe The Bug**
Using the webos jellyfin player which is a thin wrapper around jellyfin-web.

Bug 1: DoVi 8.1 file triggers transcoding on LG G4.
Bug 2: Disabling transcoding for this user allows playing this file DV content just fine with direct stream, but the two audio tracks which are DTS HDMA and Dolby TrueHD with Atmos get transcoded to AAC (DTS and TrueHD passthrough options are checked in the settings)

**Media Information**
Video
ID                                       : 1
Format                                   : HEVC
Format/Info                              : High Efficiency Video Coding
Format profile                           : Main 10@L5.1@High
HDR format                               : Dolby Vision, Version 1.0, Profile 8.1, dvhe.08.06, BL+RPU, no metadata compression, HDR10 compatible / SMPTE ST 2086, Version HDR10, HDR10 compatible
Codec ID                                 : V_MPEGH/ISO/HEVC
Duration                                 : 1 h 55 min
Bit rate                                 : 75.5 Mb/s
Width                                    : 3 840 pixels
Height                                   : 2 160 pixels
Display aspect ratio                     : 16:9
Frame rate mode                          : Constant
Frame rate                               : 23.976 (24000/1001) FPS
Color space                              : YUV
Chroma subsampling                       : 4:2:0 (Type 2)
Bit depth                                : 10 bits
Bits/(Pixel*Frame)                       : 0.380
Stream size                              : 60.7 GiB (79%)
Default                                  : Yes
Forced                                   : No
Color range                              : Limited
Color primaries                          : BT.2020
Transfer characteristics                 : PQ
Matrix coefficients                      : BT.2020 non-constant
Mastering display color primaries        : Display P3
Mastering display luminance              : min: 0.0001 cd/m2, max: 1000 cd/m2
Maximum Content Light Level              : 877 cd/m2
Maximum Frame-Average Light Level        : 180 cd/m2

Audio #1
ID                                       : 2
Format                                   : MLP FBA 16-ch
Format/Info                              : Meridian Lossless Packing FBA with 16-channel presentation
Commercial name                          : Dolby TrueHD with Dolby Atmos
Codec ID                                 : A_TRUEHD
Duration                                 : 1 h 55 min
Bit rate mode                            : Variable
Bit rate                                 : 4 406 kb/s
Maximum bit rate                         : 7 623 kb/s
[...]

PR #6530, opened 2025-02-17
Add option to limit hls segment length for webOS

patrickd77-eng:
Changes

The change aims to allow users to toggle a suggested workaround/fix for an outstanding bug with DV playback via the WebOS client. The new setting lets users limit segment length to 1 instead of the default (I believe 6?). This PR was created due to suggestions made in another [PR](https://github.com/jellyfin/jellyfin/pull/13561#issuecomment-2660701231). ~~I have provided some translations, all of them (except en-GB and en-US) were generated via GitHub CoPilot. Feel free to reject those, as I know there is a good system for community translations in place already.~~

I have tested the change against my local server with the client running against VS Code. The setting is false by default and is correctly toggled on/off when I changed it and saved it, so the state is correctly remembered. I can also see it being correctly set in the transcoding profiles, and they are undefined if the user leaves it disabled, so the server will ignore that property.

A more technical explanation is seen here: https://github.com/jellyfin/jellyfin-webos/issues/248#issuecomment-2622502439

Issues

https://github.com/jellyfin/jellyfin-webos/issues/248

PR #6678, opened 2025-03-29
Enable hls.js with WebOS

timitt:
**Changes**
This patch will put hls.js in use when using LG WebOS 4 or newer. Native HLS in WebOS seems to be broken in many ways. Worst way is that it breaks multichannel audio and everything plays as stereo. External amplifiers that are connected to TV through HDMI ARC or Toslink fail to show Dolby Digital marking no matter what channel count is forced to be used when native HLS is in use.

**Note that we need 1.6.0 or newer version of hls.js** to use this patch if we want Dolby Vision 8 to work.
Hls.js 1.6.0 now supports supplemental codecs and it also has much better support for Dolby Vision files generally.

Also note that users should enable use of fMP4 when this patch is enabled unlike with native hls.

Possible negatives with this patch is the fact that we need to transcode AC-3 and EAC-3 when HLS is needed as WebOS MSE does not support playing videos with those audio codecs.
It also kills Dolby Atmos, but I am not sure if it worked correctly with native hls either (at least it did not work when using HDMI ARC).
There is possible solution for that too here: https://forum.jellyfin.org/t-webos-with-hls-js

**Issues**

Fixes https://github.com/jellyfin/jellyfin-webos/issues/275
Fixes https://github.com/jellyfin/jellyfin-web/issues/6522

Probably fixes https://github.com/jellyfin/jellyfin-webos/issues/248
Probably fixes https://github.com/jellyfin/jellyfin-webos/issues/285
Probably fixes https://github.com/jellyfin/jellyfin-webos/issues/284

Relates https://github.com/jellyfin/jellyfin-webos/issues/278
Relates https://github.com/jellyfin/jellyfin-webos/issues/274
Relates https://github.com/jellyfin/jellyfin-webos/issues/262
Relates https://github.com/jellyfin/jellyfin-web/issues/6533
Relates https://github.com/jellyfin/jellyfin-web/issues/6484


timitt:
> > Possible negatives with this patch is the fact that we need to transcode AC-3 and EAC-3 when HLS is needed as WebOS MSE does not support playing videos with those audio codecs.
> 
> This is probably a bug in either ffmpeg or hls.js. Looking at the chromium sources webOS MSE does support AAC, MP3, Vorbis, AC3, EAC3, OPUS, FLAC, AC4, DTS(E/X) (only on select models). Supported video codecs are H264/5 VP8/9.

Problem is WebOS. If you use isTypeSupported for audio/mp4; codecs="whatever" then you get true in those formats. If you use isTypeSupported for video/mp4; codecs="whatever-video,whatever-audio" then supported audio codecs are much more limited. For that reason MSE does not play videos with AC3, but it does play plain audio with AC3.

I have talked about this issue on mentioned forum post and also on first "fixes" ticket I list.

Also the solution would be to separate audio and video to their own mp4 streams (which hls allows). That would solve also other problems.

Fixing that problem is just not required to make this pull request a must have for every WebOS user that have multichannel audio system at home.

sundermann:
> Problem is WebOS. If you use isTypeSupported for audio/mp4; codecs="whatever" then you get true in those formats. If you use isTypeSupported for video/mp4; codecs="whatever-video,whatever-audio" then supported audio codecs are much more limited. For that reason MSE does not play videos with AC3, but it does play plain audio with AC3.

Maybe it's just the check that's broken? Did you try removing isTypeSupported from hls.js?

timitt:
> > Problem is WebOS. If you use isTypeSupported for audio/mp4; codecs="whatever" then you get true in those formats. If you use isTypeSupported for video/mp4; codecs="whatever-video,whatever-audio" then supported audio codecs are much more limited. For that reason MSE does not play videos with AC3, but it does play plain audio with AC3.
> 
> Maybe it's just the check that's broken? Did you try removing isTypeSupported from hls.js?

Yes, I have tried to ignore check, but actual playback request using MSE does not work with eac3 either.

Skin80:
Is there any update to this being implemented?

timitt:
> Is there any update to this being implemented?

I don't know what I can do to make this reviewed.

There is not much code to there, but quite a lot of testing to make sure it is good patch to offer. I have also explained why it is made like that and so on (here and in that forum thread).

Of course I am not able to test other WebOS televisions (and WebOS versions) than my own so there is that.

rschneider1509:
i tried it by manualy compiling it and put it in my server. indeed, the sound problem was fixed, but then i had issues with dolby vision :( 
can someone confirm?

aedalzotto:
This PR is helpful due to the numerous bugs in LG's HLS. However, killing Dolby Atmos makes it unfit for merging, in my opinion.

An alternative is to enable hls.js only when content is delivered through HLS (e.g., when Prefer fMPG-HLS Media Container is selected).  This would keep the default behavior of using LG's implementation of ts.

Recently, on [webos forums](https://forum.webostv.developer.lge.com/t/stereo-only-for-dolby-vision-video-with-multi-channel-audio-in-hls-streams/23069/24), we learned that LG is reviewing the fix for the bug that originated the need for this PR, but it will take some time to arrive through an update.

timitt:
> This PR is helpful due to the numerous bugs in LG's HLS. However, killing Dolby Atmos makes it unfit for merging, in my opinion.
> 
> An alternative is to enable hls.js only when content is delivered through HLS (e.g., when Prefer fMPG-HLS Media Container is selected). This would keep the default behavior of using LG's implementation of ts.
> 
> Recently, on [webos forums](https://forum.webostv.developer.lge.com/t/stereo-only-for-dolby-vision-video-with-multi-channel-audio-in-hls-streams/23069/24), we learned that LG is reviewing the fix for the bug that originated the need for this PR, but it will take some time to arrive through an update.

Obviously Dolby Atmos works like before if HLS is not needed (if video is in container that can be played directly). I don't see we lose anything when using this patch. I wrote in my forum post (linked in my initial message) about possible way to make Dolby Atmos work too, but I do not have motivation to try to fix that since this pull request (and my forum post) is ignored even when fix is very much needed.

If LG actually fixes this problem then this patch is not needed to make multi channel audio to work, but other problems still remain. For example that pause issue and fast forward issue. There was some sort of patch for those already implemented, but apparently it does not fully fix those problems.

Skin80:
> This PR is helpful due to the numerous bugs in LG's HLS. However, killing Dolby Atmos makes it unfit for merging, in my opinion.
> 
> An alternative is to enable hls.js only when content is delivered through HLS (e.g., when Prefer fMPG-HLS Media Container is selected). This would keep the default behavior of using LG's implementation of ts.
> 
> Recently, on [webos forums](https://forum.webostv.developer.lge.com/t/stereo-only-for-dolby-vision-video-with-multi-channel-audio-in-hls-streams/23069/24), we learned that LG is reviewing the fix for the bug that originated the need for this PR, but it will take some time to arrive through an update.

How is it "killing Dolby Atmos"? Dolby Atmos doesn't work with native HLS on LG WebOS, so this fix can't be "killing" something that is already "dead" :-p

Also, timitt has provided a solution to the Atmos issue (separating video and audio streams). So, I would suggest the Jellyfin developers try to implement that change instead of leaving it up to LG to (maybe) fix it in a future update.

Overall, I would argue that we should move away from the default behaviour of LG, given how broken their software is. Given that the issue is at least a year old (affecting WebOS 23 and 24), would it not be prudent to move away from utilising native software when Jellyfin can utilise its own version?

Case in point, LG's implementation of hls.js broke Jellyfin for all LG TVs. It would be best practise to use embedded components in the Jellyfin-web client code, instead of relying on the native components provided by TV manufacturers.

Clearly the likes of LG, Samsung, Sony, etc. can break Jellyfin during any of their numerous firmware updates, so the less Jellyfin relies on native software, the better.

Skin80:
> > This PR is helpful due to the numerous bugs in LG's HLS. However, killing Dolby Atmos makes it unfit for merging, in my opinion.
> > An alternative is to enable hls.js only when content is delivered through HLS (e.g., when Prefer fMPG-HLS Media Container is selected). This would keep the default behavior of using LG's implementation of ts.
> > Recently, on [webos forums](https://forum.webostv.developer.lge.com/t/stereo-only-for-dolby-vision-video-with-multi-channel-audio-in-hls-streams/23069/24), we learned that LG is reviewing the fix for the bug that originated the need for this PR, but it will take some time to arrive through an update.
> 
> Obviously Dolby Atmos works like before if HLS is not needed (if video is in container that can be played directly). I don't see we lose anything when using this patch. I wrote in my forum post (linked in my initial message) about possible way to make Dolby Atmos work too, but I do not have motivation to try to fix that since this pull request (and my forum post) is ignored even when fix is very much needed.
> 
> If LG actually fixes this problem then this patch is not needed to make multi channel audio to work, but other problems still remain. For example that pause issue and fast forward issue. There was some sort of patch for those already implemented, but apparently it does not fully fix those problems.

I feel your pain! The number of open pull requests and issues on Github, many dating back months (and even years), is frustrating.

The devs really need to set up a testing repository so that fixes can be tested and signed off quicker.

MrHumanRebel:
Would be nice to finally play HLS on my very expensive LG OLED...

thornbill:
> > > This PR is helpful due to the numerous bugs in LG's HLS. However, killing Dolby Atmos makes it unfit for merging, in my opinion.
> > > An alternative is to enable hls.js only when content is delivered through HLS (e.g., when Prefer fMPG-HLS Media Container is selected). This would keep the default behavior of using LG's implementation of ts.
> > > Recently, on [webos forums](https://forum.webostv.developer.lge.com/t/stereo-only-for-dolby-vision-video-with-multi-channel-audio-in-hls-streams/23069/24), we learned that LG is reviewing the fix for the bug that originated the need for this PR, but it will take some time to arrive through an update.
> > 
> > 
> > Obviously Dolby Atmos works like before if HLS is not needed (if video is in container that can be played directly). I don't see we lose anything when using this patch. I wrote in my forum post (linked in my initial message) about possible way to make Dolby Atmos work too, but I do not have motivation to try to fix that since this pull request (and my forum post) is ignored even when fix is very much needed.
> > If LG actually fixes this problem then this patch is not needed to make multi channel audio to work, but other problems still remain. For example that pause issue and fast forward issue. There was some sort of patch for those already implemented, but apparently it does not fully fix those problems.
> 
> I feel your pain! The number of open pull requests and issues on Github, many dating back months (and even years), is frustrating.

Trust me, I don't think anyone is more frustrated by the current state of reviews in this repo than me... this project with hundreds of thousands of users has around 2 regular contributors and the vast majority of reviews are done by just me. I am a volunteer the same as every other contributor with a day job, family, other life responsibilities, and other projects that I maintain within the Jellyfin ecosystem. I also struggle with getting reviews on my own PRs since I do the majority of reviews and obviously cannot review my own PRs.

I am more than open to any suggestions on improving the process.

PRs dealing with specific platforms can also be challenging as I have no webOS devices to test or verify this change. In the past, I relied pretty heavily on another contributor that tested TV specific changes, but they seem to be less active of late. It does look like a few people have tested and verified this change besides the author now, so I do agree that it can be considered tested, but generally I don't know how to communicate this to webOS users that a PR requires testing beyond using GitHub labels.

> The devs really need to set up a testing repository so that fixes can be tested and signed off quicker.

Again I'm welcome to any specific ideas on this front. We do have preview builds published to Cloudflare pages for every PR that is opened and of course anyone is able to pull the changes themselves, but if you have additional ideas please share them or join the matrix/discord chat to discuss it!

Thanks for your patience.

thornbill (approved):

timitt:
> Trust me, I don't think anyone is more frustrated by the current state of reviews in this repo than me... this project with hundreds of thousands of users has around 2 regular contributors and the vast majority of reviews are done by just me. I am a volunteer the same as every other contributor with a day job, family, other life responsibilities, and other projects that I maintain within the Jellyfin ecosystem. I also struggle with getting reviews on my own PRs since I do the majority of reviews and obviously cannot review my own PRs.

Your (and other volunteers) work is much appreciated. Especially as I know that I would not be able to put myself to your spot and be active contributor.

> I am more than open to any suggestions on improving the process.

For me the issue seems to be that there are several different offered ways to communicate about possible fixes or issues and the ones I have used does seem to be not read (or at least answered by anyone that does any development). Of course if there is mostly you then there is obviously too much work for one (or two) person.

I did not try Matrix as I am on different timezone (Finland) but forums does not seem to have devs talking too much. Github discussions seems to be quite quiet too. So only way seemed to be to just do pull request.
I joined through Discord (as I use that with my friends) now.

> PRs dealing with specific platforms can also be challenging as I have no webOS devices to test or verify this change. In the past, I relied pretty heavily on another contributor that tested TV specific changes, but they seem to be less active of late. It does look like a few people have tested and verified this change besides the author now, so I do agree that it can be considered tested, but generally I don't know how to communicate this to webOS users that a PR requires testing beyond using GitHub labels.

I should probably have made builds to my repository with patch to make testing easier, but then again I myself find precompiled stuff from random people (like I am to all of you) bit dangerous. Still that would probably added more testers especially if I would have guessed it could help.

> > The devs really need to set up a testing repository so that fixes can be tested and signed off quicker.
> 
> Again I'm welcome to any specific ideas on this front. We do have preview builds published to Cloudflare pages for every PR that is opened and of course anyone is able to pull the changes themselves, but if you have additional ideas please share them or join the matrix/discord chat to discuss it!
> 
> Thanks for your patience.

Thank you for merging. I believe that we have many more happy users now. I also hope that all those "probably fixes" tickets got fixed by this pull request. I could not confirm them even though they do seem to be ones that got fixed too.

src/components/htmlMediaHelper.js
```diff
@@ -48,6 +48,12 @@ export function enableHlsJsPlayer(runTimeTicks, mediaType) {
         return false;
     }
 
+    // Native HLS support in WebOS only plays stereo sound. hls.js works better, but works only on WebOS 4 or newer.
+    // Using hls.js also seems to fix fast forward issues that native HLS has.
+    if (browser.web0sVersion >= 4) {
```
IDisposable, line 53: Shouldn't this be `webOsVersion` (e.g. a capital O, not a Zero)?

sundermann: No, the variable name is correct like this

IDisposable: Is this some jellyfin infill, since the only GitHub search hits are in these in these (and clone) repos...

sundermann: Maybe you've been searching wrong?

https://github.com/jellyfin/jellyfin-web/blob/855276e8afe0ef42d4754918d9e8c6d647ff9f65/src/scripts/browser.js#L296

timitt: I have no idea why it is written with zero, but it is certainly correct.

dmitrylyzo: This is just because of the UserAgent string:
https://webostv.developer.lge.com/develop/specifications/web-api-and-web-engine
> Mozilla/5.0 (Web0S; Linux/SmartTV) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/68.0.3440.106 Safari/537.36 WebAppManager

IDisposable: > > Is this some jellyfin infill, since the only GitHub search hits are in these in these (and clone) repos...
>
> Maybe you've been searching wrong?
> 
> https://github.com/jellyfin/jellyfin-web/blob/855276e8afe0ef42d4754918d9e8c6d647ff9f65/src/scripts/browser.js#L296

That's in a JellyFin repo...

sundermann: Not sure if I understand correctly. jellyfin-webos is just a slim wrapper around jellyfin-web so it's correct that this code is in jellyfin-web.

IDisposable: > Not sure if I understand correctly. jellyfin-webos is just a slim wrapper around jellyfin-web so it's correct that this code is in jellyfin-web.

It's fine, I was previously unaware that the Web0S was driven by the Agent string (per [@dmitrylyzo 's comment](https://github.com/jellyfin/jellyfin-web/pull/6678#discussion_r2175518358) and thus asked a question about if the spelling was intentional. I did a Google and GitHub search to see if that spelling was used elsewhere and found that it was unique to the jellyfin-web repo (and the forks) and it was only here and being that it was `browser.web0sVersion` , I was concerned that no browser actually implemented that.

I didn't check to see that it was actually a infilled method declared in brower.js and filled into the global `browser` scope because I'm not in the habit of adding things to the global `browser` scope except for polyfills... and since this was not a spec-mentioned thing, I was surprised.

I am **very sorry** for wasting so much time just asking a question... I should have dived into the code first. Mea culpa.

CapivaraTupiniquim:
> cíficas sobre este

I have an LG OLED C4 TV here and I'm willing to test it if you need it! I really like Jellyfiin's project

tuzidaguai:
@timitt

My TV is a CX (webOS 5.6.1), but using your hls.js causes Jellyfin 10.11.0-(rc3-rc8) to transcode EAC3 audio to AAC when playing DV P5 EAC3 (MKV) videos(direct stream mode), 10.11.(rc1-rc2) can play these videos using remuxing mode. I modified the two files 'src/components/htmlMediaHelper.js' and 'src/scripts/browserDeviceProfile.js', changing 'if (browser.web0sVersion >= 4)' to '>=6' and repackaged. Does this effectively disable your new hls.js feature while still allowing me to use all the new features of rc8? Is this the correct approach?
thank you.

damienl560:
> @timitt
> 
> My TV is a CX (webOS 5.6.1), but using your hls.js causes Jellyfin 10.11.0-(rc3-rc8) to transcode EAC3 audio to AAC when playing DV P5 EAC3 (MKV) videos(direct stream mode), 10.11.(rc1-rc2) can play these videos using remuxing mode. I modified the two files 'src/components/htmlMediaHelper.js' and 'src/scripts/browserDeviceProfile.js', changing 'if (browser.web0sVersion >= 4)' to '>=6' and repackaged. Does this effectively disable your new hls.js feature while still allowing me to use all the new features of rc8? Is this the correct approach? thank you.

@timitt 

After updating to 10.11.0, dolby vision is now broken on both my C1 (03.53.45) and my B4 (23.20.28). Whereas the previous version was properly remuxing mkv files with DV profile 8 into a format compatible with my TVs, it now falls back to HDR10. This happens whether fMP4 is enabled or not. Additionally, it unnecessarily degrades the audio by converting (E)AC3 to AAC, even though AC3 tracks also play flawlessly on both TVs with the remuxed DV profile 8 video (as tuzidaguai also pointed out).

I suspect that the bugs that this change was supposed to fix is not common to all LG WebOS TVs but only to some of them. Unfortunately, this makes Jellyfin unusable for me. 

When I remove this change and replace the files manually, it works again as expected.

I believe we should use the same approach as #6530 and make this a toggleable option.

timitt:
I am sorry to hear that. I think you should make a new bug report about that. Making it toggleable is likely a good idea and probably an easy thing to do for some people. If it were easy to me I would have probably made it there originally.

On my LG C2 this patch was pretty much only way to make it work. I wrote a long solo conversation on Jellyfin forums about this too.

I primarily blame LG for this as I dont understand how it is possible to make it work so differently wrong on every version update.
This patch seemed to be really important for at least C2, C3 generations (including G-models). Now there is a new update that may make this patch less relevant but it is not yet available for my C2. It seems to be quite impossible to make it work well on all WebOS versions unless someone had all of them to test.

sundermann:
@damienl560 Have you checked that fMP4 is enabled using the logs? I've seen quite often that after selecting fMP4 in the options people forget to scroll down to the bottom and press the save button. It might also be necessary to restart the app afterwards. I feel like fMP4 should be enabled by default on webOS >= 4 to circumvent this issue.

PR #6686, opened 2025-03-31
Implement playback inactivity timeout

simonerlic:
Implements an "Are you still there?" popup after user-defined criterion are met.
User may change the behaviour in playback settings. Users can set the prompt to appear after a set number of episodes with no ineraction (e.g. mouse movement), or after a set amount of time with no interaction.

**Changes**

- Created a `playTimeout` plugin
- Added user settings for adjusting feature parameters
  - Added controls to `playbackSettings.js`, `playbackSettings.template.html`, and `userSettings.js`
- Added a bind to `playmanager.js`
- Added localization for en-US and en-GB

**Issues**

Work based off of conversations on [feature request 814](https://features.jellyfin.org/posts/814/are-you-still-watching-prompt)
Planned work discussed on [GitHub discussions](https://github.com/jellyfin/jellyfin-web/discussions/6603)


jhparker88:
I think you should squash your commit as it contains an unrelated commit history.

> Added localization for en-US and en-GB

Translations except for en-US are provided by weblate. Also something happened with the updated en-US json which now contains unrelated format changes which makes the whole commit look huge (+4,537 −3,899 ).

dmitrylyzo (commented):
It's also a good idea to make changes in a separate branch rather than using `master`.

PR #6694, opened 2025-04-02
Improve Album Sort on Songs Tab

Derek4aty1:
**Changes**
* Adds a secondary sorting of Album Artist to the Album sort in the Songs tab

It is rare for the average person to have albums with the same name across different artists, but it is possible. Using "Greatest Hits" as an example, with the current logic, albums songs from Fleetwood Mac's "Greatest Hits" album could be intertwined with The Bangles' "Greatest Hits" album. This PR adds a secondary sort on Album Artist such that this won't happen. The example I used is directly from the issue linked below, where you can see it visually. I reproduced on my end, which prompted me to fix it.

**Issues**
Fixes jellyfin/jellyfin#5733

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -96,6 +96,7 @@
 - [K. Kyle Puchkov](https://github.com/kepper104)
 - [ItsAllAboutTheCode](https://github.com/ItsAllAboutTheCode)
 - [Jxiced](https://github.com/Jxiced)
+- [Derek Huber](https://github.com/Derek4aty1)
 
 ## Emby Contributors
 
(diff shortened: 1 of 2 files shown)


thornbill:
It seems like this should follow `SortName` instead of being second in the list?

Derek4aty1:
Why? I'm not sure what SortName represents, but every other sort method has it last, so kept it last in the Album sort as well.

thornbill:
`SortName` is a normalized name field used for sorting. Primarily it's useful for things with a set order that is not alphabetical (episodes, tracks, etc.). It can also be set manually for specifying a custom order (useful for things like movie franchises). However as you pointed out `SortName` is already lower priority in the music view, so perhaps it's just not as relevant here. This is probably fine as is. :+1:

thornbill (approved):

PR #6706, opened 2025-04-04
Use server name for document title

thornbill:
**Changes**
Continuing the work on our branding configuration (https://github.com/jellyfin/jellyfin-meta/discussions/81) this uses the server name for the default document title.

**Issues**
N/A

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -80,8 +80,6 @@ build: ${__JF_BUILD_VERSION__}`);
     // Update localization on user changes
     Events.on(ServerConnections, 'localusersignedin', globalize.updateCurrentCulture);
     Events.on(ServerConnections, 'localusersignedout', globalize.updateCurrentCulture);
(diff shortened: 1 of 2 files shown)


src/index.jsx
```diff
@@ -80,8 +80,6 @@ build: ${__JF_BUILD_VERSION__}`);
     // Update localization on user changes
     Events.on(ServerConnections, 'localusersignedin', globalize.updateCurrentCulture);
     Events.on(ServerConnections, 'localusersignedout', globalize.updateCurrentCulture);
-    // Localize the document title
-    document.title = globalize.translateHtml(document.title, 'core');
```
thornbill, line 84: I really don't understand why we were doing this to begin with... the title starts out as "Jellyfin" which isn't translatable :man_shrugging:

viown (approved):

PR #6714, opened 2025-04-05
Fix filter name collisions

### LINKED ISSUE #6702: ParentalRating filter query parameter gets corrupted

### Describe the bug

When selecting parental ratings, the selection gets corrupted in certain cases, leading to inconsistent behaviour.

### Reproduction Steps

1. Go to library
2. Open Parental Rating Filter
3. Tick "FSK-12"
4. Tick "AL"
5. Tick "A"
6. Observe the requests in devtools

These are the filter parameters sent:
1. `OfficialRatings=FSK-12`
2. `OfficialRatings=FSK-12|AL`
3. `OfficialRatings=FSK-12L|A`

### Expected/Actual behaviour

The query to be correctly constructed, the Parameter should be `OfficialRatings=FSK-12|AL|A`

### Logs

_No response_

### Server version

master

### Web version

master

### Build version

1b797bbe184aa1515285362ee94c580470df5388

### Platform

Windows

### Browser

Firefox

### Additional information

_No response_

jagadam97:
fix: filter with delimiter generation closes #6702

**Changes**
for filters with delimiter current implementation uses string.replace function to remove duplicate filter if added. but if i add a filter that is prefix of existing filter then the prefix for item is getting removed and creates a filter that does not represent previous filters

example as in #6702
FSK-12 and AL are preapplied filters, user now tryes to add A
```javascript
("FSK-12|AL").replace("|A")
```
will run resulting in "FSK-12L" after filtering existing
then after addition it will become "FSK-12L|A"

this PR replaces the logic for above by converting the existing filter to Array and removing the checked / unchecked / selected Filter.

Screenshot From 2025-04-06 04-35-02

**Issues**

Fixes #6702

### DIFF
--- a/src/components/filterdialog/filterdialog.js
+++ b/src/components/filterdialog/filterdialog.js
@@ -351,7 +351,10 @@ class FilterDialog {
                 const filterName = chkGenreFilter.getAttribute('data-filter');
                 let filters = query.Genres || '';
                 const delimiter = '|';
-                filters = (delimiter + filters).replace(delimiter + filterName, '').substring(1);
+                filters = filters
+                    .split(delimiter)
+                    .filter((f) => f !== filterName)
+                    .join(delimiter);;
                 if (chkGenreFilter.checked) {
                     filters = filters ? (filters + delimiter + filterName) : filterName;
                 }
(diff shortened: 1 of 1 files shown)


thornbill (approved):

thornbill:
The write up here is excellent btw 💜

jagadam97:
> The write up here is excellent btw 💜

Thanks. :grinning:

PR #6748, opened 2025-04-11
Remove trailers tab

### LINKED ISSUE #4375: install the trailers channel

thornbill:
**Changes**
* Removes the trailers tab since there are no longer maintained plugins that provide trailers

**Issues**
Fixes #4375 
Closes #5846


nielsvanvelzen (approved):

vreelb:
Really going to great lengths to never approve my PR, huh? So much for ever contributing to this project. 😞

PR #6766, opened 2025-04-14
Update subtitle position styling

thornbill:
**Changes**
* Updates the positioning of subtitles when they are bottom aligned to maintain a constant spacing from the bottom of the screen

**Single Line**

| position | current | updated |
|---|---|---|
| -4 | | |
| -3 | line 1 | line 1 |
| -2 | | |
| -1 | | |

**Two Lines**

| position | current | updated |
|---|---|---|
| -4 | | line 1 |
| -3 | line 1 | line 2 |
| -2 | line 2 | |
| -1 | | |

**Four Lines**

| position | current | updated |
|---|---|---|
| -6 | | line 1 |
| -5 | | line 2 |
| -4 | line 1 | line 3 |
| -3 | line 2 | line 4 |
| -2 | line 3 | |
| -1 | line 4 | |

**Issues**
https://features.jellyfin.org/posts/3210/subtitle-position-correction-adjustment-line-1-2

### DIFF
--- a/src/components/subtitlesettings/subtitleappearancehelper.js
+++ b/src/components/subtitlesettings/subtitleappearancehelper.js
@@ -103,7 +103,7 @@ function getTextStyles(settings, preview) {
         const lineHeight = 1.35; // FIXME: It is better to read this value from element
         const line = Math.abs(pos * lineHeight);
         if (pos < 0) {
-            list.push({ name: 'margin-bottom', value: `${line}em` });
+            list.push({ name: 'margin-bottom', value: `${line - 1}em` });
             list.push({ name: 'margin-top', value: '' });
         } else {
             list.push({ name: 'margin-bottom', value: '' });
--- a/src/components/subtitlesettings/subtitleappearancehelper.js
+++ b/src/components/subtitlesettings/subtitleappearancehelper.js
@@ -101,13 +101,14 @@ function getTextStyles(settings, preview) {
     if (!preview) {
         const pos = parseInt(settings.verticalPosition, 10);
         const lineHeight = 1.35; // FIXME: It is better to read this value from element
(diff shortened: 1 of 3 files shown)


dmitrylyzo:
Does native subtitles (Chrome) work the same way?
https://github.com/jellyfin/jellyfin-web/blob/a840b1e8eec6174397ef4406f819255ec166e91e/src/plugins/htmlVideoPlayer/plugin.js#L1529

barronpm (approved):

thornbill:
Good catch @dmitrylyzo! This should be working in both cases now. I also adjusted the values since my previous change left you unable to position the subtitles at the very bottom of the screen.

src/components/subtitlesettings/subtitleappearancehelper.js
```diff
@@ -103,10 +103,10 @@ function getTextStyles(settings, preview) {
         const lineHeight = 1.35; // FIXME: It is better to read this value from element
         const line = Math.abs(pos * lineHeight);
         if (pos < 0) {
-            list.push({ name: 'min-height', value: `${line}em` });
+            list.push({ name: 'margin-bottom', value: `${line - 1}em` });
```
dmitrylyzo, line 106: Shouldn't `lineHeight` be subtracted instead of `1` to zero out the bottom margin?

dmitrylyzo (approved):
Seems to work.

GithubEnjoye:
The fix works on 10.11.0, but only when "Vertical position" is set to negative values like -3. When I set it to 13 to move the subtitles a little bit down, the subtitle growing/stacking behavior is the incorrect old one. Is there a possibility for a fix of the growing/stacking or even a subtitle rework for more granular position control and even custom font specification?

PR #6771, opened 2025-04-15
Add splash screen backdrop to login page

thornbill:
**Changes**
* Fixes the alignment of backdrops in the experimental layout since the pinned side bar is removed
* Removes the unused `enableImageRotation` function parameter from `setBackdrops()`
* Adds support for using the splash screen as a backdrop and enables it on the login page

Screenshot 2025-04-15 at 12-45-41 Stable Demo

**Issues**
N/A


nielsvanvelzen (approved):
Seems to work for me, one sidenote; the "forgot password" page could probably also have the backdrop

PR #6777, opened 2025-04-16
Update experimental layout library controls

thornbill:
**Changes**
* Updates the UI of the library controls in the experimental layout

**Movies Before**
Screenshot 2025-04-16 at 14-13-07 Jellyfin Edge

**Movies After**
Screenshot 2025-05-02 at 17-17-42 Jellyfin Edge

**Music Before**
Screenshot 2025-04-16 at 14-12-45 Jellyfin Edge

**Music After**
Screenshot 2025-05-02 at 17-18-13 Jellyfin Edge

**Issues**
N/A


thornbill:
- Shuffle for songs is gone
  - It seems that it was never enabled in the experimental layout, but now it is. 
- Shuffle jumps in after Play all. Do they have different conditions?
  - Shuffle is only shown if more than one item is displayed. I've updated it so none of those buttons are displayed until the items are fetched. 
- Also why is first button labeled e.g. `Play All`, next Time `Shuffle`, if we have no `Play all`
  - It was a design decision I made. It's nice to have text on the main action, but it might be a bit much on all of them and it definitely wouldn't fit on mobile. 🤷‍♂️

thornbill:
I've made some adjustments so pagination gets a full row on mobile. I have no data to back this up, but I _suspect_ Jellyfin users would prefer seeing the item count over page numbers. 😅

joshuaboniface (approved):

PR #6789, opened 2025-04-21
Update link for setup guide

felix920506:
**Changes**
Update link and string to go with https://github.com/jellyfin/jellyfin.org/pull/1379

**Issues**

### DIFF
--- a/src/strings/en-us.json
+++ b/src/strings/en-us.json
@@ -119,7 +119,7 @@
     "ButtonPause": "Pause",
     "ButtonPlayer": "Player",
     "ButtonPreviousTrack": "Previous track",
-    "ButtonQuickStartGuide": "Post-Install Setup Guide",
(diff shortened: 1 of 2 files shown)


JPVenson (approved):

src/strings/en-us.json
```diff
@@ -119,7 +119,7 @@
     "ButtonPause": "Pause",
     "ButtonPlayer": "Player",
     "ButtonPreviousTrack": "Previous track",
-    "ButtonQuickStartGuide": "Quick Start Guide",
+    "ButtonQuickStartGuide": "Post-Install Setup Guide",
```
thornbill, line 122: I think calling it "Post-Install" while you are going through the first time setup process to be confusing...

```suggestion
    "ButtonQuickStartGuide": "Setup Guide",
```

PR #6792, opened 2025-04-21
refactor: forgotPassword page to react component

MontejoJorge:
**Changes**

Refactor the forgotPassword  from the legacy controllers to react component.


src/apps/stable/routes/session/forgotPassword/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+            return apiClient.ajax({
+                type: 'POST',
+                url: apiClient.getUrl('Users/ForgotPassword'),
+                dataType: 'json',
+                contentType: 'application/json',
+                data: JSON.stringify({
+                    EnteredUsername: enteredUsername
+                })
+            });
+        },
```
viown, line 31: We should use the TS SDK when possible as the api client is considered legacy.

https://typescript-sdk.jellyfin.org/classes/generated-client.UserApi.html#forgotpassword

viown: Maybe you can use `ServerConnections.getCurrentApi();` to get an unauthenticated SDK client.

src/apps/stable/routes/session/forgotPassword/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+            const response = await getUserApi(currentApi).forgotPassword({
+                forgotPasswordDto: {
+                    EnteredUsername: enteredUsername
+                }
+            });
+
+            return response.data;
+        },
+        onSuccess: (result) => {
+            if (result.Action == 'ContactAdmin') {
```
viown, line 31: There is an enum for these.

https://typescript-sdk.jellyfin.org/enums/generated-client.ForgotPasswordAction.html

src/strings/es.json
```diff
@@ -714,6 +714,7 @@
     "MessageEnablingOptionLongerScans": "Activar esta opción implicará escaneos de la biblioteca más largos.",
     "MessageFileReadError": "Ha habido un error leyendo el fichero. Por favor, inténtalo más tarde.",
     "MessageForgotPasswordFileCreated": "Se ha creado el siguiente archivo en tu servidor y contiene instrucciones de cómo proceder",
+    "MessageForgotPasswordPinReset": "Introdice el PIN aqui para finalizar el reseteo de contraseña",
```
viown, line 717: You should only edit `en-us.json` when adding strings, as the other files are handled by weblate.

src/apps/stable/routes/session/forgotPassword/index.tsx
MontejoJorge: okay, fixed!

src/apps/stable/routes/session/forgotPassword/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+                msg += '<br/><br/>';
+                msg += globalize.translate('MessageForgotPasswordPinReset');
+                msg += '<br/><br/>';
+                msg += result.PinFile;
+                msg += '<br/>';
+                Dashboard.alert({
+                    message: msg,
+                    title: globalize.translate('ButtonForgotPassword'),
+                    callback: function () {
+                        navigate('forgotpasswordpin');
```
viown, line 57:
```suggestion
                        navigate('/forgotpasswordpin');
```
When I click submit I get navigated to a not found page. Possible fix? (haven't tried)

MontejoJorge: Yes, i have tried and it works, ty

viown (approved):

src/apps/stable/routes/session/forgotPassword/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+    }, [username, forgotPasswordMutation]);
+
+    const handleUsernameChange = useCallback((e: React.ChangeEvent<HTMLInputElement>) => {
+        setUsername(e.target.value);
+    }, []);
+
+    return (
+        <Page
+            id='forgotPasswordPage'
+            className='page standalonePage forgotPasswordPage mainAnimatedPage'
```
thornbill, line 80: The `Page` componenet automatically adds the `page` class.

```suggestion
            className='standalonePage forgotPasswordPage mainAnimatedPage'
```

src/apps/stable/routes/session/forgotPassword/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+                msg += result.PinFile;
+                msg += '<br/>';
+                Dashboard.alert({
+                    message: msg,
+                    title: globalize.translate('ButtonForgotPassword'),
+                    callback: function () {
+                        navigate('/forgotpasswordpin');
+                    }
+                });
+            }
```
thornbill, line 60: Seems like a good place to use a switch statement and it would be better to use the alert component directly instead of `Dashboard.alert` in my opinion.

thornbill (changes requested):
Looks great! I just had a couple minor suggestions. :+1:

src/apps/stable/routes/session/forgotPassword/index.tsx
viown: `SimpleAlert` is a mui component, we shouldn't use it here. I think thornbill meant that you should just call [alert directly](https://github.com/jellyfin/jellyfin-web/blob/bb810a183a3a083f81095e1486ac063f62c8f838/src/components/alert.js#L6).

MontejoJorge: Okay, fixed!

thornbill (approved):

PR #6795, opened 2025-04-22
Use local copy of connection manager

thornbill:
**Changes**
* Introduces a local copy of `ConnectionManager` to use instead of the version published in `jellyfin-apiclient`. Currently this is the same as the published version with some small cleanup applied. We want to introduce some changes (in future PR(s)) to support detecting when the server is starting up.
* Moves the minimum version handling from `ServerConnections` to `ConnectionManager`
* Moves some other apiclient related code to `lib/jellyfin-apiclient`

**Issues**
N/A


src/lib/jellyfin-apiclient/connectionManager.js
```diff
@@ -0,0 +1,760 @@
...
+            });
+        }
+
+        function validateAuthentication(server, serverUrl) {
+            return ajax({
+                type: 'GET',
+                url: `${serverUrl}/System/Info`,
+                dataType: 'json',
+                headers: {
+                    'X-MediaBrowser-Token': server.AccessToken
```
nielsvanvelzen, line 227: https://gist.github.com/nielsvanvelzen/ea047d9028f676185832e51ffaf12a6f :-)

Last time I checked this is the only part in jellyfin-web using a deprecated header so worth fixing now.

nielsvanvelzen (approved):

PR #6796, opened 2025-04-22
Update server update required handling

thornbill:
**Changes**
* Updates the handling of the server update required state

**Issues**
N/A


nielsvanvelzen (approved):

thornbill:
Ok I updated this to remove the dedicated route. Redirecting to/from the dedicated route when the connection state changed was overly complicated. Instead just rendering the error state in the current route context seems like a better approach.

A future PR will update handling for the `Unavailable` `ConnectionState`.

PR #6799, opened 2025-04-23
Fix lazy loading on the genre pages

### LINKED ISSUE #6779: Genre tab's sorting gets applied each time a row is scrolled into view

### Describe the bug

When visiting the Genre tab on shows page, the list of items for a genre gets show with the "Random" sort. When scrolling down and then up again, the order is not preserved.

### Reproduction Steps

1. Visit Shows page
2. Click on Genre
3. note order of first genre items

Image

4. Scroll items out of view
5. Scroll back up to genre

Image

Order is not preserved

### Expected/Actual behaviour

Order of items is preserved when scrolling

### Logs

_No response_

### Server version

10.10.7

### Web version

10.10.7

### Build version

10.10.7

### Platform

docker

### Browser

FF

### Additional information

_No response_

Jorgensonnj:
**Changes**

Changed the query object's `SortBy` value to be `SortName` from `Random`. This change keeps the ordering of the shows consistent when scrolling to and from a genre section of the genre tab.

**Issues**

Fixes #6779 Genre tab's sorting gets applied each time a row is scrolled into view


viown:
Using `Random` here is intended in order to avoid showing the same items over and over again. The issue is that the requests will still be made when the rows are scrolled into view.

I'm assuming this issue is a side-effect from the lazy loader which loads the rows as the user scrolls down. If we want to fix this we should prevent loading rows that have already been loaded.

JPVenson (approved):

Jorgensonnj:
**Changes**

- Undid my previous commit per the previous comment from @viown. 
- Added logic around the callback function of the `intersectionObserver` that executes the callback and then prevents it from future execution.

**Notes**

This change will impact movie genres, show genres, and images.

viown:
Awesome, it looks like the lazy loader never even worked as intended for the genres pages and your change fixes it. However, the image loader relies on the multiple execution of the callback in order to fill and empty the image as it [intersects and goes out of view](https://github.com/jellyfin/jellyfin-web/blob/20816d4edbadb035dede0617cf5447275697f876/src/components/images/imageLoader.js#L85).

I believe after your change, it'll never get emptied, so I'm thinking we could add an option to the lazy loader e.g `invokeOnce` to implement your behaviour and use that on the movies and tv show genre pages.

Jorgensonnj:
Third time is the charm!

**Changes**

- The creation of the `intersectionObserver` within the context of the `lazyLoader` will now pass along the `observer` to the callback function. This will allow control of the `observer` from within the callback function. 
- Lazy loaded movie and show items now explicitly `unobserve` the item after they have been loaded.

**Notes**

Image lazy loading functionality seemed to be unaffected by these changes, but please let me know if I missed something.

src/controllers/movies/moviegenres.js
```diff
@@ -124,6 +128,8 @@ export default function (view, params, tabContent) {
             if (result.Items.length >= query.Limit) {
                 tabContent.querySelector('.btnMoreFromGenre' + id + ' .material-icons').classList.remove('hide');
             }
+        }).then(() => {
+            observer.unobserve(elem);
```
viown, line 132: Calling unobserve after the items were fetched can leave a small window where `fillItemsContainer` could be called more than once. I think you can just call unobserve closer to the beginning of the function and that would be fine.

viown (approved):

PR #6803, opened 2025-04-24
Use playback manager to set default audio stream index

gnattu:
**Changes**

Use the method directly on the player might select a stream that is not playable and/or cause some inconsistencies on GUI. (Like track 1 is selected on the OSD but actually playing track 2). 

If we are doing more 10.10 releases we might want to backport this.

**Issues**


viown:
If the previous method isn't needed anymore, then we can remove it?

gnattu:
> If the previous method isn't needed anymore, then we can remove it?

Actually I'm not sure if it is needed or not. But if it is not being used by current code probably it is no longer needed.

PR #6824, opened 2025-05-02
Separate and refactor SubtitleSync and OffsetController

armanckeser:
Cleans up the implementation, separating offset controller and the subtitle sync feature itself.

**Changes**

**Issues**

This is to pave way for this feature https://github.com/jellyfin/jellyfin-web/pull/6735

## Summary by CodeRabbit

- **Refactor**
  - Improved the subtitle synchronization interface for better modularity and maintainability. No changes to user-facing features or interactions.


src/components/subtitlesync/subtitlesync.js
```diff
@@ -1,170 +1,225 @@
...
+        this.offsetController.adjustOffset(+this.subtitleSyncSlider.step);
     }
 
     decrementOffset() {
-        this.update(-subtitleSyncSlider.step);
+        this.toggle();
+        this.offsetController.adjustOffset(-this.subtitleSyncSlider.step);
+    }
+
+    destroy() {
```
dmitrylyzo, line 190: Maybe move `destroy` and `toggle` right after `constructor` to keep the old sequence?
Or do you have any thoughts on their current positions?

src/components/subtitlesync/subtitlesync.js
```diff
@@ -1,170 +1,225 @@
...
+        const value = parseFloat(this.slider.value) + delta;
+        this.slider.updateOffset(value);
+    }
 
-    instance.element = parent;
+    setOffset(offset) {
+        this.currentOffset = offset;
+        this.slider.value = offset.toString();
+        this.textField.updateOffset(offset);
+    }
```
dmitrylyzo, line 110: Alternatively (to `currentOffset` and `setOffset`), we can declare a property with a custom getter and setter.
```js
get currentOffset() {
    return parseFloat(this.slider.value);
}

set currentOffset(offset) {
    this.slider.value = offset.toString();
    this.textField.updateOffset(offset);
}
```

dmitrylyzo (commented):
Haven't tested, but looks good.

src/components/subtitlesync/subtitlesync.js
armanckeser: Good call, thanks! I also added the playback manager offset call into it

src/components/subtitlesync/subtitlesync.js
armanckeser: Nope, the old sequence sounds fine

src/components/subtitlesync/subtitlesync.js
armanckeser:
```
        slider.addEventListener('change', () => {
            // eslint-disable-next-line no-self-assign
            this.currentOffset = this.currentOffset;
        });
```

not a JS native, but I wanted the slider change to trigger setting of the offset so I am assigning the value to itself, is there a better way to do this, or am I missing something

armanckeser:
@dmitrylyzo I don't want to nag but wanted to confirm, for reviews do you prefer being tagged or do you just go through your notifications when you have the time?

src/components/subtitlesync/subtitlesync.js
```diff
@@ -1,125 +1,138 @@
...
+
+        textField.addEventListener('keydown', (event) => {
+            if (event.key === 'Enter') {
+                // if input key is enter search for float pattern
+                let inputOffset = /[-+]?\d+\.?\d*/g.exec(textField.textContent);
+                if (inputOffset) {
+                    inputOffset = parseFloat(inputOffset[0]);
+                    this.currentOffset = inputOffset;
+                } else {
+                    textField.textContent = (playbackManager.getPlayerSubtitleOffset(this.player) || 0) + 's';
```
dmitrylyzo, line 77:
```suggestion
                    textField.updateOffset(playbackManager.getPlayerSubtitleOffset(this.player) || 0);
```

In the next PR, we need to introduce a new translatable string for offset and use it in `textField.updateOffset`.

src/components/subtitlesync/subtitlesync.js
```diff
@@ -1,125 +1,138 @@
...
-        this.value = sliderValue === undefined ? 0 : sliderValue;
+    get currentOffset() {
+        return parseFloat(this.slider.value);
+    }
 
-        updateSubtitleOffset();
-    };
+    set currentOffset(value) {
+        this.slider.value = value.toString();
+        playbackManager.setSubtitleOffset(value, this.player);
```
dmitrylyzo, line 34:
```suggestion
        this.slider.value = value.toString();

        // rely on slider value trimming
        value = this.currentOffset;

        playbackManager.setSubtitleOffset(value, this.player);
```
Re-assign value to fix floating point rounding errors.
float-1
Previously, it was `const value = parseFloat(subtitleSyncSlider.value);`.
You may want to rephrase the comment.

src/components/subtitlesync/subtitlesync.js
dmitrylyzo: > not a JS native, but I wanted the slider change to trigger setting of the offset so I am assigning the value to itself, is there a better way to do this, or am I missing something (would assigning to `parseFloat(this.slider.value)` be more desirable?)

I'm fine with self-assigning approach. Alternatively, you can extract the playbackManager and textField updates in a separate function, as it was before: https://github.com/jellyfin/jellyfin-web/blob/fec58cc8aaa9dc665105d2677b5b721415360aee/src/components/subtitlesync/subtitlesync.js#L75-L81.

dmitrylyzo (approved):
LGTM. I think we should use "squash merge".

src/components/subtitlesync/subtitlesync.js
```diff
@@ -1,125 +1,142 @@
...
+        value = this.currentOffset;
+
+        playbackManager.setSubtitleOffset(value, this.player);
+        this.textField.updateOffset(value);
     }
 
-    subtitleSyncSlider.updateOffset = function (sliderValue) {
-        // default value is 0s = 0ms
-        this.value = sliderValue === undefined ? 0 : sliderValue;
+    _initSlider() {
```
thornbill, line 42: Small code style feedback: we should use actual private methods instead of the pseudo standard underscore prefix.

Also for future reference, [we don't use "conventional" commits](https://jellyfin.org/docs/general/contributing/development#pull-request-guidelines).

armanckeser: Ah, not a JS native (coming from Python), didn't even know JS had special syntax for private methods, good to learn. I made the change


For the commit, I am used to a squash merge and changed the name of the PR to non conventional, hopefully that is good.

thornbill:
Not a blocker but I would suggest removing the underscore from the method names

armanckeser:
@dmitrylyzo @thornbill I think this is ready now?
Also, is there any show or movie in the demo deployment with subtitles? If not adding one might make sense?

armanckeser:
@dmitrylyzo @thornbill, opened another small PR exposing the current track events. If we can get these in I can start working on the subtitle sync view if that's okay. Thank you!

armanckeser:
@thornbill hey there! Don't mean to be a thorn on you guys' side but the squeaky wheel gets the grease and I would love to contribute the subtitle sync rendering to this project. I can only imagine how hard it must be to maintain and develop this, so I am happy to accommodate for any preferences, anything that will decrease the merge time for this and the following PRS to days instead of months. Do let me know the conditions that would make you feel comfortable in merging these changes and ultimately the final version of #6735. The new PR I opened doesn't use conventional commits and I have squashed it down to a singular commit and plan to do the same for the final commit adding the timeline render.

src/components/subtitlesync/subtitlesync.js
```diff
@@ -1,125 +1,142 @@
...
+        playbackManager.setSubtitleOffset(value, this.player);
+        this.textField.updateOffset(value);
+    }
+
+    #initSlider() {
+        const slider = this.slider;
+
+        if (layoutManager.tv) {
+            slider.classList.add('focusable');
+            // TODO: Delay required for Firefox – wait for custom element to attach
```
dmitrylyzo: `TODO` is something that needs to be done.

`HACK` is something that already exists, but there may be a better way. It's just a reminder.

The delay has already been applied, so this is not a `TODO`. 🤷‍♂️

As an option, drop `TODO`, but I'd prefer it to be marked as `HACK`.
```suggestion
            // Delay required for Firefox – wait for custom element to attach
```

AFAIK, Bill can skip some inadequate/minor warnings from Sonar Cloud. Moreover, I believe that this code will gone after moving to React.

thornbill:
Hey @armanckeser I hate to leave you hanging on this, but I'm just really busy at the moment. On top of my normal responsibilities I am in the process of moving, so really unless something is critical for 10.11 or was work I already had in progress it's not getting the attention it deserves.

Thanks for your patience!

armanckeser:
Thanks for the heads up, good luck with the move

src/components/subtitlesync/subtitlesync.js
```diff
@@ -1,125 +1,142 @@
...
+            } else {
+                // keep focus to prevent fade with osd
+                textField.hasFocus = true;
+                if (event.key.match(/[+-\d.s]/) === null) {
+                    event.preventDefault();
+                }
+            }
 
-    subtitleSyncSlider.addEventListener('change', () => updateSubtitleOffset());
+            // TODO: TV layout will require special handling for navigation keys. But now field is not focusable
```
dmitrylyzo: same
```suggestion
            // TV layout will require special handling for navigation keys. But now field is not focusable
```

dmitrylyzo (approved):

armanckeser:
@thornbill do you see a world where this feature gets merged? i got some time for a few weeks and would love to get to that world in the way you would feel comfortable

dmitrylyzo (approved):
Just in case. The changes still look fine. Tested with the current `master` and #8153.

Issue #22, opened 2019-01-12
Refresh page after deleting a media element


dinki:
When deleting a media file (tv show) from the web interface from the 'Tv Shows' cover view, the page should refresh to reflect the change.  Currently if you delete a file you really have no way of knowing if it worked unless you manually refresh the page.  Eventually it would be cool if the page would be handled like Plex does and the item would be deleted and then the others would shift to take it's place.  This, of course, requires more than just a 'static' HTML page.

dinki:
Additionally, if I go into the web ui, select a movie and click 'edit images'.  I upload a new image.  At that point the movie has a new image but it does not update the web page I previously had with the wrong image.  I have to hit F5 to see the new image change on the page.  JF should call a page reload/refresh to show this updated image.

dinki:
Excuse the mess.  I tried combining what I thought were similar issues but ended up splitting them (https://github.com/jellyfin/jellyfin-web/issues/125)

PR #6832, opened 2025-05-03
Immediately update UI to reflect media deletion

markreading:
**Changes**

Currently, when deleting a media item using the context menu, the request seemingly completes, but no changes to the UI are made until the user refreshes the page, and finds that the media is now gone.  I have added a few lines that will immediately remove the deleted media from the list in the UI after the delete API call succeeds.

This change reflects this feature request found [here](https://features.jellyfin.org/posts/169/refresh-page-after-deleting-a-media-element), and is also a fix I have wanted for a little while as well.

**Issues**

N/A

### DIFF
--- a/src/components/itemContextMenu.js
+++ b/src/components/itemContextMenu.js
@@ -557,7 +557,18 @@ function executeCommand(item, id, options) {
                 getResolveFunction(resolve, id)();
                 break;
             case 'delete':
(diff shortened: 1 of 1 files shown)


markreading:
@dinki
https://github.com/jellyfin/jellyfin-web/issues/22

PR #6835, opened 2025-05-07
Add ability to use React component in old Emby code

TheBosZ:
**Changes**
This PR shows how the old Emby front-end code can be modified to use the React components so React can be incrementally implemented "bottom-up" rather than "top down" (wait, which one is which? 😄). 

The main method here is to use `createRoot` from ReactDom to create new React areas in the old HTML. Because the React components are written in TSX, I wrote a [wrapper helper](src/controllers/itemDetails/reactUtils.tsx) to act as a bridge between the old code and the new code. 

Unfortunately, using `createRoot` loses the context that some of the React components rely on. To prevent that from being a problem, the bridge wraps each root in the same context components as they are down in the "real" root. This presents a problem though: we don't have access to the router at this level so anything that uses the router hooks (like `useLocation`) won't work. I'm not sure if there's a way to pull it in as well, but it seemed unlikely. So, as a band-aid, the bridge doesn't use the regular `UserThemeProvider` context but instead uses a plain `ThemeProvider` hardcoded to use the dark theme. Since the only place that this matters is the Dashboard, I figured it was an okay shortcut. We could potentially change `UserThemeProvider` to not use `useLocation` and fix the issue by just using window.location instead.

This PR invokes the bridge in two places: the media info bar and the Play/Resume button. Why those two? To show two very different usages. The info bar uses the PrimaryMediaInfo component which takes a bunch of props and sets up the whole bar while the Play/Resume button uses the PlayOrResumeButton component which is really just a straight-forward component.

To assist in the PrimaryMediaInfo component, I added a "go-between" component called MiscInfo. It deconstructs the options object from the old JS code and uses it to mount PrimaryMediaInfo. This isn't the only way this could work, but seemed like an okay solution to eventually be removed.

Additionally, I fixed up a bunch of issues in the components to make them look/act as close to the current code as possible.

**Issues**
This probably belongs with the [project to migrate to React and Typescript](https://github.com/orgs/jellyfin/projects/43?pane=issue&itemId=30109885).

The [React docs](https://react.dev/learn/add-react-to-an-existing-project) were an invaluable help.


TheBosZ:
@thornbill This is probably one you'll be interested in.

thornbill:
~~@jellyfin-bot rebase~~ whoops that is no longer supported

TheBosZ:
> ~@jellyfin-bot rebase~ whoops that is no longer supported

@thornbill Did you want me to rebase and fix conflicts? I wasn't sure if this was code you actually wanted to merge or not

thornbill:
No I was trying to be lazy and get the conflicts resolved before picking the changes into my branch. Thanks again for this! You can see the change in https://github.com/jellyfin/jellyfin-web/pull/6918 but your authorship is retained. :purple_heart:

TheBosZ:
> No I was trying to be lazy and get the conflicts resolved before picking the changes into my branch. Thanks again for this! You can see the change in #6918 but your authorship is retained. 💜

Nice! I'll close this PR then since it's outdated.

PR #6838, opened 2025-05-09
add remember me option at login

fluffy88:
Add "Remember Me" option to login screen.

**Changes**
This PR adds a new remember me checkbox to the login screen for the visible profiles. The intention here is to make the profile selection page show up every time the app/page is loaded. This is inline with Netflix/Disney style profile selection screens.

By default the checkbox is checked therefore current existing behaviour remains unchanged.
Users who deselect the checkbox will be shown the profile selection screen on every app start or page refresh.

Screenshot with the new checkbox.
Screenshot from 2025-05-09 18-08-28

I believe this feature is a desirable one as the Android client already provides a similar option and I have found a few Reddit threads talking about this behaviour. However, I understand if this particular implementation is not ideal. If you are not happy with the UX please provide me some feedback/suggestions and I can try implementing a different approach.

### DIFF
--- a/src/controllers/session/login/index.html
+++ b/src/controllers/session/login/index.html
@@ -31,6 +31,12 @@
         <div class="visualLoginForm" style="text-align: center;">
             <h1 style="margin-top:1em;">${HeaderPleaseSignIn}</h1>
             <div id="divUsers" class="itemsContainer vertical-wrap centered"></div>
+            <div class="itemsContainer vertical-wrap centered">
+                <label class="checkboxContainer" style="width: auto;">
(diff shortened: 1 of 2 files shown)


Chris4942:
@fluffy88 I checked out your branch, but I couldn't figure out how to reproduce the steps. It looks like the intended behavior is
> By default the checkbox is checked therefore current existing behaviour remains unchanged.
> Users who deselect the checkbox will be shown the profile selection screen on every app start or page refresh.

It looks like even if I deselect the checkbox, I still don't see the new behavior on refresh. I just stay logged into the same user as before without the option to add another

Issue #6261, opened 2024-10-27
Backdrop screensaver black borders


Arrow420:
**Describe The Bug**

Backdrop images smaller than screen resolution results in ugly black borders around instead of scaling up the image to screen size.

**Steps To Reproduce**

1. Have a backdrop images smaller than your screen resolution. (eg. 1280x720 image on a 1920x1080 display)
2. Enable backdrop screensaver.
3. Wait for it cycle to the image.
4. Image is padded with black borders

**Expected Behavior**

Lower resolution backdrop images should be scaled to cover the whole screen when using the screensaver.

**Logs**

\-

**Screenshots**
![](https://i.imgur.com/rPVKamL.jpeg)

![](https://i.imgur.com/YxgVqZ7.jpeg)

**System (please complete the following information):**
 - Platform: Winows
 - Browser: Jellyfin Media Player
 - Jellyfin Version: 10.10.0

**Additional Context**

\-

PR #6840, opened 2025-05-11
Backdrop Screen Saver enhancements

Kashugoyal:
### Backdrop Screensaver enhancements
This PR adds netflix/plex like screensaver backdrop look and feel to jellyfin web. 

#### Changes
1. Make backdrop image **cover the entire screen**
2. Fix title and description text display
   * Description is community rating (if available) and production year (if available)
4. Add animation to backdrop images and improve text layout and rendering
5. Increase default backdropScreensaverInterval **from 5 to 30 seconds**, to better align with animation effects

This MR also resolves the issue #6261 ✅ 

|Before|After|
|---|---|
|old| new|
|Video: https://github.com/user-attachments/assets/a70f1819-8aa2-45ce-9acb-c71172aee0ea| Video: https://github.com/user-attachments/assets/e1ca12d6-09c6-44c4-b782-39af0f1cc09c|

#### Future Work
We can allow the user to customize the overlay text through the settings.

#### Seeing it in action
To see the changes in action, 
1. Open the preview url from the cloudflare deployment below
2. Once you sign in, go to Profile settings -> Display
3. Change Screensaver from None to Backdrop
4. Let your screen be idle for 180s and voila !!

### DIFF
--- a/src/components/slideshow/slideshow.js
+++ b/src/components/slideshow/slideshow.js
@@ -412,7 +412,9 @@ export default function (options) {
         return getSwiperSlideHtmlFromSlide({
             originalImage: getImgUrl(item, currentOptions.user),
             Id: item.Id,
-            ServerId: item.ServerId
+            ServerId: item.ServerId,
+            title: item.Name,
+            description: getSlideDescriptionFromItem(item)
         });
     }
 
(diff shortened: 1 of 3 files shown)


ZachLobato (commented):
Hey this seems pretty neat. A few things that may be worth addressing:

- The description says the default is raised from 5 to 30 seconds, but the code change sets `|| 20`. Could you clarify which is correct and update the description or the code?
-  Animating `margin-left` triggers a full layout recalculation on every frame which could cause performance issues on lower-end devices. A composited equivalent would be something like:

    ```css
    @keyframes backgroundScroll {
        0%   { transform: translateX(-10%); }
        100% { transform: translateX(10%); }
    }
    .swiper-slide-img {
        height: 120%;
        width: 120%; 
        object-fit: cover;
    }
    ```

    `transform` is also GPU-composited and won't mess with the layout on neighboring elements.

PR #6842, opened 2025-05-12
Hardcode vidaa hevc playback is unsupported

93Pd9s8Jt:
Hardcode vidaa lack of support for hevc

See https://forum.jellyfin.org/t-setting-up-jellyfin-on-a-hisense-vidaa-os-tv for more context on why this is desirable.

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -11,7 +11,7 @@ function canPlayHevc(videoTestElement, options) {
         return true;
     }
 
-    if (browser.ps4) {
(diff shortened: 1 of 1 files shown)


dmitrylyzo:
https://github.com/jellyfin/jellyfin-web/pull/6595

93Pd9s8Jt:
> #6595 More precisely, [#6595 (comment)](https://github.com/jellyfin/jellyfin-web/pull/6595#issuecomment-2745185970)

Thanks, my bad for not seeing that. I'll close this.

PR #6843, opened 2025-05-12
Add targetABI to plugin repo page

enter-a-random-username:
**Changes**

Add the targetABI. Especially useful for Unstable/master and if the plugin is updated or not. Fast check for reproting reporting something, build/patch something or maybe deactivate it.

### DIFF
--- a/src/apps/dashboard/features/plugins/components/PluginDetailsTable.tsx
+++ b/src/apps/dashboard/features/plugins/components/PluginDetailsTable.tsx
@@ -86,6 +86,18 @@ const PluginDetailsTable: FC<PluginDetailsTableProps> = ({
                         }
                     </TableCell>
                 </TableRow>
+                <TableRow>
+                    <TableCell variant='head'>
+                        {globalize.translate('TargetABI')}
+                    </TableCell>
+                    <TableCell>
+                        {
+                            (isRepositoryLoading && <Skeleton />)
+                            || pluginDetails?.version?.targetAbi
+                            || globalize.translate('Unknown')
+                        }
+                    </TableCell>
+                </TableRow>
             </TableBody>
         </Table>
     </TableContainer>


src/apps/dashboard/features/plugins/components/PluginDetailsTable.tsx
```diff
@@ -86,6 +86,18 @@ const PluginDetailsTable: FC<PluginDetailsTableProps> = ({
                         }
                     </TableCell>
                 </TableRow>
+                <TableRow>
+                    <TableCell variant='head'>
+                        {globalize.translate('TargetABI')}
```
viown, line 91: This string wasn't added?

jumoog:
Good idea!

nielsvanvelzen:
I don't think it makes sense to add this to the UI. The TargetAbi property in the plugin manifest is meant for the server, not for the user.

jumoog:
> > I don't think it makes sense to add this to the UI. The TargetAbi property in the plugin manifest is meant for the server, not for the user.
> 
> I disagree, the current situation is more than optimal  given that https://github.com/jellyfin/jellyfin/pull/12460 has been closed with users being allowed to install incompatible plugins that will in the worse case crash the system, I think it is commendable to give users more information about a plugin and its expected API compatibility.


No it's still a problem. I closed my PR because I solved this by reading the jellyfin user agent and serving the correct json file. 

See: https://github.com/intro-skipper/server_configs/blob/main/Caddyfile

src/apps/dashboard/features/plugins/components/PluginDetailsTable.tsx
viown: If you don't intend to localize the string, then you can just use "Target ABI" directly because `globalize.translate` will error in console logs when it can't find the key.

thornbill:
Closing based on Niels' feedback above.

PR #6866, opened 2025-05-15
 Add Play All and Shuffle buttons to Music Videos Libraries 

theguymadmax:
Adds Play All and Shuffle buttons to the Music Video libraries. For home media, these functions are only available for photo playback in the Home Videos and Photos libraries. This update provides users with a dedicated space to place videos and use these playback features.

**Issues**
 N/A

### DIFF
--- a/src/controllers/list.js
+++ b/src/controllers/list.js
@@ -890,7 +890,7 @@ class ItemsView {
                     // Folder, Playlist views
                     && itemType !== 'UserView'
                     // Only Photo (homevideos) CollectionFolders are supported
-                    && !(itemType === 'CollectionFolder' && item?.CollectionType !== CollectionType.Homevideos)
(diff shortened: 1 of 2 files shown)


src/controllers/list.js
```diff
@@ -890,7 +890,7 @@ class ItemsView {
                     // Folder, Playlist views
                     && itemType !== 'UserView'
                     // Only Photo (homevideos) CollectionFolders are supported
-                    && !(itemType === 'CollectionFolder' && item?.CollectionType !== CollectionType.Homevideos)
+                    && !(itemType === 'CollectionFolder' && item?.CollectionType !== CollectionType.Homevideos && item?.CollectionType !== CollectionType.Musicvideos)
```
thornbill, line 893: For both of these checks the comment preceding it should also be updated

PR #6868, opened 2025-05-15
Fix link - skip redirection

ireun:
**Changes**
Old link is being redirected, this commit skips the redirection


thornbill:
I believe there is another one of these links somewhere. If you could fix it also that would be awesome!

Issue #4779, opened 2023-09-16
`min-width` and `max-width` conditional for header goes into two containers even when on higher resolutions


Zelaf:
**Describe The Bug**
So I noticed that my Jellyfin web client has a larger header than some screenshots and noticed that on my laptops resolution of `2880x1800` with a Windows scaling on 200%, which is the default for this resolution, Jellyfins web client goes into the "mobile" or rather `max-width` mode on the styling inside of the `@media (max-width: 100em)` conditional and have the sections split instead of inline. Since I have such a high res display it shouldn't.

**Steps To Reproduce**

1. Have a potentially odd-ish resolution / scaling enabled
2. Go to website and see the header sections split.

**Expected Behavior**
Since it's a laptop display with a high resolution the header should be inline.

**Logs**
I don't think logs are necessary, will be provided if asked.

**Screenshots**

Windows scaling set to 200%
![image](https://github.com/jellyfin/jellyfin-web/assets/10819462/89f41efe-72fd-4114-81f4-8090632d2a03)
Zooming out site to 80%
![image](https://github.com/jellyfin/jellyfin-web/assets/10819462/a0e4e7db-784f-4b70-9590-7c9af53ee336)

Windows scaling set to 175% (No page zoom)
![image](https://github.com/jellyfin/jellyfin-web/assets/10819462/0679cf28-1233-4054-bb4a-d3b211abacc0)

**System (please complete the following information):**
 - Platform: Windows
 - Browser: Firefox
 - Jellyfin Version: 10.8.10

**Additional Context**
Setting the scaling to 175% in Windows generally fixes the issue. I was also able to fix the issue by editing the CSS in the developer tools. Setting `@media (max-width: 90em)` fixed the issue as well and doing a very small bit of testing of the different phone presets in the developer tools didn't seem to have an impact on how it works on mobile devices. However such a fix should be tested further.

The issue is also similar to the one reported at #2894

I would suggest to use another value to guess what the device needs. With Media Queries Level 4 there's the possibilities of querying for portrait or landscape with `orientation`, `device-aspect-ratio` which could include common phone/desktop aspect ratios, `device-width` which could include range of common phone resolutions or `pointer` which can detect if there's a mouse cursor or not. Any of these or a combination of these could potentially be more stable when detecting if the device is a phone or not. I'd love to hear some thoughts of these suggestions!

GouravKumar13:
@Zelaf  can please assign this to me

[...]

PR #6872, opened 2025-05-16
Fix header layout issue on high-DPI displays with scaling

Kusnezow94:
**Changes**
Fix header layout issues on high-DPI displays with scaling by implementing a targeted CSS solution. Analyzed how the media query breakpoint at 1130px interacts with scaled displays, identifying that content was being overlapped by the two-row header. Maintained the original breakpoint for compatibility while adding targeted margin-top to page content elements. Created specific exclusions for detail pages to prevent unnecessary spacing. Tested the solution across multiple display configurations to ensure consistent behavior on both standard and high-DPI displays with various scaling factors.

**Issues**
https://github.com/jellyfin/jellyfin-web/issues/2894
https://github.com/jellyfin/jellyfin-web/issues/4779

### DIFF
--- a/src/components/layoutManager.js
+++ b/src/components/layoutManager.js
@@ -43,11 +43,34 @@ class LayoutManager {
             }
         }
 
+        // Add a class to handle high-DPI displays with scaling
+        this.handleHighDpiDisplays();
+
         Events.trigger(this, 'modechange');
     }
 
     getSavedLayout() {
         return appSettings.get('layout');
+    }
+
+    handleHighDpiDisplays() {
+        // Check if we're on a high-DPI display with scaling
+        const devicePixelRatio = window.devicePixelRatio || 1;
+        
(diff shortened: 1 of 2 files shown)


theguymadmax:
That looks so much better, I always thought it looked weird. One thing I’ve noticed in experimental mode: the spacing between the menu bar and the My Media section now seems too much.

How the current master displays on my screen: 






How it displays with this fix: 

Fix 1
Fix 2

Kusnezow94:
@theguymadmax thanks for the kudos! Actually this "bigger" margin, is the standard master margin. It just gets smaller, because the master pushes the navigation down, without adjusting the margin for it - worst case (which i luckly did not see), is that it could even hide content behind it, like pagination for example. If you zoom out on master, just set the page zoom to 80%, you'll see it pop right back in its place and the margin will be bigger then. :-)

PR #5669, opened 2024-06-06
Force DoVi on browser.xboxOne as edgeUWP says it can't play it

RaafatAkkad:
Unsure if DoVi can be used with profile 8 on Xbox but video play

https://ott.dolby.com/codec_test/index.html
Screenshot 2024-06-10 at 8 50 16 PM

PR #6900, opened 2025-05-25
Enter FullScreen before playback, passing video info to enable DoVi support on Xbox 

RaafatAkkad:
**Changes**

In order for DoVi videos to play on Xbox, the client must switch the resolution to DolbyVisionLowLatency. Currently `enableFullscreen()` is called after the video begins playing but the Xbox expects the display to be in DoVi mode before. This PR moves `enableFullscreen()` & `disableFullscreen()` to playbackmanager.js. This PR also extends the `window.NativeShell.enableFullscreen()` to take in video information so that NativeShell can switch the display resolution, hdr mode. This also allow a client to attempt to match the display to the resolution & refresh rate.

Note: The Xbox is an always `full screen` device, but I have used `enableFullscreen()` to switch the resolution to match the video (typically 24hz) and then `disableFullscreen()` to switch it back to the default mode (typically 60hz)

The changes can be tested on Xbox One/Series by compiling `jellyfin-uwp` using my fork https://github.com/jellyfin/jellyfin-uwp/pull/46 . https://github.com/jellyfin/jellyfin-web/pull/5669 is also required.

Note: This PR is based on https://github.com/jellyfin/jellyfin-web/pull/5628 with conflicts resolved

image

**Testing:**
I have tested `jellyfin-web` on
- Xbox Series X (The 1080p & 4K profile 5 video from https://developer.dolby.com/tools-media/sample-media/video-streams/dolby-vision-streams plays.)
- Windows 11 (to check enableFullScreen/disableFullScreen works but I don't have a HDR capable monitor so playback doesn't work)

**Issues**

### DIFF
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -12,6 +12,7 @@ import itemHelper from '../itemHelper';
 import { pluginManager } from '../pluginManager';
 import PlayQueueManager from './playqueuemanager';
 import * as userSettings from '../../scripts/settings/userSettings';
+import shell from '../../scripts/shell';
 import globalize from '../../lib/globalize';
 import loading from '../loading/loading';
 import { appHost } from '../apphost';
@@ -2707,6 +2708,18 @@ export class PlaybackManager {
                     playerData.maxStreamingBitrate = maxBitrate;
                     playerData.streamInfo = streamInfo;
 
+                    if (streamInfo.fullscreen && streamInfo.mediaType === 'Video') {
+                        const videoStream = mediaSource.MediaStreams.find(ms => ms.Type === 'Video');
+                        if (videoStream) {
+                            shell.enableFullscreen({
+                                videoWidth: videoStream.Width,
+                                videoHeight: videoStream.Height,
+                                videoFrameRate: videoStream.ReferenceFrameRate || videoStream.RealFrameRate,
+                                videoRangeType: videoStream.VideoRangeType
+                            });
+                        }
+                    }
+
                     return player.play(streamInfo).then(function () {
                         loading.hide();
                         onPlaybackStartedFn();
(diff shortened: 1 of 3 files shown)


src/components/playback/playbackmanager.js
```diff
@@ -3412,6 +3425,10 @@ export class PlaybackManager {
                 }
             }
 
+            if (streamInfo?.fullscreen && streamInfo?.mediaType === 'Video') {
+                shell.disableFullscreen();
+            }
```
dmitrylyzo, line 3430: Doesn't it do the same via `onPlaybackStopped.call`?

src/components/playback/playbackmanager.js
```diff
@@ -2707,6 +2708,18 @@ export class PlaybackManager {
                     playerData.maxStreamingBitrate = maxBitrate;
                     playerData.streamInfo = streamInfo;
 
+                    if (streamInfo.fullscreen && streamInfo.mediaType === 'Video') {
+                        const videoStream = mediaSource.MediaStreams.find(ms => ms.Type === 'Video');
+                        if (videoStream) {
+                            shell.enableFullscreen({
+                                videoWidth: videoStream.Width,
+                                videoHeight: videoStream.Height,
+                                videoFrameRate: videoStream.RealFrameRate,
```
dmitrylyzo: There's `ReferenceFrameRate` (https://github.com/jellyfin/jellyfin/pull/12603). Maybe we should use it instead of (or in addition to) `RealFrameRate`?
```suggestion
                                videoFrameRate: videoStream.ReferenceFrameRate || videoStream.RealFrameRate,
```

RaafatAkkad: Thanks. done.

JPVenson:
No longer required as now handled by the xbox via its own plugin

PR #6904, opened 2025-05-28
Fix comic page sorting by including natural path and filename order, Fixing #6642

### LINKED ISSUE #6642: Page Order issues with CBZ files containing folders.

### Describe the bug

When a CBZ file that has folders containing images (a common method of handling chapters) the page Order is incorrect with the current order being used sorting at the file-name level then folder level (EG: c01/p01.png -> c02/p01.png -> c03/p01.png -> v01/p02.png) instead of sorting by absolute truncated name sequences.

Chrome and Firefox both preform different sorting to each other, but neither are the correct method. 

- Firefox sports in reverse chapter but correct page order.
- Chrome sports in correct chapter and correct page order.
- Both go through one page in each chapter before moving on to the next page, again in each chapter.

### Reproduction Steps

1. Download the following file: [TestCBZ.zip](https://github.com/user-attachments/files/19405348/TestCBZ.zip)
2. Rename from `TestCBZ.zip` to `TestCBZ.cbz` and upload to Jellyfin in a book library.
3. Attempt to view the book in the web-viewer
4. Page through the book in Chrome or Firefox
5. Note that you are moving through chapters then moving to then to the next page number.

### Expected/Actual behaviour

Intended behavior is as follows:

1. Sort the level you are at in alphabetical order
2. Select the first item.
3. If a folder, open the folder and repeat from step 1 at the new level
4. If an image, display the image.
5. If there is another file at the folder-level, repeat from step 3
6. If there is not another file, exit the folder and resume the parent level of recursion.
7. If there are no higher levels of recursion, you have finished the file.

### Logs

This is a logic / sort issue, as such there are no log errors available.

### Server version

10.10.6

### Web version

10.10.6

### Build version

10.10.6

### Platform

Chrome OS, Windows 10

### Browser

Chrome, Chrome & Firefox

### Additional information

~~I'll also be adding an issue regarding chapter support in CBZ files momentarily.~~ It may be worthwhile to add in chapter support for CBZ files at the same time as they can naively be defined as the first image in a folder.

terokorp:
Previously, folder names (i.e. file paths) were ignored during sorting, which caused pages from different chapters or volumes to be mixed together. Additionally, lexicographic filename sorting caused issues such as "Vol. 12 Ch. 56" appearing before "Vol. 1 Ch. 1".

Changes:
- Sorting now considers both `file.path` and `file.name`, using a natural sort algorithm.
- Paths and names are normalized (trimmed and cleaned) to handle inconsistencies like extra spaces or missing values.
- This ensures that pages within the same chapter or volume appear grouped and in the correct numerical order, including support for subchapters like `Ch. 4.5`.

**Issues**
Fixes #6642


terokorp:
This PR has been open over 7 months. Is here something to change/fix or problems with my PR or what should I do to get this merged?
Initially github-actions[bot] Requested changes and those are already fixed, but request didn't go away.
@thornbill

pedroadame:
+1 to merging this, some mangas are impossible to read on Jellyfin

PR #6906, opened 2025-05-29
Add Proxy polyfill

dmitrylyzo:
https://github.com/jellyfin/jellyfin-web/pull/6787 broke old browsers (webOS 1.2) because tanstack-query started using `Proxy`.

**Changes**
Add Proxy polyfill.

**Issues**
N/A

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -53,6 +53,7 @@
         "material-react-table": "3.2.1",
         "native-promise-only": "0.8.1",
         "pdfjs-dist": "3.11.174",
+        "proxy-polyfill": "0.3.2",
         "react": "18.3.1",
         "react-blurhash": "0.3.0",
         "react-dom": "18.3.1",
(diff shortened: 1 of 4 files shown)


ferferga:
Just as an FYI, Proxy is impossible to fully polyfill, hence why after the Vue 3 update it was discussed on team that Jellyfin Vue will never be a suitable 1:1 match with jf-web because of the lack of back compat.

At some point I guess we can "bump" the minimum required version for Web and this might not be necessary, but in the meantime perhaps you should look into replacing TanStack (or asses how Proxy is used internally to avoid it's use altogether, in case TanStack only uses it for specific methods).

thornbill (approved):

thornbill:
Regarding ferferga's comment, I assume you tested and verified the polyfill fixes the issue?

ferferga:
@thornbill Given Dimitry's track record when giving all the attention to the details in PRs, I assumed so rightaway.

My point is that it could work now, but it might not at some point (depends on which extent TanStack continues using Proxy internally). Perhaps sooner than later they might make use of unpolyfiable Proxy features, so I'd keep an eye specially since you're actively incorporating it to the codebase, it's best to rip it now in case it's needed than doing it when it's used everywhere (I'm already living this with Vuetify at Vue lol 😅).

PR #6915, opened 2025-06-02
Prevent unwanted navigation history changes during horizontal scroll for trackpads

fnbutton:
**Summary:**
This PR introduces a `preventDefault` mechanism to prevent unintended backward/forward navigation in the browser history when performing a two-finger horizontal scroll gesture on a trackpad.

**Details:**
Previously, horizontal trackpad scrolling could inadvertently trigger navigation history changes, especially during aggressive or boundary-hitting scroll movements. This change specifically restricts history navigation by evaluating the scroll start, end, and the intensity of the scroll movement.

**Visual Reference:**
Please refer to the accompanying GIF for a demonstration of the fix.

Before:
annoying-behavior

After:
behavior-fixed

### DIFF
--- a/src/lib/scroller/index.js
+++ b/src/lib/scroller/index.js
@@ -599,6 +599,13 @@ const scrollerFactory = function (frame, options) {
         let delta = normalizeWheelDelta(event);
 
         if (transform) {
+            if (o.horizontal && event.deltaX !== 0 &&
+                (event.deltaY >= -5 && event.deltaY <= 5) &&
+                (pos.dest + o.scrollBy * delta > 0) &&
+                (pos.dest + o.scrollBy * delta < pos.end)
+            ) {
+                event.preventDefault();
+            }
             self.slideBy(o.scrollBy * delta);
         } else {
             if (isSmoothScrollSupported) {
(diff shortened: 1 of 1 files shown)


src/lib/scroller/index.js
```diff
@@ -794,7 +801,7 @@ const scrollerFactory = function (frame, options) {
             if (o.mouseWheel) {
                 // Scrolling navigation
                 dom.addEventListener(scrollSource, wheelEvent, scrollHandler, {
-                    passive: true
+                    passive: false
```
dmitrylyzo, line 804: I believe you also need to adjust `removeEventListener`: https://github.com/jellyfin/jellyfin-web/blob/83522f88b6087455011152077e9a2b70439f6236/src/lib/scroller/index.js#L639-L641
and probably the second `addEventListener`: https://github.com/jellyfin/jellyfin-web/blob/83522f88b6087455011152077e9a2b70439f6236/src/lib/scroller/index.js#L811-L813

fnbutton: Thanks , i forgot it, fixed now.

thornbill (approved):

PR #6928, opened 2025-06-07
Resumes playback after reload during playback (#559)

### LINKED ISSUE #559: Refresh page during stream breaks web player
**Describe the bug**

When playing a video stream in the web player, if you refresh the page, you get given a plain error page saying "Value cannot be null. Parameter name: path".

When watching the stream, before refreshing, the URL is always the same: `http://<server address>/web/index.html#!/videoosd.html`

After refreshing, the URL is set to `http://<server address>/web`. At this point if you press the browser's back button it will keep redirecting you back to the same error page, unless you press back fast enough to get back into the main UI (still doesn't get you back to the stream).

It looks like there is no browser URL routing being done for streams so when you refresh it it doesn't know what stream to go back to, it only knows you were on `videoosd.html`.

**To Reproduce**

1. Start watching a movie/TV episode
2. Refresh the page
3. See error

**Expected behavior**

I would expect refreshing the page to continue the movie/TV episode where you left off, just like Plex does.

**Logs**

**Screenshots**

screenshot

 - OS: Server is running Debian 10.1, client is running Arch Linux
 - Browser: qutebrowser
 - Jellyfin Version: 10.4.0
 - Reverse proxy: No

**Additional context**

portalman32:
**Changes**
I modified the following files to fix the bug from issue 559:
- `src/controllers/playback/video/index.js`
- `src/components/playback/playbackmanager.js`

The `playbackmanager` saves the item id in session storage once playback has begun.

If a reload is performed during playback, the item id is retrieved from session storage. Then the last saved position ticks are retrieved using the `apiClient`.  If there are no saved position ticks for that item id, the playback position defaults to zero ticks. Finally, the video is resumed using the `playbackmanager`.

**Current:**
The current functionality is that when a user reloads the page during playback, the user goes back to `/home`

**Now:**
After my changes, the user will resume the last saved playback position on the video after a reload during playback.

https://github.com/user-attachments/assets/ce88b889-83e8-4a51-9744-07bdfdd38392

**Issues**
Fixes #559

### DIFF
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -3255,6 +3255,10 @@ export class PlaybackManager {
                 playerData.secondarySubtitleStreamIndex = null;
             }
 
+            if (streamInfo?.item?.Id) {
+                sessionStorage.setItem('lastPlayedItemId', streamInfo.item.Id);
+            }
+
             self._playNextAfterEnded = true;
             const isFirstItem = playOptions.isFirstItem;
             const fullscreen = playOptions.fullscreen;
--- a/src/controllers/playback/video/index.js
+++ b/src/controllers/playback/video/index.js
@@ -1620,6 +1620,37 @@ export default function (view) {
         }
     }
 
+    function isPageReloaded() {
+        // Detects if the current page load was a result of a reload.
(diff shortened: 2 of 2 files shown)


gompa:
this seems to work for series and movies, it doesn't auto play doh
i could see this be added for music too but that's a personal opinion

PR #6929, opened 2025-06-08
Remove indexing and accept a full array of albumartists into the func…

JohnCaveson:
…tion

**Changes**
in cardBuilder.js, pass `item.AlbumArtists` instead of one AlbumArtist.
in itemHelper.js, check for array and length on AlbumArtists. if > 1 map all the names and display with a `/` between each otherwise display the singular one

**Issues**

Fixes [14183](https://github.com/jellyfin/jellyfin/issues/14183)

### DIFF
--- a/src/components/itemHelper.js
+++ b/src/components/itemHelper.js
@@ -16,6 +16,8 @@ export function getDisplayName(item, options = {}) {
         throw new Error('null item passed into getDisplayName');
     }
 
+    console.debug('getDisplayName', item);
+
     if (item.Type === 'Timer') {
         item = item.ProgramInfo || item;
     }
@@ -50,6 +52,12 @@ export function getDisplayName(item, options = {}) {
         if (number) {
             name = name ? (number + nameSeparator + name) : number;
         }
+    }
+
+    if (item instanceof Array && item.length > 1) {
+        return item.map(i => getDisplayName(i, options)).join(' / ');
+    } else if (item.length === 1) {
+        return item[0].Name;
     }
 
     return name;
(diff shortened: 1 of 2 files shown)


whyman:
Bump on this, its a small patch but nice improvement.

@thornbill / @nielsvanvelzen apologies for the ping, I realise everyone is busy, but is there anything needed to move it along?

src/components/itemHelper.js
```diff
@@ -52,6 +52,12 @@ export function getDisplayName(item, options = {}) {
         }
     }
 
+    if (item instanceof Array && item.length > 1) {
+        return item.map(i => getDisplayName(i, options)).join(' / ');
+    } else if (item.length === 1) {
+        return item[0].Name;
+    }
```
thornbill:
```suggestion
    if (Array.isArray(item) {
        if (item.length > 1) {
            return item.map(i => getDisplayName(i, options)).join(' / ');
        } else if (item.length === 1) {
            return item[0].Name;
        }
    }
```

thornbill (changes requested):

src/components/itemHelper.js
```diff
@@ -52,6 +52,14 @@ export function getDisplayName(item, options = {}) {
         }
     }
 
+    if (Array.isArray(item) {
```
thornbill:
```suggestion
    if (Array.isArray(item)) {
```

thornbill (approved):

csaavedra:
Isn't a bit risky to merge a change like this a few days before making a stable release? Shouldn't this have waited until after the major release that 10.11.0 was?

thornbill:
It was a 10 line change to fix a bug. I accepted the risk and have already fixed a related issue for 10.11.1.

PR #6930, opened 2025-06-08
chore(library): Add configurations for minimum collection size

darioackermann:
Jellyfin Frontend changes to allow an arbitrary amount of movies required for a collection to be formed (at least 1)

grafik

The default value corresponds to the previously hardcoded value of 2.
Counterpart of https://github.com/jellyfin/jellyfin/pull/14250

Is the relevant translation automatically flagged for adjustment or how is this done?


darioackermann:
not stale

PR #6962, opened 2025-06-16
Fix split idiom shutdown -> shut down

systemcrash:
**Changes**
Fix split idiom shutdown -> shut down in en-us and en-gb


thornbill:
Only changes to the default language (en-us) can be committed here. All other languages are managed via our weblate instance at translate.jellyfin.org.

thornbill:
Looking through the translations, I see three other usages of "shutdown"... Is there any reason to not correct the others?

systemcrash:
Fair. I only went for the string I saw in the GUI. Only one other is a verbal phrase (where you would not say "I am shutdowning the server").

thornbill (approved):

PR #6965, opened 2025-06-18
Fix ends at not always accounting for playback position

### LINKED ISSUE #6964: "Ends At" time on details page doesn't account for playback position

### Describe the bug

The ends at time on the details page doesn't account for playback position.
This means that when you click on an item in the "continue watching" section, it shows its end time as if you hadn't started watching it.

### Reproduction Steps

1. Partially watch a tv show or movie
2. Go back to the main page and note that the progress bar shows that you have partially watched it
Image
3. Click on the thumbnail and note that the "Ends At" time doesn't account for what you've already watched
Image
4. Resume watching the episode/movie and note that the "Ends At" in the player is different
Image

### Expected/Actual behaviour

I expect the "Ends At" time to account for what I've already watched and it should match the "Ends At" time in the player.

### Logs

_No response_

### Server version

10.10.7

### Web version

10.10.7 (f4b8aa0ed4)

### Build version

Release

### Platform

Linux

### Browser

All versions

### Additional information

_No response_

johnnyg:
Update all "Ends At" times to use the same function which takes playback position into account.
Note that the "Ends At" time doesn't always account for playback rate however this can be addressed in another PR.

**Changes**
Update all the "Ends At" code to use `getEndsAtFromPosition`

**Issues**
Fixes #6964

### DIFF
--- a/src/components/mediainfo/EndsAt.tsx
+++ b/src/components/mediainfo/EndsAt.tsx
@@ -16,10 +16,7 @@ const EndsAt: FC<EndsAtProps> = ({ runTimeTicks, positionTicks, className }) => 
         className
     );
 
-    const playbackRate = 1;
-    const includeText = true;
-
-    const displayTime = mediainfo.getEndsAtFromPosition(runTimeTicks, positionTicks, playbackRate, includeText);
+    const displayTime = mediainfo.getEndsAtFromPosition(runTimeTicks, positionTicks, 1, true);
 
     return (
         <Box className={cssClass}>
--- a/src/components/mediainfo/mediainfo.js
+++ b/src/components/mediainfo/mediainfo.js
@@ -322,9 +322,7 @@ export function getMediaInfoHtml(item, options = {}) {
 export function getEndsAt(item) {
     if (item.MediaType === 'Video' && item.RunTimeTicks && !item.StartDate) {
         const positionTicks = item.UserData?.PlaybackPositionTicks;
-        const playbackRate = 1;
-        const includeText = true;
-        return getEndsAtFromPosition(item.RunTimeTicks, positionTicks, playbackRate, includeText);
+        return getEndsAtFromPosition(item.RunTimeTicks, positionTicks, 1, true);
     }
 
     return null;


src/components/mediainfo/mediainfo.js
```diff
@@ -321,11 +321,10 @@ export function getMediaInfoHtml(item, options = {}) {
 
 export function getEndsAt(item) {
     if (item.MediaType === 'Video' && item.RunTimeTicks && !item.StartDate) {
-        let endDate = new Date().getTime() + (item.RunTimeTicks / 10000);
-        endDate = new Date(endDate);
-
-        const displayTime = datetime.getDisplayTime(endDate);
-        return globalize.translate('EndsAtValue', displayTime);
+        const positionTicks = item.UserData?.PlaybackPositionTicks ?? 0;
```
dmitrylyzo: You can omit `?? 0` since `getEndsAtFromPosition` has `|| 0`.
https://github.com/jellyfin/jellyfin-web/blob/18f3083e69e17634c52ff842e003b1b01c17fcb0/src/components/mediainfo/mediainfo.js#L335

If you want a defined value, use `|| 0` since it more portable (_legacy browsers_) and will produce the same result.

src/components/mediainfo/EndsAt.tsx
```diff
@@ -1,28 +1,29 @@
...
         className
     );
 
-    const endTime = new Date().getTime() + (runTimeTicks / 10000);
-    const endDate = new Date(endTime);
-    const displayTime = datetime.getDisplayTime(endDate);
+    const playbackRate = 1;
+    const includeText = true;
+
+    const displayTime = mediainfo.getEndsAtFromPosition(runTimeTicks, positionTicks, playbackRate, includeText);
```
thornbill:
```suggestion
    const displayTime = mediainfo.getEndsAtFromPosition(runTimeTicks, positionTicks, 1, true);
```

src/components/mediainfo/mediainfo.js
```diff
@@ -321,11 +321,10 @@ export function getMediaInfoHtml(item, options = {}) {
...
     if (item.MediaType === 'Video' && item.RunTimeTicks && !item.StartDate) {
-        let endDate = new Date().getTime() + (item.RunTimeTicks / 10000);
-        endDate = new Date(endDate);
-
-        const displayTime = datetime.getDisplayTime(endDate);
-        return globalize.translate('EndsAtValue', displayTime);
+        const positionTicks = item.UserData?.PlaybackPositionTicks;
+        const playbackRate = 1;
+        const includeText = true;
+        return getEndsAtFromPosition(item.RunTimeTicks, positionTicks, playbackRate, includeText);
```
thornbill:
```suggestion
        return getEndsAtFromPosition(item.RunTimeTicks, positionTicks, 1, true);
```

johnnyg:
Assignments were purely for readability so happy to remove them (which I've done).

PR #6969, opened 2025-06-23
Fix issues fix the filter menu

Eyt-Lev:
**Changes**
now when you select the "played" filter, the "unplayed" is unchecked and vice versa,
also added the same functionality in the series status filters.

**Issues**
before, you could select both filters and experience buggy behavior when the applied filters didn't match the UI (when you are filtering for items that are both played and unplayed you expect to get nothing, but that's not what you get)


Eyt-Lev:
can this be merged? it's just 9 lines of code that fix annoying buggy behaviour

Eyt-Lev:
Really?

Eyt-Lev:
thanks for the review, how do we merge this?

thornbill:
I don't think this change makes any sense from a UX perspective. You are making checkboxes act in a non-standard way because you were not aware that the filters are an `OR` operation instead of an `AND`. I'm not saying that the current behavior is the best, but this change doesn't help imo.

On a more general, best practices note, avoid merging `master` to feature branches unless there is a conflict. It invalidates any prior review(s) and just generates noise on the PR.

PR #6974, opened 2025-06-25
Fix checkbox toggle on controller

### LINKED ISSUE #6973: Toggling a checkbox on a controller toggles twice.

### Describe the bug

Toggling a checkbox on controller toggles it twice. This leads to difficulties changing settings when relying on a controller.

### Reproduction Steps

1. Settings->Display->Display mode. Set to TV.
2. Settings->Controls->Enable gamepad support=On
3. Reload page
4. Select a checkbox and press A on your controller.

### Expected/Actual behaviour

The checkbox should toggle as normal. Instead it toggles twice. Once on keydown, once on keyup.

### Logs

_No response_

### Server version

10.10.7

### Web version

10.10.7

### Build version

10.10.7

### Platform

Windows, Xbox

### Browser

Firefox, Edge, Edge UWP WebView2

### Additional information

_No response_

brad1111:
Currently toggling a checkbox on a controller causes it to be triggered twice due to a keydown event being handled by the checkbox, and gamepadtokey triggering a click on keyup.

**Changes**
- Now ignores keyDown event of a controller if gamepads have been enabled in settings.

**Issues**
Fixes #6973

### DIFF
--- a/src/scripts/gamepadtokey.js
+++ b/src/scripts/gamepadtokey.js
@@ -81,11 +81,11 @@ _ButtonPressedState.setgamepadA = function (newPressedState) {
     const newPressedEvent = raiseKeyEvent(_gamepadAPressed, newPressedState, _GAMEPAD_A_KEY, _GAMEPAD_A_KEYCODE, false, true, _gamepadADownEvent);
     _gamepadAPressed = newPressedState;
 
-    if (newPressedState && newPressedEvent != null) {
-        _gamepadADownEvent = newPressedEvent;
-    } else if (!newPressedState) {
+    if (!newPressedState) {
         // reset on keyup
         _gamepadADownEvent = null;
+    } else if (newPressedEvent != null) {
+        _gamepadADownEvent = newPressedEvent;
     }
 };
 
@@ -202,6 +202,7 @@ function raiseEvent(name, key, keyCode) {
     event.key = key;
     event.keyCode = keyCode;
     (document.activeElement || document.body).dispatchEvent(event);
+
     return event;
 }
 
@@ -243,6 +244,7 @@ function raiseKeyEvent(oldPressedState, newPressedState, key, keyCode, enableRep
             clickElement(document.activeElement || window);
         }
     }
+
     return newPressedEvent;
 }
 


thornbill:
@brad1111 is this change still needed? I wasn't sure if your other PR fixed this also. :sweat_smile:

brad1111:
This fix is still useful for anyone using the gamepadtokey controller support. I.e. controller support on any web browser that isn't UWP based.

src/elements/emby-checkbox/emby-checkbox.js
```diff
@@ -6,6 +7,11 @@ import 'webcomponents.js/webcomponents-lite';
 const EmbyCheckboxPrototype = Object.create(HTMLInputElement.prototype);
 
 function onKeyDown(e) {
+    // Ignore GamepadA onKeyDown if using gamepadtokey since it will click the checkbox onKeyUp.
+    if (appSettings.enableGamepad && e.key === 'GamepadA') {
+        return;
+    }
```
dmitrylyzo: `enableGamepad` is a function, but you are just checking if it exists (always `true`).

dmitrylyzo: Doesn't it click anyway by this code?
https://github.com/jellyfin/jellyfin-web/blob/5046b1df0ff862de822ded654a405a913c76f940/src/scripts/keyboardNavigation.js#L212-L214

brad1111: Because gamepadtokey at some point has been modified to emulate keyboard inputs, it acts as if it has sent the enter key.
https://github.com/jellyfin/jellyfin-web/blob/5046b1df0ff862de822ded654a405a913c76f940/src/scripts/gamepadtokey.js#L41-L50

If it had been handled as a GamepadA in keyboardnavigation it would actually cause a triple click because of those changes.

Essentially what is happening right now is: 
- gamepadtokey sends a keydown event with keyCode 13 and keyname 'GamepadA'.
- KeyboardNavigation ignores the keyname entirely, and based on keyCode it interprets it as Enter, therefore not handling it as GamepadA.
- Emby checkbox handles the keydown event, changing the state.
- On keyup, gamepadtokey clicks the focused element (the checkbox) because clickOnKeyUp is enabled for GamepadA, changing the state back.

Thinking about it now, it might be better to revert the keycodes in gamepadtokey to be the UWP Controller keycodes and disable clickOnKeyUp, that way it should have the same behaviour as on UWP.

dmitrylyzo: > Because gamepadtokey at some point has been modified to emulate keyboard inputs, it acts as if it has sent the enter key.

Yes, it's hard to say if it was intended.

> If it had been handled as a GamepadA in keyboardnavigation it would actually cause a triple click because of those changes.

:+1: 

> Thinking about it now, it might be better to revert the keycodes in gamepadtokey to be the UWP Controller keycodes and disable clickOnKeyUp, that way it should have the same behaviour as on UWP.

For now, I'd think of `gamepadtokey` as JoyToKey application - treat the gamepad as a keyboard. We just need to make it work the same way.

1. We could try to ignore the default action (click) on keyup if the default action of the previous keydown was prevented. _This is probably preferable._

2. We could try to change `clickonKeyUp` block to `clickonKeyDown` (move it to the keydown block), but we still shouldn't perform click if the default action was prevented. _This is probably easier._

brad1111:
I resolved this based on ignoring clickonkeyup if keydown or keyup have been prevented as suggested. I agree with the bots assessment of the code being a bit hard to follow.

I'm thinking that removing the check for keycode existing would be a nice way to simplify it, since keycode is always provided by gamepadtokey.

src/scripts/gamepadtokey.js
```diff
@@ -219,18 +224,24 @@ function raiseKeyEvent(oldPressedState, newPressedState, key, keyCode, enableRep
...
+            newPressedEvent = raiseEvent('keyup', key, keyCode);
         }
-        if (clickonKeyUp) {
+        if (clickonKeyUp && !oldPressedEvent?.defaultPrevented && !newPressedEvent.defaultPrevented) {
             clickElement(document.activeElement || window);
         }
+
+        if (newPressedEvent) {
+            return newPressedEvent;
+        }
```
dmitrylyzo: I think we can return `null` here.
```suggestion
        return newPressedEvent;
```

We could make the function always return `newPressedEvent`:
```js
let newPressedEvent = null;

// No-op if oldPressedState === newPressedState
if (newPressedState === true) {
    ...
} else if (newPressedState === false && oldPressedState === true) {
    ...
}

return newPressedEvent;
```

src/scripts/gamepadtokey.js
```diff
@@ -219,18 +224,24 @@ function raiseKeyEvent(oldPressedState, newPressedState, key, keyCode, enableRep
...
         resetThrottle(key);
 
+        let newPressedEvent = null;
         // button up
         if (keyCode) {
-            raiseEvent('keyup', key, keyCode);
+            newPressedEvent = raiseEvent('keyup', key, keyCode);
         }
-        if (clickonKeyUp) {
+        if (clickonKeyUp && !oldPressedEvent?.defaultPrevented && !newPressedEvent.defaultPrevented) {
```
dmitrylyzo: `newPressedEvent` can be `null` if there's no `keyCode` or `!allowInput` (in `raiseEvent`).
```suggestion
        if (clickonKeyUp && !oldPressedEvent?.defaultPrevented && !newPressedEvent?.defaultPrevented) {
```

src/scripts/gamepadtokey.js
```diff
@@ -77,8 +78,11 @@ _ButtonPressedState.getgamepadA = function () {
 };
 
 _ButtonPressedState.setgamepadA = function (newPressedState) {
-    raiseKeyEvent(_gamepadAPressed, newPressedState, _GAMEPAD_A_KEY, _GAMEPAD_A_KEYCODE, false, true);
+    const newPressedEvent = raiseKeyEvent(_gamepadAPressed, newPressedState, _GAMEPAD_A_KEY, _GAMEPAD_A_KEYCODE, false, true, _gamepadADownEvent);
     _gamepadAPressed = newPressedState;
+    if (newPressedEvent != null) {
```
dmitrylyzo: Store keydown only.
```suggestion
    if (newPressedState && newPressedEvent != null) {
```
It might be a good idea to reset the stored event:
```js
if (newPressedState) {
    if (newPressedEvent != null) {
        _gamepadADownEvent = newPressedEvent;
    }
} else {
    _gamepadADownEvent = null;
}
```

src/scripts/gamepadtokey.js
brad1111: Implemented the second suggestion as I think it ends up looking cleaner.

src/scripts/gamepadtokey.js
brad1111: I think reseting the event sounds sensible so I chose that one. Slightly modified to compress the nested ifs into a single if/else if as I think it's easier to follow.

dmitrylyzo: > Slightly modified to compress the nested ifs into a single if/else if as I think it's easier to follow.

But `newPressedState` is checked twice.

src/scripts/gamepadtokey.js
```diff
@@ -77,8 +78,15 @@ _ButtonPressedState.getgamepadA = function () {
...
-    raiseKeyEvent(_gamepadAPressed, newPressedState, _GAMEPAD_A_KEY, _GAMEPAD_A_KEYCODE, false, true);
+    const newPressedEvent = raiseKeyEvent(_gamepadAPressed, newPressedState, _GAMEPAD_A_KEY, _GAMEPAD_A_KEYCODE, false, true, _gamepadADownEvent);
     _gamepadAPressed = newPressedState;
+
+    if (newPressedState && newPressedEvent != null) {
+        _gamepadADownEvent = newPressedEvent;
+    } else if (!newPressedState) {
+        // reset on keyup
+        _gamepadADownEvent = null;
+    }
```
dmitrylyzo: If you don't want nested conditions:
```suggestion
    if (!newPressedState) {
        // reset on keyup
        _gamepadADownEvent = null;
    } else if (newPressedEvent != null) {
        _gamepadADownEvent = newPressedEvent;
    }
```
This way `newPressedState` is tested once.

src/scripts/gamepadtokey.js
```diff
@@ -219,19 +230,20 @@ function raiseKeyEvent(oldPressedState, newPressedState, key, keyCode, enableRep
...
         if (keyCode) {
-            raiseEvent('keyup', key, keyCode);
+            newPressedEvent = raiseEvent('keyup', key, keyCode);
         }
-        if (clickonKeyUp) {
+        if (clickonKeyUp && !oldPressedEvent?.defaultPrevented && !newPressedEvent?.defaultPrevented) {
             clickElement(document.activeElement || window);
         }
     }
+    return newPressedEvent;
```
dmitrylyzo: _nitpicking_
Just to "beautify" the code. Iirc, it was recommended to leave a blank line after the `if/else` block.
```suggestion
    }

    return newPressedEvent;
```

src/scripts/gamepadtokey.js
```diff
@@ -194,6 +202,7 @@ function raiseEvent(name, key, keyCode) {
     event.key = key;
     event.keyCode = keyCode;
     (document.activeElement || document.body).dispatchEvent(event);
+    return event;
```
dmitrylyzo: _nitpicking_
Same - to split "dispatch" and "return" activities.
```suggestion
    (document.activeElement || document.body).dispatchEvent(event);

    return event;
```

src/scripts/gamepadtokey.js
brad1111: Sorry, I forgot to respond to the previous comment. I was going to suggest this as an alternative, so I've accepted all the changes.

PR #6978, opened 2025-06-26
Catch the exception when the user has no access to the epgs

### LINKED ISSUE #6977: Uncaught exception when get into the detail page of a show

### Describe the bug

The uncaught exception occurs because the client is unable to get /LiveTv/Programs and server returns 403. After finding similar issues, i found one (https://github.com/jellyfin/jellyfin/issues/11089) , and the solution of the issue is to grant the user the permission of accessing Live TV. But i wanna eliminate such message without granting the permission. So i tried to catch the exception.

### Reproduction Steps

1. Revoke the permission of a user (https://jellyfin.org/docs/general/server/users/adding-managing-users/#feature-access)
2. go back to home, and launch the Devtools of the browser
3. click a show to present its detail
4. exception occurs

### Expected/Actual behaviour

It's sensible to catch the exception and then do the same thing as there is no available live tv epgs, but now it prints the response to the console.

### Logs

Image

### Server version

10.10.7

### Web version

10.11.0-rc2

### Build version

Dev Server

### Platform

Windows

### Browser

Chrome

### Additional information

Image

I have tried to make minor modification to the source, later i will file a pr

To be honest, i think this is an enhancement more than a bug, but i don't think it should be filed as a feature, so i still file it here. Sorry for any inconvenience caused!

tobelogin:
Catch exception when the user has no access to tv epgs

**Changes**
When the server returns 403, the client will do as there is 0 items in response

**Issues**
Fixes #6977

### DIFF
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -1631,6 +1631,10 @@ function renderSeriesSchedule(page, item) {
         imageLoader.lazyChildren(scheduleTab);
 
         loading.hide();
+    }).catch(function (resp) {
+        if(resp.status === 403) {
+            page.querySelector('#seriesScheduleSection').classList.add('hide');
+        }
     });
 }
 


tobelogin:
Well, i don't know if there should be loading.hide() at the end of catch block, since everything works so well before i catch the exception

tobelogin:
Sorry, i'm a github newcomer, if there is anything i should do

PR #6979, opened 2025-06-26
Add fallback to 4K resolution if window size is infinite

### LINKED ISSUE #6415: Primary Images cannot be loaded and stay blurry on webOS
**Describe The Bug**
I use the webOS Jellyfin client on my LG TV.
When I open the the app, all images are blurred.
After reconnecting to the server (profile -> change server -> connect) the images are loaded correctly.
After turning off the TV and turning it on again, the images stay blurred again.

My nginx proxy server shows the following requests for the images:
`/Items/8db0/Images/Primary?fillHeight=Infinity&fillWidth=Infinity&quality=96&tag=bea50` (I shortened the hashes)

The response is as follows:
```
{
	"type": "https://tools.ietf.org/html/rfc9110#section-15.5.1",
	"title": "One or more validation errors occurred.",
	"status":400,
	"errors": {
		"fillWidth": ["The value \u0027Infinity\u0027 is not valid."],
		"fillHeight": ["The value \u0027Infinity\u0027 is not valid."]
	},
	"traceId": "00-b268cbb3a4e5f45a30e9f50e4c9c572d-b5103b9c688405f0-00"
}
```

This issue only occurs on the webOS client. All other clients work fine.

**Steps To Reproduce**

1. Turn off the TV (webOS 04.06.75)
2. Turn on the TV and start the Jellyfin app (1.2.2)
4. All images are blurred

**Expected Behavior**
All images are loaded correctly.

**Workaround**
Reconnecting to the server fixes the issue until the TV is turned off and on again.

Another workaround that works for me is to remove the invalid URL parameters in the Nginx proxy server before passing it to Jellyfin.
```
location /Items {
    if ($args ~* (.*)(fillHeight=Infinity&fillWidth=Infinity&?)(.*)) {
        set $args $1$3;
    }
}
```

**Logs**
`/Items/xyz/Images/Primary?fillHeight=Infinity&fillWidth=Infinity`

 - Platform: Linux (Unraid 6.12.13)
 - Browser: webOS (04.06.75) Jellyfin client (1.2.2)
 - Jellyfin Version: 10.10.3

thornbill:
**Changes**
* Moves the dom utils to the utils directory
* Adds a fallback to use 4K dimensions if `window` reports an infinite screen size (reported issue for webOS)

Marking as a draft until confirmed

**Issues**
Fixes #6415 

## Summary by CodeRabbit

* **Refactor**
  * Updated internal references to a utility module across the application for improved consistency and maintainability.
  * Enhanced documentation formatting for better readability.
  * Improved handling of window size calculations to address rare edge cases on certain devices.
  
No changes to user-facing features or functionality.

### DIFF
--- a/src/apps/dashboard/features/sessions/utils/getNowPlayingImageUrl.ts
+++ b/src/apps/dashboard/features/sessions/utils/getNowPlayingImageUrl.ts
@@ -1,7 +1,7 @@
 import type { BaseItemDto } from '@jellyfin/sdk/lib/generated-client/models/base-item-dto';
 import { ImageType } from '@jellyfin/sdk/lib/generated-client/models/image-type';
 import { ServerConnections } from 'lib/jellyfin-apiclient';
-import dom from 'scripts/dom';
+import dom from 'utils/dom';
 
 const getNowPlayingImageUrl = (item: BaseItemDto) => {
     if (!item.ServerId) return null;


FranzDeschler:
I didn´t test the fix yet but it looks promising.
But I would like to suggest further changes:
Only save the window size in the `windowSize` variable, if it is not `Infinite`.

The problem is - as I was able to inspect it so far - that the first time, the window size is requested - and it returns `Infinite` - the value is cached in the `windowSize` variable. So all further calls of `getWindowSize()` result in `Infinite`.
In your solution, the function would always return 4K. It surely fixes the original bug, but is not the correct value for non-4K screens.

Simply put:
If `innerWidth` or `innerHeight` is `Infinite`, return 4K but **don´t** cache the value in the `windowSize` variable. Otherwise, write it in the cache-variable.

This attempt means that not all, but only some calls of `getWindowSize()` result in an incorrect value.

thornbill:
Good call @FranzDeschler! I have updated the code so the fallback value will not be saved.

src/components/activitylog.js
```diff
@@ -2,7 +2,7 @@ import escapeHtml from 'escape-html';
 import Events from '../utils/events.ts';
 import globalize from '../lib/globalize';
 import { ServerConnections } from 'lib/jellyfin-apiclient';
-import dom from '../scripts/dom';
+import dom from '../utils/dom.js';
```
IDisposable: _nit_ all the other places we don't have the `.js` on the import.

FranzDeschler:
I was now able to test your solution on my TV but it does NOT fix the issue (yet)!
I had to make one little adjustment:

I had to replace
`const { innerWidth, innerHeight } = window;`
with
```
const innerWidth = window.innerWidth;
const innerHeight = window.innerHeight;
```
and that finally fixed the issue.

I don´t know why this happens, but taking a look in the server logs, I was able to validate that the issue still exists with your current solution. The logs still showed
`... /Images/Primary?fillHeight=Infinity&fillWidth=Infinity ...`

After my adjustment, the query parameters were correct. No `Infinity` in the logs anymore.

src/components/activitylog.js
thornbill: I really don't know why VS code occasionally adds this when files are moved :man_facepalming:

thornbill:
@coderabbitai review

thornbill:
@coderabbitai review

thornbill:
@coderabbitai review

FranzDeschler (approved):
Looks good to me.

PR #6993, opened 2025-06-29
Use keycodes for UWP controller instead of gamepadtokey.

brad1111:
**Changes**
Adds keycodes that Xbox controllers produce in UWP WebViews to KeyboardNavigation. Force disables gamepadtokey on Xbox so that key events aren't duplicated.

**Considerations**
- I chose to add all the different names for the controller buttons but it would equally work if they were called by the existing key names.
- I chose to include keycodes that are not currently used on Xbox. This makes it easier to test Jellyfin-UWP on desktop, and it covers the case where WebView 2 goes back to using the documented keycodes.
- I included NavigationUp/Down/Left/Right, but am not 100% sure what it is used for since that isn't documented. I included it for completions sake since the TVJS sample code includes it. I suspect it might be for the Xbox media remote.

### DIFF
--- a/src/scripts/keyboardNavigation.js
+++ b/src/scripts/keyboardNavigation.js
@@ -67,19 +67,19 @@ const KeyNames = {
 
 const KeyAliases = {
     // GamepadA needs special case handling
-    'GamepadB': 'Escape',
-    'NavigationUp': 'ArrowUp',
-    'NavigationDown': 'ArrowDown',
-    'NavigationLeft': 'ArrowLeft',
-    'NavigationRight': 'ArrowRight',
-    'GamepadDPadUp': 'ArrowUp',
-    'GamepadDPadDown': 'ArrowDown',
-    'GamepadDPadLeft': 'ArrowLeft',
-    'GamepadDPadRight': 'ArrowRight',
-    'GamepadLeftThumbUp': 'ArrowUp',
-    'GamepadLeftThumbDown': 'ArrowDown',
-    'GamepadLeftThumbLeft': 'ArrowLeft',
-    'GamepadLeftThumbRight': 'ArrowRight'
+    GamepadB: 'Escape',
+    NavigationUp: 'ArrowUp',
+    NavigationDown: 'ArrowDown',
+    NavigationLeft: 'ArrowLeft',
+    NavigationRight: 'ArrowRight',
+    GamepadDPadUp: 'ArrowUp',
+    GamepadDPadDown: 'ArrowDown',
+    GamepadDPadLeft: 'ArrowLeft',
+    GamepadDPadRight: 'ArrowRight',
+    GamepadLeftThumbUp: 'ArrowUp',
+    GamepadLeftThumbDown: 'ArrowDown',
+    GamepadLeftThumbLeft: 'ArrowLeft',
+    GamepadLeftThumbRight: 'ArrowRight'
 };
 
 /**
@@ -297,6 +297,6 @@ if (navigator.getGamepads && appSettings.enableGamepad() && canEnableGamepad()) 
 export default {
     enable: enable,
     getKeyName: getKeyName,
-    isNavigationKey: isNavigationKey,
-    canEnableGamepad: canEnableGamepad
+    isNavigationKey,
+    canEnableGamepad
 };


JPVenson (approved):

src/elements/emby-select/emby-select.js
```diff
@@ -79,7 +79,8 @@ function onMouseDown(e) {
 }
 
 function onKeyDown(e) {
-    if (e.keyCode === 13 && !enableNativeMenu()) {
+    // Xbox controller for UWP WebView2 uses keycode 195 to select.
+    if (e.keyCode === 13 && !enableNativeMenu() || (e.keyCode === 195 && browser.edgeChromium)) {
```
dmitrylyzo: Previously, it used native menu:
https://github.com/jellyfin/jellyfin-web/blob/a7d443ab231a1666c2678ee2f75c802ab5c36f52/src/elements/emby-select/emby-select.js#L10-L12
Now it always uses a custom implementation.

Doesn't UWP open its own menu?

Also, `browser.edgeChromium` or `(browser.edgeUwp || browser.xboxOne)`?

src/scripts/keyboardNavigation.js
```diff
@@ -145,27 +173,43 @@ export function enable() {
         let capture = true;
 
         switch (key) {
+            case 'NavigationLeft':
+            case 'GamepadDPadLeft':
+            case 'GamepadLeftThumbLeft':
```
dmitrylyzo: Do we want to use both the left thumb and the D-pad for navigation?

src/scripts/keyboardNavigation.js
```diff
@@ -145,27 +173,43 @@ export function enable() {
...
+            case 'NavigationDown':
+            case 'GamepadDPadDown':
+            case 'GamepadLeftThumbDown':
             case 'ArrowDown':
                 inputManager.handleCommand('down');
                 break;
 
+            case 'GamepadA':
+                inputManager.handleCommand('select');
+                break;
```
dmitrylyzo, line 214: Doesn't UWP perform a "click" on its own?

src/elements/emby-slider/emby-slider.js
```diff
@@ -527,12 +527,18 @@ function stepKeyboard(elem, delta) {
      */
 function onKeyDown(e) {
     switch (keyboardnavigation.getKeyName(e)) {
+        case 'NavigationLeft':
+        case 'GamepadDPadLeft':
+        case 'GamepadLeftThumbLeft':
```
dmitrylyzo: Technically, we can map all `*Left/Right/Up/Down` to `ArrowLeft/Right/Up/Down` in `keyboardnavigation.getKeyName` (if we don't need to know the source of the event) to simplify the event handling.

src/scripts/keyboardNavigation.js
brad1111: With the virtual mouse it does, however this is to allow us to not need the virtual mouse.
It doesn't if the virtual mouse is disabled.
Maybe behaviour was different in UWP WebView 1, or you are thinking of the click event caused by gamepadtokey?

src/scripts/keyboardNavigation.js
brad1111: It would be very unusual for an Xbox app to not respond to both Left Thumbstick and Dpad.

thornbill:
We should hide the gamepad setting... or the entire "controls" section since the setting is ignored.

src/scripts/keyboardNavigation.js
brad1111: Just reread this, did you mean combining these into one event, or either only accepting input from one of the two?
I interpreted it as the latter.

src/elements/emby-slider/emby-slider.js
brad1111: If that would be the preferred way of doing things I can easily change that. Especially if it makes maintaining things easier. I'm not sure if we need the source either.

src/elements/emby-select/emby-select.js
brad1111: Honestly it was just a way of working around it not providing a click event. Though personally I find the native menu a bit small on xbox. I'll attach a screenshot a later to show.

EdgeUWP currently doesn't detect WebView2 as it is based on Chromium Edge and not old Edge. I'm not sure whether to change that so EdgeUWP detects WebView2 instead, detects both, or have a separate EdgeUWPWebView2 detection thing. 
Your suggestion would currently work on Xbox, but not on Desktop which makes testing more difficult.

brad1111: This is what it looks like on Xbox. Screenshot_2025-06-30_14-01-46

However I was only able to get this to show in a remote debugger with a mouse click. Even worse once it's shown it doesn't respond to anything other than analog stick, even then it's scrolling but not selecting. Native select with WebView2 on Xbox doesn't work.

src/scripts/keyboardNavigation.js
dmitrylyzo: > I interpreted it as the latter.

Yes, I meant to use one of the two.

> It would be very unusual for an Xbox app to not respond to both Left Thumbstick and Dpad.

If this is the standard behavior, then I'm fine with handling both.

src/elements/emby-slider/emby-slider.js
dmitrylyzo: I was thinking of wrapping the keydown/up handling with some kind of keybinding system that allows the user to change keyboard shortcuts in the `Controls` settings. It should produce "commands" depending on the context (video player, audio player, book player, general use). I haven't done anything yet, just thoughts.

At the moment, our shortcuts are inconsistent (video player, audio player, book player).

Making `ArrowLeft/...` as an alias for `ArrowLeft/...`, `NavigationLeft/...`, `GamepadDPadLeft/...`, `GamepadLeftThumbLeft/...` will reduce the number of changes.

src/elements/emby-select/emby-select.js
dmitrylyzo: If you want to use the custom menu on XBox, you should adjust the `enableNativeMenu` accordingly.

src/scripts/keyboardNavigation.js
```diff
@@ -20,6 +20,36 @@ const KeyNames = {
     38: 'ArrowUp',
     39: 'ArrowRight',
     40: 'ArrowDown',
+
+    // UWP WebView section start --
+    // Navigation Up/Down/Left/Right is part of TVJS directionalnavigation-1.0.0.0.js
+    // Unsure what this is used for. Media remote?
+    138: 'ArrowUp',
```
dmitrylyzo: The `keyCode` property is deprecated. Therefore, if it is missing (unlikely, but...), `getKeyName` will return the `key` property which is `Navigation*`, `GamepadLeftThumb*` or `GamepadDPad*`.

I was thinking about introducing a key mapping and using it in `getKeyName` like this:
```js
const KeyAliases = {
    'NavigationLeft': 'ArrowLeft',
    'GamepadLeftThumbLeft': 'ArrowLeft',

    ...

    // Not sure, as it would probably conflict with the default behavior
    //'GamepadA': 'Enter',

    'GamepadB': 'Escape'
};

function getKeyName(event) {
    const key = KeyNames[event.keyCode] || event.key;
    return KeyAliases[key] || key;
}
```

src/controllers/user/controls/index.js
```diff
@@ -21,6 +22,11 @@ export default function (view) {
     view.addEventListener('viewshow', function () {
         view.querySelector('.smoothScrollContainer').classList.toggle('hide', !layoutManager.tv);
 
+        if (browser.edgeUwp) {
+            view.querySelector('.enableGamepadContainer').classList.add('hide');
```
dmitrylyzo: Use `classList.toggle` when possible.

src/elements/emby-slider/emby-slider.js
brad1111: I’ve consolidated these now

src/components/apphost.js
```diff
@@ -265,7 +265,11 @@ const supportedFeatures = function () {
     }
 
     features.push(AppFeature.DisplayLanguage);
-    features.push(AppFeature.DisplayMode);
+
+    if (!browser.xboxOne) {
+        features.push(AppFeature.DisplayMode);
+    }
```
dmitrylyzo: This check is unnecessary for apps (have `NativeShell`), unless you want to pin TV layout ([default for XBox](https://github.com/jellyfin/jellyfin-web/blob/855276e8afe0ef42d4754918d9e8c6d647ff9f65/src/components/layoutManager.js#L57-L59)) for native browser.

src/scripts/keyboardNavigation.js
brad1111: Oh, so should this PR wait on those changes then?
In terms of keyCode though, WebView2 only provides keyCode and not the named key.

dmitrylyzo: > Oh, so should this PR wait on those changes then?

No, I thought you would do that. 😅

> In terms of keyCode though, WebView2 only provides keyCode and not the named key.

Then fill `KeyNames` with the correct names for consistency. And add `KeyAliases` to remap keys - this will be easier to revert.

src/components/apphost.js
brad1111: Ah I was unaware of NativeShell. I assume this needs to be resolved on the UWP side then.

src/scripts/keyboardNavigation.js
brad1111: I can be really dopey sometimes, sorry about that. Added it.

src/controllers/user/controls/index.js
```diff
@@ -20,7 +21,7 @@ export default function (view) {
 
     view.addEventListener('viewshow', function () {
         view.querySelector('.smoothScrollContainer').classList.toggle('hide', !layoutManager.tv);
-
+        view.querySelector('.enableGamepadContainer').classList.toggle('hide', browser.edgeUwp);
```
dmitrylyzo: Just to separate hiding fields and restoring their values.
```suggestion
        view.querySelector('.enableGamepadContainer').classList.toggle('hide', browser.edgeUwp);

```
I would also sort them alphabetically.

dmitrylyzo: It would probably be a good idea to extract `browser.edgeUwp` into a `canEnableGamepad` function (in `keyboardNavigation`) or something like that. So we can use it here, in the Settings page and in  `keyboardNavigation` to eliminate duplication.

src/scripts/keyboardNavigation.js
```diff
@@ -40,11 +65,27 @@ const KeyNames = {
...
+    'GamepadLeftThumbDown': 'ArrowDown',
+    'GamepadLeftThumbLeft': 'ArrowLeft',
+    'GamepadLeftThumbRight': 'ArrowRight'
+};
+
 /**
  * Keys used for keyboard navigation.
  */
 const NavigationKeys = ['ArrowLeft', 'ArrowRight', 'ArrowUp', 'ArrowDown', 'BrowserHome', 'Find'];
-
```
dmitrylyzo: Please return the blank line.

dmitrylyzo (commented):
Can't test if it works on Xbox, but LGTM. Just nitpicking.

brad1111:
> Can't test if it works on Xbox, but LGTM. Just nitpicking.

I've tried to test it as best as thouroughly as I can on Xbox on both my WIP PR https://github.com/jellyfin/jellyfin-uwp/pull/64 to enable controller support, and the current version of Jellyfin on the store.

There are a couple things that don't work perfectly yet with my build:
- The QuickConnect page in settings is unusable with my build, however this also seems to be the case on my Hisense TV and in the WebOS simulator.
- A user can change the layout of the app away from TV breaking controller input. This is because the UWP app does not yet implement NativeShell.

With my build I tested:
- All the settings pages, and controls
- Video pages and playback
- Audio pages and playback
- Collections pages
- Books
- Login with QuickConnect and manually

With the current build I ensured the cursor still exists, and that keydown events don't interfere with the settings controls or video playback (seems like they are disabled if the virtual mouse is enabled)

If you wanted to and have a Windows 10/11 computer and a controller you could try my build, even if the behaviour is not identical it should be mostly similar.

src/scripts/keyboardNavigation.js
```diff
@@ -40,6 +65,23 @@ const KeyNames = {
...
+    'NavigationLeft': 'ArrowLeft',
+    'NavigationRight': 'ArrowRight',
+    'GamepadDPadUp': 'ArrowUp',
+    'GamepadDPadDown': 'ArrowDown',
+    'GamepadDPadLeft': 'ArrowLeft',
+    'GamepadDPadRight': 'ArrowRight',
+    'GamepadLeftThumbUp': 'ArrowUp',
+    'GamepadLeftThumbDown': 'ArrowDown',
+    'GamepadLeftThumbLeft': 'ArrowLeft',
+    'GamepadLeftThumbRight': 'ArrowRight'
```
thornbill: The quotes around the object keys are not necessary:

```suggestion
    GamepadB: 'Escape',
    NavigationUp: 'ArrowUp',
    NavigationDown: 'ArrowDown',
    NavigationLeft: 'ArrowLeft',
    NavigationRight: 'ArrowRight',
    GamepadDPadUp: 'ArrowUp',
    GamepadDPadDown: 'ArrowDown',
    GamepadDPadLeft: 'ArrowLeft',
    GamepadDPadRight: 'ArrowRight',
    GamepadLeftThumbUp: 'ArrowUp',
    GamepadLeftThumbDown: 'ArrowDown',
    GamepadLeftThumbLeft: 'ArrowLeft',
    GamepadLeftThumbRight: 'ArrowRight'
```

src/scripts/keyboardNavigation.js
```diff
@@ -239,12 +290,13 @@ function attachGamepadScript() {
...
+if (navigator.getGamepads && appSettings.enableGamepad() && canEnableGamepad()) {
     window.addEventListener('gamepadconnected', attachGamepadScript);
 }
 
 export default {
     enable: enable,
     getKeyName: getKeyName,
-    isNavigationKey: isNavigationKey
+    isNavigationKey: isNavigationKey,
+    canEnableGamepad: canEnableGamepad
```
thornbill: You can use the shorthand syntax for these:

```suggestion
    isNavigationKey,
    canEnableGamepad
```

thornbill (commented):
Just a couple syntax suggestions. Once those are resolved I think we should land this and see what feedback we get. :slightly_smiling_face:

PR #7001, opened 2025-07-03
Update global.d.ts

sakgoyal:
Improve the types for the `NativeShell` window object


sakgoyal:
squash and merge? I'd rather not force push if thats an option

PR #7003, opened 2025-07-03
Typescript Emby Components

sakgoyal:
* use class components instead of functions
* use proper custom web elements

* will this work on older platforms or does this require using the `webcomponents.js` package


sakgoyal:
https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_custom_elements

the examples always call `super` in the constructor. even if there is nothing else in there. so i'm not sure if it can be removed like eslint says, or it is required. 

they keycode deprecation has always been there. I didn't change that part

timreichen:
This won't work because safari does not support customized built-in elements: https://caniuse.com/custom-elementsv1 .

PR #7009, opened 2025-07-09
Fix unsorted results in Edit Images #7008

### LINKED ISSUE #7008: Unsorted results in Edit Images results

### Describe the bug

Going through the results of the "Edit Images" tool is cumbersome because the results are unsorted by defaults which leads to pages of irrelevant results.

Being able to see the number of votes is a good feature, but it is only halfway implemented if the user has to go through pages of random posters.

### Reproduction Steps

1. Open the "Edit Images" menu of any movie/show
2. Click on a magnifier icon
3. Observe that it is, indeed, not ordered. You are often forced to scroll down to find a relevant result.

See the images below for examples.

Image

Image

### Expected/Actual behaviour

I expect Jellyfin to show me the results in the correct order (most to least popular), or at least give me a similar option.

### Logs

_No response_

### Server version

10.10.2

### Web version

10.10.2

### Build version

10.10.2

### Platform

All platforms

### Browser

All browsers

### Additional information

I am planning to create a PR to correct this.

biskweet:
Going through the results of the "Edit Images" tool is cumbersome because the results are unsorted by defaults which leads to pages of irrelevant results.

Being able to see the number of votes is a good feature, but it is only halfway implemented if the user has to go through pages of random posters.

**Changes**
- Now sorts the results in descending order automatically

**Issues**
Fixes #7008

### DIFF
--- a/src/components/imageDownloader/imageDownloader.js
+++ b/src/components/imageDownloader/imageDownloader.js
@@ -61,6 +61,16 @@ function reloadBrowsableImages(page, apiClient) {
     }
 
     apiClient.getAvailableRemoteImages(options).then(function (result) {
+        result.Images.sort((a, b) => {
+            // Sort images from the most to the least popular
+            if (a.CommunityRating && b.CommunityRating) {
+                // Lazy eval will make it 0 if needed
+                return (b.VoteCount || 0) - (a.VoteCount || 0);
+            } else {
+                return 0;
+            }
+        });
+
         renderRemoteImages(page, apiClient, result, browsableImageType, options.startIndex, options.limit);
 
         page.querySelector('#selectBrowsableImageType').value = browsableImageType;


theguymadmax:
[Image priority ordering](https://github.com/jellyfin/jellyfin/blob/38d0367c428fdce7e42c8e70da11327ecbba2c6c/MediaBrowser.Model/Extensions/EnumerableExtensions.cs#L27): 
  - Images that match the requested language
  - Images with no language
  - Images in English
  - Images that don't match the requested language

Then by community rating and vote count.

The order from your issue appears to be incorrect, as it shows posters without titles first, since there are none in your requested language. This should be addressed on the server. Here's a recent attempt to do so: [#14236](https://github.com/jellyfin/jellyfin/pull/14236)

thornbill:
Yup this should be fixed server side 👍

biskweet:
@theguymadmax
> [Image priority ordering](https://github.com/jellyfin/jellyfin/blob/38d0367c428fdce7e42c8e70da11327ecbba2c6c/MediaBrowser.Model/Extensions/EnumerableExtensions.cs#L27):
> 
> * Images that match the requested language
> * Images with no language
> * Images in English
> * Images that don't match the requested language

This choice of ordering is, IMO, arguable. Generally speaking, the English poster is the international poster and considered the official or second-official one. I would argue that:
* "Images with no language" should definitely not come before English, and
* Even if one accepts the current ordering, the results should be sorted within each "category"

As someone who spends a lot of time tweaking the artworks (because of questionable defaults), I can tell how much time is spent at just comparing the number of votes to find the "(semi-)official" poster/logo/thumbnail.

Concerning the place of implementation (whether it should be server- or client-side), I understand your point though.

PR #7017, opened 2025-07-10
Add unified plugin page

thornbill:
**Changes**
Replaces the "My Plugins" and "Catalog" pages with a single unified plugin page that allows filtering based on installation status and category

Screenshot 2025-07-25 at 17-38-13 Plugins

**Issues**
N/A


sjorge:
Personal preference maybe but I feel like the it should default to showing installed plugins instead of all, usually when I visit the page I want to see if there is an update for a plugin or I want to make changing to the plugins configuration.

With it defaulting to all it's very busy and cluttered.

Edit:

Additionally hitting the cog to go to the repo configuration has no easy way to go back to the previous page. (Unless I am not seeing it)

thornbill:
@sjorge thanks for the feedback! I might do a poll to gather some feedback on defaulting to installed vs all.

https://github.com/jellyfin/jellyfin-web/pull/7024 adds the "back" button back to the dashboard so hopefully that helps with leaving the repositories page. :slightly_smiling_face:

nielsvanvelzen:
I also think the filter should be set to just the installed plugins by default. Additionally the installed/all filter should be separate from the category filter so I can query all my installed music plugins for example.

Managing the repositories using the the gear icon does not feel really intuitive to me. With the amount of room there is adding a label "manage repositories" should make it easier to find.

Otherwise looks great, merging the installed/catalog pages makes a lot of sense.

sjorge:
> #7024 adds the "back" button back to the dashboard so hopefully that helps with leaving the repositories page. 🙂

Yes, that works well :)

> @sjorge thanks for the feedback! I might do a poll to gather some feedback on defaulting to installed vs all.

I wonder if that will also speed up the initial load as you don't need to wait for the catalog files to download.

src/apps/dashboard/routes/plugins/index.tsx
```diff
@@ -1,44 +1,82 @@
...
-        isError: isConfigurationPagesError,
-        isPending: isConfigurationPagesPending
-    } = useConfigurationPages();
+        data: pluginDetails,
+        isError,
+        isPending
+    } = usePluginDetails();
+    const [ category, setCategory ] = useState<string>();
     const [ searchQuery, setSearchQuery ] = useState('');
+    const [ status, setStatus ] = useState<PluginStatusOption>(PluginStatusOption.Installed);
```
viown: Maybe we can use search params instead? To avoid always defaulting to installed when going back after clicking on a plugin.

theguymadmax:
Thanks for addressing all the feedback. It functions and looks great! 
One other nitpick. I think it would help if there were some visual separator between the two main sections. <br>

nielsvanvelzen:
The "View all plugins" link when there are no results should probably have 2 behaviors:

1. If a category filter is active only reset that one
2. If a installed/catalog filter is active without a category filter reset that one

src/apps/dashboard/routes/plugins/index.tsx
nielsvanvelzen: This would be my last nitpick on the feature. Adding two query parameters for the category & status would be nice to have.

thornbill: :+1: done

nielsvanvelzen (approved):

PR #7028, opened 2025-07-18
Add Play All button to movies

theguymadmax:
**Changes**
 Adds a Play All button to the movies library 

**Issues**
N/A


timreichen:
I think this button makes it kinda confusing next to the shuffle button.
It reminds me of an audio app that normally has a shuffle toggle and a play button to start a certain selection/album etc.
These two buttons do not what they imply in context to each other imo.

PR #7036, opened 2025-07-22
imageLoader: fix upscrolling showing unloaded images

evan314159:
**Changes**
When scrolling upwards images that have been unloaded as they went out of view show as blank boxes. Fix by not unloading images that have already been lazy-loaded.  Tested with Safari and Google Chrome on macOS.

**Issues**
No issue opened.  If you scroll upwards on any long page with images you will see images unload as you scroll up, leaving blank boxes.

### DIFF
--- a/src/components/images/imageLoader.js
+++ b/src/components/images/imageLoader.js
@@ -82,12 +82,8 @@ export function fillImage(entry) {
         source = entry;
     }
 
-    if (entry.isIntersecting) {
-        if (source) {
-            fillImageElement(target, source);
-        }
-    } else if (!source) {
-        emptyImageElement(target);
+    if (entry.isIntersecting && source) {
+        fillImageElement(target, source);
     }
 }
 


viown:
While modern browsers are smart enough at handling this automatically, I think emptying the image does help with memory-limited systems such as TVs (webos/tizen).

ghost:
For apps I only have iOS and AppleTV, they don't seem bothered with native apps (Jellyfin and Swiftfin on iOS, Swiftfin on Apple TV) but all are more powerful than an average TV.  It's up to 100 small images with default config.

In any case, there is a real problem with Jellyfin web page rendering.  This was the smallest change that I could identify that resolved the problem.  I would welcome someone who is more expert having a go.

viown:
> In any case, there is a real problem with Jellyfin web page rendering.

Is your issue reproducible on 10.10.z or only on master?

thornbill:
As @viown pointed out this is by design for low end systems.

The most common culprit for page loading issues is browser extensions like dark reader in my experience.

PR #7037, opened 2025-07-23
fix(syncPlay): Playlist synchronization

tppadn:
Replace translateItemsForPlayback with simple sorting to prevent automatic expansion of albums/playlists that broke synchronization.

**Changes**
Replace translateItemsForPlayback with simple sorting to prevent automatic expansion of playlists that broke synchronization. 
The function was expanding the playlist on client side when joining a group, causing playlist content to change from server intent.

**Issues**
I think this solves this issue reported on jellyfin/jellyfin but it needs to be tested : https://github.com/jellyfin/jellyfin/issues/12422

### DIFF
--- a/src/plugins/syncPlay/core/QueueCore.js
+++ b/src/plugins/syncPlay/core/QueueCore.js
@@ -40,9 +40,7 @@ class QueueCore {
 
         console.debug('SyncPlay updatePlayQueue:', newPlayQueue);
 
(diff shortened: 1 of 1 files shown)


thedreaddpirate:
Was this PR intended just to fix this issue?  https://github.com/jellyfin/jellyfin/issues/12422

Because it appears to have been fixed by https://github.com/jellyfin/jellyfin/pull/17797 on the server side.

PR #7045, opened 2025-07-26
Dynamically add "Now Playing" information to the <title>

edent:
**Changes**
This causes the `<title>` element to be rewritten with the name of the currently playing media.

When the next or previous track plays, the title is updated.

Navigating around the UI and changing page may result in the title reverting to the default. It will update when the media changes.

Tested on Firefox and Chrome, on server 10.10.7
**Issues**

As per #7043


gompa:
this seems to work fine,
when playing series it shows the series name not the current episode
not sure if it might be an idea to show artist-title for music and series-episode for shows, and just movie titles for movies?(but i think this is mostly a personal preference)

edent:
> it might be an idea to show artist-title for music and series-episode for shows, and just movie titles for movies?(but i think this is mostly a personal preference)

Agreed. I wasn't sure of the exact format of all the various media metadata, so tried to keep it basic.

Can I suggest starting with this and then, if people find it useful, enhancing it with further information?

timreichen:
> > it might be an idea to show artist-title for music and series-episode for shows, and just movie titles for movies?(but i think this is mostly a personal preference)
> 
> Agreed. I wasn't sure of the exact format of all the various media metadata, so tried to keep it basic.
> 
> Can I suggest starting with this and then, if people find it useful, enhancing it with further information?

I think working out a good default format should be done in this initial PR.

I like the idea of the proposed
- music: `artist - title`
- shows: `series - title`
- movies: `title`

Though shows might also benefit having `S\d{2}E{\d}2` in the title: `Blackadder S02E03 - Head`, though the title might get too long.

In addition, I think it would be cool if there was a setting to enable/disable dynamic document titles.

edent:
I don't know the data structure well enough to make further changes.

If there's documentation on what metadata various media can contain, I'll try to give it a go. If not, I'm happy for you to take this over.

timreichen:
> I don't know the data structure well enough to make further changes.
> 
> If there's documentation on what metadata various media can contain, I'll try to give it a go. If not, I'm happy for you to take this over.

There is https://typescript-sdk.jellyfin.org/interfaces/generated-client.BaseItemDto.html where all the props are listed.

edent:
This now shows the title based on either:

## Audio Now Playing
Screenshot from 2025-08-23 22-26-13

## Video Now Playing

### TV including episode name / number
Screenshot from 2025-08-23 22-25-43

### Movie Now Playing
Screenshot from 2025-08-23 22-24-26

nielsvanvelzen (changes requested):
This change is not going to work as-is. The title is currently managed by React when navigating between pages. Changing the `document.title` manually will just get overridden when navigating again.

PR #7047, opened 2025-07-27
Fix wrong dashboard settings string

timreichen:


thornbill:
There's nothing "wrong" about this... it's general settings

PR #7048, opened 2025-07-27
Refactor updateUserData to reduce Cognitive Complexity

BrainTeazer:
**Changes**
Sonarcloud identified updateUserData as a Cognitive Complexity of 21 whereas 15 is allowed. Some nesting is removed and the function is broken down into different functions.

**Issues**
[Relevant Sonarcloud finding/issue](https://sonarcloud.io/project/issues?issueStatuses=OPEN%2CCONFIRMED&types=CODE_SMELL&id=jellyfin_jellyfin-web&open=AXepJqbaYTlMEX6vFIIA&tab=code)

### DIFF
--- a/src/components/cardbuilder/cardBuilder.js
+++ b/src/components/cardbuilder/cardBuilder.js
@@ -1265,6 +1265,41 @@ function ensureIndicators(card, indicatorsElem) {
 }
 
 /**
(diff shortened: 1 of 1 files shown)


gompa:
this seems to work fine, I'm not that familiar with the card builder so take that with a grain of salt

PR #7049, opened 2025-07-28
Enable HEVC for MPEG-TS HLS over hls.js 1.6+

nyanmisaka:
**Changes**
- Enable HEVC for MPEG-TS HLS over hls.js 1.6+

hls.js supports HEVC in MPEG-TS since 1.6.0. In some cases users may want to switch back from fMP4 to MPEG-TS for compatibility and to alleviate browser issues.


nyanmisaka:
Converted to draft for now, because the implementation in hls.js is unstable in some corner cases.

PR #7050, opened 2025-07-29
Fix vertical alignment during fetch of media aditional content

Thiteago:
**Changes**  
Adjust the vertical alignment of the loader and label when fetching additional content of a media.

| BEFORE | AFTER |
|--------|-------|
| image | image |

### DIFF
--- a/src/styles/videoosd.scss
+++ b/src/styles/videoosd.scss
@@ -174,6 +174,8 @@
 
 .osdMediaStatus {
     margin-left: auto;
+    display: flex;
+    align-items: center;
 }
 
 .osdMediaStatus .animate {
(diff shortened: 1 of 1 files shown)


src/styles/videoosd.scss
```diff
@@ -174,6 +174,9 @@
 
 .osdMediaStatus {
     margin-left: auto;
+    display: flex;
+    align-items: center;
+    gap: 0.125rem;
```
thornbill: Gap isn't supported on some of the browser engines that we support. Please use equivalent margins instead.

Thiteago: Thank you! I've made the necessary adjustment and replaced the gap.

pedrosrc (approved):

crocha12 (approved):

DougNeo (approved):

PR #7054, opened 2025-08-01
fix header icon and description

timreichen:
**Changes**
The header button that currently indicates `user` is not really about the user but settings in general, as it opens a settings menu. This PR fixes the icon and description.

**Issues**


gompa:
this just removes the jellyfin logo on the right ? : 
image

timreichen:
> this just removes the jellyfin logo on the right ? : image

That is weird, the icon shows up on dev and production builds locally. I'll have another look.

timreichen:
@gompa Seems like the button gets overridden depending on whether the user has a profile image or not. Fixed.

Issue #6451, opened 2025-01-16
Not escaping directory names leading to XSS risk


anthonyryan1:
**Describe The Bug**
Within the `#/wizardlibrary.html` template page, we are not currently escaping the directory name, and it can lead to XSS problems.

I'm reporting this publicly, rather than to the security email because there's very little risk in my opinion. With the wizard page being used only on new setups, and Jellyfin and the dashboard libraries page being buried in settings. Also Jellyfin mostly being private use, it's a low priority security bug.

**Steps To Reproduce**
1. Create a new folder with with a maliciously crafted name like `"><img src=x onerror=alert("Not escape")>"`
2. Start setting up a new Jellyfin instance
3. When you setup up the media libraries, select that suspiciously named folder as your media directory
4. Once you save the folder, you will see an alert when returned to the wizardlibrary.html templatepage.

If you continue through the setup, this same XSS will also occur in `#/dashboard/libraries`

**Expected Behavior**
The folder name should be rendered as text, rather than HTML. Using either textContent in javascript, or an escaping function if rendered server side.

**Logs**
N/A

**Screenshots**
N/A

**System (please complete the following information):**
 - Platform: All
 - Browser: All
 -  Jellyfin Version: 10.10.3

**Additional Context**
All credit to Fahimhusain Raydurg, who discovered this issue. I'm just reporting it.

anthonyryan1:
Bump to avoid auto-close.

LucianoHanna:
I have tested it but I didn't see an alert in wizardlibrary.html or #/dashboard/libraries. I don't know if it was something specific on my setup, but there's a bug anyway, probably another payload would work in my setup.

PS: I saw this on devtools: `<img src="x" onerror="alert(&quot;Not" escape")="">`... I don't know there's a quote between "Not escape"

PR #7062, opened 2025-08-04
Fix XSS at wizard and dashboard library

LucianoHanna:
More details at Issue #6451

### DIFF
--- a/src/apps/dashboard/controllers/library.js
+++ b/src/apps/dashboard/controllers/library.js
@@ -1,5 +1,3 @@
-import escapeHtml from 'escape-html';
-
 import taskButton from 'scripts/taskbutton';
 import loading from 'components/loading/loading';
 import globalize from 'lib/globalize';
@@ -255,7 +253,6 @@ function getCollectionTypeOptions() {
 }
 
 function getVirtualFolderHtml(page, virtualFolder, index) {
-    const elementId = virtualFolder.elementId ? `id="${virtualFolder.elementId}" ` : '';
     const cardDiv = document.createElement('div');
     cardDiv.setAttribute('class', 'card backdropCard scalableCard backdropCard-scalable');
     cardDiv.setAttribute('data-index', index);
--- a/src/apps/dashboard/controllers/library.js
+++ b/src/apps/dashboard/controllers/library.js
@@ -255,17 +255,30 @@ function getCollectionTypeOptions() {
 }
 
 function getVirtualFolderHtml(page, virtualFolder, index) {
-    let html = '';
-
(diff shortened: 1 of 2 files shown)


src/apps/dashboard/controllers/library.js
```diff
@@ -255,108 +253,176 @@ function getCollectionTypeOptions() {
...
+    const cardTextLocations = document.createElement('div');
+    cardTextLocations.className = 'cardText cardText-secondary';
     if (virtualFolder.showLocations === false) {
-        html += "<div class='cardText cardText-secondary'>";
-        html += '&nbsp;';
-        html += '</div>';
+        cardTextLocations.innerHTML = '&nbsp;';
     } else if (virtualFolder.Locations.length && virtualFolder.Locations.length === 1) {
-        html += "<div class='cardText cardText-secondary' dir='ltr' style='text-align:left;'>";
-        html += virtualFolder.Locations[0];
```
dmitrylyzo, line 348: Too many changes. You just need to make an escape here:
```js
html += escapeHtml(virtualFolder.Locations[0]);
```

src/apps/wizard/controllers/library.js
```diff
@@ -282,90 +292,178 @@ function getVirtualFolderHtml(page, virtualFolder, index) {
...
+    const cardTextLocations = document.createElement('div');
+    cardTextLocations.className = 'cardText cardText-secondary';
     if (virtualFolder.showLocations === false) {
-        html += "<div class='cardText cardText-secondary'>";
-        html += '&nbsp;';
-        html += '</div>';
+        cardTextLocations.innerHTML = '&nbsp;';
     } else if (virtualFolder.Locations.length && virtualFolder.Locations.length === 1) {
-        html += "<div class='cardText cardText-secondary' dir='ltr' style='text-align:left;'>";
-        html += virtualFolder.Locations[0];
```
dmitrylyzo, line 357: Too many changes. You just need to make an escape here:
```js
html += escapeHtml(virtualFolder.Locations[0]);
```

src/apps/dashboard/controllers/library.js
LucianoHanna: Yes, it should solve that specific XSS but you really shouldn't build a HTML concatenating a lot of string like this, this is a really fragile pattern

dmitrylyzo: You are currently doing 2 things: fixing and refactoring.

1. You should split the fix and the rewrite. Thus, the fix will be accepted ASAP.
2. We are moving to React.

PR #7066, opened 2025-08-05
Make hamburger menu button always the first header item

timreichen:
**Changes**
The hamburger menu should always be the first item in the header for consistency.

**Before**
Bildschirmfoto 2025-08-05 um 13 32 02
**After**
Bildschirmfoto 2025-08-05 um 13 31 52

### DIFF
--- a/src/scripts/libraryMenu.js
+++ b/src/scripts/libraryMenu.js
@@ -40,9 +40,9 @@ function renderHeader() {
     let html = '';
     html += '<div class="flex align-items-center flex-grow headerTop">';
     html += '<div class="headerLeft">';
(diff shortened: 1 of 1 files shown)


erikbucik:
If you are going to reorder the buttons, I'd reorder the `Back` and `Home` buttons as well.
Or even better, remove `Home` because it's already in the hamburger menu.

timreichen:
> If you are going to reorder the buttons, I'd reorder the `Back` and `Home` buttons as well. Or even better, remove `Home` because it's already in the hamburger menu.

I think removing the `home` button makes sense but as another PR. I don't want to combine multiple issues in one PR which are not related. I imagine that makes the PR easier to review and speeds up a potential merge.

PR #7067, opened 2025-08-05
Hide cast button if menu is empty

timreichen:
**Changes**
This PR hides the chromecast button if chromecast is not supported instead of opening an empty menu where it says that it is not supported.

**Before**
Bildschirmfoto 2025-08-05 um 13 59 04

Bildschirmfoto 2025-08-05 um 13 59 05

**After**
Bildschirmfoto 2025-08-05 um 13 59 15

### DIFF
--- a/src/scripts/libraryMenu.js
+++ b/src/scripts/libraryMenu.js
@@ -163,7 +163,9 @@ function updateUserInHeader(user) {
             headerSearchButton.classList.remove('hide');
         }
 
-        if (!layoutManager.tv) {
+        const isChromecastPluginLoaded = !!pluginManager.plugins.find(plugin => plugin.id === 'chromecast');
+
+        if (!layoutManager.tv && isChromecastPluginLoaded) {
             headerCastButton.classList.remove('hide');
         }
 


nielsvanvelzen:
That is not the "chromecast" button. It's the "play on" button, of which chromecasting is just a part. By removing this button you're effectively making it impossible to use the remote control functionality of Jellyfin.

maru801:
This would also break the use of the MPV Shim client.

timreichen:
> That is not the "chromecast" button. It's the "play on" button, of which chromecasting is just a part. By removing this button you're effectively making it impossible to use the remote control functionality of Jellyfin.

> This would also break the use of the MPV Shim client.

I see, sorry for the mixup, but isn't the icon the chromecast icon?
The goal is to hide the button if the menu is empty.

Is the MPV Shim also an element in `playbackManager.getTargets()`? If so the code can be changed to 
```ts
const targets = await playbackManager.getTargets();
if (!layoutManager.tv && targets.length) {
    headerCastButton.classList.remove('hide');
}
```

maru801:
It's a general casting button. It allows you to "cast" to any other present and running Jellyfin (webUI) client on the server's local network.

What it really does is allows you to remote control another client and have it switch to certain pages and play media. Like Niels states, it's a remote control.

MPV Shim is a casting client for Jellyfin. It can only be used when another webUI client opens that casting icon and selects it.

timreichen:
> It's a general casting button. It allows you to "cast" to any other present and running Jellyfin (webUI) client on the server's local network.
> 
> What it really does is allows you to remote control another client and have it switch to certain pages and play media. Like Niels states, it's a remote control.
> 
> MPV Shim is a casting client for Jellyfin. It can only be used when another webUI client opens that casting icon and selects it.

Yes, but what I mean is that if there isn't any selectable option, the button should be hidden. If there are already clients or a client is connected then the button should appear. 

Is there some kind of event or callback whenever a client is connected that gets broadcast to all clients?

viown:
> Is there some kind of event or callback whenever a client is connected that gets broadcast to all clients?

No, but when clicking on the play on button, it does reload the sessions. If we hide the button, the user has to refresh the page to be able to see the button again after launching their cast client.

Regardless, I personally don't think we should hide on empty, because the button does make the user aware of the feature and hiding it can make it more confusing when troubleshooting scenarios like cast client not connecting, etc.

PR #7069, opened 2025-08-06
Add Artists to detail pages, deduplicate Albums and Appears On

evan314159:
**Changes**
Add Artists to detail pages, deduplicate Albums and Appears On

**Issues**
Implements the UI component of jellyfin/jellyfin#14591

### DIFF
--- a/src/controllers/itemDetails/index.html
+++ b/src/controllers/itemDetails/index.html
@@ -199,6 +199,13 @@
                     </div>
                 </div>
 
+                <div id="artistsCollapsible" class="verticalSection detailVerticalSection hide">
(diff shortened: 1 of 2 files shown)


ghost:
For those trying this with classical collections where you may see a lot more Appears on matches, you might like to do set like this so that Appears On shows as a table instead of a scroller:

#itemDetailPage .moreFromArtistSection .emby-scroller { margin-right: 0; }
#itemDetailPage .moreFromArtistSection .emby-scrollbuttons { display: none; }
#itemDetailPage .moreFromArtistSection .itemsContainer { flex-wrap: wrap; }

ghost:
Resubmitted as jellyfin/jellyfin#14654

PR #5285, opened 2024-03-17
Refactor dashboard user

grafixeyehero:
**Changes**
- Replace Legacy `Api` to new `SDK API`
- Added React Hook Form for efficient form handling.
- Add React Hook Form MUI for seamless integration with Material-UI components.
- Implemented Zod for schema validation and error handling
- Replace components to MUI
- Add a New `users/settings` route that handle **profile**, **access**, **parental control**, and **password** tab page
- Remove `Library Access permission setting` From user add page
- Add `EnableUserPreferenceAccess` setting in dashboard User Profile Form to Allow user to edit profile image and password
- Hide `reset password` button for `Admin` and `Normal` user. 
- Admin can `reset password` for `non-admin user`.
- Disable `Manage Server` option if no password is configured.
- Minor UI enhancements.

if I miss something let me know

**Issues**
https://github.com/jellyfin/jellyfin-web/issues/5199

<details>

<summary>Screenshot Dashboard Users</summary>

| Dashboard User Page|  |
| --- | --- |
| Screenshot 2024-04-10 205527 |  | 
| Dashboard Profile Page|  |
| User with Admin Privilege | Mange Server Disabled if Password not provided for normal user. To activate Manage Server option password must set for user first |
| Screenshot 2024-11-06 204241 | Screenshot 2024-11-06 205000 |
| Dashboard Password Page|  |
| Password Required for Admin to change | Admin Can Reset Password for normal users |
| Screenshot 2024-11-06 204804 | Screenshot 2024-04-17 223447 |
</details>

<details>

<summary>Screenshot Profile Page</summary>

| Admin User | Allowed to Edit profile image and password |
| --- | --- |
| Screenshot 2024-11-07 040312 | Screenshot 2024-11-07 230740 | 
| | Not Allow to Edit profile image and password when Preference Access Disable |
| | Screenshot 2024-11-07 230905 | 

</details>

<details>

<summary>Screenshot UI Change</summary>

| Before | After |
| --- | --- |
| Before | After | 

</details>

PR #7089, opened 2025-08-11
Migrate dashboard user pages to use TS SDK

viown:
Since it looks like https://github.com/jellyfin/jellyfin-web/pull/5285 is inactive, I've re-opened https://github.com/jellyfin/jellyfin-web/pull/6403

**Changes**
Migrate the dashboard user pages to use TS SDK (Step 1 of 2)

**Issues**
N/A


viown:
For the nested functions / code smells, I plan to refactor them in follow-up PRs re-designing the pages in material UI.

PR #7090, opened 2025-08-12
Add overview text to episode cards in "Next Up" and "More from" 

aiudirog:
**Changes**
Adds the `Overview` text for each episode to their cards in the "Next Up" and "More from" sections. 

This is to avoid the interaction cost of navigating back out to the main episode list to figure out which episode is which if you aren't familiar with the titles.

Examples:

<details>
<summary>Next up</summary>

image

</details>

<details>
<summary>More from</summary>

image

</details>

Please let me know if I should look into adding a display setting to toggle this. I also wasn't quite sure if/where I should put any test cases, so any guidance would be great!

**Issues**
N/A

### DIFF
--- a/src/components/cardbuilder/cardBuilder.js
+++ b/src/components/cardbuilder/cardBuilder.js
@@ -573,6 +573,10 @@ function getCardFooterText(item, apiClient, options, footerClass, progressHtml, 
     }
 
     if (showOtherText) {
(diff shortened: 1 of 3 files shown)


MBR-0001:
@aiudirog in my opinion this should have a setting since not everyone would want this

PR #7100, opened 2025-08-15
Mark item played when starting next episode through upNextDialog

### LINKED ISSUE #7039: Episode not marked as played when starting next episode through upNextDialog

### Describe the bug

I've found that episodes are not guaranteed to be marked as watched when starting the next episode through the `upNextDialog`. They are only marked as played if the left over time at the end of the episode falls below the `Maximum resume percentage`. I've found this not to be the case for relatively short content with long credits. Note that this setup uses introskipper to mark to extend the outro segment beyond the default outro segment length.

### Reproduction Steps

1. Playback a short episode with a long outro segment.
2. Click to play next episode from `upNextDialog`
3. (Watch a few episodes like this)
4. Return to home screen and observe all 'watched' eposides in the `Continue Watching` section with just the outro segments remaining.

### Expected/Actual behaviour

I expect when starting the new episode through this dialog popup to mark the current episode as watched always, regardless of time remaining. 
I.e. users should not have to modify the `Maximum resume percentage` for this to work as expected.

### Logs

_No response_

### Server version

10.10.7

### Web version

10.10.7

### Build version

10.10.7

### Platform

Anything that wraps jellyfin-web

### Browser

Firefox, Chrome, ...

### Additional information

When looking for the cause of this issue I've found a possible solution to explicitly mark the episode as watched in the [`onStartNowClick()` function of `upnextdialog.js`](https://github.com/jellyfin/jellyfin-web/blob/801b96413b23a6f22c3f4c6b438b68724f31152e/src/components/upnextdialog/upnextdialog.js#L104).

I am not familiar with the correct method of implementing this, but a potential way to implement this would be something like this:
```js

async function onStartNowClick() {
    const options = this.options;

    if (options) {
        const player = options.player;

        await this.hide();
        
        // Explicitly mark Played when starting next episode:
        const currentItem = playbackManager.currentItem(player);
        if (itemHelper.canMarkPlayed(currentItem)) {
            const apiClient = ServerConnections.getApiClient(currentItem);
            apiClient.markPlayed(apiClient.getCurrentUserId(), currentItem.Id, new Date());
        }
        playbackManager.nextTrack(player);
    }
}
```

RoboMagus:
**Changes**

Explicitly marks episodes as watched when starting next episode through `upNextDialog`. 
This fixes a problem where for episodes with relatively long outros that fall outside the `Maximum resume percentage` will remain in the `continue watching` section while only the outro credits remain.

**Issues**

Fixes #7039

**Notes**

Changes have been tested locally and appear to work. This is however my first contribution to this code-base and this solution may not be the best option to solve this particular issue.

### DIFF
--- a/src/components/upnextdialog/upnextdialog.js
+++ b/src/components/upnextdialog/upnextdialog.js
@@ -109,6 +109,11 @@ async function onStartNowClick() {
 
         await this.hide();
 
+        const currentItem = playbackManager.currentItem(player);
+        if (itemHelper.canMarkPlayed(currentItem)) {
+            const apiClient = ServerConnections.getApiClient(currentItem);
+            apiClient.markPlayed(apiClient.getCurrentUserId(), currentItem.Id, new Date());
+        }
         playbackManager.nextTrack(player);
     }
 }


thornbill:
Any idea if this would cause double counting in the case that it is within the end duration?

RoboMagus:
> Any idea if this would cause double counting in the case that it is within the end duration?

I have not explicitly tested for double counting, but have not ran into any issues in my time running with this change.

Looking at the receiving end of that function call, if explicitly marking episodes as played _does_ cause issues it seems best to be solved on the server side as there all required data is available to implement a sort of _debounce_ based on the known last played DateTime.

P.S.;
I found that in `jellyfin/jellyfin/Emby.Server.Implementations/Session/SessionManager.cs` the `PlayCount` increment appears inconsistent. In `OnPlaybackStart()` the PlayCount is already updated, and on `OnPlaybackStopped()` this may be incremented again or not at all. (The `playedToCompletion` does not seem to correlate with the `PlayCount` update.
This might be intended behavior, but feels worth noting as it might be problematic.

PR #7109, opened 2025-08-22
Improve Artists support on detail pages

evan314159:
**Changes**
On detail pages:
- For Albums, show Artists and hide People
- for Artists, deduplicate albums where the artist is both an artist and album artist

### DIFF
--- a/src/controllers/itemDetails/index.html
+++ b/src/controllers/itemDetails/index.html
@@ -199,6 +199,13 @@
                     </div>
                 </div>
 
+                <div id="artistsCollapsible" class="verticalSection detailVerticalSection hide">
+                    <h2 id="artistsHeader" class="sectionTitle sectionTitle-cards padded-right">${Artists}</h2>
+                    <div is="emby-scroller" class="padded-top-focusscale padded-bottom-focusscale no-padding" data-centerfocus="true">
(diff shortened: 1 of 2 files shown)


ghost:
I think this should address jellyfin/jellyfin#14749 and jellyfin/jellyfin#14750

thornbill:
This was caused by a server bug that should be fixed by https://github.com/jellyfin/jellyfin/pull/14971

ghost:
At least at the time when I submitted this PR there was no present support for showing Artists on music detail pages.  With change #14971, Artists are no longer People, so the part of this PR that hides People on music detail pages may not be as necessary, but unless there is another UI change in the works something is needed to show the Artists on music detail pages?

ghost:
Confirming that with #14971 Artist icon does not show up on album detail page, and on Artist detail page artist appears twice for each album -- once under Albums for album artist, and then again under Appears On for artist.  The PR avoided these.  Please review.

theguymadmax:
@evan314159  The remaining issues you mentioned should be resolved by [#14991](https://github.com/jellyfin/jellyfin/pull/14991) in RC9. As for broader UI changes, those are on hold until the upcoming redesign.

Issue #7107, opened 2025-08-20
Volume slider is not working in Firefox


rgabbo:
### This issue respects the following points:

- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-web/issues?q=is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

When I play a video using firefox 141.0.3 and I try to adjust volume with volume slider keeping it pressed, the volume changes of few points but after became unresponsive and require another click to make further adjustments. This issue doesn't occur on chromium. The video show

### Reproduction Steps

1. play a video
2. try to adjust the volume
3. issue occurs

### Expected/Actual behaviour

I expect the possibility to adjust the volume smoothly with the slider

### Server version

10.10.7

### Web version

10.10.7

### Build version

10.10.7

### Platform

Windows

### Browser

Firefox 141.0.3

### Additional information

https://github.com/user-attachments/assets/6798b5b9-2726-4d6e-b755-d334f3209e04

antoinefortin:
Hello, thank you for pointing this is out. I fixed it locally on my machine, its my first time commiting to a open source project so I will try to open a MR

FurmanSK:
Still happens on 10.11.11 on firefox.

PR #7110, opened 2025-08-25
Fixed audio change playback on Firefox

antoinefortin:
**Changes**
- Custom drag handling - mouse & pointer event listeners to  smooth volume slider dragging in Firefox  HTML range slider behavior is limited
- Real-time volume updates - Calculates volume levels in real-time as the user drags, provide immediate audio feedback without waiting for mouse release
- Precise position mapping - I took the mouse coordinate in Screen to be  mouse coordinates to accurate volume percentages (0-100) based on the slider's dimensions and min/max values 
**Issues**
[](https://github.com/jellyfin/jellyfin-web/issues/7107)

### DIFF
--- a/src/controllers/playback/video/index.js
+++ b/src/controllers/playback/video/index.js
@@ -1850,8 +1850,8 @@ export default function (view) {
         playbackManager.setVolume(e.target.value, currentPlayer);
     });
 
-        // Firefox vol slider
-    if (browser.firefox) {
+    // Firefox vol slider
+    if (browser.firefox)  {
         let isDragging = false;
         const updateVolumeFromMousePosition = (e) => {
             const slider = nowPlayingVolumeSlider;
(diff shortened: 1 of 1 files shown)


dmitrylyzo:
The root of the problem is that the OSD switches focus to the Pause button because the volume slide is not focusable:
https://github.com/jellyfin/jellyfin-web/blob/18f3083e69e17634c52ff842e003b1b01c17fcb0/src/components/focusManager.js#L96
https://github.com/jellyfin/jellyfin-web/blob/18f3083e69e17634c52ff842e003b1b01c17fcb0/src/components/focusManager.js#L107-L109

As a workaround, we could make the volume slider focusable (by adding the `focusable` class) for non-TV layouts (_It cannot be focusable on TV, as this will disrupt navigation_).
Similar to:
https://github.com/jellyfin/jellyfin-web/blob/18f3083e69e17634c52ff842e003b1b01c17fcb0/src/controllers/playback/video/index.js#L1662-L1664

pierrelasse:
*Any plans to finish this MR? Would be greatly appreciated*

PR #7118, opened 2025-08-28
Update LanNetworksHelp description for clarity

linkandzelda91:
**Changes**
previously, it indicated that if you leave it blank, only the server's local network is considered local. I tested this and afaik it's wrong, leaving it blank sets all RFC1918 addresses to local, and additionally it has an effect on the remote access settings below it.

steps I used to confirm the effect on remote access:
1. set lan networks blank
2. disable remote access
3. attempt connection from RFC1918 outside of network
I got locked out, had to edit network.xml to either put my clients LAN subnet in the lan networks, or enable remote again.

steps I used to confirm the effect on bandwidth:
1. set lan networks blank
2. set global internet bandwidth limit blank
3. play a direct play compatible video
4. it was direct play for me
5. change lan networks to something outside of your client subnet. I used 192.168.0.0/24, and my client is from 172.16.0.0/24
6. set global internet bandwidth to something below your test media's bitrate
7. play a direct play compatible video
8. it was being transcoded

**Issues**
N/A


linkandzelda91:
also if I'm reading the server side code correctly this makes sense.
I think this is the part that checks for server access, it returns a quick allow right at the start if the IsInLocalNetwork(remoteIP) is true.
```
public RemoteAccessPolicyResult ShouldAllowServerAccess(IPAddress remoteIP)
    {
        var config = _configurationManager.GetNetworkConfiguration();
        if (IsInLocalNetwork(remoteIP))
        {
            return RemoteAccessPolicyResult.Allow;
        }
```

initially I thought remoteIP maybe was only for remote clients, but the comment for remote IP:
"/// <param name="remoteIP">IP address of the client.</param>"

But what does IsInLocalNetwork do?
and inside the definition for IsInLocalNetwork:
```
private bool CheckIfLanAndNotExcluded(IPAddress address)
    {
        foreach (var lanSubnet in _lanSubnets)
        {
            if (lanSubnet.Contains(address))
            {
                foreach (var excludedSubnet in _excludedSubnets)
                {
                    if (excludedSubnet.Contains(address))
                    {
                        return false;
                    }
                }

                return true;
            }
        }

        return false;
    }
```
so if the IP is inside _lansubnets, we get a pass.

_lansubnets definition
```
/// <summary>
    /// Unfiltered user defined LAN subnets (<see cref="NetworkConfiguration.LocalNetworkSubnets"/>)
    /// or internal interface network subnets if undefined by user.
    /// </summary>
    private IReadOnlyList<IPNetwork> _lanSubnets;
```

Now I think I just need to show that _lanSubnets gets all RFC1918 by default:
```
  // If no LAN addresses are specified, all private subnets and Loopback are deemed to be the LAN
            if (!NetworkUtils.TryParseToSubnets(subnets, out var lanSubnets, false) || lanSubnets.Count == 0)
            {
                _logger.LogDebug("Using LAN interface addresses as user provided no LAN details.");

                var fallbackLanSubnets = new List<IPNetwork>();
                if (IsIPv6Enabled)
                {
                    fallbackLanSubnets.Add(NetworkConstants.IPv6RFC4291Loopback); // RFC 4291 (Loopback)
                    fallbackLanSubnets.Add(NetworkConstants.IPv6RFC4291SiteLocal); // RFC 4291 (Site local)
                    fallbackLanSubnets.Add(NetworkConstants.IPv6RFC4193UniqueLocal); // RFC 4193 (Unique local)
                }

                if (IsIPv4Enabled)
                {
                    fallbackLanSubnets.Add(NetworkConstants.IPv4RFC5735Loopback); // RFC 5735 (Loopback)
                    fallbackLanSubnets.Add(NetworkConstants.IPv4RFC1918PrivateClassA); // RFC 1918 (private Class A)
                    fallbackLanSubnets.Add(NetworkConstants.IPv4RFC1918PrivateClassB); // RFC 1918 (private Class B)
                    fallbackLanSubnets.Add(NetworkConstants.IPv4RFC1918PrivateClassC); // RFC 1918 (private Class C)
                }

                _lanSubnets = fallbackLanSubnets;
            }
            else
            {
                _lanSubnets = lanSubnets;
            }
```

so, if _lanSubnets is empty, it gets set to fallbackLanSubnets, which looks like it contains all RFC1918, plus loopback and some ipv6 stuff I dont understand.

From there it's used as the reference for if a client's IP is in LAN and not excluded, then finally used to decide if the user can access the server.

src/strings/en-us.json
```diff
@@ -1034,7 +1034,7 @@
     "LabelYear": "Year",
     "LabelYoureDone": "You're Done!",
     "LabelZipCode": "Zip Code",
-    "LanNetworksHelp": "Comma separated list of IP addresses or IP/netmask entries for networks that will be considered on local network when enforcing bandwidth restrictions. If set, all other IP addresses will be considered to be on the external network and will be subject to the external bandwidth restrictions. If left blank, only the server's subnet is considered to be on the local network.",
+    "LanNetworksHelp": "Comma separated list of IP addresses or IP/netmask entries for networks that will be considered on local network when enforcing bandwidth and remote access restrictions.If left blank, all RFC1918 addresses are considered local.",
```
viown:
```suggestion
    "LanNetworksHelp": "Comma separated list of IP addresses or IP/netmask entries for networks that will be considered on local network when enforcing bandwidth and remote access restrictions. If left blank, all RFC1918 addresses are considered local.",
```

PR #7124, opened 2025-09-01
Fix pdf extension check

### LINKED ISSUE #7126: Cannot read PDF with capital extension

### Describe the bug

Cannot read pdf file when the extension .pdf is not in lowercase

### Reproduction Steps

Try to open a pdf file with the extension in capital (.PDF)

### Expected/Actual behaviour

To be able to open the PDF file

### Logs

_No response_

### Server version

 10.10.7

### Web version

10.10.7 (f4b8aa0ed)

### Build version

10.10.7

### Platform

Ubuntu

### Browser

Brave

### Additional information

_No response_

S1NJED:
fix pdf file with uppercase extension could not be played

**Changes**
It can now read pdf files with the extension uppercase (.PDF), it was already the case for .cbz file (fixed #7119 )

**Issues**
Fixes #7126


thornbill:
Did you test this? The CBZ issue was for files **inside** the CBZ file. Also there is no way a capitalization issue would cause different behavior between web and JMP as reported in https://github.com/jellyfin/jellyfin-media-player/issues/977.

S1NJED:
Yes I have test it.

current code 
image

with my fix
image

edit: indeed the second issue is irrelevant mb

theguymadmax:
Tested this and it works, but it is also irrelevant to issue 1 - comic reader issue, which was already fixed.  This fixes an undocumented issue with Jellyfin not rendering PDFs with uppercase extensions.

thornbill (approved):

PR #7125, opened 2025-09-02
Changed the erroneous spelling of the OK button in localizations.

eje211:
**Changes**
The "OK" button that appears everywhere is wrongly spelt. "Ok" is not a possible spelling. While the origins of the word OK are disputed, [the most frequently accepted one](While the origins of the word are disputed, the most frequently accepted i) is a comedic abbreviation for "oll korrect" from early-mid-nineteenth century America. That makes "OK" a possible spelling and the spelling most often requires from style books.  "Okay", which is phonetic, is also acceptable. "Ok" is not a possible spelling. It's basically a misspelled "ock" sound. It makes Jellyfin look amateurish.

**Issues**
I resolved the wrong spelling of the "OK" button.


viown:
Changes should only be done to the `en-us.json` file, as all other files are managed by weblate.

eje211:
That was a simple request to fulfill!

And I didn't know about weblate. Sorry.

PR #7129, opened 2025-09-08
fix: endless loop when saving metadata people with no type

S1NJED:
**Changes**

Fix infinite loading circle when saving metadata of an item with people with no `type`.

- Now the `type` field is required in the add people form.
- I handled the error with the code `400` that was causing the infinite loading (i added it in case) 

**Issues**

#6559


thornbill:
Appreciate the PR but it seems like the easier solution is to just remove the blank option from the form. 

Superseded by #7197

PR #6346, opened 2024-11-25
Add "reset all filters" button to filter dialog

othmar52:
**Changes**
Currently it can be tedious to find active filters in those filter dialog dropdowns  
This PR adds a button to reset all filters in case any filter is active  

image  

As a follow-up it would make sense to add indicators to all dropdown items that have active filters.  
But i have no idea where to put such a suggestion/enhancement-request!?

**Issues**
N/A

PR #7130, opened 2025-09-09
Add a Reset Filters button to bottom of filter dropdown

1hitsong:
**Changes**
- Adds Reset Filters button to bottom of the filter dropdown
- On click, all applied filters are removed and the page content is refreshed
- If no filters are applied, disable button
- Apply to both standard and experimental layouts

**Screenshots**

Standard Layout
image

Experimental Layout
image

### DIFF
--- a/src/components/filterdialog/filterdialog.js
+++ b/src/components/filterdialog/filterdialog.js
@@ -9,6 +9,7 @@ import '../../elements/emby-collapse/emby-collapse';
 import './style.scss';
 import template from './filterdialog.template.html';
 import { stopMultiSelect } from '../../components/multiSelect/multiSelect';
+import { getFilterStatus } from 'components/filterdialog/filterIndicator';
 
 function merge(resultItems, queryItems, delimiter) {
     if (!queryItems) {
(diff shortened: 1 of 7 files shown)


src/components/filterdialog/filterdialog.js
```diff
@@ -145,6 +154,22 @@ function setVisibility(context, options) {
     if (options.mode === 'episodes') {
         showByClass(context, 'episodeFilter');
     }
+
+    if (!options.hasFilters) {
```
thornbill, line 158: Would it make sense to just call `getFilterStatus` here also instead of passing an `option`?

tnyeanderson: Looks like the `option` is already passed to the func anyway. Unless there's reason to not trust that value (I'm unfamiliar with the code so can't say whether this is a valid concern), perhaps it's more efficient to use the single boolean negation rather than calling `getFilterStatus` which evaluates ~20 values to return the result.

Like I said, I'm unfamiliar with this code base and y'all would know better than me. Just a user who desperately wants this feature merged, and wanting to make sure it stays moving :)

Thanks for all your hard work on this project!

tnyeanderson:
#6346 was approved over a year ago, but was never merged. This PR has the same goal, and I think we all agree this is an extremely useful (and needed) feature. If there's absolutely anything I can do to help get this merged (even if it's just this little poking reminder in a flood of PRs), let me know, I am happy to get my hands dirty!

Once again, thank you all for your hard work on this project. I use it daily and really appreciate the time y'all have put into it :)

PR #7132, opened 2025-09-10
Added frame by frame control by using "," and "."

### LINKED ISSUE #3087: Allow per-frame playback navigation
On media players such as youtube, it's possible to seek through the video one frame at a time using the `,` and `.` keys. This is a super useful feature, for example when you want to take a screenshot of one specific frame. I appreciate that there's more granularity in seeking intervals than other apps such as Plex, but I've always felt myself longing to be able to seek on a frame-to-frame basis. Didn't see any existing issues or feature requests for this, so thought I'd make one myself.

On a technical level, this should be easily accomplishable. There's only really two places where changes need to be made:
1. https://github.com/jellyfin/jellyfin-web/blob/master/src/controllers/playback/video/index.js#L1041  
A key listener on the playback controller for the new navigation keys (ex. `,` and `.`), which would call:
2. https://github.com/jellyfin/jellyfin-web/blob/master/src/components/playback/playbackmanager.js#L3465  
New functions on the playbackManager which would call seekRelative using the tick equivalent of one frame.

This is what I gleaned from a quick glance-over of the source code. The most challenging part, in my opinion, would be determining how many "ticks" a single frame lasts for. A single second is equivalent to 50,000,000 ticks, so all we'd need to determine is how many frames per second there are in the current video - I'm sure that kind of information would exist _somewhere_, just would need to look further into it.

If this is something you'd all fine with but don't want to invest time into, I'd be more than happy to try implementing this myself. Thank you for your time!

nixietab:
**Changes**
Added frame-by-frame navigation functionality to the video player.
When the video is paused, users can now advance one frame forward by pressing "." or go one frame backward by pressing ",".

**Issues**
Resolves #3087

_This is a proof of concept on how would work_

### DIFF
--- a/src/plugins/syncPlay/ui/players/HtmlVideoPlayer.js
+++ b/src/plugins/syncPlay/ui/players/HtmlVideoPlayer.js
@@ -27,6 +27,10 @@ class HtmlVideoPlayer extends NoActivePlayer {
                 return this.player.currentTimeAsync();
             };
         }
+
+        // Frame stepping constants
+        this._frameStepListener = null;
+        this._frameDuration = 1 / 25; // Default to 25fps, adjust if you can get actual framerate
     }
 
     /**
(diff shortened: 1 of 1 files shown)


flying-sheep:
Superseded by #7315, where I use the stream’s FPS

PR #7142, opened 2025-09-16
Make DirectoryBrowser and UserSettings available to plugins

zachdeibert:
**Changes**
The DirectoryBrowser and UserSettings modules can now be accessed from plugins to use (e.g. in their configuration pages).
This also sets up a pluginImport() function that can be used for similar needs in the future in a cleaner way that importing everything into globals.  It deprecates the old globals that used to exist for this purpose.


zachdeibert:
Any thoughts on this?

PR #7145, opened 2025-09-23
Set MaxParentalRating and MaxParentalSubRating when setting parental controls

### LINKED ISSUE #14838: Australian Classifications still lumped under 15 in 10.11.0rc5
### Description of the bug

After upgrading to 10.11.0RC5 and waiting for the first library scan to finish, then saving a random change to my son's user profile, PG through MA15+ are still all listed as 15 in the user options.

### Reproduction steps

Install 10.11.0 rc 5 in Australia 
Wait for a first library scan to complete
Adjust a users settings
Attempt to enable PG for that user but not MA15+

Same behaviour as 10.10.7

Ratings split into different options

### Jellyfin Server version

Unstable

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.0 rc 5

### Environment

```markdown
- OS: OpenSUSE
- Linux Kernel:
- Virtualization: docker
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Jellyfin Data Storage:
- Media Storage:
- External Integrations:
```

### Jellyfin logs

```shell
NA
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

Shadowghost:
**Changes**
After http://github.com/jellyfin/jellyfin/pull/12615 sub-ratings need to be set for proper control if the current user's ratings support them.

**Issues**
Fixes https://github.com/jellyfin/jellyfin/issues/14838

### DIFF
--- a/src/apps/dashboard/routes/users/parentalcontrol.tsx
+++ b/src/apps/dashboard/routes/users/parentalcontrol.tsx
@@ -1,4 +1,4 @@
-import type { AccessSchedule, ParentalRating, UserDto } from '@jellyfin/sdk/lib/generated-client';
+import type { AccessSchedule, ParentalRating, ParentalRatingScore, UserDto } from '@jellyfin/sdk/lib/generated-client';
 import { UnratedItem } from '@jellyfin/sdk/lib/generated-client/models/unrated-item';
 import { DynamicDayOfWeek } from '@jellyfin/sdk/lib/generated-client/models/dynamic-day-of-week';
 import escapeHTML from 'escape-html';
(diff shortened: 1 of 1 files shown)


LovesTha:
Preview server doesn't retain setting of PG, switches to no selection (and other clients fail to load with that user when in this state) (server running 10.11.0 rc5)

crobibero:
This PR doesn't work.

`parentalRating` is set to the select HtmlElement.
The option value is also set to `[object Object]` which I don't think will work even if we extracted the selected option.

image

Shadowghost:
@crobibero @LovesTha should work now. But I'm not that well-versed in React, so code-wise this can for sure be solved better.

LovesTha:
That appears to work to me.

LovesTha:
But a test in another 'problem' rating system would be sane.

thornbill (approved):

PR #7146, opened 2025-09-25
feat: improve user creation/update error messages and fix backdrops display

alghyt:
**Changes**

**Issues**


viown:
https://github.com/jellyfin/jellyfin-web/pull/7147#issuecomment-3341540935

PR #7147, opened 2025-09-26
fix: support PDF files with uppercase extension (.PDF)

alghyt:
**Changes**

**Issues**


viown:
This PR does multiple unrelated things and is unreviewable in this state.

PR #7150, opened 2025-09-28
Add download all support for music albums and collections

thornbill:
**Changes**
Extends download all support for music albums and collections

**Issues**
N/A

### DIFF
--- a/src/components/itemContextMenu.js
+++ b/src/components/itemContextMenu.js
@@ -426,17 +426,19 @@ function executeCommand(item, id, options) {
             case 'downloadall': {
                 const downloadItems = items => {
                     import('../scripts/fileDownloader').then((fileDownloader) => {
-                        const downloads = items.map(item => {
-                            const downloadHref = apiClient.getItemDownloadUrl(item.Id);
-                            return {
-                                url: downloadHref,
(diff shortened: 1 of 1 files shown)


viown (approved):
- When I try to download an album with more than 10 tracks, not all tracks are downloaded with some missing. Looks like a browser or server limitation to me. When I experimented with adding a 100ms delay between downloads the issue seemed to be fixed.
- When downloading a collection with a music album inside, the download will fail due to an unsupported item type.

The server supporting downloading albums as a zip would likely fix both issues I mentioned.

thornbill:
> * When I try to download an album with more than 10 tracks, not all tracks are downloaded with some missing. Looks like a browser or server limitation to me. When I experimented with adding a 100ms delay between downloads the issue seemed to be fixed.

This is interesting since we should already be adding a delay: https://github.com/jellyfin/jellyfin-web/blob/master/src/scripts/multiDownload.js#L49

I don't think it's a blocker for now, but we might need to investigate further if it is an issue for people.

> * When downloading a collection with a music album inside, the download will fail due to an unsupported item type.

This should be fixed by the filtering I added on `CanDownload` .

PR #7167, opened 2025-10-02
Fix the tonemap box not hidden in software transcoding

nyanmisaka:
**Changes**
- Fix the tonemap box not hidden in software transcoding

**Issues**
- a3872ff is not included in b3de4af


viown (approved):

viown:
Should we remove the `AllowTonemappingSoftwareHelp` localization string if it's not needed anymore?

nyanmisaka:
> Should we remove the `AllowTonemappingSoftwareHelp` localization string if it's not needed anymore?

Yes. It can be removed now.

PR #7173, opened 2025-10-07
Added option to invert episode order

need4swede:
See [PR #14956](https://github.com/jellyfin/jellyfin/pull/14956)

**Changes**
Introduces a new checkbox to allow inverting episode order for series in the metadata editor. Updates the UI to show or hide this option based on item type, handles its value on submit and reset, and adds corresponding string resources for English locales.

**Issues**
Fixes [Display Order: Aired (newest first) #11209](https://github.com/jellyfin/jellyfin/discussions/11209)


thornbill:
Tagging as "discussion needed" since **every** client would require updating to support this.

PR #7177, opened 2025-10-08
Show scroller content that overflows

thornbill:
**Changes**
Allows scroller content that overflows the container to still be visible allowing for an "edge to edge" experience.

Screenshot 2025-10-08 at 12-42-17 Jellyfin Edge

**Issues**
N/A


viown:
There is an issue where the persons can sometimes intersect with the poster if the body is short:
image

thornbill:
Blocked until #7195 lands

PR #7190, opened 2025-10-13
Fix "More From" section to show all albums by album artist

theguymadmax:
**Changes**

With the contributing artist query fix ([#14991](https://github.com/jellyfin/jellyfin/pull/14991)) on the server, we need to update the condition for the "More from" section.

Before: "More from" only showed albums where they were a contributing artist 

After: "More from" shows all albums where they are the main album artist

**Issues**
 N/A

### DIFF
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -1194,7 +1194,7 @@ function renderMoreFromArtist(view, item, apiClient) {
         if (item.Type === 'MusicArtist') {
             query.ContributingArtistIds = item.Id;
         } else {
-            query.AlbumArtistIds = item.AlbumArtists.map(artist => artist.Id);
+            query.AlbumArtistIds = item.AlbumArtists.map(artist => artist.Id).join(',');
         }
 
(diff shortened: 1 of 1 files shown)


src/controllers/itemDetails/index.js
```diff
@@ -1194,7 +1194,7 @@ function renderMoreFromArtist(view, item, apiClient) {
         if (item.Type === 'MusicArtist') {
             query.ContributingArtistIds = item.Id;
         } else {
-            query.ContributingArtistIds = item.AlbumArtists.map(artist => artist.Id).join(',');
+            query.AlbumArtistIds = item.AlbumArtists.map(artist => artist.Id);
```
thornbill: It might be better to keep the explicit `.join(',')` here. It seems like this works, but the apiclient doesn't have any specific handling for an array so it's dependent on `toString()` returning a joined string when encoding the parameters.


```suggestion
            query.AlbumArtistIds = item.AlbumArtists.map(artist => artist.Id).join(',');
```

PR #7223, opened 2025-10-20
Add error handling to rest of dashboard pages

viown:
(review with whitespace disabled)

Users experiencing errors may usually be met with an infinitely loading page with no indication that something went wrong. While we've already added error handling as part of the migration to react, it was not done consistently on all pages.

**Changes**
- Add error handling to rest of dashboard pages

**Issues**
N/A


thornbill:
What do you think about using React Router's `loader` + `errorElement` for error handling?

viown:
The issue I had with loaders is that the TS SDK would be unauthenticated during the loading stage for react router. I haven't looked into it deeply, but I think to fix this, we need to refactor the TS SDK to not rely on the legacy apiclient in order to be authed.

PR #7226, opened 2025-10-20
Add a toggleable option (in appSettings) to turn hls.js on/off on webOS (building upon #6678)

damienl560:
#6678 added use of hls.js for WebOS >= 4, to fix issues with native WebOS not playing multichannel sound correctly on some TVs and to fix fast forward issues.

These issues are however specific to certain TV (and/or firmware versions) and #6678 also introduces other issues, namely loss of native (E)AC3 audio, which is transcoded to AAC. I have also experienced some problems getting Dolby Vision Profile 8 to work reliably with hls.js (enabling fMP4 seems to improve this issue, but not always reliably).

Furthermore, fast forward issues were also addressed by #6530, which limits the usefulness of #6678.

This PR adds a toggle in the playback settings, storing it as an appSetting, which is used in conjunction with checking whether WebOS >= 4 is used to determine if the native HLS player or hls.js will be used. This introduces greater flexibility and allows users who are not experiencing problems with the native player to enjoy the original audio.

This is my first PR so I'm sure there is room for improvement :-)


tuzidaguai:
I packaged your modified jellyfin-web. In the webOS client's playback settings, an option appeared: “Use hls.js instead of native HLS player.” When I disabled it and played a DV P8 EAC3 video, Jellyfin displayed that it was working in remuxing mode. If the option is enabled, it works in direct stream mode, with the audio transcoded to AAC.
Due to time constraints, I've only tested a few files like P8 MKV, P5 MP4, and P7 MKV. I'll continue testing more files soon.
Thank you very much for your PR. It works perfectly.

timitt:
As a person behind https://github.com/jellyfin/jellyfin-web/pull/6530 I think merging this patch is really needed. Especially if those new LG patches are going to fix some of the issues with newer models. I should have done it myself, but I had quite limited dev tools and also was skeptical that tv model based option would be allowed (that fast forward patch was not there when I started my work).

fkroener:
This patch fixes my issues described in https://github.com/jellyfin/jellyfin-webos/issues/310#issuecomment-3448495558. With the new option turned off, as well as fMP4-HLS disabled, Dolby Vision with EAC3+Atmos works again, remuxed from mkv to ts.

agneevX:
If this is merged, users will get two options:

hls.js being enabled will prevent seeking issues with remuxed DV files (#6678) but leave users with broken Dolby Atmos support and transcoding of most audio codecs to AAC.

With hls.js disabled, users will have direct playback of supported files and remuxed playback of supported DV profiles but the core issue of seeking in remuxed DV files will remain.

Is there a compromise possible?

damienl560:
> If this is merged, users will get two options:
> 
> hls.js being enabled will prevent seeking issues with remuxed DV files (#6678) but leave users with broken Dolby Atmos support and transcoding of most audio codecs to AAC.
> 
> With hls.js disabled, users will have direct playback of supported files and remuxed playback of supported DV profiles but the core issue of seeking in remuxed DV files will remain.
> 
> Is there a compromise possible?

Didn't #6530 fix seeking issues? It seemed to have fixed the freezing after seeking problems I was experiencing with mkv files when they were remuxed. So you could just turn off hls.js with my change and turn on the shorter hls segments with #6530 (already in production).

thornbill:
It sounds like the core issues that #6678 addressed have been fixed by other changes... so why not just revert it? Adding a bunch of settings to maybe address edge cases isn't a practice we want to encourage. Playback should just work with minimal tweaking (even if that means remuxing/transcoding).

damienl560:
> It sounds like the core issues that #6678 addressed have been fixed by other changes... so why not just revert it? Adding a bunch of settings to maybe address edge cases isn't a practice we want to encourage. Playback should just work with minimal tweaking (even if that means remuxing/transcoding).

Reverting the change would solve the problem that I encountered since everything was working fine until the rc when it was introduced. However, my understanding is that the original impetus for this change was that its author was experiencing problems with multichannel output when using the native hls player. I have not experienced such problems, but I'm also using different TVs than the author. So it may be that, with their TV model, there would still be issues that can only be addressed by using hls.js.

At any rate, if the goal is that playback should just work out of the box, it is not the case at the moment since the default settings strip Dolby Vision from mkv files that were remuxed properly before and, even when it works with fMP4 (not enabled by default), there is unnecessary transcoding that can only have a negative impact on sound quality and features (e.g. if there is an Atmos soundtrack, it should not be transcoded to AAC ; even a vanilla AC3 soundtrack should not be unnecessarily transcoded, especially for those that still use an SPDIF output since that you'd do AC3 -> AAC -> AC3 for SPDIF output).

Considering that webOS seems to be a bit of a mess, with a lot of different models each with their own idiosyncrasies, it may be necessary to sometimes have manual settings for some of these things, such as that found in #6530. Of course, ideally, we would have enough users with different TV models to do proper tests with proper criteria (e.g. DV should play as DV, Atmos as Atmos, ...).

I understand the need to keep things as streamlined as possible, which is why the new setting is only shown on webOS. It's a way to revert the breaking change without breaking other things for those for whom hls.js was working as (they) expected. The native player would be used out of the box, with an option to have hls.js for those who really need it (instead of making it the default as is the case now, when the native player has been the 'tried and true' option for several major versions).

timitt:
> It sounds like the core issues that #6678 addressed have been fixed by other changes... so why not just revert it? Adding a bunch of settings to maybe address edge cases isn't a practice we want to encourage. Playback should just work with minimal tweaking (even if that means remuxing/transcoding).

It would be very nice if everyone would report their TV model and possibly WebOS version if they have a problem with #6678 .
For now it seems that everyone has LG CX which is older model than C2 (which I have).
It is very much needed for my C2 and apparently also everyone who have C3 or G2 or G3. Maybe even C4 and G4. It is possible that new WebOS patch will fix issues, but at least on C2 it is not yet available. Also it is unknown how well it works for all use cases.

So that toggle is very much needed unless someone has all the LG televisions to actually know what version should have that patch in use or not. For me current 10.11.1 works very well.

> > It sounds like the core issues that #6678 addressed have been fixed by other changes... so why not just revert it? Adding a bunch of settings to maybe address edge cases isn't a practice we want to encourage. Playback should just work with minimal tweaking (even if that means remuxing/transcoding).
> 
> Reverting the change would solve the problem that I encountered since everything was working fine until the rc when it was introduced. However, my understanding is that the original impetus for this change was that its author was experiencing problems with multichannel output when using the native hls player. I have not experienced such problems, but I'm also using different TVs than the author. So it may be that, with their TV model, there would still be issues that can only be addressed by using hls.js.
>
> At any rate, if the goal is that playback should just work out of the box, it is not the case at the moment since the default settings strip Dolby Vision from mkv files that were remuxed properly before and, even when it works with fMP4 (not enabled by default), there is unnecessary transcoding that can only have a negative impact on sound quality and features (e.g. if there is an Atmos soundtrack, it should not be transcoded to AAC ; even a vanilla AC3 soundtrack should not be unnecessarily transcoded, especially for those that still use an SPDIF output since that you'd do AC3 -> AAC -> AC3 for SPDIF output).

I did choose to put WebOS 4 as limit because that was the first version where using hls.js would actually be possible. Unfortunately nobody reported that on CX it was not good thing (and why would anyone have reported it as they has working setup). I got multiple reports saying that the patch was fixing the issues and was needed and many had different LG TVs.

At least on C2 AC3 and EAC3 did not work correctly with native HLS. TV would tell that Dolby Atmos was in use, but AV Receiver received stereo sound (which my AV receiver can make some sort of 5.1, but not actual Dolby Digital). If video did not need HLS (either native or hls.js) then EAC3 and AC3 did play correctly on TV. So even when TV showed "Dolby Atmos" it did not actually play it through AV receiver. Maybe this works on CX.
hls.js uses MediaSource API on background and that is also bit broken on LG C2. It does not play EAC3 or AC3 if video and audio is sent to it in same container. It just reports that media is not supported. It is able to play EAC3 and AC3 if audio is separated as it usually is on HLS streams, but Jellyfin does not separate it (at least it did not on 10.10.x).
So if AV receiver is in use and video needs HLS (either native or hls.js) then EAC3 and AC3 can't be used. Of course it is possible that older CX works differently.

> Considering that webOS seems to be a bit of a mess, with a lot of different models each with their own idiosyncrasies, it may be necessary to sometimes have manual settings for some of these things, such as that found in #6530. Of course, ideally, we would have enough users with different TV models to do proper tests with proper criteria (e.g. DV should play as DV, Atmos as Atmos, ...).

Also it seems to me that #6530 is not needed when hls.js is used. The blame is for LG making a mess like WebOS, but I doubt they care about Jellyfin or similar apps.
Also we can't really fully trust to WebOS version either as can be seen with DTS support. That differs between TV models even if they have same WebOS versions. Option to manually choose to enable or disable hls.js is pretty much the only good option.

> I understand the need to keep things as streamlined as possible, which is why the new setting is only shown on webOS. It's a way to revert the breaking change without breaking other things for those for whom hls.js was working as (they) expected. The native player would be used out of the box, with an option to have hls.js for those who really need it (instead of making it the default as is the case now, when the native player has been the 'tried and true' option for several major versions).

I have no problem having native HLS as default especially if LG actually fixes the problem for newest models. But at least at the time of the patch most of the LG OLED TV models worked better with hls.js (C2, G2, C3, G3 and I think C4 and G4 too). Although if you did not have AV receiver then it was not that important improvement.

damienl560:
Personally, I have a C1 and a B4, both connected to an AV receiver via SPDIF.

With the native hls player, they both output AC3 to the receiver and DV, even without fMP4 enabled.
With hls.js, with fMP4 enabled, I usually get DV output but AC3 is transcoded to AAC, then re-encoded to AC3 by the TV (which is not great because it's best to bitstream the original AC3 stream rather than re-encode it twice, on top of the original encoding)
Without fMP4 enabled, only the HDR fallback is played, not DV.

timitt:
> Personally, I have a C1 and a B4, both connected to an AV receiver via SPDIF.
> 
> With the native hls player, they both output AC3 to the receiver and DV, even without fMP4 enabled. With hls.js, with fMP4 enabled, I usually get DV output but AC3 is transcoded to AAC, then re-encoded to AC3 by the TV (which is not great because it's best to bitstream the original AC3 stream rather than re-encode it twice, on top of the original encoding) Without fMP4 enabled, only the HDR fallback is played, not DV.

I did test using SPDIF and HDMI ARC and they both worked identically (meaning poorly).

Native HLS does need you to disable fMP4 if you wish to use Dolby Vision (it was general consensus with LG WebOS when I made that patch). And it is completely opposite with hls.js. That requires fMP4 to be enabled to have working Dolby Vision. So HDR fallback is expected in that case. Just like enabling fMP4 on native HLS breaks DoVi (atleast for me).

Dolby Vision video plays fine with my TV on native HLS, but sound is always stereo. That is confirmed on LG forums, but apparently it is said to be fixed in future [patch](https://forum.webostv.developer.lge.com/t/stereo-only-for-dolby-vision-video-with-multi-channel-audio-in-hls-streams/23069/31). So when that patch is eventually available then my patch is less relevant (unless there are models that don't get that update at all). B4 probably already has WebOS 25.

Still I think it much better to choose to pull your patch instead of reverting #6678 . Otherwise there are some users that do not have any option to fix the playback except by patching the code themselves.

But since relatively new TV like B4 works fine then it seems to be better to keep hls.js disabled by default.

Also reportedly new WebOS 25 internal media player (which Jellyfin and many or even most apps use) is not particularly stable at the moment. It means crashing videos that are not fault of Jellyfin. Nice...

cmd51247:
WebOS here behaving very erratically as well. Old DV movies (previously trigger DV on 10.10.7 and earlier revs) run fine for the most part. BUT, the second I swap out the external subtitle (download a new format, change the name, etc), the movie ends up going in this weird transcoding but not reported as transcode loop. Behavior is:

On 10.10.7, it would should as "Direct play" whereas now it'll show as "Auto - 120 Mbps" signifying transcoding is happening. But when you go to "Playback info," no transcoding is shown. It'll start playing with "Dolby Vision" showing in the upper right corner with fMP4 enabled, will show standard "HDR" with it disabled.

Again, these are with movies that played with zero issues on 10.10.7. Remuxing the movie physically and/or renaming the subtitle back to its original name doesn't fix the issue, it still does the weird "ghost transcode."

Strange thing is: If I don't touch the previously mkv -> m2ts/ts remuxed files and their corresponding external subtitle files (ass/srt), Dolby Vision continues to trigger fine with "Direct play."

I'm on a LG OLED C3 with latest firmware + Jellyfin running on a Windows server. Again, zero issues before with 10.10.7 or earlier. I seriously hope this code makes it into 10.11.2, will be a huge help all of these WebOS/DV issues.

cmd51247:
Welp, I stand corrected: Even those previously remuxed from mkv -> m2ts videos that show "Direct play" are kicking off transcoding due to the strange AC3 -> AAC behavior mentioned above. What's interesting is, on my C3, it will show "Dolby Vision" in the upper right corner of the screen, show "Direct play," and everything will work almost as if Direct play is 100% in effect (including jumping forward/backwards in the video). 

Log below for reference, really really really hope this gets resolved in the next update:

```
Stream #0:0[0x1015](und): Video: hevc (Main 10) (HDMV / 0x564D4448), yuv420p10le(tv, bt2020nc/bt2020/smpte2084), 3840x2076 [SAR 1:1 DAR 320:173], 23.98 fps, 23.98 tbr, 90k tbn
      Side data:
        DOVI configuration record: version: 1.0, profile: 8, level: 6, rpu flag: 1, el flag: 0, bl flag: 1, compatibility id: 1, compression: 3
  Stream #0:1[0x1100](eng): Audio: truehd (mlpa / 0x61706C6D), 48000 Hz, 5.1(side), s32 (24 bit)
  Stream #0:2[0x1100]: Audio: ac3 (mlpa / 0x61706C6D), 0 channels, fltp
  Stream #0:3[0x1101](eng): Audio: ac3 (AC-3 / 0x332D4341), 48000 Hz, 5.1(side), fltp, 640 kb/s
Stream mapping:
  Stream #0:0 (hevc) -> setparams:default
  overlay_qsv:default -> Stream #0:0 (hevc_qsv)
  Stream #0:1 -> #0:1 (truehd (native) -> aac (libfdk_aac))
Press [q] to stop, [?] for help
[Parsed_subtitles_2 @ 0000026e3632af80] libass API version: 0x1704000
[Parsed_subtitles_2 @ 0000026e3632af80] libass source: tarball: 0.17.4
[Parsed_subtitles_2 @ 0000026e3632af80] Shaper: FriBidi 1.0.16 (SIMPLE) HarfBuzz-ng 10.4.0 (COMPLEX)
[Parsed_subtitles_2 @ 0000026e3632af80] Using font provider directwrite (with GDI)
[hls @ 0000026e2dc84040] Opening 'C:\ProgramData\Jellyfin\Server\cache\transcodes\ac523309e5d93d601e4739a79dbde252-1.mp4' for writing
Output #0, hls, to 'C:\ProgramData\Jellyfin\Server\cache\transcodes\ac523309e5d93d601e4739a79dbde252.m3u8':
  Metadata:
    encoder         : Lavf61.7.100
  Stream #0:0: Video: hevc (hvc1 / 0x31637668), qsv(tv, bt709, progressive), 3840x2076 [SAR 1:1 DAR 320:173], q=2-31, 25896 kb/s, 23.98 fps, 24k tbn
      Metadata:
        encoder         : Lavc61.19.101 hevc_qsv
      Side data:
        Mastering Display Metadata, has_primaries:1 has_luminance:1 r(0.6800,0.3200) g(0.2650,0.6900) b(0.1500 0.0600) wp(0.3127, 0.3290) min_luminance=0.000100, max_luminance=1000.000000
        Content Light Level Metadata, MaxCLL=0, MaxFALL=0
        cpb: bitrate max/min/avg: 25896104/0/25896103 buffer size: 103584412 vbv_delay: N/A
  Stream #0:1: Audio: aac, 48000 Hz, 5.1, s16, 640 kb/s
      Metadata:
        encoder         : Lavc61.19.101 libfdk_aac
      Side data:
        cpb: bitrate max/min/avg: 640000/640000/640000 buffer size: 0 vbv_delay: N/A
frame=    0 fps=0.0 q=0.0 size=N/A time=N/A bitrate=N/A speed=N/A    
frame=    0 fps=0.0 q=0.0 size=N/A time=N/A bitrate=N/A speed=N/A    
[Parsed_subtitles_2 @ 0000026e3632af80] fontselect: (σ╛«Φ╗ƒΘ¢àΘ╗æ, 400, 0) -> ArialMT, 0, ArialMT
[Parsed_subtitles_2 @ 0000026e3632af80] Glyph 0x99AC not found, selecting one more font for (σ╛«Φ╗ƒΘ¢àΘ╗æ, 400, 0)
[Parsed_subtitles_2 @ 0000026e3632af80] fontselect: (σ╛«Φ╗ƒΘ¢àΘ╗æ, 400, 0) -> YuGothicUI-Semibold, 2, YuGothicUI-Semibold
[Parsed_subtitles_2 @ 0000026e3632af80] fontselect: (σ╛«Φ╗ƒΘ¢àΘ╗æ, 700, 0) -> Arial-BoldMT, 0, Arial-BoldMT
[Parsed_subtitles_2 @ 0000026e3632af80] Glyph 0x6559 not found, selecting one more font for (σ╛«Φ╗ƒΘ¢àΘ╗æ, 700, 0)
[Parsed_subtitles_2 @ 0000026e3632af80] fontselect: (σ╛«Φ╗ƒΘ¢àΘ╗æ, 700, 0) -> YuGothicUI-Semibold, 2, YuGothicUI-Semibold
frame=    0 fps=0.0 q=0.0 size=N/A time=N/A bitrate=N/A speed=N/A    
frame=   21 fps=8.3 q=-0.0 size=N/A time=00:00:00.00 bitrate=N/A speed=   0x    
frame=   42 fps= 14 q=-0.0 size=N/A time=00:00:00.87 bitrate=N/A speed=0.289x    
frame=   64 fps= 18 q=-0.0 size=N/A time=00:00:01.79 bitrate=N/A speed=0.507x
```

cmd51247:
> It would be very nice if everyone would report their TV model and possibly WebOS version if they have a problem with #6678 . For now it seems that everyone has LG CX which is older model than C2 (which I have). It is very much needed for my C2 and apparently also everyone who have C3 or G2 or G3. Maybe even C4 and G4. It is possible that new WebOS patch will fix issues, but at least on C2 it is not yet available. Also it is unknown how well it works for all use cases.
> 

As you can see from my comments above, I'm on a C3 with latest firmware, and 10.11.x completely broke my DV playback and introduced the AC3 - AAC transcoding bug. So unfortunately, #6530 also has the potential to break C3's as well.

> At least on C2 AC3 and EAC3 did not work correctly with native HLS. TV would tell that Dolby Atmos was in use, but AV Receiver received stereo sound (which my AV receiver can make some sort of 5.1, but not actual Dolby Digital). If video did not need HLS (either native or hls.js) then EAC3 and AC3 did play correctly on TV. So even when TV showed "Dolby Atmos" it did not actually play it through AV receiver. Maybe this works on CX. hls.js uses MediaSource API on background and that is also bit broken on LG C2. It does not play EAC3 or AC3 if video and audio is sent to it in same container. It just reports that media is not supported. It is able to play EAC3 and AC3 if audio is separated as it usually is on HLS streams, but Jellyfin does not separate it (at least it did not on 10.10.x). So if AV receiver is in use and video needs HLS (either native or hls.js) then EAC3 and AC3 can't be used. Of course it is possible that older CX works differently.
> 

My C3 is hooked up to a Samsung sound bar + surround system, and AC3/EAC3 processed without stereo sound issues on 10.10.7 and earlier versions.

timitt:
> > It would be very nice if everyone would report their TV model and possibly WebOS version if they have a problem with #6678 . For now it seems that everyone has LG CX which is older model than C2 (which I have). It is very much needed for my C2 and apparently also everyone who have C3 or G2 or G3. Maybe even C4 and G4. It is possible that new WebOS patch will fix issues, but at least on C2 it is not yet available. Also it is unknown how well it works for all use cases.
> 
> As you can see from my comments above, I'm on a C3 with latest firmware, and 10.11.x completely broke my DV playback and introduced the AC3 - AAC transcoding bug. So unfortunately, #6530 also has the potential to break C3's as well.

There were C3 and G3 users that did think they needed that patch. What is that latest firmware you have? It WebOS25 is 33.xx and it might be available in some countries. Others have WebOS24 with 23.xx firmwares (I think). On WebOS25 that multichannel playback is supposedly fixed.
AC3 - AAC transcoding is not a bug as I explained in original patch and also in my previous messages here.
#6530 you mention is option to shorten hls segment length. Are you sure you are talking about that or maybe #6678 ?

> > At least on C2 AC3 and EAC3 did not work correctly with native HLS. TV would tell that Dolby Atmos was in use, but AV Receiver received stereo sound (which my AV receiver can make some sort of 5.1, but not actual Dolby Digital). If video did not need HLS (either native or hls.js) then EAC3 and AC3 did play correctly on TV. So even when TV showed "Dolby Atmos" it did not actually play it through AV receiver. Maybe this works on CX. hls.js uses MediaSource API on background and that is also bit broken on LG C2. It does not play EAC3 or AC3 if video and audio is sent to it in same container. It just reports that media is not supported. It is able to play EAC3 and AC3 if audio is separated as it usually is on HLS streams, but Jellyfin does not separate it (at least it did not on 10.10.x). So if AV receiver is in use and video needs HLS (either native or hls.js) then EAC3 and AC3 can't be used. Of course it is possible that older CX works differently.
> 
> My C3 is hooked up to a Samsung sound bar + surround system, and AC3/EAC3 processed without stereo sound issues on 10.10.7 and earlier versions.

Could you tell me more precisely what sound bar it is and what kind of surround system? And how do you know it actually gets real AC3? On my Onkyo AV receiver I see display that says "Dolby Digital" when it actually receives AC3 or EAC3. When it does not get real thing then it shows something else and by default it tries to make it surround. So it might not be instantly obvious by just listening.

Besides, I think we both support the idea of making it possible to disable or enable hls.js?

cmd51247:
> There were C3 and G3 users that did think they needed that patch. What is that latest firmware you have? It WebOS25 is 33.xx and it might be available in some countries. Others have WebOS24 with 23.xx firmwares (I think). On WebOS25 that multichannel playback is supposedly fixed. AC3 - AAC transcoding is not a bug as I explained in original patch and also in my previous messages here. #6530 you mention is option to shorten hls segment length. Are you sure you are talking about that or maybe #6678 ?
>

My C3 is currently on 33.22.15 firmware, and it's the latest available where I'm located. As for the AC3 - AAC transcoding, I did not observe this behavior before 10.11.x. And yes, my apologies, I meant #6678 and not #6530.

And with respect, I don't see how you can say AC3 - AAC transcoding "is not a bug." The exact same files that were playing both video and audio streams direct with 10.10.x are now transcoding the audio to AAC and losing Dolby Atmos in the process. To me, this is a huge downgrade, and definitely falls under "a major bug" in my book.

I understand WebOS25 can be kind of wonky, but these playback issues did not occur until the Jellyfin server upgrade to 10.11.x. All other software/hardware/versions remained constant.
 
> Could you tell me more precisely what sound bar it is and what kind of surround system? And how do you know it actually gets real AC3? On my Onkyo AV receiver I see display that says "Dolby Digital" when it actually receives AC3 or EAC3. When it does not get real thing then it shows something else and by default it tries to make it surround. So it might not be instantly obvious by just listening.
>  

Samsung Q990 with satellites. With Jellyfin 10.10.7, the same remuxed m2ts videos would trigger "Dolby Vision" and "Dolby Atmos," and with TV connected via eARC, my soundbar would also display "Atmos" as expected. I have played multiple files that have both Atmos and Stereo streams, it's very obvious between the Atmos surround vs the Stereo streams being stereo.

> Besides, I think we both support the idea of making it possible to disable or enable hls.js?
> 

Absolutely! Hopefully this patch gets integrated into 10.11.2 as the change from 10.10.7 to 10.11.x thus far has broken DV playback for several movies as well as introduced other odd behaviors on my end. At this point, I do somewhat regret upgrading so quickly, but I'm confident the folks at Jellyfin will get this option/fix implemented quickly.

cmd51247:
Looks like this didn't make it into 10.11.2 either... 😞

agneevX:
The short HLS segments PR was merged in April and made it to the general release in October :/

cmd51247:
I can confirm on my C3 LG OLED running latest WebOS that 10.11.3 has addressed this issue completely, behavior has reverted to what it was on 10.10.7: DV activating along with Atmos without any unnecessary AAC transcoding. Thank you Jellyfin Team!!! 👍😁

damienl560:
> I can confirm on my C3 LG OLED running latest WebOS that 10.11.3 has addressed this issue completely, behavior has reverted to what it was on 10.10.7: DV activating along with Atmos without any unnecessary AAC transcoding. Thank you Jellyfin Team!!! 👍😁

Indeed, @thornbill reverted the change that had been made a while ago to use hls.js with webOS (see #7296). This was his preferred solution rather than make it a user-selectable option. It has fixed my issues too.

Interestingly, my daughter mistakenly accepted the latest updated of webOS on C1 (to 03.53.45) and this has also fixed the compatibility problems between fMP4 and DV. We'll see which option is best once we play files for a bit longer.

w-marco:
I am running Jellyfin on an LG OLED G5, using the latest Firmware (33.22.69). 
Here's my observations from running different versions of Jellyfin (10.11.2, so hls.js & 10.11.5, so native LG hls):

Starting with 10.11.2, it looks as follows:
- Prefer fmp4 enabled: Dolby Vision works, Audio always transcodes to AAC, no matter what
- Prefer fmp4 disabled: Dolby Vision does not work, falls back to regular HDR, Audio always transcodes to AAC

Running 10.11.5 looks like this:
- Prefer fmp4 enabled: Dolby Vision does not work, falls back to regular HDR, AC3 audio works as intended
- Prefer fmp4 disabled: Dolby Vision works, AC3 audio works as intended

So in my case, the only option that actually fully works is using the default LG hls (not hls.js) but with fmp4 disabled. 
Now, hls.js would have the benefit of being able to use fmp4, which might bring improvements, however, as long as it has the audio transcode issue, that isn't really viable either. 

My 2 cents would either be implementing this hls.js switch, or making hls.js the default again with fmp4 also being enabled by default, but only IF the AAC audio transcode issue of hls.js can be fixed.

SjoerdV:
I would still very much prefer to have an option to toggle between native webos functionality and the jellyfin hls.js. Has this been ruled out completely and if so, why? Will JF 10.12 have any improvements on the WebOS compatibility front (or better a (partial) reversal/switch-back, to the 10.10.7 system)?

PR #7228, opened 2025-10-20
horizontal scrollers intentional horizontal movement

dannymichel:
horizontal scrollers intentional horizontal movement at least 2px deltaX and deltaY  2x deltaX for wheel, 2.5x ratio for drag for accidentally catching during vertical page scrolling


thornbill:
Please note the comment at the top of that file:

> NOTE: **This file should not be modified.**
> It is a legacy library that should be replaced at some point.

Issue #7225, opened 2025-10-20
Web client does not report Buffering correctly to SyncPlay


PeachesMLG:
### This issue respects the following points:

- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-web/issues?q=is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

When doing SyncPlay if one client is buffering the video its supposed to call the SyncPlay/Buffering endpoint to tell all the other clients to pause whilst we catch up.

### Reproduction Steps

1. Create a syncplay session on two decives
2. on one device disable the browsers network (you can usually do this via devtools depending on your browser)
3. Observe that when the video is buffering due to no new updates, no SyncPlay/Buffering POST request is made

(Incidently there is also no buffering animation which may or may not be a regression, I veigly remember this being in 10.10.7)

### Expected/Actual behaviour

SyncPlay/Buffering POST request to be made when we are buffering

### Server version

10.11.0

### Web version

10.11.0

### Build version

10.11.0

### Platform

Linux

### Browser

Firefox

PeachesMLG:
So im a bit out of my comfort zone here, since I have little knowledge of HLS video or the existing code base.

However after messing around for a bit, I have come up with a fix: https://github.com/PeachesMLG/jellyfin-web/commit/aa4f6069ca5a7592d2da304b84829fd172228421 by manually binding hlsEvents which then calls the onWaiting method.

Im not sure why this broke in the first place, and even reverting back multiple versions (even back to 10.9) I couldnt find a version this worked on.


As I said I am massively unfamiliar with hlsPlayers aswell as the existing codebase hence why I havent submited this as a pull request since im sure its far from ideal code (I havent even written any unbind logic), however hopefully it can give some insight into whats going wrong to somone smarter than myself.


Note: I am happy to do the code myself if somone can give me some guidance if im going down the right path (e.g. is this binding supposed to happen somewhere else but is failing?)

PeachesMLG:
So I figured this out and raised a PR to fix this. turns out the `'waiting'` event wasn’t triggering correctly because the handler’s `this` context was lost — it was defined as a normal function instead of an arrow function, so `this` no longer referred to the video player instance.

PR #7244, opened 2025-10-22
Fixed issue where waiting event is not being called correctly

PeachesMLG:
**Changes**
Converts the onWaiting method to arrow function to fix the `this` context so that events can be triggered correctly

Fixes: https://github.com/jellyfin/jellyfin-web/issues/7225


dmitrylyzo (approved):
This seems to have been overlooked in https://github.com/jellyfin/jellyfin-web/pull/1091

You need to rebase this change onto `release-10.11.z`.

PeachesMLG:
Appologies, I clearly completly messed this up so badly github just closed the pr (I might have accidently deleted all the commits so there is no diff)

I have raised a new PR here: https://github.com/jellyfin/jellyfin-web/pull/7245

PR #7245, opened 2025-10-22
Fixed issue where waiting event is not being called correctly

PeachesMLG:
**Changes**
Converts the onWaiting method to arrow function to fix the `this` context so that events can be triggered correctly

Fixes: https://github.com/jellyfin/jellyfin-web/issues/7225

### DIFF
--- a/src/plugins/htmlVideoPlayer/plugin.js
+++ b/src/plugins/htmlVideoPlayer/plugin.js
@@ -1066,7 +1066,7 @@ export class HtmlVideoPlayer {
 
     onWaiting = () => {
         Events.trigger(this, 'waiting');
-    }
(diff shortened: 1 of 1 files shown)


src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1064,7 +1064,7 @@ export class HtmlVideoPlayer {
         Events.trigger(this, 'pause');
     };
 
-    onWaiting() {
+    onWaiting = () => {
         Events.trigger(this, 'waiting');
     }
```
nielsvanvelzen:
```suggestion
    };
```
for the linter issue

PR #7258, opened 2025-10-26
Add Titan OS detection

nielsvanvelzen:
**Changes**

- Add detection for Titan OS (a chromium based TV OS in Europe)
- Support displaying Titan OS in dashboard
  

**Issues**

### DIFF
--- a/src/scripts/browser.js
+++ b/src/scripts/browser.js
@@ -3,23 +3,23 @@ function isTv() {
     const userAgent = navigator.userAgent.toLowerCase();
 
     // The OculusBrowsers userAgent also has the samsungbrowser defined but is not a tv.
-    if (userAgent.indexOf('oculusbrowser') !== -1) {
+    if (userAgent.includes('oculusbrowser')) {
         return false;
     }
 
-    if (userAgent.indexOf('tv') !== -1) {
+    if (userAgent.includes('tv')) {
         return true;
     }
 
-    if (userAgent.indexOf('samsungbrowser') !== -1) {
+    if (userAgent.includes('samsungbrowser')) {
         return true;
     }
 
-    if (userAgent.indexOf('viera') !== -1) {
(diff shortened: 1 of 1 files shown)


dmitrylyzo:
Please update declaration: https://github.com/jellyfin/jellyfin-web/blob/0cd8dcf946fe039a7d27287180ecd25820582f5f/src/scripts/browser.d.ts#L8

src/scripts/browser.js
```diff
@@ -19,6 +19,10 @@ function isTv() {
         return true;
     }
 
+    if (userAgent.indexOf('titanos') !== -1) {
```
thornbill:
```suggestion
    if (userAgent.includes('titanos')) {
```

nielsvanvelzen: While I don't mind making this change, I did it this way to be consistent with the other if statements. Do you want me to change all of those to use `includes()` as well?

src/scripts/browser.js
```diff
@@ -209,6 +214,7 @@ const uaMatch = function (ua) {
         || /(iphone)/.exec(ua)
         || /(windows)/.exec(ua)
         || /(android)/.exec(ua)
+        || /(titanos)/.exec(ua)
```
dmitrylyzo, line 217: `match` and `platformMatch` both set `browser.titanos`:
https://github.com/jellyfin/jellyfin-web/blob/6ce3e579c200d95fe299c736da664d8823c412c8/src/scripts/browser.js#L256-L264

Currently, there's no harm. Just a note.

nielsvanvelzen: I think that's what we want, the platform itself is a browser based on chromium.

src/scripts/browser.js
thornbill: You can... it's just an old pattern that hasn't been updated here yet.

PR #7259, opened 2025-10-26
Avoid native browser alerts

nielsvanvelzen:
Custom modals don't block the browser thread and look a lot better, additionally native alert/promp/confirm dialogs don't work in most TV based browsers. While this change is specifically targetted towards Titan OS, I don't see any reason to keep a workaround list instead of always using our custom modal.

**Changes**
Remove "canUseNativeAlert" check and always show our own dialog component.

**Issues**


dmitrylyzo:
There's also a confirmation dialog.

dmitrylyzo (approved):

PR #7271, opened 2025-10-28
Fix tuner devices list

viown:
**Changes**
Fix tuner devices list not scaling correctly

**Issues**
Fixes https://github.com/jellyfin/jellyfin-web/issues/7265


thornbill:
The `Form` import needs removed. Looks good otherwise!

thornbill (approved):

PR #7282, opened 2025-11-01
Upgrade to NPM 11 and Node 24 active LTS

larouxn:
**Related**
- https://github.com/jellyfin/jellyfin-web/pull/7207
- https://github.com/jellyfin/jellyfin-packaging/pull/116

**Changes**

~Simply updates the NPM upperbound allowed version to `<12.0.0` so that it allows NPM 11.~ Upgrades the entire repo to use Node 24 and thus NPM 11. The reason for doing is because Node 24, as of [v24.11.0](https://github.com/nodejs/node/releases/tag/v24.11.0), has been promoted to be the new LTS release _and_ Node 24 includes NPM 11 by default. Hence in order to support Node 24 we need to support NPM 11. Node 20 is also coming up on EOL this April. ([source](https://nodejs.org/en/about/previous-releases))
Screenshot From 2026-01-11 12-00-02

~The primary change of concern in NPM 11 is that it drops support for older Node versions. That said, 1. all Node older than v20 are not supported ([source](https://nodejs.org/en/about/previous-releases)) and 2. people can still use NPM >=9.6.4 and NPM v10.x.x if they need older Nodes.~

Required for the possible Node upgrade portion of bring the Flatpak up to date.
- https://github.com/flathub/org.jellyfin.JellyfinServer/pull/684

**Testing**

Tested production builds and various checks.

Screenshot From 2026-01-11 12-13-10
Screenshot From 2026-01-11 12-13-17
image
Screenshot From 2026-01-11 12-15-50
Screenshot From 2026-01-11 12-16-01

Also tried testing the devcontainer since it needed an image upgrade but I hit an error. That said, the same error happens on my machine on `master` branch so. :man_shrugging: 

image


scallaway:
Is there any reason this has been delayed? Would be really cool to see this in since it was an immediate blocker for me when pulling down the repository to get up and running.

larouxn:
Just rebased on `master` branch for good measure. :rocket:

thornbill:
The issue with npm 11 is that it introduces breaking changes to the lockfile format without properly bumping the spec to a new version. We need to ensure everywhere that builds or otherwise depends on web is bumped simultaneously. This is especially true for the packaging repo and ci in this repo.

.github/workflows/__quality_checks.yml
```diff
@@ -48,9 +48,9 @@ jobs:
           show-progress: false
 
       - name: Setup node environment ⚙️
-        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        uses: actions/setup-node@395ad3262231945c25e8478fd5baf05154b1d79f # v6.1.0
```
larouxn, line 51: For some reason CI is using the older `actions/setup-node` (v4.4.0) despite it being changed here.
Image

larouxn: Seems workflows may be using `master` so we need to ship https://github.com/jellyfin/jellyfin-web/pull/7207 first before this PR.

thornbill:
Just a reminder that the [packaging repo](https://github.com/jellyfin/jellyfin-packaging) will also need updating.

We should update the CI workflows to use `node-version-file` to make future updates a little easier. ([Related SDK PR](https://github.com/jellyfin/jellyfin-sdk-typescript/pull/982))

larouxn:
> Just a reminder that the [packaging repo](https://github.com/jellyfin/jellyfin-packaging) will also need updating.

Thanks for the heads up. PR is up here.
- https://github.com/jellyfin/jellyfin-packaging/pull/116

> We should update the CI workflows to use `node-version-file` to make future updates a little easier. ([Related SDK PR](https://github.com/jellyfin/jellyfin-sdk-typescript/pull/982))

Great idea. Workflows have been migrated to use `node-version-file`. :heavy_check_mark:

larouxn:
Thanks for the merge @thornbill. Apologies but I made one typo. Fix: https://github.com/jellyfin/jellyfin-web/pull/7522 :pray:

Issue #7231, opened 2025-10-21
[Tizen 10.11.0] Jellyfin remuxes instead of direct playing of DV 7.6 with HDR fallback or with HDR+


dusanhatvani:
### Description of the bug

After upgrading to version 10.11.0 Jellyfin is remuxing content with DV profile 7.6 on my Tizen TV.
Previously on 10.10.7 Jellyfin used to directly play such content with no issues.
In such cases Jellyfin used to utilize either HDR10 fallback or HDR10+ if present.

The main problem is then tied to TRUE HD direct play, which is possible when eARC soundbar is connected to Tizen TV. This forced remuxing breaks the TRUE HD playback. If audio transcoding is enabled, jellyfin transcodes the TRUE HD to AAC. If it is not enabled, TV is reporting TRUE HD audio playback, but no sound comes out.

Similar issue if not the same occured on version 10.9.2 last year 
https://github.com/jellyfin/jellyfin/issues/11794#issue-2310987953

### Reproduction steps

1. Play any 4K remux with DV profile 7.6 with HDR fallback or HDR10+ on Samsung Tizen TV
2. Observe remuxing taking place instead of direct playing.

### What is the current _bug_ behavior?

Remuxing/Transcode instead of direct play

### What is the expected _correct_ behavior?

Jellyfin should detect HDR fallback or present HDR10+ profile and should direct play the file

### Jellyfin Server version

10.10.0+

### Specify the build version

10.11.0

### Environment

```markdown
- OS: Linux
- Virtualization: docker
- Clients: Tizen
```

### Jellyfin logs

```shell
Attached
```

### FFmpeg logs

```shell
Attached
```

### Additional information

[ffmpeg log.txt](https://github.com/user-attachments/files/23014609/ffmpeg.log.txt)
[jellyfin logs.txt](https://github.com/user-attachments/files/23014608/jellyfin.logs.txt)

gnattu:
If you believe it is due to TrueHD:

<img width="834" height="263" alt="Image" src="https://github.com/user-attachments/assets/8783c065-86a8-488b-a79d-8c8e62d765bd" />

[...]

PR #7283, opened 2025-11-02
Restrict recently added max-32-streams limit to Tizen < v6.5

kinke:
v10.11 introduced that limit for Tizen in general, causing all formerly direct-playable media with more than 32 streams to be remuxed - one cause of regression #7231.

According to https://github.com/jellyfin/jellyfin-tizen/issues/289#issuecomment-2581270727, Tizen v5 only supports max 32 streams. This limit seems to have been lifted in v6, according to https://github.com/jellyfin/jellyfin-web/issues/7231#issuecomment-3474685133. I'm using Tizen v9 and don't have any problems with around 100 streams.

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -924,7 +924,7 @@ export default function (options) {
 
     profile.ContainerProfiles = [];
 
-    if (browser.tizen) {
+    if (browser.tizenVersion < 6.5) {
         // Tizen doesn't support more than 32 streams in a single file
         profile.ContainerProfiles.push({
             Type: 'Video',


src/scripts/browserDeviceProfile.js
```diff
@@ -924,7 +924,7 @@ export default function (options) {
 
     profile.ContainerProfiles = [];
 
-    if (browser.tizen) {
+    if (browser.tizen && browser.tizenVersion < 6) {
```
dmitrylyzo: According to https://github.com/jellyfin/jellyfin-web/issues/7231#issuecomment-3474161085, QN90B is Tizen 6.5. We need someone to test this on Tizen 6.0.

You can omit `browser.tizen &&`, because `browser.tizenVersion` is only defined when `browser.tizen` is `true`.

kinke: Oh, I didn't realize that there are apparently weird v5.5 and v6.5 versions.

If it helps getting this in quicker, without having to find a 6.0 tester, I can change it to `< 6.5`.

kinke: Okay, changed it to `< 6.5` to be on the safe side.

kinke:
Ping, would be nice to have this land in v10.11.3.

thornbill (approved):

PR #7284, opened 2025-11-02
Enable HDR10+ and DoVi fallback profiles

kinke:
The 10.11 server introduced new `VideoRangeType` values in https://github.com/jellyfin/jellyfin/pull/13277, but the web client apparently wasn't updated accordingly. This is one cause of regression #7231, leading e.g. to `DOVIWithHDR10Plus` media being remuxed instead of direct-played.

I've tested `HDR10Plus`, `DOVIWithHDR10` and `DOVIWithHDR10Plus` media successfully on Tizen v9. The author of the linked issue additionally tested `DOVIWithELHDR10Plus` (and I think `DOVIWithEL`) and `DOVIInvalid` on Tizen 6.

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -1188,7 +1188,18 @@ export default function (options) {
         vp9VideoRangeTypes += '|HDR10';
         av1VideoRangeTypes += '|HDR10';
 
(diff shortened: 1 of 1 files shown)


nyanmisaka:
@gnattu can you take a look?

PR #7285, opened 2025-11-02
Playback info: Promote out of settings menu to OSD footer icon

kinke:
To make it less cumbersome to toggle it on/off. On TV (the only platform I've tested this on), this saves about 4 key presses on the remote, for each toggling. In particular, closing it takes a single OK-press while the OSD is still visible (from toggling it on).

### DIFF
--- a/src/controllers/playback/video/index.html
+++ b/src/controllers/playback/video/index.html
@@ -87,6 +87,9 @@
                 <button is="paper-icon-button-light" class="btnVideoOsdSettings autoSize" title="${Settings}">
                     <span class="largePaperIconButton material-icons settings" aria-hidden="true"></span>
                 </button>
+                <button is="paper-icon-button-light" class="btnPlayerStats autoSize" title="${PlaybackData}">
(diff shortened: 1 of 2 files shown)


src/controllers/playback/video/index.html
```diff
@@ -87,6 +87,9 @@ <h3 class="osdTitle"></h3>
                 <button is="paper-icon-button-light" class="btnVideoOsdSettings autoSize" title="${Settings}">
                     <span class="largePaperIconButton material-icons settings" aria-hidden="true"></span>
                 </button>
+                <button is="paper-icon-button-light" class="btnPlayerStats autoSize" title="${PlaybackData}">
```
kinke, line 90: I've placed the new info button/icon to the right of the settings button. That used to be the right-most button on TV, in the bottom-right corner; that's the info now. The number of key presses to get to the settings button is hence unchanged, so muscle memory shouldn't be affected.

thornbill:
Most people do not care or should ever need to view the playback info, so I don't see a point in letting it consume valuable real estate here.

PR #7286, opened 2025-11-02
Playback info: Add frame rate, bit depth, color space, color transfer and pixel format

kinke:
This is an admittedly totally subjective list of stats that I find useful and have so far had to look up in the 'Media info' instead (which is currently barely usable on TV for me, cannot scroll down and so only see about half of the video stream stats).

Also display the original `VideoRangeType` for DoVi media, not just the 'nice' `VideoDoViTitle` (which e.g. currently doesn't differentiate between HDR10 and HDR10+ fallbacks).

### DIFF
--- a/src/components/playerstats/playerstats.js
+++ b/src/components/playerstats/playerstats.js
@@ -289,13 +289,6 @@ function getMediaSourceStats(session, player) {
         });
     }
 
-    if (videoStream.BitDepth) {
-        sessionStats.push({
-            label: globalize.translate('MediaInfoBitDepth'),
-            value: `${videoStream.BitDepth} bit`
-        });
-    }
-
     if (videoStream.VideoRangeType) {
         sessionStats.push({
             label: globalize.translate('LabelVideoRangeType'),


src/components/playerstats/playerstats.js
```diff
@@ -274,17 +274,55 @@ function getMediaSourceStats(session, player) {
...
+    }
+
     if (videoStream.VideoRangeType) {
         sessionStats.push({
             label: globalize.translate('LabelVideoRangeType'),
-            value: videoStream.VideoDoViTitle || videoStream.VideoRangeType
+            value: videoStream.VideoDoViTitle ?
+                `${videoStream.VideoDoViTitle} [${videoStream.VideoRangeType}]` :
+                videoStream.VideoRangeType
+        });
```
nyanmisaka, line 303: It would be better to list `VideoDoViTitle` on a separate line and add `translate('LabelVideoDoViTitle')`

src/components/playerstats/playerstats.js
```diff
@@ -274,17 +274,55 @@ function getMediaSourceStats(session, player) {
...
+    }
+
+    if (videoStream.ColorTransfer) {
+        sessionStats.push({
+            label: globalize.translate('LabelColorTransfer'),
+            value: videoStream.ColorTransfer
+        });
+    }
+
```
nyanmisaka, line 319: `ColorPrimaries` is missing.

src/components/playerstats/playerstats.js
```diff
@@ -274,17 +274,55 @@ function getMediaSourceStats(session, player) {
...
     if (videoStream.BitRate) {
         sessionStats.push({
             label: globalize.translate('LabelVideoBitrate'),
             value: getDisplayBitrate(videoStream.BitRate)
         });
     }
 
+    if (videoStream.BitDepth) {
```
nyanmisaka: `BitDepth` is useless here because `PixelFormat` and even `Profile` already provide the depth value.

```
yuv4xxp -> [8] <- Main
yuv4xxp10 -> [10] <- Main10
```

src/components/playerstats/playerstats.js
```diff
@@ -274,17 +274,55 @@ function getMediaSourceStats(session, player) {
         });
     }
 
+    const frameRate = videoStream.AverageFrameRate || videoStream.RealFrameRate;
+    if (frameRate) {
+        sessionStats.push({
+            label: globalize.translate('Framerate'),
```
nyanmisaka, line 280: Better to use `Label*` here to keep consistency: `LabelFramerate`.

src/components/playerstats/playerstats.js
```diff
@@ -274,17 +274,55 @@ function getMediaSourceStats(session, player) {
...
+    if (videoStream.ColorTransfer) {
+        sessionStats.push({
+            label: globalize.translate('LabelColorTransfer'),
+            value: videoStream.ColorTransfer
+        });
+    }
+
+    if (videoStream.PixelFormat) {
+        sessionStats.push({
+            label: globalize.translate('MediaInfoPixelFormat'),
```
nyanmisaka, line 329: Better to use `Label*` here to keep consistency: `LabelPixelFormat`.

src/components/playerstats/playerstats.js
kinke: Yeah, useless for me too, but I guess there's a good number of people who cannot derive it that way, so I felt like including it explicitly.

src/components/playerstats/playerstats.js
kinke: It doesn't exist; I've tried to use the `Label…` keys where available, and otherwise fell back to the existing strings used for the media info dialog.

src/components/playerstats/playerstats.js
kinke: Okay, as in the media info.

src/components/playerstats/playerstats.js
kinke: Okay, added.

src/components/playerstats/playerstats.js
nyanmisaka: The list is already long enough compared to YouTube's "Stats for nerds", especially for mobile clients. I once considered simplifying them. IMO adding duplicate items won't help much. Besides, aside from developers and advanced users, almost no one will actually check this.

src/components/playerstats/playerstats.js
nyanmisaka: You can add it here: https://github.com/jellyfin/jellyfin-web/blob/master/src/strings/en-us.json

kinke: And what about the 99 other language files?

src/components/playerstats/playerstats.js
kinke: Okay, removed.

src/components/playerstats/playerstats.js
nyanmisaka: It is handled by the community in Weblate https://translate.jellyfin.org/projects/jellyfin/jellyfin-web/

src/components/playerstats/playerstats.js
nyanmisaka: My plan is to utilize horizontal space, rather than vertically expanding the items.
This can prevent the overlay from extending beyond the screen on mobile devices.

osd

kinke: Looks good to me! :+1:

nyanmisaka: You can give #7291 a try.

kinke:
Closing in favor of https://github.com/jellyfin/jellyfin-web/pull/7291.

PR #7294, opened 2025-11-04
Use experimental layout by default

thornbill:
**Changes**
Enables the "experimental" layout by default

**Issues**
N/A

### DIFF
--- a/src/RootAppRouter.tsx
+++ b/src/RootAppRouter.tsx
@@ -13,13 +13,16 @@ import { STABLE_APP_ROUTES } from 'apps/stable/routes/routes';
 import { WIZARD_APP_ROUTES } from 'apps/wizard/routes/routes';
 import AppHeader from 'components/AppHeader';
 import Backdrop from 'components/Backdrop';
+import { SETTING_KEY as LAYOUT_SETTING_KEY } from 'components/layoutManager';
 import BangRedirect from 'components/router/BangRedirect';
 import { createRouterHistory } from 'components/router/routerHistory';
+import { LayoutMode } from 'constants/layoutMode';
+import browser from 'scripts/browser';
 import appTheme from 'themes/themes';
 import { ThemeStorageManager } from 'themes/themeStorageManager';
 
-const layoutMode = localStorage.getItem('layout');
-const isExperimentalLayout = layoutMode === 'experimental';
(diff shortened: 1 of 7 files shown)


src/RootAppRouter.tsx
```diff
@@ -13,13 +13,16 @@ import { STABLE_APP_ROUTES } from 'apps/stable/routes/routes';
...
 import BangRedirect from 'components/router/BangRedirect';
 import { createRouterHistory } from 'components/router/routerHistory';
+import { LayoutMode } from 'constants/layoutMode';
+import browser from 'scripts/browser';
 import appTheme from 'themes/themes';
 import { ThemeStorageManager } from 'themes/themeStorageManager';
 
-const layoutMode = localStorage.getItem('layout');
-const isExperimentalLayout = layoutMode === 'experimental';
+const layoutMode = browser.tv ? LayoutMode.Tv : localStorage.getItem(LAYOUT_SETTING_KEY);
```
pedrormonteiro, line 24: setting the layoutMode only affects localStorage? there is no "server" config for it?

src/components/displaySettings/displaySettings.template.html
```diff
@@ -169,10 +169,10 @@ <h2 class="sectionTitle">
     <div class="selectContainer fldDisplayMode hide">
         <select is="emby-select" class="selectLayout" label="${LabelDisplayMode}">
             <option value="">${Auto}</option>
+            <option value="experimental">${Experimental}</option>
             <option value="desktop">${Desktop}</option>
             <option value="mobile">${Mobile}</option>
             <option value="tv">${TV}</option>
```
pedrormonteiro, line 175: you could use `LayoutMode` here

src/RootAppRouter.tsx
thornbill: Yes, layout mode is client side only.

src/components/displaySettings/displaySettings.template.html
thornbill: That isn't really available in the html templates. We only really do substitution on translation keys.

thornbill:
Closing this as I think an alternative approach that allows us to continue using "experimental" mode for larger changes would be better.

PR #7296, opened 2025-11-04
Revert "Enable use of hls.js when LG WebOS 4 or newer is used."

thornbill:
This reverts commit 85fc9de45f774f5f4a4baf78f226544dcacc8775.

**Changes**
This should resolve the multitude of issues with DV on webOS in 10.11

**Issues**
Fixes https://github.com/jellyfin/jellyfin-webos/issues/310

### DIFF
--- a/src/components/htmlMediaHelper.js
+++ b/src/components/htmlMediaHelper.js
@@ -48,12 +48,6 @@ export function enableHlsJsPlayer(runTimeTicks, mediaType) {
         return false;
     }
 
-    // Native HLS support in WebOS only plays stereo sound. hls.js works better, but works only on WebOS 4 or newer.
-    // Using hls.js also seems to fix fast forward issues that native HLS has.
-    if (browser.web0sVersion >= 4) {
-        return true;
-    }
-
     // The native players on these devices support seeking live streams, no need to use hls.js here
     if (browser.tizen || browser.web0s) {
         return false;
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -137,14 +137,6 @@ function supportsEac3(videoTestElement) {
 }
 
 function supportsAc3InHls(videoTestElement) {
-    // We use hls.js on WebOS 4 and newer and hls.js uses Media Sources Extensions (MSE) API.
-    // On WebOS MSE does support AC-3 and EAC-3 only on audio mp4 file but not on audiovideo mp4
-    // therefore until audio and video is not separated when generating stream and m3u8 this should
-    // return false.
-    if (browser.web0sVersion >= 4) {
-        return false;
-    }
-
     if (browser.tizen || browser.web0s) {
         return true;
     }


timitt:
I really don't understand this decision. You were given pull request that would give option to choose between hls.js and native HLS. But you decided to just revert the my pull request that was kind of useful for many of us.

Now I am back in the situation where every single HLS stream is played in stereo.

viown:
I can't speak for the author, but ideally playback should work without requiring the user to change through settings.

The average user doesn't know what HLS is and why they have to change it to get their media working.

timitt:
Desperately I downloaded newest firmware from LG US website as it is compatible with my european tv. Now EAC3 and AC3 work fine, but some Dolby Vision mkv videos are just flickering mess.

For me the decision to revert was absolute train wreck and there is no way back unless LG fixes their player or possibly I patch hls.js back to my version (not sure if it fixes these new issues). Though if I repackage mkv to mp4 with exact same video it plays fine.

andrebower:
> Now I am back in the situation where every single HLS stream is played in stereo.

I have the same issue now!

thornbill:
> I really don't understand this decision. You were given pull request that would give option to choose between hls.js and native HLS. But you decided to just revert the my pull request that was kind of useful for many of us.
>
> Now I am back in the situation where every single HLS stream is played in stereo.

We received **many** reports that the original change in 10.11 completely broke playback for webOS users. Given the choice between having audio play as stereo instead of proper multichannel vs completely failing to play content at all it was pretty clear that we should not keep the hls.js change (by default).

The problem with the pull request to add hls.js as an option is threefold:

1. As Max mentioned the goal should be for playback to just work. No one should have to modify settings to get basic playback working.
2. A new option really represents a new feature and not just a simple bugfix which makes it ineligible for including in a patch/bugfix release.
3. The new option requires new translations. Translation changes are not included in patch/bugfix releases.

It **could** be considered for the next major release if a better solution is not found in the meantime.

timitt:
> > I really don't understand this decision. You were given pull request that would give option to choose between hls.js and native HLS. But you decided to just revert the my pull request that was kind of useful for many of us.
> > Now I am back in the situation where every single HLS stream is played in stereo.
> 
> We received **many** reports that the original change in 10.11 completely broke playback for webOS users. Given the choice between having audio play as stereo instead of proper multichannel vs completely failing to play content at all it was pretty clear that we should not keep the hls.js change (by default).
> 
> The problem with the pull request to add hls.js as an option is threefold:
> 
> 1. As Max mentioned the goal should be for playback to just work. No one should have to modify settings to get basic playback working.
> 2. A new option really represents a new feature and not just a simple bugfix which makes it ineligible for including in a patch/bugfix release.
> 3. The new option requires new translations. Translation changes are not included in patch/bugfix releases.
> 
> It **could** be considered for the next major release if a better solution is not found in the meantime.

I appreciate explaining, but if you only choose one then there are WebOS TV owners that can't really use Jellyfin. Obviously LG is the one to blame as they seem to constantly break things in new versions while fixing other problems and at the same time things work differently between diffent TV models.
It is nice though to have working playback without tuning settings, but it just isn't possible as there seems to be no way to know what settings are needed as there is no way for Jellyfin to know which model of TV one has and not even exact version of WebOS. Some reported that Atmos actually worked for them while my system got only stereo. That makes me think that maybe only HDMI ARC was broken and not eARC or something like that. But Jellyfin can not know which one is used either. There would have been solution for that AAC too, but it would have needed more rework to Jellyfin HLS streaming logic so it was completely unrealistic for me to even try to make that change.

Also at least some of the reports of videos that did not play in 10.11. were not fixed after hls.js was not used anymore ([#14581](https://github.com/jellyfin/jellyfin/issues/14581)). Most of the complains were that AC3/EAC3 (including Atmos) was reencoded to AAC when HLS was needed. And I found that much lesser issue than dropping to stereo sound. Of course people who has Atmos working found that worse.

Anyway, I solved the problem by buying Android TV box (and also new AV receiver) to get everything to work as I just found it too much hassle to make a personal build for Jellyfin every time it gets new release. I had many DV videos that failed to show actual image when native HLS was in use after updating to WebOS 25 (if I repackage them to MP4 then they play fine so video). All of those work without any problem with this new Android TV box.

jacekowski:
@thornbill You have also received reports that not using hls.js breaks things. Toggleable option is the best way out here, default can be left at internal HLS.

thornbill:
Ok. Open a PR targeting master that does so. Like I said in my last comment.

PR #7298, opened 2025-11-05
Relax dynamic HDR device Profile

gnattu:
HDR10+ are designed to be backward-compatible with HDR10 which means we can just advertise HDR10+ as supported for HDR10 supported clients. This will avoid the server to force a remux to remove the HDR10+ metadata to ensure compatibility. At least the server-side removal is not needed for major web browsers and the tvs that uses the web client.

For TVs, we need to relax the device profile even more by exposing not-supported but can be played ranges. This is due to the HLS support on TVs are often lacking features and the remux on server side would end up removing more than just the dynamic range. I don't personally own any of the TVs, so I'm not sure if this relaxed profile would break anything though.

**Changes**

**Issues**

Fixes https://github.com/jellyfin/jellyfin/discussions/15223

Fixes https://github.com/jellyfin/jellyfin/issues/15271

Fixes https://github.com/jellyfin/jellyfin-webos/issues/310

Related: https://github.com/jellyfin/jellyfin-web/pull/7296

The revert is still needed to retain full features, but the profile also has to be relaxed.

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -1193,6 +1193,9 @@ export default function (options) {
             // Tizen TV does not support Dolby Vision at all, but it can safely play the HDR fallback.
             // Advertising the support so that the server doesn't have to remux.
             hevcVideoRangeTypes += '|DOVIWithHDR10|DOVIWithHDR10Plus|DOVIWithEL|DOVIWithELHDR10Plus|DOVIInvalid';
+            // Although no official tools exist to create AV1+DV files yet, some of our users managed to use community tools to create such files.
+            // These files should also be playable on Tizen TVs.
+            av1VideoRangeTypes += '|DOVIWithHDR10|DOVIWithHDR10Plus|DOVIWithEL|DOVIWithELHDR10Plus|DOVIInvalid';
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/scripts/browserDeviceProfile.js
```diff
@@ -1184,12 +1184,15 @@ export default function (options) {
...
-        av1VideoRangeTypes += '|HDR10';
+        // HDR10+ videos can be safely played on all HDR10 capable devices, just without the dynamic metadata.
+        hevcVideoRangeTypes += '|HDR10|HDR10Plus';
+        vp9VideoRangeTypes += '|HDR10|HDR10Plus';
+        av1VideoRangeTypes += '|HDR10|HDR10Plus';
 
         if (browser.tizenVersion >= 3 || browser.vidaa) {
-            hevcVideoRangeTypes += '|DOVIWithHDR10';
+            // Tizen TV does not support Dolby Vision at all, but it can safely play the HDR fallback.
+            // Advertising the support so that the server don't have to do the remux.
```
Shadowghost:
```suggestion
            // Advertising the support so that the server doesn't have to remux.
```

krishhari2105:
Tizen TVs above v6 support AV1 as far as I know. we could add av1videorange type too in the relaxed profile types.

thornbill:
@gnattu does this supersede https://github.com/jellyfin/jellyfin-web/pull/7284 ?

gnattu:
> @gnattu does this supersede #7284 ?

Yes it would. I supplied extra av1 ranges and now this will supersede that.

PR #7299, opened 2025-11-05
Fix sonarcloud findings

pedrormonteiro:
**Changes**
This PR encompasses changes to multiple files, based on current practices and focus specifically on issues flagged on SonarCloud.

**Issues**
None that I know of.


pedrormonteiro:
This PR focus on high-level SonarCloud issues. Sadly, I am not 100% familiar with all flows affected during this refactor and would appreciate it if someone could help me test :)

PR #7300, opened 2025-11-05
Add playback speed controls for audiobooks

caleb-conner:
**Changes**
Adds playback speed controls (0.5x to 2.5x) for audiobook playback and a settings button to the audio player's now playing bar. Playback speed preference is persisted across sessions using sessionStorage.

**Issues**
Adds feature parity with other audiobook apps (Audible, LibreVox) for playback speed control.

### DIFF
--- a/src/components/nowPlayingBar/nowPlayingBar.js
+++ b/src/components/nowPlayingBar/nowPlayingBar.js
@@ -90,6 +90,8 @@ function getNowPlayingBarHtml() {
 
     html += `<button is="paper-icon-button-light" class="openLyricsButton mediaButton hide" title="${globalize.translate('Lyrics')}"><span class="material-icons lyrics" style="top:0.1em" aria-hidden="true"></span></button>`;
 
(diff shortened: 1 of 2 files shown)


jordanmichaelrushing:
I've been looking into this functionality and saw this PR. Unfortunately I think there's a fatal flaw with it: It should work great for Android and Web, however you'll have issues with iOS where it will sound like it's stuttering or skipping due to pitch control issues. The solution is unfortunately to either:
1. Transcode directly fron the server itself (so the client is always seeing it as playing back at "1.0x")
2. Web Audio API + client-side time-stretching library. However this adds a library dependency (~100KB), CPU-heavy on mobile, non-trivial to wire into the existing audio graph, potential latency/buffer issues

I'm opting for option 1 for a better long term solution. I'll post a PR in this repo and the server repo soon. Not sure why this PR was never reviewed in the past 6 months or so, but I think it should unfortunately be closed out.

thornbill:
While it may not have existed yet at the time this was opened, this PR clearly violates our llm policy.

PR #7315, opened 2025-11-08
Add frame scrubbing controls

### LINKED ISSUE #3087: Allow per-frame playback navigation

flying-sheep:
**Changes**
Register `,` and `.` as controls to scrub frame-by-frame.

These are the same keys that YouTube uses for this functionality, which seems to be the interface you’re mirroring with these keybinds.

**Issues**
Resolves #3087, closes #7132
Implements https://features.jellyfin.org/posts/3545/seeking-video-frame-by-frame

TODO:
- [x] Use current media stream FPS to calculate ticks to jump forward/backward

### DIFF
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -1346,8 +1346,8 @@ export class PlaybackManager {
         };
 
         self.getFps = function (player) {
-            self.currentMediaSource(player).MediaStreams.find(s => s.Type === "Video")?.RealFrameRate
-        }
+            return self.currentMediaSource(player).MediaStreams.find(s => s.Type === 'Video')?.RealFrameRate;
+        };
 
         function getSavedMaxStreamingBitrate(apiClient, mediaType) {
             if (!apiClient) {
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -1347,7 +1347,7 @@ export class PlaybackManager {
         };
 
         self.getFps = function (player) {
-            return self.currentMediaSource(player).MediaStreams.find(s => s.Type === 'Video')?.RealFrameRate;
+            return self.currentMediaSource(player).MediaStreams.find(s => s.Type === 'Video')?.ReferenceFrameRate;
         };
 
         function getSavedMaxStreamingBitrate(apiClient, mediaType) {
@@ -3891,14 +3891,14 @@ export class PlaybackManager {
 
     nextFrame(player = this._currentPlayer) {
         const fps = this.getFps(player);
-        if (fps != null) {
+        if (fps) {
             this.seekRelative(TICKS_PER_SECOND / fps, player);
         }
     }
 
     previousFrame(player = this._currentPlayer) {
         const fps = this.getFps(player);
-        if (fps != null) {
+        if (fps) {
             this.seekRelative(-TICKS_PER_SECOND / fps, player);
         }
     }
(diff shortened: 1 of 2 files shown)


flying-sheep:
OK! Ready for review now!

flying-sheep:
@thornbill what do I have to do to get some eyes on this?

src/components/playback/playbackmanager.js
```diff
@@ -1345,6 +1346,10 @@ export class PlaybackManager {
             }
         };
 
+        self.getFps = function (player) {
+            return self.currentMediaSource(player).MediaStreams.find(s => s.Type === 'Video')?.RealFrameRate;
```
thornbill:
```suggestion
            return self.currentMediaSource(player).MediaStreams.find(s => s.Type === 'Video')?.ReferenceFrameRate;
```

src/components/playback/playbackmanager.js
```diff
@@ -3884,6 +3889,20 @@ export class PlaybackManager {
         this.seekRelative(offsetTicks, player);
     }
 
+    nextFrame(player = this._currentPlayer) {
+        const fps = this.getFps(player);
+        if (fps != null) {
```
thornbill:
```suggestion
        if (fps) {
```

src/components/playback/playbackmanager.js
```diff
@@ -3884,6 +3889,20 @@ export class PlaybackManager {
...
+            this.seekRelative(TICKS_PER_SECOND / fps, player);
+        }
+    }
+
+    previousFrame(player = this._currentPlayer) {
+        const fps = this.getFps(player);
+        if (fps != null) {
```
thornbill:
```suggestion
        if (fps) {
```

thornbill (changes requested):
Left some minor suggestions.

Thanks for your patience as I've been busy with real-life stuff.

flying-sheep:
Thank you for taking a look! Sorry for being annoying and contacting you repeatedly. Hope your life stuff is turning out well!

flying-sheep:
Hia @thornbill do you think this is ready now?

flying-sheep:
@thornbill?

thornbill (approved):

flying-sheep:
awesome, thanks!

thornbill:
Thanks for contributing! Sorry reviews are so slow currently. 💜

flying-sheep:
no worries, as long as things are alive!

PR #206, opened 2019-03-24
Conversion to webpack - part4 - module resolution, appStorage refactor, minor cleanup

vitorsemeano:
This is the fourth part of webpack conversion and it will be the last that will work as it is. Next parts will be merged in a different branch so future changes into master don't get affected. 

This part focus on name module resolution, with some refactors and minor cleanup.

PR #7319, opened 2025-11-10
Avoid clipping thumbnails in episode lists

duckinator:
**Changes**

Currently, `.listItemImage` has `background-size: cover`, which scales the picture to avoid empty space _even if that hides part of the image_. This can cause unwanted cropping:

Screenshot 2025-11-09 at 21-03-04 Jellyfin

With this change, it has `background-size: contain`, which scales the picture to ensure the entire image is visible, but _may leave empty space around the image_. This is how the same page looks with the change applied by (mis)using the "Custom CSS Code" in the server settings:

Screenshot 2025-11-09 at 21-03-13 Jellyfin

### DIFF
--- a/src/components/listview/listview.scss
+++ b/src/components/listview/listview.scss
@@ -144,7 +144,7 @@
     min-width: 2.78em;
     min-height: 2.78em;
     background-repeat: no-repeat;
-    background-size: cover;
+    background-size: contain;
     flex-shrink: 0;
     background-position: center center;
     position: relative;


duckinator:
I rebased the PR.

My Jellyfin instance has had this change manually applied via "Custom CSS Code" for almost 3 months now. It fixes the cropped thumbnails and I haven't encountered any new issues.

dfederm:
This change seems overly broad as it would impact all content type that use list views, not just episode thumbnails.

The existing codebase already has a `.listItemImage-channel` class that specifically overrides to `contain` for channel images. This suggests to me that the developers intentionally wanted `cover` as default and `contain` only for specific cases.

I would suggest either following the same pattern as `.listItemImage-channel`, or even just reusing (and renaming) `.listItemImage-channel` if you believe this change should apply to episode lists. I'm not really a UI expert though to know which "looks better" in this particular scenario or whether the Jellyfin devs agree on this change at all.

duckinator:
I appreciate where you're coming from, but making speculative changes before hearing back from the actual maintainers doesn't feel conducive to getting them to review it. 

If they reply and agree with you, I'll happily look into it. 

In the meantime, I'll periodically rebase it so it doesn't get stale.

duckinator:
I decided to dig into this a bit more.

It looks like the change from `contain` to `cover` was made in 67f04532bed2eac19abc0216f951dd872997bb29, which this commit would undo. That was done in the middle of the huge refactor in #206 around 7 years ago.

Going off the commit message, I think the change was made to address the alignment of the "watched" indicator. This is with my change, which is the same as before that 7-year-old commit:

image

With this PR as is, the choice is:

1. merge this PR, which prioritizes not cropping images no matter what, at the expense of sometimes having the watched indicator misaligned.
2. close this PR, which prioritizes aligning the watched icon no matter what, at the expense of sometimes cropping the images.

However, the root cause of the problem seems to be using a 16:9 aspect ratio image, while Jellyfin is making divs with the `listItemImage-large` class 19.5vw wide by 13vw tall (so 39:26 aspect ratio).

Given the cause, I don't think there is an option that avoids both of these problems. I think it becomes a matter of which do you care about more, and which aspect ratios do you want to prioritize.

E.g., if you decide you want to prioritize 16:9 images looking the cleanest, the height of `.listItemImage-large` could be changed to be 10.96875vw tall, it would wind up with a 16:9 aspect ratio. (Calculated via: `16/9 = 19.5/x` -> `x = (9 * 19.5) / 16` -> 10.96875.)

It seems most of the CSS related to this was inherited from Emby.

PR #7321, opened 2025-11-10
Tizen: Enable direct-play of anamorphic videos

kinke:
Anamorphic videos currently require a remux, which apparently breaks the aspect ratio: https://github.com/jellyfin/jellyfin-tizen/issues/376

On my Tizen v9 TV, the web player handles the aspect ratio just fine directly, correctly stretching the 720x472 pixels of some DVD source to a 873x472 view. **Edit**: Tested with HEVC and H264 media.

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -297,6 +297,11 @@ function supportedDolbyVisionProfileAv1(videoTestElement) {
     return videoTestElement.canPlayType?.('video/mp4; codecs="dav1.10.06"').replace(/no/, '');
 }
 
+function supportsAnamorphicVideo() {
+    // Tizen applies the aspect ratio correctly
+    return browser.tizenVersion >= 6;
+}
+
 function getDirectPlayProfileForVideoContainer(container, videoAudioCodecs, videoTestElement, options) {
     let supported = false;
     let profileContainer = container;
@@ -1219,12 +1224,6 @@ export default function (options) {
 
     const h264CodecProfileConditions = [
         {
-            Condition: 'NotEquals',
-            Property: 'IsAnamorphic',
-            Value: 'true',
-            IsRequired: false
-        },
-        {
             Condition: 'EqualsAny',
             Property: 'VideoProfile',
             Value: h264Profiles,
@@ -1245,12 +1244,6 @@ export default function (options) {
     ];
 
     const hevcCodecProfileConditions = [
-        {
-            Condition: 'NotEquals',
-            Property: 'IsAnamorphic',
-            Value: 'true',
-            IsRequired: false
-        },
         {
             Condition: 'EqualsAny',
             Property: 'VideoProfile',
@@ -1282,12 +1275,6 @@ export default function (options) {
 
     const av1CodecProfileConditions = [
         {
-            Condition: 'NotEquals',
-            Property: 'IsAnamorphic',
-            Value: 'true',
-            IsRequired: false
-        },
-        {
             Condition: 'EqualsAny',
             Property: 'VideoProfile',
             Value: av1Profiles,
@@ -1306,6 +1293,29 @@ export default function (options) {
             IsRequired: false
         }
     ];
+
+    if (!supportsAnamorphicVideo()) {
+        h264CodecProfileConditions.push({
+            Condition: 'NotEquals',
+            Property: 'IsAnamorphic',
+            Value: 'true',
+            IsRequired: false
+        });
+
+        hevcCodecProfileConditions.push({
+            Condition: 'NotEquals',
+            Property: 'IsAnamorphic',
+            Value: 'true',
+            IsRequired: false
+        });
+
+        av1CodecProfileConditions.push({
+            Condition: 'NotEquals',
+            Property: 'IsAnamorphic',
+            Value: 'true',
+            IsRequired: false
+        });
+    }
 
     if (!browser.edgeUwp && !browser.tizen && !browser.web0s) {
         h264CodecProfileConditions.push({


src/scripts/browserDeviceProfile.js
```diff
@@ -297,6 +297,11 @@ function supportedDolbyVisionProfileAv1(videoTestElement) {
     return videoTestElement.canPlayType?.('video/mp4; codecs="dav1.10.06"').replace(/no/, '');
 }
 
+function canDirectPlayAnamorphicVideo() {
+    // Tizen 9 applies the aspect ratio correctly
+    return browser.tizenVersion >= 9;
```
kinke: I guess this is supported in older versions too, but I'm only able to test a Tizen 9 device.

Bambus3000: I have a QN95A, 2021 year, Tizen 6 afaict, and I'd love to be able to test this. 
My tv had no issues direct playing without remuxing on 10.10.7. Can you point me towards a truehd enabled wgt to test with, I use the docker method from Georift for installing if that matters.

Shadowghost:
Since this could be considered a bugfix for the broken remux playback, should this be targeting the 10.11 release branch?

src/scripts/browserDeviceProfile.js
kinke: Sorry, I don't have a TrueHD-enabled build, but the .wgt in https://github.com/jellyfin/jellyfin-web/issues/7231#issuecomment-3514430066 is hopefully enough to quickly test some anamorphic video.

kinke:
> Since this could be considered a bugfix for the broken remux playback, should this be targeting the 10.11 release branch?

I can re-target obviously; I just can't confirm that this has actually regressed in v10.11 (as I haven't tested anamorphic videos with 10.10 earlier). These profile-conditions seem to be ancient (added around 5 years ago, for H264/H265 at least), so it didn't look like a regression to me, and I wasn't expecting direct-play to ever have worked.

src/scripts/browserDeviceProfile.js
Bambus3000: I tried to install it, but I couldn't. I have no idea how to install other than using the method above, and I think that is somewhat hardcoded to use jeppevinkel's builds. Pointing to your file locally just failed on me.
But would that even make sense to try, didn't you set tizen 9 as minimum version?

kinke: > But would that even make sense to try, didn't you set tizen 9 as minimum version?

Yeah it would, as I've changed the condition for that build to `browser.tizen`, specifically so that direct-play can be tested on all Tizen devices: https://github.com/kinke/jellyfin-web/commit/efee821fe24a93228a08bd14f368862cfb158ca0

FWIW, this is the script I'm using on Linux (incl. resigning the .wgt so that I can install it; had to create my own Samsung dev account... - **edit**: but I had to do that for jeppevinkel's builds too):
```bash
#!/usr/bin/env bash

set -euxo pipefail

tv_ip="192.168.178.20"
wgt="~/Downloads/Jellyfin.wgt"

# enable dev mode on TV, choosing *this* machine's IP, then reboot TV
curl -f http://$tv_ip:8001/api/v2/

~/tizen-studio/tools/sdb connect $tv_ip

# certificate creation: https://developer.samsung.com/smarttv/develop/getting-started/setting-up-sdk/creating-certificates.html
# then fix up `password`s in ~/tizen-studio-data/profile/profiles.xml
~/tizen-studio/tools/ide/bin/tizen package -t wgt -s MyProfile -- $wgt

# -t: see device name in `sdb devices` output
~/tizen-studio/tools/ide/bin/tizen install -n $wgt -t GQ65S95FATXZG
```

kinke: Okay, changed to `tizenVersion >= 6.5` after successful tests: https://github.com/jellyfin/jellyfin-web/issues/7231#issuecomment-3518347251

Bambus3000: Ok, am I reading this right, my tv is tizen 6.0, and minimum version is 6.5? But somewhere else I read 5.5? github is messy for me as a mere consumer.
The reason I went with Georifts method was that I have no interest in running tizen studio or having a samsung dev account, I'm no developer... 
10.11 has been just a depressing ordeal, and I wish I never "upgraded", especially now that my 4 year old tv suddenly is "outdated".

Bambus3000: Ok, I remembered that someone mentioned a one click installer made by PatrickSt1991, I found out that it could install a .wgt from a custom location, and after installing yours I can happily confirm that all my troublesome media plays beautifully, and your two example clips play just as well, this is on Tizen 6.0 on my 2021 65QN95A. 
Thank you so much for your work!

kinke: Perfect, changed to `tizenVersion >= 6` accordingly.

Bambus3000: Hi, just wanted to give a heads up in regards to an issue with experimental pgs sub rendering using this version, the subs are out of sync if enabled. If I disable remuxing in the user settings and select "Always burn in subtitle when transcoding" they show up on time, but with poor quality AAC audio. Reverting back to the default 10.11.z branch fixes exp. pgs rendering, but breaks anamorphic shows as expected since  your patch is not yet implemented.
EDIT: After more testing, I've found that even the normal branch has issues, and it was only by coincidence that the subtitles seemed to be working, they are in fact still broken.

Bambus3000:
Can somone review and approve this pr perhaps?

krishhari2105:
Someone pls review this PR. I don't see any issue though.

src/scripts/browserDeviceProfile.js
```diff
@@ -297,6 +297,11 @@ function supportedDolbyVisionProfileAv1(videoTestElement) {
     return videoTestElement.canPlayType?.('video/mp4; codecs="dav1.10.06"').replace(/no/, '');
 }
 
+function canDirectPlayAnamorphicVideo() {
```
thornbill: nitpicking, I think this is more inline with other functions in this file

```suggestion
function supportsAnamorphicVideo() {
```

kinke: Renamed.

Bambus3000:
Sweet, after a quick test on my tizen 6 tv with an episode of Stargate SG1 DVD which is anamorphic in an mkv container, I can happily report that it no longer remuxes, it direct plays as it used to in <=10.10.7.
Might as well note that I needed to take what thornbill said literally and install the jellyfin-master-truehd.wgt, as the jellyfin-truehd.wgt I normally install didn't work to my huge dissapointment. 🤦 TMYK. 
Thanks for all your work. 😁

Issue #7324, opened 2025-11-10
Disable 'Scan' button while scanning is in progress


erikbucik:
### This issue respects the following points:

- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-web/issues?q=is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

After initiating a library rescan, the 'Scan All Libraries' button remains active.

<img width="831" height="625" alt="Image" src="https://github.com/user-attachments/assets/e30b7eed-0050-4ec5-a339-f21bee4e3db6" />

### Reproduction Steps

1. Go to the Dashboard
2. Click on the 'Scan All Libraries' button

### Expected/Actual behaviour

The button should become disabled (both in look and functionality) while a library scan is running.

### Server version

10.11.2

### Web version

10.11.2

### Build version

10.11.2

### Platform

Windows

### Browser

Brave (Chromium)

PR #7325, opened 2025-11-11
Disable scan button when scanning

viown:
**Changes**
Disable scan button in dashboard when scanning

**Issues**
Fixes https://github.com/jellyfin/jellyfin-web/issues/7324

### DIFF
--- a/src/apps/dashboard/routes/index.tsx
+++ b/src/apps/dashboard/routes/index.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useState } from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
 import Page from 'components/Page';
 import globalize from 'lib/globalize';
 import Box from '@mui/material/Box';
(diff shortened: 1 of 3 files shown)


src/apps/dashboard/routes/index.tsx
```diff
@@ -94,6 +99,7 @@ export const Component = () => {
                                 onScanLibrariesClick={onScanLibraries}
                                 onRestartClick={promptRestart}
                                 onShutdownClick={promptShutdown}
+                                isScanning={librariesTask?.State == TaskState.Running}
```
larouxn: Should we also check if a scan is still cancelling? I imagine we should keep disabled until it's idle.

```TS
librariesTask?.State == TaskState.Cancelling
```

Suppose an alternative approach could be to check if _not_ idle.

```TS
librariesTask?.State !== TaskState.Idle
```

viown: Makes sense, thanks!

thornbill (approved):

PR #7328, opened 2025-11-11
Support DOVI in mkv container on WebOS 25 and newer

ErebZx:
This change enables Direct Play of DOVI files in MKV container on WebOS 25, while still blocking it on earlier versions that do not support it, allowing remuxing on C1, CX and other devices that won't receive the latest update.

LG’s internal player is still far from perfect, but it is improving. A recent software update (version 33.22.52) fixed the stuttering issue I experienced when playing MKV files with many embedded subtitles on earlier version 33.22.15 (tested on an LG C2). Since the update, I haven’t encountered any other issues when directly playing MKV files with DOVI, so this might be a good time to merge the change.

Seeking forward and backward now works much better compared to remuxing to TS or MP4 containers. I also haven’t been able to get subtitles out of sync like I could before. This change may fix other issues like audio passthrough problems that I'm unable to test myself.

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -1500,7 +1500,6 @@ export default function (options) {
         // On WebOS 25 and newer, mp4, ts, and mkv containers are allowed.
         // On WebOS 24 and lower, only mp4 and ts containers are allowed.
         const allowedContainers = browser.web0sVersion >= 10 ? ['mp4', 'ts', 'mkv'] : ['mp4', 'ts'];
-        
         profile.CodecProfiles.push({
             Type: 'Video',
             Container: '-' + allowedContainers.join(','),
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -1496,10 +1496,14 @@ export default function (options) {
     });
 
     if (browser.web0s && supportsDolbyVision(options)) {
-        // Disallow direct playing of DOVI media in containers not ts or mp4.
+        // Adjust DOVI container rules based on WebOS version.
+        // On WebOS 25 and newer, mp4, ts, and mkv containers are allowed.
+        // On WebOS 24 and lower, only mp4 and ts containers are allowed.
+        const allowedContainers = browser.web0sVersion >= 10 ? ['mp4', 'ts', 'mkv'] : ['mp4', 'ts'];	
+        
         profile.CodecProfiles.push({
             Type: 'Video',
-            Container: '-mp4,ts',
+            Container: '-' + allowedContainers.join(','),
             Codec: 'hevc',
             Conditions: [
                 {
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -1499,7 +1499,7 @@ export default function (options) {
         // Adjust DOVI container rules based on WebOS version.
         // On WebOS 25 and newer, mp4, ts, and mkv containers are allowed.
         // On WebOS 24 and lower, only mp4 and ts containers are allowed.
-        const allowedContainers = browser.web0sVersion >= 10 ? ['mp4', 'ts', 'mkv'] : ['mp4', 'ts'];	
+        const allowedContainers = browser.web0sVersion >= 10 ? ['mp4', 'ts', 'mkv'] : ['mp4', 'ts'];
         
         profile.CodecProfiles.push({
             Type: 'Video',
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -1499,7 +1499,7 @@ export default function (options) {
         // Adjust DOVI container rules based on WebOS version.
         // On WebOS 25 and newer, mp4, ts, and mkv containers are allowed.
         // On WebOS 24 and lower, only mp4 and ts containers are allowed.
-        const allowedContainers = browser.web0sVersion >= 10 ? ['mp4', 'ts', 'mkv'] : ['mp4', 'ts'];
+        const allowedContainers = browser.web0sVersion >= 25 ? ['mp4', 'ts', 'mkv'] : ['mp4', 'ts'];
         profile.CodecProfiles.push({
             Type: 'Video',
             Container: '-' + allowedContainers.join(','),
--- a/src/scripts/browser.js
+++ b/src/scripts/browser.js
@@ -115,7 +115,11 @@ function web0sVersion(browser) {
 
         // The next is only valid for the app
 
-        if (browser.versionMajor >= 94) {
+        if (browser.versionMajor >= 120) {
+            return 25;
+        } else if (browser.versionMajor >= 108) {
+            return 24;
+        } else if (browser.versionMajor >= 94) {
             return 23;
         } else if (browser.versionMajor >= 87) {
             return 22;


tuzidaguai:
I built your PR and ran it together with Jellyfin 10.11.2. On the CX TV, this pr forces the EAC3 in P8 MKV files to transcode to AAC, but it does not work in remuxing mode.

dmitrylyzo:
AFAIK, DoVi in MKV was denied because DoVi Profile 5 was played with incorrect colors - https://github.com/jellyfin/jellyfin-web/pull/4916
Also, it didn't play DoVi, but a HLG/HDR10 fallback, iirc.

Is DoVi Profile 5 in MKV now playing with the correct colors?
Is DoVi actually being played instead of the backup stream?

tuzidaguai:
> AFAIK, DoVi in MKV was denied because DoVi Profile 5 was played with incorrect colors - #4916 Also, it didn't play DoVi, but a HLG/HDR10 fallback, iirc.
> 
> Is DoVi Profile 5 in MKV now playing with the correct colors? Is DoVi actually being played instead of the backup stream?

This PR can activate Dolby Vision for p8 MKV on webOS as long as you set Prefer fMP4-HLS Media Container: enable; otherwise, it will be HDR. P5 MP4 can work in direct play mode while also enabling Dolby Vision on tv. I currently don't have a sample for P5 MKV, so it hasn't been tested. 


> Seeking forward and backward now works much better compared to remuxing to TS or MP4 containers. I also haven’t been able to get subtitles out of sync like I could before. This change may fix other issues like audio passthrough problems that I'm unable to test myself.

Additionally, I personally think that the PR that solves the CX subtitle desynchronization issue is 'Add option to limit HLS segment length for webOS (https://github.com/jellyfin/jellyfin-web/pull/6530). As long as Use shorter HLS segments: enable, which means hls_time=1, it can fix the desynchronization on CX. This applies to both fMP4 and TS containers.When I use your PR with 'Use shorter HLS segments: disable,' I still encounter the issue of subtitles being out of sync.

tv cx
Sorry, I used online translation.

ErebZx:
> I built your PR and ran it together with Jellyfin 10.11.2. On the CX TV, this pr forces the EAC3 in P8 MKV files to transcode to AAC, but it does not work in remuxing mode.

This should not change anything on CX, it shoud work the same as before. When testing, make sure you **enable** Prefer fMP4-HLS container on 10.11.2 because it uses HLS.js so it needs to remux to mp4 container for DV to work. If you test this with 10.10.7, make sure to **disable** Prefer fMP4-HLS container because it uses native hls so it needs to remux to ts container which is more stable. 





> AFAIK, DoVi in MKV was denied because DoVi Profile 5 was played with incorrect colors - #4916 Also, it didn't play DoVi, but a HLG/HDR10 fallback, iirc.
> 
> Is DoVi Profile 5 in MKV now playing with the correct colors? Is DoVi actually being played instead of the backup stream?

This was the case until WebOS 25, now we do have support for DOVI in MKV container.
Profile 5 plays fine with correct colors, Profile 8 plays in DV mode as well, it does not fallback to HDR10.
The only regression is with Profile 7, which is still unsupported, but on WebOS 25 it does not fallback to HDR10 stream, like it did on earlier versions. This issue is unrelated to this change, it manifests on latest stable version of Jellyfin as well, it fails to play and transcodes to SDR. For that problem we ether need some other workaround, or to wait for LG to fix its internal player.

src/scripts/browserDeviceProfile.js
```diff
@@ -1463,10 +1463,13 @@ export default function (options) {
     });
 
     if (browser.web0s && supportsDolbyVision(options)) {
-        // Disallow direct playing of DOVI media in containers not ts or mp4.
+        // Adjust DOVI container rules based on WebOS version.
+        // On WebOS 25 and newer, mp4, ts, and mkv containers are allowed.
+        // On WebOS 24 and lower, only mp4 and ts containers are allowed.
+        const allowedContainers = browser.web0sVersion >= 10 ? ['mp4', 'ts', 'mkv'] : ['mp4', 'ts'];
```
dmitrylyzo: You need to update `web0sVersion`: https://github.com/jellyfin/jellyfin-web/blob/77df7d5d855d3622b01bbde01d40c955c97dea76/src/scripts/browser.js#L123-L129

crimsonspecter:
> AFAIK, DoVi in MKV was denied because DoVi Profile 5 was played with incorrect colors - #4916 Also, it didn't play DoVi, but a HLG/HDR10 fallback, iirc.
> 
> Is DoVi Profile 5 in MKV now playing with the correct colors? Is DoVi actually being played instead of the backup stream?

I recently upgraded from webOS 24 to 25 and can also confirm that webOS 25 now can playback dolby vision in mkv container as well as mp4 and ts. Tested profile 5, 8, and 10 in mkv container and the tv triggers dolby vision mode with correct colors.

andresilva:
Can confirm this worked for me on LG G4 with webOS 25. Without this patch some files would result in black screen unless HLS was used, and with HLS it would always fallback to HDR and never use DV. Only tested with P8 DV. #7526 also fixes it for me, but I think that's accidental (rather than this PR which is explicitly enabling support).

Skin80:
Can also confirm this works for me. Recently upgrade my LG C3 to webOS 25 and tested using self-built jellyfin-web with patch applied.
Incidently, I tried applying the patch prior to upgrading to webOS 25 and (as expected) it didn't change anything. Therefore, I would also like to confirm that it has no adverse affect on users running webOS 24 (and lower).

SjoerdV:
> > I built your PR and ran it together with Jellyfin 10.11.2. On the CX TV, this pr forces the EAC3 in P8 MKV files to transcode to AAC, but it does not work in remuxing mode.
> 
> This should not change anything on CX, it shoud work the same as before. When testing, make sure you **enable** Prefer fMP4-HLS container on 10.11.2 because it uses HLS.js so it needs to remux to mp4 container for DV to work. If you test this with 10.10.7, make sure to **disable** Prefer fMP4-HLS container because it uses native hls so it needs to remux to ts container which is more stable.

Could we have the option back so my C1 behaves like it did on 10.10.7? That is with the `Prefer fMP4-HLS container` option **disabled**: P5 MP4 and P8.1 MKV did just work fine in that scenario/version. In the current 10.11 versions I **have** to enable the mentioned option to not get a HDR fallback, but then the audio and subs eventually mess up when seeking/resuming play. Please also keep supporting pre WebOS 24/25 LGTVs. Thanks for all your great work!

lorincmate:
I tested the change on my C4 and it works.
I wonder about the change in the `src/scripts/browserDeviceProfile.js`
Updating the if condition to this (bellow) would deliver .mkv compatibility for webOS25+ and this PR could be finally closed.
`    if (browser.web0s && browser.web0sVersion < 25 && supportsDolbyVision(options)) {`

Skin80:
Any chance we can merge this pull request? Thanks

DrHurt:
Also working on LG C2 with WebOS 25

agneevX:
Can we get this merged please? webOS 25 is now rapidly rolling out across supported devices.

Skin80:
> Can we get this merged please? webOS 25 is now rapidly rolling out across supported devices.

Not holding my breath! Reviewer hasn't been seen since beginning of March. And no way of contacting anyone...

asardaes:
If anyone would like to test this and is using the linuxserver docker image, you can add [this mod](https://github.com/asardaes/jellyfin-mod), but it will only work if all your WebOS devices are already on v25 because I simply added mkv as allowed without any version check. You can audit the change [here](https://github.com/asardaes/jellyfin-mod/blob/main/root/etc/s6-overlay/s6-rc.d/init-mod-jellyfin-mod/up).

DrHurt:
I believe confirming profile 5 compatibility was blocking this review 

> AFAIK, DoVi in MKV was denied because DoVi Profile 5 was played with incorrect colors - #4916 Also, it didn't play DoVi, but a HLG/HDR10 fallback, iirc.
> 
> Is DoVi Profile 5 in MKV now playing with the correct colors? Is DoVi actually being played instead of the backup stream?

@dmitrylyzo I can confirm both profile 8.1 (DV with HDR10)  & profile 5 (DV with no fallback layer) work correctly on LG C2 WebOS 25. 

DV logo pops up and colours render correctly.

Hope this helps.

Thanks.

theolaa:
I've found that if I force a transcode by picking a bitrate lower than the content bitrate, then setting it to a bitrate higher than the content bitrate (not auto though), it will start direct playing or direct streaming with fallback HDR10 working. Still can't get proper DoVi working though except through the system video player (not through Jellyfin).

agneevX:
> If anyone would like to test this and is using the linuxserver docker image, you can add [this mod](https://github.com/asardaes/jellyfin-mod), but it will only work if all your WebOS devices are already on v25 because I simply added mkv as allowed without any version check. You can audit the change [here](https://github.com/asardaes/jellyfin-mod/blob/main/root/etc/s6-overlay/s6-rc.d/init-mod-jellyfin-mod/up).

This change forces media with DV profile 7.6 to transcode video, whose audio would otherwise transcode to AAC on `10.11.*` and direct play on `10.10.*`.

asardaes:
> This change forces media with DV profile 7.6 to transcode video, whose audio would otherwise transcode to AAC on `10.11.*` and direct play on `10.10.*`.

Are you sure it's transcoding? I get the HDR fallback instead of DV for profile 7.6, but the player still reports "direct streaming."

Skin80:
WebOS doesn't fully support DV profile 7 (DOVIwithEL).

In addition, the Jellyfin team have updated the device profiles in the past year or so, to be stricter about DOVI compatibility.
In the past, Jellyfin just checked if it was DOVI compatible, but now it checks for different variants (DOVIWithHDR10, DOVIWithHLG, DOVIWithSDR, DOVIWithEL, DOVIWithHDR10Plus, DOVIWithELHDR10Plus, DOVIInvalid).

So, depending on what your specific model of TV reports back to Jellyfin, it will either play it as DoVi (without the EL), fallback to HDR or it will transcode.

Off topic:
You're better off either "baking" the EL of a DoVi P7 file into a single BL or converting it to an 8.1 profile. 
There are very few devices that support proper hardware decoding of P7 encoded media files with an FEL. The Ugoos AM6B+ is the only one I can name off the top of my head. All other devices that can play P7 files normally just display the MEL (eg Nvidia Shield) or drop the EL completely (all P7 compatible TVs).

Appreciate it might change in the future, but I doubt it. P5 was created for broadcast, P8 for media files/players and P7 for physical media (UHD Blurays). TV manufacturers don't need to support P7 decoding, as that's what your Bluray player does on its behalf.

DrHurt:
> 
> So, depending on what your specific model of TV reports back to Jellyfin, it will either play it as DoVi (without the EL), fallback to HDR or it will transcode.
> 

No, the enhancement layer is the DoVi metadata!

Jellyfin strips the DoVi EL and streams the HDR10 base layer via HLS. 

Without that, some devices end up with a black screen as they can’t handle the dual layer file.

TV manufacturers don’t support P7 because it requires an additional license that wouldn’t make financial sense for them to buy.

asardaes:
Is it possible to remux to mp4 discarding the EL but keeping the RPU? Or is that only possible with transcoding?

I've tested a file with DV Profile 7.6 using the latest version of Jellyfin with and without the change from this PR. Without the change it remuxes to mp4, with the change it uses direct streaming, but in both cases it uses the HDR10 fallback, so the PR doesn't seem to break anything.

DrHurt:
> Is it possible to remux to mp4 discarding the EL but keeping the RPU? Or is that only possible with transcoding?
> 
> I've tested a file with DV Profile 7.6 using the latest version of Jellyfin with and without the change from this PR. Without the change it remuxes to mp4, with the change it uses direct streaming, but in both cases it uses the HDR10 fallback, so the PR doesn't seem to break anything.

Theoretically it’s possible to convert 7.6 to 8.1 on the fly like Kodi does but not a simple task.

Pr0mises:
Can confirm it's working great on WebOS25 C4 with `Prefer fMP4-HLS Media Container:` `disabled` and `enabled` with the base of 10.11.8
Dolby Vision (P5 and P8) as well as Dolby Atmos is correctly played as direct streaming.

Skin80:
@dmitrylyzo Any chance you can review and commit? Thanks

thornbill (approved):

PR #7335, opened 2025-11-15
Blacklist Firefox native mkv playback

gnattu:
**Changes**

Firefox 145 introduced native playback of mkv files but it is not optimized for streaming use case. For a lot of files, it would require downloading the entire file before the playback could even start, which is not desirable for video streaming.

**Issues**

Fixes https://github.com/jellyfin/jellyfin/issues/15521

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -205,6 +205,13 @@ function testCanPlayMkv(videoTestElement) {
         return true;
     }
 
+    if (browser.firefox) {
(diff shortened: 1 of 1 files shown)


scratchmex:
Wondering if there is a way to force Firefox to buffer it when needed instead of blacklisting. At the end, performance wise it's only mixing it to MP4. But just wondering. I can probably research on it and make it my first contribution. What are your thoughts?

thornbill:
We can't fix a browser's broken playback handling.

scratchmex:
I'm going to continue the discussion on the issue, if you want to check out

PR #7345, opened 2025-11-21
fix book playback in continue reading home section

dkanada:
**Changes**

Right now the web client thinks books are unplayable - this is probably a relic from before the book players were added and I question whether or not the canPlay method is even required at this point. Playback works fine everywhere else in the interface.

**Issues**

None


thornbill:
This seems like a good candidate for a backport. Could you change the target branch to be the release?

PR #7347, opened 2025-11-21
Refactor light theme to use base theme

thornbill:
**Changes**
Refactors the light theme to extend the base theme. There should only be minor differences in the appearance. The most significant being the header is lighter which was already the case in the experimental layout.

**Issues**
N/A

### DIFF
--- a/src/styles/dashboard.scss
+++ b/src/styles/dashboard.scss
@@ -76,7 +76,7 @@ progress[aria-valuenow]::before {
 }
 
 a[data-role=button] {
-    background: #292929 !important;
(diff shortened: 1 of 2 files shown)


oddstr13:
The dropdown to select theme (and other things) is white in all themes

thornbill:
> The dropdown to select theme (and other things) is white in all themes

The only theme using the new base theme is the dark (and light with this PR) themes. It was an issue in the dark theme though. Styling for native select elements is pretty poorly supported so the best I could do was use a slightly different (non-transparent) color for them.

PR #7348, opened 2025-11-23
Add easier links to video details from playing and playlist screens

elialbert:
**Changes**
I added two new links. While a video is playing, the video title was just text - I made it a link to video details. Super handy.
On a video playlist, I added a video details link in the context menu. Also super handy. 

**Issues**
[Fider page](https://features.jellyfin.org/posts/3600/web-client-links-to-video-detail)

Meta:
I've been wanting these two features for a while. I'm new to Jellyfin dev and I see there's a bit of a PR backlog here, sorry to add to it! Let me know how I can help.

Videos of changes:
Desktop, video playing title link:
https://github.com/user-attachments/assets/1ba9d224-4617-4db6-857c-ba603f63834d

Playlist context:
https://github.com/user-attachments/assets/aa08a8f4-940b-4206-a58a-e2ca58e041f1

TV, video playing focus state for title link:

https://github.com/user-attachments/assets/e27cf93e-fa3a-4cc8-9ed4-ba94b8e2a5c8


src/controllers/playback/video/index.js
```diff
@@ -262,6 +262,26 @@ export default function (view) {
 
         LibraryMenu.setTitle(title);
 
+        // Make the page title clickable to navigate to the item detail page
+        const pageTitleElement = document.querySelector('.pageTitle');
```
dmitrylyzo: Better to query from `view` rather than `document` (to narrow down the search).

src/controllers/playback/video/index.js
```diff
@@ -262,6 +262,26 @@ export default function (view) {
...
+            pageTitleElement.style.cursor = 'pointer';
+
+            // Remove any existing click listener
+            const oldListener = pageTitleElement._videoDetailClickListener;
+            if (oldListener) {
+                pageTitleElement.removeEventListener('click', oldListener);
+            }
+
+            // Add new click listener
+            const clickListener = function() {
```
dmitrylyzo: Why not move out `clickListener` and call it `showItemDetails` (or similar)? You can use `currentItem` variable. This way you don't need `_videoDetailClickListener`.

src/components/itemContextMenu.js
```diff
@@ -369,6 +369,16 @@ export async function getCommands(options) {
         });
     }
 
+    // Add "View Details" option for video items in playlists
+    if (options.openItemDetails !== false
+        && (item.Type === 'Movie' || item.Type === 'Episode' || item.Type === 'Video' || item.Type === 'MusicVideo')) {
```
dmitrylyzo: Use values from  `BaseItemKind` to check the item type.

Or maybe `itemdetails` should be enabled for any item type?

src/strings/en-us.json
```diff
@@ -1697,6 +1697,7 @@
     "VideoAudio": "Video Audio",
     "ViewAlbum": "View album",
     "ViewAlbumArtist": "View album artist",
+    "ViewItemDetails": "View details",
```
dmitrylyzo: Alphabetical order.

dmitrylyzo (commented):
Not tested, but to make it work on TV, title should have `focusable` class. Or it could be a link, then using `href` it can be opened in a new tab.
https://github.com/jellyfin/jellyfin-web/blob/c23b722426efbbdce4903c53f52037c93cec4ce7/src/scripts/libraryMenu.js#L46

elialbert:
@dmitrylyzo thanks so much for the review! I made 3/4 changes. the title wasn't in the view, it's in the header, so not sure if there's a better way you'd want to refer to it but I didn't see one. I did add focusable on the h3 - if you'd rather do a link let me know. I have tested in browser but not on TV.

src/controllers/playback/video/index.js
dmitrylyzo: Oh, my bad, it belongs to the library menu.

As an option, we could add `url` argument to `LibraryMenu.setTitle` and add/remove click handling or add/change the link url there. Or maybe add another function for title "link", but `setTitle` should reset it anyway. I'm for `setTitle(title, url = '')`, but let's see what others say.

dmitrylyzo:
> I have tested in browser but not on TV.

You can change the mode in `Settings / Display / Display mode`. Then you can use the keyboard for navigation.

elialbert:
Thanks @dmitrylyzo ! I'm new to this codebase and your reviews so helpful. I have moved the title link logic to the library as you suggested, and I've made sure it works on TV mode (added a video in PR description). I set up the link to be focusable but didn't make it a button, so it looks the same as before in non-TV mode. It turns blue rather than highlighting when you focus it in TV mode.

elialbert:
@dmitrylyzo sorry to bug you. are you waiting on anything from me? anything I could do to get this PR across the finish line?

dmitrylyzo:
1. Maybe use `emby-linkbutton` element?
I tried to use it for the title link so that it works similarly to links to a TV series, season, or artist on the item details page.
https://github.com/jellyfin/jellyfin-web/blob/46ddff668b2836de6f3a57a389361926fd878b9d/src/controllers/itemDetails/index.js#L446
In TV mode, it probably looks weird when focused (like a scaled button), but consistent.
Or maybe someone has other thoughts?

2. Currently, we don't support back navigation to video playback well. It falls back to the homepage or shows the player with the theme song (if one exists and is enabled). Ideally, this behavior should be fixed. But if we can live with it for now, then there's no problem.

elialbert:
thanks hi - 
1. I tried with emby-linkbutton and it ended up adding a box highlight with same color as text, that didn't work at all. I think the way it looks now as seen in the video in description is very clear and easy? 
2. I would love to help develop back button behavior in a separate PR if you'd like

dmitrylyzo:
> it ended up adding a box highlight with same color as text, that didn't work at all.

I think that's because you haven't removed the new style: https://github.com/jellyfin/jellyfin-web/blob/ed7e8633a3133d0602033144fb157f3aa03cff93/src/styles/videoosd.scss#L345-L362
You also need to add the `button-link` class and an inline style `color:inherit;`. At least that's what I did.

> I think the way it looks now as seen in the video in description is very clear and easy?

This is more about consistency - to make links look the same. This may change with the new UI design.

elialbert:
great, happy to try that

elialbert:
https://github.com/user-attachments/assets/f355367c-14fb-4de2-b14b-bbae530a4f6a

is this what you meant? on tv mode it looks great, I got rid of the extra styles, I guess I liked them better but I definitely don't care either way and I totally agree on consistency.
in non tv mode it's just a regular link.

dmitrylyzo:
`View details` is available on the item details page. Disable `itemdetails` command here https://github.com/jellyfin/jellyfin-web/blob/01f1e97f3b24a4a774e5754e00db11bd41cc89c7/src/controllers/itemDetails/index.js#L101-L117

elialbert:
@dmitrylyzo thank you for the continued feedback and help. Truly there is a lot to consider when trying to make what seemed like a small change. I've confirmed that the more details link in the menu appears on playlists but not on the item detail page now.

src/scripts/libraryMenu.js
```diff
@@ -771,7 +771,25 @@ function setTitle (title) {
...
+            const linkElement = document.createElement('a', 'emby-linkbutton');
+            linkElement.classList.add('button-link');
+            linkElement.style.color = 'inherit';
+            linkElement.href = url;
+            linkElement.innerText = html || '';
+
+            // Add show-focus class for TV mode to enable focus indicator
+            if (layoutManager.tv) {
+                linkElement.classList.add('show-focus');
+            }
```
dmitrylyzo: Oh, `emby-linkbutton` adds focus itself.

https://github.com/jellyfin/jellyfin-web/blob/10625a03604b568a9634d477f2070f9decaba547/src/elements/emby-button/emby-button.js#L39-L43

```suggestion
```

elialbert:
ok by me!

dmitrylyzo (approved):
Both features work.

The only problem is the "back" navigation: after "back", it redirects to the home page (if there is no theme song) or stays on the video player page and plays the theme song.

A possible solution is to remove `#/video` from the history when navigating to other pages. But I think it goes beyond the scope of this PR.

elialbert:
thanks for all your help! i don't know next steps to get merged, Im assuming you do, but if there's anything I can help with I'm around.

nyanmisaka (commented):
> While a video is playing, the video title was just text - I made it a link to video details.

In my opinion, adding a clickable link to the video title when the video player **doesn't support background playback** increases the risk of accidentally exiting the video. In other words, it acts as a larger exit button, but without any warning.

elialbert:
that's an interesting point. I do see your perspective. I would argue tho that most video streaming sites link the title text back to detail, so this is following a pattern people are used to. 
I am open to other ways of doing it if you have a suggestion?

Issue #3791, opened 2022-08-05
Save progress on reading epubs


RemcoSchrijver:
Hello,

I've encountered the issues regarding saving of your reading progress with epubs. At the moment it only saves your progress on a chapter basis. Some debugging later this is caused because all locations get the same number per chapter coming from epub.js. Inside a location it is stored which 'page' you are at in the chapter, but in saving the progress to jellyfin this is lost.

The only thing that does keep pages is the CFI entry coming from a location. [Some discussion on this is happening in the epubjs git.](https://github.com/futurepress/epub.js/issues/744#issuecomment-842729281)

So saving the CFI entry as a string to Jellyfin doesn't seem like a great idea, or books should get their own entity. But it seems like only the CFI in this library has enough data to actually save your progress on a 'page' by page basis.

PR #7353, opened 2025-11-26
Use ProgressState for precise EPUB location tracking

bluesam1:
## Summary

To make the 'Books' feature useful, this updates the BookPlayer to read and write CFI (Canonical Fragment Identifier) strings via the new `ProgressState` field for precise EPUB location resumption. Falls back to percentage-based resume for backward compatibility.

## Problem

EPUB reading position cannot be accurately synced across devices because the current implementation uses percentage-based position tracking, which breaks when text reflows on different screen sizes.

## Solution

- Add `getProgressState()` method to `BookPlayer` to expose CFI string
- Store CFI on page navigation via `relocated` event
- Read `ProgressState` from API on book load for precise resume
- Add `self.getProgressState()` to `PlaybackManager` following existing patterns
- Include `ProgressState` in playback progress reports

## Changes

- **BookPlayer:** 
  - Added `getProgressState()` method returning current CFI
  - Store CFI in `progressState` on `relocated` event
  - Read `item.UserData?.ProgressState` on load, fallback to percentage if not present
  
- **PlaybackManager:**
  - Added `self.getProgressState()` method for consistency with other player state methods
  - Include `ProgressState` in `getPlayerState()` output

## Testing

- Tested EPUB resume with CFI strings
- Verified fallback to percentage-based resume for items without `ProgressState`
- Confirmed progress reports include `ProgressState` in API calls
- Confirmed other media types still work as before

## Related

Complements backend PR: jellyfin/jellyfin#15633 (Add ProgressState field for precise EPUB location tracking)
Discussion here: jellyfin/jellyfin-web/issues/3791

image<br/>
image

### DIFF
--- a/src/apps/dashboard/features/users/api/useAuthProviders.ts
+++ b/src/apps/dashboard/features/users/api/useAuthProviders.ts
@@ -1,0 +1,21 @@
+import { Api } from '@jellyfin/sdk';
+import { useQuery } from '@tanstack/react-query';
+import { useApi } from 'hooks/useApi';
+import { getSessionApi } from '@jellyfin/sdk/lib/utils/api/session-api';
+import type { AxiosRequestConfig } from 'axios';
+
+const fetchAuthProviders = async (api: Api, options?: AxiosRequestConfig) => {
+    const response = await getSessionApi(api).getAuthProviders(options);
+
+    return response.data;
+};
+
+export const useAuthProviders = () => {
+    const { api } = useApi();
+
+    return useQuery({
+        queryKey: [ 'AuthProviders' ],
+        queryFn: ({ signal }) => fetchAuthProviders(api!, { signal }),
+        enabled: !!api
+    });
+};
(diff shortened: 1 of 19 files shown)


dkanada:
@bluesam1 this is a great improvement but the server portion might need some changes - there has been an effort to avoid more attributes that only apply to specific library types. This is why it originally used the percentage value.

For what it's worth, the percentage is *slightly* more granular than per-chapter and keeps progress per-page as long as you use the same device. I believe most of the reports in #2582 and #3791 are from changing devices or exiting and reloading without refreshing the page, which will try to load the previous state in some cases.

bluesam1:
> @bluesam1 this is a great improvement but the server portion might need some changes - there has been an effort to avoid more attributes that only apply to specific library types. This is why it originally used the percentage value.
> 
> For what it's worth, the percentage is _slightly_ more granular than per-chapter and keeps progress per-page as long as you use the same device. I believe most of the reports in #2582 and #3791 are from changing devices or exiting and reloading without refreshing the page, which will try to load the previous state in some cases.

Thanks for the feedback. My thinking was that in order for this to be an ebook reader worth using, an almost-exact location is necessary. I had found the page was still off when using the same device, so I'll do some testing to see if I can replicate the issue.

PR #7354, opened 2025-11-26
initial sections for book library view

dkanada:
**Changes**

All three of these views are working excellent - series view should also be straightforward once I get some work done on the server. Is there any benefit to adding a view in the stable router or do you plan on enabling experimental by default in the next release?

Authors will need more work because the Person API is not nearly as complete as the Artist one. Once they have feature parity I think an "Author" tab can be shown that has a "Person Role" dropdown to select other types - same for movies and other libraries.

**Issues**

None

### DIFF
--- a/src/components/router/appRouter.js
+++ b/src/components/router/appRouter.js
@@ -401,7 +401,7 @@ class AppRouter {
         }
 
         if (context !== 'folders' && !itemHelper.isLocalItem(item)) {
-            const layoutMode = localStorage.getItem('layout') ?? LayoutMode.Experimental;
+            const layoutMode = localStorage.getItem('layout') || LayoutMode.Experimental;
 
             if (layoutMode === LayoutMode.Experimental && item.CollectionType == CollectionType.Books) {
                 return `#/books?topParentId=${item.Id}`;


thornbill:
Experimental layout should be the default in the next release except for TVs 👍

dkanada:
There are a number of prerequisites before series view will be usable so this is fine to review or merge. Author view is also unfeasible until API improvements are made to the Persons controller.

src/components/router/appRouter.js
```diff
@@ -401,6 +401,12 @@ class AppRouter {
         }
 
         if (context !== 'folders' && !itemHelper.isLocalItem(item)) {
+            const layoutMode = localStorage.getItem('layout');
```
thornbill: Well this is broken on master... we should use experimental if this is not set also now 😅 

```suggestion
            const layoutMode = localStorage.getItem('layout') || LayoutMode.Experimental;
```

dkanada:
@thornbill everything mentioned has been addressed - it turns out collections are *never* scoped to a single library even in the current stable release, so rather than discuss the pros and cons of whether global collections should exist I've removed it to prioritize everything else.

src/components/router/appRouter.js
```diff
@@ -401,6 +401,12 @@ class AppRouter {
         }
 
         if (context !== 'folders' && !itemHelper.isLocalItem(item)) {
+            const layoutMode = localStorage.getItem('layout') ?? LayoutMode.Experimental;
```
thornbill:
```suggestion
            const layoutMode = localStorage.getItem('layout') || LayoutMode.Experimental;
```

layout can be an empty string

PR #7357, opened 2025-11-28
Filterdialog: Turn years into strings for correct de-duplication

Collin-Swish:
# Summary
In the renderFilters function the query and api result years are passed into the lodash union function to be de-duplicated. Previously the api result was passed in as an array of integers while the query a split comma separated string. Because the lodash union function performs strict comparisons we should ensure that all values passed in are of the same type. This change converts the list of integer years into a list of string years. 

**Changes**
* Convert api result years into strings.

**Issues**
Fixes jellyfin/jellyfin-web#7361


theguymadmax (approved):

theguymadmax:
This should target the release branch:
`git rebase -i upstream/release-10.11.z` and drop all commits that are not part of this PR.

Issue #7358, opened 2025-11-28
Can't highlight artist/album in web UI


ajsnyde:
### This issue respects the following points:

- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-web/issues?q=is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

Not sure what CSS does it, but I can't highlight artist and sometimes album name when playing music. Makes it annoying to copy/paste for searching/sharing.

### Reproduction Steps

Play a song from an album, try to highlight the album _and_ artist:

<img width="1216" height="802" alt="Image" src="https://github.com/user-attachments/assets/66300bec-0fa7-4b7c-b01c-9511a199b516" />

Or play a single song, and try to highlight the song name, album, and artist:
https://github.com/user-attachments/assets/b25273a7-5032-4c90-9e6e-02cb2b6cbf6a

### Expected/Actual behaviour

I expect the relevant text to be selectable within the WebUI.

### Server version

10.10.7

### Web version

10.10.7

### Build version

10.10.7

### Platform

Linux

### Browser

Librewolf, also chromium

PR #7359, opened 2025-11-28
Update remotecontrol.scss

ajsnyde:
Correcting to attempt to fix https://github.com/jellyfin/jellyfin-web/issues/7358

Untested, not built, going by what corrected the issue in the browser by manually changing the CSS values.

### DIFF
--- a/src/styles/librarybrowser.scss
+++ b/src/styles/librarybrowser.scss
@@ -653,6 +653,10 @@
     margin: 0.15em 0 0.2em;
 }
 
+.musicParentName a {
+    user-select: text;
+}
+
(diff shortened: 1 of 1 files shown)


uxdesignerhector:
Your fix doesn't work but adding these lines to `src\styles\librarybrowser.scss` will fix it. You can remove your changes
``` css
.musicParentName a {
    /* let users select artist names in music libraries */
    user-select: text;
}
```

jellyfin-artist-name-selection

ajsnyde:
Removed my earlier changes, added your suggestion. Haven't built/tested locally.

PR #7360, opened 2025-11-29
Fix media session isLocalPlayer detection on first player change

sildur:
**Changes**
PlaybackSubscriber was firing media session updates before rebinding the new player, so the initial payload sent `isLocalPlayer: false` (player undefined) to NativeShell. Android then assumed remote playback and showed the cast/remote volume rocker instead of the phone’s local volume, making volume control cumbersome. Use the current playbackManager player as a fallback and set `this.player` on playerchange before emitting the first update to restore correct isLocalPlayer reporting.

**Issues**
Fixes: https://github.com/jellyfin/jellyfin-android/issues/1745
Fixes: https://github.com/jellyfin/jellyfin-android/issues/1854

### DIFF
--- a/src/strings/sk.json
+++ b/src/strings/sk.json
@@ -998,7 +998,7 @@
     "TvLibraryHelp": "Pozrite sa na {0}sprievodcu pomenovania TV programov{1}.",
     "LabelLineup": "Lineup",
     "ErrorAddingListingsToSchedulesDirect": "Nastala chyba pri pridávaní lineupu do účtu vášho Direct plánovača. Direct plánovač umožňuje len obmedzený počet lineupov na účet. Možno sa budete musieť prihlásiť na web Direct plánovača a odstrániť ostatné záznamy z vášho účtu pred pokračovaním.",
-    "Features": "Features",
+    "Features": "Funkcie",
     "XmlTvPathHelp": "Cesta k XMLTV súboru. Jellyfin tento súbor načíta a pravidelne ho skontroluje, či neobsahuje aktualizácie. Ste zodpovedný za vytvorenie a aktualizovanie tohoto súboru.",
     "XmlTvNewsCategoriesHelp": "Programy s týmito kategóriami budú zobrazené ako spravodajské programy. Viacej kategórií oddelte \"|\".",
     "XmlTvMovieCategoriesHelp": "Programy s týmito kategóriami budú zobrazené ako filmy. Viacej kategórií oddeľte \"|\".",


viown:
This looks like it can go in a patch release. Could you rebase your changes on top of the `release-10.11.z` branch?

dmitrylyzo:
We shouldn't set `player` in derived classes.

The problem is probably that `PlaybackSubscriber` subscribes `bindPlayerEvents` on the `PlayerChange` event after the `on*` handlers.
https://github.com/jellyfin/jellyfin-web/blob/1b54ae6caca05b6fc60ce325664d9a90c38f63ea/src/apps/stable/features/playback/utils/playbackSubscriber.ts#L81-L83
https://github.com/jellyfin/jellyfin-web/blob/1b54ae6caca05b6fc60ce325664d9a90c38f63ea/src/apps/stable/features/playback/utils/playbackSubscriber.ts#L86
Iirc, events are processed in a sequence of subscriptions. So we should swap them. Or we should call `onPlayerChange` (don't subscribe) after calling `bindPlayerEvents`.

sildur:
> We shouldn't set `player` in derived classes.
> 
> The problem is probably that `PlaybackSubscriber` subscribes `bindPlayerEvents` on the `PlayerChange` event after the `on*` handlers.
> 
> https://github.com/jellyfin/jellyfin-web/blob/1b54ae6caca05b6fc60ce325664d9a90c38f63ea/src/apps/stable/features/playback/utils/playbackSubscriber.ts#L81-L83
> 
> https://github.com/jellyfin/jellyfin-web/blob/1b54ae6caca05b6fc60ce325664d9a90c38f63ea/src/apps/stable/features/playback/utils/playbackSubscriber.ts#L86
> 
> Iirc, events are processed in a sequence of subscriptions. So we should swap them. Or we should call `onPlayerChange` (don't subscribe) after calling `bindPlayerEvents`.

Thanks for the catch. I removed the subclass assignment and rewired PlaybackSubscriber so PlayerChange binds the player before calling handlers (as you suggested about the subscription order). That ensures this.player is set before media session updates and stops Android from seeing isLocalPlayer: false.

sildur:
> This looks like it can go in a patch release. Could you rebase your changes on top of the `release-10.11.z` branch?

Of course. The pull request against `release-10.11.z` is here: https://github.com/jellyfin/jellyfin-web/pull/7376

thornbill:
Since the other PR was merged into the release branch, our tooling will handle getting it committed to master. Thanks!

PR #7376, opened 2025-12-01
Bind PlayerChange before handlers to report correct isLocalPlayer

sildur:
**Changes**
PlayerChange was firing before the subscriber rebound its player, so the first media session update could send `isLocalPlayer: false` (player undefined) and Android treated playback as remote (cast volume UI). Rewire PlaybackSubscriber so PlayerChange binds the current player before invoking handlers, ensuring media session updates always have a bound player and report the correct local/remote state.

**Issues**
Fixes: https://github.com/jellyfin/jellyfin-android/issues/1745
Fixes: https://github.com/jellyfin/jellyfin-android/issues/1854

### DIFF
--- a/src/styles/librarybrowser.scss
+++ b/src/styles/librarybrowser.scss
@@ -501,6 +501,20 @@
     background-size: contain;
     background-repeat: no-repeat;
     background-position: center center;
+
+    [dir="rtl"] & {
+        left: 25vw;
+        right: unset;
+    }
+
+    .layout-mobile &,
+    .layout-tv & {
+        display: none;
+    }
+
+    @media all and (max-width: 68.75em) {
+        display: none;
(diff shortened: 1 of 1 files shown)


sildur:
> Nice find!
> 
> In the future please avoid adding merge commits unless there is a conflict.

Thanks! Yes, I'll do so from now on.

viown:
@sildur There seems to be a minor regression here where the cast icon does not update after disconnecting, likely due to [this event not firing](https://github.com/jellyfin/jellyfin-web/blob/2851963299fcc854aabc693a44c9a45182a1df0f/src/scripts/libraryMenu.js#L852).

I have not looked into this, but I might do so later. Although if you can take a look that'd be greatly appreciated.

PR #7380, opened 2025-12-02
Add help link to backups page

theguymadmax:
**Changes**
Add help link to the backups page

**Issues**
N/A


thornbill:
Please add the link to https://github.com/jellyfin/jellyfin.org/blob/master/scripts/data/jellyfin-web-urls.json so people don't break it

PR #7384, opened 2025-12-03
Generalize multiselect across different cards/list items

makotech222:
**Changes**
Add a new parameter to ListItem and Card, 'isMultiselectable', which adds the '.multiselectable' class to the item. The multiselect component now matches on this class, so theoretically it can be expanded to other items, as long as they have 'data-id' prop.

Also fixes issue with some cards being selectable that shouldn't be, like tv series posters, next up, and cast and crew.

I believe I have passed this property in to all constructors where it is appropriate, but its possible I missed some previously used cases, but I don't have some things like live-tv to test.

**Issues**
Addresses request https://features.jellyfin.org/posts/187/multi-select-in-lists . Music, TvShows, etc are now multiselect capable and can be mass-added to playlists and such.

### DIFF
--- a/src/components/itemContextMenu.js
+++ b/src/components/itemContextMenu.js
@@ -133,7 +133,7 @@ export async function getCommands(options) {
     if (!browser.tv) {
         // Multiselect is currrently only ran on long clicks of card components
         // This disables Select on any context menu not originating from a card i.e songs
-        if (options.positionTo && (dom.parentWithClass(options.positionTo, 'card') !== null)) {
+        if (options.positionTo) {
             commands.push({
                 name:  globalize.translate('Select'),
                 id: 'multiSelect',
--- a/src/components/itemContextMenu.js
+++ b/src/components/itemContextMenu.js
@@ -132,8 +132,8 @@ export async function getCommands(options) {
 
     if (!browser.tv) {
         // Multiselect is currrently only ran on long clicks of card components
-        // This disables Select on any context menu not originating from a card i.e songs
-        if (options.positionTo) {
+        // This disables Select on any context menu not originating from a multiselectable.
+        if (options.positionTo && (dom.parentWithClass(options.positionTo, 'multiselectable') !== null)) {
             commands.push({
                 name:  globalize.translate('Select'),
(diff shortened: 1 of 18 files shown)


dfederm:
I believe `itemContextMenu.js` needs an update around line 535 as well.

Current:
```js
            case 'multiSelect':
                import('./multiSelect/multiSelect').then(({ startMultiSelect }) => {
                    const card = dom.parentWithClass(options.positionTo, 'card');
                    startMultiSelect(card);
                });
                break;
```

I think "card" needs to be changed to "multiselectable" based on your other changes.

src/components/itemContextMenu.js
```diff
@@ -133,7 +133,7 @@ export async function getCommands(options) {
     if (!browser.tv) {
         // Multiselect is currrently only ran on long clicks of card components
         // This disables Select on any context menu not originating from a card i.e songs
-        if (options.positionTo && (dom.parentWithClass(options.positionTo, 'card') !== null)) {
+        if (options.positionTo) {
```
dfederm: Consider restoring this but changing it to:

```js
if (options.positionTo && (dom.parentWithClass(options.positionTo, 'multiselectable') !== null))
```

src/components/itemContextMenu.js
```diff
@@ -133,7 +133,7 @@ export async function getCommands(options) {
     if (!browser.tv) {
         // Multiselect is currrently only ran on long clicks of card components
         // This disables Select on any context menu not originating from a card i.e songs
```
dfederm: This comment no longer seems relevant, or if it is, replace "card"

makotech222:
thank you for the review. ill take a look at those change requests and have them fixed up shortly.

src/components/itemContextMenu.js
makotech222: Okay re-added

src/components/itemContextMenu.js
makotech222: cleaned it up

makotech222:
Okay all the issues have been fixed

PR #7398, opened 2025-12-07
Fix keyboard controls working only on latin keyboard layouts

### LINKED ISSUE #7397: J/K/L playback control keys don't work on non-latin keyboard layouts

### Describe the bug

J/K/L keys allow you to go back, pause and go forward, but it doesn't work on cyrillic keyboard layouts.

To make sure this is a valid report I've checked and this works on youtube

### Reproduction Steps

1. Switch to russian keyboard layout
2. Click J/K/L | Й/К/Л (same location on keyboard)
3. Observe that it doesn't work

### Expected/Actual behaviour

It should

### Logs

_No response_

### Server version

10.11.4

### Web version

10.11.4

### Build version

10.11.4?

### Platform

Linux

### Browser

Brave

### Additional information

Image

`code` should probably be used instead of `key`

https://github.com/jellyfin/jellyfin-web/blob/068f2e691bea9ee6e66dd2e613763cceb7272d85/src/scripts/keyboardNavigation.js#L125-L128

MBR-0001:
**Changes**
Made keyboardNavigation return event.code instead of key, which resolves to same string regardless of layout

image

Applied changes wherever getKeyName is used, tested with media playback and it seems to be working just fine.

Not sure about TVs and stuff like that, maybe someone more knowledgeable can say if this breaks anything. I hope it doesn't :eyes: 

Tested:
[Screencast_20260407_190512.webm](https://github.com/user-attachments/assets/4e4a5038-78cb-4336-9acf-aefa6757b8f5)

**Issues**

Fixes #7397

### DIFF
--- a/src/scripts/keyboardNavigation.js
+++ b/src/scripts/keyboardNavigation.js
@@ -124,7 +124,7 @@ if (!hasFieldKey) {
  */
 export function getKeyName(event) {
     const key = KeyNames[event.keyCode] || event.key;
-    return KeyAliases[key] || key;
+    return KeyAliases[key] || event.code;
 }
 
 /**
--- a/src/controllers/playback/video/index.js
+++ b/src/controllers/playback/video/index.js
@@ -1280,8 +1280,7 @@ export default function (view) {
                     e.stopPropagation();
                 }
                 break;
-            case 'k':
-            case 'K':
+            case 'KeyK':
                 if (!e.shiftKey) {
                     e.preventDefault();
                     playbackManager.playPause(currentPlayer);
@@ -1302,8 +1301,7 @@ export default function (view) {
                     playbackManager.volumeDown(currentPlayer);
                 }
                 break;
-            case 'l':
-            case 'L':
+            case 'KeyL':
             case 'ArrowRight':
             case 'Right':
                 if (!e.shiftKey) {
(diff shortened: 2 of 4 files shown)


thornbill (approved):
I don't think we have a good way of testing ever input possibility here, so let's get it in for the 12.0 RCs and see what shakes out. 😅

src/scripts/keyboardNavigation.js
```diff
@@ -124,7 +124,7 @@ if (!hasFieldKey) {
  */
 export function getKeyName(event) {
     const key = KeyNames[event.keyCode] || event.key;
-    return KeyAliases[key] || key;
+    return KeyAliases[key] || event.code;
```
dmitrylyzo, line 127: This broke old browsers: `code` is `undefined`.

I think we should move to `code` this way:
```js
const code = KeyCodes[event.keyCode] || event.code;
return KeyAliases[code] || code;
```
and change `hasFieldKey` block.

MBR-0001: thanks! I'll try to make a PR for this soon

dmitrylyzo: No need. I'm almost ready to submit the changes.

PR #7400, opened 2025-12-08
Fix trickplay images not displaying for alternate movie versions

theguymadmax:
**Changes**
Use `mediaSourceId` instead of `item.Id`

**Issues**
https://github.com/jellyfin/jellyfin/issues/15297#issuecomment-3560137897


thornbill (changes requested):
We're passing the `MediaSourceId` as a query param per the API spec... if that isn't working, this sounds like a server bug.

theguymadmax:
Fixing on server: https://github.com/jellyfin/jellyfin/pull/15757

PR #7409, opened 2025-12-09
Fix subtitles need to be uploaded twice

### LINKED ISSUE #7194: Subtitles need to be uploaded twice

### Describe the bug

When uploading an .srt file nothing happens on the first try. I need to do it twice for the change to be visible. Even refreshing the webpage does not solve it so I think it is not an ui problem.

### Reproduction Steps

1 Use your mobile device with Google Chrome
2 Go to a movie
3 Three dot menu
4 Edit subtitle
5 German
6 Click +
7 Upload .srt
8 Save
9 Wait
10 Nothing happen so repeat 3 to 8 and see that it is only being saved if you do it twice

### Expected/Actual behaviour

Should be saved on first try

### Logs

_No response_

### Server version

10.10.7

### Web version

10.10.7

### Build version

10.10.7

### Platform

Android 16

### Browser

Chrome

### Additional information

Jellyfin is on unraid. App data is on SSD and media on HDD.

Even after successfully uploading the first subtitle and trying to add English ones immediately it also needs to be done twice.

shempignon:
**Changes**
Fix the need to upload subtitles twice.
The issue was that since the `preventDefault` method was called after an asynchronous task, it was never reach. Hence the page reload which is the default behavior of an `onSubmit` form event. 

**Issues**
Fixes #7194

### DIFF
--- a/src/components/subtitleuploader/subtitleuploader.js
+++ b/src/components/subtitleuploader/subtitleuploader.js
@@ -80,11 +80,12 @@ function setFiles(page, files) {
 }
 
 async function onSubmit(e) {
+    e.preventDefault();
+
     const file = currentFile;
 
     if (!isValidSubtitleFile(file)) {
         toast(globalize.translate('MessageSubtitleFileTypeAllowed'));
-        e.preventDefault();
         return;
     }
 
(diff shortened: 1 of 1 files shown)


dmitrylyzo (commented):
LGTM. But I think this should be rebased onto `release-10.11.z` to fix this behavior in the next release.

PR #7411, opened 2025-12-09
Fix browser detection overwrites

thornbill:
**Changes**
* Moves browser detection logic to a function so it is testable
* Fixes issues where the `browser.tv` flag is overwritten for PS4 and Xbox
* Fixes the removal of the `browser.operaTv` flag for Titan OS being overwritten
* Removes duplicated `toLowerCase` calls

**Issues**
N/A

### DIFF
--- a/src/scripts/browser.js
+++ b/src/scripts/browser.js
@@ -242,100 +242,98 @@ const uaMatch = function (ua) {
     };
 };
 
-/* eslint-disable @stylistic/indent */
 export const detectBrowser = (userAgent = navigator.userAgent) => {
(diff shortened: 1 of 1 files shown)


src/scripts/browser.js
```diff
@@ -248,9 +242,11 @@ const uaMatch = function (ua) {
     };
 };
 
-const userAgent = navigator.userAgent;
+/* eslint-disable @stylistic/indent */
```
thornbill: I'll fix this in a subsequent commit so this is easier to review

nielsvanvelzen (approved):

PR #7414, opened 2025-12-10
fix(subtitles): Correction for subtitle shifts when transcoding

### LINKED ISSUE #7413: Subtitles shift when playing transcoded media

### Describe the bug

I noticed a few months ago, that when playing transcoded media in the web player the subtitle track shifts in time a seemingly random amount upon switching video tracks or seeking. I could correct the shift by applying a subtitle offset manually, but upon seeking again, it would require me to adjust it. 

I've tested this with clients on Windows and MacOS, in browsers like Chrome, Zen, Edge and Safari and through Jellyfin servers installed via Docker on Linux and a fresh Windows install. Happened with all kinds of different media.

### Reproduction Steps

1. Open any Jellyfin web client
2. Play a media of your choice
3. Turn on subtitles
4. Set a lower bit rate, where the server must transcode
5. Seek to a random time
6. Observer the subtitle not aligning with the video track

### Expected/Actual behaviour

The subtitle track should stay consistently aligned with the video and audio track.

### Logs

_No response_

### Server version

10.10.2

### Web version

10.10.2

### Build version

10.10.2

### Platform

Windows, MacOS

### Browser

Chrome, Zen (Firefox), Edge, Safari

### Additional information

_No response_

6Leoo6:
Fixed #7413 submitted by me

**Cause**
The problem was that the native video player did not follow the changes made to the initial PTS while playing an HLS stream. This caused a desync in the original subtitle timings and the HLS library's current time and the subtitles were display a few seconds off. This usually only happened while playing transcoded content after a seek was made.

**Changes**
Hooked up a listener to the **INIT_PTS_FOUND** event of the HLS library, that fires, when the initial PTS value changes. I keep track of the shift overtime and apply a correctional offset to the start and end times of the current text track's cues each time. Now when loading a new text track, it auto-adjusts the start times for the cues. This implementation does not interfere with the existing manual subtitle shifting mechanism. 

**Note**
I attempted a fix for my current instance, while testing in multiple browsers, tho I'm not sure if I covered all possible scenarios for on-client subtitle displaying. Further testing might be needed. 

**Issues**
#7413

### DIFF
--- a/src/plugins/htmlVideoPlayer/plugin.js
+++ b/src/plugins/htmlVideoPlayer/plugin.js
@@ -492,10 +492,10 @@ export class HtmlVideoPlayer {
                             this.hideTextTrackWithActiveCues(textTrack);
                         }
 
-                        Array.from(cues).forEach(cue => {
+                        for (const cue of Array.from(cues)) {
                             cue.startTime -= offsetToApply;
                             cue.endTime -= offsetToApply;
-                        });
+                        }
 
                         if (shouldClearActiveCues) {
                             this.forceClearTextTrackActiveCues(textTrack);
--- a/src/plugins/htmlVideoPlayer/plugin.js
+++ b/src/plugins/htmlVideoPlayer/plugin.js
@@ -293,6 +293,14 @@ export class HtmlVideoPlayer {
      * @type {number | null | undefined}
      */
     #currentTime;
+    /**
+     * @type {number | null | undefined}
+     */
+    #hlsStreamTimeShift;
+    /**
+     * @type number
+     */
+    #currentSubtitleCorrectionOffset;
 
     /**
      * @private (used in other files)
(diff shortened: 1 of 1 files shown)


PeachesMLG:
Does this actually work?

The issue is Jellyfin tells ffmpeg to start transcoding at a specific time (say 10:00) however it provides it the -noaccurate_seek flag, so ffmpeg will actually start the transcoding at the closest keyframe to 10:00, which may actually be 09:48

This then starts streaming to the web client, so the hls player will think its playing 10:00, however the transcoded media actually starts at 09:48.


Im unsure how this code would solve that problem, and if we did want to solve the problem it would be better to fix it on the server side, not applying a hacky patch onto the web client no?

6Leoo6:
After trying with many different media sources I think it does work.

I admit that this is kind of a _hacky patch_, but while we cannot make changes to the server code itself due to the FFmpeg mess, we can mitigate the effects a bit. I'm sure this fix could be improved further, but I'm not yet sure if it's worth going down that path. Feedback would be appreciated. 

> the hls player will think its playing 10:00

The player thinks it has encountered a discontinuity in the stream and continues as normal with the fragments. It will know exactly what the offset was and provides an easy way of bearing with the problem through an event after the discontinuity and the offset provided (as initPTS).

I think FFmpeg might be doing its work as intended (the -noaccurate_seek **_DOES_** suggest a **_non accurate seek_**) and the culprit could be the HLS.js library itself. I will look further into their implementation and what could be done.

PR #7415, opened 2025-12-10
Fix the return value for 'behavior' property

dmitrylyzo:
**Changes**
Fix the return value for 'behavior' property. Change `null` to `'auto'`.

**Issues**
> TypeError: Failed to execute 'scrollTo' on 'Element': Failed to read the 'behavior' property from 'ScrollOptions': The provided value 'null' is not a valid enum value of type ScrollBehavior.

### DIFF
--- a/src/components/scrollManager.js
+++ b/src/components/scrollManager.js
@@ -41,7 +41,7 @@ try {
     const opts = Object.defineProperty({}, 'behavior', {
         get: function () {
             supportsScrollToOptions = true;
-            return null;
+            return 'auto';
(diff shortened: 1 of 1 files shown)


6Leoo6:
LGTM. Confirmed fix in both Chromium and Firefox browsers. Fix also aligns with official documentation on [allowed properties of behavior](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTo#behavior).

thornbill (approved):

Issue #7352, opened 2025-11-25
Dashboard Series/Episodes and Albums/Songs counts merge at a specific width


rpatterson:
### This issue respects the following points:

- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-web/issues?q=is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

In the admin "Dashboard", directly under the "Scan All Libraries" button, the two media counts in each of the boxes for "Series Episodes" and "Albums Songs" lose any spacing between the two numbers and visually merge together appearing as one number at a certain width when those numbers are large enough. With larger widths they are separated by enough space to distinguish the numbers and with narrower widths the grid is responsive, rearranges the layout, and again the counts are separated by enough space. So this happens only when:

- the counts are large, I see it when the total number of digits between the two counts is 8-9 digits
- only a very specifc, small range of widths just before the grid system responds and rearranges

### Reproduction Steps

Testing in Chrome, as I increase and decrease the width using the browser's zoom, first 9 digit combined counts then 8 digit combined counts merge, then the grid responds, rearranges the layout, and restores the spacing. This happens when the counts grid changes from 3 to 2 columns as well as when the outer grid system places the "Activity" and "Paths" boxes to the right then below. 

### Expected/Actual behaviour

The responsive UI should maintain visual separation even when the counts are large, even if it has to elide (`...`) those counts. Personally, I'd prefer the counts be changed to use magnitude suffixes when the numbers are large enough: `1,001 10,001` -> `1k 10k` .

### Server version

10.11.3

### Web version

10.11.3

### Build version

10.11.3

### Platform

Linux Docker: `image: "ghcr.io/jellyfin/jellyfin"`

### Browser

Chrome v144.0.7534.0

### Additional information

Very minor, I would prioritize this below pretty much everything else, but I thought I'd report while noticing it and maybe it's a good beginner UI contributor bug.

uxdesignerhector:
I confirm it is happening. I will give a look to this.
![Image](https://github.com/user-attachments/assets/e9e7eb75-8aa3-43d6-b57e-2b4dcba8ab8e)

PR #7419, opened 2025-12-10
Add a gap to stack elements to prevent numbers from other adjacent stack to collapse and appear as one

uxdesignerhector:
Adds a gap to stack elements in `admin dashboard` to prevent numbers from other adjacent stack to collapse and appear as one at some positions. It is better explained with images:

Issue:
jellyfin-no-margin-between-numbers

Fix:
jellyfin-gap-between-numbers

**Changes**
```
src\apps\dashboard\features\metrics\components\MetricCard.tsx
```

**Issues**
Fixes https://github.com/jellyfin/jellyfin-web/issues/7352

### DIFF
--- a/src/apps/dashboard/features/metrics/components/MetricCard.tsx
+++ b/src/apps/dashboard/features/metrics/components/MetricCard.tsx
@@ -35,12 +35,12 @@ const MetricCard: FC<MetricCardProps> = ({
         >
             <Stack
                 direction='row'
+                spacing={2}
                 sx={{
                     width: '100%',
                     padding: 2,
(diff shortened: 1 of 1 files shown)


6Leoo6:
LGTM. A simple and obvious fix. Confirmed in both Chromium and Firefox.

soultaco83:
Seems to work fine on my end. Keeping proper spacing between the objects without colliding the variables.

viown:
The `gap` property isn't supported on Jellyfin Media Player on certain operating systems. Maybe use `spacing={1}` instead?

PR #7425, opened 2025-12-13
Fix title truncation without ellipsis in Identify search results

### LINKED ISSUE #7368: [UI] Truncated titles in Identify search do not display "..." indicating truncation

### Describe the bug

I reported this on Jellyfin-Media-Player and after 9 months the issue got closed asking me to report it here:

https://github.com/jellyfin/jellyfin-media-player/issues/852

This issue occurs in both the MacOS jellyfin-media-player client and the web client.

Search for a title in the Identify pane. If the title is more than ~15 characters long, the result will be truncated in the display. But the title will not display "..." indicating truncation of the title. Lbrary listings do indicate truncation in this manner, so the behaviour is inconsistent.

### Reproduction Steps

Click identify on a movie, search for "all the world's a stooge"
Observe the title in the result is truncated without any indication of truncation

### Expected/Actual behaviour

Ideally you would rethink the UI so that titles are not truncated and are clearly identified. Second best would be to show the full title as a tooltip during mouseover. Regardless, the truncation itself should be noted with "..." in all possible views.

See screenshot in the original report

### Logs

_No response_

### Server version

10.10.2

### Web version

10.10.2

### Build version

10.10.2

### Platform

MacOS

### Browser

N/A

### Additional information

_No response_

Shardy678:
**Changes**

Add a targeted CSS override to restore the text-overflow ellipsis indicator for truncated titles specifically in the Item Identifier search results.

**Issues**

Fixes #7368

### DIFF
--- a/src/components/cardbuilder/card.scss
+++ b/src/components/cardbuilder/card.scss
@@ -317,6 +317,12 @@ button::-moz-focus-inner {
     text-overflow: initial;
 }
 
+.identificationSearchResultList .cardText {
+    white-space: nowrap !important;
+    overflow: hidden !important;
+    text-overflow: ellipsis !important;
+}
+
 .cardText-secondary {
     font-size: 86%;
 }


MenotiFilho:
I tested this using the Cloudflare preview build.

I opened the Identify dialog for a movie and searched for
`all the world's a stooge`, as described in #7368. With a narrow
viewport, long result titles are now truncated with an ellipsis
instead of being cut off without any visual indication.

The change is scoped to Identify search results, and I did not
notice layout regressions in that dialog.

dkanada (approved):

Issue #6301, opened 2024-11-05
When playing music, Jellyfin doesn't show artist not song name on tab


alangrafu:
**Describe The Bug**

Most music players provide information about artist and song being played in the tab, while Jellyfin doesn't.

**Steps To Reproduce**

1. Play some music on Jellyfin
2. Check the tab of the jellyfin web app
3. The information is "Jellyfin" and "Playing".

**Expected Behavior**

I would expect the name of the artist and song.

**Logs**

**Screenshots**

![Image](https://github.com/user-attachments/assets/94063385-29a7-4382-a128-38b415191482)

**System (please complete the following information):**
 - Platform:  Debian
 - Browser: Firefox
 - Jellyfin Version: 10.10.0

**Additional Context**

alangrafu:
BTW I know this is more a feature request than a bug, but it seems to be a **relatively** small fix, instead of a whole feature.

Dan-613:
I can look at this, but can someone guide me as to where the code lives that handles playback of music?

viown:
@Dan-613 Sure! A lot of the audio playback is handled here https://github.com/jellyfin/jellyfin-web/blob/master/src/plugins/htmlAudioPlayer/plugin.js. There's also the [now playing bar](https://github.com/jellyfin/jellyfin-web/tree/master/src/components/nowPlayingBar), because it may be worth considering whether we should add it for casted media as well.

PR #7429, opened 2025-12-14
[UI] Update document title (Tab) with artist `name - song` when playing songs

uxdesignerhector:
Update document title (Tab) with artist `name - song` when playing songs and reverts to default document title when music is stopped.

This only affects music library. Other types of library are not affected.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-web/issues/6301

jellyfin-update-tab-with-artist-song


tylers-username:
@uxdesignerhector, this is a nice enhancement. It seems to work as reported. 

⚠️ The only UX issue that stands out to me is that if the user changes tabs, the title is changed to that of the tab until the next audio in the list starts. 

💡 We may want a check such as `block title changes if audio is active.`

Issue #6221, opened 2024-10-18
Activity Page does not support Filtering and Pagination together correctly


PeachesMLG:
**Describe The Bug**
The Activity Page does not support Filtering and Pagination together due to doing pagination on server side, and filtering on the client side.
This means If you set a filter for a user, it only filters the results on the current page you are on.
e.g. Page1 may return 0 results, Page 2 may return 2 results, page 3 5 ect.

**Steps To Reproduce**
This is demenstrated best on an account that doesnt use jellyfin often
1. Go to the activity page
2. Create a filter (e.g. filter by a user)
3. Navigate through the pages and see the problem

**Screenshots**
Page 1
![image](https://github.com/user-attachments/assets/efe18edf-b035-4bc3-8e45-35bdb25ac7c0)
Page 2
![Untitled design](https://github.com/user-attachments/assets/2fe9de4f-f4c0-4010-b885-4173323a9559)

**System (please complete the following information):**
 - Platform: Browser
 - Browser: Firefox
 - Jellyfin Version: 10.9.11

A7Lavinraj:
That means filtering should be done first before pagination, right?

thornbill:
It's definitely not ideal, but this is currently working as intended. Pagination happens on the backend because the query can be slow (I have 126,000+ entries on my main server) and filtering happens on the frontend because the api does not support filtering. We might be able to add some logic to fetch additional rows if filtering results in less than the page size but that doesn't appear to be something that is supported "out of the box" with the table library.

PeachesMLG:
> It's definitely not ideal, but this is currently working as intended. Pagination happens on the backend because the query can be slow (I have 126,000+ entries on my main server) and filtering happens on the frontend because the api does not support filtering. We might be able to add some logic to fetch additional rows if filtering results in less than the page size but that doesn't appear to be something that is supported "out of the box" with the table library.

When you say the API doesnt support it, do you mean the Jellyfin server API? or the Material React Table Library?

I think the best option here would be to add filtering onto the backend Jellyfin api no?

thornbill:
The user activity API in the Jellyfin server does not support filtering.

A7Lavinraj:
So what we have to build anyways, like give some hints to do it

PeachesMLG:
> The user activity API in the Jellyfin server does not support filtering.

So from what I can see it has Date and UserId filtering
[...]

PR #7431, opened 2025-12-17
Implemented sorting and filtering for Activity page

bjorntp:
**Changes**
- Added support for backend filters and sorting in Activity page.
- Changed filters for Severity and Name to single-select to comply with backend. 
- Added range support filter for dates
- Disabled the regular search bar as there is no search support in the backend. In my opinion it is redundant as well now that the filters work.

This relies on code from the unstable backend build. Keep that in mind if testing.

**Issues**
 #6221


thornbill (changes requested):
1. You removed search param support so links from elsewhere in the dashboard are not behaving as expected.
2. The filter options should not be shown by default.
3. You shouldn't be making unrelated whitespace changes.

bjorntp:
FYI, I have made the changes. I just want to add the comment since I suspect that you get a lot of comments on GitHub so you might not see a request, and since you have a large workload here I do not want it to slip through the cracks.

thornbill (approved):

PR #7433, opened 2025-12-17
Implement continuous checking of the behavior of track control buttons in a playlist

### LINKED ISSUE #7432: [UX] Track control buttons won't disappear if the user reaches last track watching the playlist

### Describe the bug

I've found this issue: if the user clicks at the last track in the TV shows / music, etc., track control buttons (`Previous track` and `Next track`) will disappear. But, this doesn't happen when user reaches last track watching the playlist.

Example: TV show of 5 episodes (1 - 2 - 3 - 4 - 5). User clicks 5 - track buttons disappear. But, if the user starts watching from somewhere except the last track and reaches this last track (3 - 4 - 5), these buttons won't disappear.

I've found a solution, but I don't actually know if it could ruin the performance to be honest.

### Reproduction Steps

User clicks on the last track of the category: `Previous track` and `Next track` buttons disappear.
User reaches the last track via watching a playlist: `Previous track` and `Next track` buttons doesn't disappear (they actually should). 

### Expected/Actual behaviour

It was expected that the behavior of these buttons would be the same in these two cases.

### Logs

_No response_

### Server version

10.11.5

### Web version

10.11.5

### Build version

10.11.5

### Platform

Everywhere

### Browser

Everywhere

### Additional information

_No response_

JustChickNugget:
Now track control buttons will disappear when the user reaches last track watching the playlist.

**Changes**
The behavior of track control buttons is checked each time a track is switched (pressing on the `Previous track`, `Next track` buttons via interface or using keyboard shortcuts).

**Issues**
Fixes #7432

### DIFF
--- a/src/apps/legacy/controllers/playback/video/index.js
+++ b/src/apps/legacy/controllers/playback/video/index.js
@@ -1343,18 +1343,12 @@ export default function (view) {
                 if (e.shiftKey) {
                     e.preventDefault();
                     playbackManager.previousTrack(currentPlayer);
-
-                    const currentPlaylist = await playbackManager.getPlaylist();
-                    updateTrackButtonsState(currentPlaylist);
                 }
                 break;
             case 'KeyN':
                 if (e.shiftKey) {
                     e.preventDefault();
                     playbackManager.nextTrack(currentPlayer);
-
-                    const currentPlaylist = await playbackManager.getPlaylist();
-                    updateTrackButtonsState(currentPlaylist);
                 }
                 break;
             case 'NavigationLeft':
@@ -1963,9 +1957,6 @@ export default function (view) {
 
     view.querySelector('.btnPreviousTrack').addEventListener('click', async function () {
         playbackManager.previousTrack(currentPlayer);
-
-        const currentPlaylist = await playbackManager.getPlaylist();
-        updateTrackButtonsState(currentPlaylist);
     });
     view.querySelector('.btnPreviousChapter').addEventListener('click', function () {
         playbackManager.previousChapter(currentPlayer);
@@ -1978,9 +1969,6 @@ export default function (view) {
     });
     view.querySelector('.btnNextTrack').addEventListener('click', async function () {
         playbackManager.nextTrack(currentPlayer);
-
-        const currentPlaylist = await playbackManager.getPlaylist();
-        updateTrackButtonsState(currentPlaylist);
     });
     btnRewind.addEventListener('click', function () {
         playbackManager.rewind(currentPlayer);
--- a/src/apps/legacy/controllers/playback/video/index.js
+++ b/src/apps/legacy/controllers/playback/video/index.js
@@ -911,20 +911,8 @@ export default function (view) {
     }
 
     async function updatePlaylist() {
-        try {
-            const playlist = await playbackManager.getPlaylist();
-
-            if (playlist && playlist.length > 1) {
-                const btnPreviousTrack = view.querySelector('.btnPreviousTrack');
-                const btnNextTrack = view.querySelector('.btnNextTrack');
-                btnPreviousTrack.classList.remove('hide');
-                btnNextTrack.classList.remove('hide');
-                btnPreviousTrack.disabled = false;
-                btnNextTrack.disabled = false;
-            }
-        } catch (err) {
-            console.error('[VideoPlayer] failed to get playlist', err);
-        }
+        const currentPlaylist = await playbackManager.getPlaylist();
+        updateTrackButtonsState(currentPlaylist);
     }
 
     function updateTimeText(elem, ticks, divider) {
@@ -1212,7 +1200,7 @@ export default function (view) {
         }
     }
 
-    function onKeyDown(e) {
+    async function onKeyDown(e) {
         clickedElement = e.target;
 
         const isKeyModified = e.ctrlKey || e.altKey || e.metaKey;
@@ -1355,12 +1343,18 @@ export default function (view) {
                 if (e.shiftKey) {
                     e.preventDefault();
                     playbackManager.previousTrack(currentPlayer);
+
+                    const currentPlaylist = await playbackManager.getPlaylist();
+                    updateTrackButtonsState(currentPlaylist);
                 }
                 break;
             case 'KeyN':
                 if (e.shiftKey) {
                     e.preventDefault();
                     playbackManager.nextTrack(currentPlayer);
+
+                    const currentPlaylist = await playbackManager.getPlaylist();
+                    updateTrackButtonsState(currentPlaylist);
                 }
                 break;
             case 'NavigationLeft':
@@ -1609,6 +1603,47 @@ export default function (view) {
         const playbackRateSpeed = sessionStorage.getItem('playbackRateSpeed');
         if (playbackRateSpeed !== null) {
             player.setPlaybackRate(playbackRateSpeed);
+        }
+    }
+
+    function updateTrackButtonsState(currentPlaylist) {
+        try {
+            const currentPlaylistItemId = playbackManager.getCurrentPlaylistIndex() + 1;
+
+            const btnPreviousTrack = view.querySelector('.btnPreviousTrack');
+            const btnNextTrack = view.querySelector('.btnNextTrack');
+
+            if (currentPlaylist && currentPlaylistItemId !== currentPlaylist.length) {
+                if (btnPreviousTrack) {
+                    btnPreviousTrack.classList.remove('hide');
+                    btnPreviousTrack.disabled = false;
+                } else {
+                    console.error('[VideoPlayer] failed to get "btnPreviousTrack"', btnPreviousTrack);
+                }
+
+                if (btnNextTrack) {
+                    btnNextTrack.classList.remove('hide');
+                    btnNextTrack.disabled = false;
+                } else {
+                    console.error('[VideoPlayer] failed to get "btnNextTrack"', btnNextTrack);
+                }
+            } else {
+                if (btnPreviousTrack) {
+                    btnPreviousTrack.classList.add('hide');
+                    btnPreviousTrack.disabled = true;
+                } else {
+                    console.error('[VideoPlayer] failed to get "btnPreviousTrack"', btnPreviousTrack);
+                }
+
+                if (btnNextTrack) {
+                    btnNextTrack.classList.add('hide');
+                    btnNextTrack.disabled = true;
+                } else {
+                    console.error('[VideoPlayer] failed to get "btnNextTrack"', btnNextTrack);
+                }
+            }
+        } catch (err) {
+            console.error('[VideoPlayer] failed to get playlist', err);
         }
     }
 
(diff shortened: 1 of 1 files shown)


tylers-username:
@thornbill, this MR performs as described for movies and tv shows — resolving the case described in https://github.com/jellyfin/jellyfin-web/pull/7433

When tested against the current functionality, music playlists continue to work as expected in all modes (shuffle, repeat, no-repeat, no-shuffle).

⚠️ Current functionality enables users to go back once they arrive at the end of a movie or tv show playlist. That is no longer possible in this MR as the playlist controls are removed at the end of the playlists, meaning no back button is available.

JustChickNugget:
> @thornbill, this MR performs as described for movies and tv shows — resolving the case described in #7433
> 
> When tested against the current functionality, music playlists continue to work as expected in all modes (shuffle, repeat, no-repeat, no-shuffle).
> 
> ⚠️ Current functionality enables users to go back once they arrive at the end of a movie or tv show playlist. That is no longer possible in this MR as the playlist controls are removed at the end of the playlists, meaning no back button is available.

I was planning to leave it as it was before, but I get your point.

tylers-username:
> > @thornbill, this MR performs as described for movies and tv shows — resolving the case described in #7433
> > When tested against the current functionality, music playlists continue to work as expected in all modes (shuffle, repeat, no-repeat, no-shuffle).
> > ⚠️ Current functionality enables users to go back once they arrive at the end of a movie or tv show playlist. That is no longer possible in this MR as the playlist controls are removed at the end of the playlists, meaning no back button is available.
> 
> I was planning to leave it as it was before, but I get your point.

Yeah, the "⚠️" was to call attention to it—unsure if it was actually right or wrong.

Your latest commit seems to handle this nicely, though. 👍

src/controllers/playback/video/index.js
```diff
@@ -1937,8 +1968,11 @@ export default function (view) {
         })) || [];
     };
 
-    view.querySelector('.btnPreviousTrack').addEventListener('click', function () {
+    view.querySelector('.btnPreviousTrack').addEventListener('click', async function () {
         playbackManager.previousTrack(currentPlayer);
+
+        const currentPlaylist = await playbackManager.getPlaylist();
+        updateTrackButtonsState(currentPlaylist);
```
dmitrylyzo: Why do you need to call `updateTrackButtonsState` from the button/keyboard handlers? Is `onStateChanged` not enough?

src/controllers/playback/video/index.js
```diff
@@ -1620,6 +1614,43 @@ export default function (view) {
...
+        try {
+            if (currentPlaylist) {
+                const btnPreviousTrack = view.querySelector('.btnPreviousTrack');
+                const btnNextTrack = view.querySelector('.btnNextTrack');
+
+                const currentPlaylistIndex = playbackManager.getCurrentPlaylistIndex() + 1;
+
+                if (btnPreviousTrack) {
+                    if (currentPlaylistIndex === 1) {
+                        btnPreviousTrack.classList.add('hide');
```
dmitrylyzo: In TV mode (`Settings / Display / Display mode`), when you hide the button, it looses focus. The focus "fixes" by itself when you press the navigation keys, but that's probably not good for the UX.

src/controllers/playback/video/index.js
JustChickNugget: Didn't see that event function, actually.

src/controllers/playback/video/index.js
JustChickNugget: I don't know if I understood this correctly, but anyway, I tried improving the focus behavior.

src/controllers/playback/video/index.js
```diff
@@ -1620,6 +1610,44 @@ export default function (view) {
...
+                    console.error('[VideoPlayer] failed to get "btnNextTrack"', btnNextTrack);
+                }
+
+                if (currentPlaylist.length === 1 || currentPlaylistIndex === currentPlaylistIndex) {
+                    showOsd();
+                }
+            }
+        } catch (err) {
+            console.error('[VideoPlayer] failed to get playlist', err);
+        }
```
dmitrylyzo: I think you should move `try..catch` to `updatePlaylist` where `getPlaylist` is called.

src/controllers/playback/video/index.js
```diff
@@ -1214,7 +1202,7 @@ export default function (view) {
         }
     }
 
-    function onKeyDown(e) {
+    async function onKeyDown(e) {
```
dmitrylyzo: Now unnecessary.
```suggestion
    function onKeyDown(e) {
```

src/controllers/playback/video/index.js
```diff
@@ -1937,7 +1967,7 @@ export default function (view) {
         })) || [];
     };
 
-    view.querySelector('.btnPreviousTrack').addEventListener('click', function () {
+    view.querySelector('.btnPreviousTrack').addEventListener('click', async function () {
```
dmitrylyzo: Now unnecessary.
```suggestion
    view.querySelector('.btnPreviousTrack').addEventListener('click', function () {
```

src/controllers/playback/video/index.js
```diff
@@ -1949,7 +1979,7 @@ export default function (view) {
     view.querySelector('.btnNextChapter').addEventListener('click', function () {
         playbackManager.nextChapter(currentPlayer);
     });
-    view.querySelector('.btnNextTrack').addEventListener('click', function () {
+    view.querySelector('.btnNextTrack').addEventListener('click', async function () {
```
dmitrylyzo: Now unnecessary.
```suggestion
    view.querySelector('.btnNextTrack').addEventListener('click', function () {
```

src/controllers/playback/video/index.js
```diff
@@ -1620,6 +1610,44 @@ export default function (view) {
...
+                const currentPlaylistIndex = playbackManager.getCurrentPlaylistIndex() + 1;
+
+                if (btnPreviousTrack) {
+                    if (currentPlaylistIndex === 1) {
+                        btnPreviousTrack.classList.add('hide');
+                        btnPreviousTrack.disabled = true;
+                    } else {
+                        btnPreviousTrack.classList.remove('hide');
+                        btnPreviousTrack.disabled = false;
+                    }
```
dmitrylyzo: Use `toggle` when possible.
```suggestion
                    const previousDisabled = currentPlaylistIndex === 1;
                    btnPreviousTrack.classList.toggle('hide', previousDisabled);
                    btnPreviousTrack.disabled = previousDisabled;
```

src/controllers/playback/video/index.js
```diff
@@ -1620,6 +1610,44 @@ export default function (view) {
...
+                }
+
+                if (btnNextTrack) {
+                    if (currentPlaylist.length === currentPlaylistIndex) {
+                        btnNextTrack.classList.add('hide');
+                        btnNextTrack.disabled = true;
+                    } else {
+                        btnNextTrack.classList.remove('hide');
+                        btnNextTrack.disabled = false;
+                    }
```
dmitrylyzo: Use `toggle` when possible.
```suggestion
                    const nextDisabled = currentPlaylist.length === currentPlaylistIndex;
                    btnNextTrack.classList.toggle('hide', nextDisabled);
                    btnNextTrack.disabled = nextDisabled;
```

src/controllers/playback/video/index.js
```diff
@@ -1620,6 +1610,44 @@ export default function (view) {
...
+                } else {
+                    console.error('[VideoPlayer] failed to get "btnNextTrack"', btnNextTrack);
+                }
+
+                if (currentPlaylist.length === 1 || currentPlaylistIndex === currentPlaylistIndex) {
```
dmitrylyzo: `currentPlaylistIndex === currentPlaylistIndex` is always `true`. :thinking:

dmitrylyzo: `currentPlaylistIndex === currentPlaylistIndex` is always `true`. :thinking: 

I believe you need to check `currentVisibleMenu === 'osd'` to keep the OSD hidden (imo, `updateTrackButtonsState` should not show the OSD).

```suggestion
                if (currentVisibleMenu === 'osd') {
```

src/controllers/playback/video/index.js
```diff
@@ -1339,13 +1327,15 @@ export default function (view) {
                 if (e.shiftKey) {
                     e.preventDefault();
                     playbackManager.previousTrack(currentPlayer);
+                    showOsd(btnPreviousTrack);
```
dmitrylyzo: Is `showOsd` called so that the user can see some kind of response to performing a disabled action?

dmitrylyzo (commented):
Tbf, I am not sure if hiding the buttons is a good idea ("jumping buttons"), so I reviewed the PR mainly from a technical point of view. Simply disabling the buttons may be a better choice - need more opinions.

src/controllers/playback/video/index.js
JustChickNugget: Whoops, what a stupid mistake. Had a hard day, sorry

JustChickNugget:
> Tbf, I am not sure if hiding the buttons is a good idea ("jumping buttons"), so I reviewed the PR mainly from a technical point of view. Simply disabling the buttons may be a better choice - need more opinions.

As I mentioned earlier, users may be confused about having different behaviors in two situations described in the issue #7432. This pull request actually addresses more details than the issue describes at the moment, giving a better UX (in my opinion) for controlling the current track on the playback, removing the "Previous track" button at the first track of the playlist, and removing the "Next track" button at the last track of the playlist.

Why? Because as a Jellyfin user, I counted the original behavior confusing, so I thought about making it a little bit better.

ftc2:
interesting.

in my opinion, the ideal behavior for series playback would be:
- when starting/resuming an episode, load **all episodes** (incl previous eps) of the series as a playlist and jump to the target episode.
  - the current behavior seems to be load the target episode + subsequent episodes. sadly, this means you can't go back to the previous episodes using the OSD unless you literally just watched them.
- while playing ep 1, hide `Skip to previous` OSD element
- when playing the final ep, hide `Skip to next` OSD element
- if playlist length is 1, hide both `Skip to previous` and `Skip to next` OSD elements

i'm not sure if my OSD element hiding recommendations are applicable to music. it probably makes sense to always have them visible for music so that you can loop around.

JustChickNugget:
> interesting.
> 
> in my opinion, the ideal behavior for series playback would be:
> 
>     * when starting/resuming an episode, load **all episodes** (incl previous eps) of the series as a playlist and jump to the target episode.
>       
>       * the current behavior seems to be load the target episode + subsequent episodes. sadly, this means you can't go back to the previous episodes using the OSD unless you literally just watched them.
> 
>     * while playing ep 1, hide `Skip to previous` OSD element
> 
>     * when playing the final ep, hide `Skip to next` OSD element
> 
>     * if playlist length is 1, hide both `Skip to previous` and `Skip to next` OSD elements
> 
> 
> i'm not sure if my OSD element hiding recommendations are applicable to music. it probably makes sense to always have them visible for music so that you can loop around.

Well, it works something like this right now. The single thing is that previous episodes aren't loading into the playlist. So, the current logic will hide the "Skip to previous" OSD element even when you are not playing the first episode of series. My PR doesn't target on fixing the playlist creation, but only "Skip to ..." OSD element visibility. So, I suppose when the playlist creation behavior is fixed, this logic will work fine as you described.

PR #7440, opened 2025-12-20
Update ASS subtitles immediately when adjusting subtitle offset

### LINKED ISSUE #7439: Unable to adjust subtitle offset backwards when using ASS

### Describe the bug

When using ASS, the subtitle offset cannot be adjusted backwards without causing some funky behaviour. Adjusting the offset backwards from its current position (whether from 0 or any other position) won't immediately update the subtitles on screen; instead, the last displayed subtitle will stay on screen until playback reaches the point where it should disappear (with the new offset), then subtitles proceed with the new adjusted offset.

This is annoying both in terms of playback and in terms of adjusting the subtitle offset as it also means you cannot preview the effects of a subtitle offset change if you are adjusting the offset backwards.

### Reproduction Steps

1. Play any media with ASS subtitles
2. Adjust the subtitle offset backwards

### Expected/Actual behaviour

As described above.

### Logs

_No response_

### Server version

10.11.5

### Web version

10.11.5

### Build version

10.11.5

### Platform

Windows

### Browser

Chrome

### Additional information

_No response_

abitofevrything:
**Changes**
Fixes an issue where the ASS renderer would not render subtitles when adjusting the offset backwards due to frames for a position further forward being cached.

This PR also makes it so that subtitles update as the subtitle offset scroller is being moved, instead of only when it is released. This is slightly more convenient for setting the offset as desired.

**Issues**
Fixes #7439

### DIFF
--- a/src/components/subtitlesync/subtitlesync.js
+++ b/src/components/subtitlesync/subtitlesync.js
@@ -87,7 +87,7 @@ function init(instance) {
         updateSubtitleOffset();
     };
 
-    subtitleSyncSlider.addEventListener('change', () => updateSubtitleOffset());
+    subtitleSyncSlider.addEventListener('input', () => updateSubtitleOffset());
 
     subtitleSyncSlider.getBubbleHtml = function (_, value) {
         return '<h1 class="sliderBubbleText">'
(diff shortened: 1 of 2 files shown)


dmitrylyzo:
I think it's better to fix the "cache stalling" in https://github.com/jellyfin/JavascriptSubtitlesOctopus (https://github.com/jellyfin/JavascriptSubtitlesOctopus/pull/68).

abitofevrything:
Sounds good, if that fixes the issue. I'll put this as a draft for now and see what happens with your PR.

abitofevrything:
Reopened without the workaround for the underlying issue that has been fixed in the subtitle renderer itself. The dependency will need to be bumped when it releases.

The other change about updating the subtitle offset as the scrollbar is being moved is still a valid change.

abitofevrything:
@nielsvanvelzen any chance this could get a review?

PR #7449, opened 2025-12-25
Fix duplicate media locations

Collin-Swish:
**Changes**
Added logic to ignore duplicate folders in the same library similar to what was done in the equivalent function in mediaLibraryCreator.js

**Other Considerations**
A server side check is also necessary to prevent this bug from happening via the api. I can probably get that done after Christmas. 

**Issues**
Fixes jellyfin/jellyfin#15828


theguymadmax:
This works when you're initially adding to media paths to a library. But if you then come back to an existing library, you can still add duplicate paths. This should probably be fixed on the server side. Maybe you can pick up where this PR left off [#14421](https://github.com/jellyfin/jellyfin/pull/14421)

I would also want to target master instead of the release branch.

Issue #5246, opened 2024-03-01
Firefox / Chromium on Linux can't playback all files, Chrome working


Quadrubo:
**Describe The Bug**
Firefox / Chromium on Linux can't playback all of the files on my server. Chrome on Linux, Firefox on Windows and my Android client are all working with the same files.

Firefox fails with the following console errors and keeps loading forever.
```
Media resource blob:https://jellyfin.l.qudr.de/6eb59baf-dc33-4910-b383-27128f884d3c could not be decoded. [index.html](https://jellyfin.l.qudr.de/web/index.html#!/home.html)
HLS Error: Type: mediaError Details: bufferStalledError Fatal: false [main.jellyfin.bundle.js:2:97852](https://jellyfin.l.qudr.de/web/main.jellyfin.bundle.js?d5f64bca077b6d6aaf33)
HLS Error: Type: mediaError Details: bufferNudgeOnStall Fatal: false [main.jellyfin.bundle.js:2:97852](https://jellyfin.l.qudr.de/web/main.jellyfin.bundle.js?d5f64bca077b6d6aaf33)
Media resource blob:https://jellyfin.l.qudr.de/6eb59baf-dc33-4910-b383-27128f884d3c could not be decoded. [index.html](https://jellyfin.l.qudr.de/web/index.html#!/home.html)
Media resource blob:https://jellyfin.l.qudr.de/6eb59baf-dc33-4910-b383-27128f884d3c could not be decoded, error: Error Code: NS_ERROR_DOM_MEDIA_FATAL_ERR (0x806e0005)
Details: mozilla::MediaResult mozilla::FFmpegDataDecoder<60>::InitDecoder(AVDictionary**): Couldn't open avcodec
```

Chromium fails with the following console errors and ends at the error message `This client isn't compatible with the media and the server isn't sending a compatible media format.`

```shell
4291.b963b8b61cf8a9570109.chunk.js:2 Uncaught DOMException: Failed to read the 'buffered' property from 'SourceBuffer': This SourceBuffer has been removed from the parent media source.
    at SourceBuffer
```

Also notable is that for Firefox & Chromium on Linux, it seems to direct stream. The other browsers are transcoding but I don't know what causes this.

**Media Information**
```
/videos/ebb5fb15-d7ad-e6c1-81bb-77295c1f0026/hls1/main/0.ts

[...]

PR #7455, opened 2025-12-29
Refactor buffer length calculation for HLS player

### LINKED ISSUE #7454: Frequent stuttering during playback.

### Describe the bug

The previous buffering limits were too strict; the maximum buffer length of 6 seconds is now easily triggered.  The current video segment length is approximately 5-6 seconds. Therefore, the loading time for the next segment is too short, leading to frequent buffering issues.
This is especially true in network environments with high latency, where the last segment finishes playing almost immediately after transmission begins.
It is recommended to dynamically adjust the buffer length based on the bitrate to ensure the buffer size remains at 50MB.

### Reproduction Steps

1. Limit bandwidth to 5Mbps
2. Keep latency below 200ms
3. Play videos with a bitrate of 4Mbps or higher

### Expected/Actual behaviour

Smooth playback without lag.

### Media info of the file

```shell

```

### Logs

_No response_

### FFmpeg logs

```shell

```

### Server version

10.11.4

### Web version

10.11.4

### Build version

10.11.4

### Platform

Wndows

### Browser

Chrome

### Additional information

_No response_

### LINKED ISSUE #15787: Jellyfin stutters when playing buffered videos.
### Description of the bug

Jellyfin stutters when playing buffered videos.
The progress bar clearly shows that the currently playing portion has been buffered, and I did not drag the progress bar at all during the entire process. The buffered length is over 30 seconds.

My network experiences packet loss, so to avoid stuttering, I pre-buffer some of the video before starting playback, but the buffered portion still stutters.

I am using Chrome for playback. The video is standard 4K HEVC SDR, and the audio is AAC Stereo.

I'm not sure how to provide debugging logs in Chrome, so please let me know how to do so if needed.

### Reproduction steps

1. simply start playback.

Just stuttering.

Just stuttering.

### Jellyfin Server version

10.11.4

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.4

### Environment

```markdown
- OS: Ubuntu 20.04
- Linux Kernel: 6.8.0-85-generic
- Virtualization: docker
- Clients: Browser
- Browser: Chrome 143.0.7499.42
- FFmpeg Version: Using the official Docker image.(don't know)
- Playback Method: Direct Play
- Hardware Acceleration: none
- CPU Model: Intel(R) Xeon(R) CPU E5-2696 v3 @ 2.30GHz
- GPU Model: none
- Plugins: none
- Reverse Proxy: nginx
- Base URL: none
- Networking: Bridge
- Jellyfin Data Storage & Filesystem:  local HDD - ext4
- Media Storage & Filesystem: local HDD - ext4
- External Integrations: Using the official Docker image.(don't know)
```

### Jellyfin logs

```shell
No relevant information is available.
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

AinzRimuru:
Updated max buffer length calculation based on bitrate and added clamping between 6s and 60s. Removed hardcoded buffer length limits for specific browsers.

**Changes**
Updated max buffer length calculation based on bitrate and added clamping between 6s and 60s. Removed hardcoded buffer length limits for specific browsers.

**Issues**
Fix [#7454](https://github.com/jellyfin/jellyfin-web/issues/7454)
Fix [#15787](https://github.com/jellyfin/jellyfin/issues/15787)


AinzRimuru:
Based on observations, the modified network resources are now being utilized more efficiently. The previous 6-second limitation caused bubbles in the buffering process, resulting in the network being idle or restricted approximately 20% of the time. This means that after the modification, the player's requirements for the network environment will be reduced to some extent.

AinzRimuru:
@thornbill Hi! I noticed you made the last commit here.  Is there a more accurate way to get the current stream's bitrate than the current method? Because if transcoding occurs due to unsupported encoding, I don't know where to get the actual bitrate.

AinzRimuru:
@theguymadmax Hi. I've been waiting for two weeks, and I need some advice.

thornbill:
Here is my advice:
1. Don't ping random people. Maintainers already get enough notifications.
2. Be patient. There are over 200 other open PRs here and this one requires specific knowledge about browser support for hls to review.

AinzRimuru:
> Here is my advice:
> 
> 1. Don't ping random people. Maintainers already get enough notifications.
> 2. Be patient. There are over 200 other open PRs here and this one requires specific knowledge about browser support for hls to review.

I'm sorry, I won't make the same mistake again.

SWW13:
I've cherry-picked this PR onto `10.11.5` and tested it for about a week on my productive Jellyfin instance. This successfully fixes the stuttering issue for remote clients with higher latency (e.g. on a mobile connection).

GeiserX:
For context, I've been working on the same problem from the native client side. Two companion PRs add a user-facing buffer size preset (Auto / Large / Extra large) that controls ExoPlayer's `DefaultLoadControl` durations:

- **Android mobile**: [jellyfin/jellyfin-android#1946](https://github.com/jellyfin/jellyfin-android/pull/1946)
- **Android TV**: [jellyfin/jellyfin-androidtv#5515](https://github.com/jellyfin/jellyfin-androidtv/pull/5515) (fixes [androidtv#2855](https://github.com/jellyfin/jellyfin-androidtv/issues/2855))

The native clients had the same underlying issue — default buffer durations too conservative for variable or high-latency connections. The dynamic approach here (50 MiB / bitrate, clamped 6–60s) is the right call for hls.js since it auto-adapts to stream bitrate without needing a user setting. The user confirmation from @SWW13 on a production instance is encouraging.

Worth noting that PR #7740 proposes a simpler revert to the hls.js 600s default, but that seems riskier given the Chrome 138 regression that motivated #6985 in the first place — even if upstream hls.js fixes are landing, not all users will be on the latest browser. The bitrate-aware clamping here is a safer middle ground.

So I'm also thinking on modifying this approach in my PRs. Maybe we need even more discussion around it. There are many ways of doing it, but you might be on the right track. Thanks @AinzRimuru . Hopefully this can get merged soon

AinzRimuru:
> For context, I've been working on the same problem from the native client side. Two companion PRs add a user-facing buffer size preset (Auto / Large / Extra large) that controls ExoPlayer's `DefaultLoadControl` durations:
> 
> * **Android mobile**: [jellyfin/jellyfin-android#1946](https://github.com/jellyfin/jellyfin-android/pull/1946)
> * **Android TV**: [jellyfin/jellyfin-androidtv#5515](https://github.com/jellyfin/jellyfin-androidtv/pull/5515) (fixes [androidtv#2855](https://github.com/jellyfin/jellyfin-androidtv/issues/2855))
> 
> The native clients had the same underlying issue — default buffer durations too conservative for variable or high-latency connections. The dynamic approach here (50 MiB / bitrate, clamped 6–60s) is the right call for hls.js since it auto-adapts to stream bitrate without needing a user setting. The user confirmation from @SWW13 on a production instance is encouraging.
> 
> Worth noting that PR #7740 proposes a simpler revert to the hls.js 600s default, but that seems riskier given the Chrome 138 regression that motivated #6985 in the first place — even if upstream hls.js fixes are landing, not all users will be on the latest browser. The bitrate-aware clamping here is a safer middle ground.
> 
> So I'm also thinking on modifying this approach in my PRs. Maybe we need even more discussion around it. There are many ways of doing it, but you might be on the right track. Thanks @AinzRimuru . Hopefully this can get merged soon

Further discussion is necessary. The current implementation is not yet sufficiently robust, but I have absolutely no understanding of how Jellyfin's transcoding mechanism works.
The current implementation is unable to accommodate transcoding triggered by a client's lack of support for a specific video codec.
If you know how to retrieve the bitrate of the current stream, please let me know; alternatively, are there any other viable solutions? Thank you.

GeiserX:
I dug through both the web client and the server code a bit.

`setSrcWithHlsJs()` does already have enough context to branch on transcoding: `options.playMethod` is available there, and for transcode playback the web client builds the actual HLS URL from `mediaSource.TranscodingUrl`.

`MediaSourceInfo.Bitrate` does not seem like the right source for this case. On the server it is inferred from the source streams, so it represents the input media bitrate, not the transcoded output bitrate.

The best client-side signal I found is the transcoding URL query itself. On the server, `StreamInfo` serializes `VideoBitrate` and `AudioBitrate` into the transcode URL when they are set, so inside `setSrcWithHlsJs()` you could do something like:

```js
const sourceBitrate =
    options.mediaSource?.MediaStreams?.find(s => s.Type === 'Video')?.BitRate || 8000000;
const maxStreamingBitrate = playbackManager.getMaxStreamingBitrate(this) || sourceBitrate;
let bitrate = Math.min(sourceBitrate, maxStreamingBitrate);

if (options.playMethod === 'Transcode') {
    const params = new URL(url, window.location.href).searchParams;
    const videoBitrate = Number(params.get('VideoBitrate')) || 0;
    const audioBitrate = Number(params.get('AudioBitrate')) || 0;
    const transcodedBitrate = videoBitrate + audioBitrate;
    if (transcodedBitrate > 0) {
        bitrate = transcodedBitrate;
    }
}
```

That should handle the codec-incompatibility case much better than using `mediaSource.MediaStreams[*].BitRate`, because it uses the negotiated transcode target instead of the source bitrate.

One caveat: for HLS this is still the bitrate encoded into the master transcode URL, not necessarily the exact currently selected ABR variant after hls.js switches levels. So it is a much better client-side estimate, but not a perfect "current rendition bitrate" in the literal sense.

If the goal is the exact active HLS level, that probably has to come from hls.js itself after startup rather than from `MediaSourceInfo`.

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -440,15 +440,16 @@ export class HtmlVideoPlayer {
...
+                // Dynamic buffer length based on bitrate
+                // Limit max buffer size to 50MB (50 * 1024 * 1024 * 8 bits = 419430400 bits)
+                const maxBufferSizeBits = 50 * 1024 * 1024 * 8;
+                const sourceBitrate = options.mediaSource?.MediaStreams?.find(s => s.Type === 'Video')?.BitRate || 8000000; // Default 8Mbps
+                const maxStreamingBitrate = playbackManager.getMaxStreamingBitrate(this) || sourceBitrate;
+                // Use the smaller of source bitrate and max streaming bitrate (actual playback bitrate)
+                const bitrate = Math.min(sourceBitrate, maxStreamingBitrate);
+                let maxBufferLength = Math.floor(maxBufferSizeBits / bitrate);
+                // Clamp buffer length between 6s and 60s
+                maxBufferLength = Math.max(6, Math.min(60, maxBufferLength));
```
GeiserX, line 575: When transcoding is triggered by codec incompatibility rather than a bitrate cap, `Math.min(sourceBitrate, maxStreamingBitrate)` may not reflect the actual transcoded output bitrate. The server serializes `VideoBitrate` and `AudioBitrate` into the transcode URL query (`StreamInfo`), so we can parse those for a more accurate estimate.

Note: for HLS this is still the bitrate from the master transcode URL, not necessarily the exact ABR variant after hls.js switches levels — but it's a much better client-side signal than the source stream bitrate for the codec-incompatibility case.

```suggestion
                // Dynamic buffer length based on bitrate
                // Limit max buffer size to 50MB (50 * 1024 * 1024 * 8 bits = 419430400 bits)
                const maxBufferSizeBits = 50 * 1024 * 1024 * 8;
                const sourceBitrate = options.mediaSource?.MediaStreams?.find(s => s.Type === 'Video')?.BitRate || 8000000; // Default 8Mbps
                const maxStreamingBitrate = playbackManager.getMaxStreamingBitrate(this) || sourceBitrate;
                let bitrate = Math.min(sourceBitrate, maxStreamingBitrate);

                // For transcoded streams, prefer the negotiated transcode target from the URL
                if (options.playMethod === 'Transcode') {
                    const params = new URL(url, window.location.href).searchParams;
                    const videoBitrate = Number(params.get('VideoBitrate')) || 0;
                    const audioBitrate = Number(params.get('AudioBitrate')) || 0;
                    const transcodedBitrate = videoBitrate + audioBitrate;
                    if (transcodedBitrate > 0) {
                        bitrate = transcodedBitrate;
                    }
                }

                let maxBufferLength = Math.floor(maxBufferSizeBits / bitrate);
                // Clamp buffer length between 6s and 60s
                maxBufferLength = Math.max(6, Math.min(60, maxBufferLength));
```

GeiserX (commented):
Nice approach with the dynamic buffer calculation. One suggested improvement to better handle transcoding triggered by codec incompatibility (where the source bitrate doesn't reflect the actual transcoded output).

GeiserX:
One more thought on this PR — right now `maxMaxBufferLength` is pinned to `maxBufferLength` (line 460), which means hls.js can never buffer beyond the dynamically calculated target.

In practice, the pinning mostly affects low/medium bitrate streams:
- **1 Mbps stream**: `maxBufferLength` = 60s = ~7.5 MB. There's room to buffer further under the 60 MB `maxBufferSize` ceiling, but hls.js stops at 60s.
- **100 Mbps stream**: `maxBufferLength` = 6s, and `maxBufferSize` would cap it around ~4.8s anyway — pinning has no effect here.

That said, the current pinning was introduced intentionally in #6985 to avoid `bufferAppendError` on Chrome 138, and I don't have browser-specific evidence that the failure mode is gone in current Chrome/Edge builds. So decoupling `maxMaxBufferLength` from `maxBufferLength` here would need separate testing and is probably better as its own PR if someone can verify it.

**What I do think pairs well with this PR** is a user-facing buffer setting — letting users choose between auto (this PR's dynamic calculation), large (~120s), and extra large (~240s) presets. The dynamic calculation would remain the smart default, and users on slow/satellite connections who know their device can handle a deeper buffer could override it.

I've opened companion PRs for the native apps following this same pattern I did (like I told you in an earlier message)

I'll open a web client PR for the user-facing setting shortly, designed to work alongside this PR's auto-calculation if both land.

GeiserX (commented):
Follow-up suggestion: decouple `maxMaxBufferLength` from `maxBufferLength` so hls.js can buffer ahead when bandwidth and memory allow it.

GeiserX:
Minor nit on the PR body: the first issue reference reads `Fix [#5246]` but links to #7454. Might want to fix that for traceability.

AinzRimuru:
> Minor nit on the PR body: the first issue reference reads `Fix [#5246]` but links to #7454. Might want to fix that for traceability.

Thank you for pointing out the mistake, it has been corrected.

PR #7469, opened 2026-01-02
Fix duplicate locations

### LINKED ISSUE #15828: Same folder can be added multiple times to the same library
### Description of the bug

When I add a folder to a library, I can add the same folder multiple times.
When I remove the duplicate folder, all occurances of this folder will be removed.

### Reproduction steps

1. Go to dashboard, libraries and click on any library.
2. now click on + for folders.
3. add C:\Test or /test as a folder.
4. repeat steps 2 and 3 multiple times

The folders are added multiple times to the library.
When you delete one entry, all entries for this folder gets removed, so you have to add it again.

Don't add the same folder to the list if it's already there.

### Jellyfin Server version

10.11.5

### Specify commit id

_No response_

_No response_

### Specify version number

10.11.5

### Specify the build version

10.11.5

### Environment

```markdown
- OS: Debian 13
- Virtualization: Proxmox VM
- Clients: Windows
- Browser: Firefox
```

### Jellyfin logs

```shell
There is nothing in the logs about this.
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

Image

### Additional information

_No response_

Collin-Swish:
**Changes**
Added logic to ignore duplicate folders in the same library in mediaLibraryEditor.js similar to what was done in the equivalent function in mediaLibraryCreator.js 

**Issues**
Fixes jellyfin/jellyfin#15828

### DIFF
--- a/src/components/mediaLibraryEditor/mediaLibraryEditor.js
+++ b/src/components/mediaLibraryEditor/mediaLibraryEditor.js
@@ -59,6 +59,13 @@ function onEditLibrary() {
 function addMediaLocation(page, path) {
     const virtualFolder = currentOptions.library;
     const refreshAfterChange = currentOptions.refresh;
+    const pathLower = path.toLowerCase();
+    const pathFilter = virtualFolder.Locations.filter(p => {
+        return p.toLowerCase() == pathLower;
+    });
+    if (pathFilter.length) {
+        return;
+    }
     ApiClient.addMediaPath(virtualFolder.Name, path, null, refreshAfterChange).then(() => {
         hasChanges = true;
         refreshLibraryFromServer(page);


thornbill (approved):

MBR-0001:
I might be wrong and I know it's a bit of a stretch, but wouldn't this break stuff on case-sensitive filesystems? Like trying to add /mnt/Movies and /mnt/movies

thornbill:
@MBR-0001 good point https://github.com/jellyfin/jellyfin-web/pull/7914

Collin-Swish:
> I might be wrong and I know it's a bit of a stretch, but wouldn't this break stuff on case-sensitive filesystems? Like trying to add /mnt/Movies and /mnt/movies

You are correct that it would cause that issue. I wrote it like that because that was how this same logic was written in the library creation dialog box. This is because I didn't want inconsistent behavior. If we want to allow for case-unique directories it might be appropriate to open that as its own issue.

PR #6919, opened 2025-06-03
Item Card component displays Critic|Community ratings from setting

shoecar:
**Changes**
Two major changes:
- New `cardRatings` display setting
- Item cards display Critic and/or/neither Community based on `cardRatings` value

**Screenshots**
New "Display > Item Card" setting type with "Item Card Ratings Display" setting:
image
image
This controls which of `CriticRating`/`CommunityRating` should be displayed (defaulting to `None`).

Here is what it looks like with `All` selected:
image
You can see no `CriticRating` is shown for shows as that data is not available. Moves have both `CriticRating` and `CommunityRating` so they are shown together.

**Issues**
[3330/rating-on-posters](https://features.jellyfin.org/posts/3330/rating-on-posters)
[discussions/6911](https://github.com/jellyfin/jellyfin-web/discussions/6911)

PR #7471, opened 2026-01-03
Jellyfin ratings

soultaco83:
**Changes**
Reopening pr: https://github.com/jellyfin/jellyfin-web/pull/6919
Two major changes:

*original notes from @shoecar*
New cardRatings display setting
Item cards display Critic and/or/neither Community based on cardRatings value

*Updated changes*
Resolved Merge conflicts
Updated CustomCSS in userSettings to resolve a null error
Updated UserSettings to change from DropDown to Checkbox
The notes from https://github.com/jellyfin/jellyfin-web/pull/6919#pullrequestreview-2970989325  have been resolved. Added checkbox instead of dropdown. Updated color as suggested. Moved community above Critic. 
Intead of at bottom of settings page I thought this would be better to be under Display section

Screenshots: 
image

image

**Issues**
[3330/rating-on-posters](https://features.jellyfin.org/posts/3330/rating-on-posters)
https://github.com/jellyfin/jellyfin-web/discussions/6911

Hoping I updated this correctly. Since master is using experimental view I needed to update react. I have no knowledge of react but basing off existing files I updated to what looked correct.


soultaco83:
ESlint should be fixed. Did not realize tab was not the same as spaces needed.

soultaco83:
demo server
image

soultaco83:
Personally I would like this to be a defaulted on option. Leaving off as I believe the jellyfin team would perfer this to be set by the user. My only concern is users may not know this exists.

soultaco83:
Not sure how I closed it but reopened

PR #7478, opened 2026-01-04
Scale subtitles based on screen size

### LINKED ISSUE #558: Inconsistent size of subtitles

ghost:
Uses viewport height and width to scale subtitles accordingly. Changes the switch statement to use a default size and scaling factor instead of set values. The default size allow for easier size changes accross all size settings and the scaling factor allows size changes for individual size settings.

Fixes #558

### DIFF
--- a/src/components/subtitlesettings/subtitleappearancehelper.js
+++ b/src/components/subtitlesettings/subtitleappearancehelper.js
@@ -5,28 +5,31 @@
 
 function getTextStyles(settings, preview) {
     const list = [];
+    let size = 0;
+    const defaultSize = 3;
 
     switch (settings.textSize || '') {
         case 'smaller':
-            list.push({ name: 'font-size', value: '.8em' });
+            size = defaultSize * 0.8;
             break;
         case 'small':
-            list.push({ name: 'font-size', value: 'inherit' });
+            size = defaultSize * 1;
             break;
         case 'larger':
-            list.push({ name: 'font-size', value: '2em' });
+            size = defaultSize * 2;
             break;
         case 'extralarge':
-            list.push({ name: 'font-size', value: '2.2em' });
+            size = defaultSize * 2.2;
             break;
         case 'large':
-            list.push({ name: 'font-size', value: '1.72em' });
+            size = defaultSize * 1.72;
             break;
         case 'medium':
         default:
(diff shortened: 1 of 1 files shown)


thornbill (commented):
I believe some browsers (TVs) that we support lack support for `vh` and `vw`. We likely need a fallback for those.

PR #7479, opened 2026-01-04
Suspend playback controls when typing in settings

ghost:
Allows typing in the SyncPlay settings while in the video player instead of jumping in the video. 

Added condition to the playback controls script that checks if an html input element is active. If the condition is true it suspends all inputs related to playback control. Alternatively the check can be moved to the number keys case to suspend only those specific inputs.

``` case '0':
case '1':
case '2':
case '3':
case '4':
case '5':
case '6':
case '7':
case '8':
case '9': { // no Shift
  e.preventDefault();
  -> HERE <-
  const percent = parseInt(key, 10) * 10;
  playbackManager.seekPercent(percent, currentPlayer);
  break;
}```

### DIFF
--- a/src/controllers/playback/video/index.js
+++ b/src/controllers/playback/video/index.js
@@ -1222,6 +1222,9 @@ export default function (view) {
         // Skip modified keys
         if (isKeyModified) return;
 
+        // Suspend video controls when clicked into input element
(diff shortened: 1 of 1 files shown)


FelixRoediger:
@thornbill
I would appreaciate a review when you have time :)

PR #7482, opened 2026-01-06
Allow sorting books by index number

### LINKED ISSUE #7404: Sort by book index number

### Describe the bug

I just tried using Jellyfin to host my comics library which is properly tagged with ComicVine data in embeded `ComicInfo.xml`.

The `Bookshelf` plugin parsed the CBZ files properly. However when viewing a comic with lots of issues, the books are sorted by the title, which does not always follow the order of the series.

I can see `Bookshelf` extracts the `Number` field in to `IndexNumber` but there is no way in the UI to sort via that field. It would make reading comic a way better experience if we can add this in the sort by field.

### Reproduction Steps

1. Go to a book library with comics
2. click "sort by" option
3. No "Index Number" option available

### Expected/Actual behaviour

"Index Number" field is available as sort.

### Logs

_No response_

### Server version

10.11.4

### Web version

10.11.4

### Build version

10.11.4

### Platform

Windows

### Browser

Chrome

### Additional information

_No response_

AlkyoneAoide:
**Changes**
- Adds Index Number to the list of available sorting methods for books in the experimental layout

**Issues**
Fixes #7404

**Comments**
Assuming you have properly formed metadata, sorting serials and periodicals by index number (equivalent to something like issue number) is really helpful for having a pleasant browsing/reading experience!
A demonstration, sorting volumes of One Piece by Name (default) vs. Index Number.  Sorting by name produces out-of-order results, while sorting by index number organizes the volumes in order.

I hope I did everything right lol, first time contributing and I absolutely could not figure out how (or if it's possible yet) to force a book-type library to use the experimental UI.  And thanks to everyone for such an awesome project!


PeachesMLG (approved):

dkanada (approved):
Tested and working on the latest commit in both layouts.

src/controllers/list.js
```diff
@@ -1070,6 +1070,12 @@ class ItemsView {
             name: globalize.translate('Runtime'),
             value: [ItemSortBy.Runtime, ItemSortBy.SortName].join(',')
         });
+
+        sortBy.push({
+            name: globalize.translate('IndexNumber'),
+            value: [ItemSortBy.IndexNumber, ItemSortBy.SortName].join(',')
+        });
```
dmitrylyzo: Shouldn't we check if the type is `Books`?

AlkyoneAoide: As far as I can tell, that sort menu is only used by book type libraries in the legacy layout, and it seems like books can't check library type the same way as everything else (`this.params.type` is `null` for books), so I didn't feel the need to worry about checking type.

thornbill: It's kind of a catch all for anything that doesn't have a dedicated view, so yes we should add a check. :slightly_smiling_face:

thornbill: Alternatively we could remove the option from the legacy list view since the "experimental" layout is the default in the next release.

AlkyoneAoide: Oh, sure!  That sounds fine!

MBR-0001: folders inside a book library still use the old UI so the option to sort by index is missing there, but considering the complexity of suporting this case we can go without this and (hopefully) have that implemented at some point in the future

MBR-0001 (approved):
Worked in my crude testing, aside from folder stuff mentioned above, LGTM

image
sorted:
image

Gauvino (approved):
LGTM, thanks for the fix, was waiting for it

PR #7489, opened 2026-01-08
Fix artist links in recently added

Shadowghost:
`ServerId` was always undefined when the Artist links are generated, leading to invalid links

**Changes**
* Pass serverId to URL builder


thornbill:
Closing since #7496 fixed this with a different approach

PR #7491, opened 2026-01-11
Fix jellyfin sdk type errors

PeachesMLG:
**Changes**
Fixed the Jellyfin type errors when updating to latest unstable SDK version

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
         "@fontsource/noto-sans-sc": "5.2.6",
         "@fontsource/noto-sans-tc": "5.2.6",
         "@jellyfin/libass-wasm": "4.2.3",
-        "@jellyfin/sdk": "0.0.0-unstable.202512091852",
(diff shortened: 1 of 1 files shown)


src/components/cardbuilder/Card/cardHelper.ts
```diff
@@ -598,6 +600,7 @@ function getMediaTitle(cardOptions: CardOptions, item: ItemDto): TextLine {
             });
 
     return getTextActionButton({
+        Key: item.Id ?? '',
```
PeachesMLG, line 603: Honestly no idea what Key actually is here, also doesnt look like its used so though id just fallback to the Id

PeachesMLG:
Not fully sure why the stale label has been added, but merge conflicts have been resolved

thornbill:
The stale label doesn't get removed immediately... it should when stalebot runs next though

PR #7496, opened 2026-01-14
Fix missing server id

thornbill:
**Changes**
Fixes “undefined” server ids in urls

**Issues**
N/A

Alternative to https://github.com/jellyfin/jellyfin-web/pull/7489


thornbill:
> I never really understood why jellyfin-web even needs to send the server id everywhere, you can't really connect to multiple servers at the same time anyway. But that's outside the scope of this PR.

My guess is that it goes back to the Emby Connect functionality... I definitely think we'll rip it out eventually

PR #7503, opened 2026-01-18
Migrate wizard pages to React

viown:
**Changes**
Migrates the startup wizard pages to React + TS SDK

| image |image |
|-------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------|
| image | image |
image

**Issues**
N/A

PR #7504, opened 2026-01-18
Navigate to wizard if not completed on server add

viown:
When adding a server to web, it will navigate to the login page even if the wizard process hasn't been completed yet. This small fix allows the user to configure the startup wizard process on the web-wrapping apps.

**Changes**
Navigate to wizard if not completed on server add

**Issues**
N/A

### DIFF
--- a/src/controllers/session/addServer/index.js
+++ b/src/controllers/session/addServer/index.js
@@ -16,7 +16,11 @@ function handleConnectionResult(page, result) {
             break;
         }
         case ConnectionState.ServerSignIn:
-            Dashboard.navigate('login?serverid=' + result.Servers[0].Id, false, 'none');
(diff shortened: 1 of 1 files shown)


dmitrylyzo:
We need to disable the setup wizard in TV mode (as we do with [Dashboard](https://github.com/jellyfin/jellyfin-web/blob/d8ef1e5137f23d6d26cfe02b2c6dffe2b9d75145/src/apps/stable/routes/user/settings/index.tsx#L257)) and add a page to inform the user or display a popup message. The latter should probably be done in the PR on master branch in case we need translations.

viown:
The current wizard should work on TVs though? Unless you're referring to my [other PR](https://github.com/jellyfin/jellyfin-web/pull/7503) in which case we could add an alert dialog if a user tries to add an incomplete server on TV.

PR #7516, opened 2026-01-22
Wait for app host to initialize before proceeding

nielsvanvelzen:
On Titan OS the only way to retrieve device information is using an asynchronous function. Unfortunately jellyfin-web registers an ApiClient immediately on load making a web-wrapper unable to execute any async code to set device info.

This PR changes the initialization code to not call appHost.init() once from the app initialization code, before attempting to do any api client stuff. Giving some room for web wrappers to do async work.

**Changes**

- Move `appHost.init()` call from `apphost.js` module load
- Await `appHost.init()` call - only affects web-wrappers that return a promise like jellyfin-android and jellyfin-titanos.

**Issues**

### DIFF
--- a/src/components/apphost.js
+++ b/src/components/apphost.js
@@ -456,5 +456,3 @@ if (window.addEventListener) {
     window.addEventListener('blur', onAppHidden);
 }
 
-// load app host on module load
-appHost.init();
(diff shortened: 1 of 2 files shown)


nielsvanvelzen:
I do not feel responsible for failing CI

PR #7523, opened 2026-01-23
Add support for Plugin Page MenuIcon

jumoog:
**Changes**
* Support Plugin Page MenuIcon
* Plugin Author must use `material-ui/material-icons`

grafik

Example Plugin Commit:

https://github.com/intro-skipper/intro-skipper/commit/8c51c33ff3d821051335a63ea1896bfa7a65050c

**Issues**
* fix TODO

### DIFF
--- a/src/apps/dashboard/components/drawer/sections/PluginDrawerSection.tsx
+++ b/src/apps/dashboard/components/drawer/sections/PluginDrawerSection.tsx
@@ -1,5 +1,6 @@
 import Extension from '@mui/icons-material/Extension';
 import Folder from '@mui/icons-material/Folder';
+import Icon from '@mui/material/Icon';
 import List from '@mui/material/List';
 import ListItemIcon from '@mui/material/ListItemIcon';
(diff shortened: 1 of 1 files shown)


jumoog:
In my example, I used `SkipNext`, but the icon is `skip_next`. In my opinion, it's easier if the plugin developer uses the name and we convert it here.

Always use the folder icon as a fallback.

jumoog:
The reason for targeting 10.11 is that it is just a small cosmetic feature

nielsvanvelzen:
Supporting icons for plugin pages is definitely a nice feature to have, however we should **NOT** have it rely on the icon set that jellyfin-web uses. There is a high change we will be switching our icon set to a different system, and even then icons may be changed/removed when the used set is updated.

It would be better to have the plugin provide an actual svg/png file. This would need changes to both the server/web.

jumoog:
> It would be better to have the plugin provide an actual svg/png file. This would need changes to both the server/web.

Yeah, that does sound like a good idea in the long term.

PR #7524, opened 2026-01-23
Add option to enable Dolby Vision and Dolby Vision Profile 10

antaljanosbenjamin:
**Changes**
I added two options to the Video Advanced section of playback settings:
1. Enable Dolby Vision: it overrides the value that the device reports from itself.
2. Enable Dolby Vision Profile 10: it enables playing Dolby Vision Profile 10 even if the browser cannot play it.

I added these options because I think I among the people who has a TV that reports these values wrong. My TV (LG 4K Ultra HD TV 55UP77006LB (2021)) reports that it doesn't support Dolby Vision and `supportedDolbyVisionProfileAv1` says it cannot play Dolby Vision Profile 10 files, but my TV has not issues with it. I tried with multiple files from [Kodi's DV sample files](https://kodi.wiki/view/Samples) and they worked. I thought it might be useful for other people, so I think it makes sense to contribute this to the project.

image

**Issues**


src/scripts/browserDeviceProfile.js
```diff
@@ -1235,7 +1235,7 @@ export default function (options) {
             hevcVideoRangeTypes += '|DOVIWithEL|DOVIWithELHDR10Plus|DOVIInvalid';
         }
 
-        if (supportedDolbyVisionProfileAv1(videoTestElement)) {
+        if (appSettings.enableDolbyVisionProfile10() || supportedDolbyVisionProfileAv1(videoTestElement)) {
```
nyanmisaka, line 1238: Have you checked that [DV Profile 5](https://mega.nz/file/2TYwVSTZ#LBh2CoX3QUfkR_kwey_aim9QWvwk3UpLQozEG9n2ww0) video (pure DV used to verify its tone mapping capabilities) plays correctly in color, even though the webOS browser on your [55UP77006LB](https://www.lg.com/uk/tvs-soundbars/smart-tvs/55up77006lb/) doesn't report DV support?

If so, then adding an advanced option should suffice. For another advanced option regarding Profile 10, simply check for AV1 support and combine it with the first advanced option.

Finally, it would be ideal if the "Video Advanced" options could only be displayed on webOS + Tizen, and hide on other platforms. Otherwise, it might confuse users and cause issues.

antaljanosbenjamin: > Have you checked that [DV Profile 5](https://mega.nz/file/2TYwVSTZ#LBh2CoX3QUfkR_kwey_aim9QWvwk3UpLQozEG9n2ww0) video (pure DV used to verify its tone mapping capabilities) plays correctly in color, even though the webOS browser on your [55UP77006LB](https://www.lg.com/uk/tvs-soundbars/smart-tvs/55up77006lb/) doesn't report DV support?

Good that you asked, because I just digged into this in the past couple of days and I am still learning about Dolby Vision. I haven't tried it before and it doesn't work. However in my opinion this shouldn't be a problem, because `supportedDolbyVisionProfilesHevc` checks separately for profile 5 and 8. Profile 5 is problematic and therefore the above mentioned function only returns 8 as supported profile. As a result Dolby Vision Profile 5 (`DOVI` if I got it correctly) video range type is not reported as directly playable. ~~As far as I understood, it is okay like this, if not, please tell me either what is problematic (and I can try to come up with a solution) or what should I do.~~

Edit: I realized that the option and setting naming is a bit confusing. It doesn't enable pure DV, but only DV with some HDR fallback. So I would say something like "Enable Dolby Vision with HDR fallback" and `enableDolbyVisionWithHdrFallback` is a much better naming. If that work for you, I rename both the option and setting.

> If so, then adding an advanced option should suffice. For another advanced option regarding Profile 10, simply check for AV1 support and combine it with the first advanced option.

Thanks for the suggestion, I don't like the second option, so removing it makes me happy.

> Finally, it would be ideal if the "Video Advanced" options could only be displayed on webOS + Tizen, and hide on other platforms. Otherwise, it might confuse users and cause issues.

I can definitely do that. Just to be sure, did you mean the whole "Video Advanced" section, including DTS and TrueHD or just the new option?

nyanmisaka: I understand now. Your LG TV model doesn't actually support Dolby Vision, as it reported to Jellyfin. Instead, it supports HDR10 and the HLG fallback layer included in DV Profile 8 & 10.

So the problem is simplified, we shouldn't need a new checkbox. It will use the same solution as Samsung TVs that do not support DV at all. We just need to modify it like this:
```diff
diff --git a/src/scripts/browserDeviceProfile.js b/src/scripts/browserDeviceProfile.js
index 593076caa7..917d424e5b 100644
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -1190,7 +1190,7 @@ export default function (options) {
     let vp9VideoRangeTypes = 'SDR';
     let av1VideoRangeTypes = 'SDR';
 
-    if (browser.tizenVersion >= 3) {
+    if (browser.tizenVersion >= 3 || (browser.web0s && !supportsDolbyVision(options))) {
         hevcVideoRangeTypes += '|DOVIWithSDR';
     }
 
@@ -1200,7 +1200,7 @@ export default function (options) {
         vp9VideoRangeTypes += '|HDR10|HDR10Plus';
         av1VideoRangeTypes += '|HDR10|HDR10Plus';
 
-        if (browser.tizenVersion >= 3 || browser.vidaa) {
+        if (browser.tizenVersion >= 3 || browser.vidaa || (browser.web0s && !supportsDolbyVision(options))) {
             // Tizen TV does not support Dolby Vision at all, but it can safely play the HDR fallback.
             // Advertising the support so that the server doesn't have to remux.
             hevcVideoRangeTypes += '|DOVIWithHDR10|DOVIWithHDR10Plus|DOVIWithEL|DOVIWithELHDR10Plus|DOVIInvalid';
@@ -1215,7 +1215,7 @@ export default function (options) {
         vp9VideoRangeTypes += '|HLG';
         av1VideoRangeTypes += '|HLG';
 
-        if (browser.tizenVersion >= 3) {
+        if (browser.tizenVersion >= 3 || (browser.web0s && !supportsDolbyVision(options))) {
             hevcVideoRangeTypes += '|DOVIWithHLG';
         }
     }
```

If you test that it works, you can modify the PR target branch to version 10.11.z so that it can be included as a minor fix in the next stable release.

antaljanosbenjamin: Ah, thanks for the guide! I will check this later today and if it works, will change the PR.

antaljanosbenjamin: I will close this PR and open another one to the version 10.11.z. I think that is the cleaner solution. I have to force push my branch or change both branches in this PR. Furthermore, my branch name also doesn't suit the new changes. I hope you don't mind.

nyanmisaka: I have no problem with that. At least the current branch name is no longer relevant.

PR #7526, opened 2026-01-24
Allow to play Dolby Vision with fallback on webOS

antaljanosbenjamin:
**Changes**
Allowed to play Dolby Vision files with fallback on webOS when Dolby Vision is not supported.

I run into this issue by noticing that my TV can play some files from a USB stick, but Jellyfin rejects to direct play them. I have created another [PR](https://github.com/jellyfin/jellyfin-web/pull/7524) previously, but because of the radically different approach that @nyanmisaka proposed, I thought it is better to open a new PR.

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -1185,7 +1185,9 @@ export default function (options) {
     let vp9VideoRangeTypes = 'SDR';
     let av1VideoRangeTypes = 'SDR';
 
-    if (browser.tizenVersion >= 3) {
+    const isWebOsWithoutDolbyVision = browser.web0s && !supportsDolbyVision(options);
+
+    if (browser.tizenVersion >= 3 || isWebOsWithoutDolbyVision) {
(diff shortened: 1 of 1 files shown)


andresilva:
Can confirm this worked for me on LG G4 with webOS 25. Without this patch some files would result in black screen unless HLS was used, and with HLS it would always fallback to HDR and never use DV. I think this fix is actually accidental though, because my TV **does** support DV, it's just that without this patch it's not advertised?

antaljanosbenjamin:
@andresilva are you sure no transcoding is going on in the background? Because I think this patch shouldn't change how dolby vision capable TVs should handle it. Maybe your TV doesn't report it dolby vision capability properly?

andresilva:
No transcoding is happening, my TV **does** support DV, so I believe this patch is fixing the issue accidentally, i.e. DV gets advertised for the fallback, but since my TV supports DV it doesn't use the fallback. #7328 is the proper explicit fix, but the end result is the same for me with either patch. I don't really know that much about jellyfin or DV, so it's possible my understanding is completely wrong here.

Issue #6456, opened 2025-01-17
“Scheduled tasks” on sidebar drawer blocked by music player bar


syfq91:
**Describe The Bug**
The “scheduled tasks” on the side bar is blocked by music player. Happens on mobile web browsers and also Jellyfin Expo app. 

**Steps To Reproduce**
1. Play some music
2. Click on 'Top Left menu' to open sidebar drawer 
3. Click on ‘Dashboard’
4. Scroll down to the bottom 
5. See “scheduled tasks” blocked behind the music player

**Expected Behavior**
Able to see the last item (scheduled tasks)

**Logs**

**Screenshots**

![Image](https://github.com/user-attachments/assets/9abc8d4d-0e65-4cf5-8ef5-eaaaa0d7c64c)

![Image](https://github.com/user-attachments/assets/0e17d40a-3d6b-455d-a27b-df41a7345849)

**System (please complete the following information):**
 - Platform: iPhone 13
 - Browser: Safari
 - Jellyfin Version: 10.10.3

**Additional Context**

syfq91:
Still happens on 10.10.7

![Image](https://github.com/user-attachments/assets/3f08c8e7-e6a8-4b59-95ab-5577fefc9cb7)

syfq91:
still not fix

syfq91:
Still happens in 10.11.5

PR #7529, opened 2026-01-26
Add padding to mobile drawer

viown:
**Changes**
Add padding to mobile drawer to make room for the now playing bar.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-web/issues/6456


dfederm:
This will always add padding even when the now playing bar is hidden? Seems like that would lead to some awkward wasted space. It does appear to match desktop behavior, although I suspect it's less of an issue in practice on desktop.

Consider instead something like:
```css
/* Only add padding when nowPlayingBar is visible (doesn't have .hide class) */
body:has(.nowPlayingBar:not(.hide)) .MuiDrawer-paper {
    padding-bottom: 4.2rem;
}
```

thornbill:
I have no strong opinion either way. What do you think @viown?

Issue #7505, opened 2026-01-18
Video title has been removed on player


PeachesMLG:
### This issue respects the following points:

- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-web/issues?q=is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

The video title no longer appears when playing media if you are using the experimental layout

Before

<img width="1072" height="203" alt="Image" src="https://github.com/user-attachments/assets/d542d9f5-2889-4191-824b-91fcb533aa8a" />

After

<img width="1720" height="117" alt="Image" src="https://github.com/user-attachments/assets/6c2c703d-731b-42e4-9508-f85728c4218b" />

### Reproduction Steps

1. Enable experimental mode
2. Play a video
3. Observe no title

### Expected/Actual behaviour

There should be a title

### Logs

N/A

### Server version

10.11.3

### Web version

10.11.3

### Build version

10.11.3

### Platform

Linux

### Browser

Firefox

theguymadmax:
Dupe of #7316

Issue #7316, opened 2025-11-08
Movie/Episode title does not show on Firefox in player


JPVenson:
### This issue respects the following points:

- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-web/issues?q=is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

Left Edge, right Firefox

<img width="3839" height="1985" alt="Image" src="https://github.com/user-attachments/assets/100ee83a-d251-4755-a09e-f92157cb8b85" />

### Reproduction Steps

play any movie/episode on firefox

### Expected/Actual behaviour

Media title shows on firefox at the same place as edge does

### Server version

10.10.1

### Web version

10.10.1

### Build version

10.10.1

### Platform

Win11

### Browser

Firefox

imheinrichlunge:
It happens to me too, but in my case a black screen appears, although only with one movie in AV1. I don't know if the video track is corrupted or something, because it's the only movie in AV1 in my entire collection that has this error; the others with AV1 don't. On the other hand, Edge plays it without any problems.

theguymadmax:
I think it might be an issue with the display mode. The title shows up when I use the desktop display mode, but doesn't appear using the experimental layout, regardless of browser type.

imheinrichlunge:
Exactly, I just tested all the display modes and what @theguymadmax says is true: if the display mode is set to automatic/experimental, the errors mentioned in my case occur. With the other values, the movie plays without any problems.

PR #7530, opened 2026-01-26
Add video title to experimental layout

viown:
**Changes**
Adds the video title to the experimental layout top bar.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-web/issues/7505
Fixes https://github.com/jellyfin/jellyfin-web/issues/7316

### DIFF
--- a/src/controllers/playback/video/index.js
+++ b/src/controllers/playback/video/index.js
@@ -192,6 +192,7 @@ export default function (view) {
         if (!item) {
             updateRecordingButton(null);
             LibraryMenu.setTitle('');
+            Events.trigger(document, EventType.VIDEO_TITLE_CHANGE, [ '' ]);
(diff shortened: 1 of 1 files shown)


src/controllers/playback/video/index.js
```diff
@@ -261,6 +261,7 @@ export default function (view) {
         }
 
         LibraryMenu.setTitle(title);
+        Events.trigger(document, EventType.VIDEO_TITLE_CHANGE, [ title ]);
```
dfederm, line 265: I believe you also need to call this above in the `if (!item)` block just after `LibraryMenu.setTitle('');`:

```js
Events.trigger(document, EventType.VIDEO_TITLE_CHANGE, [ '' ]);
```

viown: Good call, thanks.

Issue #5040, opened 2023-03-03
[Issue]: Collections delete prompt is misleading


cinderblockgames:
### Please describe your bug

If you click Delete on a Collection, you get the following prompt:

```
Deleting this item will delete it from both the file system and your media library. Are you sure you wish to continue?
```

Collections don't exist on the file system, and it gives me a pause every time, like "am I about to delete my movies by accident?"

### Jellyfin Version

10.8.0

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

alexfornuto:
It looks like this uses the value from the key `ConfirmDeleteItem`, which is used by `deleteHelper.js`, which is used by  `itemContextMenu.js`, which is used by... you get the idea.

I wholeheartedly agree with this improvement and wanted to submit a patch myself, but it's over my non-developer head to do. I reckon there would either need to be a new `itemContextMenu` specific for collections, or some sort of catch for when it's used in the Collections page.

alexfornuto:
Non-stale update. The problem has been identified, and solutions proposed, it just needs a brave soul to make it happen.

ibizaman:
Not stale just yet!

kevin201:
Agree. It's a very misleading wording for a warning message. It's the equivalent of deleting a media library, and should be treated as such, IMHO.

lichenophile:
Wild that this issues so old haha

> Deleting this collection will delete it from your media library as well as the collection folder on the file system and all its content. The original media files will be left untouched. Are you sure you wish to continue?

Can anyone clarify for me? I now understand from this that the `Delete` option under the collection `...` menu does not delete my media, but I want to confirm that it also does not remove media from my pre-existing libraries. (IE it restores my library view to the state it was before the Collection 'grouping')

Logically, I want to assume deleting a collection does nothing but delete the grouping and is entirely safe to do but boy it'd be a pain if I had to readd things to my libraries and lost playback stats, metadata etc

(I created a massive collection for anime series + anime movies a while back, have been scared to delete it but want to remove it as it hides movies from the library view, they're all grouped under the collection instead)

NoricSteel:
> Wild that this issues so old haha
> 
> Can anyone clarify for me? I now understand from this that the `Delete` option under the collection `...` menu does not delete my media, but I want to confirm that it also does not remove media from my pre-existing libraries. (IE it restores my library view to the state it was before the Collection 'grouping')
> 
[...]

PR #7534, opened 2026-01-26
Clarify Playlist and Collections delete prompt

### LINKED ISSUE #3837: Deleting a playlist have a too much "threatening" alert message
**Describe The Bug**
When deleting a playlist a dialogue with the following message is displayed:

> **Delete Item?**
> Deleting this item will delete it from both the file system and your media library. Are you sure you wish to continue?

It makes think that media inside the playlist are going to be deleted too.

**Steps To Reproduce**

1. Create a playlist
2. Try to delete it

**Expected Behaviour**

The message should be more specific to playlists or less "threatening".

**System**
 - Platform: Windows 10
 - Browser: Firefox 103
 - Jellyfin Version: 10.8.1

dfederm:
Fixes #3837,#5040

This change simply special-cases the text shown when deleting a Playlist or Collection. This clarifies that only the playlist or collection is being deleted, not the underlying media.

### DIFF
--- a/src/strings/en_US.json
+++ b/src/strings/en_US.json
@@ -166,7 +166,7 @@
     "CinemaModeConfigurationHelp": "Cinema mode brings the theater experience straight to your living room with the ability to play trailers and custom intros before the main feature.",
     "SelectAudioNormalizationHelp": "Track gain - Adjusts the volume of each track so they playback at the same loudness.\nAlbum gain: Adjusts the volume of all tracks in an album as a group, preserving the album’s original dynamic range. Switching between “Off” and other options requires restarting the current playback.",
     "ClearQueue": "Clear queue",
-    "ClientSettings": "Client side settings",
+    "ClientSettings": "Client Settings",
     "Collections": "Collections",
     "Colorist": "Colorist",
     "ColorPrimaries": "Primary colors",


src/strings/en-us.json
```diff
@@ -178,6 +178,8 @@
     "ConfigureDateAdded": "Set up how metadata for 'Date added' is determined in the Dashboard > Libraries > Display",
     "ConfirmDeleteImage": "Delete image?",
     "ConfirmDeleteItem": "Deleting this item will delete it from both the file system and your media library. Are you sure you wish to continue?",
+    "ConfirmDeleteCollection": "Deleting this collection will remove it from your media library. The items in this collection will not be deleted. Are you sure you wish to continue?",
+    "ConfirmDeletePlaylist": "Deleting this playlist will remove it from your media library. The items in this playlist will not be deleted. Are you sure you wish to continue?",
```
viown: How about just a simple "Are you sure you'd like to delete this playlist / collection?" I don't see a reason why the user would assume the items would be deleted as well, although I'm aware the current message can seem misleading.

dfederm: Sounds good, I did leave these quite verbose. Might as well just simplify it since the "scarier" delete message is more verbose for the reason of clarity, while these require less so.

PR #1945, opened 2020-09-24
SyncPlay for TV series (and Music)

OancaAndrei:
**Changes**
This PR aims at making SyncPlay more tv-series friendly, by adding new states that handle the transition between one episode and another.

This PR is related to the [server side one](https://github.com/jellyfin/jellyfin/pull/3194).

**Issues**
Fixes #1777.

PR #7535, opened 2026-01-27
Remove old irrelevant comment about 'LabelSyncPlayTimeOffset'

dfederm:
This simply removes a TODO in `src/components/playerstats/playerstats.js` since the strings are now gone

### DIFF
--- a/src/components/playerstats/playerstats.js
+++ b/src/components/playerstats/playerstats.js
@@ -352,7 +352,6 @@ function getSyncPlayStats() {
     });
 
     syncStats.push({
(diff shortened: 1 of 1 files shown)


dfederm:
@nielsvanvelzen  are you sure? The string was removed from en-us.json back in 2020 (#1945) and it still isn't cleaned up from the other language files.

nielsvanvelzen:
Yeah there was a configuration issue so it wasn't working... it's fixed now, thanks for catching that haha

dfederm:
Well, that trivializes this PR then, just a comment removal 😆

PR #7554, opened 2026-02-01
Add Still watching prompt. (Netflix style playback sleep after n plays)

ShanePadellaro:
Adds a user configurable Netlfix style still watching prompt, that asks if a users is still watching after N auto plays. When the prompt is triggered the user can either disable the check for the duration ("Don't Ask me again") of that playback session or click "Ask me again later" and reset the play counter.

- Manually skipping by using next and previous buttons resets the playback counter (user interaction).
- Starting a new session resets the counter

Screenshot 2026-02-01 at 17 43 34

Settings -> Playback
- Enable or disable Are you still watching prompt
- Configure how many plays until the prompt is displayed

Screenshot 2026-02-01 at 17 43 57

**Changes**
playbackmanger.js
- Added an autoplay counter and logic for incrementing the counter after each item is played
- Added state variable to track if the user has selected "Don't ask me again" to suppress the prompt for the session
- Extracted counter logic into class methods.
- Trigger event when still watching threshold is reached so that the prompt can be triggered from a different class and maintain separation of logic and UI in playbackmanger.js

/src/controllers/playback/video/index.js
- Added an event handler that listens for the event triggered by playbackmanger.js and display the prompt to the user as well as handles the button click callbacks.


src/strings/en-us.json
```diff
@@ -302,6 +302,14 @@
...
     "EnableRewatchingNextUp": "Enable Rewatching in Next Up",
     "EnableRewatchingNextUpHelp": "Enable showing already watched episodes in 'Next Up' sections.",
+    "AreYouStillWatchingTitle": "Are you still watching?",
+    "AreYouStillWatchingSettingTitle": "Ask if the user is still watching after several consecutive auto-plays.",
+    "AreYouStillWatchingText": "Do you want to keep watching?",
+    "AreYouStillWatchingSettingDescription": "If user is not active, ask if the user is still watching.",
+    "LabelAreYouStillWatchingCount": "Auto-plays before prompting",
+    "LabelAreYouStillWatchingCountHelp": "Number of consecutive auto-plays allowed before asking if you're still watching (default 5).",
+    "ButtonKeepWatchingAskLater": "Keep watching and ask me again later",
+    "ButtonDontAskAgain": "Don't ask me again",
```
dmitrylyzo: Stick to the alphabetical order.

PR #7557, opened 2026-02-02
Add default tab options for Home Videos and Photos library 

theguymadmax:
**Changes**
In the experimental layout, the Home Videos and Photos library now supports tabs. This PR allows selecting which tab (Photos, Photo Albums, or Videos) is used as the default when opening the library.

**Issues**
 N/A

### DIFF
--- a/src/components/homeScreenSettings/homeScreenSettings.js
+++ b/src/components/homeScreenSettings/homeScreenSettings.js
@@ -52,123 +52,133 @@ function getLandingScreenOptions(type) {
     const list = [];
 
     if (type === 'movies') {
-        list.push({
-            name: globalize.translate('Movies'),
-            value: LibraryTab.Movies,
-            isDefault: true
-        });
-        list.push({
-            name: globalize.translate('Suggestions'),
-            value: LibraryTab.Suggestions
-        });
-        list.push({
-            name: globalize.translate('Favorites'),
-            value: LibraryTab.Favorites
-        });
-        list.push({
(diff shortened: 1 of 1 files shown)


src/components/homeScreenSettings/homeScreenSettings.js
```diff
@@ -155,6 +155,20 @@ function getLandingScreenOptions(type) {
             name: globalize.translate('Series'),
             value: LibraryTab.SeriesTimers
         });
+    } else if (type === 'homevideos') {
+        list.push({
```
thornbill: sonarcloud makes a good point... we can just make one call to `list.push` here and pass all three options

src/components/homeScreenSettings/homeScreenSettings.js
```diff
@@ -248,7 +262,7 @@ function getPerLibrarySettingsHtml(item, user, userSettings) {
         html = `<div class="checkboxListContainer">${html}</div>`;
     }
 
-    if (item.CollectionType === 'movies' || item.CollectionType === 'tvshows' || item.CollectionType === 'music' || item.CollectionType === 'livetv') {
+    if (item.CollectionType === 'movies' || item.CollectionType === 'tvshows' || item.CollectionType === 'music' || item.CollectionType === 'livetv' || item.CollectionType === 'homevideos') {
```
thornbill: this is probably getting to the point that it would be better to do `['movies', ... ].includes(item.CollectionType)` instead

PR #7559, opened 2026-02-04
Update 'Trickplay' capitalization in en-us.json

HammyHavoc:
Corrected a lingual inconsistency.

**Changes**

**Issues**


thornbill:
"Trick play" is a general industry term for this functionality. I don't think it should be capitalized.

Issue #7561, opened 2026-02-05
Dashboard no longer update all activity


solbu:
### This issue respects the following points:

- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-web/issues?q=is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

I don't know when it happened, but after upgrading from 10.10.x to 10.11.x, the Dashboard no longer updates the Activity column, and not properly updates the Devices column.
To update it I have to refresh the page or go to another page – such as «Users» – and then go back with the back button updates everything to the actual status.

When people log in or start to play something, they show up. But if they stop playing by closing the page, or the ap on their phone or TV devices, the dashboard says they are playing, but the timestamp doesn't update. Again, refreshing or navigating away and then go bak shows the actual state of things.
If they log out they used to disapear from the Devices column and this was reflected in the Activity solumn, this no longer happens.

I have a browser on a system that does nothing but show the dashboard, as I have 14 users (everyone is close friends of family) and need to keep an eye on things. I do this for my own local jellyfin server, and an external one belonging to a friend. Both have the same behaviour.

I have tested this on two different computers and other browsers, the behaviour is identical.
Both jellyfin servers are installed using the upstream debian repo, so not installed using docker.

### Reproduction Steps

1: Open the Dashboard.
2: Play something from another browser or device.
3: Observe in the Dashboard that the Devices column update the status with the user playing something, but the Activity column does Not.
4: Close the playing session without stopping the playback. 
5: Notice that the user playing is lister as having paused the playback, due to the time stamp not advancing every 2 seconds. And the Activity column does not update regardless of activity.

### Expected/Actual behaviour

The Dashboard should show the actual state of the system. If a user stop playing something, the Activity and Devices column's should keep updating the status without the admin needing to constantly refresh the page.

### Server version

10.11.6

### Web version

10.11.6

### Build version

10.11.6

### Platform

Mageia Linux

### Browser

Firefox 139.0.4 (Flatpak version)

### Additional information

Something changed in 10.11.x that broke this, and apparently no one noticed, or nobody who noticed cared enough to report this.

viown:
If you open your console logs, do you see any "failed to connect to websocket" messages?

[...]

PR #7565, opened 2026-02-06
Add polling fallback to sessions

viown:
**Changes**
Add polling fallback to sessions in case of a failed websocket connection

**Issues**
~~Fixes https://github.com/jellyfin/jellyfin-web/issues/7561~~ Doesn't seem to fix the original issue, but a useful change nonetheless.


thornbill:
Could we rely on the built-in functionality of tanstack query to specify a `refetchInterval` to `useQuery` for this?

viown:
I think I've tried this, but ran into issues with checking websocket connectivity.

PR #7571, opened 2026-02-10
Add configurable VR projection modes to HTML video playback

Enuratique:
**Changes**
This adds initial VR video projection support to the HTML5 video player with an extensible mode system and auto-detection.  
Users can now choose `VR / 3D Mode` in player settings (Off, Auto, SBS/TAB, fisheye variants), and the selection is persisted in app settings.  
Auto mode resolves from `Video3DFormat` metadata and filename/path heuristics.  
PlaybackManager command support was added (`SetVrProjection`) so the mode is integrated with existing player command plumbing.  
Includes unit tests for projection normalization/detection and new `en-us` localization keys for the VR labels.

**Issues**
Follow-up work for immersive WebXR head tracking will be submitted separately to keep this PR focused and reviewable.


Enuratique:
Crap, this should probably not be set to merge into master... Can I get an upstream experimental/vr branch that I can push this corpus of work into rather than try to push to master? I'm trying to keep PRs manageable / reviewable.

PR #7587, opened 2026-02-12
Fix Chromecast disconnect crash due to null player in MediaSessionSubscriber

### LINKED ISSUE #7585: Disconnecting from chromecast while video is playing produces an uncaught runtime error

### Describe the bug

Basically what it says in the title. In the current master branch of jellyfin-web, disconnecting from a cast while a video is playing produces a big backtrace about player being null.

### Reproduction Steps

1. Clone the repository
2. `npm install && npm start`
3. Click the cast icon in the upper right and select a google chromecast device
4. Start playing a video
5. While the video is playing, click the cast icon in the upper right and click "disconnect"
6. Error appears

### Expected/Actual behaviour

I expect the app to disconnect from the chromecast without error, returning to a non-casting state where it is free to cast to other chromecasts. Instead, it produces a big runtime error backtrace across the entire app with three copies of the backtrace.

### Logs

ERROR
player cannot be null
    at PlaybackManager.self.getPlayerState (webpack-internal:///./components/playback/playbackmanager.js:2125:15)
    at MediaSessionSubscriber.onMediaSessionUpdate (webpack-internal:///./apps/stable/features/playback/utils/mediaSessionSubscriber.ts:148:62)
    at MediaSessionSubscriber.onPlayerChange (webpack-internal:///./apps/stable/features/playback/utils/mediaSessionSubscriber.ts:197:14)
    at eval (webpack-internal:///./utils/events.ts:59:22)
    at Array.forEach (<anonymous>)
    at Object.trigger (webpack-internal:///./utils/events.ts:58:14)
    at triggerPlayerChange (webpack-internal:///./components/playback/playbackmanager.js:244:61)
    at setCurrentPlayerInternal (webpack-internal:///./components/playback/playbackmanager.js:1013:7)
    at PlaybackManager.self.setActivePlayer (webpack-internal:///./components/playback/playbackmanager.js:876:9)
    at PlaybackManager.setDefaultPlayerActive (webpack-internal:///./components/playback/playbackmanager.js:3981:12)

### Server version

10.11.6-1

### Web version

a30dd7e0dc8b62a0d48bd2ef90ef562f28ffb918

### Build version

a30dd7e0dc8b62a0d48bd2ef90ef562f28ffb918

### Platform

Linux

### Browser

Chrome Version 144.0.7559.132 (Official Build) (64-bit)

### Additional information

_No response_

dfederm:
When disconnecting from a Chromecast, `setCurrentPlayerInternal(null, null)` is called, which triggers a `playerchange` event. By the time `MediaSessionSubscriber.onPlayerChange()` fires, `this.player` has already been set to null by `PlaybackSubscriber.bindPlayerEvents()`. The original code used a default parameter that eagerly called `getPlayerState(this.player)`, which throws when player is null.

This change guards against a null player by checking `this.player` before calling `getPlayerState()`. When there is no active player, reset the media session instead.

Note: The `?.` optional chaining on `state` in the rest of the function is a TypeScript concession. After the guard, `state` is always defined, but TS can't narrow through reassignment of an optional parameter.

Fixes #7585

### DIFF
--- a/src/apps/stable/features/playback/utils/mediaSessionSubscriber.ts
+++ b/src/apps/stable/features/playback/utils/mediaSessionSubscriber.ts
@@ -96,15 +96,17 @@ class MediaSessionSubscriber extends PlaybackSubscriber {
 
     private onMediaSessionUpdate(
         { type: action }: Event,
-        stateOverride?: PlayerState
+        state?: PlayerState
     ) {
-        if (!this.player) {
-            console.debug('[MediaSessionSubscriber] no active player; resetting media session');
-            return resetMediaSession();
+        if (!state) {
+            if (!this.player) {
+                console.debug('[MediaSessionSubscriber] no active player; resetting media session');
+                return resetMediaSession();
+            }
+            state = this.playbackManager.getPlayerState(this.player);
         }
 
-        const state: PlayerState = stateOverride || this.playbackManager.getPlayerState(this.player);
-        const item = state.NowPlayingItem;
+        const item = state?.NowPlayingItem;
 
         if (!item) {
             console.debug('[MediaSessionSubscriber] no now playing item; resetting media session', state);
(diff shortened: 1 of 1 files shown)


dfederm:
Looks like @thornbill fixed this same issue in #7645 (and somehow had basically the same fix and _exactly_ the same debug log message?)

My approach is still probably cleaner and he current code in master unconditionally bails when `this.player` is null, even if the caller provided valid state. Callers like `onPlayerPlaybackStart(e, state)` and `onPlayerStateChange(e, state)` pass state explicitly and the current code throws it away and resets if the player happens to be null. My approach only falls back to the player when state isn't provided, which is the right semantic: "use what you were given, only derive state when you need to."

dfederm:
Closing as this is already fixed.

PR #7588, opened 2026-02-12
Fix duplicate DOM elements when navigating to same page

### LINKED ISSUE #7570: Navigation to Home from Movie details duplicates #indexPage element
### Description of the bug

When navigating the web interface, specifically returning to the Home screen from a Movie detail page, the `#indexPage` div is duplicated in the DOM.
        
One instance of the div remains visible, while the duplicate has the class `hide`. This results in two elements with `id="indexPage"`, which violates HTML standards (IDs must be unique). This breakage affects custom CSS targeting the ID and can potentially break JavaScript functionality that relies on `getElementById` or similar selectors expecting a single element. It also wastes client resources by retaining the duplicated DOM tree.

### Reproduction steps

1. Open the web client (tested on https://demo.jellyfin.org/stable/web)
2. Log in (using the demo user)
3. Click on any Movie title to enter the item detail view
4. Click the "Home" button (house icon) in the top-left navigation bar
5. Open the browser Developer Tools (F12) and inspect the DOM <body>
6. Observe that there are now two `div` elements with `id="indexPage"`

The `div #indexPage` and its children are duplicated in the DOM. One is hidden, one is visible. Unique IDs are duplicated

The existing `#indexPage` should be reused, or the previous one should be fully removed from the DOM before creating a new one. IDs must remain unique

### Jellyfin Server version

10.11.6

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

N/A

### Environment

```markdown
- OS: Official Demo Server (Linux)
- Clients: Web Client
- Browser: Chrome 144
```

### Jellyfin logs

```shell
N/A
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

Image

### Additional information

_No response_

dfederm:
The view container uses a 3-slot page rotation. When navigating to a page via push navigation (e.g. clicking the Home button), `loadView` creates a fresh page in the next slot without removing any existing cached instance of the same page in another slot. This results in duplicate DOM elements with the same id attribute.

For example: Home (slot 0) → Movie (slot 1) → Home (slot 2) leaves two `#indexPage` divs in the DOM, one visible and one hidden.

This change cleans up any existing cached page with the same URL before creating a new one. It also adds null checks in `beforeAnimate`/`afterAnimate` for cleared slots

Fixes #7570

### DIFF
--- a/src/components/viewContainer.js
+++ b/src/components/viewContainer.js
@@ -38,6 +38,18 @@ export function loadView(options) {
             pageIndex = 0;
         }
 
+        // Remove any previous cached page for the same URL to prevent
+        // duplicate DOM elements (e.g. duplicate #indexPage ids)
+        const mainAnimatedPagesContainer = getMainAnimatedPages();
+        for (let i = 0; i < allPages.length; i++) {
+            if (i !== pageIndex && currentUrls[i] === options.url && allPages[i]) {
+                triggerDestroy(allPages[i]);
+                mainAnimatedPagesContainer?.removeChild(allPages[i]);
+                allPages[i] = null;
+                currentUrls[i] = null;
+            }
+        }
+
         const isPluginpage = options.url.includes('configurationpage');
         const newViewInfo = normalizeNewView(options, isPluginpage);
         const newView = newViewInfo.elem;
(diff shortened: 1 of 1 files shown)


dfederm:
@thornbill Gentle poke for this PR as it's several months old

dfederm:
@thornbill Nudge on this PR

dfederm:
@dmitrylyzo This something you'd be willing to review?

PR #7572, opened 2026-02-10
Add password visibility toggle to login

aaron13100:
**Changes**
- Add a show/hide toggle to the manual login password field.
- Style the toggle for visibility and focus in light/dark themes.
- Add en-us strings for Show/Hide password.

**Issues**
- Refs https://features.jellyfin.org/posts/2924/add-a-show-password-button-on-login-screen

Note: changes written by codex (gpt-5.2-codex)

PR #7593, opened 2026-02-12
Fix token persistence when Remember Me disabled

### LINKED ISSUE #516: Security issue with access token
Hi,

Sorry if the issue does not exist anymore on Jellyfin, I have not (yet) migrated from Emby. I have spotted a security issue with the way access tokens are stored in the local storage.

When the user is connected, Emby stores an access token under a `servercredentials3` key, with a user id. Emby adds also a `enableAutoLogin` key set to true or false, determining if the user should be remembered.

When the user closes the browser, **all user information [access token + user id]** is kept into the browser, even if the user has opted out from the remember me feature.

When the user comes back to Emby, if the `enableAutoLogin` key is set to false, then the user id and the access token are cleared.

This is clearly a security issue. If a user does opt out from the Remember me feature, then the whole local storage should be cleared when the user is disconnected.

Moreover, the local storage contains a whole policy stored under a `user-xxx` key with many boolean values, I would not be surprised if this would allow any user to bypass some security features (I'm puerly speculating here).

If the issue also exists in Jellyfin, I would encourage the local storage usage to be replaced by a proper session, with all authorization procedures being done in the backend.

--

Raw reproduction scenario:

- Access your emby instance at `example.com`
- Enter your credentials and opt out the Remember me option
- Close your browser
- Access your emby instance at `example.com/randomstring` (to get a 404 error)
- Open the dev tools, go in the local storage and replace the value of `enableAutoLogin` to `true`
- Access `example.com`
- You're now connected <-- Unexpected behavior

aaron13100:
**Changes**
- Fixes credential persistence in `ConnectionManager` so an access token is only saved when auto-login is enabled.
- Prevents the #516 bypass where users could toggle `enableAutoLogin` in localStorage and regain a remembered session after logging in with Remember Me disabled.
- Adds unit tests covering both paths: no token persistence when auto-login is disabled, and normal persistence when enabled.

**Issues**
Fixes #516

### DIFF
--- a/src/lib/jellyfin-apiclient/connectionManager.js
+++ b/src/lib/jellyfin-apiclient/connectionManager.js
@@ -3,6 +3,7 @@ import { getAuthorizationHeader } from '@jellyfin/sdk/lib/utils';
 import { MINIMUM_VERSION } from '@jellyfin/sdk/lib/versions';
 import { ApiClient } from 'jellyfin-apiclient';
 
+import appSettings from 'scripts/settings/appSettings';
 import events from 'utils/events';
 import { ajax } from 'utils/fetch';
 import { equalsIgnoreCase } from 'utils/string';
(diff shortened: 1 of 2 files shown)


thornbill:

PR #7594, opened 2026-02-13
Fix dialog box not appearing for chromecast disconnect

### LINKED ISSUE #7591: Shut down player dialog no longer appears for Chromecast disconnect

### Describe the bug

When connected to a chromecast, if you hit the "disconnect" button, there used to be a dialog box asking if you wanted to shut down the video, or just disconnect from the player and leave it running, but that dialog box no longer appears. This is because 'EndSession' was removed as a supported message for the chromecast in 588e9e38f7 (fair enough, the player itself doesn't support that message), but you **can** end a session on a chromecast, you just call the endSession method of the chromecast sdk instead. But, the code that displays this dialog only checks for a supported EndSession method to decide whether to display the dialog or not, so it misses this is possible and doesn't display the dialog.

### Reproduction Steps

1. Clone repo
2. `npm install && npm start`
3. Open the app (localhost:8080)
4. Connect to a server
5. Cast to a chromecast through the UI cast button
6. Play a video
7. Click the UI cast button and then "disconnect"
8. No dialog box appears

### Expected/Actual behaviour

The dialog box should appear asking if I want to shut down the app on the chromecast or just disconnect from it.

### Logs

_No response_

### Server version

10.11.6

### Web version

a30dd7e0dc8b62a0d48bd2ef90ef562f28ffb918

### Build version

a30dd7e0dc8b62a0d48bd2ef90ef562f28ffb918

### Platform

Linux

### Browser

Google Chrome Version 144.0.7559.132 (Official Build) (64-bit)

### Additional information

_No response_

HazardousPeach:
Fix a bug introduced in 588e9e38f7 preventing the chromecast from being detected as supporting ending a session, causing the dialog box about whether you want to end the session or not to not appear.

**Background**

There are two supported ways for Jellyfin to end a session with a PlayerPlugin: sending an "EndSession" command, or calling endSession on the chromecast SDK. The code used to check if the PlayerPlugin listed "EndSession" as a supported command, and if so, called the endSession method. But, the chromecast doesn't actually support "EndSession" as a message, so someone removed it from the list of supported messages, even though its the only PlayerPlugin that has the endSession method that is actually called.

I think this logic also would have failed if a remote session player reported EndSession as a supported command, since the remote session player plugin doesn't have an endSession method.

**Changes**

I added a type annotation into plugin manager that would surface some of the duck-typed weirdness leading to these edge cases. Instead of calling endSession when the "EndSession" **message** is supported, I just call sendMessage with that message. Then, I changed the chromecast plugin to report the EndSession message as supported, but added another case in sendMessage that would call endSession on anything that had it on the EndSession command. This should fix both the chromecast issue and any weirdness with other remote sessions that did support EndSession as a message (not a method).

The type annotation surfaced one other bug, that this.player was typed as PlayerPlugin | undefined, but it was copied from a place that was actually PlayerPlugin | null, so I fixed that there too, hope that's the right thing for this maintainer community.

**Issues**
Fixes #7591

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -103,6 +103,7 @@
 - [TheBosZ](https://github.com/thebosz)
 - [qm3jp](https://github.com/qm3jp)
 - [johnnyg](https://github.com/johnnyg)
+- [Alex Sanchez-Stern](https://github.com/HazardousPeach)
 
 ## Emby Contributors
 
--- a/src/apps/experimental/components/AppToolbar/menus/RemotePlayActiveMenu.tsx
+++ b/src/apps/experimental/components/AppToolbar/menus/RemotePlayActiveMenu.tsx
@@ -67,7 +67,9 @@ const RemotePlayActiveMenu: FC<RemotePlayActiveMenuProps> = ({
                 onMenuClose();
 
                 if (id === 'yes') {
-                    playbackManager.getCurrentPlayer().endSession();
+                    playbackManager.sendCommand(
+                        { Name: 'EndSession' },
+                        playbackManager.getCurrentPlayer());
                 }
                 playbackManager.setDefaultPlayerActive();
             }).catch(() => {
--- a/src/apps/stable/features/playback/utils/playbackSubscriber.ts
+++ b/src/apps/stable/features/playback/utils/playbackSubscriber.ts
@@ -42,7 +42,7 @@ export interface PlaybackSubscriber {
 
 // eslint-disable-next-line @typescript-eslint/no-unsafe-declaration-merging
 export abstract class PlaybackSubscriber {
-    protected player: PlayerPlugin | undefined;
+    protected player: PlayerPlugin | null = null;
 
     private readonly playbackManagerEvents = {
         [PlaybackManagerEvent.PlaybackCancelled]: this.onPlaybackCancelled?.bind(this),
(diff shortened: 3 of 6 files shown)


HazardousPeach:
I think I actually addressed all the issues requested by the bot? Those checks are passing now, not sure why it still says changes are requested.

HazardousPeach:
Agh okay this actually wasn't doing the right thing in testing, sorry for wasting anyones time, I'm going to test this a lot better before un-drafting it.

HazardousPeach:
Okay, **now** this is tested. I squashed the 'oops' commits so the history is clear and its rebased off the latest master. Note that with just this PR, selecting either option will result in the same crash you got before when you hit disconnect, fixed by PR #7587 (but it will still do the disconnecting or session ending that the dialog box is letting you choose between). The two change sets combine cleanly to produce a build which has the box *and* prevents the crash, but they also should be reviewable separately.

HazardousPeach:
Oops just fixed for the classic menu too, sorry about the force pushes, but I don't think anyone had looked in between those.

HazardousPeach:
Hey @nielsvanvelzen @thornbill , just wanted to make sure this PR doesn't get ignored because its marked as "Changes requested"; those changes were requested by the bot on an earlier commit because of linter errors, and all the linter issues have now been fixed, so the request is stale. If someone could dismiss the stale bot review so that the status shows properly for reviewers, it would be much appreciated!

PR #7600, opened 2026-02-14
ux: right align form dialog buttons

brandon1024:
**Changes**
    
Adjust the button placement in the metadata editor and various dialogs for a more visually-appealing experience and improved usability. This revision right aligns the buttons in a flex container, rather than stacking or centering them.

Happy to make any adjustments, just let me know :-)

Chat on Matrix -> https://matrix.to/#/!xrSDQsdjElWFYUAMoG:matrix.org/$LgwuQXenP0CF5fzyXQbuF9trRcacuTRkCcoUrXu_RL0?via=bonifacelabs.ca&via=t2bot.io&via=matrix.org

**Screenshots**

Before:

jellyfin home brandonrichardson ca_web_ (2)
jellyfin home brandonrichardson ca_web_ (4)

After:

jellyfin home brandonrichardson ca_web_ (1)
jellyfin home brandonrichardson ca_web_ (3)

## Disclosure

These changes are completely organic, written by myself, **without** the use of AI. I just want to make this clear, given the current landscape..

### DIFF
--- a/src/components/metadataEditor/metadataEditor.template.html
+++ b/src/components/metadataEditor/metadataEditor.template.html
@@ -266,7 +266,7 @@
                 </div>
             </div>
             <br />
-            <div class="formDialogFooter">
+            <div class="formDialogFooter" style="display: flex; justify-content: flex-end;">
                 <button is="emby-button" type="button" class="raised button-cancel block btnCancel formDialogFooterItem">
                     <span>${ButtonCancel}</span>
                 </button>
--- a/src/components/formdialog.scss
+++ b/src/components/formdialog.scss
@@ -61,13 +61,13 @@
     left: 0;
     right: 0;
     display: flex;
+    justify-content: flex-end;
     position: absolute;
     padding: 1em 1em;
 
     /* Without this emby-checkbox is able to appear on top */
     z-index: 1;
     align-items: flex-end;
-    justify-content: center;
     flex-wrap: wrap;
 }
 


brandon1024:
Perhaps this would be best fixed in a second PR, but there's a ton of padding in the metadata editor that I would like to remove. It seems that this padding is there to help render the popup metadata editor correctly, so fixing this isn't so trivial it seems.

image

Edit: I've managed to sort out the padding. I'll open a separate PR once this is merged.

Chris4942 (approved):
Checked out your branch and tested locally and it looks great in both places! Well done! You'll still need an approval from @thornbill though I think

brandon1024:
Thanks for the review @Chris4942!

geebru (changes requested):
Right aligned buttons can be an accessibility problem for magnification users and don't align with any other actions in Jellyfin.

If they need adjusting, they should be changed to match other button alignments to match the rest of the system and not be a one off design change.

brandon1024:
> Right aligned buttons can be an accessibility problem for magnification users and don't align with any other actions in Jellyfin.

I'm happy to left align or center the buttons if that is better. I opted to right align them here to align with the options at the top of the modal -- see the "..." icon in last image in the PR description.

> If they need adjusting, they should be changed to match other button alignments to match the rest of the system and not be a one off design change.

I agree, consistency is important. What would you propose as an alternative? I believe the current UX is quite poor, so doing something is better than nothing. I couldn't find any other pages where we have buttons organized like this. In the dashboard the "Scan Libraries", "Restart" and "Shutdown" buttons are inline, left-aligned:

image

geebru:
Agreements or disagreements on the visual language of actions in the web-view aside - the current behavior _is_ normal as far as I can tell. I'd raise the question of whether there's an appetite to adjust these at the given moment in the UX channel, and the PR could adjust from there.

Once there's alignment on a direction, if a different one is chosen, I can see some open arms to update actions across applications.

To your credit, some of the dialog components do include right-aligned actions (like Metadata repository management and backups), but they're locked to smaller sizes where the items don't float as far out.

image

brandon1024:
FYI, changes look good on mobile.

localhost_8080_(iPhone 12 Pro)

@geebru I've opened a discussion in the matrix channel as suggested. Hoping to move this review along, I've got some more fixes that I would like to submit once this is merged.

brandon1024:
Following a bit of healthy and positive discussion in Matrix, I've opted to promote this style into the main dialog CSS for consistency across Jellyfin pages. Here's a few screenshots.

localhost_8080_

localhost_8080_ (1)

image

localhost_8080_ (2)

Scanned through Jellyfin and changes look quite nice. Should be ready for review @thornbill.

Additional UI fixes will follow once this is merged.

brandon1024:
This has been open for a few months now awaiting maintainer review. Any chance I could have someone look at this? I have some further fixes to the metadata editor ready as well, but it depends on these changes here.

/cc @thornbill

brandon1024:
Rebased onto the tip of main to ensure changes still look good in 12.0. Everything looks good :+1:

dkanada (approved):

src/components/metadataEditor/metadataEditor.template.html
```diff
@@ -266,7 +266,7 @@ <h2>${HeaderMetadataSettings}</h2>
                 </div>
             </div>
             <br />
-            <div class="formDialogFooter">
+            <div class="formDialogFooter" style="display: flex; justify-content: flex-end;">
```
geebru, line 269: You don't need this inline styling if the class in the .scss is applying. This just duplicates the same CSS but with higher specificity.

brandon1024: Unfortunately we do, because the metadata manager has some logic which manipulates this element and removes the `formDialogFooter` class depending on whether the content is rendered in a modal window or as a standalone page. If you drop the inline style, the buttons will be displayed stacked as it was before.

See further: https://github.com/jellyfin/jellyfin-web/blob/master/src/components/metadataEditor/metadataEditor.js#L1151

geebru: Very unfortunate :( but heard! Thanks for the extra info!

geebru (approved):
I can't seem to "Dismiss" my old review, and while I'm not totally aligned on the execution (in terms of the visual output), I think the consistency is a major win, and it sounds like the core team is onboard as well. Thanks!

brandon1024:
> I can't seem to "Dismiss" my old review, and while I'm not totally aligned on the execution (in terms of the visual output), I think the consistency is a major win, and it sounds like the core team is onboard as well. Thanks!

Thanks @geebru :-) I appreciate your comment.

PR #7601, opened 2026-02-14
Add "Remove from Continue" context menu action for in progress items

aaron13100:
This is for feature https://features.jellyfin.org/posts/517/add-an-option-to-remove-an-item-from-continue-watching and takes https://github.com/jellyfin/jellyfin-meta/discussions/50 into account.

https://github.com/jellyfin/jellyfin-meta/discussions/50 lays out two possible approaches for "remove from continue" behavior.

A: Hide without losing resume position
Store a per user flag and adjust Continue queries to exclude flagged items while preserving resume progress. (cleaner end state, server side data and query changes, concerns about added complexity).

B: Clear resume progress to fall out of Continue
(client side only, easy)

This PR implements Option B only so this is not "hide while keeping resume." It's "clear resume so it falls out of Continue."

Implementation details:
- Adds a new item context menu action shown for in-progress items (has UserData.PlaybackPositionTicks) that are not played and not local items.
- The action is shown after Refresh metadata in the context menu.
- Selecting it calls the existing Playstate API markUnplayed to clear resume progress so it falls out of the relevant Continue row.
- Labels are Clear watch progress / Clear listening progress / Clear reading progress depending on media type (video/audio/books).

If you throw this away it's fine. It's only about 36 lines of code.

As a user I personally think option A is better, but that this might be better than nothing currently. I do think that a label change is kind of important though. As far as I understand it, many major services use option A where the person does not lose their progress even though they remove the item from the continue watching list.

Netflix: option A
Disney+: option A
Apple TV: option A
Hulu: option B
Amazon: dunno

So I think that if you accept this PR, it must be clear for the user that it's not doing the same thing as those other services, which is what they might expect. Here are some of the label options. I went with number one.

1. Clear watch progress / Clear listening progress / Clear reading progress
2. Reset watch progress / Reset listening progress / Reset reading progress
3. "Remove from Continue (clear progress)" (too long I guess)
4. "Remove (clear watch progress)" (is "Remove" ambiguous in this case?)
5. "Clear progress (remove from list)"

In general, I think that you could accept this PR without closing the feature request because I think that this is probably not what most people want and it's just kind of a band-aid on the situation. Or we could look at this as a separate feature which is similar to option A, but doesn't necessarily have to close the feature request.

Testing:

- npm run lint
- npm run stylelint
- npm run build:check
- npm test
- npm run build:es-check


src/components/itemContextMenu.js
```diff
@@ -42,6 +43,18 @@ function getDeleteLabel(type) {
     }
 }
 
+function getRemoveFromContinueLabel(item) {
```
Chris4942, line 52: Please add a [JSDoc](https://jsdoc.app/) with type annotations for this. Specifically, what is the type of `item` and what type does this return? As a reader, it's super confusing to not know what `item` is

Chris4942 (approved):
Works as expected! just add some JSDoc type annotations! You'll still need an approval from a maintainer though

src/components/itemContextMenu.js
aaron13100: Claude Opus 4.6: Added @param {BaseItemDto} and @returns {string} in c0558ab.

src/components/itemContextMenu.js
```diff
@@ -42,6 +43,22 @@ function getDeleteLabel(type) {
     }
 }
 
+/**
+ * @param {import('@jellyfin/sdk/lib/generated-client/models/base-item-dto').BaseItemDto} item
```
Chris4942: Don't do it this way. Instead, it's better to define a typedef at the top so that it's not such a long definition and it can be reused easily

```js
/**
* @typedef {import(...).BaseItemDto} BaseItemDto
*/
```
Then you can reuse it below

aaron13100: done

thornbill:

PR #7604, opened 2026-02-14
Fix Up Next dialog after seeking near end

### LINKED ISSUE #3673: "Show next video info during playback" depends on "Play next episode automatically"
I like getting prompted to play the next episode so I don't need to go back to the episode list and select the next one. 
However, this setting seems to be dependent on "Play next episode automatically" being enabled. 
These options do not have any indication that they should be dependent on one another in the settings UI, so opening a bug here.  

**Steps To Reproduce**
Disable "Play next episode automatically"
Enable "Show next video info during playback"
Try to play an episode, no pop up prompt will happen in the last 30 seconds of the video. 
Try to "play all from here", no pop up prompt will happen in the last 30 seconds of the video. 

**Expected Behavior**
The player should pop up to ask if I want to play the next episode without a countdown timer to play the next one automatically unless there is nothing to play after the current episode. 

Server: Ubuntu Linux 22.04
Client: Windows 11
Chrome: 101.0.4951.67
Jellyfin 10.8.0

aaron13100:
**Changes**
Fix an issue where the Up Next dialog may never appear after seeking near the end of an episode when “Show next video info” is enabled.

If `playbackManager.nextItem(player)` fails (e.g. play queue not ready yet), we now avoid setting `comingUpNextDisplayed` so subsequent `timeupdate` events can retry.

**Issues**
Fixes #3673

**Testing**
- npm run lint
- npm run stylelint
- npm run build:check
- npm test
- npm run build:es-check


src/controllers/playback/video/index.js
```diff
@@ -686,15 +686,19 @@ export default function (view) {
...
                     currentUpNextDialog = new UpNextDialog({
                         parent: view.querySelector('.upNextContainer'),
                         player: player,
                         nextItem: nextItem
                     });
                     Events.on(currentUpNextDialog, 'hide', onUpNextHidden);
-                }, onUpNextHidden);
+                }, function () {
+                    // If the play queue isn't ready yet, allow a later retry on subsequent timeupdate events.
+                    comingUpNextDisplayed = false;
```
Chris4942: Is it necessary to also set this to `false`? I think we'd just need to the one for success

Chris4942 (commented):
See if we can simplify this a bit, otherwise, this looks fine to me

Chris4942:
Tested on preview by skipping to the end of a few episodes of Pioneer and making the sure the up next pop up still shoes up. It seems to work as expected

src/controllers/playback/video/index.js
aaron13100: From Claude Opus 4.6: You're right - since comingUpNextDisplayed = true only happens inside the success callback now, it's never true when the rejection handler runs. Removed the redundant reset and went back to passing onUpNextHidden directly. See c388022.

thornbill:

PR #7608, opened 2026-02-16
Add Type Annotations To Login and Logout Code Paths

Chris4942:
### Changes

- add type annotations to the functions involved in login and logout code paths

### Testing

Log in and out several times to multiple accounts through both methods and make sure nothing broke


OmegaGiven:
After looking this over i went ahead and tested locally login works nothing changed as no functional changes are included by my look over

as far as annotations go just a few pointers but really unsubstatial so if you really want to touch up here are my thoughts to really get annotations to near perfect (is there truely ever anything that is perfect?)

login/index.js:50
number rather than int

boolean rather than bool if we want to follow type scripting:
connectionManager.js:195
connectionManager.js:197
connectionManager.js:233

dialogHelper.js:50 
-remove the extra scrollY
-removeOnClose is listed 2 times one as required and other time as optional.

dialog.js:17
include id and name, both used in dialog.js:106

Chris4942:
@OmegaGiven Thank you for the review! I updated those fields as requested!

Issue #6679, opened 2025-03-29
Unable to zoom in when reading books


redstonedesigner:
### Description of the bug

I'm a user with accessibility needs.  One of these needs is the ability to zoom in when reading books, as I struggle to read standard font size(s).

When attempting to read a book in Jellyfin, it is not possible to zoom in.  Even browser zoom is ignored.  This occurs on at least the web and iPadOS versions of the app.

### Reproduction steps

1. Sign into Jellyfin on web or iPadOS
2. Start reading a book
3. Attempt to zoom in

### What is the current _bug_ behavior?

I am unable to zoom in when reading a book, both natively and within the browser.

### What is the expected _correct_ behavior?

Zooming is respected, scroll bars being added as necessary.

### Jellyfin Server version

10.10.0+

### Specify the build version

10.10.6

### Environment

```markdown
- OS: Ubuntu Server 24.04
- Virtualization: Docker
- Clients: Browser, iPadOS
- Browser: Microsoft Edge
- Hardware Acceleration: No
- GPU Model: N/A
- Reverse Proxy: NGINX Proxy Manager
- Base URL: none
- Storage: local
```

### Jellyfin logs

```shell
No relevant log output - issue is client-side
```

### FFmpeg logs

```shell

```

thornbill:
What file format are the affected books?

redstonedesigner:
Both PDF and EPUB file formats are affected, though in different ways.

When viewing a PDF, I only get the option to close (and no zoom options at all are available or respected).
When viewing an EPUB, I have the option to increase the font size however it's not taking any action.

OmegaGiven:
I have a fix for this as i wanted a feature like this for comics this will only be applied to pdfs though and not epub format. i may try fixing that in a differnt PR but ill try commiting what i got. and see if it works. i have confirmed its functionality is 100X better than what the pdf reader is currently.

FatherDivine:
> I have a fix for this as i wanted a feature like this for comics this will only be applied to pdfs though and not epub format. i may try fixing that in a differnt PR but ill try commiting what i got. and see if it works. i have confirmed its functionality is 100X better than what the pdf reader is currently.

Any update?

Tag me on anything, I'll look at the PDF and see if an epub fix can be made easily. I have over 1300 books and mostly epub.

gdeddi:
really looking forward to seeing something like this implemented, as i just downloaded manga and can hardly read it bc its zoomed out and i cant zoom in

[...]

PR #7609, opened 2026-02-16
Adding header and zoom accessability pill to pdf reader

OmegaGiven:
**Changes**
added header to the top of pdf reader with title/issue/chapter of the pdf with a page counter so you can know what page you are on.

added an accessability pill for zooming/going to next/previous page, closing out, resize zoom see screeshot for what that looks like. that pill will be on the bottom of the page for mobile and the top will have the page below the title for mobile

mobile features:
- tripple tap to hide/show header
- pinch zooming now supported
- swipe gestures maintained for going to next page
- able to move around while zoomed and scroll for browser works now with that.

exposed the touch gesture hardcoded values to css for people to be able to adjust for their own web clients if they want. May feel different for the gestures depending on resolution of screen. currently on my s24 and 1440p monitor it feels fine and usable. 10x better than what the pdf reader was

**Issues**
(https://github.com/jellyfin/jellyfin-web/issues/6679)

BEFORE
brave_screenshot_localhost (2)

AFTER
Screenshot 2026-03-20 at 11 41 15 AM


src/plugins/pdfPlayer/plugin.js
```diff
@@ -92,6 +124,135 @@ export class PdfPlayer {
         return true;
     }
 
+    getTitle() {
```
Chris4942, line 174: With new functions like this, please include type annotations so that es lint can help keep us honest and it's easier for other people to trace what you're doing. I think this is returning a `string`, but it's not totally obvious from just looking at it.
```suggestion
    /**
     * @returns {string}
     * 
     */
    getTitle() {
```

src/plugins/pdfPlayer/plugin.js
```diff
@@ -92,6 +124,135 @@ export class PdfPlayer {
...
+        return this.item?.SeriesName || this.item?.Album || this.item?.Name || '';
+    }
+
+    getChapter() {
```
Chris4942, line 181: Same thing here, I think this returns a `string`, but it would be great if it was annotated so I could be sure es lint had ensured it

Chris4942 (commented):
I didn't even know that jellyfin had a PDF viewer. It would be good to get some added to the demo account so that we could test against those there, but I'm not sure how the demo server. I'll try building your branch locally so I can test there

OmegaGiven:
thanks for the suggestion, i didnt know what kind of comments this community liked. i will try to add some commenting when i have time tomorrow or the day after. thanks for looking at this, you are the true hero!

src/plugins/pdfPlayer/plugin.js
OmegaGiven: thanks added this with additional type annotations in commit

src/plugins/pdfPlayer/plugin.js
OmegaGiven: thanks added this with additional type annotations in commit

Chris4942:
Testing Note

- When I zoom in, the page stops being centered and is instead left aligned. If I go back to 100% zoom, then it's centered again. This is a bit jarring. It should probably just stay centered.

Steps:
- OS: Linux (NixOS), DE: Hyprland
- Browser: Brave
- [Test File Used](https://archive.org/details/crime-must-pay-ace-46)

1. I added the test file to
```
/mnt/main/Books/Comics/Crime Must Pay The Penalty! (1951)/Comic #001 (1955).pdf
```
2. I added `/mnt/main/Books` as a jellyfin Books library in the admin dashboard
3. I built the project using `npm start`
4. Accessed it through `localhost:8080` in my browser

---

I'm still working on getting the app on my phone to test the swipe features

OmegaGiven:
Yea im aware of that. It has to do with the viewport only being as big as the x direction so when zooming out it wont out zoom in a direction that is locked. It may require a lot of wizardry to fix so I decided to just get this out since it's ner 3x better than what exists and maybe touch up later as its a low demand thing.

OmegaGiven:
fixed the merge conflict: just the contributors page adding names (accepted both)

dkanada:
All of the features look great but it would probably be better to clone the existing video OSD rather than use a new top header and pill shape.

The interface itself should also be moved into a new class - the PDF controls can be used without changes in the comic viewer if they're separate from the plugin. The EPUB player would need a settings button added for text style but it's nearly a drop-in replacement there as well.

OmegaGiven:
@dkanada Ill take a look and test if the exsiting OSD functionality can be copied or used but i think these may use entirely different libraries and methods to display in the viewport so probably not viable.

On the note of the EPUB reader, that is handled differently and actually has features in place that don't cover the scope of this PR because these are different view formats as EPUB is text based (relatively) and PDF is image based (for the most part) so that would be out of scope and handled differently.

otherwise let me do some additional testing and scraping of the other format plugins and see if i can make this look "Unified" but im still just pushing to hopefully get this in as at its current state PDF reader is essentially in a unusable state.

OmegaGiven:
(sorry if this feels like a book lol)

@dkanada as a followup for your suggestions i looked into how bookplayer and comicplayer do their page rendering and headers and just the file handling will prevent us from making a more unified method of handling the pages.
eg: comicplayer uses sway to handle the crb files types which are image based but how pdfs render would really hinder how its handled and cause lots of slowdowns in handling resulting in a bad user experience. 

what my logic does is inherently a proprietary setup with no libraries but handles the pdf zoom in the view port to keep the resolution as is as pdfs are beefier so its easier to handle.

i did go ahead and unify the pill if that was what was subtly suggesting and i liked that.

I combined the header and the pill so this makes it so we use less screen real estate which is needed on mobile but keeps the title and page number which maybe we add to the other players in a different PR but that is out of scope of this PR. the scope of this pr is to add accessibility and functionality to just the pdf player and touch nothing else.

if you want to test this here is some quick commands you can use to pull and boot up if you got the dev tools already and use your existing jellyfin backend:

git clone https://github.com/jellyfin/jellyfin-web.git
gh pr checkout 7609
npm install # gets dependencies
npm run serve -- --port 8097

For host make sure to use YOUR-IP:8086 which would be your existing backend port so you can test on existing files in your jellyfin server

dkanada:
I think I've explained my thoughts poorly so I'll list them separately for clarity. I agree that only the PDF interface is in scope for this pull request but it should look like a Jellyfin UI component and use React under the hood.

**Interface**

My suggestion was to more closely match the existing OSD interface so everything looks unified. I'll attach a screenshot below as a loose example - elements like the favorite button aren't in scope but were added for future reference.

**Code**

We are migrating to React so there's no reason to implement this in the existing layout. I was recommending a new React component called `BookOsd` that would layer the controls (and swipe logic) over the book view. Only the PDF plugin needs changes right now, but someone can go through later and mount the same component for EPUBs and comics without code duplication. See #7656 for an example on how to mount components in older layouts.

example

OmegaGiven:
@dkanada i see what you mean, look at the screenshots of the new look i updated my screenshots so you can see i unified it like you suggested so this should be fine for now.

@FatherDivine 
[FatherDivine](https://github.com/FatherDivine)
Just tagging you as you seemed invested in the different book formatting plugins. would love your take. this is ready to ship whenever there is a code merge but always open for more eyes

FatherDivine:
Ah just noticed. You are right, it's just pdf with issues. I was exporting mass documents as epub but it was defaulting to PDF, which is the one with issues. ePub "could" be faster in opening large books (the 800+ page kinds) but that's for another day.

Thanks and hope they merge soon! if not I'll compile from source for this, as it's worth it. @OmegaGiven

OmegaGiven:
I love jellyfin but ill be using another homelab tool for PDF comics specifically as any good changes I can PR up probably wont merge before they get conflicts with other peoples changes, hope others can have more patience than me for fixing issues and old deprecated functions.

PR #7615, opened 2026-02-19
Add toggle for truncating titles in list view

harrisonmg:
Media titles or filenames that lack [normal word break points](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/overflow-wrap#normal) are hidden with ellipsis. In the "Media Info" view, these titles are allowed to overflow. Long titles become impossible to view fully, particularly on platforms where mouse hover tooltips are unavailable.

**Changes**
- Add an option for the List view that allows the user to disable truncating long titles. Defaults to checked, which is the existing behavior.
  Screenshot from 2026-02-19 14-11-15

  Checked (existing behavior):
  Screenshot from 2026-02-19 14-11-30

  Unchecked:
  Screenshot from 2026-02-19 14-11-42

- Tweak Media Info CSS to allow wrapping.

  Before:
  Screenshot from 2026-02-19 12-07-18

  After:
  Screenshot from 2026-02-19 12-07-00
  
**Alternatives**
The setting complexity could be avoided by simply changing the CSS and making this the default behavior moving forward.

### DIFF
--- a/src/components/listview/listview.js
+++ b/src/components/listview/listview.js
@@ -130,7 +130,7 @@ function getChannelImageUrl(item, size) {
     }
 }
 
-function getTextLinesHtml(textlines, isLargeStyle) {
+function getTextLinesHtml(textlines, isLargeStyle, truncate) {
     let html = '';
 
     const largeTitleTagName = layoutManager.tv ? 'h2' : 'div';
@@ -154,6 +154,10 @@ function getTextLinesHtml(textlines, isLargeStyle) {
         }
 
         elem.classList.add('listItemBodyText');
+
+        if (truncate) {
+            elem.classList.add('listItemBodyText-truncate');
+        }
 
         elem.innerHTML = '<bdi>' + escapeHtml(text) + '</bdi>';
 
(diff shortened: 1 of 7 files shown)


src/components/listview/listview.js
```diff
@@ -130,7 +130,7 @@ function getChannelImageUrl(item, size) {
     }
 }
 
-function getTextLinesHtml(textlines, isLargeStyle) {
+function getTextLinesHtml(textlines, isLargeStyle, truncate) {
```
Chris4942, line 133: It would be great to add type annotations to new parameters, so if this is a `boolean` then this.

```suggestion
/**
 * @param {boolean} truncate
 */
function getTextLinesHtml(textlines, isLargeStyle, truncate) {
```

If it's nullable, then you'd use `[truncate]` instead of `[truncate]`

src/controllers/list.js
Chris4942, line 1214: Similarly, here, since you're changing the return type, it would be great to add a type annotation for the parameter you're adding, e.g.,

```suggestion
    /**
     * @return {{ truncateLongTitles: boolean }}
     */
    getViewSettings() {
```

Chris4942 (commented):
I'm not a maintainer, I think that what you've defined as an optional behavior should be the default behavior. I don't think there's a situation where truncating would be preferred to wrapping without a space. As you mentioned, that would be a lot simpler and be less code to maintain

harrisonmg:
I appreciate the feedback from Chris, but without any interaction from maintainers I must assume this isn't getting merged.

PR #7475, opened 2026-01-03
Fix library menu button disappearing when name is too long

FelixRoediger:
Fixes the titles "text-overflow: ellipsis;" property by setting parents overflow property to "hidden".

Currently the button to manage a library in the dashboard would be pushed into the overflow if the library name was too long. The text element has the text-overflow property set to "ellipsis" but this doesn't do anything unless the parent element has its overflow property set to "hidden".

Fixes #7464

PR #7627, opened 2026-02-25
Custom aspect ratio to zoom / scale videos

ckcr4lyf:
**Changes**
This adds a new "custom" option to the aspect ratio selector, allowing zoom (like in Kodi)

The motivation behind this is I have an ultrawide OLED, and if I have movies with black bars, the space gets wasted since my monitor is wide enough for what _should_ have been the original aspect ratio. 

Using this I can scale it a custom amount to get rid of the "fake" black bars.

**Issues**
N/A

**Code assistance**
The initial codegen for the classes etc was from cursor after prompting it with the requirement of:
* "Custom" aspect ratio using `transform: scale()` CSS
* Told it to copy the subtitle sync style slider

After that I tweaked it a bit.

**Caveats**
* The PR isn't fully complete, I need to figure out the "state" behind when to show the slider etc. a bit more, since it is not exactly the same as subtitle sync
* This does not work with "native" subtitles; in that case the subtitles scale with the video and disappear off the screen. With "custom" they remain in place.

I just kinda wanted to get the ball rolling, see if there is any immediate feedback, before I continue

**Alternatives Considered**
My initial idea was just a JS prompt that asks the user to enter the scale amount (I remember it for most aspect ratios on my particular monitor), but the slider is probably nicer?

** Demo **

https://github.com/user-attachments/assets/e802ab48-879a-45c7-80bc-e9fb622926f3


ckcr4lyf:
I branched off of `10.11.6` tag coz master seemed broken, btw. (e.g subtitleSync was not working on master); should I try and rebase? The last four commits at the end are mine.

viown:
Since you're adding a new feature, you'll want to target master. You can find instructions to do this [here](https://github.com/jellyfin/jellyfin-web/pull/7475#issuecomment-3799163752) however you'll need to swap the places of `release-10.11.z` and `master` in the rebase command, so instead you'd do: `git rebase --onto master release-10.11.z` which should correctly place your commits on top of master.

ckcr4lyf:
Thanks for the feedback @viown . Since the subtitle sync (and therefore this slider which is inspired from there) is broken on master, I hacked together some `scss` to at least make it display. Not sure if someone is already looking into fixing this.

Anyway, I've figured out the state stuff depending on how/when the "Custom" dialog is opened, should be fine now. There are still some random JS `console.log()` around in case I need to debug further, but I can clean those up.

Let me know what you guys think, if this is too big, or anything I can do to make your guys job easier. Cheers!

ckcr4lyf:
!unstale

PR #7631, opened 2026-02-26
Order MediaMetadata.artwork from highest resolution to smallest

### LINKED ISSUE #7630: Fetching album artwork via MPRIS->Firefox->Jellyfin uses a low resolution image

### Describe the bug

When playing media in Jellyfin on Firefox on Linux, the artwork that Firefox provides via the MPRIS interface is a low resolution image. For example, when displayed in noctalia-shell:

Image

### Reproduction Steps

Open Jellyfin in Firefox on Linux. Choose a track with album artwork and play it.
While the player is running, use an MPRIS tool like `playerctl` to get the artwork metadata:

```
$ playerctl metadata mpris:artUrl
```

This should return a URL like `file:///home/goose/.mozilla/firefox/firefox-mpris/23163_59.png`.

### Expected/Actual behaviour

Full-resolution artwork is expected from the image returned by MPRIS. However, inspecting the file shows its resolution is 96x96.

### Logs

_No response_

### Server version

10.11.4

### Web version

10.11.4

### Build version

10.11.4

### Platform

Linux

### Browser

Firefox 147.0.4

### Additional information

It's suspected that Firefox uses the first image in the `MediaMetadata.artwork` array for MPRIS.
Right now, when Jellyfin creates that array, the images are ordered from smallest resolution (96x96) to highest (512x512).

agausmann:
**Changes**

Reverse the order of `DEFAULT_IMAGE_SIZES`, which makes `getArtwork` put the highest available resolution image first in the array of images. 

**Issues**

Fixes #7630. When I tested this, Firefox now chooses the highest resolution image to provide to MPRIS.

**Code assistance**

None

### DIFF
--- a/src/apps/stable/features/playback/utils/mediaSessionSubscriber.ts
+++ b/src/apps/stable/features/playback/utils/mediaSessionSubscriber.ts
@@ -11,8 +11,12 @@ import type { ItemDto } from 'types/base/models/item-dto';
 import type { PlayerState } from 'types/playbackStopInfo';
 import type { Event } from 'utils/events';
 
-/** The default image resolutions to provide to the media session */
-const DEFAULT_IMAGE_SIZES = [96, 128, 192, 256, 384, 512];
+/** The default image resolutions to provide to the media session.
+ * 
+ * Highest-to-lowest order matters; Firefox on Linux seems to use the first
+ * image in the artwork array for its MPRIS interface. (#7630)
+ */
+const DEFAULT_IMAGE_SIZES = [512, 384, 256, 192, 128, 96];
 
 const hasNavigatorSession = 'mediaSession' in navigator;
 const hasNativeShell = !!window.NativeShell;


viown:
This looks like it can go in a patch release. Could you rebase your PR on top of the `release-10.11.z` branch? The steps to do so are [outlined here](https://github.com/jellyfin/jellyfin-web/pull/7475#issuecomment-3799163752).

agausmann:
> Could you rebase your PR on top of the `release-10.11.z` branch? 

Should be good now

viown (approved):

thornbill:
It would be good to include a comment that the order matters and why. 🙃

agausmann:
> It would be good to include a comment that the order matters and why. 🙃

I agree, thanks for the feedback. I also spent some time better understanding this behavior, looking for a stronger justification for this change.

I read through parts of the [Media Session spec](https://w3c.github.io/mediasession/#mediametadata-artwork-images). The behavior that I observed (first item in the artwork array) is not specified there. In the [fetch image algorithm](https://w3c.github.io/mediasession/#update-metadata-algorithm) (run as part of the artwork setter), it refers to selecting a "preferred artwork image" but does not specify any order of preference for the user agent.

After that, I also tested in Chromium. It always scales its MPRIS image to 150px, and I can't really tell any difference in the scaled image's quality from this change. So it is pretty clearly a browser-specific behavior.

PR #7637, opened 2026-02-28
Add webOS 24/25 version detection and allow DoVi MKV on webOS 25

NoFear0411:
webOS 25 (LG 2025 TVs) natively supports Dolby Vision playback from MKV containers. Without proper version detection, these TVs report as webOS 23, and the device profile blocks DoVi in MKV — forcing HLS remux that drops DV metadata to HDR10 and causes audio issues with DTS-HD MA in mpegts segments.

- Add version detection for webOS 24 (Chromium 108+) and webOS 25 (Chromium 120+)
- Allow `mkv` as a valid container for DoVi HEVC direct play on webOS 25+
- Older webOS versions keep the existing mp4/ts restriction

Tested on LG C4 (webOS 25) with DoVi Profile 8.1 MKV.

Addresses jellyfin/jellyfin#16092, jellyfin/jellyfin#16179

### DIFF
--- a/src/scripts/browser.js
+++ b/src/scripts/browser.js
@@ -115,7 +115,11 @@ function web0sVersion(browser) {
 
         // The next is only valid for the app
 
-        if (browser.versionMajor >= 94) {
+        if (browser.versionMajor >= 120) {
+            return 25;
+        } else if (browser.versionMajor >= 108) {
+            return 24;
+        } else if (browser.versionMajor >= 94) {
             return 23;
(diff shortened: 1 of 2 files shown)


dmitrylyzo:
There's another PR with the same changes: https://github.com/jellyfin/jellyfin-web/pull/7328

NoFear0411:
My bad, I realized when I open it that it is a duplicate. Closing.

PR #7528, opened 2026-01-26
Use display missing episodes setting in search

lmaotrigine:
This PR re-implements the changes in #5661.

**Changes**
Respects the `DisplayMissingEpisodes` user setting while displaying episodes in search results.

**Issues**
Fixes #5594

PR #7640, opened 2026-03-01
Respect DisplayMissingEpisodes setting in item details pages

lmaotrigine:
Follows up on PR #7528

This PR does not display missing episodes in item details pages unless the user has `DisplayMissingEpisodes` enabled.

I noticed this happening in Person details pages, and I'm not sure if anything else is affected. I was initially wary of attempting this because the `scripts` directory had an ominous here-be-dragons warning, but this file does not seem to be referenced anywhere else.

I haven't based this on master because this is what I was told to do in the previous PR.

**Changes**
The query parameters for child items in `/details` pages include `isMissing` based the user's preferences. Not sure if this is ideal but from initial testing it seems to work, since I think
this parameter is just ignored for non-episode items..? Unless there is an edge case not present in my own library.

I did consider just passing around the relevant boolean instead of the user object but I see that other such usage in the library passes around the `user` itself, so I have done the same.

**Issues**
None reported

**Code assistance**
no.

### DIFF
--- a/src/scripts/itemsByName.js
+++ b/src/scripts/itemsByName.js
@@ -381,7 +381,7 @@ function getQuery(options, item, user) {
         Limit: 100,
         StartIndex: 0,
         CollapseBoxSetItems: false,
-        isMissing: !user.Configuration.DisplayMissingEpisodes ? false : undefined,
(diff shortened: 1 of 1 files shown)


lmaotrigine:
sorry, messed this up.

PR #7645, opened 2026-03-02
Fix crashes on playback exit

thornbill:
**Changes**
Fixes some unhandled exceptions when exiting the video player

**Issues**
N/A

**Code assistance**
VS Code autocomplete

### DIFF
--- a/src/apps/stable/features/playback/utils/mediaSessionSubscriber.ts
+++ b/src/apps/stable/features/playback/utils/mediaSessionSubscriber.ts
@@ -104,8 +104,8 @@ class MediaSessionSubscriber extends PlaybackSubscriber {
         }
 
         const state: PlayerState = stateOverride || this.playbackManager.getPlayerState(this.player);
-
         const item = state.NowPlayingItem;
+
         if (!item) {
             console.debug('[MediaSessionSubscriber] no now playing item; resetting media session', state);
             return resetMediaSession();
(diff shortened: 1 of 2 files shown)


dmitrylyzo:
What about this one? https://github.com/jellyfin/jellyfin-web/blob/8ea720bf0ea5ece5650331672bfc6171e8527423/src/plugins/htmlVideoPlayer/plugin.js#L1481

thornbill:
I didn't observe a crash there, but shouldn't hurt to add the check there also.

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1479,6 +1482,9 @@ export class HtmlVideoPlayer {
 
         // download the track json
         this.fetchSubtitles(track, item).then(function (data) {
```
dmitrylyzo: To preserve `this`.
```suggestion
        this.fetchSubtitles(track, item).then((data) => {
```

src/apps/stable/features/playback/utils/mediaSessionSubscriber.ts
```diff
@@ -96,10 +96,16 @@ class MediaSessionSubscriber extends PlaybackSubscriber {
...
     ) {
-        const item = state.NowPlayingItem;
+        if (!this.player) {
+            console.debug('[MediaSessionSubscriber] no active player; resetting media session');
+            return resetMediaSession();
+        }
 
+        const state: PlayerState = stateOverride || this.playbackManager.getPlayerState(this.player);
+
+        const item = state.NowPlayingItem;
```
dmitrylyzo, line 107: If it is not against the code style, you can add a new line before the `if` (_as it was_) to make this line unchanged.

```suggestion
        const item = state.NowPlayingItem;

```

PR #7656, opened 2026-03-06
improve author and book support on itemDetails page

dkanada:
**Changes**

1. Show authors on book details page and vice versa
2. Migrate the relevant code to a React component

Not super familiar with React so feel free to suggest any changes. The logic felt bespoke enough to require a unique component so I went that route.

**Issues**

None

### DIFF
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -992,9 +992,9 @@ function renderDetails(page, instance, item, apiClient, context) {
 
         for (const type of metadataTypes) {
             const renderTarget = document.createElement('div');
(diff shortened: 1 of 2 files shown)


src/controllers/itemDetails/index.js
```diff
@@ -2099,6 +2016,8 @@ export default function (view, params) {
     function init() {
         const apiClient = getApiClient();
 
+        self._unmount = [];
```
thornbill, line 2019: I think we should be able to use an actual private property:

```suggestion
        self.#unmountCallbacks = [];
```

dkanada: went with the underscore because this isn't a class so private properties won't build

PR #7659, opened 2026-03-06
Add Safari DoVi detection fallback for profiles 5 and 8

NoFear0411:
## Summary

Safari's `canPlayType()` does not report Dolby Vision support — even on devices with hardware DoVi decoding (VideoToolbox). Apple treats DoVi as a managed HLS capability: `AVPlayer` handles it when the master playlist signals `dvh1`, but `canPlayType` tests bare MP4 playback which is a separate code path.

This means devices like the iPad A16 (2025) that fully support DoVi never get to play it because the browser API returns empty, Jellyfin builds a profile without DoVi, and the server transcodes unnecessarily.

### What this does

Adds inline Safari fallbacks for P5 and P8 detection in `supportedDolbyVisionProfilesHevc()`, matching the existing webOS pattern (`browser.web0sVersion >= 4`).

**Conditions:** Safari on iOS 13+ or macOS.

### Safety

Devices without DoVi hardware are not broken by this change. The HLS master playlist fallback chain from jellyfin/jellyfin#16362 lists both a `dvh1` and an `hvc1` PQ variant pointing to the same stream URL. When a device cannot decode DoVi, the HLS player selects the `hvc1` variant and decodes the base HEVC layer with PQ tone mapping — better than a forced SDR transcode.

### Related

- jellyfin/jellyfin#16362 — Adds spec-compliant `dvh1` variant + `hvc1` PQ fallback in HLS master playlist
- jellyfin/jellyfin#16179 — HLS CODECS signaling discussion

## Test plan

- [ ] Verify Safari on macOS reports DoVi P5 and P8 in device profile
- [ ] Verify Safari on iOS 13+ reports DoVi P5 and P8 in device profile
- [ ] Verify Chrome/Firefox behavior is unchanged (no Safari fallback applied)
- [ ] Verify webOS detection still works independently
- [ ] Play DoVi P5 content on Safari — confirm dvh1 variant selected, no transcode

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -273,9 +273,16 @@ function supportedDolbyVisionProfilesHevc(videoTestElement) {
     const supportedProfiles = [];
     // Profiles 5/8 4k@24fps
     if (videoTestElement.canPlayType) {
-        if (videoTestElement
-            .canPlayType('video/mp4; codecs="dvh1.05.06"')
-            .replace(/no/, '')) {
+        if (
+            videoTestElement
+                .canPlayType('video/mp4; codecs="dvh1.05.06"')
+                .replace(/no/, '')
+            // Safari does not expose DoVi via canPlayType even on capable devices.
+            // AVPlayer handles DoVi when the HLS manifest signals dvh1; canPlayType
+            // only tests bare MP4 which Apple treats as a separate path.
+            // Devices without DoVi fall back to the hvc1 PQ variant in the manifest.
+            || (browser.safari && ((browser.iOS && browser.iOSVersion >= 13) || browser.osx))
+        ) {
             supportedProfiles.push(5);
         }
         if (
@@ -284,10 +291,13 @@ function supportedDolbyVisionProfilesHevc(videoTestElement) {
                 .replace(/no/, '')
             // LG TVs from at least 2020 onwards should support profile 8, but they don't report it.
             || (browser.web0sVersion >= 4)
+            // Safari: same canPlayType limitation as P5. P8 also has an HDR10 base layer as fallback.
+            || (browser.safari && ((browser.iOS && browser.iOSVersion >= 13) || browser.osx))
         ) {
             supportedProfiles.push(8);
         }
     }
+
     return supportedProfiles;
 }
 


NoFear0411:
Analysis of the canPlayType gap and safety implications:

  This PR fixes an inconsistency in the existing Safari detection logic:

  - `supportsDolbyVision()` (line 265) already returns true for Safari on iOS 13+ / macOS - this is existing code, not part of this PR
  - `supportedDolbyVisionProfilesHevc()` (line 270) returns [] because `canPlayType('video/mp4; codecs="dvh1.05.06"')` returns empty on Safari
  - Result: supportsDolbyVision says yes, but zero profiles are declared --> no DOVI range types sent to server --> P5 content is transcoded to SDR

Why canPlayType fails: Apple treats Dolby Vision as a managed HLS capability. AVPlayer handles DoVi when the manifest signals dvh1 in CODECS, but canPlayType only tests bare MP4 codec support a separate code path in AVFoundation that doesn't expose DoVi. This is the same limitation that webOS has (line 293: web0sVersion >= 4 fallback for P8).

  Safety by profile:
  - P8: Safe on all devices. The base layer is standard BT.2020 PQ (HDR10). Devices without DoVi hardware decode the HDR10 base layer with correct colors.
  - P5: Correct on DoVi-capable devices (iPhone 12+, Apple Silicon Macs). On older hardware (Intel Macs, pre-A14 iPhones), P5's IPT-PQ color space would display incorrectly without DoVi processing. However, the alternative without this PR is SDR transcode also a significant quality loss. This is the same optimistic tradeoff that webOS uses.

iPadOS 13+ detection: browser.js has a `maxTouchPoints > 1` workaround (line 269-274) that correctly identifies iPadOS 13+ as browser.iOS, despite Apple's UA reporting as desktop macOS. The condition `browser.iOS && browser.iOSVersion >= 13` correctly captures these devices.

Companion PR: https://github.com/jellyfin/jellyfin/pull/16362 adds a spec-compliant dvh1 variant to the HLS master playlist alongside the existing hvc1 variant. Together, these two PRs complete the DoVi P5 pipeline: this PR tells the server "Safari supports DoVi", and #16362 ensures the manifest properly signals it.

nyanmisaka:
> Safari's canPlayType() does not report Dolby Vision support — even on devices with hardware DoVi decoding (VideoToolbox).

This statement is incorrect. Dolby provides a webpage using CanPlayType to test `dvh1` and `dav1` support: https://ott.dolby.com/codec_test/index.html

Apple simply prefers not to report Dolby Vision support via CanPlayType on older devices where the screen doesn't support Dolby Vision, but the GPU renderer can handle it. Workarounds can be added to these quirks instead of completely disregarding CanPlayType.

IMG_0391

src/scripts/browserDeviceProfile.js
```diff
@@ -273,9 +273,16 @@ function supportedDolbyVisionProfilesHevc(videoTestElement) {
...
-            .replace(/no/, '')) {
+        if (
+            videoTestElement
+                .canPlayType('video/mp4; codecs="dvh1.05.06"')
+                .replace(/no/, '')
+            // Safari does not expose DoVi via canPlayType even on capable devices.
+            // AVPlayer handles DoVi when the HLS manifest signals dvh1; canPlayType
+            // only tests bare MP4 which Apple treats as a separate path.
+            // Devices without DoVi fall back to the hvc1 PQ variant in the manifest.
+            || (browser.safari && ((browser.iOS && browser.iOSVersion >= 13) || browser.osx))
```
nyanmisaka, line 284: Have you tested this on some older devices that only support up to iOS 15? In my experience, their Safari directly reject Dolby Vision HLS playlist entries. I forget whether they select SDR fallbacks.

NoFear0411: Apple's HLS player explicitly:
  1. Reads all variants
  2. Filters to only playable variants (based on CODECS, device capabilities)
  3. Picks highest bitrate/SCORE among the filtered set

When older Safari "rejects" dvh1 HLS entries, does the player fail entirely, or does it fall back to an hvc1/SDR variant? That's the critical distinction. The Apple spec says it should fall back, but if there's a known Safari bug that I am not aware of it, please do share more info.

_Ref: https://developer.apple.com/documentation/http-live-streaming/hls-authoring-specification-for-apple-devices-appendixes_

NoFear0411:
@nyanmisaka You're correct that `canPlayType` does report DoVi support on devices ONLY with DoVi-capable screens. The PR description was poorly worded, my bad.

> Apple simply prefers not to report Dolby Vision support via CanPlayType on older devices where the screen doesn't support Dolby Vision, but the GPU renderer can handle it.

That is true to some extent. I noticed on the latest iPad A16 (2025-2026) models, Apple claims DoVi support, but JF fails to detect it and sends transcoded SDR version or if you force it, you get a green screen. _I own the A16 2025 model and did the tests there._

  The specific gap this PR targets is devices where the GPU (VideoToolbox) can decode and tonemap DoVi but `canPlayType` returns empty because Apple gates the response on screen DoVi capability, not decoder capability. 
For example, the 2025/2026 iPad models with the A16 chip can process DoVi through VideoToolbox but the display doesn't meet Apple's DoVi screen criteria, so `canPlayType('video/mp4; codecs="dvh1.05.06"')` returns empty.

  The blanket iOS >= 13 condition is too broad though, I agree. 

Regarding older iOS 15 devices, Apple's own https://developer.apple.com/documentation/http-live-streaming/hls-authoring-specification-for-apple-devices-appendixes reference playlist includes dvh1.05.06 variants alongside hvc1 SDR variants, and their docs state the player filters to "_a set of variants, any of which could be played_" before selecting. I think this implies older Safari should gracefully skip dvh1 variants and fall back to hvc1/SDR rather than failing entirely.

That said, could you clarify what you observed on iOS 15 and below, did Safari fail the entire playlist or did it fall back to an SDR/hvc1variant when encountering dvh1entries? That determines whether to keep iOS >= 13 or need to narrow it down more.

thornbill:
Invalid per llm policy

PR #7661, opened 2026-03-07
Subtitle offset increased to [-90, 90] seconds

ThatZiv:
**Changes**

Increased subtitle offset from +/-30s  to 90s. Not sure how it was originally decided but increasing the range makes it easier to match less-than ideal subtitles. More often than not, I find myself needing to offset beyond the 30s range. Others would agree:
- https://www.reddit.com/r/jellyfin/comments/12yk65w/max_subtitle_delay_offset_value/
- https://github.com/jellyfin/jellyfin-web/issues/3783
- https://forum.jellyfin.org/t-subtitle-offset-of-more-less-than-30s
- https://features.jellyfin.org/posts/2308/increase-subtitles-offset


dmitrylyzo:
For reference: #5094

PR #7676, opened 2026-03-12
Extract getCardImageUrl and remove duplicate hook

thornbill:
**Changes**
* Removes duplicate code for react and legacy card components
* Fixes react cards failing to display images for books and music videos

**Issues**
N/A

**Code assistance**
VS Code autocomplete


thornbill:
This is likely dependent on https://github.com/jellyfin/jellyfin-web/pull/7673 to resolve the type error.

PR #7677, opened 2026-03-12
Add display version util for 12.0

thornbill:
**Changes**
Updates the dashboard to show 2 digit version numbers for anything over 10.x

**Issues**
N/A

**Code assistance**
VS Code autocomplete

### DIFF
--- a/src/utils/versions.ts
+++ b/src/utils/versions.ts
@@ -30,7 +30,11 @@ export function getDisplayVersion(version: string | null | undefined) {
 
     // For versions above 10.X, we switch to only showing major.minor to reflect the new versioning scheme.
     if (compareVersions(version, '11') >= 0) {
-        return version.split('.').slice(0, 2).join('.');
+        const parts = version.split('.');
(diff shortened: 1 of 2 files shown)


src/utils/versions.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    return 0;
+}
+
+/** Formats a version (e.g., "1.2.3") string for display. */
+export function getDisplayVersion(version: string | null | undefined) {
+    if (!version) return;
+
+    // For versions above 10.X, we switch to only showing major.minor to reflect the new versioning scheme.
+    if (compareVersions(version, '11') >= 0) {
```
nielsvanvelzen, line 32: We're never gonna release a 11.0, after 10.11.0 comes 12.0. So this comment doesn't matter _that_ much, but wouldn't it be better to check for 12 instead of 11 here? And maybe even check if the last digit is 0 and display it anyway if it isn't.

nielsvanvelzen (approved):

PR #7681, opened 2026-03-13
Fix Audiobook Playback on iOS

### LINKED ISSUE #6113: iOS (Safari): background audio playback broken
**Describe The Bug**
Using Safari on iOS 18, audio playback stops working when the app goes to the background (phone going to standby or when switching the app).
The play time does continue (as can be seen in the media controls) and when switching back to safari, the track/audiobook continues. As if the jellyfin tab/Safari was simply muted; the playback does not get paused.

Furthermore, Trying to pause the playback, some seconds of the track are looped for a moment.

**Steps To Reproduce**
1. Go to a music / audiobook library 
2. Click on a song to start playing it
3. Put phone to standby/switch app
4. go back to safari, hear audio again
5. try pausing the playback

**Expected Behavior**
Playback should continue in the background. Pausing the playback should not cause looping, but directly stop playing audio.

**Logs**
None

**Screenshots**
Not applicable. If needed, I can create videos.

 - Platform: iOS (18)
 - Browser: Safari
 - Jellyfin Version: 10.9.9

**Additional Context**
Other playback works in the background as intended. See (or hear) https://tools.woolyss.com/html5-audio-video-tester/?u=woolyss.com/f/audio-sample.mp3 for example

wustus:
**Changes**
This change skips the setup of an `AudioContext` by returning early before `addGainElement(elem);` is called. The outcome is that audio playback uses the `<audio>` element directly without routing through the Web Audio API.
Since the volume cannot be controlled programmatically on iOS Safari, the skipped `GainNode` code did not work as intended anyways, it just bricked playback on iOS.

**Issues**
This PR aims to fix #6113. Due to `AudioContext` being suspended when Safari on iOS is put in the background, playback of the Audioplayer is stopped and cannot be reattached to when going back to the application.

**Code assistance**
The code needing adjustment and problem space was found using Claude. I did read up on the commented issues a bit but cannot speak to any of this with certainty. In all honesty I just wanted this issue resolved to listen to audiobooks.


wustus:
The above PR was tested on an iPhone 14 Pro, running iOS 26.3.1
With the above change I was able to listen to an Audiobook while the Jellyfin Safari app was put into background.

wustus:
https://github.com/user-attachments/assets/4ae3cf7e-4646-4fa9-a7ba-8b47003dbee4

Here's some video I recorded showing the bug and how it behaves with the implemented fix. The first two examples use my main Jellyfin server, once using the Jellyfin iOS app and the second one using the WebUI. Both times whenever I background safari, playback stops but the time progresses so that when I come back to safari and continue playing, the audio playback sometimes gets stuck in a loop.
The last example is of the patched WebUI version running in development mode on my machine.

LTe:
@wustus I do not think it is good approach. I think we should use https://developer.mozilla.org/en-US/docs/Web/API/AudioSession API and set this to `playback`. I guess this is the right context for audio in jellyfin.

With this browser should know that AudioContext should not be suspended.

wustus:
With the last commit, audio playback should work on iOS devices by default, even is Safari is put into background. Scrubbing through the audio will not work by default. To enable scrubbing, Audio Normalization must be turned off in the user settings.

LTe:
Tested manually

**Steps to reproduce:**

0. Use Safari on iOS
1. Start jellyfin web with `npm run`
2. Login as active user
3. Click on the top right corner in your profile image
4. Settings
5. Playback
6. Make sure that you have Audio Normalization enabled
7. Go to home page
8. Pick audio file (music / book).
9. Put application to background

**Expected behavior**

Music is still playing


**Actual behavior**

Music stops playing

## Video

Recording from `master` branch:

https://github.com/user-attachments/assets/54615bdc-aef0-4e02-9416-099111e1813a

Recording from this PR:

https://github.com/user-attachments/assets/775867b0-9173-4d0f-8974-3bcd2801cff5

wustus:
Thank you for extensive testing!

LTe (approved):
Tested manually https://github.com/jellyfin/jellyfin-web/pull/7681#issuecomment-4083595508

dkanada:
This seems to have fixed the background playback issue but I noticed the audio pitch was frequently jumping higher for 10-20 seconds then reverting to normal. Not sure if it's related.

wustus:
> This seems to have fixed the background playback issue but I noticed the audio pitch was frequently jumping higher for 10-20 seconds then reverting to normal. Not sure if it's related.

Thanks for testing. I'll take a look.

wustus:
> This seems to have fixed the background playback issue but I noticed the audio pitch was frequently jumping higher for 10-20 seconds then reverting to normal. Not sure if it's related.

I apologize for replying late. I was not able to reproduce this. I've listened to an audiobook and some song with high dynamics, playback was fine for both.

thornbill:
Someone else tested this and encountered the same issues that dkanada reported. This will need some additional investigation/work.

jordanmichaelrushing:
This behavior is driving my wife crazy for audiobooks so I'm going to pull it down and test with my slew of audiobook changes. I'll report back on my end in a few hours what happens with her experience. 

Edit: oh it's because of the normalization path, sorry I missed that. Ok so ideally the native player, that I guess isn't ready (is there a roadmap for that somewhere btw?), solves this (it does, I tested it). However, to fix this now I would say this is backwards. Here's my suggestion:
Instead of...
```
if (userSettings.selectAudioNormalization() == 'TrackGain') {
    normalizationGain = options.item.NormalizationGain
        ?? options.mediaSource.albumNormalizationGain;
} else if (userSettings.selectAudioNormalization() == 'AlbumGain') {
    normalizationGain =
        options.mediaSource.albumNormalizationGain
        ?? options.item.NormalizationGain;
} else {
    console.debug('normalization disabled');
    return;
}

// this avoids the AudioContext being suspended when Safari is put into background
// to allow scrubbing, Audio Normalization must be turned off in the settings
if ("audioSession" in navigator) {
    navigator.audioSession.type = "playback";
}
```

Do the audioSession.type = "playback" in the early return (normalization turned off) path:
```
if (userSettings.selectAudioNormalization() == 'TrackGain') {
    normalizationGain = options.item.NormalizationGain
        ?? options.mediaSource.albumNormalizationGain;
} else if (userSettings.selectAudioNormalization() == 'AlbumGain') {
    normalizationGain =
        options.mediaSource.albumNormalizationGain
        ?? options.item.NormalizationGain;
} else {
    console.debug('normalization disabled');
    if ("audioSession" in navigator) { // <- HERE
        navigator.audioSession.type = "playback";
    }
    return;
}
```

This means background play on iOS would only work when normalization if turned OFF, however it means users wouldn't run into a pitch issue. We would need to explain that distinction to users. Otherwise you'd want this line in both pathways (and accept that turning on normalization would introduce pitch issues). 

**Note that I have not tested this yet (I will in a few hours).**

jordanmichaelrushing:
^ Confirmed, putting this change on the denormalized early return path works and is what I would suggest we do here. If that's the agreed upon path, I might suggest a note in the normalization setting outlining this if device is iOS

thornbill:
@jordanmichaelrushing it would be a bit easier to test if you opened a pr with your change

Also what are you referring to by native player? One of the options in the Jellyfin iOS app?

anultravioletaurora:
Tested this branch on my end - background playback is working on iOS with Safari in the background and the screen locked

jordanmichaelrushing:
@thornbill correct. In the iOS hidden menu that [you built](https://github.com/jellyfin/jellyfin-ios/pull/339)

I don't really like stepping on other people's changes, that being said we haven't heard back from @wustus since April. So I'll create a separate PR this time. https://github.com/jellyfin/jellyfin-web/pull/8211

Let me know if you want me to add a little something to the description under the normalization setting

wustus:
@jordanmichaelrushing thanks for continuing on this.

MBR-0001:
superseded by https://github.com/jellyfin/jellyfin-web/pull/8211

PR #7685, opened 2026-03-14
Fix login loop

### LINKED ISSUE #7683: Logout on browser reload

### Describe the bug

Since a few days login is not reload persistent in different browsers on master (without patches) and also JMP.
Browsing works, but a reload will put me back to the login. Never had the issue before.

### Reproduction Steps

1. use current master
2. Press F5 or so
3. End up at login form

### Expected/Actual behaviour

Browsing after F5

### Logs

_No response_

### Server version

10.12.0

### Web version

10.12.0

### Build version

Release

### Platform

Linux

### Browser

Firefox, JMP

### Additional information

3rd party clients are fine.

Crashdummyy:
**Changes**

* Import AUTHORIZATION_HEADER from `@jellyfin/sdk/lib/constants` ( got moved from `/lib/api` )

**Issues**

* Fixes #7683 

**Additional Details**
The constants have been moved in this commit 'https://github.com/jellyfin/jellyfin-sdk-typescript/commit/b4c554a2ba9b42d34797a09955552cf1c200149e#diff-769911c416ccf8514d8fd941ae0abe8fb5c606ade0c218e22151a5f5f9f3d700'

### DIFF
--- a/src/strings/tr.json
+++ b/src/strings/tr.json
@@ -666,7 +666,7 @@
     "ShowTitle": "Başlığı göster",
     "Raised": "Yükseltilmiş",
     "LabelPlaylist": "Oynatma listesi",
-    "LabelPlayer": "Oyanatıcı",
+    "LabelPlayer": "Oynatıcı",
     "LabelPlaceOfBirth": "Doğum yeri",
     "LabelPersonRoleHelp": "Örnek: Dondurma kamyonu şoförü",
     "LabelPersonRole": "Rol",


src/lib/jellyfin-apiclient/connectionManager.js
```diff
@@ -225,6 +225,13 @@ export default class ConnectionManager {
         }
 
         function validateAuthentication(server, serverUrl) {
+            if (!server.AccessToken) {
```
dmitrylyzo: `!server.AccessToken` is `false` when  `validateAuthentication` is called:
https://github.com/jellyfin/jellyfin-web/blob/3169493515995962e2983645ce157c8d5a4fa8eb/src/lib/jellyfin-apiclient/connectionManager.js#L575-L576

Crashdummyy: I can safely remove it.
I thought it would be okay to have it as an additional safeguard.

The "fix" was importing the correct package

klaki892 (approved):
Tested - (Debian 13, Chrome 140.0.7339.81) 

Can confirm this fixes the reproducible issue of 'Remember Me' functionality not working currently. 

(I just spent _too much time_ tracking this down myself)

Issue #7569, opened 2026-02-09
Cannot play album using web interface.


LiamFry:
### This issue respects the following points:

- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-web/issues?q=is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

Clicking album ">" (play) button - either in album art or banner - does "nothing." No song begins to play.

### Reproduction Steps

1) Open main music "folder."
2) Any album, click in center of album art (faint "play" symbol). Nothing happens.
3) Go into the album.
4) On the album "banner" (which shows album, artist, total track info), click ">" (play) button. Nothing happens.

No error whatsoever is reported. Clicking on the ">" (play) symbol(s) simply "does nothing."

### Expected/Actual behaviour

I expect the 1st song to begin playing and (with "Play next episode automatically" enabled) each subsequent song to auto-play thereafter.

### Logs

When you click on the faint ">" (play) button at the center of an album's cover art:
```
[2026-02-09 15:26:13.929 -05:00] [INF] Playback stopped reported by app "Jellyfin Web" "10.11.6" playing "Summer Of Love". Stopped at "0" ms
[...]

PR #7686, opened 2026-03-15
fix for #7569 where an album cannot be played from web interface

stopforumspam:
fix for #7569 where an album cannot be played from web interface

**Changes**
A simple logic change to prevent jellyfin-web from requesting a recursive list of items to play when there is none in the album

**Issues**
whilst Ive tested this myself with albums and boxset albums (one of the conditions in the logic), Ive not been able fault playback.  I 

**Code assistance**
I would appreciate far more learned contributors to test this


src/components/playback/playbackmanager.js
```diff
@@ -1952,20 +1952,22 @@ export class PlaybackManager {
                 }, queryOptions));
             } else if (firstItem.IsFolder) {
                 let sortBy = null;
+                let recursive = true;
                 if (options.shuffle) {
                     sortBy = 'Random';
```
yutamago, line 1957: Are we ignoring the case where options.shuffle = true?

This solution seems to address only one use case out of many...

src/components/playback/playbackmanager.js
```diff
@@ -1952,20 +1952,22 @@ export class PlaybackManager {
...
                 } else if (firstItem.Type !== 'BoxSet') {
                     if (firstItem.CollectionType === 'music' || firstItem.MediaType === 'Audio') {
                         sortBy = 'Album,ParentIndexNumber,IndexNumber,SortName';
                     } else {
+                        recursive = false;
```
yutamago, line 1962: Please explain why this works.
Seems a bit random to me.

stopforumspam: the rest call to the backend returns no data when the parameters are built for the request with the recursive flag set.  setting it false returns the requested data.  maybe a backend api bug?

src/components/playback/playbackmanager.js
stopforumspam: every other use case resulted in a correct rest response

stopforumspam:
im going to close this as 10.11.11 seems to have fixed the issue causing this

PR #7697, opened 2026-03-18
When deleting a single item, include the name of the item in the confirmation message

joshuaobrien:
## Changes
I was deleting something from my library recently and thought it would be helpful if the confirmation text included what's actually being deleted.

If this change looks okay, I will raise further changes to handle multiple items, playlists, etc.

### Before
image

### After
image

## Issues
N/A

## Code assistance
N/A


MBR-0001 (approved):
LGTM, some languages might have to change the wording a bit to make the sentence sound normal but otherwise fine

PR #7732, opened 2026-03-19
Hold and drag on media no longer activates Multi-select

### LINKED ISSUE #7288: Hold and drag on media shelves on home also triggers multi-select UI

### Describe the bug

Once logged in and on the Jellyfin home, if one has many libraries they can click and drag to see more of their media libraries.

However, my partner and I frequently find ourselves accidentally triggering the multi-select UI, which then prevents us from accessing the library we want until we've closed out of it.

### Reproduction Steps

1. Login to Jellyfin.
2. Create multiple libraries if you don't have at least 5. I have 10 for example.
3. In the **My Media** shelf, drag the shelf around to scroll through the libraries.
4. Observe the multi-select UI appears despite actively scrolling.

## Expected

Hold to trigger multi-select makes sense to me, and is quite intuitive. However, if I'm actively moving my cursor or finger around, it should cancel the multi-select behavior entirely until I let go.

For example:

* If I click and hold and either don't move the cursor at all, or perhaps for accessibility keep the cursor within 1rem of the starting location, it will trigger multi-select.
* If I move the cursor further than 1rem away from one I initially held, for the entire click and drag action, the multi-select UI should never appear. (Even if during this same click and drag, I hold my position.)

## Actual

If I drag slowly to get an overview of all libraries, it will almost always trigger the multi-select UI. This must be closed before I'm actually able to access a library.

### Logs

_No response_

### Server version

10.11.1

### Web version

10.11.1

### Build version

10.11.1

### Platform

Linux

### Browser

Firefox 144.0.2 (64-bit) and Jellyfin Media Player

### Additional information

https://github.com/user-attachments/assets/93083cbd-53bb-4cda-be5b-4cbdde696be5

You can see here that I'm trying to drag around and trigger the UI every time.

### Workarounds

* One can use the buttons at the top-right to paginate.
* One can do very short drags and move the mouse back and forth dragging and letting go.
* One can click and drag then move the cursor outside the shelf while dragging. This allows the user to drag without triggering the multi-select UI.

bjorntp:
**Changes**
When moving the mouse atleast 5px in either X or Y direction multi-select will no longer activate. Same logic was already in place for touch screen so was almost a copy paste.

**Issues**
Fixes #7288 

**Code assistance**
Codex helped identify that there was already logic in place for touch screens in multiSelect.js

### DIFF
--- a/src/components/multiSelect/multiSelect.js
+++ b/src/components/multiSelect/multiSelect.js
@@ -495,13 +495,35 @@ export default function (options) {
     }
 
     function onMouseDown(e) {
+        touchTarget = null;
+        touchStartX = e.clientX || 0;
+        touchStartY = e.clientY || 0;
+
+        const element = e.target;
+        if (!dom.parentWithClass(element, 'card')) {
+            return;
+        }
+
         if (touchStartTimeout) {
             clearTimeout(touchStartTimeout);
             touchStartTimeout = null;
         }
 
-        touchTarget = e.target;
+        touchTarget = element;
         touchStartTimeout = setTimeout(onTouchStartTimerFired, 550);
+    }
+
+    function onMouseMove(e) {
+        if (!touchTarget) {
+            return;
+        }
+
+        const deltaX = Math.abs((e.clientX || 0) - (touchStartX || 0));
+        const deltaY = Math.abs((e.clientY || 0) - (touchStartY || 0));
+
+        if (deltaX >= 5 || deltaY >= 5) {
+            onMouseOut();
+        }
     }
 
     function onMouseOut() {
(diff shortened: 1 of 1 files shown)


Mayme123:
Tested the fix locally and works as described. The issue is reproducible on master where trying to hold and drag on cards triggers multi-select. These changes prevent that if the mouse is moved 5 or more pixels, which should be the intended behavior.

MBR-0001 (approved):
LGTM, tested and works

thornbill:
This needs rebased on master instead of the release branch to be merged.

thornbill (approved):

PR #7741, opened 2026-03-21
Add a folder view to the modern home videos library layout

theguymadmax:
### Changes

Adds a new default folder tab to the Home Videos and Photos library in the modern/experimental layout. Update corresponding default landing screen options. 

This enables users to view media with the same folder structure as in the legacy view.

### Issues
N/A

### Code assistance
N/A

### DIFF
--- a/src/apps/experimental/features/libraries/constants/libraryRoutes.ts
+++ b/src/apps/experimental/features/libraries/constants/libraryRoutes.ts
@@ -178,25 +178,24 @@ export const LibraryRoutes: LibraryRoute[] = [
         views: [
             {
                 index: 0,
-                label: 'Photos',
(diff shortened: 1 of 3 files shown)


thornbill:
I'm curious if this shouldn't be **the** default :man_shrugging:

theguymadmax:
Yes, I think that makes sense. Right now, if you have videos (no photos) in this library, the default tab lands on the photos tab, which will be empty. This may confuse some users at first. 

So the tab layout will be: 
1) Folders  (default tab)
2) Photos
3) Photo Albums
4) Videos

This way, the default tab will always land on a page with media. Then the user can customize the landing screen options if they prefer another tab.

thornbill:
Yes that is what I was thinking also.

PR #7751, opened 2026-03-24
'Ends at' takes into account if video not started yet

HaraldNordgren:
It bothers me that when checking the info on an assets, it reads "Ends at" even when I'm not playing it yet or even intended the play it now.

I ran the copy through Cluade Sonnet 4.6, but I seen now that it seems to be handled by Weblate for this project. So how to do that? Will it work to just leave the `globalize.translate('EndsAtIfStartedNowValue', time)` call but no translation strings and the correct translation jobs will be kicked off automatically upon merging?


ferferga:
I don't think the current wording it's misleading or confusing by any way, so personally I don't see a reason to change it.

thornbill:
Closed as invalid. You did not read the contributing guide or complete the PR template.

HaraldNordgren:
> I don't think the current wording it's misleading or confusing by any way, so personally I don't see a reason to change it.

@ferferga I feel like it is a bit jarring. Because why am I being told that it ends at a certain time when I didn't start any video yet? 🤗

HaraldNordgren:
> Closed as invalid. You did not read the contributing guide or complete the PR template.

My bad!

PR #7774, opened 2026-03-28
Add configurable trickplay image preloading

rzalawad:
### Changes

Add an optional per-user setting to preload all trickplay sprite sheet images when a video starts playing. When enabled, the client fires GET requests for all sprite sheet tiles at playback start using `new Image()` objects, which populates the browser's HTTP cache. When the user later hovers the progress bar, the same URLs are served from cache instead of making new network requests, eliminating the ~300ms delay per tile.

The setting is a checkbox in the user's Playback settings page (Advanced section), defaulting to off.

### Issues

Related discussion: https://github.com/orgs/jellyfin/discussions/14182

---

### DIFF
--- a/src/components/playbackSettings/playbackSettings.js
+++ b/src/components/playbackSettings/playbackSettings.js
@@ -230,6 +230,7 @@ function loadForm(context, user, userSettings, systemInfo, apiClient) {
     context.querySelector('.chkEnableDts').checked = appSettings.enableDts();
     context.querySelector('.chkEnableTrueHd').checked = appSettings.enableTrueHd();
     context.querySelector('.chkEnableHi10p').checked = appSettings.enableHi10p();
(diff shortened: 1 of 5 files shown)


trumanmiller:
follows existing patterns in the codebase, i tested and it works as expected

with default trickplay settings an hours worth of previews is 4x ~1MB requests that occur after the first chunk of video is loaded so it shouldn't slow down initial video load.

however, it should be enabled by default; 99% of users wont think to look for the setting, it makes the experience noticeably more snappy, and there isn't really a downside.

rzalawad:
@trumanmiller Thank you for trying out the changes! Regarding what the default setting should be, I'm gonna defer to what a PR approver has to say although I do prefer (like you) for it to be enabled by default as well.

picarica:
incredible work

PR #7795, opened 2026-04-02
fix: show artist name in now playing bar, fullscreen player, and queue

siltus:
## Changes

- **Now playing bar**: Use full item data from API to update artist text, since \NowPlayingItem\ from playback state often lacks \ArtistItems\/\Artists\ fields
- **Fullscreen player**: Simplify artist rendering logic using \.map().join()\, add \.catch()\ error handling for the async item fetch
- **Queue/playlist**: Batch-fetch full item details for items missing \ArtistItems\ before rendering, ensuring all songs display their artist name

## Problem

For some songs, the artist name was missing in the player UI (now playing bar, fullscreen player, and queue). The root cause is that the \NowPlayingItem\ object from the playback state and playlist items from \getPlaylist()\ don't always have \ArtistItems\ or \Artists\ populated.

## Solution

- In the now playing bar, the existing \piClient.getItem()\ call (used for user data buttons) now also updates the secondary text with artist info from the full item.
- In the fullscreen player, added error handling to the existing async fetch.
- In the queue, added a batch fetch (\piClient.getItems()\ with \Fields: ArtistItems\) for playlist items missing artist data before rendering the list view.

### DIFF
--- a/src/components/nowPlayingBar/nowPlayingBar.js
+++ b/src/components/nowPlayingBar/nowPlayingBar.js
@@ -544,6 +544,18 @@ function updateNowPlayingInfo(state) {
     if (nowPlayingItem.Id) {
         const apiClient = ServerConnections.getApiClient(nowPlayingItem.ServerId);
         apiClient.getItem(apiClient.getCurrentUserId(), nowPlayingItem.Id).then(function (item) {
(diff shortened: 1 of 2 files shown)


siltus:
Apologies, my stupid LLM opened the PR upstream instead of my fork. Please ignore

PR #7796, opened 2026-04-03
Load device information lazily in connection manager

nielsvanvelzen:
### Changes

We want to avoid using the apphost before it is initialized. To accomplish this 2 changes were needed:

- Lazily load values from apphost in connection manager instead of reading them immediately on connectionmanager import
- Fix the servernotifications initializing itself on import, causing apiclients to be instantiated before apphost init

### Issues

Fixes https://github.com/jellyfin/jellyfin/issues/16534

### Code assistance

If only AI could make sense of this code 🫠 

---

### DIFF
--- a/src/scripts/serverNotifications.js
+++ b/src/scripts/serverNotifications.js
@@ -201,10 +201,12 @@ function bindEvents(apiClient) {
     Events.on(apiClient, 'message', onMessageReceived);
 }
 
-ServerConnections.getApiClients().forEach(bindEvents);
-Events.on(ServerConnections, 'apiclientcreated', function (e, newApiClient) {
-    bindEvents(newApiClient);
-});
+export function initializeServerConnections() {
+    ServerConnections.getApiClients().forEach(bindEvents);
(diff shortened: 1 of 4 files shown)


src/scripts/serverNotifications.js
```diff
@@ -201,10 +201,12 @@ function bindEvents(apiClient) {
...
-ServerConnections.getApiClients().forEach(bindEvents);
-Events.on(ServerConnections, 'apiclientcreated', function (e, newApiClient) {
-    bindEvents(newApiClient);
-});
+export function initializeServerConnections() {
+    ServerConnections.getApiClients().forEach(bindEvents);
+    Events.on(ServerConnections, 'apiclientcreated', function (e, newApiClient) {
+        bindEvents(newApiClient);
+    });
+}
```
nielsvanvelzen, line 209: Alternative here is to just not call the ` ServerConnections.getApiClients().forEach(bindEvents);` line, as the event listener is registered before any apiclients are made. But that still leaves side-effects so I'd say this approach of adding an initialize function would be better.

thornbill (approved):

PR #7798, opened 2026-04-03
Add chapter name inside osd slider bubble

will-blk:
The goal of this PR is to restores the chapter name into the osd slider bubble. This was previously requiring an chapter image, something that I personally never seen before.

### Changes

Before:
image

After:
image

### Issues

- None

### Code assistance

- None

---

### DIFF
--- a/src/controllers/playback/video/index.js
+++ b/src/controllers/playback/video/index.js
@@ -1575,22 +1575,24 @@ export default function (view) {
         }
 
         const src = getImgUrl(item, chapter, index, 400, apiClient);
+        let html = '<div class="chapterThumbContainer chapterBubblePosition">';
 
         if (src) {
-            let html = '<div class="chapterThumbContainer">';
             html += '<img class="chapterThumb" src="' + src + '" />';
             html += '<div class="chapterThumbTextContainer">';
-            html += '<div class="chapterThumbText chapterThumbText-dim">';
-            html += escapeHtml(chapter.Name);
-            html += '</div>';
-            html += '<h2 class="chapterThumbText">';
-            html += datetime.getDisplayRunningTime(positionTicks);
-            html += '</h2>';
-            html += '</div>';
-            return html + '</div>';
+        } else {
+            html += '<div class="chapterThumbTextContainer chapterBubblePosition">';
         }
 
-        return null;
+        html += '<div class="chapterThumbText chapterThumbText-dim">';
+        html += escapeHtml(chapter.Name);
+        html += '</div>';
+        html += '<h2 class="chapterThumbText">';
+        html += datetime.getDisplayRunningTime(positionTicks);
+        html += '</h2>';
+        html += '</div>';
+
+        return html + '</div>';
     }
 
     let playPauseClickTimeout;
@@ -1920,7 +1922,7 @@ export default function (view) {
         ticks *= value;
         const item = currentItem;
 
-        if (item?.Chapters?.length && item.Chapters[0].ImageTag) {
+        if (item?.Chapters?.length) {
             const html = getChapterBubbleHtml(ServerConnections.getApiClient(item.ServerId), item, item.Chapters, ticks);
 
             if (html) {
--- a/src/styles/videoosd.scss
+++ b/src/styles/videoosd.scss
@@ -54,7 +54,10 @@
 .chapterThumbContainer {
     box-shadow: 0 0 1.9vh #000;
     flex-grow: 1;
-    position: relative;
+}
+
+.chapterBubblePosition {
+    position: relative !important;
 }
 
 .chapterThumb {
@@ -88,6 +91,7 @@
     background: none;
     padding: 0.25em 0.5em;
     user-select: none;
+    min-width: fit-content;
 }
 
 .chapterThumbText {
@@ -96,6 +100,8 @@
     opacity: 1;
     color: #fff;
     text-shadow: 0 0 25px #000, 0 0 6px #000;
+    white-space: nowrap;
+    min-width: fit-content;
 }
 
 .chapterThumbText-dim {


will-blk:
This is my second PR for an open source project, take it easy on me :laughing:

MBR-0001 (changes requested):
I like the idea 🙂
I've got an observation though - the bubble is not centered, it looks a bit out of place

https://github.com/user-attachments/assets/db73f811-446d-4bcf-98be-6e815ad25568

And also, make sure your PR has all the sections that are in the template (changes, issues, code assistance etc)

will-blk:
> I like the idea 🙂 I've got an observation though - the bubble is not centered, it looks a bit out of place
> 
>  Screencast_20260605_000043.webm 
> And also, make sure your PR has all the sections that are in the template (changes, issues, code assistance etc)

Mind to help out? This bubble width is always set to 0px, so the offset never centers it above the cursor. My css/ui knowledge isn't that good, so this is proving to be a headache. One partial progress was to reorder
```
function updateBubble(range, percent, value, bubble) {
    requestAnimationFrame(function () {
        if (!range.updateBubbleHtml?.(bubble, value)) {
            let html;

            if (range.getBubbleHtml) {
                html = range.getBubbleHtml(percent, value);
            } else {
                if (range.getBubbleText) {
                    html = range.getBubbleText(percent, value);
                } else {
                    html = value.toLocaleString();
                }
                html = '<h1 class="sliderBubbleText">' + html + '</h1>';
            }

            bubble.innerHTML = html;
        }

        requestAnimationFrame(function () {
            const bubbleTrackRect = range.sliderBubbleTrack.getBoundingClientRect();
            const bubbleRect = bubble.getBoundingClientRect();

            let bubblePos = bubbleTrackRect.width * percent / 100;
            if (globalize.getIsElementRTL(range)) {
                bubblePos = bubbleTrackRect.width - bubblePos;
            }
            bubblePos = Math.min(Math.max(bubblePos, bubbleRect.width / 2), bubbleTrackRect.width - bubbleRect.width / 2);

            bubble.style.left = bubblePos + 'px';
        });
    });
}
```
moving ` bubble.innerHTML = html;` before any attempt of `bubbleTrackRect.width` this gave width to the texts contents but not the container.

thornbill:
Wow I thought I had left feedback on this one a long time ago. 😅

We probably would want to keep a check that the chapter has either an image or a name.

will-blk:
> Wow I thought I had left feedback on this one a long time ago. 😅
> 
> We probably would want to keep a check that the chapter has either an image or a name.

There is already a guard in place at line 1573. :grin: 

```
if (!chapter) {
  return null;
}
```

MBR-0001:
I spent 30 mins yesterday trying to solve this and gave up, asked claude for help and it managed to achieved this

[Screencast_20260605_222058.webm](https://github.com/user-attachments/assets/3404344b-6543-4f01-b36d-23143c767ac2)

with the following change
image

```css
.chapterThumbContainer:not(:has(.chapterThumb)) .chapterThumbTextContainer {
    position: relative;
}
```
honestly this part of code is really puzzling and I have no idea if that breaks something, but at least I tried :smile:

will-blk:
> I spent 30 mins yesterday trying to solve this and gave up, asked claude for help and it managed to achieved this
> 
>  Screencast_20260605_222058.webm 
> with the following change image
> 
> ```css
> .chapterThumbContainer:not(:has(.chapterThumb)) .chapterThumbTextContainer {
>     position: relative;
> }
> ```
> 
> honestly this part of code is really puzzling and I have no idea if that breaks something, but at least I tried 😄

If it works, it works. :laughing: Updated the scss with it. Locally also looks fine and it didn't replace the current design without chapters

MBR-0001:
The image view is now a bit broken

master:
image

now:
image

will-blk:
> The image view is now a bit broken
> 
> master: image
> 
> now: image

Could you send me this file? I need something with the images to test it out.

src/styles/videoosd.scss
```diff
@@ -53,7 +53,10 @@
 
 .chapterThumbContainer {
     box-shadow: 0 0 1.9vh #000;
-    flex-grow: 1;
+    position: relative;
+}
+
+.chapterThumbContainer:not(:has(.chapterThumb)) .chapterThumbTextContainer {
```
MBR-0001:
```suggestion
.chapterThumbContainer:not(:has(.chapterThumbWrapper)) .chapterThumbTextContainer {
```

It seems the img element is never rendered but instead is replaced with a div with a background-image

Image

MBR-0001:
The image themselves are generated by jellyfin by enabling trickplay in library settings @will-blk

src/styles/videoosd.scss
MBR-0001: it would seem there are two different codepaths here, one for media that has trickplay images enabled and then one for media without trickplay but with chapter images, first time hearding that exists. Maybe the selector should support both

EDIT: and it seems that's a library option too, chapter images!

will-blk:
It seems trickplay uses a different bubble from the chapters images. 

Trickplay:
```
# src/controllers/playback/video/index.js:1466
function updateTrickplayBubbleHtml
```

Chapter:
```
# src/controllers/playback/video/index.js:1560
function getChapterBubbleHtml
```

So, the best approach would be replacing the css from the bubble to something unique :thinking: Thoughts?

src/styles/videoosd.scss
dmitrylyzo: We can't use `:has` because old web engines don't support it.
https://caniuse.com/css-has

MBR-0001:
I think the easiest solution would be add a new css class next to `chapterThumbContainer` (so at approximately line 1578 - where you create the html variable) that would tell css whether there is an image or not, so there's need for `:not(:has(.chapterThumb))`

will-blk:
Ok, moved the relative position to a class only used by the chapterBubble.

src/styles/videoosd.scss
will-blk: Yes 😆  I wonder why they have different options and all if they look the same and has almost the same generation.

MBR-0001:
seems it is still not centered

[Screencast_20260616_200429.webm](https://github.com/user-attachments/assets/b41f9473-e554-4d9d-99ef-931877203870)

will-blk:
> seems it is still not centered
> 
>  Screencast_20260616_200429.webm

My bad :man_facepalming: I forgot to add the change to the commit

MBR-0001:
sorry to nag but there is now a difference between the trickplay view and the chapter images view
trickplay (div and h2):
image

chapter image previews (affected by changes, h4):
image
could you revert the changes in hierarchy so they match?

idea: you could use the h4 when there is no `src` and the previous hierarchy when there is

will-blk:
@MBR-0001 How do you switch between the trickplay and chapter views? 

Also, I will set both to use h2 them, too much noise code/styling making them apart, but honestly, I would use h4 on all of them, h2 font is too big with/without images.

MBR-0001:
> @MBR-0001 How do you switch between the trickplay and chapter views?
> 
> Also, I will set both to use h2 them, too much noise code/styling making them apart, but honestly, I would use h4 on all of them, h2 font is too big with/without images.

I symlinked the same movie in a few folders and then added multiple libraries each pointing to its own folder, one has trickplay, one has chapter images and one has none of those

src/controllers/playback/video/index.js
```diff
@@ -1563,22 +1575,24 @@ export default function (view) {
...
-            html += datetime.getDisplayRunningTime(positionTicks);
-            html += '</h2>';
-            html += '</div>';
-            return html + '</div>';
+        } else {
+            html += '<div class="chapterThumbTextContainer chapterBubblePosition">';
         }
 
-        return null;
+        html += '<h2 class="chapterThumbText chapterThumbText-dim">';
```
MBR-0001:
```suggestion
        html += '<div class="chapterThumbText chapterThumbText-dim">';
```

src/controllers/playback/video/index.js
```diff
@@ -1563,22 +1575,24 @@ export default function (view) {
...
+        html += escapeHtml(chapter.Name);
+        html += '</h2>';
```
MBR-0001:
```suggestion
        html += '</div>';
```

MBR-0001: Image

This would make the chapter name less huge and also make it so all variants of the bubble look the same

chapter images

Image

trickplay

Image

MBR-0001 (approved):

PR #7799, opened 2026-04-04
Remove redundant mixed pt locale in favor of pt-BR and pt-PT

Blackspirits:
Changes included:
- removed `src/strings/pt.json`
- removed `pt` from the supported locale list and UI language options
- mapped generic `pt` to `pt-pt` for backward compatibility


Blackspirits:
Hi @thornbill ,

I understand the PR was closed, so before proposing any similar change again, I would like to clarify the intended locale policy for Portuguese.

My main concern is that a generic `pt` locale does not really have a clear editorial standard once both `pt-PT` and `pt-BR` already exist.

In practice, `pt` usually ends up becoming a mixed locale over time, with wording, spelling, and terminology from both European Portuguese and Brazilian Portuguese. That makes the translation inconsistent, less natural, and ultimately less professional for everyone.

From a localization quality perspective, I believe it is better to keep only explicit variants such as `pt-PT` and `pt-BR`, while handling backward compatibility separately if needed.

My intention with the PR was to reduce ambiguity, improve consistency, and still preserve compatibility by mapping generic `pt` to `pt-PT`, but I understand that this may not match the project’s expectations or technical decisions.

Could you please clarify what the preferred direction is for Portuguese locales in Jellyfin before I consider any follow-up proposal?

Thanks.

thornbill:
The PR is invalid because the template was not followed. The changes have not been reviewed.

PR #7810, opened 2026-04-08
Respect DisplayMissingEpisodes setting in item details pages

lmaotrigine:
Follows up on PR https://github.com/jellyfin/jellyfin-web/pull/7528

This PR does not display missing episodes in item details pages unless the user has `DisplayMissingEpisodes` enabled.

I noticed this happening in Person details pages, and I'm not sure if anything else is affected. I was initially wary of attempting this because the `scripts` directory had an ominous here-be-dragons warning, but this file does not seem to be referenced anywhere else.

### Changes
The query parameters for child items in `/details` pages include `isMissing` based the user's preferences. Not sure if this is ideal but from initial testing it seems to work, since I think
this parameter is just ignored for non-episode items..? Unless there is an edge case not present in my own library.

I did consider just passing around the relevant boolean instead of the user object but I see that other such usage in the library passes around the `user` itself, so I have done the same.

### Issues
None reported

### Code assistance
no.

---

### DIFF
--- a/src/scripts/itemsByName.js
+++ b/src/scripts/itemsByName.js
@@ -381,7 +381,7 @@ function getQuery(options, item, user) {
         Limit: 100,
         StartIndex: 0,
         CollapseBoxSetItems: false,
-        isMissing: !user.Configuration.DisplayMissingEpisodes ? false : undefined,
+        isMissing: !user?.Configuration?.DisplayMissingEpisodes ? false : undefined,
     };
     query = Object.assign(query, options || {});
     addCurrentItemToQuery(query, item);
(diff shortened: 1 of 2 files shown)


src/scripts/itemsByName.js
```diff
@@ -261,7 +261,8 @@ function renderSection(item, element, type) {
                 ArtistIds: '',
                 AlbumArtistIds: '',
                 Limit: 6,
-                SortBy: 'SortName'
+                SortBy: 'SortName',
+                isMissing: !user?.Configuration?.DisplayMissingEpisodes ? false : undefined
```
dmitrylyzo, line 265: You should probably use `user?.Configuration?.DisplayMissingEpisodes === false`. Otherwise, if the option isn't set (no user / no configuration), you will disable missing episodes.

dmitrylyzo: Nvm. Here the same logic ("No explicit `DisplayMissingEpisodes: true` - no missing episodes"):  https://github.com/jellyfin/jellyfin-web/blob/5fd5996b020396167598c9360095c0da722157d1/src/apps/stable/features/search/api/useSearchItems.ts#L77

thornbill (approved):

PR #7811, opened 2026-04-08
Add configurable playback buffer size for HLS player

GeiserX:
### Changes

FYI: This is written by me, not by LLMs.
This PR adds a playback buffer size setting (Auto / Large / Extra Large) under Settings > Playback that controls how far ahead the HLS player can buffer video.

- **Auto**: preserves current master behavior (30s default, 6s for high-bitrate streams on Chrome/Edge/Firefox per #6985) - check the discussion at https://github.com/jellyfin/jellyfin-web/pull/7455
- **Large**: allows hls.js to buffer up to 120s ahead with a 500 MB memory budget
- **Extra Large**: allows hls.js to buffer up to 240s ahead with a 3 GB memory budget (enough for ~240s of a 100 Mbps stream)

Of course any other defaults, or having maybe a slider (?) could be considered. But having a really long buffer could be really interesting for some devices with access to a lot of resources which have a quite weak Internet connection.

All presets keep the same initial buffer target from #6985 (6s or 30s depending on bitrate). The larger presets raise both the time ceiling (`maxMaxBufferLength`) and the memory ceiling (`maxBufferSize`), so hls.js can actually use the deeper buffer. These are explicit opt-in overrides — users choosing them accept higher memory usage and the possibility of re-exposing browser-side buffering issues on some clients (e.g. the Chrome 138 class of issue from #6985).

FYI2: If #7455 lands first, this branch will need a rebase so that Auto uses the new dynamic bitrate-based calculation instead of the current 30s / 6s behavior.

Companion PRs for the native apps, I will probably add the new "Auto" behavior proposed in the other PR as well when they're merged:
- https://github.com/jellyfin/jellyfin-android/pull/1946
- https://github.com/jellyfin/jellyfin-androidtv/pull/5515

### Issues

Helps with https://github.com/jellyfin/jellyfin-web/issues/7454
Helps with https://github.com/jellyfin/jellyfin/issues/15787

### DIFF
--- a/src/components/playbackSettings/playbackSettings.js
+++ b/src/components/playbackSettings/playbackSettings.js
@@ -236,6 +236,7 @@ function loadForm(context, user, userSettings, systemInfo, apiClient) {
     context.querySelector('.chkRememberAudioSelections').checked = user.Configuration.RememberAudioSelections || false;
     context.querySelector('.chkRememberSubtitleSelections').checked = user.Configuration.RememberSubtitleSelections || false;
     context.querySelector('.chkExternalVideoPlayer').checked = appSettings.enableSystemExternalPlayers();
+    context.querySelector('#selectPlaybackBufferSize').value = appSettings.playbackBufferSize();
     context.querySelector('.chkLimitSupportedVideoResolution').checked = appSettings.limitSupportedVideoResolution();
     context.querySelector('#selectPreferredTranscodeVideoCodec').value = appSettings.preferredTranscodeVideoCodec();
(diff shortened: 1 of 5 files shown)


GeiserX:
This was ideated after a discussion with @AinzRimuru in https://github.com/jellyfin/jellyfin-web/pull/7455 (already mentioned it in the PR desc, but worth mentioning it explicitly)
AND that the bug could still trigger on Chrome if setting extra_large setting. But we're counting on that this change is for expert users who understand this and want to get away from the "auto" buffer which only gives you stuttering stream.

PR #7822, opened 2026-04-14
Add folders tab to experimental view in book libraries

dkanada:
### Changes

Adds back the folder view to book libraries. I envision this as a stopgap until Author and Series tabs are supported, at which point folders can be sidelined.

### Issues

None

### Code Assistance

None


theguymadmax:
You have a lint issue,

theguymadmax (approved):
You may want to add a books case to the homeScreenSettings, so users can set the preferred default tab.

PR #7827, opened 2026-04-14
Add per-title subtitle appearance settings

ocp1006:
### Summary

Subtitle appearance settings (font, size, color, etc.) are currently global and apply to all content. This adds per-title overrides so users can set subtitle appearance independently for each media item.

### Changes
- Adds a **Subtitle Appearance** option to the playback settings menu (gear icon)
- Saves appearance under a per-item key (`subtitleappearance_<itemId>`) in localStorage
- Falls back to the global subtitle appearance when no per-title setting exists
- Works for all subtitle rendering paths (native VTT cues and custom element rendering)

Addresses https://features.jellyfin.org/posts/3868

### Code assistance
Claude Code with Sonnet 4.6 was used for to brainstorm the implementation.

---
- [x] I have provided a substantive review of https://github.com/jellyfin/jellyfin-web/pull/7842


src/components/playback/playersettingsmenu.js
```diff
@@ -282,6 +287,11 @@ function handleSelectedOption(id, options, player) {
                 options.onOption('suboffset');
             }
             return Promise.resolve();
+        case 'subtitleappearance':
+            if (options.onOption) {
+                options.onOption('subtitleappearance');
+            }
```
DerGernTod: you could use optional chaining here:                 
```ts
options.onOption?.('subtitleappearance');
```

src/controllers/playback/video/index.js
```diff
@@ -1184,6 +1186,57 @@ export default function (view) {
...
 
+    function showSubtitleAppearanceDialog() {
+        const item = currentItem;
+        if (!item) return;
+
+        const appearanceKey = `subtitleappearance_${item.Id}`;
+        const apiClient = ServerConnections.getApiClient(item.ServerId);
+        const userId = apiClient.getCurrentUserId();
+
+        Promise.all([
```
DerGernTod: you could get rid of one level of nesting here if you make `showSubtitleAppearanceDialog` async, makes it a bit simpler to understand imho (especially line 1200)

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -1434,7 +1439,7 @@ export class HtmlVideoPlayer {
             document.getElementsByTagName('head')[0].appendChild(styleElem);
         }
 
-        styleElem.innerHTML = this.getCueCss(subtitleAppearanceHelper.getStyles(userSettings.getSubtitleAppearanceSettings()), '.htmlvideoplayer');
+        styleElem.innerHTML = this.getCueCss(subtitleAppearanceHelper.getStyles(userSettings.getSubtitleAppearanceSettingsWithFallback(this.#getSubtitleAppearanceItemKey())), '.htmlvideoplayer');
```
DerGernTod: there's 4 levels of nesting here now. would be easier to read if you add descriptive variables (so you don't have to read it inside-out to recognize the order), e.g.
```ts
const subtitleAppearanceKey = this.#getSubtitleAppearanceItemKey();
const appearanceSettings = userSettings.getSubtitleAppearanceSettingsWithFallback(subtitleAppearanceKey);
const styles = subtitleAppearanceHelper.getStyles(appearanceSettings);
styleElem.innerHTML = this.getCueCss(styles, '.htmlvideoplayer');
```

src/scripts/settings/userSettings.js
```diff
@@ -621,6 +621,19 @@ export class UserSettings {
...
 
+    /**
+     * Get subtitle appearance settings for an item, falling back to global settings if none are saved for that item.
+     * @param {string|undefined} itemKey - Per-item settings key (e.g. 'subtitleappearance_<itemId>').
+     * @return {Object} Subtitle appearance settings.
+     */
+    getSubtitleAppearanceSettingsWithFallback(itemKey) {
+        const raw = itemKey && this.get(itemKey, false);
+        if (raw) {
+            return Object.assign({}, defaultSubtitleAppearanceSettings, JSON.parse(raw));
```
DerGernTod: does it make sense to guard this `JSON.parse`? or are we certain `raw` is always parsable?

DerGernTod (approved):
just a few nits, but i think this should work

src/components/playback/playersettingsmenu.js
ocp1006: Fixed.

src/controllers/playback/video/index.js
ocp1006: Fixed.

src/plugins/htmlVideoPlayer/plugin.js
ocp1006: Implemented.

src/scripts/settings/userSettings.js
ocp1006: Fixed.

ocp1006:
Thanks @DerGernTod for the CR, all are good suggestions. I've implemented them all 💪

DerGernTod (approved):

ocp1006:
Hi! This PR is updated with current master, merge conflicts are resolved, and SonarCloud is passing again.

Could someone take another look when available? cc @DerGernTod since you previously reviewed this. Happy to address anything else needed to get this merged.

DerGernTod:
lgtm but i don't have write access so my approval isn't enough for the review gate

PR #7835, opened 2026-04-16
fix broken gamepad controls

### LINKED ISSUE #7834: Gamepad navigation broken

### Describe the bug

Gamepad navigation has the following issues (I'm using an Xbox Wireless Controller):
* pressing A/B repeatedly confirms/cancels, up to the point that holding the button just a few milliseconds navigates deep in / out , starts playback or exits the whole app
* using the left thumbstick to navigate triggers lots of navigations (e.g. pushing down for a split second selects the very bottom element, making it impossible to navigate)

Navigating with the D-Pad seems to work fine.

### Reproduction Steps

1. Connect a gamepad (I tested it with an Xbox Wireless Controller and Microsoft Edge)
2. Try to navigate using the left thumbstick and A/B buttons

### Expected/Actual behaviour

The respective input event is fired way too often, making it impossible to navigate.

### Logs

_No response_

### Server version

10.12.0

### Web version

10.12.0

### Build version

10.12.0

### Platform

Windows 11

### Browser

Edge

### Additional information

_No response_

DerGernTod:
fixes #7834 

sorry for the big change, but the code was really hard to read and i thought it could use a rework. i tried to keep the original sematics as close to the original as possible.


DerGernTod:
sorry for the bad PR description!

PR #7836, opened 2026-04-16
Make library header sticky

thornbill:
### Changes
Makes the library header "sticky" so it remains fixed when scrolling and removes the bottom pagination

Best reviewed with whitespace off

### Issues
N/A

### Code assistance
Zed autocomplete

---


nielsvanvelzen (changes requested):
Image

There is a 1px gap

nielsvanvelzen (approved):

MBR-0001 (approved):
Seems good

Image

the letter picker looks a tiny bit "odd" here but can't think of a way to do it differently, making it move below the sticky bar would cause it to jump around which is worse

thornbill:
Yeah still need a better solution for the letter filter but it was already causing layout issues 🫠

PR #7837, opened 2026-04-17
Fix Gamepad controls repeatedly firing key and click events

### LINKED ISSUE #7834: Gamepad navigation broken

DerGernTod:
### Changes
* Thumbstick directions now evaluated independently. Earlier, diagonal thumbstick movement wouldn't be captured properly (directions were evaluated mutually exclusive).
* Adapted throttling to behave os-like. Earlier, throttling had a bug that caused it only to work for D-Pad navigation, not thumbsticks and buttons. Now, thumbstick navigation throttles the keyboard events in an os-like behavior: the first input is throttled 400ms, then with a shorter interval between repeats (100ms).
* Release debounce added. Sometimes controllers fire releases due to hardware constraints (e.g. if the user accidentaly releases the stick down beneath the threshold and moves it immediately afterwards again) - this feels way more natural and is linked to the new axis hysteresis (see below).
* Axis hyteresis added. This is basically "dead-zone" handling and won't trigger input beneath a certain degree of axis movement (also there's a small threshold band to prevent repeated "swapping" between detected and non-detected input)
* Axis smoothing added - prevents stick jitter between very low values (also for a smoother and more natural feeling)
* Button repeating locked. Buttons now don't repeat anymore (this is one of the main reasons of the bug #7834). Now buttons won't repeat, and their input is locked until they were released. If you hold A, you won't navigate multiple levels deep anymore.

### Issues

Fixes #7834 

### Code assistance

I used Github Copilot (Sonnet 4.6) to explain the existing code and debug the issue.  Once I found the problems, I gave it a few roundtrips to resolve it, but the code was so convoluted that it was hard to find a way to fix it without breaking something else. I let AI write a test that finally showed the issue. From that point on I carefully guided it to the rewrite the whole thing, then cleaned up the code. I think it's way easier to understand now.

---

* [x] I have provided a *substantive* review of [another web PR](https://github.com/jellyfin/jellyfin-web/pulls?q=is%3Apr+is%3Aopen+-label%3Adependencies+-label%3Ablocked+-label%3Abackend+-label%3Ainvalid+-label%3A"merge+conflict"+-label%3Astale+-author%3A%40me). (here: https://github.com/jellyfin/jellyfin-web/pull/7827#pullrequestreview-4131146616 hope this is substantial enough)


lmaotrigine:
It's difficult to give a line-by-line review of this, as much as I'd want to, due to the sheer size of the diff. From the contributing guide, PRs
> SHOULD represent a singular focus (i.e. a PR to fix a bug should not include unrelated refactoring).

the `src/scripts` directory is also filled with cruft, as indicated in the contributing guide. There's also an instruction to not create new files in there.

The title of this PR - along with the fact that it links to single other pull request (which should really be an issue instead) - suggests that it is a simple bugfix. If you want to include a refactor (which probably warrants more discussion before proceeding), this should be reworded to better explain the scope, and split into smaller changesets.

Please also review point 1. under [this heading of the LLM policies](https://jellyfin.org/docs/general/contributing/llm-policies#llm-code-contributions-to-official-projects).

lmaotrigine:
in any case, i can confirm this makes the left thumbstick behave similar to the D-Pad.
as for the buttons, I actually *expect* holding down A to fire multiple presses in any other case. i didn't find the delay to be noticeably short either. in my opinion, not repeating on long press is surprising behaviour; but ultimately, it is up to the maintainers.

DerGernTod:
thanks for the feedback! much appreciated.

ah sorry, i screwed up the issue numbers. i fixed it and linked to the correct ticket: #7834.

i agree that it's hard to review line by line, one needs to review the whole thing. it was just as hard for me to understand and honestly, i couldn't find a way to hack this fix into the existing code without refactoring everything. i basically wrote this from scratch and re-introduced the special behavior for electron (to not change anything in that direction, i'm not sure if electron is even used... i thought the desktop client actually uses something else).

that's also the reason why a singular change wasn't really possible for me - one of the reasons this bug exists is that dpad and sticks were implemented completely different, so the natural fix was refactoring these to behave the same, and that is already changing ~60% of the file (i would even argue the remaining code stayed rather similar - i could revert these cleanup changes to reduce the diff, but i think with the nature of this change, this wouldn't make a big difference for review efforts).

regarding the a/b button navigation: you're right that this can be opinionated, but for me in the current release it's unusable because if i press the A button just for a split second (like you would press it to "confirm" or "abort"), it triggers dozens of events. in the current state it's unusable for me. but i agree, it's probably worth a discussion if it should repeatedly trigger the event (similar to the dpad/sticks where if you hold longer, it first triggers just 1, then multiple events in quicker succession) or, like the current implementation does, only accepts the button once before it releases.

do you have any recommendations how i could improve the pr?

lmaotrigine:
> it was just as hard for me to understand and honestly, i couldn't find a way to hack this fix into the existing code without refactoring everything

that is understandable. this code is 10 years old, [vendored from here](https://github.com/microsoft/TVHelpers/blob/master/tvjs/src/GamepadToVK/gamepadtokey-1.0.0.0.js). it is definitely in serious need of modernisation. however, such efforts must be in line with the contributing guidelines. I've mentioned some relevant parts in my previous comment. Furthermore, for the issue you mentioned here, merely changing the number on [this line](https://github.com/jellyfin/jellyfin-web/blob/e0994bb68349cb1f917d0ede9c898f37659b5f02/src/scripts/gamepadtokey.js#L179) should be sufficient, i would think. feel free to correct me if i'm wrong.

>  one of the reasons this bug exists is that dpad and sticks were implemented completely different

judging from [here](https://github.com/jellyfin/jellyfin-web/blob/e0994bb68349cb1f917d0ede9c898f37659b5f02/src/scripts/gamepadtokey.js#L101) and further down, they are implemented exactly the same way. can you please clarify?

> regarding the a/b button navigation: you're right that this can be opinionated

in such cases, i would prefer this be configurable by the user. if this PR is going to be reworked as a refactor effort, it could be a part of that.

large changes like this should at the very least be broken into smaller commits, if not separate PRs, so that review can be more focussed.

as for the behaviour of the code itself, another thing that i noticed was that thumb stick navigation is now noticeably more sluggish when scrolling through a large enough item list. i do not want to undermine the original issue, but i personally use the thumb stick to speed scroll, and the D-Pad for more granular movement. i don't think this is unusual usage, and this PR would change how this works.

DerGernTod:
> merely changing the number on [this line](https://github.com/jellyfin/jellyfin-web/blob/e0994bb68349cb1f917d0ede9c898f37659b5f02/src/scripts/gamepadtokey.js#L179) should be sufficient

unfortunately it's not that easy. if you check [here](https://github.com/jellyfin/jellyfin-web/blob/e0994bb68349cb1f917d0ede9c898f37659b5f02/src/scripts/gamepadtokey.js#L217), you'll notice that the way throttling is triggered and reset is depending on lots of factors, one of which caused it not to throttle at all for the thumbsticks and buttons (which in the end was the root cause of the issue). 

> judging from [here](https://github.com/jellyfin/jellyfin-web/blob/e0994bb68349cb1f917d0ede9c898f37659b5f02/src/scripts/gamepadtokey.js#L101) and further down, they are implemented exactly the same way. can you please clarify?

i think didn't clarify this well. the code _looks_ the same, though dpad presses are considered buttons, while sticks are axis in the gamepad api. treating them the way they are (with further down i guess you mean [this function](https://github.com/jellyfin/jellyfin-web/blob/e0994bb68349cb1f917d0ede9c898f37659b5f02/src/scripts/gamepadtokey.js#L252)) means completely ignoring diagonal movement (else-if, one example), and ignoring the fact that one could move the stick from left to right without resetting them in the middle (e.g. moving left, then up, then right) - the stick state for left and up will stay true even if the user actually moves right. so i thought introducing some kind of abstraction before normalizing this would be a good idea.

> in such cases, i would prefer this be configurable by the user
agree, makes sense to make the A/B interaction configurable, but i think throttling still makes sense in the repeating scenario.

> i personally use the thumb stick to speed scroll, and the D-Pad for more granular movement.
i agree that stick and dpad navigation should be different than what the PR currently implements, i think reducing the throttling for the sticks should be sufficient.

do you think it makes sense to summarize all of our findings on the issue and close this PR for now until there's more feedback from maintainers? or should i try to clean up the PR (separate it into commits easier to review, fix the things we discussed)?

sorry if this is causing more work than it benefits, it's my first real open source contribution

lmaotrigine:
> unfortunately it's not that easy. if you check [here](https://github.com/jellyfin/jellyfin-web/blob/e0994bb68349cb1f917d0ede9c898f37659b5f02/src/scripts/gamepadtokey.js#L217), you'll notice that the way throttling is triggered and reset is depending on lots of factors, one of which caused it not to throttle at all for the thumbsticks and buttons (which in the end was the root cause of the issue).

i changed the threshold to 2000 to make it more obvious and found that both the thumbstick and the d-pad are both throttled, so i simply cannot reproduce your reported behaviour. i also cannot find anywhere in this code where either input is special cased.

> treating them the way they are (with further down i guess you mean [this function](https://github.com/jellyfin/jellyfin-web/blob/e0994bb68349cb1f917d0ede9c898f37659b5f02/src/scripts/gamepadtokey.js#L252)) means completely ignoring diagonal movement (else-if, one example), and ignoring the fact that one could move the stick from left to right without resetting them in the middle (e.g. moving left, then up, then right) - the stick state for left and up will stay true even if the user actually moves right.

this seems valid. in practice, though, in relation to the linked issue, i was only able to bypass throttling one single time using diagonal movement, ~~so it appears to not a big contributor to the large repeats you're experiencing (this is using my contrived large throttling mentioned above)~~ EDIT: was able to reproduce absence of throttling when going full circle on the thumbstick (was able to fix with a single line change) . however, i understand the rationale for this, and don't have objections as such. i only mention that it is unrelated to the title and the linked issue.

> do you think it makes sense to summarize all of our findings on the issue and close this PR for now until there's more feedback from maintainers?

i am not an arbiter of the validity of PRs themselves. that is up to the maintainers. you may leave this open until someone else gets around to reviewing further.

again, i don't mean to invalidate this wholesale, because i feel that rework of this is much needed. i only find it odd that it purports to be a bug fix, and i feel that it has far exceeded that scope.

at this point, it is probably wise to wait for a maintainer to advise further. i don't think i have anything new to add. however, i will test and provide feedback on any new changes to this branch.

DerGernTod:
now i'm really curious, because as i mentioned, in the state `main` currently is, controller navigation is unusable for me. maybe it's related to hardware? i have 4 xbox wireless controllers (the new ones), and all of them experience this behavior. may i ask which gamepad you're using?

lmaotrigine:
i'm using an old xbox 360 wireless controller. I'm not sure why I'm not able to reproduce most of this, but I am able to trigger some repeated presses when it shouldn't be happening (although I have to actively try) which this commit does fix.

dmitrylyzo:
Could it be that the Gamepad API in Edge (or Edge itself) sends a key press on its own, and we get a "doubling event"?

DerGernTod:
hmm i tried it both in edge and in the jellyfin desktop client... not sure if that uses edge internally on windows. that would mean though that gamepad controls would also work without the setting enabled, and that doesn't seem to be the case

PR #7842, opened 2026-04-18
Refactor Metadata editor

edave64:
I wanted to make some changes to the metadata editor (hopefully a future PR) and stumbled across the `setFieldVisibilities` which I feel has room for improvement. I spun this change into its own PR to not pollute the other one I'm working on with unrelated changes.

### Changes

I've changed blocks such as:

```ts
    if (item.Type === 'Series') {
        showElement('#fldStatus', context);
        showElement('#fldAirDays', context);
        showElement('#fldAirTime', context);
    } else {
        hideElement('#fldStatus', context);
        hideElement('#fldAirDays', context);
        hideElement('#fldAirTime', context);
    }
```

Into:

```ts
    toggleElement('#fldStatus, #fldAirDays, #fldAirTime', item.Type === BaseItemKind.Series, context);
```

- Making the code shorter and less redundant, since the selectors are not repeated.
- Consistently uses the `BaseItemKind` and `MediaType` constants, which were only used occasionally before.

### Issues
This PR shouldn't cause any behaviour changes and does not address any open Issues I'm aware of.

### Code assistance
Claude Sonnet 4.6 has been used for code review before submitting, but all code and this text has been written by me.

---

* [x] I have provided a *substantive* review of https://github.com/jellyfin/jellyfin-web/pull/7843#pullrequestreview-4136407478.


src/components/metadataEditor/metadataEditor.js
```diff
@@ -500,165 +501,76 @@ function loadExternalIds(context, item, externalIds) {
...
-        }
-        hideElement('#fldCustomRating', context);
-    } else {
-        showElement('#fldCommunityRating', context);
-        showElement('#genresCollapsible', context);
-        showElement('#studiosCollapsible', context);
-        showElement('#fldOfficialRating', context);
-        showElement('#fldCustomRating', context);
-    }
+    toggleElement('#fldOfficialRating', item.Type === 'TvChannel', context);
```
ocp1006: The original logic was:
- **Person / Genre / Studio / MusicGenre** → hidden
- **TvChannel** → shown
- **Everything else (Movie, Series, Episode, …)** → shown

This line shows it only for `TvChannel` and hides it for all other types, including normal media. The Official Rating field will silently disappear for movies, series, episodes, etc.

Suggested fix:
```js
toggleElement('#fldOfficialRating',
    ![BaseItemKind.Person, BaseItemKind.Genre, BaseItemKind.Studio, BaseItemKind.MusicGenre].includes(item.Type),
    context);
```

Also a nit: this line still uses the string `'TvChannel'` instead of `BaseItemKind.TvChannel`, inconsistent with the rest of the PR.

src/components/metadataEditor/metadataEditor.js
```diff
@@ -500,165 +501,76 @@ function loadExternalIds(context, item, externalIds) {
...
-        hideElement('#fldPremiereDate', context);
-        hideElement('#fldDateAdded', context);
-        hideElement('#fldYear', context);
-    } else {
-        showElement('#metadataSettingsCollapsible', context);
-        showElement('#fldPremiereDate', context);
-        showElement('#fldDateAdded', context);
-        showElement('#fldYear', context);
-    }
+    toggleElement('#metadataSettingsCollapsible, #fldPremiereDate, #fldDateAdded, #fldYear',
```
ocp1006, line 569: Nit: `'TvChannel'` string literal here should be `BaseItemKind.TvChannel` - same goal as the rest of this PR.

src/components/metadataEditor/metadataEditor.js
```diff
@@ -500,165 +501,76 @@ function loadExternalIds(context, item, externalIds) {
...
-function hideElement(selector, context, multiple) {
-    context = context || document;
-    if (typeof selector === 'string') {
-        const elements = multiple ? context.querySelectorAll(selector) : [context.querySelector(selector)];
+/**
+ * Sets the "hide" class to the elements matching the selector
+ * @param {string} selector - Css selector to search with
+ * @param {Element} [context] - Element to search in, defaults to document
+ */
+function hideElement(selector, context) {
```
ocp1006, line 509: The original `hideElement`/`showElement` accepted either a CSS selector string *or* a raw DOM element (`typeof selector === 'string'` branch). The new signatures only accept strings. Worth double-checking that no callers outside this file pass a raw element - if they do, the toggle will silently do nothing.

edave64: All uses of this function I could find only pass CSS selectors, which is why I removed the functionality.

The functions aren't exported, and no functions `hideElement` or `showElement` are ever called outside this file.

src/components/metadataEditor/metadataEditor.js
edave64: Thanks! I was thinking about why the conditions were so strangely nested, but came to the entirely wrong conclusion about the behavior. 😅

Guess that makes it all the better to unify the conditions like this.

Also, it's not a nitpick, since using the constants was an explicit goal of this PR. Fixed

src/components/metadataEditor/metadataEditor.js
edave64: Fixed

09bravo (approved):
I have reviewed the changes. This PR simplifies the codebase without affecting functionality, and all the issues have been already been noted and fixed. I approve.

PR #7843, opened 2026-04-19
Added a button to switch between paginated and scrolled mode in the book reader

09bravo:
### Changes
Before: 
image
After (default in paginated mode)
image
the message
image
and finally scrolled mode
image

### Issues
not any I am aware of, but this is a problem I had myself and I wanted to read in scroll mode

### Code assistance
Some help from claude (mostly explanation of code base) explening why JS is throwing errors at me (I am a rust developer not a javascript developer, although I didn't have many problems and almost no problem with css and html since I use those but the logic was all on me (and for what is worth I picked the icon myself from https://fonts.google.com/icons) 

---

* [X] I have provided a *substantive* review of [Refactor Metadata editor#7842](https://github.com/jellyfin/jellyfin-web/pull/7842).


src/plugins/bookPlayer/plugin.js
```diff
@@ -1,3 +1,4 @@
+import toast from '../../components/toast/toast';
```
edave64: This import should probably be in the segment with the other `../../component` imports

src/plugins/bookPlayer/plugin.js
```diff
@@ -257,7 +262,10 @@ export class BookPlayer {
             this.tocElement = new TableOfContents(this);
         }
     }
-
+    toggleScroll() {
+        this.scroll = !this.scroll;
+        toast(`Reading mode changed to ${this.scroll ? 'scrolled' : 'paginated'}. Please reopen the book to apply`);
```
edave64: What exactly is meant by "reopen the book"? The scroll property isn't persisted as far as I can see and defaults to false again, so it can't be reloading the page?

If that reload is required, can't we do it for the user?

Either way, this is clunky. Calling `this.rendering?.flow(this.scroll ? 'scrolled' : 'paginated');` seems to switch these modes at runtime without needing to reload anything.

(The logic below and the `this.scroll` variable should still be kept in case this method runs before `this.rendering` is initialized, or ever reinitializes)

Also the string isn't translatable. If you look at other uses of `toast`, it's always written as `toast(globalize.translate('[UniqueKey]'))`, with a corresponding translation in `src/strings`.

All the more reason not to ask the user to do things :P

src/plugins/bookPlayer/plugin.js
```diff
@@ -374,12 +382,11 @@ export class BookPlayer {
                 const api = toApi(ServerConnections.getApiClient(item));
                 const downloadHref = getLibraryApi(api).getDownloadUrl({ itemId: item.Id });
                 const book = epubjs(downloadHref, { openAs: 'epub' });
-
+                const flow = (this.scroll) ? "scrolled":"paginated";
```
edave64: ESLint doesn't like the lack of space around the colon. Can be automatically fixed

edave64 (changes requested):
The core idea is correct, and given the old `//TODO` entry clearly intended at some point.

But it really should happen automatically without additional user action and currently I can't get it to work for me at all, because I don't understand how to reload the book without resetting the variable.

src/plugins/bookPlayer/plugin.js
09bravo: ~Will fix this in the next commit~ I removed the toast as it is no longer necessary

src/plugins/bookPlayer/plugin.js
09bravo: Hello, Thanks for the feedback, I will fix the toast part, also what I meant by reopening the book is either pressing escape or clicking the x button, ~I wasn't aware about this.rendering I will try it out~ update: it turns out the function is called this.rendition not this.rendering.

src/plugins/bookPlayer/plugin.js
09bravo: I will fix it

src/plugins/bookPlayer/plugin.js
09bravo: I have fixed all of the issues and pushed the code :)

edave64: You're right. Not sure why I didn't just copy the code I already made and retyped it with a typo instead 😅

I'll recheck the branch once I get back from work, than approve if it works correctly.

src/plugins/bookPlayer/plugin.js
edave64: I recommend running `npm run lint` in the console to run the linter and see if it complains about something else. (In your code, no need to fix anything in outside files)

I guess it also wouldn't like the double quotes.

I have my IDE configured to run/fix eslint automatically, so I don't see these errors unless I carefully look 😅

src/plugins/bookPlayer/plugin.js
```diff
@@ -382,7 +381,7 @@ export class BookPlayer {
                 const api = toApi(ServerConnections.getApiClient(item));
                 const downloadHref = getLibraryApi(api).getDownloadUrl({ itemId: item.Id });
                 const book = epubjs(downloadHref, { openAs: 'epub' });
-                const flow = (this.scroll) ? "scrolled":"paginated";
+                const flow = (this.scroll) ? "scrolled" : "paginated";
```
edave64: Ok, so I just checked, eslint really does complain about the double quotes, too.
I don't quite know what the process in the repo is, if the maintainer would just run "npm run -- --fix" on their own, but I'd suggest you change these quotes to single quotes.
But from a logic perspective, this runs as it should, so I'll mark it as approved.

edave64 (approved):

src/plugins/bookPlayer/plugin.js
09bravo: I should probably run ```npm run -- --fix``` when I get home (or change it to single quotes manually). 
Anyway thanks for all the help (especially ```this.rendition ``` I wouldn't have figured it out on my own)

edave64 (approved):

PR #7848, opened 2026-04-22
migrate book OSD to react component

dkanada:
### Changes

Book controls are migrated to a React component and the UI more closely matches the video controls. Next steps would be moving more logic into the element (settings, TOC management, maybe swipe and tap controls) and using it in the PDF and comic players as well.

- title is shown in the top element
- close button changed to back navigation
- overlay above book content
- fade transition after two seconds

### Issues

None

### Code assistance

General research and debugging the SCSS error message.

### DIFF
--- a/src/plugins/bookPlayer/plugin.js
+++ b/src/plugins/bookPlayer/plugin.js
@@ -6,15 +6,16 @@ import browser from 'scripts/browser';
 import TouchHelper from 'scripts/touchHelper';
 import { toApi } from 'utils/jellyfin-apiclient/compat';
 
-import layoutManager from '../../components/layoutManager';
 import loading from '../../components/loading/loading';
 import keyboardnavigation from '../../scripts/keyboardNavigation';
 import dialogHelper from '../../components/dialogHelper/dialogHelper';
 import TableOfContents from './tableOfContents';
+import BookOsd from './BookOsd/BookOsd';
 import { translateHtml } from '../../lib/globalize';
 import * as userSettings from '../../scripts/settings/userSettings';
 import { PluginType } from '../../types/plugin.ts';
 import Events from '../../utils/events.ts';
+import { renderComponent } from '../../utils/reactUtils';
 
 import 'material-design-icons-iconfont';
 import '../../elements/emby-button/paper-icon-button-light';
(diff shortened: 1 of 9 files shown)


MBR-0001:
image
the new UI covers the content, was this intended?
I know it goes away after a few seconds but still, seems a bit annoying

dkanada:
@MBR-0001 it was intended, and I originally planned to address your comment with some behavior changes to toggle visibility (rather than always trigger showOsd) for onPointerUp events. Ran into issues with the swipe listener though, so I've removed it for the time being.

src/plugins/bookPlayer/plugin.js
```diff
@@ -365,10 +325,6 @@ export class BookPlayer {
             }
         };
 
-        if (!Screenfull.isEnabled) {
```
MBR-0001, line 368: Seems this part was lost? I don't see any new code that hides the fullscreen button

MBR-0001 (changes requested):
Looks a tiny bit odd in landscape mode
[Screencast_20260526_231654.webm](https://github.com/user-attachments/assets/dfa572aa-8b44-4a90-b730-eebced2dd366)



Other than that, much much better than the last iteration

src/plugins/bookPlayer/plugin.js
dkanada: Added conditional display to several buttons including this one and tweaked landscape. Long term I'd rather just enforce portrait on mobile devices though.

src/plugins/bookPlayer/plugin.js
```diff
@@ -63,12 +63,13 @@ export class BookPlayer {
...
 
         loading.show();
-        const elem = this.createMediaElement();
+        const elem = this.createMediaElement(options);
         return this.setCurrentSrc(elem, options);
     }
 
     stop() {
         this.unbindEvents();
+        this.unmountBookOsd();
```
dfederm, line 72: Consider `?.` here since `unmountBookOsd` is assigned only inside `createMediaElement`. It's probably unlikely but does match other defensive checks below (eg `mediaElement` is also only assigned in `createMediaElement`).

MBR-0001 (approved):

thornbill (commented):
This looks great! The only thing I noticed is the buttons lack titles.

thornbill (approved):

PR #7864, opened 2026-04-28
Move bitrate detection to web

viown:
Moves the bitrate detection logic from the legacy apiclient to web. No change in detection behavior is made, other than an additional guard to prevent more than one test from running at the same time.

Moving the bitrate detection behavior to web paves the way to eventually revising how the bitrate calculation behavior is done

### Changes
Move bitrate detection to web

### Issues
N/A

### Code assistance
None

---

### DIFF
--- a/src/utils/bitrateTest.ts
+++ b/src/utils/bitrateTest.ts
@@ -145,6 +145,7 @@ const detectBitrateInternal = (api: Api, tests: BitrateTest[], index: number, cu
     const test = tests[index];
 
     return getDownloadSpeed(api, test.bytes).then(
+        // eslint-disable-next-line sonarjs/function-return-type
         (bitrate) => {
             if (bitrate < test.threshold) {
                 return normalizeReturnBitrate(bitrate);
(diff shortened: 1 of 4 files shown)


src/utils/bitrateTest.ts
thornbill, line 1: Was this more or less copied from the apiclient?

viown: Only partially. I re-wrote almost all of it by hand to make sure it works with the TS SDK.

thornbill: I was just curious. It doesn't really matter either way. 😄

src/utils/bitrateTest.ts
```diff
@@ -0,0 +1,224 @@
...
+    if (index >= tests.length) {
+        return Promise.resolve(normalizeReturnBitrate(currentBitrate || 0));
+    }
+
+    const test = tests[index];
+
+    return getDownloadSpeed(api, test.bytes).then(
+        (bitrate) => {
+            if (bitrate < test.threshold) {
+                return Promise.resolve(normalizeReturnBitrate(bitrate));
```
thornbill: Since we're already in a promise chain here, returning `Promise.resolve` doesn't seem necessary (as sonar flagged).

viown: I think I originally added this because Sonar complained about mismatched types

thornbill: Well that's dumb. I would just add a comment to ignore the "error".

thornbill (approved):

PR #7867, opened 2026-05-01
Promote delete to top-level action and fix post-delete UI refresh

xanderphillips:
### Changes

Delete was previously only accessible through the More (...) overflow menu,
requiring several extra clicks for a very common activity.
This brings it up into the primary action row alongside
Info, Mark Played, and Favorites, conditioned on `item.CanDelete` so it
only appears where the user has permission. Delete remains available in
the More menu as well, both paths exist.

Covers the Season episode list (both the legacy list view and the React
`ListViewUserDataButtons` path) and the item detail page (movies, episodes,
etc.). 

Added `ItemAction.Delete` to the `ItemAction` enum and `enableDeleteButton` 
to `ListOptions`, following the existing opt-in pattern
used by `enablePlayedButton`, `enableRatingButton`, etc. 

Added a `DeleteIconButton` React component following the `InfoIconButton` /
`MoreVertIconButton` pattern.

This also fixes how deleted items required a manual refresh to go away. 
The deleted item now disappears immediately on confirmation
regardless of which delete path was used. 

Wires `notifyRefreshNeeded` on the season episode container to re-call 
`renderChildren` on the API matching the pattern used by collection containers.

Suppresses an unhandled promise rejection when the user cancels the
confirmation dialog.

### Issues

N/A

### Code assistance

I used Claude Code to assist with this implementation to ensure I didn't
miss anything upstream or break compatibility with existing patterns
outside of the specific changes I was working on.

---

- [X] I have read and followed the [contributing guidelines](https://github.com/jellyfin/jellyfin-web/blob/master/CONTRIBUTING.md).
- [X] I have tested these changes.
- [X] I have verified that this is not duplicating changes in an existing PR.
- [X] I have provided a *substantive* review of [another web PR](https://github.com/jellyfin/jellyfin-web/pulls?q=is%3Apr+is%3Aopen+-label%3Adependencies+-label%3Ablocked+-label%3Abackend+-label%3Ainvalid+-label%3A"merge+conflict"+-label%3Astale+-author%3A%40me).

### DIFF
--- a/src/components/common/DeleteIconButton.tsx
+++ b/src/components/common/DeleteIconButton.tsx
@@ -1,0 +1,24 @@
+import React, { type FC } from 'react';
+import IconButton from '@mui/material/IconButton';
+import DeleteIcon from '@mui/icons-material/Delete';
+
+import { ItemAction } from 'constants/itemAction';
+import globalize from 'lib/globalize';
+
+interface DeleteIconButtonProps {
+    className?: string;
+}
+
+const DeleteIconButton: FC<DeleteIconButtonProps> = ({ className }) => {
+    return (
+        <IconButton
+            className={className}
+            data-action={ItemAction.Delete}
+            title={globalize.translate('Delete')}
+        >
+            <DeleteIcon />
+        </IconButton>
+    );
+};
+
+export default DeleteIconButton;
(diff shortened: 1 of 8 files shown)


thornbill:
Deleting should not be a top level option.

PR #7876, opened 2026-05-05
Replace deprecated DOMImplementation.hasFeature in scroller

Thomas05000005:
### Changes

The `wheelEvent` constant in `src/lib/scroller/index.js` was using `document.implementation.hasFeature('Event.wheel', '3.0')` to pick between the modern `wheel` event and the legacy `mousewheel` event. That API has been deprecated since DOM Level 4 and modern browsers always make it return `true`, so the legacy fallback was effectively dead code.

I replaced the check with `'onwheel' in document.documentElement`, which is the standard feature-detection pattern for events and actually returns `false` on browsers that genuinely don't support the `wheel` event. Behavior on modern browsers is unchanged; older user agents that don't expose `onwheel` now correctly fall back to `mousewheel`.

```diff
-const wheelEvent = (document.implementation.hasFeature('Event.wheel', '3.0') ? 'wheel' : 'mousewheel');
+const wheelEvent = ('onwheel' in document.documentElement ? 'wheel' : 'mousewheel');
```

### Issues

None directly. This resolves the `@typescript-eslint/no-deprecated` warning reported by `npm run lint` for `src/lib/scroller/index.js:76`.

### Code assistance

Used Claude (LLM) to scan the project for deprecated APIs flagged by ESLint and to verify that `'onwheel' in document.documentElement` is the safe replacement for `hasFeature` given the project's `browserslist`.

---

### DIFF
--- a/src/lib/scroller/index.js
+++ b/src/lib/scroller/index.js
@@ -73,7 +73,7 @@ function within(number, num1, num2) {
 // Other global values
 const dragMouseEvents = ['mousemove', 'mouseup'];
 const dragTouchEvents = ['touchmove', 'touchend'];
-const wheelEvent = (document.implementation.hasFeature('Event.wheel', '3.0') ? 'wheel' : 'mousewheel');
+const wheelEvent = ('onwheel' in document.documentElement ? 'wheel' : 'mousewheel');
 const interactiveElements = ['INPUT', 'SELECT', 'TEXTAREA'];
 
 const scrollerFactory = function (frame, options) {


Thomas05000005:
Closing this one — I missed the `NOTE: This file should not be modified` comment at the top of [src/lib/scroller/index.js](https://github.com/jellyfin/jellyfin-web/blob/master/src/lib/scroller/index.js#L1-L4) when I opened the PR. The deprecated `DOMImplementation.hasFeature` warning will go away naturally when the legacy scroller is replaced. I am opening a tracking issue so the suggested patch is not lost. Sorry for the noise.

PR #7879, opened 2026-05-05
Fix blurry card images on high-DPI displays

Maxr1998:
I was trying out Jellyfin Unstable 10.12 (2026050514) and noticed that all image cards were suddenly blurry compared to my main server that's still on 10.10.7.
Specifically, the URLs used for images were halved compared to those used on 10.10.7.

Looking through the git history, I identified 8924ea487df6ade72ce0e5a7b49b9b4a37f01ae5 as a possible cause. With the help of Claude Sonnet 4.6, I found the reason and a potential fix for this issue. Running the fixed Jellyfin Web locally confirms the images are sharp again. However, I'm unsure if the location I am (well, Claude is) applying the fix to is optimal, so please take it with a grain of salt and modify the PR if necessary.

### Changes
`getItemImageUrlById` does not apply device pixel ratio scaling, unlike the legacy `apiClient.getScaledImageUrl` which multiplied dimensions by `window.devicePixelRatio` automatically. Scale the fill dimensions explicitly to restore full-resolution images on HiDPI screens.

### Code assistance
I used Claude Sonnet to find the cause after noticing this bug on my own test server. See above for more details.

---

### DIFF
--- a/src/apps/dashboard/features/libraries/components/LibraryCard.tsx
+++ b/src/apps/dashboard/features/libraries/components/LibraryCard.tsx
@@ -44,9 +44,10 @@ const LibraryCard = ({ virtualFolder }: LibraryCardProps) => {
 
     const imageUrl = useMemo(() => {
         if (virtualFolder.PrimaryImageItemId && virtualFolder.ItemId && api) {
+            const dpr = window?.devicePixelRatio || 1;
             return getImageApi(api)
                 .getItemImageUrlById(virtualFolder.ItemId, ImageType.Primary, {
-                    maxWidth: Math.round(dom.getScreenWidth() * 0.40)
+                    maxWidth: Math.round(dom.getScreenWidth() * dpr * 0.40)
                 });
         }
     }, [ api, virtualFolder ]);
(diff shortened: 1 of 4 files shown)


thornbill:
This is probably the most noticeable, but should other usages of `getItemImageUrlById` that pass dimensions also be updated?

Maxr1998:
I'm honestly unsure, I will have to take another look at the code. But I agree that all of these should be fixed in a single change/PR, if applicable.

Maxr1998:
I looked through the code and found a few more uses of `getItemImageUrlById` that need to be updated.

The main ones were those in `listHelper.ts`. Like the legacy `listview.js`, these use hardcoded sizes of 80 and 500 (in `useList`), but don't scale them by the `devicePixelRatio` like before (the legacy implementation also used `getScaledImageUrl`).

The same applies to `cardHelper.ts` / `cardBuilder.js`. There, the (footer?) `logoHeight` is hardcoded to 40 in both cases but not scaled anymore in the new implementation.

Another one I found is in `LibraryCard.tsx`. It is not actually noticeable there because its image width is set to `0.40 * dom.getScreenWidth()`, even though the actual size on screen is much smaller than 40%. Thus, even at “half” the render size (`getScreenWidth()` returns unscaled pixels), it was fine. Nonetheless, I applied the new scaling to it, but I'd argue that “0.40 * screen width” is messy in the first place and should be fixed eventually.

thornbill (approved):

PR #7896, opened 2026-05-08
Replace deprecated UserDto.HasConfiguredPassword with HasPassword

Thomas05000005:
### Changes

`UserPasswordForm` was reading `user.HasConfiguredPassword` to decide whether to show the "current password" and "reset password" fields. That property is marked `@deprecated` in the Jellyfin TypeScript SDK's `UserDto` model. The non-deprecated sibling `HasPassword` (defined just above on the same model) carries the same boolean signal — whether the account currently has a password set — so I switched the read to `user.HasPassword` and updated the inline comment that referenced the old name.

```diff
-        if (user.HasConfiguredPassword) {
+        if (user.HasPassword) {
             ...
-                // This should only happen when user.HasConfiguredPassword is false, but this information is not passed on
+                // This should only happen when user.HasPassword is false, but this information is not passed on
```

### Issues

None directly. This resolves the `@typescript-eslint/no-deprecated` warning reported by `npm run lint` for `src/components/dashboard/users/UserPasswordForm.tsx:35`.

### Code assistance

Used Claude (LLM) to scan the SDK's UserDto model for the non-deprecated successor of HasConfiguredPassword and to confirm the semantics line up with how the form uses the value (decide whether to render password-change fields).

---

* [ ] I have provided a substantive review of another web PR.

### DIFF
--- a/src/components/dashboard/users/UserPasswordForm.tsx
+++ b/src/components/dashboard/users/UserPasswordForm.tsx
@@ -32,7 +32,7 @@ const UserPasswordForm: FunctionComponent<IProps> = ({ user }: IProps) => {
 
         (await libraryMenu).setTitle(user.Name);
 
-        if (user.HasConfiguredPassword) {
+        if (user.HasPassword) {
             if (!user.Policy?.IsAdministrator) {
                 (page.querySelector('#btnResetPassword') as HTMLDivElement).classList.remove('hide');
             }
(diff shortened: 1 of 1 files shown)


Thomas05000005:
Closing this — I checked the Jellyfin SDK and **both** `HasConfiguredPassword` **and** `HasPassword` are marked `@deprecated` on `UserDto`:

```ts
// node_modules/@jellyfin/sdk/lib/generated-client/models/user-dto.d.ts
'HasPassword'?: boolean | null;            // @deprecated
'HasConfiguredPassword'?: boolean | null;  // @deprecated
'HasConfiguredEasyPassword'?: boolean | null; // @deprecated
```

I confirmed locally that after the swap the ESLint warning just moves from `HasConfiguredPassword` to `HasPassword` (lint output identical, only the name in the warning changes). So this PR doesn't actually fix the deprecation — it just shifts it.

A real fix needs to know what the non-deprecated way to ask "does the user have a password set?" is on the API now (or whether the password-state UI in `UserPasswordForm` should use a different mechanism altogether). I'm opening a tracking issue to ask the maintainers.

PR #7907, opened 2026-05-10
Use innerHTML instead of replaceChildren to empty div

mrxz:
### Changes
The itemDetails page is using `Element.replaceChildren` since #7656. Older browsers don't support this ([caniuse](https://caniuse.com/mdn-api_element_replacechildren)) resulting in an exception that prevents the page from properly loading. This PR changes the usage of `replaceChildren` with an `.innerHTML = ''` as the purpose is to clear all children of the `itemDetailsGroup` div.

### Issues

N/A

### Code assistance

None

---

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -71,6 +71,13 @@ export default tseslint.config(
             'no-redeclare': 'off',
             '@typescript-eslint/no-redeclare': ['error', { builtinGlobals: false }],
             'no-restricted-globals': ['error'].concat(restrictedGlobals),
+            'no-restricted-properties': [
+                'error',
+                {
+                    property: 'replaceChildren',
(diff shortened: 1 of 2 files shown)


dmitrylyzo (approved):
Babel (core-js) doesn't seem to polyfill `replaceChildren`. We could do this ourselves, but cleaning with `innerHTML = ''` is portable and doesn't require new code.

thornbill:
Why is linting not failing for this?

UPDATE: Related issue: https://github.com/amilajack/eslint-plugin-compat/issues/526

thornbill:
Let's add a `no-restricted-properties` eslint rule to prevent any usages of `replaceChildren` as a property.

https://eslint.org/docs/latest/rules/no-restricted-properties

thornbill (approved):

Issue #7651, opened 2026-03-03
[Bug]: Chrome DeviceProfile hardcodes MKV in DirectPlayProfiles — causes silent 0ms playback failure for codec-compatible MKV files


xiNeRoar:
### This issue respects the following points:

- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-web/issues?q=is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

Jellyfin Web 10.11.6 includes `mkv` in the `DirectPlayProfiles` sent to the server in the `PlaybackInfo` POST request when using Chrome. Chrome's HTML5 `<video>` element does not support the MKV (Matroska) container for direct playback.

When an MKV file contains codecs that are otherwise fully compatible with Chrome (e.g. H.264 + AAC), the server selects DirectPlay and serves the raw `.mkv` file directly. Chrome cannot play it, and playback fails silently at 0ms — no error message in the UI, no ffmpeg invocation, no error in server logs.

This contradicts a prior maintainer decision. In PR #2289 (Jan 2021), nyanmisaka attempted to enable MKV DirectPlay for Chrome, but it was closed with the reasoning: "Chrome supports WebM but not MKV proper. The two standards are different and, while it may work for some files, it won't for others." Despite this, MKV is present in Chrome's DirectPlayProfiles in 10.11.6.

Additionally, the user setting `preferFmp4HlsContainer` does not prevent this. It was confirmed saved server-side as `"true"` via the DisplayPreferences API, but this setting only affects TranscodingProfiles (choosing fMP4 vs TS for HLS segments) — it does not remove MKV from the DirectPlay path.

Related issues with the same or similar symptoms (none identified the root cause):
- jellyfin-web#3690 — "Playback error when playing mkv files without transcoding on version 10.8.0"
- jellyfin#12204 — "Can't play mkv videos in browsers"
- jellyfin#11298 — DirectPlay MKV "Stopped at 0 ms"

### Reproduction Steps

1. Have an MKV file with H.264 (High profile, Level 4.1) video and AAC (LC, stereo) audio — no codec transcoding required for Chrome
2. Open Jellyfin Web in Chrome and play the file
[...]

PR #7913, opened 2026-05-12
Handle media element errors before playback starts

### LINKED ISSUE #7912: DirectPlay startup errors can leave the client stuck on backdrop

### Describe the bug

Jellyfin Web can get stuck on the backdrop when native DirectPlay fails before playback actually starts.

This is related to #7651, but focuses on the error handling after DirectPlay has already been selected.

On an affected system, Chromium fails during media startup:

```
PipelineStatus::PIPELINE_ERROR_DECODE: VaapiVideoDecoder: failed Initialize()ing the frame pool
```

The `<video>` element exposes the failure:

```
error.code = 3 // MEDIA_ERR_DECODE
readyState = 1
currentTime = 0
```

However, the startup failure does not show “Playback failed” and does not retry with transcoding. The UI remains stuck on the backdrop and can only be recovered by refreshing the page or opening the web client in a new tab.

### Reproduction Steps

1. Linux Chromium/Chrome with VA-API hardware decoding enabled.
2. Play an MKV/HEVC item that Jellyfin selects for DirectPlay.
3. Chromium fails during startup with `VaapiVideoDecoder: failed Initialize()ing the frame pool`, and `<video>.error.code` becomes `3 (MEDIA_ERR_DECODE)`.
4. Jellyfin Web stays stuck on the backdrop instead of showing an error or retrying with transcoding.

### Expected/Actual behaviour

Startup media errors should use the existing playback error handling. If transcoding is available, retry with transcoding; otherwise it should show the existing playback failure message.

The startup media error is exposed on the `<video>` element, but it does not reach the existing playback error handling. Jellyfin Web remains stuck on the backdrop.

### Logs

_No response_

### Server version

10.11.8

### Web version

10.11.8

### Build version

10.11.8

### Platform

Linux

### Browser

Chrome

### Additional information

_No response_

skrmc:
### Changes
Handle media element errors that happen during playback startup, before `play()` resolves.

DirectPlay can fail after the source is assigned but before Jellyfin Web has attached the normal media error listener. In that case the `<video>` element may already expose `MediaError.code`, but the error never reaches the existing playback error path.

This change listens for startup media errors inside `playWithPromise` and rejects the startup play promise instead of calling the normal media error handler directly. That keeps startup failures on existing `playbackmanager` handling for clearing startup state and retrying with transcoding or showing the playback error.

### Issues
Fixes #7912. Related to #7651.

### Code assistance
None.

---

### DIFF
--- a/src/components/htmlMediaHelper.js
+++ b/src/components/htmlMediaHelper.js
@@ -195,27 +195,60 @@ function onSuccessfulPlay(elem, onErrorFn) {
     elem.addEventListener('error', onErrorFn);
 }
 
-export function playWithPromise(elem, onErrorFn) {
+function getMediaElementError(elem) {
+    return elem.error || new Error('Media element error');
+}
+
+function isIgnoredPlayError(err) {
+    const errorName = (err?.name || '').toLowerCase();
+
+    // safari uses aborterror
+    return errorName === 'notallowederror'
+        || errorName === 'aborterror';
+}
+
+function playElement(elem) {
     try {
-        return elem.play()
-            .catch((e) => {
-                const errorName = (e.name || '').toLowerCase();
-                // safari uses aborterror
-                if (errorName === 'notallowederror'
-                        || errorName === 'aborterror') {
-                    // swallow this error because the user can still click the play button on the video element
+        return Promise.resolve(elem.play())
+            .catch((err) => {
+                if (isIgnoredPlayError(err)) {
+                    // swallow this error because the user can still click the play button on the media element
                     return Promise.resolve();
                 }
-                return Promise.reject(e);
-            })
-            .then(() => {
-                onSuccessfulPlay(elem, onErrorFn);
-                return Promise.resolve();
+
+                throw err;
             });
     } catch (err) {
         console.error('error calling video.play: ' + err);
-        return Promise.reject();
+        throw err;
     }
+}
+
+export function playWithPromise(elem, onErrorFn, options = {}) {
+    let cleanupStartupError;
+
+    const startupErrorPromise = new Promise((resolve, reject) => {
+        const onStartupError = () => reject(getMediaElementError(elem));
+
+        cleanupStartupError = () => {
+            elem.removeEventListener('error', onStartupError);
+        };
+
+        elem.addEventListener('error', onStartupError);
+    });
+
+    const playPromise = Promise.resolve()
+        .then(options.beforePlay)
+        .then(() => playElement(elem));
+
+    return Promise.race([playPromise, startupErrorPromise])
+        .then(() => {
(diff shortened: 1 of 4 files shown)


skrmc:
@thornbill Was this closed because the substantive review checkbox was unchecked?

thornbill:
No. Sections of the PR template were omitted entirely.

skrmc:
@thornbill Apologies, I've included all template sections. Let me know if anything else is missing.

moontwister (commented):
Tested this on webOS (LG C4) and in a desktop browser.

**It does what it says, where the element reports the failure.** I built a file with intact MP4 metadata but corrupted media data, so the server hands it over untouched and the decoder gets garbage. In a desktop browser without this patch, the element fires `error` but nothing surfaces — the spinner just runs forever. With the patch, the same file produces an error message. That is the bug described here, and this fixes it.

**On webOS this file's failure isn't reported at all.** Same file, same server: the element never fires `error` and `MediaError` is never set — it stays null while the app hangs. There is nothing for the new listener to catch in that case, so that hang has a different cause. Not an argument against this change.

**No regressions** on webOS: end-of-episode auto-advance, manual skip to next, rapid skipping before the picture settles, and aborting during buffering all worked correctly. It cherry-picks cleanly onto current master, and lint, tests and build pass.

**Two questions on `getMediaError`:**

- `MEDIA_ERR_ABORTED` (code 1) isn't mapped and falls through to `PLAYER_ERROR`. Intentional?
- The switch reads `code` before `name`, and `DOMException` carries a legacy `code` too (`NotSupportedError` is 9, `AbortError` is 20). Nothing collides today, but the intent isn't obvious from the code.

Minor: the synchronous `catch` around `elem.play()` used to `return Promise.reject()` with no argument, so a synchronous throw lost its error. `throw err` fixes that.

moontwister:
Correction to my earlier comment — I said webOS can't benefit from this, and that was wrong.

I had only tried a file with corrupted media data, where the element never reports anything at all. With a file the server can't read (the stream URL 500s), webOS does report it, and your listener catches it:

```
error                         MediaError.code = 4
startup error listener fired  code 4
race rejected                 4
onPlaybackError               MEDIA_NOT_SUPPORTED
```

So the change works there, on webOS too.

The app still ends up stuck in that scenario, but only past this point: `onPlaybackError` retries with transcoding rather than surfacing anything, and on the third attempt `play()` resolves while the element is still at `networkState 2 / readyState 0` with no data — so playback counts as started and the spinner never ends. Nothing for this PR to fix; I just wanted to correct what I claimed earlier.

thornbill:
@moontwister All communications must be human authored per our llm policy.

PR #7915, opened 2026-05-13
Remove LAN check in automatic bitrate detection

### LINKED ISSUE #6535: Auto bandwidth selection is broken when connecting to a server on the same LAN

### Describe the bug

When connecting to a server on the same LAN, the auto bandwidth is always 120mbps regardless of actual speed.
This can be problematic when using VPNs or on clients with bad Wifi reception.

### Reproduction Steps

1. Have your client and the server be on the same LAN. A VPN will work, as long as the IP punched in is the same LAN as the server.
2. Start playing video
3. Open the gear icon and see auto bandwidth = 120mbps

### Expected/Actual behaviour

Expected:
The client should select a bitrate based on the actual bandwidth between the server and the client

Actual:
The client selects 120mbps every time and high bitrate video stutters.

### Logs

_No response_

### Server version

10.10.3, Unstable

### Web version

10.10.3, Unstable

### Build version

10.10.3, Unstable

### Platform

macOS

### Browser

Firefox

### Additional information

_No response_

viown:
Currently, there is an [enforced 140Mbps minimum](https://github.com/jellyfin/jellyfin-web/blob/b2fc95df1b723bda9d486968a2d29d55214f3a08/src/utils/bitrateTest.ts#L8) for automatic bitrate detection in LAN networks. Clients connecting over WiFi may experience buffering if they play media their connection cannot handle.

This PR removes that. The current consensus with removing this check is that it can risk unnecessary transcoding under possibly rare circumstances. Counterpoint: Better to risk transcoding than risk unplayable media as this would only apply to the 'Auto' option.

Alternative methods related to testing the video playback directly can also be looked at, although a lot more complex.

### Changes
Removes the LAN bitrate minimum from the automatic bitrate detection

### Issues
Fixes https://github.com/jellyfin/jellyfin-web/issues/6535

### Code assistance
None

---

### DIFF
--- a/src/utils/bitrateTest.ts
+++ b/src/utils/bitrateTest.ts
@@ -1,11 +1,7 @@
 import { Api } from '@jellyfin/sdk';
-import type { EndPointInfo } from '@jellyfin/sdk/lib/generated-client/models/end-point-info';
-import { getSystemApi } from '@jellyfin/sdk/lib/utils/api/system-api';
 
 /** Maximum bitrate (Int32) */
 const MAX_BITRATE = 2147483647;
-/** Approximate LAN bitrate */
-const LAN_BITRATE = 140000000;
 /** Bitrate test timeout in milliseconds */
 const BITRATETEST_TIMEOUT = 5000;
 /** Bitrate cache time in milliseconds */
@@ -158,13 +154,23 @@ const detectBitrateInternal = (api: Api, tests: BitrateTest[], index: number, cu
 };
 
 /**
- * Detects the bitrate using a series of tests
- * @param {Api} api The api client
- * @param {EndPointInfo} endpointInfo Endpoint info for special handling on local networks
- * @returns {number} Normalized bitrate
+ * Detects the bitrate on the current device
+ * @param {Api} api The Api client
+ * @param {boolean} force Ignore the cached bitrate and force a re-detection
+ * @returns {number} The detected bitrate
  */
-const detectBitrateWithEndpointInfo = (api: Api, endpointInfo: EndPointInfo) => {
-    return detectBitrateInternal(
+export const detectBitrate = (api: Api, force: boolean) => {
+    if (!force
+        && lastDetectedBitrate
+        && Date.now() - (lastDetectedBitrateTime || 0) <= BITRATE_CACHE_DURATION) {
+        return Promise.resolve(lastDetectedBitrate);
+    }
+
+    if (pendingBitrateDetection) {
+        return pendingBitrateDetection;
+    }
+
+    pendingBitrateDetection = detectBitrateInternal(
         api,
         [
             {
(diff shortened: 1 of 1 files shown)


viown:
@dmitrylyzo Do you have any opinions on this? I'd appreciate your feedback since you mostly wrote this code.

dmitrylyzo:
Previously, it was exactly 140Mbps without real testing, assuming that LAN is at least 100Mbps (140 * 0.7 = 98).

Then, in https://github.com/jellyfin-archive/jellyfin-apiclient-javascript/pull/424, it becomes the minimum. Iirc, for safety.

For the use case from the issue (tunnel) and for WiFi 2.4GHz, we have no options but to remove this minimum. :shrug: 

Assuming that the bitrate tester will work in most cases :crossed_fingers:, users will rarely need to set the bitrate manually (at least they have this option/workaround).

viown:
In my limited testing under several different network conditions, the bitrate tester accurately outputs the correct bitrate capable of being streamed from the server (albeit reduced by 30% to account for network overhead). Anyone else wanting to test is welcome to, the detected bitrate is logged in the console under `[bitratetest]`

A significant portion of users connect over limited connections (iOS and Android web-wrapped clients), removing this minimum would result in a significant net increase in playable media across the board.

HaraldNordgren:
Great initiative, I have also noticed that Auto on wifi gives a higher bitrate than can be actually handled without buffering problems.

src/utils/bitrateTest.ts
```diff
@@ -185,41 +191,10 @@ const detectBitrateWithEndpointInfo = (api: Api, endpointInfo: EndPointInfo) =>
     ).then(result => {
         const bitrateInMbps = (result / 1048576).toFixed(2);
         console.debug(`[bitratetest] bitrate detected as ${bitrateInMbps} Mbps`);
-        if (endpointInfo.IsInNetwork) {
```
HaraldNordgren, line 188: Should we maybe not remove this logic but instead fallback to `LAN_BITRATE` only when no result is found?

```ts
if (!result && endpointInfo.IsInNetwork) {
    result = LAN_BITRATE;
```

HaraldNordgren (commented):
Great idea! We need a test for this logic.

You can copy as much as you want from my parallel PR: https://github.com/jellyfin/jellyfin-web/pull/8270

src/utils/bitrateTest.ts
viown: I'd prefer moving away from LAN checking entirely as this has not been a reliable way to test LAN for a long time due to the complex network configurations users can have.

HaraldNordgren: @viown That's a bit sad if it's the case.

dmitrylyzo: @HaraldNordgren In what cases may a result not be found?

Network failure? No playback either.

What else?

PR #7917, opened 2026-05-13
Fix keyboard in old browsers

dmitrylyzo:
### Changes

- Fix the check for the presence of the `KeyboardEvent.key|code` property.
- Fix the `KeyboardEvent.code` fallback.
- Add `KeyboardEvent.key` and `KeyboardEvent.code` polyfills.

Technically, the issue is fixed by 1st and 2nd commits.

We could use the [`keyboardevent-key-polyfill`](https://www.npmjs.com/package/keyboardevent-key-polyfill?ref=pkgstats.com) module for the `key` property, but now we need the `code` property.

### Issues

Regressions #4632
The check for the presence of the `KeyboardEvent.key` property is broken.
Regressions #7398
`KeyboardEvent.code` is `undefined` in old browsers (https://github.com/jellyfin/jellyfin-web/pull/7398#discussion_r3232719068). This caused navigation to stop working.

### Code assistance

None.

---

### DIFF
--- a/src/lib/legacy/keyboardEvent.js
+++ b/src/lib/legacy/keyboardEvent.js
@@ -1,6 +1,8 @@
 /**
  * Polyfill for KeyboardEvent
  * - Constructor.
+ * - 'code' property.
+ * - 'key' property.
  */
 
 (function (window) {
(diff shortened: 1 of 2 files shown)


thornbill:
@MBR-0001 would you mind confirming this still works for you?

MBR-0001:
Everything works ok here
@dmitrylyzo what about comma and period? Those are also used in playback control

dmitrylyzo:
> Everything works ok here @dmitrylyzo what about comma and period? Those are also used in playback control

They weren't mapped before, but I can add them.

MBR-0001 (approved):

PR #7947, opened 2026-05-21
Encode auth params when creating ApiClients

### LINKED ISSUE #7930: `X-Emby-Authorization` header construction fails when device name contains non-ISO-8859-1 characters

### Describe the bug

When the device name contains any character outside the ISO-8859-1 range (e.g. U+2019 RIGHT SINGLE QUOTATION MARK, em dashes, emoji, CJK characters), every API call from the web client throws synchronously before reaching the network:

```
TypeError: Failed to execute 'fetch' on 'Window': Failed to read the 'headers' property from 'RequestInit': String contains non ISO-8859-1 code point.
```

The user-visible symptom is a generic "Connection Failure" with no actionable information. The server is reachable and `/System/Info/Public` returns 200 (visible in logs via `tryReconnect`/`Reconnect succeeded`), but every subsequent authenticated request fails because the `X-Emby-Authorization` header — which embeds the device name verbatim — is rejected by the browser's Fetch implementation per [RFC 7230 §3.2.4](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.4) (header field values are ISO-8859-1 only, and modern browsers enforce this strictly via the [Fetch spec](https://fetch.spec.whatwg.org/#concept-header-value-normalize)).

### Reproduction Steps

1. On macOS, set the computer name to anything containing a typographic apostrophe (the default when typing `'` in System Settings due to Smart Quotes substitution): e.g. `Michaël’s MacBook Pro`.
2. Launch Jellyfin Desktop, which derives its device name from `ComputerName`.
3. Attempt to connect to any server.
4. Observe `Connection Failure` in the UI; logs show the `TypeError` above on every fetch.

This is also reproducible by setting a custom device name containing any of: U+2018/U+2019 smart quotes, em/en dashes, emoji, CJK, Arabic, Hebrew, or any character > U+00FF.

### Expected/Actual behaviour

Either:

1. **Sanitize at construction time** — percent-encode or RFC 5987-encode the device name component when building the `X-Emby-Authorization` header, so any Unicode device name produces a valid header. This is the preferred fix and would match how `User-Agent` is typically handled.
2. **Validate and warn** — detect non-Latin-1 characters in the configured device name on startup and surface a clear error/warning rather than the opaque `Connection Failure`.

### Logs

```
2026-05-16T11:19:20 INFO    [JS] Requesting url without automatic networking: https://<server>/System/Info/Public
2026-05-16T11:19:20 ERROR   [JS] [LibraryMenu] failed to fetch system info TypeError: Failed to execute 'fetch' on 'Window': Failed to read the 'headers' property from 'RequestInit': String contains non ISO-8859-1 code point.
```

### Server version

10.11.8

### Web version

10.11.8

### Build version

10.11.8

### Platform

Macos Silicon

### Browser

Jellyfin Desktop

### Additional information

_No response_

thornbill:
### Changes
Fixes our handling of client and device details to ensure they are properly encoded in the legacy ApiClient and not double encoded in the TS SDK Api

### Issues
Fixes https://github.com/jellyfin/jellyfin-web/issues/7930

### Code assistance
Zed autocomplete

---

### DIFF
--- a/src/lib/jellyfin-apiclient/connectionManager.js
+++ b/src/lib/jellyfin-apiclient/connectionManager.js
@@ -1,7 +1,6 @@
 import { AUTHORIZATION_HEADER } from '@jellyfin/sdk/lib/constants';
 import { getAuthorizationHeader } from '@jellyfin/sdk/lib/utils';
 import { MINIMUM_VERSION } from '@jellyfin/sdk/lib/versions';
-import { ApiClient } from 'jellyfin-apiclient';
 
 import events from 'utils/events';
 import { ajax } from 'utils/fetch';
@@ -756,7 +755,7 @@ export default class ConnectionManager {
     /**
      * Gets the ApiClient for a given BaseItem or ServerId.
      * @param {import('@jellyfin/sdk/lib/generated-client').BaseItemDto | string | undefined} item
-     * @returns {ApiClient}
+     * @returns {import('jellyfin-apiclient').ApiClient}
      */
     getApiClient(item) {
         if (!item) {
--- a/src/lib/jellyfin-apiclient/connectionManager.js
+++ b/src/lib/jellyfin-apiclient/connectionManager.js
@@ -5,6 +5,7 @@ import { ApiClient } from 'jellyfin-apiclient';
 
 import events from 'utils/events';
 import { ajax } from 'utils/fetch';
+import { createApiClient } from 'utils/jellyfin-apiclient/createApiClient';
 import { equalsIgnoreCase } from 'utils/string';
 import { compareVersions } from 'utils/versions';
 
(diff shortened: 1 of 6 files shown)


src/lib/jellyfin-apiclient/connectionManager.js
nielsvanvelzen: Unused import!!

nielsvanvelzen (changes requested):

nielsvanvelzen:
I assume you already looked into it, but wouldn't it be better to set the information on the legacy ApiClient from the TS SDK client instead?

src/lib/jellyfin-apiclient/connectionManager.js
thornbill: Helps if I don't get distracted and push my changes... :moyai:

thornbill:
Yes, eventually. Currently the auth and credential management are still heavily tied to the legacy ApiClient. We should make it a priority for v13 to finally tackle that.

nielsvanvelzen (approved):

PR #7953, opened 2026-05-24
Guard itemDetails page against empty GUID and lookup failures

RumbleCrumbleRS:
## Summary

The Jellyfin server resolves the empty GUID (`Guid.Empty` = `00000000000000000000000000000000`) for the `/Users/{userId}/Items/{itemId}` endpoint to the user root view ("Media Folders") instead of returning a 404. As a result, navigating to `#/details?id=00000000000000000000000000000000` (e.g. from a broken bookmark or share link) silently renders the wrong item with no indication that the id is invalid.

## Changes (`src/controllers/itemDetails/index.js`)

- **`isValidItemId(id)`** — new helper that rejects all-zero GUIDs (with or without dashes), causing them to fall through to the `Invalid request` branch in `getPromise`.
- **`getPromise`** — returns `Promise.reject(new Error('Invalid request'))` instead of throwing synchronously, so the `Promise.all().catch()` consumer receives the rejection rather than the thrown error escaping the chain.
- **`reload` catch handler** — calls `loading.hide()` and `appRouter.goHome()` on any detail-page lookup failure (invalid/empty id, 404, deleted item, no permission), preventing the prior stale detail page from remaining rendered.

## Reproduction

1. Log in to a Jellyfin instance.
2. Navigate to `#/details?id=00000000000000000000000000000000`.
3. **Before:** "Media Folders" detail page rendered (server silently returns the user root view for `Guid.Empty`).
4. **After:** redirect to home; console logs `failed to get item or current user: Error: Invalid request`.

## Risk

Very small surface — only the detail-page lookup failure path is changed. Successful item lookups are completely unaffected. Worst case on an unexpected regression: a 404 or permission error redirects to home instead of leaving the user on a blank/stale page (still an improvement).


andrewrabert:
This is not ready for review:
- a ton of unrelated changes
- conflicts

PR #7971, opened 2026-05-30
Disable touch switcher on experimental and disable scroll buttons on touch devices

Shadowghost:
On the old layout we had `Home` and `Favorite` tabs on the main screen which you could switch between with swiping on touch devices. These tabs do not exist in that constellation on the experimental layout, leading to unintended switches when you swiped on the main screen.
iPads are considered desktop devices by Apple, so the trigger for disabling the scroll buttons on non-desktop does not trigger, leading to weird usability issues. IMO the deciding factor should be touch device and not mobile device.

### Changes
* Disable scroll buttons on touch devices
* Disable tab switcher on experimental layout

### Code assistance
None

---

### DIFF
--- a/src/elements/emby-scroller/emby-scroller.js
+++ b/src/elements/emby-scroller/emby-scroller.js
@@ -116,7 +116,7 @@ ScrollerPrototype.attachedCallback = function () {
     }
 
     const scrollFrame = this;
-    const enableScrollButtons = layoutManager.desktop && horizontal && this.getAttribute('data-scrollbuttons') !== 'false';
+    const enableScrollButtons = layoutManager.desktop && !browser.touch && horizontal && this.getAttribute('data-scrollbuttons') !== 'false';
 
     const options = {
         horizontal: horizontal,
--- a/src/components/maintabsmanager.js
+++ b/src/components/maintabsmanager.js
@@ -1,5 +1,6 @@
 import dom from '../utils/dom';
 import browser from '../scripts/browser';
+import layoutManager from './layoutManager';
 import Events from '../utils/events.ts';
 import '../elements/emby-tabs/emby-tabs';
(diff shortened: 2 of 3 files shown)


src/elements/emby-scroller/emby-scroller.js
```diff
@@ -116,7 +116,7 @@ ScrollerPrototype.attachedCallback = function () {
     }
 
     const scrollFrame = this;
-    const enableScrollButtons = layoutManager.desktop && horizontal && this.getAttribute('data-scrollbuttons') !== 'false';
+    const enableScrollButtons = layoutManager.desktop && !browser.touch && horizontal && this.getAttribute('data-scrollbuttons') !== 'false';
```
dmitrylyzo, line 119: Some laptops are equipped with touch screens (Huawei Matebook 16s).

Shouldn't we use `Mobile` layout for iPad? Imo, you should not use `Desktop` layout if you are going to use a touch screen.

Shadowghost: Well, Apple reports iPads as desktop. There is also an argument to be made that mobile does not necessarily mean touch screen

dmitrylyzo: > Well, Apple reports iPads as desktop.

Apple or our browser detector?

> There is also an argument to be made that mobile does not necessarily mean touch screen

But `Mobile` layout is designed for touch screen. 🤷‍♂️ You can still use a mouse with it though.
`Desktop` is more for mouse.

Shadowghost: From what I gathered in my tests, iPad safari reports itself as desktop to websites.
I don't have any strong opinions on how to achieve the changes, I'm fine with however we do it 😄

thornbill (approved):

PR #7975, opened 2026-05-31
Fix navigation with game controllers

### LINKED ISSUE #7973: [12] XBox client can't go "Back" in unstable

### Describe the bug

Running unstable, with Display set to either "Auto" or "TV", when watching a video, the "B" or "Back" buttons do not work. Sometimes I am able to bring up the OSD and browse the the back button in the upper left and hit that to escape the video, sometimes I need to either let the video stop or kill the client. I am thinking this is a web-based mapping issue, but happy to be wrong and put it on Venson 😄 

This also seems to happen during browsing libraries.. can't hit B/Left to go "back".

### Reproduction Steps

1. Run unstable server image with latest prod xbox client
2. Watch a video
3. Try to exit the playing video by hitting "B" or back button on the controller (Back button is on my [8bitdo remote](https://www.8bitdo.com/media-remote-for-xbox/), not sure what it maps to on a normal controller.. guessing just "left" on the dpad, but unsure.

### Expected/Actual behaviour

I expected the app to stop playing the video, but it seems to ignore the button. If the OSD is not on the screen, it won't bring it up.

### Logs

No useful logs on server.

### Server version

12.0

### Web version

12.0

### Build version

2026052922

### Platform

XBox Series X

### Browser

XBox Client

### Additional information

_No response_

MBR-0001:
### Changes
Various changes in keyboard navigation broke controller navigation. This fixes it by bringing back the KeyNames which are used by controllers.

image

image

### Issues
Fixes #7973

### Code assistance
Clanker told me how to turn on controller navigation in browser cuz I couldn't find it :clown_face: 

---

### DIFF
--- a/src/scripts/keyboardNavigation.js
+++ b/src/scripts/keyboardNavigation.js
@@ -13,13 +13,12 @@ import appSettings from './settings/appSettings';
  */
 const KeyNames = {
     13: 'Enter',
-    19: 'Pause',
     27: 'Escape',
-    32: 'Space',
     37: 'ArrowLeft',
     38: 'ArrowUp',
     39: 'ArrowRight',
     40: 'ArrowDown',
+
     // UWP WebView section start --
     // Navigation Up/Down/Left/Right is part of TVJS directionalnavigation-1.0.0.0.js
     // Unsure what this is used for. Media remote?
--- a/src/scripts/keyboardNavigation.js
+++ b/src/scripts/keyboardNavigation.js
@@ -12,6 +12,14 @@ import appSettings from './settings/appSettings';
  * Key name mapping.
  */
 const KeyNames = {
+    13: 'Enter',
+    19: 'Pause',
+    27: 'Escape',
+    32: 'Space',
+    37: 'ArrowLeft',
+    38: 'ArrowUp',
+    39: 'ArrowRight',
+    40: 'ArrowDown',
     // UWP WebView section start --
     // Navigation Up/Down/Left/Right is part of TVJS directionalnavigation-1.0.0.0.js
     // Unsure what this is used for. Media remote?


MBR-0001:
Added release critical because gamepad navigation is currently real borked

src/scripts/keyboardNavigation.js
```diff
@@ -12,6 +12,14 @@ import appSettings from './settings/appSettings';
...
  */
 const KeyNames = {
+    13: 'Enter',
+    19: 'Pause',
+    27: 'Escape',
+    32: 'Space',
+    37: 'ArrowLeft',
+    38: 'ArrowUp',
+    39: 'ArrowRight',
+    40: 'ArrowDown',
```
dmitrylyzo, line 20: `Pause` and `Space` are unnecessary.
https://github.com/jellyfin/jellyfin-web/blob/8e8cdf548990c57d38b4612b0935eaaed12217f6/src/scripts/gamepadtokey.js#L41-L50

Add a blank line at the end of block.

But I would simply add `event.code = key;` here (since we switched to `code`):
https://github.com/jellyfin/jellyfin-web/blob/8e8cdf548990c57d38b4612b0935eaaed12217f6/src/scripts/gamepadtokey.js#L202

On the other hand, in this case, `Enter` won't work because there's no alias for `GamepadA`. :thinking:

dmitrylyzo (commented):
Oops. 😅

src/scripts/keyboardNavigation.js
MBR-0001: yeah GamepadA is not in aliases (unsure why)

I thought about going the other way and mapping by key (instead of keyCode) but KeyboardNavigation uses the string `GamepadLeftThumbDown` while gamepadtokey uses `GamepadLeftThumbStickDown`. I had no idea why the difference so I opted out for this approach instead

dmitrylyzo: > yeah GamepadA is not in aliases (unsure why)

Probably because it doesn't produce a click (on UWP), and we have to do a click ourselves in a dedicated case-block.

> I thought about going the other way and mapping by key (instead of keyCode)

`key` is not populated for special keys (TVs), iirc. So we have to use `keyCode`.

> but KeyboardNavigation uses the string `GamepadLeftThumbDown` while gamepadtokey uses `GamepadLeftThumbStickDown`. I had no idea why the difference so I opted out for this approach instead

We can probably change them in KeyboardNavigation, because I just suggested the names:
https://github.com/jellyfin/jellyfin-web/pull/6993#discussion_r2175441848
But currently, reverting KeyNames is safer.

dmitrylyzo (approved):

PR #7985, opened 2026-06-02
Fix empty encoding preset on new installation

nyanmisaka:
### Changes
- Fix empty encoding preset on new installation

### Issues
- Fixes b3de4af

### Code assistance

---

### DIFF
--- a/src/apps/dashboard/routes/playback/transcoding.tsx
+++ b/src/apps/dashboard/routes/playback/transcoding.tsx
@@ -735,7 +735,7 @@ export const Component = () => {
 
                             <TextField
                                 name='EncoderPreset'
-                                value={config.EncoderPreset}
+                                value={config.EncoderPreset || 'auto'}
                                 onChange={onConfigChange}
                                 label={globalize.translate('LabelEncoderPreset')}
                                 helperText={globalize.translate('EncoderPresetHelp')}


src/apps/dashboard/routes/playback/transcoding.tsx
dmitrylyzo, line 715: https://github.com/jellyfin/jellyfin-web/blob/4c8ae195eb6d95229a63b392092658635602e2bf/src/apps/dashboard/controllers/encodingsettings.js#L27

dmitrylyzo, line 494: https://github.com/jellyfin/jellyfin-web/blob/4c8ae195eb6d95229a63b392092658635602e2bf/src/apps/dashboard/controllers/encodingsettings.js#L38

dmitrylyzo, line 516: https://github.com/jellyfin/jellyfin-web/blob/4c8ae195eb6d95229a63b392092658635602e2bf/src/apps/dashboard/controllers/encodingsettings.js#L39

dmitrylyzo, line 532: page.querySelector('#selectTonemappingRange').value = config.TonemappingRange || 'auto';

dmitrylyzo (commented):
There are other selectable options that are non-nullable, so they have values on a fresh launch. We can add fallbacks to them, but on the other hand, it would be better to fix the `EncoderPreset` on the server side and use empty fields as a sign that something is wrong.

So maybe instead we should make `EncoderPreset` non-nullable?
https://github.com/jellyfin/jellyfin/blob/53eda14dccddbcbd8806150489ff8f4495814631/MediaBrowser.Model/Configuration/EncodingOptions.cs#L220

nyanmisaka:
> There are other selectable options that are non-nullable, so they have values on a fresh launch. We can add fallbacks to them, but on the other hand, it would be better to fix the `EncoderPreset` on the server side and use empty fields as a sign that something is wrong.
> 
> So maybe instead we should make `EncoderPreset` non-nullable? [jellyfin/jellyfin@`53eda14`/MediaBrowser.Model/Configuration/EncodingOptions.cs#L220](https://github.com/jellyfin/jellyfin/blob/53eda14dccddbcbd8806150489ff8f4495814631/MediaBrowser.Model/Configuration/EncodingOptions.cs#L220)

It should be fixed in https://github.com/jellyfin/jellyfin/pull/17007, so I guess this PR is no longer relevant?

dmitrylyzo:
> It should be fixed in [jellyfin/jellyfin#17007](https://github.com/jellyfin/jellyfin/pull/17007), so I guess this PR is no longer relevant?

I think so, but let's wait for the server team decision.

src/apps/dashboard/routes/playback/transcoding.tsx
```diff
@@ -735,7 +735,7 @@ export const Component = () => {
 
                             <TextField
                                 name='EncoderPreset'
-                                value={config.EncoderPreset}
+                                value={config.EncoderPreset || 'auto'}
```
dfederm, line 738: Doesn't this only fix the rendered value? The UI will display "auto" but if the user saves without touching the field, it still sends an empty value.

PR #7996, opened 2026-06-03
Decouple gamepadtokey repeat rate from framerate

brad1111:
### Changes

Changes the throttle function to prevent repeated keydown events from firing too fast, and rename the existing function to delay. I set the repeat rate to 30hz, as that felt closest to how fast the page moves on keyboard on Windows (maybe should be configurable?).
Also moves the timer away from triggering on vsync, meaning input delay should be consistent across refresh rates.

### Issues

Might fix part of #7834 

### Code assistance

The Duckduckgo AI summary for keyboard repeat rate suggested that the two parts were called delay and throttle, and they felt good enough as names. No other AI was used.

---

### DIFF
--- a/src/scripts/gamepadtokey.js
+++ b/src/scripts/gamepadtokey.js
@@ -170,17 +170,29 @@ _ButtonPressedState.setdPadRight = function (newPressedState) {
     _dPadRightPressed = newPressedState;
 };
 
-const times = {};
+const delayTimes = {};
 
-function throttle(key) {
-    const time = times[key] || 0;
+function delay(key) {
+    const time = delayTimes[key] || 0;
     const now = new Date().getTime();
 
     return (now - time) >= 200;
 }
 
+function resetDelay(key) {
+    delayTimes[key] = new Date().getTime();
+}
+
+const throttleTimes = {};
+
+function throttle(key) {
+    const time = throttleTimes[key] || delayTimes[key];
+    const now = new Date().getTime();
+    return (now - time) >= 33;
+}
+
 function resetThrottle(key) {
-    times[key] = new Date().getTime();
+    throttleTimes[key] = new Date().getTime();
 }
 
 const isElectron = navigator.userAgent.toLowerCase().indexOf('electron') !== -1;
@@ -225,16 +237,17 @@ function raiseKeyEvent(oldPressedState, newPressedState, key, keyCode, enableRep
         // always fire if this is the initial down press
         if (oldPressedState === false) {
             fire = true;
-            resetThrottle(key);
+            resetDelay(key);
         } else if (enableRepeatKeyDown) {
-            fire = throttle(key);
+            fire = delay(key) && throttle(key);
         }
 
         if (fire && keyCode) {
             newPressedEvent = raiseEvent('keydown', key, keyCode);
+            resetThrottle(key);
         }
     } else if (newPressedState === false && oldPressedState === true) {
-        resetThrottle(key);
+        resetDelay(key);
 
         // button up
         if (keyCode) {
--- a/src/scripts/gamepadtokey.js
+++ b/src/scripts/gamepadtokey.js
@@ -356,18 +356,16 @@ function runInputLoop() {
             }
         }
     }
-    // Schedule the next one
-    inputLoopTimer = requestAnimationFrame(runInputLoop);
 }
 
 function startInputLoop() {
     if (!inputLoopTimer) {
-        runInputLoop();
+        inputLoopTimer = setInterval(runInputLoop, 15);
     }
 }
 
 function stopInputLoop() {
-    cancelAnimationFrame(inputLoopTimer);
+    clearInterval(inputLoopTimer);
     inputLoopTimer = undefined;
 }
 
--- a/src/scripts/gamepadtokey.js
+++ b/src/scripts/gamepadtokey.js
@@ -170,29 +170,22 @@ _ButtonPressedState.setdPadRight = function (newPressedState) {
     _dPadRightPressed = newPressedState;
 };
 
-const delayTimes = {};
-
-function delay(key) {
-    const time = delayTimes[key] || 0;
-    const now = new Date().getTime();
-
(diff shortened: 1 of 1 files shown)


MBR-0001 (approved):
Seems to be working from my testing, code looks good
master:

[Screencast_20260616_222717.webm](https://github.com/user-attachments/assets/5eeff403-ea21-4fb5-8d17-cd0e9fe1a998)

this PR:

[Screencast_20260616_222735.webm](https://github.com/user-attachments/assets/dd34bd6a-019c-44fb-9491-8b75bdbf4a69)

src/scripts/gamepadtokey.js
```diff
@@ -225,16 +237,17 @@ function raiseKeyEvent(oldPressedState, newPressedState, key, keyCode, enableRep
...
-            resetThrottle(key);
+            resetDelay(key);
         } else if (enableRepeatKeyDown) {
-            fire = throttle(key);
+            fire = delay(key) && throttle(key);
         }
 
         if (fire && keyCode) {
             newPressedEvent = raiseEvent('keydown', key, keyCode);
+            resetThrottle(key);
```
dmitrylyzo, line 241: I believe that only this line should be added to fix fast `keydown`.

1. We fire first `keydown` event and set the time.
2. In the next frame, `throttle` denies another `keydown` because not enough time has passed.
3. After some frames `throttle` allows us to fire another `keydown`, but we don't reset the time.
4. In the next frame, `throttle` allows another `keydown` because the time was not reset.

With this line, in the 4th step, `throttle` will not allow us to fire the event.

src/scripts/gamepadtokey.js
```diff
@@ -225,16 +237,17 @@ function raiseKeyEvent(oldPressedState, newPressedState, key, keyCode, enableRep
...
         }
     } else if (newPressedState === false && oldPressedState === true) {
-        resetThrottle(key);
```
dmitrylyzo, line 237: This line can definitely be removed, because we only throttle `keydown` and set the time on the first fire.

src/scripts/gamepadtokey.js
```diff
@@ -225,16 +237,17 @@ function raiseKeyEvent(oldPressedState, newPressedState, key, keyCode, enableRep
         // always fire if this is the initial down press
         if (oldPressedState === false) {
             fire = true;
-            resetThrottle(key);
```
dmitrylyzo, line 228: This line can also be removed (moved to the fire block).

src/scripts/gamepadtokey.js
brad1111: The way this works is:
1. We fire first `keydown` event and set the delay time.
2. The next time the clock triggers it denies another `keydown` because we haven't reached the delay time.
3. After 200ms `delay()` will allow us a to trigger another `keydown` event. The repeat time is set here.
4. The next time the clock triggers it denies another `keydown` event since we haven't reached the repeat time.
5. After 33ms `throttle()` will allow us to trigger another `keydown` event. The repeat time is set here again.
6. Repeat 4-5 until key is no longer pressed.

If we were to remove `resetDelay()` from Line 240 it would take out the pause that happens before the key starts repeating.

dmitrylyzo: If you want to simulate this ("ms" is the time that has passed since the previous event):
```
keydown
keydown 600ms
keydown 40ms
keydown 41ms
keydown 39ms
keydown 62ms
keydown 17ms
keydown 42ms
keydown 53ms
keydown 40ms
keydown 41ms
keydown 39ms
keydown 41ms
keydown 40ms
keydown 39ms
keydown 40ms
keydown 41ms
keydown 40ms
keydown 41ms
keyup 34ms
```
you don't need 2 time arrays - you just need to compare with the corresponding delay (600, 40).

The question is, do we need to simulate this?

Btw, the easiest way to simulate this with minimal effort is to use the old throttling (probably increase from 200 to 600) + 40ms interval (instead of an animation frame). But in this case, `keyup` will also be delayed by the interval.

brad1111: I think it is best to have to have a longer delay at the start and then repeat at different rate if that’s what you mean. Most TV apps I've used work this way, and I feel it would be odd not to have it.

I did consider having the repeat set by the interval, and I think on reflection that's a fair approach if we're going to have that interval be a static 25hz as you've suggested (I was initially concerned that a below 60hz polling rate might feel unresponsive, but it feels fine in practice).

The only reason I can think of to have it handled separately is if it would be desirable for the repeat rate to be configurable, as a long interval for a slow repeat rate would probably make normal input unresponsive.

I'm fine with either improving how I handle the time array or doing the simple interval option depending on what you think is a better solution.

dmitrylyzo: > I'm fine with either improving how I handle the time array or doing the simple interval option depending on what you think is a better solution.

From my tests, I got at least 23ms from `keydown` to `keyup` (single press). It is mostly over 39ms. We lose faster `keyup` only after repeated `keydown` which is probably unnoticeable. So I think simple interval (40ms) is fine. You can lower it a bit (30-33ms).

If you really want to make it "perfect", use single `time` array (as it was). You still need a 2nd array to select the delay to compare with. You can add an array of `keydown` counters (better, imo) or current delays.

Test script:
```js
let start = performance.now();
function handler(e) {
    console.debug(e.type, performance.now() - start);
    start = performance.now();
}
['keydown', 'keyup'].forEach(t => document.addEventListener(t, handler));
```

brad1111: I ended up using an array of counters from your suggestion.

Since changing the interval to 20ms polling interval instead of it gives a fairly consistent 40ms-50ms between keydown events if there's nothing changing on the page. With keyup generally being 20-30ms after.

MBR-0001 (approved):
Works correctly, scrolling is a bit slower than last time I tested but still fine, actually better :D

https://github.com/user-attachments/assets/d67cce02-0264-4ba2-92b9-89fc05eeecfd

PR #8002, opened 2026-06-05
Make Played and Unplayed filters mutually exclusive for the Modern view

### LINKED ISSUE #16297: Filter Played / Unplayed broken logic
> Setting the filter to both "Played" and "Unplayed" is also logically broken.  It will still actually filter out videos that have been played.  It functions the same as just filtering for "Unplayed", so it doesn't solve the problem. 

 _Originally posted by @Majoraslayer in [#11667](https://github.com/jellyfin/jellyfin/issues/11667#issuecomment-3765041844)_

Might have found the issue for the inconsistent filtering result in ItemsController.cs

From my understanding when both IsPlayed and IsUnplayed are present, the last one processed overwrites the first, making the combined semantic of "show all items" impossible to express correctly via this foreach loop:

https://github.com/jellyfin/jellyfin/blob/56a469d8c3c0eed0d809d42a60d847e67ab61671/Jellyfin.Api/Controllers/ItemsController.cs#L389
```
foreach (var filter in filters)
{
    switch (filter)
    {
        case ItemFilter.IsPlayed:
            query.IsPlayed = true;   //  <- first set to true...
            break;
        case ItemFilter.IsResumable:
            query.IsResumable = true;
            break;
        case ItemFilter.IsUnplayed:
            query.IsPlayed = false;  //  <- ...then overwritten to false
            break;
    }
}
```

theguymadmax:
Selecting both Played and Unplayed filters causes a Bad Request ([#16398](https://github.com/jellyfin/jellyfin/pull/16398))

### Changes
Makes the Played and Unplayed filters mutually exclusive for the modern view. The PR does not touch the legacy filters.

### Issues
Fixes: https://github.com/jellyfin/jellyfin/issues/16297

### Code assistance
N/A

---

### DIFF
--- a/src/apps/experimental/components/library/filter/FiltersStatus.tsx
+++ b/src/apps/experimental/components/library/filter/FiltersStatus.tsx
@@ -15,6 +15,11 @@ const defaultFiltersOptions = [
     { label: 'ContinueWatching', value: ItemFilter.IsResumable }
 ];
 
+const mutuallyExclusiveFilters: Partial<Record<ItemFilter, ItemFilter>> = {
+    [ItemFilter.IsPlayed]: ItemFilter.IsUnplayed,
+    [ItemFilter.IsUnplayed]: ItemFilter.IsPlayed
+};
+
 interface FiltersStatusProps {
     viewType: LibraryTab;
     libraryViewSettings: LibraryViewSettings;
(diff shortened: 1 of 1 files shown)


MBR-0001:
This only fixes the experimental UI, what about the regular one?

theguymadmax:
> This only fixes the experimental UI, what about the regular one?

Yep, that's correct.

thornbill:
Should this also apply to the continue watching filter?

theguymadmax:
No, an item can be played and still be in progress. That's probably a new behavior in 12.0  [#15762](https://github.com/jellyfin/jellyfin/pull/15762)

thornbill (approved):

nielsvanvelzen:
Ideally we'd use radio buttons or a tri-state toggle (Played, Unset, Unplayed) instead of checkboxes. Yes I'm too late 🗿

thornbill:
Yes, but I plan to make some adjustments to the UI anyway :wink:

PR #8004, opened 2026-06-05
migrate comics and PDFs to new BookOsd component

dkanada:
### Changes

PDF and comic players now use the BookOsd component with their existing subset of UI features.

- neither player supports fullscreen at this time
- side arrow navigation was removed from the comics player

Subsequent pull requests will focus on adding fullscreen everywhere, implementing a unified page indicator in React, and extracting book navigation controls into a global helper function for user customization.

### Issues

None

### Code Assistance

None


MBR-0001 (commented):
A few observations:
The new OSD for pdf hides/blocks the content at the top and bottom 

https://github.com/user-attachments/assets/e48a51ab-b5c7-41a0-aacf-14dfb3ed7e1f

On the comic player this does not happen

https://github.com/user-attachments/assets/4ef25a55-a75f-4330-b832-25e64cc90f10

BUT, it seems dragging to change page does not work:

https://github.com/user-attachments/assets/6d742b1d-9af4-4ae1-b2c2-4cdb97d63155

src/plugins/bookPlayer/BookOsd/BookOsd.tsx
```diff
@@ -78,6 +103,22 @@ const BookOsd: FC<BookOsdProps> = ({
                     />
                 )}
 
+                {onToggleDirection && (
+                    <IconButton
+                        onClick={onClickDirection}
+                        icon={direction ? 'arrow_circle_left' : 'arrow_circle_right'}
+                        title={direction ? 'Right to Left' : 'Left to Right'}
```
MBR-0001: this should use i18n

src/plugins/bookPlayer/BookOsd/BookOsd.tsx
```diff
@@ -78,6 +103,22 @@ const BookOsd: FC<BookOsdProps> = ({
...
+                    />
+                )}
+
+                {onToggleLayout && (
+                    <IconButton
+                        onClick={onClickLayout}
+                        icon={layout ? 'import_contacts' : 'devices_fold'}
+                        title={layout ? 'Double Page View' : 'Single Page View'}
```
MBR-0001: this should use i18n

src/plugins/pdfPlayer/plugin.js
```diff
@@ -277,11 +259,13 @@ export class PdfPlayer {
...
                 this.pages[page] = document.createElement('canvas');
                 this.renderPage(this.pages[page], parseInt(page.slice(4), 10));
+
+                this.pages[page].id = 'canvas';
             }
         }
 
         // show the requested page
-        this.replaceCanvas(this.pages[prefix + number], number);
+        document.querySelector('#canvas')?.replaceWith(this.pages[prefix + number]);
```
MBR-0001: I think it may be better to use the old parent.replaceChild approach since not all browsers we target support this

MBR-0001 (changes requested):
Previously mentioned issues are resolved, now the pdf viewer is broken on mobile

https://github.com/user-attachments/assets/62414d09-4f8d-48ec-83dd-9055b43d5846

Image

MBR-0001 (approved):

PR #7838, opened 2026-04-17
Migrate WebSockets to SDK Subscriptions

anultravioletaurora:
### Changes

Removes previous websocket connection implementations in exchange for the subscription model provided by the `@jellyfin/sdk`.

Updates TS-based components to use subscription model in `useEffect`s instead of establishing polling intervals.

Updates JS-based components to use subscription model on mounting.

Removes usage of `ensureWebSocket` from the `serverConnections.js`, since websocket connections will be made lazily when needed by the SDK.

### Code assistance

I used `gemma4:e4b` and Claude Haiku 4.5 to assist in finding all of the locations where the legacy usage exists, and Claude Sonnet 4.6 to assist with my understanding and editing of larger files (`sessionPlayer.js`, `serverConnections.js`, `connectionManager.js`)

---

* [x] I have read and followed the [contributing guidelines](https://github.com/jellyfin/jellyfin-web/blob/master/CONTRIBUTING.md).
* [x] I have tested these changes.
* [x] I have verified that this is not duplicating changes in an existing PR.
* [x] I have provided a *substantive* review of [another web PR](https://github.com/jellyfin/jellyfin-web/pulls?q=is%3Apr+is%3Aopen+-label%3Adependencies+-label%3Ablocked+-label%3Abackend+-label%3Ainvalid+-label%3A"merge+conflict"+-label%3Astale+-author%3A%40me).

PR #8011, opened 2026-06-08
Fix Duplicated `/socket` Connections

anultravioletaurora:
Fixes duplicated connections to the `/socket` endpoint caused by #7838. Ensures that a single socket connection is created as long as subscriptions are active, and disconnected when subscriptions are deregistered.

### Changes

- Adds a `new Map<string, Api>` to the `connectionManager` for distinctly tracking `Api` instances
- Updates retrieval of the `Api` in the `ServerConnections` to leverage the `Api` map
- Updates the `toApi` hook to retrieve the corresponding `Api` instance from the `ServerConnections`

### Code assistance

I used `gemma4:12b` and `Claude Haiku 4.5` for asking questions, debugging assistance and performing a code review.

---


src/lib/jellyfin-apiclient/connectionManager.js
```diff
@@ -771,6 +786,19 @@ export default class ConnectionManager {
         })[0];
     }
 
+    getApi(item) {
+        if (!item) {
+            throw new Error('item or serverId cannot be null');
+        }
+
+        // Accept string or object
+        if (item.ServerId) {
```
thornbill: Unless you have a good reason to, let's not continue the pattern of allowing a Server ID or BaseItem here. It should only accept a Server ID.

thornbill (changes requested):

src/lib/jellyfin-apiclient/connectionManager.js
anultravioletaurora: Oh did - did you not order the copypasta? 

Who here at the table ordered the copypasta?

In all seriousness, I have no reason to follow the pattern, so I'll remove 👍

thornbill:
Currently there is an issue that after logging out a websocket connection continues to try to be reestablished.

nielsvanvelzen:
After signing in, when using the "select server" function the websocket is not closed until the next server is chosen. This does work correctly when signing out though.

thornbill:
I can confirm that the behavior of not signing out immediately when switching server is preexisting. It was not introduced by these changes.

nielsvanvelzen:
Looks like syncplay is broken (creating a room should automatically join it, that doesn't happen. In fact you cannot join any room at all) and remote control is also not handled. My best guess is that this relates to the listeners in serverNotifications.js.

src/lib/jellyfin-apiclient/connectionManager.js
```diff
@@ -330,13 +346,21 @@ export default class ConnectionManager {
...
+                events.trigger(self, 'localusersignedout', [logoutInfo]);
+            };
+
+            apiClient._sdk ??= toApi(apiClient);
+            const api = apiClient._sdk;
+
+            const sessionApi = api ? getSessionApi(api) : undefined;
+
+            return Promise.allSettled([
+                sessionApi.reportSessionEnded(),
```
thornbill: `sessionApi` could be undefined from the assignment above

src/lib/jellyfin-apiclient/ServerConnections.js
```diff
@@ -53,26 +53,14 @@ class ServerConnections extends ConnectionManager {
...
-                return _sdkApi.subscribe(messageTypes, onMessage, subscriptionIntervals);
-            };
+            const api = this.getApi(apiClient.serverId());
+
+            if ((!this.api && api) || this.localApiClient === apiClient) {
+                this.api = api;
+            }
+
+            apiClient.subscribe = (messageTypes, onMessage, subscriptionIntervals) =>
+                apiClient._sdk.subscribe(messageTypes, onMessage, subscriptionIntervals);
```
thornbill:
```suggestion
            apiClient.subscribe = apiClient._sdk.subscribe.bind(apiClient._sdk);
```

thornbill (changes requested):
🎉  seems to be working now! Just some minor suggestions.

src/lib/jellyfin-apiclient/connectionManager.js
anultravioletaurora: Added optional chaining 👍

thornbill (approved):

PR #8025, opened 2026-06-10
Prevent screensaver while viewing photos or reading books

### LINKED ISSUE #7873: Screensaver rises while browsing photos using Jellyfin

### Describe the bug

While browsing photos using Jellyfin, screensaver always rises when idle timeout is reached since browsing starts.
This is a result of code in screensavermanager.js, which measures lastFunctionalEvent only for video playback, but not photos and checks it every 5000ms (see code in src/scripts/screensavermanager.js) 
So, while browsing photos screesaver always rises when idle timeout is reached.

### Reproduction Steps

1. Set idle timeout for backdrop screensaver and activate it
2. Open photos library
3. Start browsing photos
4. When idle timeout is reached while browsing, you will see the screensaver

### Expected/Actual behaviour

When you browse photos no screensaver should arise.

### Logs

_No response_

### Server version

10.11.8

### Web version

10.11.8

### Build version

10.11.8

### Platform

Tizen, Linux

### Browser

Firefox, TizenBrowser

### Additional information

_No response_

dfederm:
### Changes
The screensaver could activate while viewing a photo slideshow or reading a book, PDF, or comic. These players don't set themselves as the current player in the `PlaybackManager` and don't emit playback events, so the screensaver's idle check had no way to know they were active.

Adds a block/unblock counter to the `ScreenSaverManager`. The photo, book, PDF, and comic players block the screensaver when they open and unblock it when they close, so it stays inhibited for as long as the viewer is open regardless of idle time.

### Issues
Fixes #7873

### Code assistance
None

---

### DIFF
--- a/src/plugins/bookPlayer/plugin.js
+++ b/src/plugins/bookPlayer/plugin.js
@@ -4,6 +4,7 @@ import Screenfull from 'screenfull';
 import { PluginType } from 'constants/pluginType';
 import { ServerConnections } from 'lib/jellyfin-apiclient';
 import browser from 'scripts/browser';
+import screenSaverManager from 'scripts/screensavermanager';
 import TouchHelper from 'scripts/touchHelper';
 
 import loading from '../../components/loading/loading';
@@ -61,6 +62,7 @@ export class BookPlayer {
         this.cancellationToken = false;
         this.loaded = false;
 
+        screenSaverManager.block();
         loading.show();
         const elem = this.createMediaElement(options);
         return this.setCurrentSrc(elem, options);
@@ -69,6 +71,7 @@ export class BookPlayer {
     stop() {
         this.unbindEvents();
         this.unmountBookOsd();
+        screenSaverManager.unblock();
 
         const stopInfo = {
             src: this.item
--- a/src/scripts/screensavermanager.js
+++ b/src/scripts/screensavermanager.js
@@ -49,6 +49,7 @@ function getScreensaverPlugin(isLoggedIn) {
 
 function ScreenSaverManager() {
     let activeScreenSaver;
+    let blockCount = 0;
 
     function showScreenSaver(screensaver) {
         if (activeScreenSaver) {
@@ -91,6 +92,22 @@ function ScreenSaverManager() {
         return activeScreenSaver != null;
     };
 
+    /**
+     * Prevents the screensaver from activating until a matching unblock() call.
+     */
+    this.block = () => {
+        blockCount++;
+    };
+
+    /**
+     * Releases a previous block(); the screensaver can activate once all blocks are cleared.
+     */
+    this.unblock = () => {
+        if (blockCount > 0) {
+            blockCount--;
+        }
+    };
+
     this.show = function () {
         let isLoggedIn;
         const apiClient = ServerConnections.currentApiClient();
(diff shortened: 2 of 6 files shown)


dmitrylyzo:
It can still activate the screensaver. Probably an edge case: screensaver time = 5s, wait some time (~1min).

I think it would be better to block the screensaver using the `ScreenSaverManager` instead of interfering with the `InputManager`. But the `PhotoPlayer` and book players work differently than the `VideoPlayer`. They don't generate playback events (`PlaybackStart`, `PlaybackStop`), the `PlaybackManager` doesn't set them as the current player.

Options:
1. Add "semaphore" (block/unblock) to the `ScreenSaverManager` and call them from the affected players.
2. Add events to the affected players. This will probably require making (a lot of) changes to the `PlaybackManager`.

dfederm:
> It can still activate the screensaver. Probably an edge case: screensaver time = 5s, wait some time (~1min).
> 
> I think it would be better to block the screensaver using the `ScreenSaverManager` instead of interfering with the `InputManager`. But the `PhotoPlayer` and book players work differently than the `VideoPlayer`. They don't generate playback events (`PlaybackStart`, `PlaybackStop`), the `PlaybackManager` doesn't set them as the current player.
> 
> Options:
> 
> 1. Add "semaphore" (block/unblock) to the `ScreenSaverManager` and call them from the affected players.
> 2. Add events to the affected players. This will probably require making (a lot of) changes to the `PlaybackManager`.

Thanks, this is a much cleaner approach.

src/plugins/bookPlayer/plugin.js
```diff
@@ -61,13 +62,15 @@ export class BookPlayer {
...
+        screenSaverManager.block();
         loading.show();
         const elem = this.createMediaElement(options);
         return this.setCurrentSrc(elem, options);
     }
 
     stop() {
         this.unbindEvents();
+        screenSaverManager.unblock();
         this.unmountBookOsd();
```
dmitrylyzo, line 73: nit - the same order in all players.
```suggestion
        this.unbindEvents();
        this.unmountBookOsd();
        screenSaverManager.unblock();
```

src/scripts/screensavermanager.js
```diff
@@ -91,6 +92,18 @@ function ScreenSaverManager() {
         return activeScreenSaver != null;
     };
 
+    // Prevents the screensaver from activating until a matching unblock() call.
```
dmitrylyzo: It is better to use JSDoc right away.
```suggestion
    /**
     * Prevents the screensaver from activating until a matching unblock() call.
     */
```

src/scripts/screensavermanager.js
```diff
@@ -91,6 +92,18 @@ function ScreenSaverManager() {
...
+    this.block = () => {
+        blockCount++;
+    };
+
+    // Releases a previous block(); the screensaver can activate once all blocks are cleared.
```
dmitrylyzo: same
```suggestion
    /**
     * Releases a previous block(); the screensaver can activate once all blocks are cleared.
     */
```

dmitrylyzo (commented):
LGTM.

I have another idea, but it doesn't have to be implemented right now.
With block/unblock, we can get a permanently blocked screensaver if something goes wrong (inconsistent block/unblock). To eliminate this, we can introduce a "ping" function, which will reset `screensavermanager.lastFunctionalEvent`. Then call it at some interval from the players and InputManager. The problem is, we have to ping screensavermanager faster than its idle time, which can be set pretty low (I'm testing with 5 seconds). We should probably increase the minimum value of screensaver idle time.

dmitrylyzo (approved):

PR #8041, opened 2026-06-15
Fix username validation on user create/edit

teach-moo:
The problem: 
We weren't validating usernames before sending them to the API.
If a user entered unsupported characters (/, +, &, spaces, etc.):

Create User -> generic error toast that doesn't tell you what's wrong.
Edit User -> spinner forever. No error. No way out besides leaving the page.

### Changes

**Fix**
- Added client-side username validation before the API call.
- Show a real error telling the user which characters are allowed.
- Stop the spinner immediately if validation fails.
- Added missing onError handlers to both mutations.
- Spinner now stops even if the API comes back with an unexpected error.

### Issues

Fixes #12480

### Code assistance

Both explanation of code base and debugging guidance were assisted with Claude code.
-->

---

* [ ] I have verified that this is not duplicating changes in an existing PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13208,6 +13208,21 @@
         }
       }
     },
+    "node_modules/html-encoding-sniffer/node_modules/@noble/hashes": {
+      "version": "2.2.0",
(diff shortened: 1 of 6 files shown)


thornbill:
The problem: slop PRs

PR #8046, opened 2026-06-16
Add selectable sub-menu to special features

wilburCforce:
### Changes
When browsing content with lots of extra features it is really difficult to find what you are looking for.  This change implements a menu button that presents a list of choices to filter the content on based on what that particular title has. It utilizes the existing elements in the API call so the content should be localized as a result. It also keeps the design consistent with other menu elements on the page. When you have 30 pieces of extra content this makes the scrolling functional.  This implements this feature request: https://features.jellyfin.org/posts/1940/selectable-sub-menu-for-special-features

This was tested in a live environment with titles that included 0, 1, and 3+ types of extra features.  This was also tested in various screen layouts to ensure this was responsive and screen friendly.
FILAMENT-GoogleChrome2026-06-1522-28-26-ezgif com-video-to-gif-converter

### Code assistance
AI was used to locate the impacted section quickly since this is my first contribution to the code base.  The CSS and design was implemented by me.

---

### DIFF
--- a/src/controllers/itemDetails/index.html
+++ b/src/controllers/itemDetails/index.html
@@ -212,7 +212,12 @@
             </div>
 
             <div id="specialsCollapsible" class="verticalSection detailVerticalSection hide">
-                <h2 class="sectionTitle sectionTitle-cards padded-right">${SpecialFeatures}</h2>
+                <div class="sectionTitleContainer specialsSectionTitleContainer">
+                    <h2 class="sectionTitle sectionTitle-cards padded-right">${SpecialFeatures}</h2>
+                    <button id="specialsFeaturesMenuButton" type="button" is="paper-icon-button-light" class="sectionTitleIconButton specialsFeaturesMenuButton hide" title="${ButtonMore}">
+                        <span class="material-icons more_vert" aria-hidden="true"></span>
+                    </button>
+                </div>
                 <div is="emby-scroller" class="padded-top-focusscale padded-bottom-focusscale no-padding" data-centerfocus="true">
                     <div id="specialsContent" is="emby-itemscontainer" class="scrollSlider focuscontainer-x itemsContainer"></div>
                 </div>
--- a/src/controllers/itemDetails/index.js
+++ b/src/controllers/itemDetails/index.js
@@ -18,6 +18,7 @@ import imageLoader from 'components/images/imageLoader';
 import itemContextMenu from 'components/itemContextMenu';
 import itemHelper from 'components/itemHelper';
 import mediaInfo from 'components/mediainfo/mediainfo';
+import actionSheet from 'components/actionSheet/actionSheet';
 import layoutManager from 'components/layoutManager';
 import listView from 'components/listview/listview';
 import loading from 'components/loading/loading';
@@ -45,6 +46,7 @@ import { OutboundWebSocketMessageType } from '@jellyfin/sdk/lib/websocket';
 import 'elements/emby-itemscontainer/emby-itemscontainer';
 import 'elements/emby-checkbox/emby-checkbox';
 import 'elements/emby-button/emby-button';
+import 'elements/emby-button/paper-icon-button-light';
 import 'elements/emby-playstatebutton/emby-playstatebutton';
 import 'elements/emby-ratingbutton/emby-ratingbutton';
 import 'elements/emby-scroller/emby-scroller';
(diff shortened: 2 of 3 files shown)


src/controllers/itemDetails/index.js
```diff
@@ -1767,8 +1769,58 @@ function getVideosHtml(items) {
...
+            }).then(function (type) {
+                displayFilteredSpecials(type);
+            }).catch(() => {
+                // Ignore when the user closes the menu without selecting
+            });
+        }
+
+        if (menuTypes.length > 1) {
+            specialsMenuButton.classList.remove('hide');
+            specialsMenuButton.addEventListener('click', showFilterMenu);
```
jordanmichaelrushing, line 1818: Minor: consider `specialsMenuButton.onclick = showFilterMenu` instead of `addEventListener`. Functionally equivalent today given Jellyfin's ring-buffer slot, but removes a re-render/leak if this code's call pattern ever changes, with no added complexity.

src/controllers/itemDetails/index.js
```diff
@@ -1767,8 +1769,58 @@ function getVideosHtml(items) {
...
+                offsetLeft: 0,
+                offsetTop: 0
+            }).then(function (type) {
+                displayFilteredSpecials(type);
+            }).catch(() => {
+                // Ignore when the user closes the menu without selecting
+            });
+        }
+
+        if (menuTypes.length > 1) {
```
jordanmichaelrushing, line 1816: I recommend checking `extraTypes` instead of `menuTypes`. As it stands if you have just one specials type, you'll have "All" & "Special X" which would be the same as far as I know. This way specials still load and only display the button if there are multiple types and not just one

src/controllers/itemDetails/index.js
```diff
@@ -1767,8 +1769,58 @@ function getVideosHtml(items) {
 function renderSpecials(page, item, user) {
     ServerConnections.getApiClient(item.ServerId).getSpecialFeatures(user.Id, item.Id).then(function (specials) {
         const specialsContent = page.querySelector('#specialsContent');
-        specialsContent.innerHTML = getVideosHtml(specials);
-        imageLoader.lazyChildren(specialsContent);
+        const specialsMenuButton = page.querySelector('#specialsFeaturesMenuButton');
+
+        const allSpecials = specials;
```
jordanmichaelrushing, line 1774: Why not just use specials directly?

src/styles/librarybrowser.scss
```diff
@@ -1245,7 +1245,19 @@ div.itemDetailGalleryLink.defaultCardBackground {
...
+    display: flex;
+    align-items: center;
+}
+
+.specialsFeaturesMenuButton {
+    margin-left: 0 !important;
+}
+
+#specialsFeaturesMenuButton .material-icons {
+  font-size: 2em !important;
```
jordanmichaelrushing, line 1258: If you scope this under .specialsSectionTitleContainer you shouldn't need the `!important`. Also I'd point you to my top html comment about the id and class clash

src/controllers/itemDetails/index.html
```diff
@@ -212,7 +212,12 @@ <h2 class="sectionTitle padded-right">${HeaderUpcomingOnTV}</h2>
             </div>
 
             <div id="specialsCollapsible" class="verticalSection detailVerticalSection hide">
-                <h2 class="sectionTitle sectionTitle-cards padded-right">${SpecialFeatures}</h2>
+                <div class="sectionTitleContainer specialsSectionTitleContainer">
+                    <h2 class="sectionTitle sectionTitle-cards padded-right">${SpecialFeatures}</h2>
+                    <button id="specialsFeaturesMenuButton" type="button" is="paper-icon-button-light" class="sectionTitleIconButton specialsFeaturesMenuButton hide" title="${ButtonMore}">
```
jordanmichaelrushing, line 217: Any particular reason there's both a specialsFeaturesMenuButton ID and a specialsFeaturesMenuButton class?

src/controllers/itemDetails/index.js
```diff
@@ -1767,8 +1769,58 @@ function getVideosHtml(items) {
...
+                positionTo: specialsMenuButton,
+                resolveOnClick: true,
+                menuItemClass: 'actionSheetMenuItem',
+                offsetLeft: 0,
+                offsetTop: 0
+            }).then(function (type) {
+                displayFilteredSpecials(type);
+            }).catch(() => {
+                // Ignore when the user closes the menu without selecting
+            });
```
jordanmichaelrushing, line 1813: Having a general catch at the end means it will also catch all errors, silently dropping them. Instead I recommend something like this:
```
actionSheet.show({
    ...
}).then(function (type) {
    displayFilteredSpecials(type);
}, () => {
    // Ignore when the user closes the menu without selecting
});
```

jordanmichaelrushing (commented):
I like the idea. A couple of tweaks and I think this would solid. FYI I'm not in the Jellyfin org, just someone who wants it as solid as possible. That's why I won't mark this as "Request changes". Great work.

src/controllers/itemDetails/index.js
wilburCforce: I see what you are getting at there, but that would be inconsistent with the rest of the page.  It appears that addEventListener is the preferred method based on other functions.  Change my mind though.

src/controllers/itemDetails/index.js
wilburCforce: from a stupid user perspective there is some satisfaction with knowing you are seeing all by clicking all.
I'm not set in stone on that, but having it render the same way every time as opposed to appearing and disappearing is more consistent to me as well.

src/controllers/itemDetails/index.js
wilburCforce: yeah that can be cleaned up.  I'll update that throughout.

PR #8047, opened 2026-06-16
improve font selection in epub files

dkanada:
### Changes

Font size handling is still a bit rough around the edges but this is a definite improvement. Previously it was common for our custom style to be overridden by more specific CSS rules.

### Issues

None

### Code Assistance

None


MBR-0001 (approved):
from my testing it works, the text is now more consistently sized

https://github.com/user-attachments/assets/d2934b58-9d6f-423a-94cf-8ebb8dc66146

PR #8048, opened 2026-06-16
Prevent invalid paths from being added to the list of folder paths

theguymadmax:
When creating a library, entering an invalid path would show an error, but the path was still added to the folder list.

### Changes
Validate the path against the API before adding it to the list. 

### Issues
Prevents:

### Code assistance
N/A

---


thornbill (commented):
We should never reference the global `ApiClient` in new code (ideally we wouldn't use the api client at all).

It also might read a bit easier if you make the function `async` and `await ServerConnections.getCurrentApiClientAsync().getDirectoryContents(path);` in a try/catch block.

thornbill (approved):

Issue #1421, opened 2020-06-16
Multiple video streams muxed in the same MKV file are not handled / selectable


Delgan:
**Describe the bug**
It's possible for a MKV file to contain several video tracks. However, the web client only shows one.

**To Reproduce**
1. Create a `.mkv` with multiple video tracks (eg. `ffmpeg -i input1.mp4 -i input2.mp4 -c copy -map 0 -map 1 output.mkv`)
2. Add the video file to the library
3. Go to the item details page on the web client
4. Notice that you can't select the video track to play

**Expected behavior**
All video tracks contained in the `.mkv` should be listed by the selector and usable by the player.

**System (please complete the following information):**
 - OS: Linux
 - Browser: Firefox
 - Jellyfin Version: 10.5.4

**Additional context**
Some remarks after discussion on Matrix:
* The selection [is purposefully disabled](https://github.com/jellyfin/jellyfin-web/blob/b57c6cacb805fc1bf10e34a4d5677c4f0c94295b/src/controllers/itemDetails.js#L195), it's been like this for years
* The selected video track is ignored in [the `play()` options](https://github.com/jellyfin/jellyfin-web/blob/b57c6cacb805fc1bf10e34a4d5677c4f0c94295b/src/controllers/itemDetails.js#L1952), thus enabling the `select` has no effect anyway
* It is therefore likely that it is also a problem related to the server not being able to handle multiple video streams

dominikh:
Shut up, bot.

itsTyrion:
anything? This would be useful for shows where creating a sub folder with multiple videos for every episode (I don't want/need 4k or so on mobile) is kinda chaotic/confusing(and VERY hit or miss)

Thalhammer:
> This would be super useful for shows.

Not really. All audio/video streams use the same "namespace" if you can call it that way. Jellyfin would have no clue which audio stream is related to which video stream. However it would make pretranscoding content to lower resolutions/bitrates/nonhdr much more painless because you don't have manage multiple files, don't need to group the videos in jellyfin and you don't need to store audio streams multiple times for each video. It would be even cooler if jellyfin could autoselect the best matching video stream depending on device caps, but thats a different issue.

It might be stupid to preserve all audio streams in the pretranscoded copy but I like to aim for the best quality possible (which is why I am CPU transcoding beforehand in the first place) and transcoding audio on the fly is trivial even on the cpu. In some cases the audio streams end up taking a significant chunk of the files.
[...]

PR #8051, opened 2026-06-17
Allow selecting a video stream before playback

ivenos:
### Changes

Some MKVs carry more than one video track in a single file (say a 4K and a 1080p version). The detail page already had a video dropdown, but it was hardcoded to disabled, so you could see the tracks and never pick one. This turns it on when there's more than one video track and sends the picked index as `videoStreamIndex` when you hit play, the same way the audio and subtitle dropdowns already do.

Needs the server side too: https://github.com/jellyfin/jellyfin/pull/17123

Before it was greyed out, now you can open it and pick a track:
Bildschirmfoto vom 2026-06-17 11-20-22

### Issues

Related to #1421

### Follow-up / known gap

This only covers the web client. The native apps (Android TV, Swiftfin, Roku, ...) have their own UI and won't pick this up for free. I did look at the Android TV app, and the spot that makes sense is a video picker right next to the version button on the detail screen. But getting the selected index threaded through its playback launch path was more than I felt sure about doing right and testing properly, so I left it out instead of shipping something half-baked. Happy to do it as a follow-up.

### Code assistance

I used Claude on this one: finding my way around the codebase, writing the change, and setting up a local test server. I went over everything myself and tested it (build, the existing StreamBuilder tests, and by hand with multi-track files and a movie that has two versions).

---

### DIFF
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -450,6 +450,9 @@ async function getPlaybackInfo(player, apiClient, item, deviceProfile, mediaSour
         query.AutoOpenLiveStream = false;
     }
 
+    if (options.videoStreamIndex != null) {
+        query.VideoStreamIndex = options.videoStreamIndex;
+    }
     if (options.audioStreamIndex != null) {
         query.AudioStreamIndex = options.audioStreamIndex;
     }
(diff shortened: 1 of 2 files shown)


src/components/playback/playbackmanager.js
```diff
@@ -550,6 +553,9 @@ function getLiveStream(player, apiClient, item, playSessionId, deviceProfile, me
     if (options.maxBitrate) {
         query.MaxStreamingBitrate = options.maxBitrate;
     }
+    if (options.videoStreamIndex != null) {
```
Grinv, line 556: Since you forward `VideoStreamIndex` into the `getLiveStream` query here, `getPlaybackMediaSource()` probably needs to null it out too. For a `RequiresOpening` source it resets `audioStreamIndex`/`subtitleStreamIndex` before calling `getLiveStream()` (the pre-open indices may not match the reopened stream layout), but not `videoStreamIndex`, so a stale index would still reach `OpenLiveStream`.

PR #8085, opened 2026-06-20
Handle partial now playing items

MenotiFilho:
### Changes

Handle partial `NowPlayingItem` objects in the remote control page. External or virtual playback items, when played through casting, may not include a populated `MediaStreams` array or a resolvable library item id. The remote control page now renders available title and image metadata before optional library lookups, guards the media stream type check, and skips library-only controls when the item cannot be resolved.

No screenshots provided because this fixes missing metadata rendering without changing the intended UI layout.

### Code assistance

Code assistance was used to inspect the remote control and Now Playing bar code paths, identify unsafe assumptions around partial `NowPlayingItem` objects, and draft the localized defensive checks. I tested the resulting change manually and ran `npm run lint -- src/components/remotecontrol/remotecontrol.js`.

---

* [x] I have provided a *substantive* review of [another web PR](https://github.com/jellyfin/jellyfin-web/pull/7425#issuecomment-4759766097).

### DIFF
--- a/src/components/remotecontrol/remotecontrol.js
+++ b/src/components/remotecontrol/remotecontrol.js
@@ -98,7 +98,10 @@ function updateNowPlayingInfo(context, state, serverId) {
         '';
     if (item) {
         const nowPlayingServerId = (item.ServerId || serverId);
-        if (item.Type == 'AudioBook' || item.Type == 'Audio' || item.MediaStreams[0].Type == 'Audio') {
+        const firstMediaStreamType = item.MediaStreams?.[0]?.Type;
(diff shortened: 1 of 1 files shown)


thornbill:
We are not going to accept any changes to support piracy plugins.

MenotiFilho:
> We are not going to accept any changes to support piracy plugins.

Understood, my intent was not to support any specific plugin, but to make the remote
control page handle partial NowPlayingItem objects consistently with
the bottom Now Playing bar. The crash comes from assuming
MediaStreams[0] exists before rendering the available metadata.

PR #8087, opened 2026-06-21
Preserve selected library view when switching libraries

Shadowghost:
### Changes
Preserve selected library view when switching libraries

### Code assistance
None

---


MBR-0001 (changes requested):
going from a movie library to playlists and back leaves you on the playlists view, which gives off buggy behavior vibes

Shadowghost:
@MBR-0001 so what do you want me to change?

MBR-0001:
I think something like this should be enough, I observed this behavior only with playlists
```js
if (view && !(view == LibraryTab.Playlists && collectionType == CollectionType.Playlists)) {
    setLastSelectedView(view);
}
```

MBR-0001 (approved):

thornbill:
This seems kind of weird to me? I would expect the default view when changing libraries but maybe that's just what I'm used to? 🤷‍♂️

Might be worth running it by Erik for ux input.

erikbucik:
This seems weird to me too (if I understand this right). Going to a new library is like a reset...

This behavior would treat library selection like a content filter, and not a server destination/folder with its own views.
^ Which we should discuss separately in meta

Shadowghost:
I'd propose to maybe save/restore per library?

PR #8092, opened 2026-06-22
Fix shuffle and repeat buttons in Remote Control

### LINKED ISSUE #8091: Loop and shuffle buttons don't do anything when using remote control

### Describe the bug

When casting to a remote device from the web client, clicking the loop or shuffle button in the bottom bar or the queue screen does nothing.

### Reproduction Steps

1. Play a track on another client using jellyfin web
2. From the controlling client, click the cast button and connect to the other client
3. Click on the loop or shuffle buttons
4. Nothing happens

### Expected/Actual behaviour

Loop and shuffle modes cycle like they do when playing locally

### Logs

_No response_

### Server version

10.11.11

### Web version

10.11.11

### Build version

10.11.11

### Platform

Linux

### Browser

Firefox

### Additional information

_No response_

wongcallum:
### Changes

I implemented `getRepeatMode` and `getQueueShuffleMode` in `SessionPlayer`, and set `PlaybackOrder` in `PlaybackManager.getPlayerState`.

`RepeatMode` was already used in both the client and server, but `PlaybackOrder` only existed in `PlayerStateInfo` on the server. I used `PlaybackOrder` since `ShuffleMode` would be ignored when sent to the server.
This required a mildly hacky ternary expression in both places to translate between the two which I can extract to a helper function if preferred.

Before:
before

After:
after

Please note that shuffling works but the queue currently does not refresh on the controlling client until other actions happen. I intend to raise a separate issue for this.

### Issues

Fixes #8091 

### Code assistance

Claude Opus 4.8 was used to identify the source of the issue and draft the initial fix.

---

* [ ] I have provided a *substantive* review of [another web PR](https://github.com/jellyfin/jellyfin-web/pulls?q=is%3Apr+is%3Aopen+-label%3Adependencies+-label%3Ablocked+-label%3Abackend+-label%3Ainvalid+-label%3A"merge+conflict"+-label%3Astale+-author%3A%40me). <sub><sup>I'll do that soon...</sup></sub>

### DIFF
--- a/src/components/playback/playbackmanager.js
+++ b/src/components/playback/playbackmanager.js
@@ -2176,6 +2176,8 @@ export class PlaybackManager {
                 state.PlayState.IsPaused = player.paused();
                 state.PlayState.RepeatMode = self.getRepeatMode(player);
                 state.PlayState.ShuffleMode = self.getQueueShuffleMode(player);
+                // Needed for remote control because ShuffleMode doesn't exist in PlayerStateInfo from the server
+                state.PlayState.PlaybackOrder = state.PlayState.ShuffleMode === 'Shuffle' ? 'Shuffle' : 'Default';
                 state.PlayState.MaxStreamingBitrate = self.getMaxStreamingBitrate(player);
 
                 state.PlayState.PositionTicks = getCurrentTicks(player);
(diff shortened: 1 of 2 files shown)


evanqhuang (approved):
The stubs for getRepeatMode and getQueueShuffleMode were just // not supported? — so they returned undefined and the buttons never reflected actual state. Reading from lastPlayerData.PlayState is the obvious fix and matches how isPlayingVideo/isPlayingAudio already work in the same class.

The PlaybackOrder line in playbackmanager.js is a bit subtle — the server sends PlaybackOrder in PlayerStateInfo rather than ShuffleMode, so without that translation the getter would always fall back to 'Sorted'. Makes sense.

Nit: the PlaybackOrder === 'Shuffle' ? 'Shuffle' : 'Sorted' mapping is implicit about what other PlaybackOrder values mean — a short comment there would help future readers. Minor though.

thornbill (changes requested):
We should use the appropriate enum values from the SDK for these instead of strings.
https://typescript-sdk.jellyfin.org/interfaces/generated-client.PlayerStateInfo.html

wongcallum:
My bad for not seeing that

thornbill (approved):

PR #8095, opened 2026-06-25
Fix Safari audio crackling by not over-reporting surround channels

evanqhuang:
### Changes

`getPhysicalAudioChannels()` infers that a device has surround speakers whenever the browser can decode AC3/EAC3, returning at least 6 channels. On desktop Safari with a stereo output device (`AudioContext.destination.maxChannelCount === 2`), this makes the server deliver a 5.1 stream that Safari then downmixes to stereo itself, producing intermittent crackling/popping throughout playback.

This is not specific to Jellyfin. The same artifact reproduces in Safari on Apple Silicon with independent multichannel AAC content (e.g. Fraunhofer's browser AAC multichannel test), does not occur in Chrome on the same machine, and does not occur when the audio is downmixed to stereo before Safari receives it. It appears to be an upstream Safari issue with multichannel audio; this change works around it by not advertising more channels than the output actually has.

For desktop Safari, when the real output is known to be stereo (`maxChannelCount <= 2`), trust it so the server performs the downmix cleanly instead of Safari. Macs connected to a surround device still report 6/8 channels and keep full surround. iOS was tested and is not affected, so the change is scoped to desktop Safari.

This replaces the earlier approach in this PR (removing AC3/EAC3 from `hlsInFmp4VideoAudioCodecs`); no codec support is removed.

### Testing

Same source (HEVC video + 5.1 AAC), Safari on an Apple Silicon Mac, unless noted. Only the audio channel count changes the outcome:

| Audio | Delivery | Result |
| --- | --- | --- |
| 5.1 (copy) | fMP4 HLS (`+frag_discont`) | crackle |
| 5.1 (copy) | fMP4 HLS (no `+frag_discont`) | crackle |
| 5.1 (copy) | MPEG-TS → hls.js/MSE | crackle |
| 5.1 (copy) | progressive MP4, direct play | crackle |
| 5.1 (re-encoded) | fMP4 HLS | crackle |
| 2.0 (downmixed) | fMP4 HLS | clean |
| 5.1 | in Chrome, same Mac | clean |
| 7.1 (Fraunhofer, non-Jellyfin) | native | crackle on AirPods Max |

Container, HLS fragmentation, the `+frag_discont` flag, and copy-vs-re-encode were all ruled out — the only variable that changes the result is channel count (with the output device / Spatial Audio state).

### Issues

Fixes jellyfin/jellyfin-web#8097

### Code assistance

Used AI tooling to generate the A/B test variants (remuxed HLS with different channel counts, containers, and codecs) and to help analyze the code path. I listened to each variant myself to identify which reproduced the crackling — that manual comparison is what isolated channel count as the trigger. The change itself is a small guard in the channel-count heuristic.

---

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -463,6 +463,13 @@ function getPhysicalAudioChannels(options, videoTestElement) {
 
     // AC3/EAC3 hinted that device is able to play dolby surround sound.
     if (isAc3Eac3Supported && isSurroundSoundSupportedBrowser) {
+        // Desktop Safari (Apple Silicon) crackles when it downmixes multichannel audio
+        // to a stereo output device. When the real output is known to be stereo, trust it
+        // so the server downmixes cleanly instead of Safari. jellyfin/jellyfin#15999
+        if (browser.safari && !browser.iOS && speakerCount > 0 && speakerCount <= 2) {
+            return speakerCount;
+        }
+
         return speakerCount > 6 ? speakerCount : 6;
     }
 


LTe:
Safari can support EAC3, so I'm not sure if we should remove that from the device profile. I was trying to reproduce this issue on the master branch but without issues. Can you provide media to reproduce this?

I was using `Safari 18.6 (20621.3.11.11.3)`

evanqhuang:
> Safari can support EAC3, so I'm not sure if we should remove that from the device profile. I was trying to reproduce this issue on the master branch but without issues. Can you provide media to reproduce this?
> 
> I was using `Safari 18.6 (20621.3.11.11.3)`

@LTe Thanks for the pushback, I've done some more A/B testing and I think I've narrowed it down. I suspect the issue is Apple silicon only (I'm on Safari 26.5 (21624.2.5.11.4)), and it's happening when `Maximum Allowed Audio Channels` is set to something higher than stereo (auto, 5.1, etc.). I think there might be a bug in Safari itself where it doesn't properly handle the downmixing. I've reworked the PR to drop the removal of AC3/EAC3 and add a different fix. 

With `Maximum Allowed Audio Channels` on `Auto`, `getPhysicalAudioChannels()` assumes surround speakers whenever the browser can decode AC3/EAC3, so it reports ≥6 channels even on a Mac whose actual output is stereo (`AudioContext.destination.maxChannelCount === 2`). The server then sends 5.1 and Safari does its own downmix, which appears to be broken (crackling/popping throughout). The fix is to trust `maxChannelCount <= 2` for desktop Safari instead of the too-broad AC3/EAC3 heuristic, so the server downmixes cleanly. Surround outputs (maxChannelCount 6/8) are unaffected.

Reproduction steps: Apple Silicon Mac, built-in/stereo output, Maximum Allowed Audio Channels = Auto, and media with a 5.1 (6-channel) AAC or EAC3 track, played in Safari desktop. I've also tested it here https://www2.iis.fraunhofer.de/AAC/multichannel.html and noticed the same issue when playing surround audio, so it does seem like somethings broken in Safari. I've submitted feedback to Apple but who knows when this will get resolved.

@austinbaccus or @gizeto Can you go to `Settings → Playback → Audio → Maximum Allowed Audio Channels` and confirm that when you experienced the issue you had it set to `Auto`? And then switch it to `Stereo` and test to see if that fixes the issue? Also can you confirm if you're using the built in speakers, bluetooth headphones, etc? I've noticed many more issues when wearing my Airpods Max and now I'm wondering if the bug is more nuanced. I toggled spatial audio on and off and I can't seem to reproduce the issue now. 

Seems like there's an issue where:

1. Jellyfin server has multichannel audio
2. Jellyfin web shows 2 channels (with Airpods connected in spatialize)
```
const c = new (window.AudioContext || window.webkitAudioContext)();
console.log('maxChannelCount:', c.destination.maxChannelCount);
"maxChannelCount:2"
```
3. Airpods have `Spatialize stereo` on and remuxes into surround.
I can get it to reliably trigger if I switch from spatialize "Head Tracking" -> "Fixed"

Not sure if the Airpods are an entirely different case at this point, I do remember the same issue happening on the built in speakers though.

nyanmisaka:
cc @gnattu

thornbill:
All of your comments read like pure llm output which would be a violation of our [llm policy](https://jellyfin.org/docs/general/contributing/llm-policies/)

gnattu:
I cannot reproduce the crack you mentioned with:
AirPods Pro 3 or MacBook Pro internal speaker on Safari 26.5.

LTe:
I tested again on `26.5.2 (21624.2.5.11.8)` on real content but without cracking (using speakers and AirPods Pro 2 with Spatialize stereo set to Head Tracking / Fixed and Off).

But I generated video

https://github.com/user-attachments/assets/73686df9-e8e3-44ed-a18f-6ef03f235d3f

and in jellyfin master I can hear a crackling sound. But this is only on Safari and only when I'm using AirPods.

```
ffprobe -v error -select_streams a -show_streams movie.mp4

[STREAM]
index=1
codec_name=aac
codec_long_name=AAC (Advanced Audio Coding)
profile=LC
codec_type=audio
codec_tag_string=mp4a
codec_tag=0x6134706d
sample_fmt=fltp
sample_rate=44100
channels=1
channel_layout=mono
bits_per_sample=0
initial_padding=0
id=0x2
r_frame_rate=0/0
avg_frame_rate=0/0
time_base=1/44100
start_pts=0
start_time=0.000000
duration_ts=1323000
duration=30.000000
bit_rate=127843
max_bit_rate=N/A
bits_per_raw_sample=N/A
nb_frames=1293
nb_read_frames=N/A
nb_read_packets=N/A
extradata_size=5
DISPOSITION:default=1
DISPOSITION:dub=0
DISPOSITION:original=0
DISPOSITION:comment=0
DISPOSITION:lyrics=0
DISPOSITION:karaoke=0
DISPOSITION:forced=0
DISPOSITION:hearing_impaired=0
DISPOSITION:visual_impaired=0
DISPOSITION:clean_effects=0
DISPOSITION:attached_pic=0
DISPOSITION:timed_thumbnails=0
DISPOSITION:non_diegetic=0
DISPOSITION:captions=0
DISPOSITION:descriptions=0
DISPOSITION:metadata=0
DISPOSITION:dependent=0
DISPOSITION:still_image=0
DISPOSITION:multilayer=0
TAG:language=und
TAG:handler_name=SoundHandler
TAG:vendor_id=[0][0][0][0]
[/STREAM]
```

For some reason I was able to reproduce this on iOS Simulator with speakers but not always (I need to restart simulator couple times to get the results).

https://github.com/user-attachments/assets/80023bdc-72a4-4329-8c7a-cbc401980a33

Even when there is a problem with a Safari I do not think jellyfin should fix that. This is a browser issue that should be fixed by the maintainers of the browser.

gnattu:
@LTe  Does direct play encounters this? If so it is not something really we can fix.

LTe:
> @LTe Does direct play encounters this? If so it is not something really we can fix.

Yes, it was direct play

LTe:
https://www.reddit.com/r/MacOS/comments/1njk2wn/crackling_or_popping_sound_from_speakers_on_macos/
https://www.reddit.com/r/MacOSBeta/comments/1l7wg5z/cracking_popping_sound_during_audio_playback_on/

For me, crackling sound was present when iOS Simulator was active. When you go to Simulator > I/O > Audio Output > [Pick another output] this should disappear.

I think this is totally macOS Tahoe 26 issue and should not be fixed inside jellyfin.

PR #8100, opened 2026-06-26
Clear the Alphabet filter when resetting filters

### LINKED ISSUE #8050: Clear `Alphabet` filter with `reset filters`

### Describe the bug

New feature Reset filter is awesome, if you don't know why a reduced amount of entries is shown. Sadly it won't clear the `Alphabet` filter as well.

Alphabet is also in local storage the indicator with a background image can easily be missed and storage is persistent. So switching library back and forth will not remove the selected value.

### Reproduction Steps

1. Filter with `Alphabet` `Q`
2. Switch library
3. Reset filters

### Expected/Actual behaviour

Clear all filters that limit amount of results in that library with `Reset filters` action

### Logs

_No response_

### Server version

12.0

### Web version

12.0

### Build version

Release

### Platform

Linux

### Browser

All of them

### Additional information

It is not in filter menu, but functionality is a filter and storage behavior is the same

Jian-Zhang08:
### Changes

`Reset filters` did not clear the **Alphabet** (A–Z) filter. `FilterDialog.resetQuery()` resets every other filter but never touched `query.NameStartsWith` (set by the alpha picker), which is persisted, so an A–Z selection survived a reset. `resetQuery()` now clears it too; on reload the controllers' `alphaPicker.updateControls(query)` clears the picker highlight.

### Issues

Fixes #8050

### Code assistance

Prepared with an AI coding agent (Claude): used to locate `resetQuery`, confirm `NameStartsWith` is the only unreset filter, and trace the `alphaPicker.updateControls` sync. Reviewed + verified before submitting.

---

* [x] I have verified this is not duplicating an existing PR.
* [ ] I have provided a substantive review of another web PR.

### DIFF
--- a/src/components/filterdialog/filterdialog.js
+++ b/src/components/filterdialog/filterdialog.js
@@ -469,6 +469,7 @@ class FilterDialog {
         query.HasTrailer = null;
         query.Tags = null;
         query.Years = '';
+        query.NameStartsWith = null;
     }
 
     show() {


MBR-0001:
This only affects the legacy layout, if we were to go with such a change it would have to be done on the new UI (primarily, with legacy being optional)

But even so, this doesn't even work without having a filter selected, otherwise the reset button can't be clicked, which makes me think you didn't even test this

enter-a-random-username:
@MBR-0001 in "modern" filter are still in sidebar and with backdrop active it is often really hard to spot.
But I understand the issue with active state and so on.

PR #8107, opened 2026-06-29
Add audiobook support for playback rate & chapters

jordanmichaelrushing:
### Changes
* Adds optional **speed** support for users to select non 1.0x playback rate, server side transcoding via atempo to be cross native (iOS breaks on html playbackrate for example) (0595e28273fbe5054d2e35620809f34c6b9a851c)
* Adds **chapter** support for baked in audiobook chapters (cc9a76d6ad66ebe5e516d2bec53e9aaed1921445)
    * Adds a new setting per user to decide if they want full audiobook view or chapter view (this is a highly contested UX in other apps doing the same thing so giving the User the decision felt incredibly necessary) 
* Adds **Chapter** support for multi-file audiobooks into 1 displayed item (2e562ec5b845463815352f250716f20c94b8f987)
    * Change Audiobook Item Page "Scenes" header to "Chapters" when dealing with a multi file audiobook (@dkanada has a solid implementation that adds a new dedicated section, I opted to reuse the specials section, but am happy to switch to their implementation, especially if theirs gets merged in first)
    * Multi-file audiobooks are also supported via a standardized "(^\d+)|(\d+$)" format (docs will need to be updated)
    * Display multifile artwork so crossing chapters (aka files) shows that files artwork (if different) (For Example: Graphic Audio has their files in parts and has unique art for each major part)
* Adds **Chapter** support for CUE sidecar files for Chapter info parsing (cc9a76d6ad66ebe5e516d2bec53e9aaed1921445)
    * Adds a new Book Playlist setting to determine if CUE chapters or baked in chapters were the default
* Misc: (6d3c7e1a306c6af06a33e81e0bd920b03b6b8a0a)
    * Expose the "Album" and "Album Artists" that are stored and viewable but not currently editable for Audiobooks in the Metadata Editor
    * Hide Artists in metadata editor as they aren't used for Audiobooks

### Issues
https://features.jellyfin.org/posts/590/adjust-audiobook-playback-speed
https://features.jellyfin.org/posts/1242/android-app-playback-speed-for-audio

### Code assistance
- Opus used to identify the locations I would need to target to match with the server change I currently have open
- Opus used to review the code changes

---

* [X] I have read and followed the [contributing guidelines](https://github.com/jellyfin/jellyfin-web/blob/master/CONTRIBUTING.md).
* [X] I have tested these changes.
* [X] I have verified that this is not duplicating changes in an existing PR.
[...]

PR #8112, opened 2026-06-30
basic chapter list for audiobook items

dkanada:
### Changes

Provides a very rudimentary chapter view for audiobooks now that jellyfin/jellyfin#16518 has been merged. I had grand plans for progress tracking but the itemDetails page isn't on React yet, so the required modifications were a pain.

I also tried a `list.js` replacement but that entire component probably needs to be rearchitected. Didn't want to tackle that at the same time.

### Issues

None

### Code Assistance

None

### DIFF
--- a/src/apps/legacy/controllers/itemDetails/index.js
+++ b/src/apps/legacy/controllers/itemDetails/index.js
@@ -22,6 +22,7 @@ import mediaInfo from 'components/mediainfo/mediainfo';
 import layoutManager from 'components/layoutManager';
 import listView from 'components/listview/listview';
 import loading from 'components/loading/loading';
+import ItemDetailsChapterList from 'components/itemDetails/ItemDetailsChapterList';
 import ItemDetailsMetadataList from 'components/itemDetails/ItemDetailsMetadataList';
 import { playbackManager } from 'components/playback/playbackmanager';
 import { appRouter } from 'components/router/appRouter';
@@ -577,7 +578,7 @@ function reloadFromItem(instance, page, params, item, user) {
 
     renderSeriesTimerEditor(page, item, apiClient, user);
     renderTimerEditor(page, item, apiClient, user);
-    setInitialCollapsibleState(page, item, apiClient, params.context, user);
+    setInitialCollapsibleState(page, instance, item, apiClient, params.context, user);
     const canPlay = reloadPlayButtons(page, item);
 
     setTrailerButtonVisibility(page, item);
@@ -818,7 +819,7 @@ function renderNextUp(page, item, user) {
     });
 }
 
-function setInitialCollapsibleState(page, item, apiClient, context, user) {
+function setInitialCollapsibleState(page, instance, item, apiClient, context, user) {
     page.querySelector('.collectionItems').innerHTML = '';
 
     if (item.Type == 'Playlist') {
@@ -848,6 +849,7 @@ function setInitialCollapsibleState(page, item, apiClient, context, user) {
         page.querySelector('.nextUpSection').classList.add('hide');
     }
 
+    renderChapters(page, instance, item);
     renderScenes(page, item);
 
     if (item.SpecialFeatureCount > 0) {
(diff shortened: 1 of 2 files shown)


jordanmichaelrushing:
Just so you're aware, I also made a [chapters PR](https://github.com/jellyfin/jellyfin-web/pull/8107) (amongst a bunch of things). To your specific change, I did a similar fix but ended up using the scenes section, but only on multi file audiobooks. To me this looks fine, if this gets merged in before mine I'll probably update mine to make sure multi-file uses your new section. Personally I have no preference either. Nice work

jordanmichaelrushing:
Played around with this and I believe there's a slight problem. When you're on the item page and edit the metadata it reloads and gives an error in the JS console and the section gets removed.

HTML wise all children of `#childrenCollapsible .itemsContainer` get removed, but both #childrenCollapsible & .itemsContainer do not get the `hide` class added, so it's not a can of being hidden, it's the html gets wiped out and not re-added (I assume because of the JS error).

image

jordanmichaelrushing:
Ok I figured out the issue:
`renderChapters` reuses same `.itemsContainer` on every metadata refetch, without unmounting prior root first. Two React roots fight over same node: removeChild crash. Also no else branch to add hide class back when chapters missing.

3 solutions:

1. Create a new dedicated section that isn't reusing another one (and follows the same structure as option 2 below)
2. Follow roughly the suggestion below (notice I'm setting hide when audiobook but no chapters & I'm setting the unmount instead of on `element` I'm setting it on the new renderTarget AND more importantly I'm setting innerHTML = '' which is helping to clear it out:
```
 function renderChapters(page, instance, item) {
     if (item.Type !== BaseItemKind.AudioBook) return;
     const container = page.querySelector('#childrenCollapsible');
     const element = container.querySelector('.itemsContainer');

     if (!item.Chapters?.length) {
        container.classList.add('hide');
        return;
    }

    element.innerHTML = '';
    const renderTarget = document.createElement('div');
    element.appendChild(renderTarget);

    instance._unmount.push(renderComponent(ItemDetailsChapterList, { item }, renderTarget));
    container.classList.remove('hide');
    element.classList.add('vertical-list');
}
```
3. (My preferred): The reason I used the `renderScenes` function in my PR is that it already calls `item.Chapters`. The only reason this doesn't currently work for single file audiobooks is because the embedded chapters don't have images. So if you simply change:
```
if (chapters.length && !chapters[0].ImageTag) {
```
to
```
if (chapters.length && !chapters[0].ImageTag && item.Type !== BaseItemKind.AudioBook) {
```
it automagically works. The only caveat is the look isn't the prettiest. To fix that go to chaptercardbuilder.js:
```
[line 71]: function getImgUrl({ Id, ImageTags }, { ImageTag }, index, maxWidth, apiClient) {
[lines 82-89]: 
    if (ImageTags?.Primary) {
        return apiClient.getScaledImageUrl(Id, {
            maxWidth: maxWidth,
            tag: ImageTags.Primary,
            type: 'Primary'
        });
    }
```
I know, "basic chapter list for audiobook items". That's why there's the other 2 options. That being said, this is IMHO still a small change and makes it look better. Options 1 & 2 look like this:
image

Option 3 without the image fix looks like this:
image

Option 3 with the image fix looks like this:
image

sevenrats:
#8207

dkanada:
@jordanmichaelrushing thanks for the analysis - pushed a small fix. @sevenrats your pull request is objectively more useful since it properly tracks playback state, but I'm going to try and get this into 12.0 before a released is tagged.

dkanada:
@sevenrats I forgot to mention the reason: your PR has ballooned to 3000 LOC so there's no chance for a 12.0 merge. We can shift gears once the new release comes out.

sevenrats:
> @sevenrats I forgot to mention the reason: your PR has ballooned to 3000 LOC so there's no chance for a 12.0 merge. We can shift gears once the new release comes out.

yeah that makes sense. if this direction changes please let me know asap because my PR is actually 1 of a series of PR's so if we are going to change direction I want as much notice as possible to try and preserve improvements that are downstream of my chapters PR.

PR #8159, opened 2026-07-04
Fix incorrect redirect after deleting an item from details page

### LINKED ISSUE #8106: [v12] Incorrect redirect after deleting an item from details page

### Describe the bug

When deleting an item from its details page, the application redirects to the previous library URL instead of the updated library URL.

### Reproduction Steps

Open a library (e.g. Shows).
Select a show to open its details page.
Click Delete and confirm the deletion.
Observe the page you are redirected to.

### Expected/Actual behaviour

After deleting an item from its details page, the application should redirect back to the current library URL.

Example:
/web/#/tv?topParentId=a656b907eb3a73532e40e44b968d0225&collectionType=tvshows

Actual behavior
After deleting an item, the application redirects to the legacy library URL instead.

Example:
/web/#/list?parentId=a123a0a04429ff3507afa806d87457ab&serverId=d5048463ce56458ca0edb183aa2b8e07

### Logs

_No response_

### Server version

12.0 RC2

### Web version

12.0 RC2

### Build version

12.0 RC2

### Platform

pc

### Browser

Brave

### Additional information

before deletion

Image

Image

after deletion

Image

Image

robert-md-or:
### Changes
When deleting an item from the details page, use appRouter.back() to return to the previous page instead of navigating to the parent item, which was generating a legacy URL.

### Issues

Fixes #8106

### Code assistance

Used Claude AI for codebase navigation. Fix and testing are my own.

**Before:**
Screenshot 2026-07-04 at 12 31 42

**After:**
Screenshot 2026-07-04 at 12 52 14

---


reecertc (changes requested):
`appRouter.back()` works when the item was opened from a library, but the previous page is not always a safe place to return to. The details page might have been opened from a bookmark, or the previous page might be an item that was also deleted.

Could we keep the existing parent/home navigation as a fallback when there is no safe page to return to? This would fix the redirect without leaving the user on a deleted item page.

robert-md-or:
Updated the fix to include a fallback navigation as suggested. Now uses appRouter.back() when history is available, falls back to parent item navigation, and goes home as last resort.

reecertc (changes requested):
One more issue: `canGoBack()` uses `window.history.length`, but opening the context menu and confirmation increases that value even after they close. This means a directly opened details page can still take the back branch instead of the new fallback, leaving the router stuck.

You can keep the fallback and just update `canGoBack()`

```diff
-return window.history.length > 1;
+return (window.history.state?.idx ?? 0) > 0;
```

Should be good to go after this 👍🏻

robert-md-or:
Updated canGoBack() to use window.history.state?.idx instead of window.history.length as suggested. Thanks for the catch!

reecertc (approved):

anultravioletaurora (approved):
lgtm - thanks for your contribution 🙂

src/components/router/appRouter.js
```diff
@@ -129,7 +129,7 @@ class AppRouter {
             return false;
         }
 
-        return window.history.length > 1;
+        return (window.history.state?.idx ?? 0) > 0;
```
thornbill, line 132: I can't find any reference that `idx` is a standard property of the history state object?

reecertc: Yeah, React Router adds idx internally, but it isn’t a standard property of the history state object. It works for now, but React Router could always change it in the future and break this, so you raise a fair point.

robert-md-or: You're right that `idx` isn't documented history state API — it's set internally by React Router and could change without notice. A couple of options to remove that dependency:

1. **Track visited `location.key` values** in a module-level Set, incremented on each navigation. `location.key` is a stable, public React Router property, so this avoids relying on internal state shape.
2. **Maintain our own navigation counter**, incremented in the same place we already listen for route changes, independent of anything React Router attaches to `history.state`.

Both avoid the undocumented `idx` field. I'd lean toward option 1 since it reuses an existing public property rather than adding new state to track, but happy to go with whichever you prefer.

thornbill: Per our [llm policy](https://jellyfin.org/docs/general/contributing/llm-policies/), communications must be **human** this is very clearly direct llm output...

PR #8161, opened 2026-07-06
fix eslint warnings and flag them as errors

dkanada:
### Changes

These have been bugging me for a while - the only remaining lint warnings are from nested functions but I'd like to fix those in a subsequent commit. I suppose FIXMEs and TODOs would make sense as a warning, but anything else only makes sense as an actual error in my opinion.

Grid2 seems to have been ignored due to Jellyfin Destkop v1 compatibility but since we have both v2 and v3 now I *think* it's fine to migrate? Unless v2 is also regarded as experimental.

### Issues

None

### Code Assistance

None


thornbill:
There is not a stable release of Jellyfin for Desktop v2 or v3

PR #8163, opened 2026-07-06
Use text numbers for audio rewind and fast forward instead of icon for all selectable values

jordanmichaelrushing:
### Changes
* The audio player has allows for users to specify rewind and fast forward times of 5, 10, 15, 20, 25 & 30 seconds. However the rewind icon is hard coded at 10 and the fast forward is hardcoded to 30. This change skips the built in icons and builds them on the fly.
* I also reordered the classes in the rewind button to match the order in the fast forward button as well removing the following unused classes (unused as far as I can see):
```
.btnNowPlayingRewind 
.btnNowPlayingFastForward 
```
* I added a const of `msPerSec` as there was already a 1000ms usage in that file

### Issues
I couldn't find any. Just me and my server users issue that's been flagged. Happy to create an issue if needed.

### Code assistance
None for this change, pretty straight forward change

---

### DIFF
--- a/src/apps/legacy/controllers/playback/queue/index.html
+++ b/src/apps/legacy/controllers/playback/queue/index.html
@@ -35,8 +35,9 @@
                             <span class="material-icons repeat" aria-hidden="true"></span>
                         </button>
 
-                        <button is="paper-icon-button-light" class="btnRewind btnNowPlayingRewind btnPlayStateCommand autoSize" title="${Rewind}">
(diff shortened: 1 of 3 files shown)


jordanmichaelrushing:
image
Manually modified the HTML to show that this is only using the arrow icon with a simple number, in this case a "17" that isn't selectable (no this doesn't modify the FF to 17 seconds, just using it as an example to show this working)

(Also please ignore the chapters lines, that's just my local NAS with added features that are not a part of this change)

PR #8086, opened 2026-06-21
Improve Upcoming view

Shadowghost:
Right now we limit the upcoming view to a certain unconfigurable date range.

### Changes
* Infinite scroll the upcoming view
* Show episode title on upcoming cards (aligned with legacy view, seems like this was missed when porting)
* Remove vertical scroll and instead wrap the cards

### Code assistance
Claude Opus 4.8 was used for getting some ideas on how to implement infinite scrolling correctly and for reviewing before committing.

---

* [x] I have read and followed the [contributing guidelines](https://github.com/jellyfin/jellyfin-web/blob/master/CONTRIBUTING.md).
* [x] I have tested these changes.
* [x] I have verified that this is not duplicating changes in an existing PR.
* [ ] I have provided a *substantive* review of [another web PR](https://github.com/jellyfin/jellyfin-web/pulls?q=is%3Apr+is%3Aopen+-label%3Adependencies+-label%3Ablocked+-label%3Abackend+-label%3Ainvalid+-label%3A"merge+conflict"+-label%3Astale+-author%3A%40me).

PR #8167, opened 2026-07-07
Add Alpha Picker to Genres view and implement infinite scroll

Shadowghost:
If you have a lot of genres, web is loading all of them right now which makes the page unusable.

### Changes
* Add alpha picker to genre view
* Make the genre view infinite scrollable with paginated loading

### Issues
Helps with: https://github.com/jellyfin/jellyfin/issues/17253

### Code assistance
Claude Opus 4.8 for local review

---


src/apps/modern/features/libraries/components/GenresItemsContainer.tsx
```diff
@@ -17,27 +19,76 @@ const GenresItemsContainer: FC<GenresItemsContainerProps> = ({
...
+        observer.observe(target);
+        return () => observer.disconnect();
+    }, [fetchNextPage, hasNextPage, isFetchingNextPage]);
+
+    const genres = useMemo(
+        () => data?.pages.flatMap((page) => page?.Items ?? []) ?? [],
+        [data]
+    );
 
     if (isLoading) {
```
Grinv: The picker is rendered below this early-return, so the first time you click a not-yet-cached letter, `isLoading` goes true and this returns only `<Loading />`. That unmounts the picker you just clicked until the fetch finishes. `ItemsView.tsx` keeps the picker above the loading branch and only swaps the list for the spinner, so it doesn't flash. Same here would keep the picker from disappearing on each new letter.

Shadowghost: should be fixed

src/apps/modern/features/libraries/components/AlphabetPicker.tsx
```diff
@@ -56,36 +49,44 @@ const AlphabetPicker: React.FC<AlphabetPickerProps> = ({
...
         >
-            <ToggleButtonGroup
-                orientation='vertical'
-                value={libraryViewSettings.Alphabet}
-                exclusive
-                color='primary'
-                size='small'
-                onChange={handleValue}
+            <Paper
+                elevation={4}
```
thornbill: Still not sure I love this but at least it's an improvement and I really don't know a better solution 😅 

Maybe just use `elevation={0}`? It seems a bit less obtrusive? 🤷‍♂️

src/apps/modern/features/libraries/components/GenresItemsContainer.tsx
```diff
@@ -17,27 +19,82 @@ const GenresItemsContainer: FC<GenresItemsContainerProps> = ({
...
-    if (isLoading) {
-        return <Loading />;
-    }
+    const observerTarget = useRef<HTMLDivElement>(null);
+
+    useEffect(() => {
+        const target = observerTarget.current;
+        if (!target) return;
+
+        const observer = new IntersectionObserver(
```
thornbill: Similar feedback to https://github.com/jellyfin/jellyfin-web/pull/8086, please use the hook for this...

src/hooks/useFetchItems.ts
```diff
@@ -97,12 +105,20 @@ const fetchGetGenres = async (
     }
 };
 
-export const useGetGenres = (itemType: BaseItemKind[], parentId: ParentId) => {
+export const useGetGenres = (
```
thornbill: ...and migrate this out of the `useFetchItems` file

src/apps/modern/features/libraries/components/AlphabetPicker.tsx
theguymadmax: Agreed, on both points. elevation={0}, makes the picker "stand out" less.

theguymadmax (approved):
The Picker now stands out less.  I still don't like the picker directly over the posters on the genres page, but at least the paper background make it more legible. I think this will do until someone comes up with a better solution.

thornbill (approved):

PR #8170, opened 2026-07-08
Avoid fetching Live TV route key as toolbar item

### LINKED ISSUE #8168: [RC2] web makes request to `/Items/livetv?userId=xyz` that fails with HTTP400 "The value 'livetv' is not valid."

### Describe the bug

When browsing `/web/#/livetv?collectionType=livetv&tab=2` page i see such request in dev tools / network:

Request URL https://jel.org/Items/livetv?userId=xyz
Request Method GET
Status Code 400 Bad Request

Response:
```
{
    "type": "https://tools.ietf.org/html/rfc9110#section-15.5.1",
    "title": "One or more validation errors occurred.",
    "status": 400,
    "errors": {
        "itemId": [
            "The value \u0027livetv\u0027 is not valid."
        ]
    },
    "traceId": "00-cd0530b4a196bbb1e85434d6ac4d663f-83207d67d62c084e-00"
}
```

### Reproduction Steps

1. Open devtools / network / xhr
2. Open `/web/#/livetv?collectionType=livetv&tab=2`
3. See request with HTTP400

### Expected/Actual behaviour

No request or one that does not fail?

### Logs

_No response_

### Server version

12.0

### Web version

12.0

### Build version

12.0-rc2

### Platform

Windows

### Browser

Chrome

### Additional information

_No response_

reecertc:
### Changes

The modern library toolbar now only fetches the parent item when item-backed controls such as Play All, Shuffle, or Queue can use it.

Live TV uses a synthetic route key as its parent id, so the toolbar no longer treats it as an item id and no longer requests `/Items/livetv`.

Adds focused unit coverage for the toolbar parent-item selection logic.

### Issues

Fixes #8168

### Code assistance

Used Codex and Claude to review.

---

* [x] I have provided a substantive review of another web PR.

### DIFF
--- a/src/apps/modern/features/libraries/utils/toolbar.ts
+++ b/src/apps/modern/features/libraries/utils/toolbar.ts
@@ -1,0 +1,32 @@
+import { CollectionType } from '@jellyfin/sdk/lib/generated-client/models/collection-type';
+
+import type { ParentId } from 'types/library';
+
+interface GetToolbarParentItemIdOptions {
+    parentId: ParentId;
+    collectionType?: CollectionType;
+    isBtnPlayAllEnabled?: boolean;
+    isBtnQueueEnabled?: boolean;
+    isBtnShuffleEnabled?: boolean;
+}
+
+export function getToolbarParentItemId({
+    parentId,
+    collectionType,
+    isBtnPlayAllEnabled,
+    isBtnQueueEnabled,
+    isBtnShuffleEnabled
+}: GetToolbarParentItemIdOptions): string | undefined {
+    const needsParentItem = isBtnPlayAllEnabled || isBtnShuffleEnabled || isBtnQueueEnabled;
+
+    if (!needsParentItem || !parentId) {
+        return undefined;
+    }
+
+    // Live TV uses a synthetic route key as parentId, not a real library item id.
+    if (collectionType === CollectionType.Livetv) {
+        return undefined;
+    }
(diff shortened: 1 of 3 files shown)


src/apps/modern/features/libraries/utils/toolbar.ts
```diff
@@ -0,0 +1,32 @@
...
+    const needsParentItem = isBtnPlayAllEnabled || isBtnShuffleEnabled || isBtnQueueEnabled;
+
+    if (!needsParentItem || !parentId) {
+        return undefined;
+    }
+
+    // Live TV uses a synthetic route key as parentId, not a real library item id.
+    if (collectionType === CollectionType.Livetv) {
+        return undefined;
+    }
```
Chris4942, line 29: nit: consider condensing these all into one
```suggestion
    if (
        !needsParentItem 
        || !parentId
        || collectionType === CollectionType.Livetv // Live TV uses a synthetic route key as parentId, not a real library item id.
    ) {
        return undefined;
    }
```

Or don't what you have is also fine

Chris4942 (approved):
I came here to "provide a substantive review of another web PR." but there's really not much to be said.

I'm not familiar with when this error occurs exactly, and there wasn't a video in the issue either. It might be nice to include a video either here or there, but maybe that's not necessary if this would be common knowledge if I had been working on this longer.

PR #8173, opened 2026-07-08
Refactor purple haze to use base theme

thornbill:
### Changes
Refactors the purple haze theme to use the base theme.
- Adds support for gradients and rounded person cards to base theme
- Adjusts text colors for contrast
- Makes app bar and dialogs more purple to better match legacy app

Screenshot 2026-07-08 at 17-15-16 Display

### Issues
None

### Code assistance
- Claude Sonnet was used to perform the bulk of the refactor with manual validation and cleanup
- Zed autocomplete

---

### DIFF
--- a/src/themes/_base/_theme.scss
+++ b/src/themes/_base/_theme.scss
@@ -2,6 +2,8 @@
 
 // Theme options
 $dark: true !default;
+// Enable rounded person cards (e.g. for actors/directors)
+$rounded-person-cards: false !default;
 
 // Color palette
 $common-white: palette.$common-white !default;
@@ -39,11 +41,23 @@ $button-inheritContainedBg: #424242 !default;
 $button-inheritContainedHoverBg: #616161 !default;
 $snackbarContent-bg: #303030 !default;
 $snackbarContent-color: rgba(255, 255, 255, 0.87) !default;
+// Secondary/accent color used for focus and interactive states.
+// Defaults to $primary-main so all existing themes are unaffected.
+$secondary-main: $primary-main !default;
+$secondary-contrastText: $primary-contrastText !default;
+// Background for legacy Jellyfin surface elements (paperList, visualCardBox, etc.).
+// Kept separate from $background-paper so themes can use a translucent overlay here
+// while $background-paper stays solid for MUI components (menus, dropdowns).
+$surface-overlay: $background-paper !default;
+// Optional CSS gradient applied over $appBar-defaultBg on the header and detail ribbon.
+// Set to a CSS gradient value to enable; leave as none to use only the solid colour.
+$appBar-gradient: none !default;
 
 // Assign custom css variables not part of the MUI theme palette
 :root {
     --jf-palette-background-defaultImage: #{$background-defaultImage};
     --jf-palette-AppBar-transparentBg: #{$appBar-transparentBg};
+    --jf-palette-AppBar-gradient: #{$appBar-gradient};
     --jf-card-borderRadius: #{$card-borderRadius};
 }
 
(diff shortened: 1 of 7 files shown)


nielsvanvelzen:
The border color of the inputs in the user login are now purple, even when the theme is set to dark (default). Is that intended?

thornbill:
OK that's due to adding support for the secondary color. Our default MUI theme sets it to purple but I don't think that is used anywhere...

src/themes/_base/_theme.scss
```diff
@@ -115,6 +136,11 @@ html {
     opacity: 0.86;
 }
 
+.dialog,
+.nowPlayingContextMenu {
+    background-color: var(--jf-palette-background-default, $background-default);
+}
```
anagnorisis2peripeteia:
```suggestion
.dialog,
.nowPlayingContextMenu {
    background-color: var(--jf-palette-background-default, $background-default);
    background-image: var(--jf-palette-background-defaultImage, $background-defaultImage);
    background-position: top center;
    background-repeat: no-repeat;
    background-size: cover;
}
```

src/themes/purplehaze/theme.scss
```diff
@@ -1,193 +1,104 @@
...
+    $filledInput-bg: $surface,
+    $filledInput-borderColor: transparent,
+    $card-borderRadius: 0.8em,
+    $alert-infoFilledBg: $info-background,
+    $alert-infoFilledColor: #0e0f2d,
+    $button-inheritContainedBg: $surface,
+    $button-inheritContainedHoverBg: $teal-bright,
+    $snackbarContent-bg: #303030,
+    $snackbarContent-color: rgba(255, 255, 255, 0.87)
+);
```
anagnorisis2peripeteia, line 55:
```suggestion
);

.dialog,
.nowPlayingPlaylist,
.nowPlayingContextMenu {
    background-image: none;
}
```

src/themes/_base/_theme.scss
anagnorisis2peripeteia: the assumption made here is that moving these dialog rules into the base theme will preserve the other themes behaviour. thats incorrect: blue radiance and wmc use background-defaultImage for their image/gradient backgrounds. restoring that handling in base and disabling it locally (see below) in purple haze is needed to prevent unrelated themes from losing their existing dialog backgrounds.

anagnorisis2peripeteia:
@thornbill left a review, looks like some sonnet slop to me here...

src/themes/_base/_theme.scss
thornbill: Except that was the wrong behavior. The context menus should not have used the background image in any theme.

anagnorisis2peripeteia: "Wrong in any theme" — yet your diff keeps `background-image` on `.nowPlayingPlaylist`, the panel right beside `.nowPlayingContextMenu` in the same now-playing view, so this doesn't remove a stray image, it mismatches two adjacent panels in blueradiance/wmc. And the same rule strips `.dialog`'s background for those themes too (which this PR never opens, unlike appletv) — restore the per-theme handling my suggestion added, or make the removal an explicit, maintainer-approved change rather than a side-effect.

nielsvanvelzen: Review our LLM guidelines and stop commenting with AI.

thornbill: It seems you were just hallucinating since `.nowPlayingContextMenu` appears to be entirely unused. Likely just a remnant from a prior refactor.

anultravioletaurora (approved):
moar purple 💜 🚀

PR #8187, opened 2026-07-09
Fix folder view card titles opening the typed library view

### LINKED ISSUE #5956: Clicking library title in folder view doesn't open folder view
**Describe The Bug**
When clicking the library title in the folder view it opens the normal library view instead of the folder view. Clicking the image correctly opens the folder view.

**Steps To Reproduce**
1. Open "Folders" on the home screen
2. Click the title of a library

**Expected Behavior**
Clicking the title should open the folder view just like clicking the image does.

 - Platform: Windows
 - Browser: Brave
 - Jellyfin Version: 10.9.9

TowyTowy:
### Changes

In the Folders view, clicking a library card's title opened the typed library view (e.g. `#/movies`) while clicking the card image opened the generic folder view (`#/list`). The image link is dispatched through `shortcuts.js`, which forwards the card's router `context` (`data-context`) to `appRouter.showItem`, but the title link built by `getTextActionButton` called `appRouter.getRouteUrl` without any context. `getRouteUrl` only skips typed-view routing when `context === 'folders'`, so the title resolved to a different page than the image.

This threads the card's `options.context` through `getTextActionButton` into `getRouteUrl`, so a card's title links to the same view as its image. When no context is set (all other views) the resolved URL is unchanged.

### Issues

Fixes #5956

### Code assistance

Developed with the assistance of an LLM (Claude). It was used to locate the root cause across `cardBuilder.js` / `shortcuts.js` / `appRouter.js`, implement the one-file change, and run ESLint and the cardbuilder Vitest suite (113 passed). The change and its (nil) impact outside the folder view are reviewed and understood.

---

### DIFF
--- a/src/components/cardbuilder/cardBuilder.js
+++ b/src/components/cardbuilder/cardBuilder.js
@@ -415,7 +415,7 @@ function getCardFooterText(item, apiClient, options, footerClass, progressHtml, 
             Type: item.Type,
             CollectionType: item.CollectionType,
             IsFolder: item.IsFolder
-        }));
+        }, null, null, options.context));
     }
 
     if (showOtherText) {
@@ -609,9 +609,10 @@ function getCardFooterText(item, apiClient, options, footerClass, progressHtml, 
  * @param {Object} item - Item used to generate the action button.
  * @param {string} text - Text of the action button.
  * @param {string} serverId - ID of the server.
+ * @param {string} [context] - Router context (e.g. `folders`) used to resolve the link URL, so the title links to the same view as the card image.
  * @returns {string} HTML markup of the action button.
  */
-function getTextActionButton(item, text, serverId) {
+function getTextActionButton(item, text, serverId, context) {
     if (!text) {
         text = itemHelper.getDisplayName(item);
     }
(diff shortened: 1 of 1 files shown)


Chris4942 (commented):
This does appear to be the correct behavior in the case that we may sometimes want clicking on the text to do something than clicking on the body of the card. I don't know if there is an instance where we would want that to be the case. If there are no instances where clicking on the body of the card should do something different than clicking on the title, then we should recycle the action from the card itself instead of creating it separately here.

Also, could you add a video show that the same thing happens when you click on the card and on the title?

TowyTowy:
Sorry for the slow reply.

On recycling the card's action: I don't know of a case where the title should go somewhere different from the card, so I agree they should never diverge. The reason I passed context in rather than reusing the card's action is that the title is its own anchor, so it needs its own href for middle-click and open-in-new-tab, and the only thing it was missing was the context the card already had. Making both derive from a single computed URL is the tidier fix but a wider change to cardBuilder. Happy to do it that way if you'd prefer, just say which you want.

On the video: I can't record one, I don't have a Jellyfin server with a library set up to demo against. The repro from #5956 is Home > Folders > click a library title versus its image. If a video is a blocker for merging, that's fair, and someone with a running instance would need to grab it.

PR #8188, opened 2026-07-09
Fix dark icons on card hover in light themes

thornbill:
### Changes
Fixes the action buttons on cards having dark icons when using a light theme making them impossible to see

Screenshot 2026-07-09 at 15-48-50 Display

### Issues
N/A

### Code assistance
Zed autocomplete

---

### DIFF
--- a/src/components/cardbuilder/card.scss
+++ b/src/components/cardbuilder/card.scss
@@ -448,6 +448,11 @@ button::-moz-focus-inner {
     z-index: 1;
     padding: 0.75em;
     font-size: 88%;
+
+    /* The "modern" layout needs additional specificity to prevent MUI styles from overriding */
+    &.paper-icon-button-light {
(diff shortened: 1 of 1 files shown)


Chris4942 (approved):
In what context does the bug you showed appear? When I ran off of `master`, the `color` for `paper-icon-button-light` was either `hsla(0, 0%, 100%, .76);` or `rgba(255, 255, 255, 0.76);` depending on context.

However, I don't see any reason we wouldn't want to set it here instead of inheriting to make sure we catch all the cases. However, I think we should keep the alpha value at `0.76` so the white doesn't end up to harsh against a dark background.

thornbill:
That was a good catch... the dark layouts using `#fff` was actually another symptom of this bug where MUI styles override the card styling.

MBR-0001 (approved):

PR #8191, opened 2026-07-10
Add type checking to js files

Chris4942:
### Changes

Type checking is completely turned off in all js files. This means that even if you add JSDoc comments to js files, they won't be enforced by anything unless imported by a typescript file. This inhibits the helpfulness of the type checker. 

This PR
- turns on type checking for js files
- Adds a `// @ts-nocheck` to the top of all `.js` and `.jsx` files

Aside from the `tsconfig.json` changes, all changes were made progmatically with
```bash
fd . src/ -e \.js -e \.jsx -x sed -i '1i// @ts-nocheck'
```

By adding annotations like this instead of turning off type checking as a blanket statement, we can add validated annotations to one file (or even line) at a time.

This does not replace the longer term goal of migrating to ts, but it provides an easier to achieve intermediate step where we can get many of the benefits of ts without all the work of migrating all the way to ts.

### Code assistance
None, just me and my language server.

---

    - no functional changes made, but I launched the app anyway and poked around a little bit to make sure nothing broke
    - https://github.com/jellyfin/jellyfin-web/pull/8187

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -413,7 +413,10 @@ export default tseslint.config(
         rules: {
             '@typescript-eslint/no-floating-promises': 'off',
             '@typescript-eslint/no-this-alias': 'off',
-
+            '@typescript-eslint/ban-ts-comment': [
+                'error',
+                { 'ts-nocheck': false }
+            ],
             'sonarjs/public-static-readonly': 'off',
 
             // TODO: Enable the following rules and fix issues
--- a/src/apps/dashboard/controllers/livetvguideprovider.js
+++ b/src/apps/dashboard/controllers/livetvguideprovider.js
@@ -1,3 +1,4 @@
+// @ts-nocheck
 import loading from 'components/loading/loading';
 import globalize from 'lib/globalize';
 import Dashboard, { pageIdOn } from 'utils/dashboard';
--- a/src/apps/dashboard/controllers/livetvtuner.js
+++ b/src/apps/dashboard/controllers/livetvtuner.js
@@ -1,3 +1,4 @@
+// @ts-nocheck
 import globalize from 'lib/globalize';
 import loading from 'components/loading/loading';
 import dom from 'utils/dom';
--- a/src/apps/legacy/controllers/edititemmetadata.js
+++ b/src/apps/legacy/controllers/edititemmetadata.js
@@ -1,3 +1,4 @@
+// @ts-nocheck
 import loading from 'components/loading/loading';
 import { getCurrentItemId, setCurrentItemId } from 'scripts/editorsidebar';
 
--- a/src/apps/legacy/controllers/favorites.js
+++ b/src/apps/legacy/controllers/favorites.js
@@ -1,3 +1,4 @@
+// @ts-nocheck
 import { BaseItemKind } from '@jellyfin/sdk/lib/generated-client/models/base-item-kind';
 import { ImageType } from '@jellyfin/sdk/lib/generated-client/models/image-type';
 import { ItemFields } from '@jellyfin/sdk/lib/generated-client/models/item-fields';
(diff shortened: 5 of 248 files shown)


reecertc (changes requested):
This will fail lint once CI runs (it's currently waiting on workflow approval, so only SonarCloud has run) because @typescript-eslint/ban-ts-comment is enabled and rejects every new @ts-nocheck directive:

```text
error  Do not use "@ts-nocheck" because it alters compilation errors  @typescript-eslint/ban-ts-comment
```
The simplest fix is to add this to the existing src/**/*.{js,jsx} override in eslint.config.mjs:

```js
'@typescript-eslint/ban-ts-comment': [
    'error',
    { 'ts-nocheck': false }
]
```
Keeping this scoped to JS/JSX means the rule will fire again when a file is migrated to TypeScript, preventing the suppression from surviving the migration.

Chris4942:
@reecertc Updated as you suggested. Good catch.

Side questions:
- Is there a magic command that runs all the CI checks together that I can run locally?
- I squashed all my commits into one even after you suggested changes. Is that the desired behavior? It looks like we don't have github squash and merge and in the git history, merge commits are usually just one commit. I'm also happy to just pile commits on top of each other as I make changes, but I wasn't sure what was desired

reecertc (approved):
@Chris4942 That's better.

There isn’t currently one local command covering every CI check. After npm ci, the quality-check commands are:

```
npm run build:es-check
npm run lint
npm run stylelint
npm run build:check
npm test
```

you don’t need to squash and force-push after each review change. Adding follow-up commits while review is ongoing makes the changes easier to inspect. They can be squashed near the end if you want. PRs are merged with merge commits here, so your PR commits remain grouped beneath that merge commit.

thornbill:
> Adding follow-up commits while review is ongoing makes the changes easier to inspect.

This is true. Generally rebasing should be avoided once a review starts unless it is a very small change.

> They can be squashed near the end if you want.

No, do not do this unless explicitly asked to do so by a maintainer. 

> PRs are merged with merge commits here, so your PR commits remain grouped beneath that merge commit.

Mostly true, but if a PR contains a large number of commits relative to the change or is otherwise considered to have a messy history, the maintainer merging it might do a squash merge.

---

Regarding these changes, what is the advantage to enabling your checking for js files instead of incrementally migrating them to typescript?

Chris4942:
@thornbill The advantage is that it allows for files to be migrated more gradually than a full typescript migration. Types can be added with all the types existing purely in comments so no logic is touched at all. It also allows for adding types to a part of file instead of the whole file at once.

In the end, a full type script migration is preferable, but enabling type checking in js files would lower the barrier to starting.

reecertc:
I guess the main benefit is that it makes type checking the default for JavaScript rather than something contributors have to opt into file by file.

Chris4942:
@thornbill I didn't realize how simple it would be to migrate to typescript. It really doesn't add very much overhead. I did [this PR](https://github.com/jellyfin/jellyfin-web/pull/8228) as a POC to compare, but I think it looks a lot nicer than the JSDoc one. I think that turning on type checking in js files would still have value, but this PR isn't blocking that effort

Chris4942:
[This](https://github.com/jellyfin/jellyfin-web/pull/8226/changes#diff-8c3c0432da202e7c3dde54580d15f1c908d6502060ba051373c4248ee8d630ef) is an example of where having type checking in js would be helpful. In this PR, the author wants to add type hints to `getIcon` in a `.js` file since they're already touching the function; however, since the file is in `.js` they can't do that without moving `getIcon` to another file. This ends up creating a weird file structure where that one function lives alone in its own file without having a good reason to live apart from the rest of the code besides type checking. It could be left in the `.js` file and still have type checking if we merge these changes which would lead to a smaller diff on that PR until someone takes on the full `.ts` migration.

PR #8194, opened 2026-07-10
Base desktop browser HDR support on display capability

KunJakob:
### Changes

I made this change as playback of (HEVC) HDR files through a chromium browser on a SDR display, would fail to transcode and tone map the file, as it would be perceived as an HDR capable display, despite HDR non being enabled in windows. This adds a media query backup to check the browser's self reported dynamic range, and won't report as HDR if the media query returns standard.

I also added the equivalent check for Firefox with their upcoming HDR in windows support (see comment)
### Code assistance

I used GH Copilot as semantic search for getting familiar with the codebase as well as researching how to determine HDR capabilities in browser API's. I also have LLM tab completion, which I cannot recall was used or not.

---

### DIFF
--- a/src/scripts/browserDeviceProfile.js
+++ b/src/scripts/browserDeviceProfile.js
@@ -253,14 +253,25 @@ function supportsHdr10(options) {
             || browser.safari && ((browser.iOS && browser.iOSVersion >= 11) || browser.osx)
             // Chrome mobile and Firefox have no client side tone-mapping
             // Edge Chromium 121+ fixed the tone-mapping color issue on Nvidia
-            || browser.edgeChromium && (browser.versionMajor >= 121)
-            || browser.chrome && !browser.mobile
+            // Desktop Chromium tone-mapping to SDR is unreliable (dark picture) with
+            // hardware decoding, so only report HDR support when the display is in
+            // HDR mode and no tone-mapping is needed. The video-dynamic-range query
+            // is still flagged as experimental in Chromium, so use dynamic-range.
+            || browser.edgeChromium && (browser.versionMajor >= 121) && window.matchMedia('(dynamic-range: high)').matches
+            || browser.chrome && !browser.mobile && window.matchMedia('(dynamic-range: high)').matches
             // Firefox 100+ has support for HDR on macOS/OS X. It requires OS support, which was
             // added in macOS 10.15 Catalina. If enabling HDR on other platforms, be careful about
             // allowing HDR VP9 in mp4 containers.
             //  * https://www.mozilla.org/en-US/firefox/100.0/releasenotes/
             //  * https://bugzilla.mozilla.org/show_bug.cgi?id=1915265
             || browser.firefox && browser.osx && (!browser.iphone && !browser.ipod && !browser.ipad) && (browser.versionMajor >= 100)
+            // Firefox reports HDR video support on other platforms through the
+            // video-dynamic-range media query (Windows support is experimental in
+            // Firefox 148). It evaluates to standard until HDR playback is available,
+            // so this cannot match versions before 148. The HDR VP9 color issue in
+            // mp4 containers noted above was fixed in Firefox 131.
+            //  * https://mozillagfx.wordpress.com/2026/01/16/experimental-high-dynamic-range-video-playback-on-windows-in-firefox-nightly-148/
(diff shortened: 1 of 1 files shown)


src/scripts/browserDeviceProfile.js
```diff
@@ -253,14 +253,25 @@ function supportsHdr10(options) {
...
             // Chrome mobile and Firefox have no client side tone-mapping
             // Edge Chromium 121+ fixed the tone-mapping color issue on Nvidia
-            || browser.edgeChromium && (browser.versionMajor >= 121)
-            || browser.chrome && !browser.mobile
+            // Desktop Chromium tone-mapping to SDR is unreliable (dark picture) with
+            // hardware decoding, so only report HDR support when the display is in
+            // HDR mode and no tone-mapping is needed. The video-dynamic-range query
+            // is still flagged as experimental in Chromium, so use dynamic-range.
+            || browser.edgeChromium && (browser.versionMajor >= 121) && window.matchMedia('(dynamic-range: high)').matches
+            || browser.chrome && !browser.mobile && window.matchMedia('(dynamic-range: high)').matches
```
nyanmisaka, line 261: You are breaking a feature that is working as intended. The `supportsHdr10` here indicates that the browser is at least capable of tone-mapping HDR video with correct colors, not that the screen's HDR mode is enabled.

Many users prefer to have the client handle tone-mapping rather than relying on server-side transcoding. For your specific use case, you should add a "Always transcode HDR video" checkbox to the playback settings, with the default state set to disabled.

nyanmisaka (changes requested):

src/scripts/browserDeviceProfile.js
KunJakob: I think I'm missing some understanding here that I hope you can help me with!

In testing, Chromium browsers are not capable of tone-mapping HDR video with correct colours. Playing an HEVC DV 8.1 video on a Chrome browser on a windows machine with an SDR display leads to a very dark image. I've tried Chrome and other chromium browsers, and none of them were able to handle tone mapping. Am I missing something, or what exactly constitutes being capable of tone-mapping HDR video with correct colors? It's capable of displaying HDR video on an HDR display, yes, but not tone-mapping it would seem.

Forcing all HDR video to be transcoded as a solution seems brutish in scenarios of the client having an HDR-capable display.

nyanmisaka: This is very common. To align with the dynamic metadata range of Dolby Vision, the brightness of the HDR10 compatibility layer is intentionally mastered significantly lower than that of a native HDR10 version. However, you cannot claim that Chrome’s tone mapping is incorrect, as static metadata is insufficient to compensate for the inherently lower brightness of the compatibility layer itself. Ultimately, Dolby simply does not prioritize the performance of the HDR10 compatibility layer, sacrificing it for the sake of Dolby Vision’s performance makes commercial sense.

KunJakob: I see! Would toggles (default off) for always transcoding dolby vision and always transcoding all HDR perhaps then be a good solution you would like? If always trancode HDR is enabled, the DV toggle is greyed out/disabled since it's redundant. I'd be happy to submit that PR instead!

KunJakob: Or would a solution still relying on dynamic detection still make sense if it only targets dolby vision media? I am a bit opposed to a client-settings playback toggle being required for this. I don't expect my family members to understand a word of this conversation to understand why they need to go deep in the settings to have things work as opposed to Jellyfin doing it's best effort by default. Surely it should be the powerusers that have to dive into settings to make things behave in particular fashion

nyanmisaka: Due to licensing issues, Chrome may never incorporate DV dynamic metadata parsing capabilities. There are also some poorly mastered HDR10 videos.

Add a dropdown menu containing three options: "DoVi Only", "All HDR Types", and a default "None" setting. This allows supportsHdr10 to be disabled for the selected HDR type whenever the client screen's HDR mode is inactive or DV profile is not supported.

If the client does not support the HDR mode detection API, this drop-down menu should always be hidden and disabled.

KunJakob: That proposed solution requires per-client intervention. It doesn't really help Jellyfin in having better usability. Individual clients should, in my eyes, not be required to have any understanding of media codecs etc.

I would love to submit a PR if you have an idea that can solve this without per-client intervention!

PR #8169, opened 2026-07-08
Add audiobook chapter controls to playback UI

quantumadmiral:
### Changes

Adds chapter-aware controls for audio playback when item chapters are provided by the server. This shows the current chapter beside the title and author text, adds chapter markers to the progress bar, adds chapters controls, and adds a chapter table-of-contents menu in the full Now Playing view. Playback item loading now requests Chapters so single-item and already-loaded audio playback paths can populate the controls. This relies on the server commit f5f75ed2e1b10dc1f4e55d5cdd9dd7fd69ea8f2b
feat/audiobook_chapters (#16518) which added the needed server side routes.

Before: 
image
image

Afterr:
image
image
image

### Issues

None.

### Code assistance

Code assistance was used for adding comments to the code, and adding buttons/editing CSS where needed.
---

* [x] I have read and followed the [contributing guidelines](https://github.com/jellyfin/jellyfin-web/blob/master/CONTRIBUTING.md).
* [x] I have tested these changes.
* [x] I have verified that this is not duplicating changes in an existing PR.

PR #8207, opened 2026-07-13
add chapters view

sevenrats:
### Changes

 - expose audiobook chapters using bespoke list-view
 - extend PlaybackManager.setCurrentPlaylistItem to support audiobook ux
 - extend PlaybackManager to keep audiobook pl item playback state live -- books are different
 - wire play buttons together in details - dont reset book state on play.
 - squash undefined id bug in playlist view (clicking title in playlist view leads nowhere because link is designed for albums)
 - the existing code has some shortcomings regarding tv-layout of multi-audio items where the listView is very long. this PR does not address them because they are preexisting, out of scope, and will be fixed when we move to tabview. the tabview will improve several aspects of multi-audio detail views on tv layouts.
 - port slider to react as jf-slider and use it
 - backport fixes from downstream react/osd
 
 ### Future PR's Related but Separate
  - Move sections to tabview in details view, improve tv focus management in details views, implement scrollbox in tabview (it wont break anymore because tabview pins it to the bottom)
 - Extend jf-slider with optional magnetic ticks (to be implemented in react/osd)
 - Implement magnetic ticks in audiobooks (react/osd)
 - Redesign chapter list items to be "better."
 
 ### Dependent PR's that will open immediately upon merge, already in progress:

- React Subtitle Settings: https://github.com/sevenrats/jellyfin-web/compare/feat/audiobook-chapter-list...sevenrats:jellyfin-web:react/subtitle-settings?expand=1
-  React Playback and OSD: https://github.com/sevenrats/jellyfin-web/compare/feat/audiobook-chapter-list...sevenrats:jellyfin-web:react/osd?expand=1

 

https://github.com/user-attachments/assets/8684de07-0b3d-4c27-875a-b35cea72c07b

### Code assistance

Assisted by Fable 5

---


jordanmichaelrushing:
@sevenrats do you have a video of it playing in non tv-mode? I'm juggling a few too many branches locally to pull this down

src/components/cardbuilder/audiobookChapterList.js
```diff
@@ -0,0 +1,530 @@
...
+
+    const chapterStart = chapter.StartPositionTicks;
+    const nextChapter = chapters[chapterIndex + 1];
+    const chapterEnd = nextChapter ? nextChapter.StartPositionTicks : null;
+
+    if (positionTicks < chapterStart) return null;
+    if (chapterEnd != null && positionTicks >= chapterEnd) return 1;
+
+    // Currently in this chapter
+    if (chapterEnd == null) return 0; // last chapter, no end boundary known
```
dfederm: The last chapter always reports progress 0 here because there is no next chapter, even though the item runtime is available for the duration calculation. When playback is inside the final chapter, the row is marked playing but the slider stays at the start. Pass the item runtime into the progress calculation, or use the same end boundary as `getChapterDurationTicks` for the last chapter.

src/components/cardbuilder/audiobookChapterList.js
```diff
@@ -0,0 +1,530 @@
...
+    if (player) {
+        _boundPlayer = player;
+        Events.on(player, 'timeupdate', _timeUpdateHandler);
+        Events.on(player, 'pause', _pauseHandler);
+        Events.on(player, 'unpause', _pauseHandler);
+        Events.on(player, 'playbackstop', _stopHandler);
+    }
+
+    // Also pick up if a player starts later
+    Events.on(playbackManager, 'playerchange', function playerChangeForChapters() {
```
dfederm: This `playerchange` listener is added every time the audiobook chapter list is built, but `unbindLiveUpdates()` only removes listeners from `_boundPlayer`. Leaving this attached means navigating away from an audiobook detail page still leaves a handler closed over the old container, and later player changes can rebind timeupdate handlers for stale chapter lists. Store the handler and remove it during cleanup.

src/components/cardbuilder/audiobookChapterList.js
```diff
@@ -0,0 +1,530 @@
...
+
+function bindLiveUpdates(container, item, chapters) {
+    unbindLiveUpdates();
+
+    _timeUpdateHandler = function () {
+        const player = playbackManager.getCurrentPlayer();
+        if (!player) return;
+        const currentItem = playbackManager.currentItem(player);
+        if (!currentItem || currentItem.Id !== item.Id) return;
+        const currentTicks = playbackManager.currentTime(player) * 10000;
```
jordanmichaelrushing: Because of transcoding audio, you'll want to use `playbackManager.getCurrentTicks(player)`.

`getCurrentTicks` will take care of the 10000 and will add any `transcodingOffsetTicks`.

jordanmichaelrushing:
I'm curious your thoughts on using card with images. I ask because while a singular audiobook will only have a singular image, multi-file audiobooks can each have their own album artwork that could be nice to show off. Example: (links only so I don't blow up the vertical scroll of your PR):
- https://github.com/user-attachments/assets/f4781b46-bd50-4c11-a165-ee892a6b7179
- https://github.com/user-attachments/assets/bf5c456a-2715-4ec6-a139-d775b3e661ea

I didn't made them square, so it's not the prettiest, just an example to think about

sevenrats:
the team is not yet in alignment on whether or not multifile audiobook support will even be accepted. its a good idea to keep this consideration in mind, but, frankly there are a lot of corners that get exposed by multifile and in many cases the solutions are not designed. i think it will be more natural to flesh out single-file support, and then make our multifile design decisions from that new reality once we get there

src/components/cardbuilder/audiobookChapterList.js
sevenrats: oh yeah this was just shitty vibes on my part. good catch.

jordanmichaelrushing:
Bummer, but understood. I'll still put up my multi-file audiobook PR when these initial chapter PRs get merged in. And if it gets merged in, cool. If not, my server is where I want it :P

sevenrats:
- ported to react (per: no new imperative js)
- addressed code feedback
- implemented emby-slider in react as jf-slider with full contract tests (example cutover and prove-out here: https://github.com/sevenrats/jellyfin-web/compare/feat/audiobook-chapter-list...sevenrats:jellyfin-web:react/subtitle-settings?expand=1)
- add test coverage for new components

TODO:

- [x] fix scroll-into-view to center instead of scrolling to edge
- [x] backport Slider name extensions and tests from react OSD branch

src/components/cardbuilder/audiobookChapterList/AudiobookChapterRow.tsx
```diff
@@ -0,0 +1,193 @@
...
+}) => {
+    const sliderRef = useRef<JfSliderHandle>(null);
+
+    const runTimeTicks = item.RunTimeTicks || 0;
+    const state = getChapterState(chapter, chapterIndex, chapters, positionTicks, runTimeTicks);
+    const progress = getChapterProgress(chapter, chapterIndex, chapters, positionTicks, runTimeTicks);
+    const chapterStart = chapter.StartPositionTicks ?? 0;
+    const chapterDurationTicks = getChapterDurationTicks(chapter, chapterIndex, chapters, runTimeTicks);
+    const timeDisplay = getChapterTimeDisplay(chapter, chapterIndex, chapters, state, positionTicks, runTimeTicks);
+    const chapterName = chapter.Name || `Chapter ${chapterIndex + 1}`;
```
jordanmichaelrushing: Does the or path need i18n?

src/components/cardbuilder/audiobookChapterList/AudiobookChapterRow.tsx
```diff
@@ -0,0 +1,193 @@
...
+            </div>
+            <div className='audiobookChapterItem-index'>{chapterIndex + 1}</div>
+            <div className='listItemBody audiobookChapterItem-body'>
+                <div className='listItemBodyText audiobookChapterItem-name'>{chapterName}</div>
+            </div>
+            {isPlaying && (
+                <span
+                    className='audiobookChapterItem-restart material-icons replay'
+                    role='button'
+                    tabIndex={-1}
```
jordanmichaelrushing, line 172: Does this being -1 mean it won't be able to be reached in tv mode?

src/components/cardbuilder/audiobookChapterList/AudiobookChapterList.tsx
```diff
@@ -0,0 +1,58 @@
...
+    const containerRef = useRef<HTMLDivElement>(null);
+
+    const playingIndex = useMemo(() => (
+        chapters.findIndex((chapter, i) => getChapterState(chapter, i, chapters, positionTicks, item.RunTimeTicks || 0) === 'playing')
+    ), [chapters, positionTicks, item.RunTimeTicks]);
+
+    // Scroll the playing chapter into view and focus it, stealing focus back
+    // from the detail page's banner play button (that auto-focus runs
+    // synchronously, before this effect).
+    useEffect(() => {
```
jordanmichaelrushing, line 30: More of a note, but it seems like this will potentially pulling focus away from whatever the user's doing elsewhere on the page. Possibly wants to run once on mount only, unless this is on purpose. I haven't tested it but I wonder if for example it would pull attention away when typing in metadata editor on the same page.

jordanmichaelrushing (commented):
I'm still looking but need to step away for a few hours. I'll come back to it soon

jordanmichaelrushing:
`getChapterProgress` and `getChapterDurationTicks` both independently derive the chapter's end boundary with slightly different null/0 fallbacks. I suggest extracting a shared `getChapterBounds(chapter, chapterIndex, chapters, itemRunTimeTicks)` that both call, rather than each re-deriving chapterEnd, limiting potential for eventual logic drift over time.

src/components/cardbuilder/audiobookChapterList/AudiobookChapterList.tsx
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// Chapter list for audiobooks, styled after the album track list, with a
+// seek slider on the playing chapter. Rendered into the detail page's
+// children container by the legacy itemDetails controller.
+const AudiobookChapterList: FC<AudiobookChapterListProps> = ({ item, chapters }) => {
+    const { positionTicks, isActiveForItem, isPaused } = usePlaybackProgress(item);
+    const containerRef = useRef<HTMLDivElement>(null);
+
+    const playingIndex = useMemo(() => (
```
jordanmichaelrushing, line 25: In a world where multi-file audiobooks (one synthesized chapter per file) get accepted, chapter counts could get into the hundreds for a single book. `playingIndex` currently does a linear scan (findIndex) over every chapter on each timeupdate. Since chapters are already sorted by `StartPositionTicks`, could this become a binary search over the start-tick boundaries instead? Would keep the per-tick cost flat (O(log n)) regardless of how many chapters a book ends up with, rather than scaling with chapter count.

jordanmichaelrushing:
Ok, that's all I've got for code review. One thing to call out, since yours is item based and [the other](https://github.com/jellyfin/jellyfin-web/pull/8169) is player based, let's make sure they play well together

Edit: That being said the person on the other Chapter PR hasn't responded at all. If they don't respond soon, I'll spin up my chapter control work into a supplementary PR that works with yours.

src/components/cardbuilder/audiobookChapterList/AudiobookChapterRow.tsx
sevenrats: yes its invisible in tv mode because the focus paradigm doesnt allow access to it.

sevenrats:
- extract getChapterBounds as the single source of truth for a chapter's
  tick range; progress, duration and time display now all derive start/end
  from it instead of re-deriving the end boundary with divergent null/0
  fallbacks
- split the chapter scroll and focus effects: scrolling still follows
  playback across chapter boundaries, but focus is claimed only once and
  skipped when an interaction is already in progress, so a boundary
  crossing no longer yanks focus away from the user
- center the playing chapter when scrolling it into view
- translate the chapter name fallback, restart title and seek slider
  aria-label; adds ChapterNumber, RestartChapter and SeekWithinChapter

src/components/cardbuilder/audiobookChapterList/AudiobookChapterList.tsx
sevenrats: I am willing to look at this later but its a bit of an explosion to handle it here, plus my initial attempt resulted in some failed fuzz tests so, we can keep an eye on this but this PR is already massive.

sevenrats:
TODO:

- [x] clicking into now-playing and pressing browser back causes the chapter list to not appear
 - [ ] ~~seek-preview bug when clicking in tv mode (do we even support clicking in tv mode?)~~ we dont support clicking in tv mode

jordanmichaelrushing (approved):
lgtm

PR #8209, opened 2026-07-13
Merge cards for crew with multiple roles

theguymadmax:
### Changes
Merge cards for people with multiple crew roles (Writer, Director, Producer).

Fixed a "Writer" display bug. The writing department has three jobs: Writer, Screenplay, Novel. We were incorrectly showing the type alongside the job, so if a person's role was listed as Screenplay, the "Writer" type label would also display, even though that's a separate credit and shouldn't show unless they're specifically credited as a writer.

Added translations for "Novel" and "Screenplay" since we narrowed the PersonType assignment in [#15872](https://github.com/jellyfin/jellyfin/pull/15872), those were the only labels not being translated. 

Before:
before

After:
after

### Issues
N/A

### Code assistance
N/A

---


src/apps/legacy/controllers/itemDetails/index.js
```diff
@@ -856,16 +856,33 @@ function setInitialCollapsibleState(page, item, apiClient, context, user) {
         page.querySelector('#specialsCollapsible').classList.add('hide');
     }
 
+    const mergeablePersonTypes = [
```
Finomosec: What sonar said makes sense: `mergeablePersonTypes` should be a `Set`, and use `mergeablePersonTypes.has()` to check existence or non-existence.

Finomosec (approved):
I reviewed it and apart from the "Set" hint, which sonar pointed out, i see no problems.

src/components/cardbuilder/cardBuilder.js
```diff
@@ -555,25 +573,10 @@ function getCardFooterText(item, apiClient, options, footerClass, progressHtml,
...
-                if ([ PersonKind.Actor, PersonKind.GuestStar ].includes(item.Type)) {
-                    // List actor roles formatted like "as Character Name"
-                    const roleText = globalize.translate('PersonRole', escapeHtml(item.Role));
-                    lines.push(`<span title="${roleText}">${roleText}</span>`);
-                } else if (item.Role.toLowerCase() === item.Type.toLowerCase()) {
-                    // Role and Type are the same so use the localized Type
-                    lines.push(escapeHtml(globalize.translate(item.Type)));
-                } else if (item.Role.toLowerCase().includes(item.Type.toLowerCase())) {
-                    // Avoid duplication if the Role includes the Type (i.e. Executive Producer)
-                    lines.push(escapeHtml(item.Role));
```
thornbill, line 568: It seems this case was lost?

src/components/cardbuilder/cardBuilder.js
```diff
@@ -555,25 +573,10 @@ function getCardFooterText(item, apiClient, options, footerClass, progressHtml,
...
-                } else {
-                    // Type and Role are unique so list both (i.e. Writer | Novel)
-                    lines.push(escapeHtml(globalize.translate(item.Type)));
-                    lines.push(escapeHtml(item.Role));
-                }
+            if (item.RoleList ?.length > 1) {
```
thornbill: Odd that lint didn't throw on this one 🙃 
```suggestion
            if (item.RoleList?.length > 1) {
```

src/components/cardbuilder/cardBuilder.js
```diff
@@ -555,25 +573,10 @@ function getCardFooterText(item, apiClient, options, footerClass, progressHtml,
...
+                lines.push(item.RoleList.map(getPeopleRoleOrTypeLabel).join(' / '));
```
thornbill, line 577: Why use `/` to join here? No strong opinion on it, just curious about the reasoning. 🙂

theguymadmax: TMDB uses `/` when an actor has two roles: https://www.themoviedb.org/movie/126889-alien-covenant.

So the separator is already used in Jellyfin via TMDB


1a

src/components/cardbuilder/cardBuilder.js
theguymadmax: No, it gets picked up by the fall through: `return escapeHtml(globalize.translate(Role));`

src/components/cardbuilder/cardBuilder.js
theguymadmax: Fixed, feel free to squash.

src/components/cardbuilder/cardBuilder.js
thornbill: Ok so this is the case we lost then...
```
// Type and Role are unique so list both (i.e. Writer | Novel)
```
With this change we would only ever list a single line per person so in this case just "Novel".

theguymadmax: Looking at the server [code](https://github.com/jellyfin/jellyfin/blob/b2ae39e0d911dd8065c4504a82b00965a18c6bfd/MediaBrowser.Controller/Entities/PeopleHelper.cs#L73), we are currently deduping by type not role. So that means a person that has multiple roles for the same type (e.g., PersonType:Writer with Roles:Novel, Writer) will only ever only ever have one Role (e.g., "Novel") in the DB.

If the role is different than the type (e.g., PersonType:Writer Role:Novel) we are automatically adding the type alongside the role. That's a bug, as it incorrectly represents any crew members who only has Screenplay/Novel role as also having a role of a Writer (there are only three roles we return: Writer, Screenplay, Novel for PersonType:Writer). So the current logic will indirectly be correct for a few people who actually have multiple roles, but incorrect for the majority.

For server changes, if we concatenate the roles on the server, the current code would be fine. If we add additional rows for each role, then the code will need adjusting. Since it's not possible for the server to return multiple roles for the same type, let's keep it as is and revisit when/if the server allows it.

theguymadmax: With a server fix pending https://github.com/jellyfin/jellyfin/pull/17466/changes/cbc2c7c32345f670d928f7840e5fd7422241845d, I was able to verify that this PR works as-is and better demonstrates the bug (blindly appending type when the role is different) 

With server fix, current web UI:
1a

With server fix, this PR:
2a

thornbill (approved):

PR #8211, opened 2026-07-14
Enable iOS background play when screen is turned off

jordanmichaelrushing:
### Changes
Set background play on iOS when Normalization is turned off

### Issues
https://github.com/jellyfin/jellyfin-web/issues/6113
https://github.com/jellyfin/jellyfin-ios/issues/799
https://github.com/jellyfin/jellyfin-ios/issues/19
https://github.com/jellyfin/jellyfin-ios/issues/487

### Code assistance
None

---

### DIFF
--- a/src/plugins/htmlAudioPlayer/plugin.js
+++ b/src/plugins/htmlAudioPlayer/plugin.js
@@ -164,6 +164,11 @@ class HtmlAudioPlayer {
                 elem.crossOrigin = crossOrigin;
             }
 
+            // This avoids the AudioContext being suspended when Safari is put into background
+            if ('audioSession' in navigator) {
+               navigator.audioSession.type = 'playback';
+            }
+
             return enableHlsPlayer(val, options.item, options.mediaSource, 'Audio').then(function () {
                 return new Promise(function (resolve, reject) {
                     requireHlsPlayer(async () => {
--- a/src/plugins/htmlAudioPlayer/plugin.js
+++ b/src/plugins/htmlAudioPlayer/plugin.js
@@ -166,7 +166,7 @@ class HtmlAudioPlayer {
 
             // This avoids the AudioContext being suspended when Safari is put into background
             if ('audioSession' in navigator) {
-               navigator.audioSession.type = 'playback';
+                navigator.audioSession.type = 'playback';
             }
 
             return enableHlsPlayer(val, options.item, options.mediaSource, 'Audio').then(function () {


jordanmichaelrushing:
Context from the other PR: https://github.com/jellyfin/jellyfin-web/pull/7681#issuecomment-4964477125

dkanada:
I agree with you and think we have been conflating two issues for iOS audio. One is that background playback doesn't work without this change. The other is that audio normalization doesn't work at all and will cause pitch shifting.

To that end I think you need to move this block of code to line 118 above all the audio normalization paths, and add a return *inside* the audioSession check. You should also keep the comment from the other pull request with another line explaining that audio normalization is broken and disabled.

jordanmichaelrushing:
> with another line explaining that audio normalization is broken and disabled.

I thought the pitch shifting was because of this change on the other PR being placed in the normalization path? But normalization on its own was fine? I don't have an iPhone, so I'm no expert here. My wife does however and up until I made this change on my own server last week she had normalization turned on and it was fine.

> To that end I think you need to move this block of code to line 118 above all the audio normalization paths and return directly inside

I want to push back here until someone explicitly confirms that normalization on its own is broken on iOS. Because if it is, then it sounds like normalization should be disabled in iOS. Even then, if it is I think that expands the scope of this change from just background audio to background audio is removing normalization from iOS as a whole. Edit: Check my next comment

jordanmichaelrushing:
Wait wait wait. This is somewhat related, but really I need to get a thought out. My completely separate audiobook playback rate PR needs to use server side atempo instead of using html playbackRate because iOS would skip over words when playing faster than 1.0x and would duplicate words when played slower than 1.0x. Is it because of this damn normalization? If so, I am FULLY in favor of disabling normalization on iOS, making my audiobook playback speed MUCH more simple and this path in this PR much more simple.

I'm going to run a quick test on my end, if html playback rate doesn't break with normalization turned off, I'm going to officially pitch disabling that setting.
Edit: IT WORKS WITH AUDIO NORMALIZATION OFF. Weeks I spent on atempo because of iOS... @dkanada I fully agree with you, rip it out!

dmitrylyzo:
> > To that end I think you need to move this block of code to line 118 above all the audio normalization paths and return directly inside
> 
> I want to push back here until someone explicitly confirms that normalization on its own is broken on iOS. Because if it is, then it sounds like normalization should be disabled in iOS. Even then, if it is I think that expands the scope of this change from just background audio to background audio is removing normalization from iOS as a whole. Edit: Check my next comment

If this block tells the OS that the current session is a playback and should run in the background (and this fixes the issue), it should be applied to any playback (normalized or not, and without 'return').

Does normalization make the playback to run in the background without `navigator.audioSession`?

Shouldn't we reset `navigator.audioSession` when the playback ended?

jordanmichaelrushing:
> If this block tells the OS that the current session is a playback and should run in the background (and this fixes the issue), it should be applied to any playback (normalized or not, and without 'return').

@dmitrylyzo I would point you to the original [PR](https://github.com/jellyfin/jellyfin-web/pull/7681) in which people were having issues with audio normalization and pitch changes. I agree that feature parody is important and called out in my comment over there that this might cause confusion for users on why background play isn't working for them because some random setting that is on by default prevents this: https://github.com/jellyfin/jellyfin-web/pull/7681#issuecomment-4876981651 

However, it seems like based upon what @dkanada said above, and my test for unreleased audiobook speed playback, that audio normalization on iOS might be broken in general for audio seeking (it flat out is on my unreleased code, see my previous comment).

I haven't tested seeking with audio normalization just yet (waiting for my wife to test). If anyone (using a real release not this code change) can test on iOS with "Track Gain" Audio Normalization turned on and seek 25% into an audiobook, does the pitch change? If it does, it's unrelated to this change and I would push hard to turn off / hide / ignore / disable Audio Normalization on iOS full stop.

> Does normalization make the playback to run in the background without navigator.audioSession?

It does not, it's just for normalizing audio, not the background playing. 

> Shouldn't we reset navigator.audioSession when the playback ended?

I'm not an iOS engineer, so I'm really unsure. I just saw what seemed like a backwards pathway in the other PR and wanted to fix it.

dmitrylyzo:
> @dmitrylyzo I would point you to the original [PR](https://github.com/jellyfin/jellyfin-web/pull/7681) in which people were having issues with audio normalization and pitch changes. I agree that feature parody is important and called out in my comment over there that this might cause confusion for users on why background play isn't working for them because some random setting that is on by default prevents this: [#7681 (comment)](https://github.com/jellyfin/jellyfin-web/pull/7681#issuecomment-4876981651)

Looked the original PR. It looks like it fixed background playback when normalization is enabled, your PR does the opposite when normalization is disabled.

Forget about buggy normalization for now - you are trying to fix the background playback. So, I suggest moving this block so that it works for both branches (_who knows, maybe Apple will fix normalization in the future_). For example, it can be placed [after this line](https://github.com/jellyfin/jellyfin-web/blob/601bd462458efe523ae4962b2988a1452166655c/src/plugins/htmlAudioPlayer/plugin.js#L113) or (_better?_) [before this one](https://github.com/jellyfin/jellyfin-web/blob/601bd462458efe523ae4962b2988a1452166655c/src/plugins/htmlAudioPlayer/plugin.js#L167).

jordanmichaelrushing:
> it fixed background playback when normalization is enabled

Ehh. It fixed it yes, until pitch issues came up for people, which is why that PR was never merged in and I thought going to denormalized only would be safer

> Your PR does the opposite when normalization is disabled

Correct because the assumption seemed to be that adding this background play exposed an underlying pitch issue (in either normalization path). I'm happy to move it before the HLS call. I'll push a commit, I'll just need someone to test whether having this in the normalization path does in fact expose an issue.

*Available on Docker Hub in case it helps: tatertechstudios/jellyfin:ios-background-play*

Edit: I tested on my wife's iPhone with the speaker and bluetooth headphones and was not able to repro locally with background in the normalization pathway. @dkanada you reprod it in the other PR, can you help? Also worth knowing: Do you remember if you reproduced with the iphone speaker or bluetooth? If it's bluetooth only, I think I _may_ know what the pitch issue is.

jordanmichaelrushing:
> or (better?) [before this one](https://github.com/jellyfin/jellyfin-web/blob/601bd462458efe523ae4962b2988a1452166655c/src/plugins/htmlAudioPlayer/plugin.js#L167).

~Wait no, it needs to be earlier otherwise the early return (what I'm calling denormalized) path won't get this background play. Commit a new change after line 113.~

dmitrylyzo:
> > or (better?) [before this one](https://github.com/jellyfin/jellyfin-web/blob/601bd462458efe523ae4962b2988a1452166655c/src/plugins/htmlAudioPlayer/plugin.js#L167).
> 
> Wait no, it needs to be earlier otherwise the early return (what I'm calling denormalized) path won't get this background play. Commit a new change after line 113.

That `return` is inside `import`. "Denormalized" path still uses Hls or direct HTML audio element.

jordanmichaelrushing:
Yup you're right. I misread it as the full function

dkanada:
@jordanmichaelrushing the audioSession check looks fine now. Normally I wouldn't suggest combining changes in a single pull request, but in this case I think we also need to add a case to line 118 that disables audio normalization on iOS devices to prevent the pitch shifting.

```
if ('audioSession' in navigator) {
  console.log('audio normalization is unsupported on iOS');
  return;
} else if (userSettings.selectAudioNormalization() == 'TrackGain') {
[...]
```

dmitrylyzo:
> Normally I wouldn't suggest combining changes in a single pull request, but in this case I think we also need to add a case to line 118 that disables audio normalization on iOS devices to prevent the pitch shifting.

Let it go to another PR. 🙃
Ideally, we should also hide it in the settings.

We also need to consider the theme song. I suppose it shouldn't play in the background. Iirc, theme player sets [`fullscreen: false` in `options`](https://github.com/jellyfin/jellyfin-web/blob/a60d9a298ad70ae97b16e738f3bb8f1fc76c93e2/src/components/themeMediaPlayer.js#L44).

We should test audio normalization on Mac. Is it  
buggy on Safari in general or mobile only?

jordanmichaelrushing:
> Ideally, we should also hide it in the settings.

I did on my local NAS and returning early on iOS and hiding the setting solves all of my problems. But yes I agree, different PR. I'm happy to spin that one up, I already have the code ready, I'm just at my 3 PR limit in this repo unfortunately. 

Edit: The commit - https://github.com/TaterTechStudios/jellyfin-web/commit/a6995227258cbb18de0eb0b03c4b9994a7d3ed90

jordanmichaelrushing:
I tested on my Mac in Safari. Still not able to reproduce it. I tried internal speakers and BT headphones. Which reminds me...

> If it's bluetooth only, I think I may know what the pitch issue is

My thought here was that AudioContext probably locks the sample rate when created and if someone starts with the internal speaker (probably 48kHz) and then switched to headphones (most likely 44kHz) while the screen is off (honestly maybe even turned on) then they might be a pitch shift until the buffer is exhausted and then is able to switch to 44kHz. That is my best guess at this point. That being said, I can't reproduce so we'll need exact steps from @dkanada 

================================

I did retest on my wife's iPhone however and this may have introduced a bug. I was listening via the phone speaker, then switched over to bluetooth headphones. Seeked and turned off the screen, everything normal listening for around 30 seconds. Then I turned the phone back on, unlocked the phone and then disconnected bluetooth. Jellyfin pauses but when I press play, the playtime continues but no audio comes out. Verified it twice and then switched over to my main personal branch (not clean master FWIW, has a lot of audiobook goodies) and wasn't able to reproduce it there. So either my personal master branch changes has fixed this OR this is causing a new issue.

jordanmichaelrushing:
These connection issues aren't my strength so I threw it at Fable and based on my local master I think my changes there may have actually inadvertently fixed this. It convinced me to test this "not playing" bug with a plain upstream master. It's a busy day for me so I'll try and repro clean master at some point today.

jordanmichaelrushing:
Phew! Good news, this bug exists on master. Not due to this change. I think we can move forward on this :)

dkanada (approved):

MBR-0001 (approved):

dkanada:
@jordanmichaelrushing please fix the lint issues

jordanmichaelrushing:
Good to go

jordanmichaelrushing:
@joshuaboniface fyi looks like auto-merge was turned off because I had to push a lint error fix

PR #8226, opened 2026-07-16
Add zoom control to slideshow

dkacperski97:
### Changes
- Added zoom control to slideshow view for precise zoom level adjustment using the mouse.
- Updated swiper-zoom-fakeimg behaviour to hide during every zoom animation instead of only during the first one.
- Moved getIcon to slideshowHelper so it can be shared between the slideshow and zoom control.
- Split the bottom bar because, in my opinion, having the zoom bar in the center looked a bit weird.

Before:
obraz

After:
obraz

### Code assistance
Copilot autocomplete 

---

   * https://github.com/jellyfin/jellyfin-web/pull/8228

### DIFF
--- a/src/components/slideshow/slideshow.js
+++ b/src/components/slideshow/slideshow.js
@@ -132,7 +132,7 @@ export default function (options) {
     let hideTimeout;
     /** Last coordinates of the mouse pointer. */
     let lastMouseMoveData;
-    /** Instance of zoom control **/
+    /** @type {ZoomControl|undefined} Instance of zoom control */
     let zoomControl;
 
     /**
--- a/src/components/slideshow/slideshow.js
+++ b/src/components/slideshow/slideshow.js
@@ -20,6 +20,8 @@ import dom from '../../utils/dom';
 import './style.scss';
 import 'material-design-icons-iconfont';
 import '../../elements/emby-button/paper-icon-button-light';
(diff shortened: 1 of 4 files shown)


src/components/slideshow/slideshow.js
```diff
@@ -144,6 +132,8 @@ export default function (options) {
     let hideTimeout;
     /** Last coordinates of the mouse pointer. */
     let lastMouseMoveData;
+    /** Instance of zoom control **/
+    let zoomControl;
```
Chris4942: Please include a type annotation here

Chris4942 (commented):
With substantive redesigns like this, they often want the changes to go through the design folks in `#ui-ux` on discord to make sure this vibes with the rest of the app.

I'm not a maintainer, so maybe this is fine, but it would probably better to run this by them.

dmitrylyzo:
The slider may cause problems with navigation on TV (_try using "TV" display mode on PC_).

dkacperski97:
> The slider may cause problems with navigation on TV (_try using "TV" display mode on PC_).

The top and bottom bars are not displayed on the TV:
https://github.com/dkacperski97/jellyfin-web/blob/4daa8aef5006bbc0838e91503342d4f8a7f44dbd/src/components/slideshow/slideshow.js#L160

PR #8227, opened 2026-07-16
add new starfield screensaver

Tone866:
### Changes

Adds a new screensaver with OLEDs in mind.
Looks like a simple starfield.
Pure black background and stars with random locations.
### Issues

/
### Code assistance

Claude helped to create the star logic.

---

* [x] I have read and followed the [contributing guidelines](https://github.com/jellyfin/jellyfin-web/blob/master/CONTRIBUTING.md).
* [x] I have tested these changes.
* [x] I have verified that this is not duplicating changes in an existing PR.
* [ ] I have provided a *substantive* review of [another web PR](https://github.com/jellyfin/jellyfin-web/pulls?q=is%3Apr+is%3Aopen+-label%3Adependencies+-label%3Ablocked+-label%3Abackend+-label%3Ainvalid+-label%3A"merge+conflict"+-label%3Astale+-author%3A%40me).

PR #8228, opened 2026-07-16
migrate login code to typescript

Chris4942:
### Changes
- Migrate `src/apps/legacy/controllers/session/login/index.ts` from js to ts 

### Code assistance

Asked chatgpt how to deal with the JSDoc `Object.<string, string>` type and it said to use `Record<string, string>` instead.

---

    - Manual login
    - Login through card
    - Login through quick connect
    - https://github.com/jellyfin/jellyfin-web/pull/8227
    - https://github.com/jellyfin/jellyfin-web/pull/8226
    - https://github.com/jellyfin/jellyfin-web/pull/8345


src/apps/legacy/controllers/session/login/index.ts
```diff
@@ -267,34 +275,36 @@ export default function (view, params) {
...
+                (view.querySelector('#txtManualName') as HTMLInputElement).value = '';
                 showManualForm(view, false, false);
             }
         }).catch().then(function () {
             loading.hide();
         });
-        apiClient.getJSON(apiClient.getUrl('Branding/Configuration')).then(function (options) {
-            const loginDisclaimer = view.querySelector('.loginDisclaimer');
+        // Should this promise be awaited?
+        void apiClient.getJSON(apiClient.getUrl('Branding/Configuration')).then(function (options: BrandingOptionsDto) {
```
dkacperski97, line 305: API calls should probably be replaced with functions from the Jellyfin TypeScript SDK, e.g. `getBrandingApi(api).getBrandingOptions()`.

dkacperski97 (commented):
I think it would be better to use the non-null assertion operator `!` instead of the `as` keyword where applicable. Especially for the user object, this would avoid duplicating type definitions.
The same applies to `as Element` after `querySelector`, the function already returns an Element.

src/apps/legacy/controllers/session/login/index.ts
Chris4942: Yeah, that would be better for sure. I’d rather not do that in this PR though because I don’t want to make any functional changes here however small they might be.

So, yes, but that’s outside the scope of migrating to typescript

Chris4942:
Good catch on using `!` instead of `as`. Will update accordingly

src/apps/legacy/controllers/session/login/index.ts
```diff
@@ -182,17 +187,18 @@ function loadUserList(context, apiClient, users) {
...
+export default function (view: HTMLElement, params: Record<string, string>) {
+    function getApiClient(): ApiClient {
         const serverId = params.serverid;
 
         if (serverId) {
             return ServerConnections.getOrCreateApiClient(serverId);
         }
 
+        // @ts-expect-error: We're returning a type definition instead of an instance. This path is probably never hit since it would break if it were.
         return ApiClient;
```
dmitrylyzo: It works because it returns `window.ApiClient`, but we shouldn't use it in the new code.

src/apps/legacy/controllers/session/login/index.ts
```diff
@@ -209,56 +215,57 @@ export default function (view, params) {
...
                 showManualForm(context, true, true);
             }
         }
     });
-    view.querySelector('.manualLoginForm').addEventListener('submit', function (e) {
-        appSettings.enableAutoLogin(view.querySelector('.chkRememberLogin').checked);
-        authenticateUserByName(view, getApiClient(), getTargetUrl(), view.querySelector('#txtManualName').value, view.querySelector('#txtManualPassword').value);
+    view.querySelector('.manualLoginForm')!.addEventListener('submit', function (e) {
+        appSettings.enableAutoLogin((view.querySelector('.chkRememberLogin') as HTMLInputElement).checked);
+        authenticateUserByName(view, getApiClient(), getTargetUrl(), (view.querySelector('#txtManualName') as HTMLInputElement).value, (view.querySelector('#txtManualPassword') as HTMLInputElement).value);
```
dmitrylyzo: AFAIK, you can specify the type:
```suggestion
        authenticateUserByName(view, getApiClient(), getTargetUrl(), view.querySelector<HTMLInputElement>('#txtManualName')!.value, view.querySelector<HTMLInputElement>('#txtManualPassword')!.value);
```

src/apps/legacy/controllers/session/login/index.ts
```diff
@@ -48,15 +50,17 @@ function authenticateUserByName(page, apiClient, url, username, password) {
     });
 }
 
-function authenticateQuickConnect(apiClient, targetUrl) {
+function authenticateQuickConnect(apiClient: ApiClient, targetUrl: string) {
     const url = apiClient.getUrl('/QuickConnect/Initiate');
+    // @ts-expect-error: The signature of ajax has just one argument. The `true` can probably be removed with no consequences
     apiClient.ajax({ type: 'POST', url }, true).then(res => res.json()).then(function (json) {
```
dmitrylyzo: It has a second argument: https://github.com/jellyfin-archive/jellyfin-apiclient-javascript/blob/0595869949a6da0b9e62f6a2c4a4e349f5282109/src/apiClient.js#L422

I think we should update the declaration:
https://github.com/jellyfin/jellyfin-web/blob/b57c74f17ab0d1cf8b81e17427b1b38f57cbbc35/src/apiclient.d.ts#L78

src/apps/legacy/controllers/session/login/index.ts
```diff
@@ -267,34 +274,36 @@ export default function (view, params) {
...
+                    view.querySelector('.btnQuick')!.classList.remove('hide');
                 }
             })
             .catch(() => {
                 console.debug('Failed to get QuickConnect status');
             });
 
-        apiClient.getPublicUsers().then(function (users) {
+        // Should we be awaiting this promise?
+        void apiClient.getPublicUsers().then(function (users) {
```
dmitrylyzo: No, it fetched data from the server "in parallel".

src/apps/legacy/controllers/session/login/index.ts
```diff
@@ -267,34 +274,36 @@ export default function (view, params) {
...
+                (view.querySelector('#txtManualName') as HTMLInputElement).value = '';
                 showManualForm(view, false, false);
             }
         }).catch().then(function () {
             loading.hide();
         });
-        apiClient.getJSON(apiClient.getUrl('Branding/Configuration')).then(function (options) {
-            const loginDisclaimer = view.querySelector('.loginDisclaimer');
+        // Should this promise be awaited?
+        void apiClient.getJSON(apiClient.getUrl('Branding/Configuration')).then(function (options: BrandingOptionsDto) {
```
dmitrylyzo: No, it fetched data from the server "in parallel".

src/apps/legacy/controllers/session/login/index.ts
```diff
@@ -209,56 +215,57 @@ export default function (view, params) {
...
-        view.querySelector('.visualLoginForm').classList.remove('hide');
-        view.querySelector('.manualLoginForm').classList.add('hide');
-        view.querySelector('.btnManual').classList.remove('hide');
+        view.querySelector('.visualLoginForm')!.classList.remove('hide');
+        view.querySelector('.manualLoginForm')!.classList.add('hide');
+        view.querySelector('.btnManual')!.classList.remove('hide');
 
-        import('components/autoFocuser').then(({ default: autoFocuser }) => {
+        // Should we be handling this promise?
+        void import('components/autoFocuser').then(({ default: autoFocuser }) => {
```
dmitrylyzo: Do you mean "await"? I think no, because it wasn't awaited before. We just need to load autoFocuser and run it.

src/apps/legacy/controllers/session/login/index.ts
```diff
@@ -48,15 +50,17 @@ function authenticateUserByName(page, apiClient, url, username, password) {
...
+    // @ts-expect-error: The signature of ajax has just one argument. The `true` can probably be removed with no consequences
     apiClient.ajax({ type: 'POST', url }, true).then(res => res.json()).then(function (json) {
         if (!json.Secret || !json.Code) {
             console.error('Malformed quick connect response', json);
             return false;
         }
 
-        baseAlert({
+        // Should we be awaiting this promise?
+        void baseAlert({
```
dmitrylyzo: Probably not, because it polls the quick connection below (interval) and closes the dialog after authorization.

src/apps/legacy/controllers/session/login/index.ts
```diff
@@ -112,31 +116,32 @@ function authenticateQuickConnect(apiClient, targetUrl) {
     });
 }
 
-function onLoginSuccessful(id, accessToken, apiClient, url) {
+function onLoginSuccessful(id: string | null | undefined, accessToken: string | null | undefined, apiClient: ApiClient, url: string) {
     Dashboard.onServerChanged(id, accessToken, apiClient);
-    Dashboard.navigate(url || 'home');
+    // Should we be handling the promise from navigate?
+    void Dashboard.navigate(url || 'home');
```
dmitrylyzo: We should just navigate and forget.

src/apps/legacy/controllers/session/login/index.ts
Chris4942: Updated the comment accordingly

src/apps/legacy/controllers/session/login/index.ts
Chris4942: That's very slick. I like it. Done

src/apps/legacy/controllers/session/login/index.ts
Chris4942: Updated!

PR #8234, opened 2026-07-17
Fix remote control queue not refreshing

### LINKED ISSUE #8182: [12.0-rc2] Remote queue is missing when connected to remote client

### Describe the bug

When starting playback with multiple tracks while connected to a remote client, the queue doesn't show up in the controlling Jellyfin Web client. It is sent to the controlled client and shown there, just not on the controlling client.  

When starting the a queue on the controlling client *before* connecting to the remote client, then connecting to the other client, and *then* starting the _same_ queue, it does show up on both clients. See the attached video for details.

### Reproduction Steps

1. Connect to other Jellyfin Web client (version doesn't seem to matter)
2. Start playing a queue with multiple tracks (e.g. an album)
3. Compare the queues between the two clients. Controlling client has no queue, remote client has the correct queue
4. Now reload both clients to stop playback
5. Start playing the album on the controlling client, but *without* connecting to the remote client. The queue will be shown
6. Now connect to the remote client. It starts playing the same queue, but the controlling client is missing its queue again
7. Now start playback of that album again, while still being connected to the remote client. The remote client reloads the queue, and now both clients finally show the queue.

### Expected/Actual behaviour

If the remote client reports the queue to the server (which it does), that queue should be shown on the controlling client, whenever it's available.

### Logs

No errors

### Server version

12.0

### Web version

12.0

### Build version

12.0-rc2

### Platform

Windows

### Browser

Edge

### Additional information

Here's a video the shows the issue exactly:

https://github.com/user-attachments/assets/8d64da19-d120-41c9-a132-51aae4cb73e5

reecertc:
### Changes
Notify queue consumers after the session player's asynchronous playlist refresh completes. Previously, `statechange` could be handled before the playlist was populated, leaving the controlling client's queue empty.

### Issues
Fixes #8182.

### Code assistance
Use of LLM for codebase navigation.

---

- [x] I have provided a *substantive* review of [another web PR](https://github.com/jellyfin/jellyfin-web/pulls?q=is%3Apr+is%3Aopen+-label%3Adependencies+-label%3Ablocked+-label%3Abackend+-label%3Ainvalid+-label%3A%22merge+conflict%22+-label%3Astale+-author%3A%40me).

### DIFF
--- a/src/plugins/sessionPlayer/plugin.js
+++ b/src/plugins/sessionPlayer/plugin.js
@@ -133,7 +133,10 @@ function updateCurrentQueue(instance, session) {
         instance.isPlaylistRendered = true;
     };
 
-    updatePlaylist(instance, current).then(finish, finish);
+    updatePlaylist(instance, current).then(() => {
+        finish();
+        Events.trigger(instance, 'playlistitemadd');
+    }, finish);
 }
 
 function processUpdatedSessions(instance, sessions, apiClient) {


gitedmond:
Hi! I took a look at the changes in `plugin.js`. Firing the `playlistitemadd` event right after the `updatePlaylist` promise resolves makes a lot of sense, it ensures that the UI components are notified once the playlist is actually populated. The logic looks really clean, and I like that it correctly falls back to calling `finish()` on errors. Nice work!

PR #8235, opened 2026-07-17
Show the server-provided name for plugin tuner types

vk496:
### Changes

The tuner device card resolves its title from a hardcoded type list (`getTunerName`) and shows "Unknown" for any type not in it, including plugin-provided tuners. This uses the display name returned by the `TunerHosts/Types` endpoint via a small `useTunerHostTypes` hook, keeping `getTunerName` as a fallback while the query loads.

A plugin can register an `ITunerHost`; it shows up by name in the "Add Tuner Device" dropdown, but once added the device card reads "Unknown". This makes the card match the dropdown.

Motivating case: https://github.com/vk496/jellyfin-htsp-tuner

### Issues

Part of jellyfin/jellyfin#17372. The guide provider list has the same limitation but needs a new API, proposed in jellyfin/jellyfin#17371.

### Code assistance

---

* [x] I have provided a *substantive* review of [another web PR](https://github.com/jellyfin/jellyfin-web/pulls?q=is%3Apr+is%3Aopen+-label%3Adependencies+-label%3Ablocked+-label%3Abackend+-label%3Ainvalid+-label%3A%22merge+conflict%22+-label%3Astale+-author%3A%40me).

### DIFF
--- a/src/apps/dashboard/features/livetv/api/useTunerHostTypes.ts
+++ b/src/apps/dashboard/features/livetv/api/useTunerHostTypes.ts
@@ -1,0 +1,14 @@
+import { useQuery } from '@tanstack/react-query';
+
+import { useApi } from 'hooks/useApi';
+import { getLiveTvApi } from '@jellyfin/sdk/lib/utils/api/live-tv-api';
+
+export const useTunerHostTypes = () => {
+    const { api } = useApi();
+
+    return useQuery({
+        queryKey: [ 'TunerHostTypes' ],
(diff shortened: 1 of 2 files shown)


vk496:
Verified against a live server: `TunerHosts/Types` returns `{ Id: "htsp", Name: "HTSP Tuner" }` and the added tuner has no `FriendlyName`, so the card title resolves to "HTSP Tuner" instead of "Unknown". Production build succeeds with the change compiled into the Live TV chunk.

PR #8237, opened 2026-07-18
Fix/8220 pagination state loss

### LINKED ISSUE #8220: Bug: Browser Back returns to first page instead of previous paginated folder list after video playback

### Describe the bug

When browsing a library that contains a large number of folders (about 110), the folder list is split into multiple pages (e.g. 1–100, 101–110, etc.).

If I open a folder from the **101–110** page, then open a video's **Details** page and play the video, pressing the browser's **Back** button does not return to the **101–110** page. Instead, it always returns to the **1–100** page.

This makes navigation very inconvenient when browsing large libraries.

### Reproduction Steps

1. Create or use a library containing approximately 130 folders.
2. Open the folder list and navigate to the **101–110** page.
3. Open any folder on that page.
4. Open a video's **Details** page.
5. Play the video.
6. Press the browser's **Back** button.

## Expected Behavior

The browser should return to the same paginated folder list (**101–110**) that was displayed before opening the folder.

## Actual Behavior

The browser returns to the first page (**1–100**) of the folder list instead of the previously displayed page.

### Logs

_No response_

### Server version

10.11.11

### Web version

10.11.11

### Build version

10.11.11

### Platform

Windows, Android (Chrome and WebView)

### Browser

Chorme

### Additional information

- This only happens when a video is played via its Details page.
- If the video is played directly from the folder view (without opening the Details page), the issue does not occur.
- This behavior is the same on both Windows and Android.
- On Android's default mobile interface, videos always open via the Details page, so the issue occurs every time.
- If the Android browser is switched to Desktop mode, direct playback from the folder view becomes available, and the issue does not occur.
- Browsing photos does not exhibit this issue.
- The problem appears to be related to navigation history or pagination state not being preserved after returning from the video Details page.

gitedmond:
**PR Title:**
`Fix pagination state loss on back navigation from video player`

---

### Changes

Currently, if you navigate deep into folders and play a video (e.g., Library -> Folder -> Details -> Video Player), returning w/ the back button resets the root library's pagination state to the first page. This happens because the legacy view cache manager (`src/components/viewContainer.js`) is hardcoded to hold a maximum of 3 views (`const pageContainerCount = 3`). When navigating 4 levels deep, the oldest view (the Library list) is forcefully evicted and destroyed, forcing it to be re-created from scratch when returning.

To fix this, I simply increased the `pageContainerCount` in `src/components/viewContainer.js` from `3` to `7`. This ensures that all views in a typical deep navigation path remain cached in the DOM, preserving their pagination parameters and scroll positions on back navigation.

### Issues

Fixes #8220

### Code assistance

I used the agy CLI (running Gemini 3.1 Pro) to trace the view stack lifecycle and pinpoint why the root view was being destroyed during deep navigation. I updated the cache size constant, checked it with ESLint, verified TypeScript compiler checks, and confirmed that all 162 unit tests passed.

---

* [x] I have provided a *substantive* review of [another web PR](https://github.com/jellyfin/jellyfin-web/pulls?q=is%3Apr+is%3Aopen+-label%3Adependencies+-label%3Ablocked+-label%3Abackend+-label%3Ainvalid+-label%3A"merge+conflict"+-label%3Astale+-author%3A%40me").


thornbill:
Not a valid fix and looks like an llm policy violation

PR #8242, opened 2026-07-21
unify fullscreen support across all book players

dkanada:
### Changes

Huge QoL change for the comic and PDF players. Also improves code quality by removing local state / custom CSS in the bookPlayer class.

### Issues

None

### Code assistance

None


src/components/playback/playbackmanager.js
```diff
@@ -2266,7 +2266,9 @@ export class PlaybackManager {
 
             options.items = items;
 
-            return player.play(options);
+            return player.play(options).then(() => {
+                onPlaybackStarted(player, options, player.streamInfo, player.streamInfo?.mediaSource);
```
jordanmichaelrushing: playOther handles MediaType === 'Photo' || 'Book'. This change works for books but Photos do not set streamInfo. This will crash on line 3521 when it tries setting `streamInfo.playbackStartTimeTicks`. PhotoPlayer also lacks currentTime()/stopped events, so even guarding the crash leaves the progress timer reporting.. I really don't know honestly. Needs either a player.streamInfo guard or excluding Photo from the new call.

src/plugins/bookPlayer/BookOsd/BookOsd.tsx
```diff
@@ -44,6 +45,20 @@ const BookOsd: FC<BookOsdProps> = ({
     const [layout, setLayout] = useState(settings.pagesPerView === 2);
     const [fullscreen, setFullscreen] = useState(false);
 
+    const updateFullscreen = useCallback((state: boolean) => {
+        if (Screenfull.isEnabled) {
+            void Screenfull.toggle();
```
jordanmichaelrushing, line 50: Are you purposefully ignoring `state` here? I think we should be explicit like below: `state ?
  Screenfull.request() : Screenfull.exit()`

src/plugins/bookPlayer/BookOsd/BookOsd.tsx
```diff
@@ -44,6 +45,20 @@ const BookOsd: FC<BookOsdProps> = ({
...
+        } else if (window.NativeShell) {
+            state ? window.NativeShell.enableFullscreen() : window.NativeShell.disableFullscreen();
+        } else if (document.webkitEnterFullscreen || document.webkitCancelFullscreen) {
```
jordanmichaelrushing: `document.webkitEnterFullscreen` doesn't exist. `webkitEnterFullscreen` gets called on video elements. `document` has `webkitExitFullscreen` and the legacy `webkitCancelFullScreen` (capital S).

`document` doesn't have an enter, it needs to be on an element: `element.webkitRequestFullscreen()` and worth calling out that this only exists on Desktop and iPad, not iPhone. Also worth calling out that Screenful handles most cases and NativeShell handles apps, meaning this logic branch would theorectically only serve iOS web (and maybe some random edge cases) and since there isn't an enter fullscreen option available: this pathway is dead and should be removed. I would recommend re-hiding the fullscreen button on iOS.

src/apps/legacy/features/playback/utils/playbackSubscriber.ts
```diff
@@ -21,7 +21,7 @@ export interface PlaybackSubscriber {
     onPlayerChange?(e: Event, player: PlayerPlugin, target: PlayTarget, previousPlayer: PlayerPlugin): void
     onPromptSkip?(e: Event, mediaSegment: MediaSegmentDto): void
     onPlayerError?(e: Event, error: PlayerError): void
-    onPlayerFullscreenChange?(e: Event): void
+    onPlayerFullscreenChange?(e: Event, fullscreen: boolean): void
```
jordanmichaelrushing, line 24: fullscreen is added as an arg, but nothing consumes the event?

src/plugins/bookPlayer/BookOsd/BookOsd.tsx
```diff
@@ -119,13 +135,11 @@ const BookOsd: FC<BookOsdProps> = ({
                     />
                 )}
 
-                {onToggleFullscreen && (
```
jordanmichaelrushing, line 122: See my previous comment about iOS web. This should still be hidden when `!Screenfull.isEnabled && !window.NativeShell`

src/components/playback/playbackmanager.js
jordanmichaelrushing: Might want to check functionality potentially talk about this change also because `onPlaybackStarted` → `setCurrentPlayerInternal(player)`. `playOther` doesn't stop the current player, which for me personally is something that I love because I can listen to an audio book and read at the same time. So we definitely shouldn't remove that functionality. But by calling `setCurrentPlayerInternal` the now playing controls detach from the audio. I don't know if this is intentional or not but as far as I can tell it does change the functionality. And it's probably worth a discussion if it was intentional.

src/plugins/pdfPlayer/plugin.js
```diff
@@ -261,7 +267,7 @@ export class PdfPlayer {
 
         // load any missing pages in the cache
         for (const page of pages) {
-            if (!this.pages[page]) {
+            if (!this.pages[page] || this.dimensions !== window.innerWidth + window.innerHeight) {
```
jordanmichaelrushing: Rotation wouldn't trigger this. You should probably store what the current width and height are and then compare those values. Summing them will always be the same upon rotation

src/plugins/pdfPlayer/plugin.js
```diff
@@ -271,6 +277,10 @@ export class PdfPlayer {
 
         // show the requested page
         canvas?.parentNode.replaceChild(this.pages[prefix + number], canvas);
+        this.currentSrc = () => this.pages[prefix + number];
```
jordanmichaelrushing: Nit: Comics returns a page number and PDF returns a canvas element. Not sure if a conversion is needed, doesn't break now it's just a null check.

src/components/playback/playbackmanager.js
```diff
@@ -2266,7 +2266,9 @@ export class PlaybackManager {
 
             options.items = items;
 
-            return player.play(options);
+            return player.play(options).then(() => {
```
jordanmichaelrushing: Nit: `playOther` no longer returns `player.play()`'s resolved value

jordanmichaelrushing (changes requested):

src/plugins/pdfPlayer/plugin.js
dkanada: BookPlayer returns the download URL so I aligned with that value.

src/plugins/bookPlayer/BookOsd/BookOsd.tsx
dkanada: Good catch - it's checking state again before toggling fullscreen.

src/apps/legacy/features/playback/utils/playbackSubscriber.ts
dkanada: I used this value in a previous iteration of the pull request. It's technically unused but I think it improves the interface so I left it for future improvements.

jordanmichaelrushing (approved):

src/plugins/bookPlayer/plugin.js
```diff
@@ -340,7 +324,7 @@ export class BookPlayer {
                     flow: 'paginated'
                 });
 
-                this.currentSrc = downloadHref;
+                this.currentSrc = () => downloadHref;
```
thornbill, line 340: This seems unrelated or unused?

src/plugins/comicsPlayer/style.scss
```diff
@@ -3,8 +3,8 @@
 
     .slideshowSwiperContainer {
         position: relative;
-        top: 5vh;
-        height: 90vh;
+        top: calc(5vh + env(safe-area-inset-top));
+        height: calc(90vh - env(safe-area-inset-top) - env(safe-area-inset-bottom));
```
thornbill, line 10: probably need to include the rules without `calc()` as fallback for legacy browsers

src/plugins/pdfPlayer/style.scss
```diff
@@ -2,8 +2,8 @@
     position: relative;
     display: grid;
     place-items: center;
-    top: 5vh;
-    height: 90vh;
+    top: calc(5vh + env(safe-area-inset-top));
+    height: calc(90vh - env(safe-area-inset-top) - env(safe-area-inset-bottom));
```
thornbill, line 9: same here

src/plugins/bookPlayer/plugin.js
dkanada: It's unused in the book players but it's currently expected somewhere in playbackManager. I have a local branch that starts documenting these methods in the PlayerPlugin class when we leave the RC cycle.

thornbill (approved):

PR #8245, opened 2026-07-21
Prevent backdrop from moving when menus are opened

Shadowghost:
Backdrops are slightly moving when opening any menu.

### Changes
* Add `disableScrollLock` to the menu components so they don't influence the backdrop

### Code assistance
None

---


theguymadmax:
@MBR-0001 I believe you had this issue?

MBR-0001:
Yeah, I noticed this a few months back, but its still happening on these buttons

image

if we are going to change this it might be a good idea to do it globally (mui supports this afaik, via `createTheme`), but I'd first ask an ux expert (@erikbucik) if it makes sense; it must be off by default for a reason

thornbill:
I believe there was some discussion on this in chat... I _think_ the conclusion was that scroll lock should be enabled when menus are opened?

I can't seem to replicate this in firefox, but it sounds like we might be able to use `100vw` on the backdrop container to include the scrollbars in the with of the element.

MBR-0001:
Afaik the conclusion was (copy paste):

> Should scrolling be allowed when there's an overlay present?
> 
> When there's a large popup (that dims the bg) - absolutely not.
> And a small menu popup - maybe, probably not?

the popups touched in this PR would be the small ones, so this may be ok ux-wise, but even if this is merged it wont solve all the instances of backdrop moving around, since there are other popups (mentioned in my previous msg). I'm unsure whether the filters popup could be categorized as a small popup.

Adding 100vw to backdrop solves this in my short in-browser testing though

Shadowghost:
I can change this PR to the 100vw and we can go ahead with that for now?

thornbill:
sounds good to me :+1:

src/components/OffsetAppBar.tsx
```diff
@@ -65,6 +66,7 @@ const OffsetAppBar: FC<PropsWithChildren<OffsetAppBarProps>> = ({
         <>
             <AppBar
                 {...props}
+                className={[ 'mui-fixed', className ].filter(Boolean).join(' ')}
```
thornbill: huh?

dkanada: @Shadowghost I think you're looking for the classNames method we use elsewhere.

src/components/OffsetAppBar.tsx
```diff
@@ -65,6 +67,7 @@ const OffsetAppBar: FC<PropsWithChildren<OffsetAppBarProps>> = ({
         <>
             <AppBar
                 {...props}
+                className={classNames('mui-fixed', className)}
```
MBR-0001: I noticed that there are two 'mui-fixed' classes in html, after some digging it seems this is not needed because mui already does this https://github.com/mui/material-ui/blob/48c6663a66b7eacbd9c7c411863ab5d18dbeea88/packages/mui-material/src/AppBar/AppBar.js#L195-L199

MBR-0001 (dismissed):

nyanmisaka:
I have this issue too. It seems this fix could be merged with just a few minor changes? @Shadowghost

src/components/OffsetAppBar.tsx
```diff
@@ -20,6 +20,7 @@ interface OffsetAppBarProps extends AppBarProps {
  */
 const OffsetAppBar: FC<PropsWithChildren<OffsetAppBarProps>> = ({
     children,
```
thornbill, line 22: This seems unnecessary now

thornbill:
~Haven't tested but we might need a fallback to 100% for the ancient browsers.~

Should be fine. It's supported starting in Chrome 26 and we support back to the (rough) equivalent of Chrome 27. https://caniuse.com/viewport-units

thornbill (approved):

PR #8282, opened 2026-07-30
Fix playbackstop listener leak on the movies page

### LINKED ISSUE #8280: listener leak in moviesrecommended.js

### Describe the bug

`src/apps/legacy/controllers/movies/moviesrecommended.js` subscribes to the `playbackstop` event
inside its `viewshow` handler but never unsubscribes. One listener is leaked per visit to the movies
page, growing without bound for the lifetime of the session.

```js
    ...
    Events.on(playbackManager, 'playbackstop', onPlaybackStop);   // line 395
    inputManager.on(window, onInputCommand);
});
view.addEventListener('viewbeforehide', function () {
    inputManager.off(window, onInputCommand);                     // line 399
    // <-- no Events.off for playbackstop
});
```

The subscription is unconditional and lives in `viewshow`, so it re-runs on every show including
cached restores. `Events.on` is a plain `push` with no dedupe (`src/utils/events.ts`), and
`onPlaybackStop` is defined per controller instance (line 350), so each show adds an entry.

**The equivalent TV file does it correctly.** `src/apps/legacy/controllers/shows/tvrecommended.js`
is the same page for Shows and has the symmetric pair:

```js
Events.on(playbackManager, 'playbackstop', onPlaybackStop);    // :371  (in viewshow)
...
Events.off(playbackManager, 'playbackstop', onPlaybackStop);   // :377  (in viewbeforehide)
```

The divergence between these two near-identical files suggests the movies version is an oversight.

Measured live by counting subscribers whose body contains this handler's unique
`triggerTabChange` call (`Events` stores callbacks on the target object as `obj._callbacks[type]`):

| Step | Leaked handlers | Total `playbackstop` subscribers |
|---|---|---|
| hard reload at `#/home` | **0** | 12 |
| after visit 1 (`#/movies` → `#/home`) | **1** | 18 |
| after visit 2 | **2** | 19 |
| after visit 3 | **3** | 20 |
| after visit 4 | **4** | 21 |

Exactly +1 per visit, perfectly linear, never released.

### Reproduction Steps

1. Set the client view setting to **Desktop (Legacy)**. This is required — see Additional
   information.
2. Hard reload to the home page.
3. In the browser console, record the number of `playbackstop` subscribers on the `playbackManager`
   singleton.
4. Navigate to the Movies page, then back to Home. Repeat several times.
5. Re-read the count: it grows by one per visit and never returns to baseline.

Filtering the subscriber list for a function body containing `triggerTabChange` isolates this
specific handler from other legitimate subscribers.

### Expected/Actual behaviour

**Expected:** the `playbackstop` subscription registered when the movies page is shown is released
when the page is hidden, as `tvrecommended.js` does — so the subscriber count returns to baseline
after leaving the page.

**Actual:** the subscription is never released. Each visit adds a permanently-registered handler.
Every one of those stale handlers is invoked on every playback stop, and each retains its controller
instance and view references.

### Logs

No errors are logged. The symptom is accumulating listeners rather than a visible failure.

### Server version

10.12.0

### Web version

12.0.0

### Build version

12.0.0

### Platform

Linux x86_64

### Browser

Firefox 140.0

### Only reachable on the legacy app

`moviesrecommended` is registered only in the legacy app routes
(`src/apps/legacy/routes/legacyRoutes/user.ts:98`). On the modern app, `/movies` resolves to a React
page, this controller never loads, and the count stays flat with zero leaked handlers — I measured
this, including with a real movie library present. So the bug requires the **Desktop (Legacy)**
client view setting to reproduce.

This is worth stating explicitly because a first pass on the modern app measured completely flat and
the defect nearly got written off as unreachable dead code.

### Impact

Each stale handler runs on every playback stop and, when the stopped item's media type is `Video`,
calls `mainTabsManager.getTabsElement().triggerTabChange()` against a possibly-stale view. Beyond
the listener and memory growth, that risks redundant tab re-renders or errors against detached DOM.

Growth here is **unbounded** — it is keyed to `viewshow` firing, not to how many views the 3-slot
view cache holds.

### Suggested fix

One line, mirroring `tvrecommended.js:377` — add to the existing `viewbeforehide` handler in
`moviesrecommended.js`:

```js
Events.off(playbackManager, 'playbackstop', onPlaybackStop);
```

### Possibly related

While measuring the above I also observed `playbackstop` subscribers accumulating from
`emby-itemscontainer` elements on cached item-detail views — the container subscribes in
`attachedCallback` and unsubscribes in `detachedCallback`, but cached views are hidden rather than
removed from the DOM, so detach never fires. That looked like a separate defect with a different
root cause, so I have not included it here.

sevenrats:
moviesrecommended subscribed to playbackstop in its viewshow handler but never unsubscribed, so one listener was leaked per visit to the page and accumulated without bound for the life of the session. Each stale handler ran on every playback stop and called triggerTabChange against a view that may no longer be current.

### Changes

Release the subscription in viewbeforehide, mirroring the symmetric pair already present in the equivalent Shows controller (tvrecommended.js).

### Issues

fixes #8280 

### Code assistance

assisted by fable

---

### DIFF
--- a/src/apps/legacy/controllers/movies/moviesrecommended.js
+++ b/src/apps/legacy/controllers/movies/moviesrecommended.js
@@ -397,6 +397,7 @@ export default function (view, params) {
     });
     view.addEventListener('viewbeforehide', function () {
         inputManager.off(window, onInputCommand);
+        Events.off(playbackManager, 'playbackstop', onPlaybackStop);
     });
     for (const tabController of tabControllers) {
         if (tabController.destroy) {
--- a/src/apps/legacy/controllers/movies/moviesrecommended.js
+++ b/src/apps/legacy/controllers/movies/moviesrecommended.js
@@ -347,7 +347,9 @@ export default function (view, params) {
         }));
     }
 
-    function onPlaybackStop(e, state) {
+    function onPlaybackStop(e, stopInfo) {
+        const state = stopInfo.state;
+
         if (state.NowPlayingItem && state.NowPlayingItem.MediaType == 'Video') {
             renderedTabs = [];
             mainTabsManager.getTabsElement().triggerTabChange();
--- a/src/apps/legacy/controllers/shows/tvrecommended.js
+++ b/src/apps/legacy/controllers/shows/tvrecommended.js
@@ -316,7 +316,9 @@ export default function (view, params) {
         });
     }
 
-    function onPlaybackStop(e, state) {
+    function onPlaybackStop(e, stopInfo) {
+        const state = stopInfo.state;
+
         if (state.NowPlayingItem && state.NowPlayingItem.MediaType == 'Video') {
             renderedTabs = [];
             mainTabsManager.getTabsElement().triggerTabChange();


dmitrylyzo:
From my unpublished changes (5 years ago 😅), there is probably an error in the handlers - `state` instead of `stopInfo`.
https://github.com/dmitrylyzo/jellyfin-web/commit/86b0b54fc648563a0702438d416e6211c3adcf36

sevenrats:
> From my unpublished changes (5 years ago 😅), there is probably an error in the handlers - `state` instead of `stopInfo`.

i dont take your meaning. is this not the correct fix?

dmitrylyzo:
> i dont take your meaning. is this not the correct fix?

It is correct - it fixes the leak. But there's another bug.
I don't remember why I didn't push the changes. Probably didn't take it seriously.

sevenrats:
nice. it seemed right in my tests.  how do you want to proceed? land this and then go after the second one? copy your fixes into this pr? close this pr and pull in your changes instead? down for whatever.

dmitrylyzo:
If my changes (`stopInfo`) are still valid (I haven't tested them since then), you can cherry pick that commit keeping your commit (symmetrical unsubscribing).

It seems that `playbackstop` from the player passes `state`, and from the playback manager - `stopInfo`.
https://github.com/jellyfin/jellyfin-web/blob/731ce7ef0bb355ad81bdc8ebd06bfc9ea92c2829/src/components/playback/playbackmanager.js#L3402-L3403

sevenrats:
@dmitrylyzo verified the playbackstop bug still exists on master and cherry picked your commit.
@thornbill this is ready for review at your convenience.

dmitrylyzo:
> cherry picked

"Cherry-picked" is when the authorship is preserved. 😉

sevenrats:
> > cherry picked
> 
> "Cherry-picked" is when the authorship is preserved. 😉

yeah I did do a cherry pick but by the time I finished I don't think there any code was left. I wound up refactoring your changes before I finalized the pick without realizing it.  if you insist that I go back and untangle it I will but I don't think any of your code is actually in here or the cherry pick would show.  If I'm wrong please lmk I don't want to be one of those noobish git ppl.

dmitrylyzo:
> yeah I did do a cherry pick but by the time I finished I don't think there any code was left. I wound up refactoring your changes before I finalized the pick without realizing it. if you insist that I go back and untangle it I will but I don't think any of your code is actually in here or the cherry pick would show. If I'm wrong please lmk I don't want to be one of those noobish git ppl.

Ah, right, the source tree has been rearranged.
I think you can rebase (to drop the merge commit) and amend the commit (add `Co-authored-by: Dmitry Lyzo <56478732+dmitrylyzo@users.noreply.github.com>` to the commit message or set the commit author using `git commit --amend --author="Dmitry Lyzo <56478732+dmitrylyzo@users.noreply.github.com>" --no-edit`).

thornbill (approved):

PR #8284, opened 2026-07-30
Fix Now Playing page not following player switches after browser back

### LINKED ISSUE #8283: Now Playing page stops following player switches after a browser-back restore

### Describe the bug

The Now Playing page (`/queue`) permanently stops reacting to player switches after you navigate
away from it and return using the browser back button.

`playback/queue/index.js` calls `remoteControl.init()` **once**, in the controller factory body,
but `remoteControl.destroy()` on **every** `viewbeforehide`. `init()` is the only place that
subscribes to the `playerchange` event; `destroy()` unsubscribes it. Nothing re-subscribes.

Because the legacy view manager caches view DOM nodes and does not re-run the controller factory on
a back-restore (guarded by the `initComplete` expando in
`src/components/viewManager/viewManager.js`), the subscription is never re-established.

Relevant code:

| File | Line | What |
|---|---|---|
| `src/apps/legacy/controllers/playback/queue/index.js` | 13-14 | `remoteControl.init()` in factory body |
| `src/apps/legacy/controllers/playback/queue/index.js` | 50-58 | `viewbeforehide` → `remoteControl.destroy()` |
| `src/components/remotecontrol/remotecontrol.js` | 901 | `init()` → `Events.on(playbackManager, 'playerchange', …)` |
| `src/components/remotecontrol/remotecontrol.js` | 910-914 | `onDialogClosed()` (via `destroy()`) → `Events.off(…)` |
| `src/components/remotecontrol/remotecontrol.js` | 916-918 | `onShow()` → `bindToPlayer` only, no re-subscribe |

I verified this by measuring the live subscriber count. `Events` stores callbacks on the target
object as `obj._callbacks[type]` (`src/utils/events.ts`), so the subscription is directly
observable:

| Step | Route | `playerchange` subscribers | |
|---|---|---|---|
| clean page load | `#/home` | **12** | baseline, `/queue` never loaded |
| fresh view load | `#/queue` | **13** | `init()` subscribes (+1) |
| navigate away | `#/home` | **12** | `destroy()` unsubscribes (−1) |
| **browser back** | `#/queue` | **12** | **never re-subscribes** ← bug |
| hard reload | `#/queue` | **13** | control: fresh load works again |

At step 4 the restored view was visible and otherwise functional, with
`nowPlayingPage.initComplete === true` confirming the controller factory did not re-run. The
sequence reproduced twice; the hard-reload control rules out a measurement artifact.

Note that `onShow()` does re-bind the **per-player** events, so the page is not dead after a
restore — only the `playbackManager`-level `playerchange` subscription is lost.

### Tracked by function identity

Raw subscriber counts are noisy — other modules subscribe and unsubscribe to `playerchange`
independently, so the total moves for unrelated reasons. Isolating the remote control's own handler
by diffing the subscriber array across a fresh `/queue` load (identity comparison, not source
matching) gives an unambiguous result:

| Step | Remote control handler present |
|---|---|
| fresh `/queue` load | **true** |
| navigate away | **false** |
| **browser back** | **false** ← the bug |
| evict from cache, load again | **true** (control) |

At the third step the view was visible with `initComplete === true`.

### Functional consequence

The missing subscription has an observable effect, confirmed by a matched-pair test.
`updateSupportedCommands()` — reachable only via `bindToPlayer`, which `onPlayerChange` calls —
toggles the `hide` class on command buttons. Corrupting those classes and then dispatching
`playerchange` exactly as `triggerPlayerChange` does
(`Events.trigger(playbackManager, 'playerchange', [player, target, previous])`) shows whether the
page can still reconcile its UI:

| | fresh page | restored page |
|---|---|---|
| handler present | true | false |
| command buttons un-hidden (corrupted) | 5 → 0 | 4 → 0 |
| re-hidden after dispatching `playerchange` | **4** | **0** |
| UI reconciled to the player | **yes** | **no** |

So on a freshly loaded Now Playing page a player change makes the page reconcile its controls to
the player's supported commands; on a restored page the identical event does nothing.

### Confirmed with a real cast switch

The above used a synthetic dispatch. The end-to-end path was then confirmed with an actual second
playback target: a second browser session logged into the same server registers as a
remote-control target, so `playbackManager.getTargets()` returns it and casting to it is a genuine
player change (`Html Audio Player` → `Remote Control`, `isLocalPlayer` false).

| | fresh page | restored page (before fix) | restored page (after fix) |
|---|---|---|---|
| handler present | true | **false** | true |
| real `playerchange` fired on cast | **yes** | yes | **yes** |
| remote control re-bound to new player | yes | **no** | **yes** |

"Re-bound" was verified by inspecting the new player's subscriber list: `bindToPlayer` is the only
thing that subscribes `playlistitemadd` / `playlistitemmove` / `playlistitemremove`, and those
appear on the newly-active remote player only when the handler is present.

So the full chain is measured, not inferred: casting to another device fires `playerchange`, and a
restored page without the subscription never receives it.

### Reproduction Steps

1. Start playing something so the Now Playing bar appears.
2. Click the Now Playing bar to open the Now Playing page (`/queue`).
3. Navigate away — click into any library item.
4. Press the **browser back button** to return to the Now Playing page.
5. Switch the active player: cast to another device, or pick a different client from the
   player-selection menu.
6. The page keeps showing and controlling the **previous** player.

Step 4 must land within the 3-slot view cache window (`src/components/viewContainer.js`) and at the
same URL string; `/queue` has no query string, so the cache key is stable. Navigating through 4+
distinct routes before returning evicts the node, forces a fresh load, and the bug does not occur —
which is a useful way to confirm the view cache is the cause.

### Expected/Actual behaviour

**Expected:** after returning to the Now Playing page via browser back, switching the active player
re-points the page at the new player, exactly as it does on a freshly loaded Now Playing page.

**Actual:** the page remains bound to the previous player. Transport controls and the volume slider
target the stale player. The position display and other per-player state still update, which makes
the failure easy to miss.

### Logs

No errors are logged; this is a silent missing subscription.

### Server version

10.12.0

### Web version

12.0.0

### Build version

12.0.0

### Platform

Linux x86_64

### Browser

Firefox 140.0

### Why the obvious fix doesn't work

Moving `init()` into `viewshow` is **not** viable — `init()` is not idempotent. It does
`insertAdjacentHTML` of the volume-control markup (`remotecontrol.js:884` / `:891`) and adds
`submit` listeners to the message forms (`:899-900`). Calling it per-show would duplicate the volume
slider and stack duplicate form handlers.

Suggested fixes, cheapest first:

1. Move the `playerchange` subscription out of `init()` into `onShow()`, paired with the
   `Events.off` already in `onDialogClosed()`. This makes subscribe/unsubscribe symmetric and
   matches what `destroy()` already tears down. `playback/queue/index.js` is the module's only
   consumer, so there is no dialog use case to regress — its `dlg` / `onDialogClosed` naming is
   vestigial from an earlier dialog-based lifecycle, which is likely how the mismatch arose.
2. Split `init()` into a one-time DOM-construction part and a re-runnable binding part.
3. Call `destroy()` only from `viewdestroy` rather than `viewbeforehide`. Weakest option:
   `viewdestroy` is not guaranteed to fire — `viewContainer.reset()` clears `innerHTML` without
   dispatching it.

### Fix verified locally

I implemented option 1 and confirmed it against a rebuilt web client. The change is two lines —
remove the `Events.on` from `init()`, and in `onShow()`:

```js
function onShow(context) {
    // Paired with the Events.off in onDialogClosed. init() only runs once per view node,
    // but the view is hidden and shown repeatedly, so subscribing there left the restored
    // page unsubscribed from playerchange for good. Off first so a repeated show cannot
    // stack duplicate subscriptions.
    Events.off(playbackManager, 'playerchange', onPlayerChange);
    Events.on(playbackManager, 'playerchange', onPlayerChange);
    bindToPlayer(context, playbackManager.getCurrentPlayer());
}
```

The defensive `Events.off` first is not strictly required today — `onShow` and `destroy` are called
1:1 from `viewshow` / `viewbeforehide` — but `Events.on` is a plain `push` with no dedupe, so it
makes a repeated show self-correcting. It mirrors the existing idiom in `emby-playstatebutton` and
`emby-ratingbutton`, where `bindEvents` calls `clearEvents` first for the same reason.

Results, same measurements as above:

| check | before | after |
|---|---|---|
| handler present after browser back | false | **true** |
| **real cast switch on restored page re-binds the remote control** | **no** | **yes** |
| UI reconciled on restored page (corrupt-and-dispatch) | false | **true** (4 buttons re-hidden) |
| handler copies after 5 away/back cycles | — | **1** (no stacking) |
| volume slider / mute button / message form count after 6 cycles | — | **1 each** (no duplicated markup) |

### Root-cause class

The general defect is *side-effectful setup at controller construction paired with destructive
teardown on `viewbeforehide`, with no restore-aware re-init.* The codebase has two valid patterns
and this controller mixes them:

- **Symmetric** — full teardown on `viewbeforehide`, unconditional re-bind on `viewshow`, never
  consults `isRestored`. Reference: the video OSD
  (`src/apps/legacy/controllers/playback/video/index.js:1664` / `:1712`).
- **Guarded** — bind once at construction, wrap re-work in `if (!e.detail.isRestored)`. Used by
  `list.js`, `livetvsuggested.js`, `session/selectServer/index.js`.

`playback/queue/index.js` binds once (guarded pattern) but tears down on every hide (symmetric
pattern), and has no `isRestored` handling anywhere.

sevenrats:
### Changes

remoteControl.init() runs once per view node, but destroy() runs on every
viewbeforehide. Since init() was the only place subscribing to playerchange
and destroy() unsubscribes it, a cached view restored by the browser back
button was left unsubscribed for good: onShow() re-binds the per-player
events but never re-established the playbackManager-level subscription.

Move the subscription to onShow(), pairing it with the Events.off already
in onDialogClosed(). Events.off first so a repeated show cannot stack
duplicate subscriptions.

Verified against a second browser session acting as a real cast target:
before, casting while on a restored Now Playing page left the page bound
to the old player; after, it re-binds. No duplicated volume control or
form handlers across repeated show/hide cycles.

### Issues

Fixes #8283 

### Code assistance

Assisted by Fable

---

### DIFF
--- a/src/components/remotecontrol/remotecontrol.js
+++ b/src/components/remotecontrol/remotecontrol.js
@@ -895,7 +895,6 @@ export default function () {
         bindEvents(context);
         context.querySelector('.sendMessageForm').addEventListener('submit', onMessageSubmit);
         context.querySelector('.typeTextForm').addEventListener('submit', onSendStringSubmit);
-        Events.on(playbackManager, 'playerchange', onPlayerChange);
 
         if (layoutManager.tv) {
             const positionSlider = context.querySelector('.nowPlayingPositionSlider');
@@ -911,6 +910,9 @@ export default function () {
     }
 
     function onShow(context) {
+        // Subscribe here, not in init(): init runs once per node but hide/show repeat.
+        Events.off(playbackManager, 'playerchange', onPlayerChange);
+        Events.on(playbackManager, 'playerchange', onPlayerChange);
         bindToPlayer(context, playbackManager.getCurrentPlayer());
     }
 


src/components/remotecontrol/remotecontrol.js
```diff
@@ -911,6 +910,9 @@ export default function () {
     }
 
     function onShow(context) {
+        // Subscribe here, not in init(): init runs once per node but hide/show repeat.
+        Events.off(playbackManager, 'playerchange', onPlayerChange);
```
thornbill, line 914: Are there cases where `onDialogClosed` is not being called to unsubscribe the event listener?

sevenrats: Not that I'm aware of, but the codebase straddles two approaches currently.  In one approach symmetry is baked into the module and in another it is left to the caller.  The extra Events.off mirrors [emby-ratingbutton](https://github.com/jellyfin/jellyfin-web/blob/8d57e28226fcd021087bb3798f197c073918e3f9/src/elements/emby-ratingbutton/emby-ratingbutton.js#L76) and [emby-playstatebutton](https://github.com/jellyfin/jellyfin-web/blob/8d57e28226fcd021087bb3798f197c073918e3f9/src/elements/emby-playstatebutton/emby-playstatebutton.js#L77), which force symmetry this same way. Events.on has no dedupe, so an unbalanced show would silently double onPlayerChange. This change makes that impossible by mimicking existing idioms. I don't think it's an absolutely necessary component of the bugfix, but it seems more right to me.

thornbill (approved):

PR #7992, opened 2026-06-03
Allow common link-protocol schemes in the login disclaimer

dfederm:
### Changes
The login page renders the admin-configured login disclaimer through `markdown-it` and `DOMPurify` with no options. DOMPurify allowed-URI regex only accepts `http(s)`, `ftp(s)`, `mailto`, `tel`, `callto`, `cid`, and `xmpp`, so anchors using `matrix:`, `tg:`, `whatsapp:`, `signal:`, `irc:`, or `ircs:` don't work.
 
This extracts the sanitize pipeline into a small helper and pass an `ALLOWED_URI_REGEXP` that mirrors `DOMPurify`'s default with those six schemes added. `javascript:`, `data:`, `vbscript:`, and `file:` remain blocked.

The disclaimer is admin-only configuration and existing post-sanitize code already sets `rel=noopener noreferrer` and `target=_blank` on every link, so the additional schemes do not widen the attack surface beyond opening an external app.

### Testing
I manually tested by using this in the login disclaimer:

```md
Contact us:

 - [Matrix](matrix:r/jellyfin:matrix.org)
 - [Telegram](tg:resolve?domain=jellyfin)
 - [WhatsApp](whatsapp://send?phone=12345)
 - [Signal](signal:+12345)
 - [IRC](irc://irc.libera.chat/jellyfin)
 - [Email](mailto:test@example.com)
```

And then logging out and seeing the actual HTML generated on the login page.

Before the fix:
```html
<div class="loginDisclaimer"><p>Contact us:</p>
<ul>
<li><a rel="noopener noreferrer" target="_blank" class="button-link emby-button" is="emby-linkbutton">Matrix</a></li>
<li><a rel="noopener noreferrer" target="_blank" class="button-link emby-button" is="emby-linkbutton">Telegram</a></li>
<li><a rel="noopener noreferrer" target="_blank" class="button-link emby-button" is="emby-linkbutton">WhatsApp</a></li>
<li><a rel="noopener noreferrer" target="_blank" class="button-link emby-button" is="emby-linkbutton">Signal</a></li>
<li><a rel="noopener noreferrer" target="_blank" class="button-link emby-button" is="emby-linkbutton">IRC</a></li>
<li><a href="mailto:test@example.com" rel="noopener noreferrer" target="_blank" class="button-link emby-button" is="emby-linkbutton">Email</a></li>
</ul>
</div>
```

After the fix:
```html
<div class="loginDisclaimer"><p>Contact us:</p>
<ul>
[...]

PR #8286, opened 2026-07-30
Replace isRelativeDay with date-fns isYesterday

Grinv:
### Changes

`isRelativeDay` returned the wrong result on the 1st of a month. It compared
against `getDate() + offsetInDays`, which is 0 on the 1st, so the last day of the
previous month never matched. Upcoming Episodes showed a full date instead of
"Yesterday".

I replaced it with `isYesterday` from `date-fns`.

### Code assistance

I used an LLM for the off-by-one and the review feedback. I checked the
replacement myself: it behaves the same as the fixed helper across timezones, and
build, lint and tests pass.

---

* [x] I have provided a *substantive* review of another web PR (#7992, #8051, #8167).


dmitrylyzo (approved):

thornbill (changes requested):
The only usages of this function are to check if a date is "yesterday". Since the goal is to replace our custom date/time handling with `date-fns`, we should just remove this function and replace the usages with `isYesterday`.

https://date-fns.org/v4.4.0/docs/isYesterday

Grinv:
@thornbill Done. Removed `isRelativeDay` and both call sites now use `isYesterday`.

Also updated the PR title and description.

PR #8289, opened 2026-07-31
Reuse the cached item query on the details page

### LINKED ISSUE #8277: Details/items using two requests for the same payload

### Describe the bug

Jellyfin web send duplicate requests with different URLs for the Item. **Can be tested in unstable demo**

Issue is on anything detail (Show, Movie Season episode) and not on library listing

Caminandes: Llamigos
https://demo.jellyfin.org/unstable/web/#/details?id=c7eb95d14156c803611cf4a78ac469f6&serverId=f0b3381645f04afb9a0e392e74b6a1b0

URLs - Payload is identical
https://demo.jellyfin.org/unstable/Users/4ed1b8b42a7c4ea682f0fe5d08d4e278/Items/c7eb95d14156c803611cf4a78ac469f6
https://demo.jellyfin.org/unstable/Items/c7eb95d14156c803611cf4a78ac469f6?userId=4ed1b8b42a7c4ea682f0fe5d08d4e278

Also, the config gets required multiple times (sally a cache hit, but the behavior changed to my knowledge)

Image

Another with Waterfall

Image

### Reproduction Steps

1. Go to detail page
2. look at network tab

### Expected/Actual behaviour

Only load the required endpoints

### Logs

_No response_

### Server version

12.0

### Web version

12.0

### Build version

Release

### Platform

Linux

### Browser

all

### Additional information

Noticed locally replicated in unstable demo

quepasaquepasa:
Closes #8277

**Cause**

Opening a details page fetches the same item twice over two different routes:

1. `appRouter.showItem(id, serverId)` resolves the string id via `queryClient.fetchQuery(getItemQuery(...))` → `GET /Items/{id}?userId={userId}` (SDK)
2. the details controller then calls `getPromise()` → `apiClient.getItem(userId, id)` → `GET /Users/{userId}/Items/{id}` (legacy apiclient)

Same payload, different URLs, so the second request can't be served from the react-query cache the first one just populated.

**Fix**

`getPromise()` now goes through the same `getItemQuery`/`queryClient` path as the router. With the client's `staleTime` of 60s the item the router already fetched is reused, so the duplicate request disappears. The legacy call is kept as a fallback for when no `Api` instance is available for the server.

**Cache correctness**

Since the details page can now be served from cache, the three in-page mutations that call `reload()` explicitly invalidate the item first, so they still refetch rather than re-rendering pre-mutation data:

- splitting alternate versions (`DELETE Videos/{id}/AlternateSources`)
- cancelling a timer
- returning from the metadata editor with `result.updated`

**Verification**
- `npx eslint src/apps/legacy/controllers/itemDetails/index.js` → 0 errors
- `npx tsc --noEmit` → clean

Note: the comment on the issue attributes the duplicate to the `regenerator-runtime` polyfill; that isn't the cause here — the two requests come from the two call paths above.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -75,6 +75,16 @@ export default tseslint.config(
                 {
                     property: 'replaceChildren',
                     message: 'replaceChildren is not supported in all target browsers'
+                },
+                {
+                    object: 'Number',
+                    property: 'parseFloat',
+                    message: 'Use the global parseFloat to match the rest of the codebase'
+                },
+                {
+                    object: 'Number',
+                    property: 'parseInt',
+                    message: 'Use the global parseInt to match the rest of the codebase'
                 }
             ],
             'no-return-assign': 'error',
--- a/src/apps/legacy/controllers/itemDetails/index.js
+++ b/src/apps/legacy/controllers/itemDetails/index.js
@@ -28,8 +28,10 @@ import itemShortcuts from 'components/shortcuts';
 import { AppFeature } from 'constants/appFeature';
 import { EventType } from 'constants/eventType';
 import { ItemAction } from 'constants/itemAction';
+import { getItemQuery } from 'hooks/useItem';
 import globalize from 'lib/globalize';
 import { ServerConnections } from 'lib/jellyfin-apiclient';
+import { queryClient } from 'utils/query/queryClient';
 import browser from 'scripts/browser';
 import datetime from 'scripts/datetime';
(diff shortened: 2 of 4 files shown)


quepasaquepasa:
Closing this myself: I didn't complete the repository's PR template and the description wasn't written to the standard your LLM policy requires. Rather than leave a non-conforming PR open for maintainers to triage, I'm withdrawing it. Apologies for the noise.

PR #8296, opened 2026-07-31
Use the axios type guard instead of any casts in queryClient

Grinv:
### Changes

The `onError` and `retry` handlers in the query cache cast the error to `any`
just to read its HTTP status. All `queryFn` calls go through the Jellyfin SDK,
which uses axios, so `isAxiosError` narrows the type without the cast.

This also drops the `status` and `statusCode` fallbacks. `statusCode` appears
nowhere else in `src/`.

### Code assistance

Used an LLM for the typing, then verified the error shapes myself against how the
SDK and axios actually behave.

---

* [x] I have provided a *substantive* review of another web PR (#7992, #8051, #8167).


thornbill (commented):
This seems unnecessarily complex just to prevent casting as `any`?

Grinv:
@thornbill Removed the helper and its tests, so it's down to a single file. I'm not fond of casts, since `as` hides what the value actually is. Title and description updated.

PR #8302, opened 2026-08-01
implement fade animations on BookOsd to match video player

dkanada:
### Changes

The final touch to align BookOsd with our existing "VideoOsd" interface - fade animations. Click or tap will toggle the visibility immediately, pointer move on desktop will show the OSD manually, and there is a global two second timeout to hide it again.

Personally, I don't *love* the pointer move listener, but it's still an improvement because now the entire screen is available for the book itself when the OSD is hidden. A future pull request will implement a unified progress indicator.

### Issues

None

### Code Assistance

Claude to debug the opacity issue on iOS devices.

### DIFF
--- a/src/plugins/bookPlayer/BookOsd/BookOsd.tsx
+++ b/src/plugins/bookPlayer/BookOsd/BookOsd.tsx
@@ -85,9 +85,10 @@ const BookOsd: FC<BookOsdProps> = ({
         };
 
         const onClick = (event: MouseEvent) => {
-            if ((event.target as Element | null)?.closest?.('.bookOsdRow')) return;
-
+            // apply this before the BookOsd check so IconButton clicks will reset the timer
             scheduleHide();
+
+            if ((event.target as Element | null)?.closest?.('.bookOsdRow')) return;
             setVisible(state => !state);
         };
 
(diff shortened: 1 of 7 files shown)


src/plugins/bookPlayer/BookOsd/BookOsd.tsx
```diff
@@ -73,14 +76,41 @@ const BookOsd: FC<BookOsdProps> = ({
...
+            if (event.pointerType !== 'mouse') return;
+
+            scheduleHide();
+            setVisible(true);
+        };
+
+        const onClick = (event: MouseEvent) => {
+            if ((event.target as Element | null)?.closest?.('.bookOsdRow')) return;
+
+            scheduleHide();
```
MBR-0001, line 89: I suggest moving this before the check above so clicking on OSD buttons delays the hiding

src/plugins/bookPlayer/plugin.js
```diff
@@ -187,26 +186,24 @@ export class BookPlayer {
 
         document.addEventListener('keydown', this.onWindowKeyDown);
         this.rendition?.on('keydown', this.onWindowKeyDown);
-
-        if (browser.safari) {
-            this.addSwipeGestures(document.querySelector('#bookPlayerContainer'));
-        } else {
-            this.rendition?.on('rendered', (e, i) => this.addSwipeGestures(i.document.documentElement));
```
MBR-0001, line 194: removing this breaks swiping on non-safari browsers

MBR-0001 (changes requested):

MBR-0001 (approved):

PR #7777, opened 2026-03-29
Replace libpgs with libbitsub and adds support for vobsub rendering

altqx:
Replace libpgs with libbitsub as suggested in https://github.com/jellyfin/jellyfin/issues/16484 and https://github.com/altqx/libbitsub/issues/1

### Changes
Swap PGS rederer library to libbitsub.
Note: Libbitsub do not exposes a real `aspectRatio` control, libbitsub computes subtitle canvas bounds internally from the video element’s box and the media’s intrinsic aspect ratio.

### Issues
https://github.com/jellyfin/jellyfin/issues/16484

### Code assistance
VS Code autocomplete

Edit: This PR will also include support for VobSub rendering using libbitsub additionally from just replacing libpgs.
Please note that this PR requires https://github.com/jellyfin/jellyfin/pull/16552 to be accepted too.

---

* [x] I have read and followed the [contributing guidelines](https://github.com/jellyfin/jellyfin-web/blob/master/CONTRIBUTING.md).
* [x] I have tested these changes.
* [x] I have verified that this is not duplicating changes in an existing PR.
* [x] I have provided a *substantive* review of [another web PR](https://github.com/jellyfin/jellyfin-web/pulls?q=is%3Apr+is%3Aopen+-label%3Adependencies+-label%3Ablocked+-label%3Abackend+-label%3Ainvalid+-label%3A"merge+conflict"+-label%3Astale+-author%3A%40me).

PR #8305, opened 2026-08-01
Revert "Replace libpgs with libbitsub and adds support for vobsub rendering"

nyanmisaka:
Reverts jellyfin/jellyfin-web#7777

This breaks the progress bar when **resuming** playback with HLS transcoding.

w/ #7777

w/o #7777


nyanmisaka:
@altqx Could you take a look at this? We are very close to the 12.0 RC4 release. If this issue persists, we will have to revert it for the time being.

altqx:
@nyanmisaka Upon checking, it's look like I accidentally changed HLS resume offset handling, causing the resume position to be counted twice. Sorry for my sloppy code. 

I will revert back the HLS offset with some update to the libbitsub worker prewarm in v1.11.0 in the new PR soon.

altqx:
The new PR should now up at https://github.com/jellyfin/jellyfin-web/pull/8308

nyanmisaka:
> The new PR should now up at #8308

Enabling graphical subtitles triggers transcoding when remuxing both video and audio (due to unsupported secondary audio stream). The option `Always burn in subtitle when transcoding` is not enabled. Can you reproduce this on your end?

nyanmisaka:
I traced it and found that these lines in `CanStreamCopyVideo()` caused the transcoding.
https://github.com/jellyfin/jellyfin/blob/master/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs#L2422-L2428

altqx:
when `AlwaysBurnInSubtitleWhenTranscoding=false` and on-the-fly extraction is disabled, the planner forces `SubtitleMethod=Encode` 

This probably an oversight. fixing 

Secondary audio forces an HLS remux, and with `EnableSubtitleExtraction=false`, `StreamBuilder.cs:1585` rejects external delivery for embedded PGS/VobSub. This pushes the request to `SubtitleMethod=Encode&SubtitleStreamIndex=3` despite `AlwaysBurnInSubtitleWhenTranscoding=false`, which correctly triggers `EncodingHelper.cs` to disable video copy since burning subtitles requires a transcode.

altqx:
> > The new PR should now up at #8308
> 
> Enabling graphical subtitles triggers transcoding when remuxing both video and audio (due to unsupported secondary audio stream). The option `Always burn in subtitle when transcoding` is not enabled. Can you reproduce this on your end?

The fix is now live at https://github.com/jellyfin/jellyfin/pull/17512

PR #8308, opened 2026-08-01
fix hls offset and update libbitsub to v1.11.0

altqx:
Fix https://github.com/jellyfin/jellyfin-web/pull/8305
And bump libbitsub to v1.11.0  with upgrate to worker pre-warm and subtitle streaming load, range requests, and a prefetch.

---


nyanmisaka (approved):
It works for me.

src/plugins/htmlVideoPlayer/plugin.js
```diff
@@ -504,6 +507,10 @@ export class HtmlVideoPlayer {
 
         if (options.resetSubtitleOffset !== false) this.resetSubtitleOffset();
 
+        void import('libbitsub')
```
thornbill: Shouldn't this be conditional on the user setting being enabled?

thornbill:
ES5 support seems to be broken again 🙃

altqx:
> ES5 support seems to be broken again 🙃

The fix should now in! i forgot that v1.11.0 update the module-worker glue asset logic.

thornbill (approved):

PR #6821, opened 2025-04-30
Use enum for app features

thornbill:
**Changes**
* Adds an enum for app feature values
* Removes unused feature values
* Fixes `dashboard.js` not checking `appHost` for multiserver support

**Issues**
N/A

PR #8314, opened 2026-08-01
Fix native shell selectServer not used when signing out

nielsvanvelzen:
### Changes

The Android app (and iOS probably too?) implements the selectServer function in the native shell to show native UI for server selection. Due to an oversight the sign out button did not use this, instead it would always redirect to the selectserver route if the multi-server feature is enabled.

This change fixes that, making sure the native shell is used for server selection.

### Code assistance

---

### DIFF
--- a/src/utils/dashboard.js
+++ b/src/utils/dashboard.js
@@ -107,8 +107,12 @@ export function logout() {
         queryClient.clear();
         // Reset cached views
         viewContainer.reset();
-        appHost.supports(AppFeature.MultiServer) ?
-            navigate('selectserver') : navigate('login');
+
+        if (appHost.supports(AppFeature.MultiServer)) {
+            selectServer();
+        } else {
+            navigate('login');
+        }
     });
 }
 


dmitrylyzo:
It shouldn't select a server when logging out unless webui is configured as a multi-server (config.json). At least that's how it worked before https://github.com/jellyfin/jellyfin-web/pull/6821

nielsvanvelzen:
The multi-server flag is needed to show the "Switch server" button. The only thing this PR does it make sure the nativeshell is invoked when it implements the server switching natively.

dmitrylyzo:
> The multi-server flag is needed to show the "Switch server" button.

But this doesn't mean that we should navigate to the server selection.

> The only thing this PR does it make sure the nativeshell is invoked when it implements the server switching natively.

In theory, when you connect to a hosted webui that is configured as multi-server, it shouldn't use `NativeShell.selectServer`. To change the server you should use `Switch server`.

I believe that if you revert `webSettings.getMultiServer()`, it will fix the problem with navigating to the server selection on logout.

nielsvanvelzen:
I'm not here to change the UX. I'm just making sure the Android app works as intended, and the webui won't show for server switching when there is a native shell implementation. This is a bug both in 10.11 and 12.0.

nielsvanvelzen:
Effectively the only thing this change does is replace this call:

```js
navigate('selectserver')
```

with

```js
if (window.NativeShell && typeof window.NativeShell.selectServer === 'function') {
    window.NativeShell.selectServer();
} else {
    navigate('selectserver');
}
```

dmitrylyzo:
Try reverting the corresponding part of https://github.com/jellyfin/jellyfin-web/pull/6821

When you logging out, you should go to the "Login" page, not to the "Select server".

nielsvanvelzen:
The exact same behavior, including the skipping of nativeshell, already existed before that PR.

```
webSettings.getMultiServer().then(multi => {
    multi ? navigate('selectserver') : navigate('login');
});
```

So even then this bug would've existed.

BotBlake:
I am so sorry I pressed the dumbest button that I could have possibly pressed

dmitrylyzo:
> So even then this bug would've existed.

It was not a bug, because that is how multi-server webui worked, iirc.

When you connect to the multi-server webui, it should show you the configured servers ("Select server"). "Logout" should go to the "Select server", because it doesn't use/have "Switch server" feature.

As I said, the problem probably started after https://github.com/jellyfin/jellyfin-web/pull/6821. So, if webui is configured as a single-server (most cases), you should go to the "login". If it is multi-server - to "selectserver".

nielsvanvelzen:
The Android app is always multiserver, because you can change the server you're connected to. And even if we ignore whether the app requests the multiserver feature or not. If a user configured it in their config.json the bug would still appear.

We **must** always use the native shell for server selection if the native shell claims to implement that functionality and **never** open the web route.

I'm not sure how to explain it any easier for you.

dmitrylyzo:
> The Android app is always multiserver, because you can change the server you're connected to.

Yes, but webui you are connecting to is not (in most cases) - you don't configure the media server after connecting, because that's where the webui is hosted.

Multi-server webui is probably designed for standalone use and shouldn't be enabled on server+webui.

> If a user configured it in their config.json the bug would still appear.

This would probably be a misuse. You shouldn't connect to a multi-server webui, as you will not have access to pre-configured servers.

> We **must** always use the native shell for server selection if the native shell claims to implement that functionality and **never** open the web route.

As for the current block, we can probably remove the navigation to selectserver, because we want to log out (switch user) and we have a dedicated "switch server". But since you don't want to change the UX, and I'm not sure why it was implemented this way, I suggest to revert Bill's changes. In both cases, the user will be switched to the login screen, rather than server selection (native or single-server webui).

dmitrylyzo:
Tracked down the author of the "Select server on logout" logic: https://github.com/jellyfin/jellyfin-web/pull/13

History:
- https://github.com/jellyfin/jellyfin-web/pull/13 Apps go to "Select server" after logout. _10.2+_
- https://github.com/jellyfin/jellyfin-web/pull/1918 Apps go to "Login" after logout unless webui is in multi-server mode. _10.7+_
- https://github.com/jellyfin/jellyfin-web/pull/6821 Apps go to "Select server" after logout (again). _10.11+_

Now I'm more convinced that we need to remove the "Select server on logout" logic, because I see no reason for this.

If someone wants to connect to a multi-server webui (i.e. standalone webui) using an app, then we will have to add another "switch server" to be able to configure and select servers within the current hosted webui. But as I said, standalone webui shouldn't be used inside the app.

thornbill (approved):
I agree we should probably just remove the select server case here, but let's merge this for now since it's actively breaking the Android app

PR #8315, opened 2026-08-02
Add setting to include episodes in progress in Next Up

slevin-7:
### Changes

The Next Up home section calls the API with a hardcoded `enableResumable: false`, and no user setting exists for it, so a series drops out of Next Up entirely while one of its episodes is only partially watched. Users who disable the Continue Watching home section then lose those in-progress episodes with no way to get them back. This adds an "Enable Episodes in Progress in Next Up" display setting, following the existing rewatching option (legacy + modern settings UI, home section and Next Up list view). Defaults to off, so nothing changes unless opted in.

### Issues

None filed.

### Code assistance

LLM-assisted: locating where the flag is consumed and drafting the patch, modeled on the existing `enableRewatchingInNextUp` implementation. I reviewed the changes and verified the behavior on my own server (12.0); `tsc --noEmit` and eslint pass.

---

### DIFF
--- a/src/apps/legacy/controllers/list.js
+++ b/src/apps/legacy/controllers/list.js
@@ -261,6 +261,7 @@ function getItems(instance, params, item, sortBy, startIndex, limit) {
             EnableImageTypes: 'Primary,Backdrop,Thumb',
             EnableTotalRecordCount: false,
             SortBy: sortBy,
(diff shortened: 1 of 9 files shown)


thornbill:
Did not complete the pr template.

PR #8316, opened 2026-08-02
Add combined Continue Watching & Next Up home section

slevin-7:
**Changes**

Adds a new opt-in home section "Continue Watching & Next Up" that merges the resume list and Next Up into a single row: in-progress movies and episodes first (most recently played), followed by each series' next unwatched episode. The trick that keeps it simple: requesting Next Up with `enableResumable: false` makes the two lists exactly complementary — a series whose next episode is partially watched surfaces in the resume part instead — so the section just concatenates two parallel queries and never needs a dedup pass. Selectable in the home screen section settings; defaults are unchanged.

**Code assistance**

LLM-assisted (mapping the section registry/settings plumbing and drafting the section module, modeled on the existing resume/nextUp sections). I reviewed the changes and verified the behavior on my own server (12.0 preview, 175k-item library): the merged row renders movies in progress, episodes in progress and upcoming episodes together, with no duplicate items; `tsc --noEmit`, eslint and a production webpack build pass.

**Issues**

None filed — happy to open a feature request first if preferred.

The section module mirrors the existing resume/nextUp sections; happy to extract a shared section-shell helper (also covering those two) in a follow-up if preferred.

### DIFF
--- a/src/components/homeScreenSettings/homeScreenSettings.template.html
+++ b/src/components/homeScreenSettings/homeScreenSettings.template.html
@@ -23,6 +23,7 @@
                 <option value="librarybuttons">${HeaderMyMediaSmall}</option>
                 <option value="activerecordings">${HeaderActiveRecordings}</option>
                 <option value="resume">${HeaderContinueWatching}</option>
+                <option value="continuewatching">${HeaderContinueWatchingAndNextUp}</option>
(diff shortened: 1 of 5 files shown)


thornbill:
Did not complete the pr template.
Home screen sections must be universally supported.

PR #8320, opened 2026-08-04
unify progress indicator across book players

dkanada:
### Changes

Previously, the comic player used a Swiper feature to show current position from local state. This pull request removes that implementation and replaces it with a ProgressIndicator element in BookOsd that hooks into playback events for all updates.

@jordanmichaelrushing I technically "fixed" the audio + book playback issue along with the photo playback issue from your previous review comments. However, I consider the former an unintentional feature and have removed it to avoid edge cases. Simultaneous playback in a single browser tab doesn't seem like something we want to support but I'll defer to @thornbill for the decision.

### Issues

None

### Code Assistance

None

### DIFF
--- a/src/plugins/comicsPlayer/plugin.js
+++ b/src/plugins/comicsPlayer/plugin.js
@@ -221,7 +221,6 @@ export class ComicsPlayer {
             elem.classList.add('slideshowDialog');
             elem.innerHTML = `<div id="bookOsdMount"></div><div dir=${this.comicsPlayerSettings.langDir} class="slideshowSwiperContainer">
                                 <div class="swiper-wrapper"></div>
-                                <div class="swiper-pagination"></div>
                             </div>`;
 
             dialogHelper.open(elem);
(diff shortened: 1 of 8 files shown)


jesse-anderson:
Hello,

Took a look and ran this against a pdf, epub, and cbz. Progress bar works great on my end.

One thing I ran into: paging through with the keyboard, the OSD hides after the 2s timer and can't be brought back, so the new progress bar goes with it. The keyboard gap isn't from this PR, it's already on master, but I think a progress bar you can actually see while paging by keyboard is worth having.

Separate from that, on the cbz the swiper fraction stays visible after the OSD hides on current master, but it's gone on this branch(replaced by progress bar I assume), so by keyboard comics position goes from always visible to not visible at all.

Full disclosure, I have #8478 open against BookOsd making the OSD respond to keydown, which would cover the keyboard part of this.

Jesse

src/plugins/comicsPlayer/plugin.js
```diff
@@ -221,7 +221,6 @@ export class ComicsPlayer {
             elem.classList.add('slideshowDialog');
             elem.innerHTML = `<div id="bookOsdMount"></div><div dir=${this.comicsPlayerSettings.langDir} class="slideshowSwiperContainer">
                                 <div class="swiper-wrapper"></div>
-                                <div class="swiper-pagination"></div>
```
jesse-anderson, line 224: This sits outside the OSD, so on master the fraction stays visible after the OSD auto hides. With it gone, comics position only shows while the OSD is up, and when you're paging using the keyboard, can't be brought back. Checked on a cbz against master and this build.

jesse-anderson (commented):
Same point as my comment above, anchored to the lines it comes.

Issue #3300, opened 2022-01-06
Download Action on "Additional Parts" list container in mobile web view


biojet1:
In Mobile web view there is no way you can download videos inside the "**Additional Parts**" list container.
I see only **Play**, **Watched**, **Favorite** buttons.
I can't access the context menu, where you can click the **Download** button.

Where in the source I can add that functionality?

PR #8328, opened 2026-08-05
default overlayMoreButton for mobile cards

BotBlake:
### Changes

Users have requested both the Download on `Additional Parts` and on `special features`.
This PR changes the default card options for `getVideosHtml` to set enable `overlayMoreButton`.
How I see  it, the cardBuilder exclusively uses `options.overlayMoreButton` for mobile layout, so setting it to true will not affect TV or desktop layouts.

I decided to directly modify the `getVideosHtml` instead of only the card options for both `specialsContent` and `additionalPartsContent`, necause `getVideosHtml` seems to be only used 3 times:
- `specialsContent`
- `additionalPartsContent`
- `musicVideosContent`

In my opinion it is sensible to just modify the default, especially since I see no reason not to allow downloading of Music Videos on mobile too.

I am aware this DOES remove the `overlayPlayButton` from those types of mini-cards, however since it is not needed this should be fine. If the UX now is "too akward", I guess enabling `centerPlayButton` would be a ?good? alternative?

Anyways...

| Before | After |
| - | - |
| grafik | grafik |

### Issues

Should fix:
- [Download Action on "Additional Parts" list container in mobile web view](https://github.com/jellyfin/jellyfin-web/issues/3300)
- [download of special features](https://github.com/jellyfin/jellyfin-android/issues/636)

### Code assistance

ChatGPT was used to find alternative solutions. However it just suggested garbage and its output was not used.

---

* [X] I have verified that this is not duplicating changes in an existing PR. (At least couldn't find one)
* [ ] I have provided a *substantive* review of [another web PR](https://github.com/jellyfin/jellyfin-web/pulls?q=is%3Apr+is%3Aopen+-label%3Adependencies+-label%3Ablocked+-label%3Abackend+-label%3Ainvalid+-label%3A"merge+conflict"+-label%3Astale+-author%3A%40me). (I wish I could :moyai: )

### DIFF
--- a/src/apps/legacy/controllers/itemDetails/index.js
+++ b/src/apps/legacy/controllers/itemDetails/index.js
@@ -1830,7 +1830,8 @@ function getVideosHtml(items) {
         action: 'play',
         overlayText: false,
         centerText: true,
-        showRuntime: true
+        showRuntime: true,
+        overlayMoreButton: true
     });
 }
 


erikbucik:
No, you're not adding a menu button over the thumbnail. 

1. No other card has it, 
2. It's very small, and 
3. The download action is missing here:

Screenshot_20260805-125350

BotBlake:
@erikbucik I think there are a few misunderstandings here.

> 1. No other card has it

`overlayMoreButton` is an existing card-builder option and is already used by multiple card views. The button is only rendered when `layoutManager.mobile` is true, so this does not introduce a new desktop or TV interaction pattern.

This PR enables that existing mobile card control for the three places that use `getVideosHtml`:

- Special features
- Additional parts
- Music videos

So this is not a one-off button being added to a single card type.

> 2. It’s very small

The button uses the existing `cardOverlayButton` class... the same class and sizing used by the play button that was previously displayed in that position. This PR does not introduce a new button size or custom styling.

> 3. The download action is missing here

The menu shown in your screenshot is a different menu. It is not the menu opened by the overlay button added by this PR.



The underlying problem this PR addresses is that users currently cannot access the download action for special features or additional parts in the mobile web layout (and therefore the Android and iOS Mobile Apps). Even if the current placement is considered visually imperfect, not having this button leaves that functionality completely inaccessible.

I’m open to using a different established UI pattern, but it needs to preserve access to the item actions on mobile. Could you clarify which existing component or placement you would prefer for these cards?

There is also a deployed preview attached to the PR, so the actual behavior and menu can be tested directly...

thornbill (approved):
For better or worse this is consistent with other cards in the UI and unblocks functionality for media that currently does not open in a dedicated view (except for maybe music videos).

PR #8329, opened 2026-08-05
add security improvements to npmrc

dkanada:
### Changes

allow-remote: Setting "none" would be better but we have a single remote package. The default on new NPM versions is none so this also fixes installs in modern environments.

ignore-scripts: I ran a search and I don't think any of our dependencies use irregular scripts?

min-release-age: Requires that a package exist for at least two weeks before it can be added to our builds by default.

### Issues

None

### Code Assistance

None

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,3 +1,7 @@
+allow-remote=root
 engine-strict=true
 fund=false
+ignore-scripts=true
+min-release-age=7
+min-release-age-exclude[]=@jellyfin/*
 save-exact=true


thornbill:
`min-release-age`: we use 7 days for renovate and the `@jellyfin` namespace should be excluded

.npmrc
```diff
@@ -1,3 +1,6 @@
 engine-strict=true
 fund=false
+allow-remote=root
+ignore-scripts=true
+min-release-age=14
 save-exact=true
```
dmitrylyzo, line 7: It looks like they were in alphabetical order. Maybe sort them?

dkanada:
@thornbill exclusions are a newer feature but I can't find the *exact* version they were introduced. I'd still like to ship that flag but will remove it if we absolutely need to have the exclusion. Even without it, a manual override followed by a lockfile install should still work.

```
npm install --min-release-age 0 @jellyfin/sdk
npm ci
```

thornbill (approved):

PR #8332, opened 2026-08-05
add WebKitGTK browser detection and device profile

IGS-GIT:
### Changes

- Add WebKitGTK User-Agent detection in browser.js and export epiphany in browser.d.ts.
- Scope browser.safari strictly to Apple platforms to prevent non-Apple WebKit engines on Linux from being misidentified as Safari.
- Add webkitGtk to BrowserName in apphost.js for accurate device name display in the UI.
- Update browserDeviceProfile.js to support 10-bit HEVC (Main 10, Level 6.2), 6-channel surround audio, and fMP4 HLS streaming for WebKitGTK.
- Disable direct MKV progressive HTTP streaming for Linux WebKit in testCanPlayMkv() to trigger on-the-fly container remuxing and prevent stream buffering hangs.
- Set preferFmp4HlsContainer to enabled by default for WebKitGTK in userSettings.js.
- Add unit tests for WebKitGTK UA detection in browser.test.ts.

### Images

Before:
Screenshot From 2026-08-09 20-44-39

After:
Screenshot From 2026-08-09 20-52-40

---

* [ x] I have read and followed the [contributing guidelines](https://github.com/jellyfin/jellyfin-web/blob/master/CONTRIBUTING.md).
* [ x] I have tested these changes.
* [ x I have verified that this is not duplicating changes in an existing PR.
* [ x] I have provided a *substantive* review of [another web PR](https://github.com/jellyfin/jellyfin-web/pulls?q=is%3Apr+is%3Aopen+-label%3Adependencies+-label%3Ablocked+-label%3Abackend+-label%3Ainvalid+-label%3A"merge+conflict"+-label%3Astale+-author%3A%40me).

### DIFF
--- a/src/scripts/browser.js
+++ b/src/scripts/browser.js
@@ -201,7 +201,7 @@ const uaMatch = function (ua) {
         || /(opera)[ /]([\w.]+)/.exec(ua)
         || /(opr)[ /]([\w.]+)/.exec(ua)
         || /(chrome)[ /]([\w.]+)/.exec(ua)
-        || /(safari)[ /]([\w.]+)/.exec(ua)
+        || /(applewebkit|webkit)[ /]([\w.]+)/.exec(ua)
         || /(firefox)[ /]([\w.]+)/.exec(ua)
         || !ua.includes('compatible') && /(mozilla)(?:.*? rv:([\w.]+)|)/.exec(ua)
         || [];
@@ -223,6 +223,10 @@ const uaMatch = function (ua) {
 
     if (browser === 'opr') {
         browser = 'opera';
+    }
+
+    if (browser === 'applewebkit') {
+        browser = 'webkit';
     }
 
     let version;
@@ -264,10 +268,6 @@ export const detectBrowser = (userAgent = navigator.userAgent) => {
 
     browser.edgeChromium = browser.edg || browser.edga || browser.edgios;
 
-    if (!browser.chrome && !browser.edgeChromium && !browser.edge && !browser.opera && normalizedUA.includes('webkit')) {
-        browser.safari = true;
-    }
-
     browser.osx = normalizedUA.includes('mac os x');
 
     // This is a workaround to detect iPads on iOS 13+ that report as desktop Safari
@@ -295,6 +295,17 @@ export const detectBrowser = (userAgent = navigator.userAgent) => {
     browser.operaTv = browser.tv && normalizedUA.includes('opr/');
 
     browser.edgeUwp = (browser.edge || browser.edgeChromium) && (normalizedUA.includes('msapphost') || normalizedUA.includes('webview'));
+
+    const isApplePlatform = browser.osx || browser.iphone || browser.ipad || browser.ipod || normalizedUA.includes('mac os x');
+    const isLinux = normalizedUA.includes('linux') && !browser.android && !browser.tizen && !browser.web0s;
+
+    if (browser.webkit) {
+        if (isApplePlatform) {
+            browser.safari = true;
+        } else if (isLinux && !browser.tv && !browser.titanos && !browser.vidaa && !browser.hisense) {
+            browser.webkitGtk = true;
+        }
+    }
 
     if (browser.web0s) {
         browser.web0sVersion = web0sVersion(browser);
--- a/src/scripts/browser.js
+++ b/src/scripts/browser.js
@@ -201,6 +201,7 @@ const uaMatch = function (ua) {
         || /(opera)[ /]([\w.]+)/.exec(ua)
         || /(opr)[ /]([\w.]+)/.exec(ua)
         || /(chrome)[ /]([\w.]+)/.exec(ua)
+        || /(epiphany)[ /]([\w.]+)/.exec(ua)
         || /(safari)[ /]([\w.]+)/.exec(ua)
         || /(firefox)[ /]([\w.]+)/.exec(ua)
         || !ua.includes('compatible') && /(mozilla)(?:.*? rv:([\w.]+)|)/.exec(ua)
@@ -264,10 +265,6 @@ export const detectBrowser = (userAgent = navigator.userAgent) => {
 
     browser.edgeChromium = browser.edg || browser.edga || browser.edgios;
 
-    if (!browser.chrome && !browser.edgeChromium && !browser.edge && !browser.opera && normalizedUA.includes('webkit')) {
-        browser.safari = true;
-    }
-
     browser.osx = normalizedUA.includes('mac os x');
 
     // This is a workaround to detect iPads on iOS 13+ that report as desktop Safari
@@ -275,6 +272,15 @@ export const detectBrowser = (userAgent = navigator.userAgent) => {
     // https://forums.developer.apple.com/thread/119186
     if (browser.osx && !browser.iphone && !browser.ipod && !browser.ipad && navigator.maxTouchPoints > 1) {
         browser.ipad = true;
+    }
+
+    const isApplePlatform = browser.osx || browser.iphone || browser.ipad || browser.ipod || normalizedUA.includes('mac os x');
+    if (!browser.chrome && !browser.edgeChromium && !browser.edge && !browser.opera && isApplePlatform && normalizedUA.includes('webkit')) {
+        browser.safari = true;
+    }
+
+    if (browser.safari && !isApplePlatform) {
+        delete browser.safari;
     }
 
     if (isMobile(normalizedUA)) {
@@ -290,6 +296,7 @@ export const detectBrowser = (userAgent = navigator.userAgent) => {
     browser.vega = normalizedUA.includes('kepler');
     browser.vidaa = normalizedUA.includes('vidaa');
     browser.web0s = isWeb0s(normalizedUA);
+    browser.epiphany = normalizedUA.includes('epiphany');
 
     browser.tv = browser.ps4 || browser.vega || browser.xboxOne || isTv(normalizedUA);
     browser.operaTv = browser.tv && normalizedUA.includes('opr/');
@@ -319,6 +326,9 @@ export const detectBrowser = (userAgent = navigator.userAgent) => {
         delete browser.safari;
         // UserAgent string contains 'Mobile Chrome', but it is a TV
         delete browser.mobile;
+    } else if (browser.epiphany) {
+        // UserAgent string contains 'Safari', but we only want 'epiphany' to be true
+        delete browser.safari;
     } else {
         browser.orsay = normalizedUA.includes('smarthub');
     }
--- a/src/scripts/browser.test.ts
+++ b/src/scripts/browser.test.ts
@@ -35,4 +35,15 @@ describe('Browser', () => {
         expect(browser.xboxOne).toBe(true);
         expect(browser.tv).toBe(true);
     });
+
+    it('should identify Epiphany browsers', () => {
+        const browser = detectBrowser('Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.0 Epiphany/45.0 Safari/605.1.15');
+        expect(browser.epiphany).toBe(true);
+        expect(browser.safari).toBeFalsy();
+    });
+
+    it('should not identify generic Linux WebKit as Safari', () => {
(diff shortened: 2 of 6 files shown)


src/scripts/settings/userSettings.js
```diff
@@ -181,7 +181,7 @@ export class UserSettings {
         }
 
         // Enable it by default only for the platforms that play fMP4 for sure.
-        return toBoolean(this.get('preferFmp4HlsContainer', false), browser.safari || browser.firefox || browser.chrome || browser.edgeChromium);
+        return toBoolean(this.get('preferFmp4HlsContainer', false), browser.safari || browser.firefox || browser.chrome || browser.edgeChromium || browser.epiphany || !browser.mobile);
```
dmitrylyzo: You are enabling fMP4 on any non-mobile platform (`|| !browser.mobile`).

src/scripts/browserDeviceProfile.js
```diff
@@ -360,7 +366,7 @@ function getDirectPlayProfileForVideoContainer(container, videoAudioCodecs, vide
             supported = browser.tizen;
             break;
         case 'mov':
-            supported = browser.safari || browser.tizen || browser.web0s || browser.chrome || browser.edgeChromium || browser.edgeUwp;
+            supported = browser.safari || browser.tizen || browser.web0s || browser.chrome || browser.edgeChromium || browser.edgeUwp || browser.epiphany || !browser.mobile;
```
dmitrylyzo: You are enabling MOV on any non-mobile platform - `|| !browser.mobile`.

src/scripts/browserDeviceProfile.js
```diff
@@ -457,7 +463,7 @@ function getPhysicalAudioChannels(options, videoTestElement) {
         return options.audioChannels;
     }
 
-    const isSurroundSoundSupportedBrowser = browser.safari || browser.chrome || browser.edgeChromium || browser.firefox || browser.tv || browser.ps4 || browser.xboxOne;
+    const isSurroundSoundSupportedBrowser = browser.safari || browser.chrome || browser.edgeChromium || browser.firefox || browser.epiphany || browser.tv || browser.ps4 || browser.xboxOne || !browser.mobile;
```
dmitrylyzo: You are enabling Surround sound on any non-mobile platform - `|| !browser.mobile`.

src/scripts/browserDeviceProfile.js
```diff
@@ -683,7 +689,7 @@ export default function (options) {
     }
 
     if (canPlayHevc(videoTestElement, options)
-        && (browser.edgeChromium || browser.safari || browser.tizen || browser.web0s || (browser.chrome && (!browser.android || browser.versionMajor >= 105)) || (browser.opera && !browser.mobile) || (browser.firefox && browser.versionMajor >= 134))) {
+        && (browser.edgeChromium || browser.safari || browser.epiphany || browser.tizen || browser.web0s || (browser.chrome && (!browser.android || browser.versionMajor >= 105)) || (browser.opera && !browser.mobile) || (browser.firefox && browser.versionMajor >= 134) || !browser.mobile)) {
```
dmitrylyzo: Again, `|| !browser.mobile` simply ignores all previous constraints on non-mobile platform. Why did you add it?

src/scripts/browserDeviceProfile.js
```diff
@@ -1150,6 +1156,11 @@ export default function (options) {
     let maxHevcLevel = 120;
     let hevcProfiles = 'main';
 
+    if (canPlayHevc(videoTestElement, options)) {
+        maxHevcLevel = 186;
+        hevcProfiles = 'main|main 10';
+    }
```
dmitrylyzo: You should check if level 186 is supported, not just HEVC.

Keep the tests in ascending order.

I think the HEVC level upgrade should be moved to a separate PR because it isn't tied to Epiphany browser.

src/scripts/browser.js
```diff
@@ -319,6 +326,9 @@ export const detectBrowser = (userAgent = navigator.userAgent) => {
         delete browser.safari;
         // UserAgent string contains 'Mobile Chrome', but it is a TV
         delete browser.mobile;
+    } else if (browser.epiphany) {
+        // UserAgent string contains 'Safari', but we only want 'epiphany' to be true
+        delete browser.safari;
```
dmitrylyzo: Don't all browsers on Apple platform use the Safari engine?

We probably shouldn't delete `browser.safari` just because it's also Epiphany.

Moreover, the match for Epiphany goes before Safari: https://github.com/IGS-GIT/jellyfin-web/blob/9681219d04a2b0eba7d2802f7f0e700e655be7fa/src/scripts/browser.js#L203-L205
So I think we can remove this block.

src/scripts/browser.js
```diff
@@ -290,6 +296,7 @@ export const detectBrowser = (userAgent = navigator.userAgent) => {
     browser.vega = normalizedUA.includes('kepler');
     browser.vidaa = normalizedUA.includes('vidaa');
     browser.web0s = isWeb0s(normalizedUA);
+    browser.epiphany = normalizedUA.includes('epiphany');
```
dmitrylyzo: I would keep the alphabetical order.

src/scripts/browser.js
```diff
@@ -277,6 +274,15 @@ export const detectBrowser = (userAgent = navigator.userAgent) => {
...
     }
 
+    const isApplePlatform = browser.osx || browser.iphone || browser.ipad || browser.ipod || normalizedUA.includes('mac os x');
+    if (!browser.chrome && !browser.edgeChromium && !browser.edge && !browser.opera && isApplePlatform && normalizedUA.includes('webkit')) {
+        browser.safari = true;
+    }
+
+    if (browser.safari && !isApplePlatform) {
+        delete browser.safari;
+    }
```
dmitrylyzo: Maybe rewrite it like this?
```suggestion
    const isApplePlatform = browser.osx || browser.iphone || browser.ipad || browser.ipod || normalizedUA.includes('mac os x');

    if (isApplePlatform) {
        if (!browser.chrome && !browser.edgeChromium && !browser.edge && !browser.opera && normalizedUA.includes('webkit')) {
            browser.safari = true;
        }
    } else {
        // No Safari on non-Apple platform
        delete browser.safari;
    }
```
Technically, this will remove `browser.safari` from webOS, Tizen, Vega, etc.

src/scripts/browser.test.ts
```diff
@@ -35,4 +35,15 @@ describe('Browser', () => {
...
     });
+
+    it('should identify Epiphany browsers', () => {
+        const browser = detectBrowser('Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.0 Epiphany/45.0 Safari/605.1.15');
+        expect(browser.epiphany).toBe(true);
+        expect(browser.safari).toBeFalsy();
+    });
+
+    it('should not identify generic Linux WebKit as Safari', () => {
+        const browser = detectBrowser('Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/60.5 Safari/605.1.15');
```
dmitrylyzo: Such a browser will become unidentified. I don't know how to handle this. 🤷

src/scripts/browserDeviceProfile.js
```diff
@@ -20,7 +20,10 @@ function canPlayHevc(videoTestElement, options) {
         && (videoTestElement.canPlayType('video/mp4; codecs="hvc1.1.L120"').replace(/no/, '')
         || videoTestElement.canPlayType('video/mp4; codecs="hev1.1.L120"').replace(/no/, '')
         || videoTestElement.canPlayType('video/mp4; codecs="hvc1.1.0.L120"').replace(/no/, '')
-        || videoTestElement.canPlayType('video/mp4; codecs="hev1.1.0.L120"').replace(/no/, ''));
+        || videoTestElement.canPlayType('video/mp4; codecs="hev1.1.0.L120"').replace(/no/, '')
+        || videoTestElement.canPlayType('video/mp4; codecs="hevc"').replace(/no/, '')
+        || videoTestElement.canPlayType('video/mp4; codecs="hvc1"').replace(/no/, '')
+        || videoTestElement.canPlayType('video/mp4; codecs="hev1"').replace(/no/, ''));
```
dmitrylyzo: Does this work? Shouldn't the codec be fully specified?

dmitrylyzo:
In addition, you haven't fully completed the PR template.

IGS-GIT:
> In addition, you haven't fully completed the PR template.

Sorry about that. I removed sections that I thought weren't applicable to my PR. I haven't had much time to review your feedback but I'll try to get it all cleared up tonight.

src/scripts/browserDeviceProfile.js
IGS-GIT: Sorry. It seemed okay at the time since I wasn't testing for mobile. I've removed all of these in the latest amendment

src/scripts/browserDeviceProfile.js
IGS-GIT: Sorry. I'll omit this. I'll save it for a future commit

src/scripts/browser.js
IGS-GIT: This was before I added the uaMatch. Forgot to delete this block. Sorry

src/scripts/browser.js
IGS-GIT: Noted. This is much cleaner. Thanks

src/scripts/browser.test.ts
IGS-GIT: This was intentional. Before, any non-Apple Webkit browser would be flagged as Safari, and since they don't have Apple's proprietary codecs, they would fallback to transcoding, or sometimes not load anything at all. This aimed to fix it, but now in hindsight, there are other browsers that would get fallbacked for no reason. I'm working on a rework of this PR to broaden its scope to WebkitGTK as a whole, not just Epiphany, since all browsers that use it also use the same codec stack, GStreamer.

src/scripts/browserDeviceProfile.js
IGS-GIT: Fair point. I'll remove these queries. Sorry

sjorge:
Would it not make more sense in src/scripts/browser.js

To first detect the browser as webkit for all webkit based ones and the further narrow it down to safari if needed?

Both seem to include "AppleWebKit/<VERSION>" in there UA string, this (https://github.com/jellyfin/jellyfin-web/pull/8332/changes#diff-378b0a54e4583efb41b757de74042838c767d39412921ff381fe1c59bfc74cd2R229) can then be dropped, and the ua matcher here (https://github.com/jellyfin/jellyfin-web/pull/8332/changes#diff-378b0a54e4583efb41b757de74042838c767d39412921ff381fe1c59bfc74cd2R204-R205) can just match on webkit.

This (https://github.com/jellyfin/jellyfin-web/pull/8332/changes#diff-378b0a54e4583efb41b757de74042838c767d39412921ff381fe1c59bfc74cd2R283-R290) can then just set browser.safari if it matches (no need to delete as it was never create in the first place).

Note: I am not super familiar with browser.js so I might have missed something obvious why this won't work.

IGS-GIT:
> Would it not make more sense in src/scripts/browser.js
> 
> To first detect the browser as webkit for all webkit based ones and the further narrow it down to safari if needed?
> 
> Both seem to include "AppleWebKit/" in there UA string, this (https://github.com/jellyfin/jellyfin-web/pull/8332/changes#diff-378b0a54e4583efb41b757de74042838c767d39412921ff381fe1c59bfc74cd2R229) can then be dropped, and the ua matcher here (https://github.com/jellyfin/jellyfin-web/pull/8332/changes#diff-378b0a54e4583efb41b757de74042838c767d39412921ff381fe1c59bfc74cd2R204-R205) can just match on webkit.
> 
> This (https://github.com/jellyfin/jellyfin-web/pull/8332/changes#diff-378b0a54e4583efb41b757de74042838c767d39412921ff381fe1c59bfc74cd2R283-R290) can then just set browser.safari if it matches (no need to delete as it was never create in the first place).
> 
> Note: I am not super familiar with browser.js so I might have missed something obvious why this won't work.

Fair point. I cleaned up and updated the detection logic. Let me know if it looks okay

IGS-GIT:
Any chance for a re-review? Let me know if there are any issues with the request

PR #8334, opened 2026-08-07
Scale remaining time by playback speed

reedstrm:
### Changes

Scaled the time remaining (negative number at the end of the timebar) by the current playback speed, so that it indicates
the actual time left to play. The clock time "Ends at" time was already scaled. To make it clear, I added a parenthetical scaling number as well. I purposefully did not scale the position in video or scrubbing numbers: those still show the actual  video stream time point, so references to specific time codes are still correct.

image

### Code assistance

Codex was used to quickly find the location in the code to change. I wrote the code, following existing variable naming conventions and code style. 

---


src/apps/legacy/controllers/playback/video/index.js
```diff
@@ -847,10 +849,10 @@ export default function (view) {
             }
 
             if (userSettings.enableVideoRemainingTime()) {
```
slickwarren, line 851:
```suggestion
            if (userSettings.enableVideoRemainingTime()) {
                const playbackRateText = playbackRate != null && Number(playbackRate) !== 1 ? ' (' + playbackRate + 'x)' : '';

```

this will only compute the const when it will actually be used

slickwarren (commented):
just suggesting to move the first const into the if block that it is used in, since this const isn't used outside of it.

PR #8338, opened 2026-08-08
Open external password reset link on different provider

### LINKED ISSUE #22: Override the "forgot password" functionality?

viown:
In cases when using an external provider (like LDAP) for password resets, the "PinFile" can be a url to an external password reset link. To avoid incorrectly asking for a pin, this PR allows instead to redirect to that provider and then return to the login page.

Ideally this should be sent as a separate action from the server.

### Changes
Redirect to external password reset provider when it's detected as a URL

### Issues
Fixes https://github.com/jellyfin/jellyfin-plugin-ldapauth/issues/22

### Code assistance
None

---

### DIFF
--- a/src/apps/legacy/routes/session/forgotPassword/index.tsx
+++ b/src/apps/legacy/routes/session/forgotPassword/index.tsx
@@ -10,6 +10,7 @@ import Input from 'elements/emby-input/Input';
 import globalize from 'lib/globalize';
 import ServerConnections from 'lib/jellyfin-apiclient/ServerConnections';
 import { getAuthenticationApi } from 'utils/sdk/authentication-api';
+import { isValidUrl } from 'utils/url';
 
 export const ForgotPasswordPage = () => {
     const navigate = useNavigate();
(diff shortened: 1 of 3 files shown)


src/apps/legacy/routes/session/forgotPassword/index.tsx
```diff
@@ -41,13 +42,19 @@ export const ForgotPasswordPage = () => {
...
-                    msg = globalize.translate('MessageForgotPasswordFileCreated');
-                    msg += '<br/><br/>';
-                    msg += globalize.translate('MessageForgotPasswordPinReset');
-                    msg += '<br/><br/>';
-                    msg += result.PinFile;
-                    msg += '<br/>';
-                    callback = () => navigate('/forgotpasswordpin');
+                    if (result.PinFile && isValidUrl(result.PinFile)) {
+                        msg += globalize.translate('MessageForgotPasswordRedirect');
+                        window.open(result.PinFile, '_blank', 'noopener,noreferrer');
```
thornbill: Should use the `openUrl` function here to support apps.

https://github.com/jellyfin/jellyfin-web/blob/8327b68236c8bee70157be3d608d829c81269512/src/scripts/shell.js#L23

thornbill (commented):
I had no clue the ldap plugin did this. 😄

thornbill (approved):

PR #8339, opened 2026-08-08
Add playback speed control for audio playback

slickwarren:
### Changes

The html audio player already implemented setPlaybackRate/getPlaybackRate and reported PlaybackRate as a supported feature, but it never exposed getSupportedPlaybackRates and no audio UI opened the rate menu, so the control was only reachable from the video OSD.

Add getSupportedPlaybackRates to the audio player and a speed button to the now playing bar and the full screen now playing screen. The button is hidden when the active player does not report PlaybackRate, so remote and cast targets do not show a dead control.

The rate menu and the rate list were previously inline in playersettingsmenu and the html video player. Both are now shared through components/playback/playbackRateMenu so the audio side does not duplicate them. The shared menu also converts the action sheet id back to a number before applying it, rather than relying on the string being coerced.

Media elements reset playbackRate to defaultPlaybackRate whenever a new source loads, so setPlaybackRate sets both. Without this the rate resets on every track, which makes the control basically useless for audiobooks.

image
image

player view:
image

### Issues

https://features.jellyfin.org/posts/1242/android-app-playback-speed-for-audio
existing PRs that didn't address comments on this feature and are stale / closed:
*  [#7281](https://github.com/jellyfin/jellyfin-web/pull/7281) 
* [#8107](https://github.com/jellyfin/jellyfin-web/pull/8107) 

### Code assistance

some code generated by LLM; also used the LLM to generate test data locally and help to ensure I followed all guidelines for contributing to this repo. 
---


slickwarren:
sorry to ping you directly but @dmitrylyzo do you know a good person to review this PR? Its a feature that's been attempted a few times in this repo and I'd like to help if possible. I'm trying to be more active in the community here.

slickwarren:
what's the best way to get something reviewed in this repo?

PR #8162, opened 2026-07-06
Fix backup list refreshing while backup jobs are active

theguymadmax:
We are currently querying for a backup list while backup jobs are running, leading to errors in the logs when the backup page is left open. 

This is a frontend fix, and a separate (independent) backend fix is in progress: https://github.com/jellyfin/jellyfin/pull/16223. 

### Changes
Don't query the backup manifest when a backup or restore is in progress. 

### Issues
Fixes: https://github.com/jellyfin/jellyfin/issues/15536

### Code assistance
N/A

---

* [x] I have read and followed the [contributing guidelines](https://github.com/jellyfin/jellyfin-web/blob/master/CONTRIBUTING.md).
* [x] I have tested these changes.
* [x] I have verified that this is not duplicating changes in an existing PR.
* [ ] I have provided a *substantive* review of [another web PR](https://github.com/jellyfin/jellyfin-web/pulls?q=is%3Apr+is%3Aopen+-label%3Adependencies+-label%3Ablocked+-label%3Abackend+-label%3Ainvalid+-label%3A"merge+conflict"+-label%3Astale+-author%3A%40me).

PR #8345, opened 2026-08-09
Fix dashboard "Scan All Libraries" button disabled without feedback

marcelmanz:
### Changes
Fixing `isScanning` condition with a proper guard before the check so it does not return also true when the `librariesTask` is undefined.

### Issues
n/a

### Code assistance
I debugged this because I am an user of jellyfin web in brave and for a while the button was disabled, so with the help of pi-coding-agent (GLM 5.2) I cloned the repo in a /tmp repository to quickly check what could be the issue. Then I did check myself that the report made sense and implemented the fix manually.

---

* [x] Review another web PR: #8162

### DIFF
--- a/src/apps/dashboard/routes/index.tsx
+++ b/src/apps/dashboard/routes/index.tsx
@@ -113,7 +113,7 @@ export const Component = () => {
                                 onScanLibrariesClick={onScanLibraries}
                                 onRestartClick={promptRestart}
                                 onShutdownClick={promptShutdown}
-                                isScanning={librariesTask?.State !== TaskState.Idle}
+                                isScanning={librariesTask && librariesTask.State !== TaskState.Idle}
                             />
                             <ItemCountsWidget />
                             <RunningTasksWidget tasks={tasks} />
--- a/src/apps/dashboard/routes/index.tsx
+++ b/src/apps/dashboard/routes/index.tsx
@@ -8,6 +8,7 @@ import { useQueryClient } from '@tanstack/react-query';
 
 import ConfirmDialog from 'components/ConfirmDialog';
 import Page from 'components/Page';
+import Toast from 'apps/dashboard/components/Toast';
 import { useApi } from 'hooks/useApi';
 import globalize from 'lib/globalize';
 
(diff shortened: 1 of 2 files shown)


Chris4942 (commented):
I was confused about what `isScanning` had to do with disabling the button so I dug through and I saw that `isScanning` is piped into `loading` on a `<Button>` and this is what `loading` on a material `Button` does:
```ts
  /**
   * If `true`, the loading indicator is visible and the button is disabled.
   * If `true | false`, the loading wrapper is always rendered before the children to prevent [Google Translation Crash](https://github.com/mui/material-ui/issues/27853).
   * @default null
   */
  loading?: boolean | null;

```
so, it does indeed make sense to me that we'd want to enable the button if there was no `libraryTask` since that would indicate that no scan is happening; however, it's still unclear to me in what case there is no `libraryTask`. Furthermore, it seems like clicking on the button wouldn't do anything anyway if there were no `libaryTask` as the `onScanLibaries` also depends on `libraryTask` (though it's just repeating the same lookup that `libraryTask` is doing.
```ts
    const onScanLibraries = useCallback(() => {
        const scanLibrariesTask = tasks?.find((value) => value.Key === 'RefreshLibrary');

        if (scanLibrariesTask?.Id) {
            startTask.mutate({
                taskId: scanLibrariesTask.Id
            });
        }
    }, [ startTask, tasks ]);
```

```ts
    const librariesTask = useMemo(() => (
        tasks?.find((value) => value.Key === 'RefreshLibrary')
    ), [ tasks ]);
```
(Unless of course the `librariesTask` memo went stale and somehow wasn't updated, which would be a whole other issue).

So I guess, the follow up to this is just the question: what would happen if there is no task with `task.Key == 'RefreshLibrary'` and you run `onScanLibrariesTask`. Would it do nothing or am I missing something?

marcelmanz:
> So I guess, the follow up to this is just the question: what would happen if there is no task with `task.Key == 'RefreshLibrary'` and you run `onScanLibrariesTask`. Would it do nothing or am I missing something?

It will do nothing, and that is not great. But I think pressing a button and seeing no effect is clearer than seeing for weeks that the button is disabled without any context. I think in my case it was the way the browser was redirecting to some wrongly cached data or something.

Reading your comment gave a second thought on it, and the best simple solution for now might be showing some feedback. I see that the already existing pattern in the repo is to show a toaster for some issues related to buttons, so I implemented that in c0c0faa978.

Given that it is a design decision I invite any maintainer to change the code as they like.

Cheers!

Chris4942:
I would have a slight preference for hover tooltip over toast because you only care about if you can run a scan if you’re trying to click the button. I’m not sure how that would be implemented in react though.

Also, I’m not a maintainer, so either one is fine with me.

PR #8349, opened 2026-08-11
Add library setting to enable directory mtime pruning during scans

jhuebel:
## Dependency

Requires jellyfin/jellyfin#17600 (`LibraryOptions.EnableDirectoryMtimePruning`) to be present on the server for the setting to have any effect. The checkbox itself is inert against older servers (the field is simply ignored).

## Summary

jellyfin-server has a new library scan optimization, directory mtime pruning (jellyfin/jellyfin#17600), that lets a library skip re-listing a folder's contents from disk when its directory hasn't changed since the last scan. It's opt-in per library (`LibraryOptions.EnableDirectoryMtimePruning`, default off) since it changes scan behavior and depends on directory mtimes updating reliably, which isn't guaranteed on some network shares (NFS/SMB).

This adds the corresponding checkbox to the shared library options editor (`libraryoptionseditor`), used by both **Add New Library** and **Edit Library**, so both flows pick it up automatically. Placed under Advanced options, next to "Enable real time monitoring". Unchecked by default, matching the server-side default.

### DIFF
--- a/src/components/libraryoptionseditor/libraryoptionseditor.js
+++ b/src/components/libraryoptionseditor/libraryoptionseditor.js
@@ -704,6 +704,7 @@ export function getLibraryOptions(parent) {
         EnableArchiveMediaFiles: false,
         EnablePhotos: parent.querySelector('.chkEnablePhotos').checked,
         EnableRealtimeMonitor: parent.querySelector('.chkEnableRealtimeMonitor').checked,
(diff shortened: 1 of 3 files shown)


jhuebel:
See my comment on the server PR about LLM usage. Unfortunate that this PR and the server PR will be tossed aside for that reason. It's a useful optimization for Jellyfin users who have very large libraries.

PR #8354, opened 2026-08-13
Add: slider markers now jump to chapter timecodes when clicked

LuisThe0ne:
### Changes

Slider markers in the player progress bar now jump to their associated timecode when clicked and get larger when hovered over.
Hovered:
image

### Issues

none
### Code assistance

Used ChatGPT for research

---

### DIFF
--- a/src/apps/legacy/controllers/playback/video/index.js
+++ b/src/apps/legacy/controllers/playback/video/index.js
@@ -148,8 +148,8 @@ export default function (view) {
             const maxWidth = window.screen.width * window.devicePixelRatio * 0.2;
             for (const [, info] of Object.entries(trickplayResolutions)) {
                 if (!bestWidth
-                        || (info.Width < bestWidth && bestWidth > maxWidth) // Objects not guaranteed to be sorted in any order, first width might be > maxWidth.
-                        || (info.Width > bestWidth && info.Width <= maxWidth)) {
+                    || (info.Width < bestWidth && bestWidth > maxWidth) // Objects not guaranteed to be sorted in any order, first width might be > maxWidth.
+                    || (info.Width > bestWidth && info.Width <= maxWidth)) {
                     bestWidth = info.Width;
                 }
             }
@@ -964,7 +964,7 @@ export default function (view) {
 
                 // show subtitle offset feature only if player and media support it
                 const showSubOffset = playbackManager.supportSubtitleOffset(player)
-                        && playbackManager.canHandleOffsetOnCurrentSubtitle(player);
+                    && playbackManager.canHandleOffsetOnCurrentSubtitle(player);
 
                 playerSettingsMenu.show({
                     mediaType: 'Video',
@@ -1138,10 +1138,10 @@ export default function (view) {
             * - primary subtitle has support
             */
         const currentTrackCanAddSecondarySubtitle = playbackManager.playerHasSecondarySubtitleSupport(player)
-                && streams.length > 1
-                && secondaryStreams.length > 0
-                && currentIndex !== -1
-                && playbackManager.trackHasSecondarySubtitleSupport(playbackManager.getSubtitleStream(player, currentIndex), player);
+            && streams.length > 1
+            && secondaryStreams.length > 0
+            && currentIndex !== -1
+            && playbackManager.trackHasSecondarySubtitleSupport(playbackManager.getSubtitleStream(player, currentIndex), player);
 
         if (currentTrackCanAddSecondarySubtitle) {
             const secondarySubtitleMenuItem = {
@@ -1926,10 +1926,15 @@ export default function (view) {
 
     nowPlayingPositionSlider.getMarkerInfo = function () {
         // use markers based on chapters
-        return currentItem?.Chapters?.map(currentChapter => ({
+        return currentItem?.Chapters?.map((currentChapter, index) => ({
             name: currentChapter.Name,
-            progress: currentChapter.StartPositionTicks / currentItem.RunTimeTicks
+            progress: currentChapter.StartPositionTicks / currentItem.RunTimeTicks,
+            index: index
         })) || [];
+    };
+
+    nowPlayingPositionSlider.seekToChapter = function (index) {
+        playbackManager.seek(currentItem?.Chapters[index].StartPositionTicks, currentPlayer);
     };
 
     view.querySelector('.btnPreviousTrack').addEventListener('click', function () {
--- a/src/apps/legacy/controllers/playback/video/index.js
+++ b/src/apps/legacy/controllers/playback/video/index.js
@@ -1927,9 +1927,9 @@ export default function (view) {
     nowPlayingPositionSlider.getMarkerInfo = function () {
         // use markers based on chapters
         return currentItem?.Chapters?.map((currentChapter, index) => ({
+            id: index,
             name: currentChapter.Name,
-            progress: currentChapter.StartPositionTicks / currentItem.RunTimeTicks,
-            id: index
+            progress: currentChapter.StartPositionTicks / currentItem.RunTimeTicks
         })) || [];
     };
 
--- a/src/apps/legacy/controllers/playback/video/index.js
+++ b/src/apps/legacy/controllers/playback/video/index.js
@@ -1929,12 +1929,12 @@ export default function (view) {
         return currentItem?.Chapters?.map((currentChapter, index) => ({
             name: currentChapter.Name,
             progress: currentChapter.StartPositionTicks / currentItem.RunTimeTicks,
-            index: index
+            id: index
         })) || [];
     };
 
-    nowPlayingPositionSlider.seekToChapter = function (index) {
-        playbackManager.seek(currentItem?.Chapters[index].StartPositionTicks, currentPlayer);
+    nowPlayingPositionSlider.onMarkerClicked = function (id) {
+        playbackManager.seek(currentItem?.Chapters[id].StartPositionTicks, currentPlayer);
     };
 
     view.querySelector('.btnPreviousTrack').addEventListener('click', function () {
--- a/src/elements/emby-slider/emby-slider.js
+++ b/src/elements/emby-slider/emby-slider.js
@@ -260,12 +260,20 @@ function updateMarkers(range, currentValue) {
             range.markerInfo = newMarkerInfo;
 
             let markersHtml = '';
-            range.markerInfo.forEach(() => {
-                markersHtml += '<span class="sliderMarker" aria-hidden="true"></span>';
+            range.markerInfo.forEach((marker) => {
+                markersHtml += `<span id="sliderMarker-${marker.index}" class="sliderMarker" aria-hidden="true"></span>`;
             });
             range.markerContainerElement.innerHTML = markersHtml;
 
             range.markerElements = range.markerContainerElement.querySelectorAll('.sliderMarker');
+
+            range.markerElements.forEach(element => {
+                element.addEventListener('click', event => {
+                    const index = event.currentTarget.id.replace('sliderMarker-', '');
+                    range.seekToChapter(index);
+                    return true;
+                });
+            });
         }
     }
 
--- a/src/elements/emby-slider/emby-slider.js
+++ b/src/elements/emby-slider/emby-slider.js
@@ -190,7 +190,7 @@ function updateValues(isValueSet) {
             backgroundLower.style.width = fraction + '%';
         }
 
-        if (range.markerContainerElement) {
+        if (range.markerOuterContainerElement) {
             updateMarkers(range, value);
         }
     });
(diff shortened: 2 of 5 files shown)


src/elements/emby-slider/emby-slider.js
```diff
@@ -260,12 +260,20 @@ function updateMarkers(range, currentValue) {
             range.markerInfo = newMarkerInfo;
 
             let markersHtml = '';
-            range.markerInfo.forEach(() => {
-                markersHtml += '<span class="sliderMarker" aria-hidden="true"></span>';
+            range.markerInfo.forEach((marker) => {
+                markersHtml += `<span id="sliderMarker-${marker.index}" class="sliderMarker" aria-hidden="true"></span>`;
```
dmitrylyzo: Maybe use `data-index` instead of `id`? Then you won't need to cut the string.

src/elements/emby-slider/emby-slider.js
```diff
@@ -260,12 +260,20 @@ function updateMarkers(range, currentValue) {
...
             });
             range.markerContainerElement.innerHTML = markersHtml;
 
             range.markerElements = range.markerContainerElement.querySelectorAll('.sliderMarker');
+
+            range.markerElements.forEach(element => {
+                element.addEventListener('click', event => {
+                    const index = event.currentTarget.id.replace('sliderMarker-', '');
+                    range.seekToChapter(index);
```
dmitrylyzo: Technically, markers can be not only chapters. So I would rename `seekToChapter` to `onMarkerClicked`.

src/apps/legacy/controllers/playback/video/index.js
```diff
@@ -1926,12 +1926,17 @@ export default function (view) {
 
     nowPlayingPositionSlider.getMarkerInfo = function () {
         // use markers based on chapters
-        return currentItem?.Chapters?.map(currentChapter => ({
+        return currentItem?.Chapters?.map((currentChapter, index) => ({
             name: currentChapter.Name,
-            progress: currentChapter.StartPositionTicks / currentItem.RunTimeTicks
+            progress: currentChapter.StartPositionTicks / currentItem.RunTimeTicks,
+            index: index
```
dmitrylyzo: If you use `index` only as an offset in the array, then you don't need to explicitly assign it - get the index during marker generation (`range.markerInfo.forEach(...`).

On the other hand, you can use `index` as an identifier (`chapter-1`, `note-3`, `resumepos`), so that the slider owner can perform various actions. Then, yes, it should be assigned in `getMarkerInfo` and probably named `id`. But I'm not sure if we need this feature.

src/elements/emby-slider/emby-slider.scss
```diff
@@ -261,12 +261,17 @@
 }
 
 .sliderMarker {
+    z-index:10;
     position: absolute;
     width: 2px;
     height: 12px;
     transform: translate3d(0, 25%, 0);
 }
```
dmitrylyzo: Add `cursor: pointer;` to indicate the ability to perform a click.

src/apps/legacy/controllers/playback/video/index.js
```diff
@@ -148,8 +148,8 @@ export default function (view) {
             const maxWidth = window.screen.width * window.devicePixelRatio * 0.2;
             for (const [, info] of Object.entries(trickplayResolutions)) {
                 if (!bestWidth
-                        || (info.Width < bestWidth && bestWidth > maxWidth) // Objects not guaranteed to be sorted in any order, first width might be > maxWidth.
-                        || (info.Width > bestWidth && info.Width <= maxWidth)) {
+                    || (info.Width < bestWidth && bestWidth > maxWidth) // Objects not guaranteed to be sorted in any order, first width might be > maxWidth.
+                    || (info.Width > bestWidth && info.Width <= maxWidth)) {
```
dmitrylyzo, line 152: Ideally, don't change any unrelated lines of code, unless it violates the style settings.

src/elements/emby-slider/emby-slider.scss
```diff
@@ -261,12 +261,17 @@
...
 
+.sliderMarker:hover {
+    transform: scale(1.5);
```
dmitrylyzo, line 278: `scale` conflicts with the calculations of `left`.
https://github.com/jellyfin/jellyfin-web/blob/02e21c830328dcafff1ab5db7a2cb13082971775/src/elements/emby-slider/emby-slider.js#L243
The marker jumps because when hovered, `markerRect.width` is x1.5 larger than normal.

To fix this, we probably need to wrap each marker (icon) with another container, which we will position. This way we also can use `translate3d(-50%, ...` instead of `calc(...)`.

To test this, set marker width to, say, 100px.

LuisThe0ne: Thank for pointing that out, i thought about it too, but in my testing the marker sizes in a uniform way
[Screencast_20260813_134842.webm](https://github.com/user-attachments/assets/42672418-a923-477e-9cdc-2ec9fe03a372)

src/apps/legacy/controllers/playback/video/index.js
LuisThe0ne: I decided to use id as a more general way in case some functionality in the future requires the markers to use an identifier for jumping.

src/apps/legacy/controllers/playback/video/index.js
LuisThe0ne: Thanks for reminding me, I read it in the contributor guide and double checked my files but I missed it.

src/elements/emby-slider/emby-slider.scss
dmitrylyzo: > Thank for pointing that out, i thought about it too, but in my testing the marker sizes in a uniform way [Screencast_20260813_134842.webm](https://github.com/user-attachments/assets/42672418-a923-477e-9cdc-2ec9fe03a372)

Click it and see.
_I should have written "when hovered and clicked"._

LuisThe0ne: You're totally right :)
I added an additional outer container and now it looks like this:
[Screencast_20260813_145818.webm](https://github.com/user-attachments/assets/02040e77-c258-4007-abf0-a6502dbeacc3)

```css
.sliderMarkerContainer {
    z-index:10;
    position: absolute;
    width: 10px;
    height: 12px;
    transform: translate3d(0, 25%, 0);
}

.sliderMarker {
    width: 100%;
    height: 100%;
    cursor: pointer;
    display: block;
}

.sliderMarker:hover {
    transform: scale(1.5);
}

```

and the container is aligned via calc()

LuisThe0ne (commented):
added container to prevent jumping

src/elements/emby-slider/emby-slider.js
```diff
@@ -260,19 +262,28 @@ function updateMarkers(range, currentValue) {
             range.markerInfo = newMarkerInfo;
 
             let markersHtml = '';
-            range.markerInfo.forEach(() => {
-                markersHtml += '<span class="sliderMarker" aria-hidden="true"></span>';
+            range.markerInfo.forEach((marker) => {
+                markersHtml += `<div class="sliderMarkerContainer" aria-hidden="true"><span data-id="${marker.id}" class="sliderMarker" aria-hidden="true"></span></div>`;
```
dmitrylyzo: AFAIK, you don't need `aria-hidden` for `div`.
```suggestion
                markersHtml += `<div class="sliderMarkerContainer"><span data-id="${marker.id}" class="sliderMarker" aria-hidden="true"></span></div>`;
```

src/elements/emby-slider/emby-slider.scss
```diff
@@ -253,20 +253,32 @@
...
     right: 0;
     margin: 0 0.54em; /* half of slider thumb size */
 }
 
-.sliderMarker {
+.sliderMarkerContainer {
+    z-index:10;
     position: absolute;
-    width: 2px;
+    width: 10px;
```
dmitrylyzo: revert debug
```suggestion
    width: 2px;
```

src/apps/legacy/controllers/playback/video/index.js
```diff
@@ -1926,12 +1926,17 @@ export default function (view) {
 
     nowPlayingPositionSlider.getMarkerInfo = function () {
         // use markers based on chapters
-        return currentItem?.Chapters?.map(currentChapter => ({
+        return currentItem?.Chapters?.map((currentChapter, index) => ({
             name: currentChapter.Name,
-            progress: currentChapter.StartPositionTicks / currentItem.RunTimeTicks
+            progress: currentChapter.StartPositionTicks / currentItem.RunTimeTicks,
+            id: index
```
dmitrylyzo: Less changes.
Bonus: alphabetical order and `id` as "primary key" 😄 
```suggestion
        return currentItem?.Chapters?.map((currentChapter, index) => ({
            id: index,
            name: currentChapter.Name,
            progress: currentChapter.StartPositionTicks / currentItem.RunTimeTicks
```

src/elements/emby-slider/emby-slider.scss
```diff
@@ -253,20 +253,32 @@
...
+.sliderMarkerOuterContainer {
     position: absolute;
     left: 0;
     right: 0;
     margin: 0 0.54em; /* half of slider thumb size */
 }
 
-.sliderMarker {
+.sliderMarkerContainer {
+    z-index:10;
```
dmitrylyzo:
```suggestion
    z-index: 10;
```

dmitrylyzo (commented):
Looks good. Maybe we only need a good names for `sliderMarkerOuterContainer`, `sliderMarkerContainer` and `sliderMarker`, but nothing comes to mind at the moment.

src/elements/emby-slider/emby-slider.js
```diff
@@ -260,19 +262,28 @@ function updateMarkers(range, currentValue) {
...
+            range.markerElements = range.markerOuterContainerElement.querySelectorAll('.sliderMarker');
 
-            range.markerElements = range.markerContainerElement.querySelectorAll('.sliderMarker');
+            range.markerElements.forEach(element => {
+                element.addEventListener('click', event => {
+                    const id = event.currentTarget.dataset.id;
+                    range.onMarkerClicked(id);
+                    return true;
+                });
+            });
```
dmitrylyzo: I overlooked this: we shouldn't add listeners if `onMarkerClicked` isn't set. Also, we need another class with `cursor: pointer`, which we will add along with the listeners.

src/elements/emby-slider/emby-slider.js
```diff
@@ -230,17 +230,19 @@ function updateBubble(range, percent, value, bubble) {
...
 
         if (!bubbleTrackRect.width || !markerRect.width) {
             // width is not set, most probably because the OSD is currently hidden
             return;
         }
 
-        marker.style.left = `calc(${valueMarker}% - ${markerRect.width / 2}px)`;
+        markerContainer.style.left = `calc(${valueMarker}% - ${markerRect.width / 2}px)`;
+
+        const marker = markerContainer.getElementsByClassName('sliderMarker')[0];
```
dmitrylyzo:
```suggestion
        const marker = markerContainer.querySelector('.sliderMarker');
```

dmitrylyzo (commented):
Tbf, 2px marker is hard to click. 😅
As an option, we can make the clickable (_with `cursor: pointer` class_) marker wider. But this is up to UI/UX team.

In addition, we could make this feature (clickable markers) available only when Ctrl is pressed.

LuisThe0ne:
I am not quite sure what benefit it would bring to only make the makers clickable when ctrl is pressed. In that case it is beneficial that the markers are only 2px wide since they do not really interrupt skipping to a specific position by "grabbing" the mouse when trying to skip to nearby timecodes. 😅

dmitrylyzo:
> I am not quite sure what benefit it would bring to only make the makers clickable when ctrl is pressed. In that case it is beneficial that the markers are only 2px wide since they do not really interrupt skipping to a specific position by "grabbing" the mouse when trying to skip to nearby timecodes. 😅

The idea with Ctrl is to use it as a "snap to" function, as well as to increase marker size for easier clicking (we can increase the "marker container" and listen clicks on it instead of the "marker tick"). This way you can switch between "free" and "snapped" modes.

On the other hand, there's no Ctrl on mobile devices, and it will be hard to tap a 2px marker. :shrug:

Anyway, UI/UX probably needs to be discussed.

LuisThe0ne:
I do not quite think that would be ideal, since new users will probably not think about pressing ctrl for clicking on a marker. I think it would be better to increase the scaling on hover, to make it easier to click, while the marker resizing when hovered is probably enough of a hint that there is associated functionality. On mobile a ui that lists the chapters is probably better suited. (as proposed in #5011).

PR #8365, opened 2026-08-16
Fix still watching prompt being unreachable by remote on TV

moontwister:
### Changes

On TV layout the "Still watching?" prompt cannot be operated with a remote. The video OSD handles keys on `document` and input commands on `window` without checking for an open dialog, so while the OSD is hidden the arrow keys take the TV branch that focuses the seek slider behind the dialog and calls `preventDefault()` — focus leaves the prompt and navigation never reaches it again.

This guards both handlers with `getOpenedDialog()`, the check the file already uses in `resetIdle` and `onWheel` to stop the player reacting behind an open dialog, and drops the now-unreachable check in the `back` and `Escape` cases.

### Issues

None — I could not find an existing report.

### Code assistance

LLM-assisted. I hit the bug on my own TV and described the symptom; the assistant traced the event chain through the code to the cause, proposed the guard and wrote the change. I checked the diagnosis against the source, built it on `master`, and tested it on my LG C4: the prompt works, and the subtitle and audio menus still navigate and close normally.

---

### DIFF
--- a/src/apps/legacy/controllers/playback/video/index.js
+++ b/src/apps/legacy/controllers/playback/video/index.js
@@ -410,6 +410,9 @@ export default function (view) {
     function onInputCommand(e) {
         const player = currentPlayer;
 
+        // Ignore commands when some dialog is opened
+        if (getOpenedDialog()) return;
+
         switch (e.detail.command) {
             case 'left':
                 if (currentVisibleMenu === 'osd') {
@@ -464,8 +467,7 @@ export default function (view) {
                 break;
 
             case 'back':
-                // Ignore command when some dialog is opened
-                if (currentVisibleMenu === 'osd' && !getOpenedDialog()) {
+                if (currentVisibleMenu === 'osd') {
                     hideOsd();
                     e.preventDefault();
                 }
@@ -1222,6 +1224,9 @@ export default function (view) {
         // Skip modified keys
         if (isKeyModified) return;
 
+        // Ignore keys when some dialog is opened
+        if (getOpenedDialog()) return;
+
         const key = keyboardnavigation.getKeyName(e);
 
         const btnPlayPause = osdBottomElement.querySelector('.btnPause');
@@ -1280,8 +1285,7 @@ export default function (view) {
                 break;
             case 'Escape':
             case 'Back':
-                // Ignore key when some dialog is opened
-                if (currentVisibleMenu === 'osd' && !getOpenedDialog()) {
+                if (currentVisibleMenu === 'osd') {
                     hideOsd();
                     e.stopPropagation();
                 }


src/apps/legacy/controllers/playback/video/index.js
```diff
@@ -410,6 +410,9 @@ export default function (view) {
     function onInputCommand(e) {
         const player = currentPlayer;
 
+        // Ignore commands when some dialog is opened
+        if (getOpenedDialog()) return;
```
Adamstuhltrager: Confirmed that this lets right and left hit the dialog on TV layout

src/apps/legacy/controllers/playback/video/index.js
```diff
@@ -464,8 +467,7 @@ export default function (view) {
                 break;
 
             case 'back':
-                // Ignore command when some dialog is opened
-                if (currentVisibleMenu === 'osd' && !getOpenedDialog()) {
+                if (currentVisibleMenu === 'osd') {
```
Adamstuhltrager: dropped dialog check is redundant now

src/apps/legacy/controllers/playback/video/index.js
```diff
@@ -1222,6 +1224,9 @@ export default function (view) {
         // Skip modified keys
         if (isKeyModified) return;
 
+        // Ignore keys when some dialog is opened
+        if (getOpenedDialog()) return;
```
Adamstuhltrager, line 1228: This line fixes the reported case. On master with TV layout and the OSD hidden, the first Right sent focus from the prompt to the seek slider behind it, and no further key got it back. With this guard, Right and Left move between the two buttons and Continue watching advances to the next episode. It also stops desktop layout from seeking the video and focusing the rewind and fast-forward buttons while the prompt is open.

Additionally it exposes a remaining problem.  Once the handler steps aside the key goes to keyboardNavigation and then focusManager. If focus has already left the dialog element as it's source (i.e. the remote has clicked somewhere other than a button.) focusManager gets a dialog element as its source and looks for a button to the right of a fullscreen rect. As this will not return a button the focus stays on the body and the prompt becomes stranded again.  This is preexisting but since the goal is for it to work with your remote I'd like to see the fix folded in.  Details and the change in the summary.

thornbill:
Invalid per llm policy.

Adamstuhltrager (commented):
I reviewed this by reproducing on a webpack dev build against a 10.11.8 server, Chrome browser, layout set to TV, keyboard arrows standing in for a remote.  to trigger i lowered the default threshold locally to 5 seconds and 1 item.  

On Master - with the OSD hidden and the prompt open focus starts on Stop watching.  The first right moves it to the seek slider behind the dialog and it can not be brought back. Desktop layout has a similar issue with a different flavor as arrow keys seek the video and focus the rewind and fast-forward buttons behind the prompt.  

This Branch - Right and Left move between the two buttons up and down stay put and continue watching starts the next episode.  The guard sits above both code paths and therefore fixes the desktop case as well.  Back and Escape still work correctly: with OSD visible and no prompt Escape hides the OSD; with prompt open back with focus on prompt button and escape with focus outside the prompt both close it and stop playback.

Open issue - if focus leaves the dialog while it's open, no arrow gets back into the dialog. this was tested three different ways, blurring the focused button, clicking the prompt text, and clicking elsewhere on the fullscreen dialog.  The LG and similar TVs are capable of all three of these.  Back still works from this state so it isn't a hard lock but the only way out is to stop watching.  The chain:

1. `inputManager.handleCommand` sees focus outside the open dialog and reroutes the command to the dialog element.
2. `focusManager.nav` calls `focusableParent` on it, finds no focusable ancestor, and gets the dialog back as the source.
3. The geometric search then looks for a button to the right of the source rect. The TV dialog is fullscreen, so nothing qualifies, and focus stays on `body`.

On Master this is hidden behind the OSD stealing focus first.  With this PR it becomes the remaining way to lose the prompt from a remote, which is the case the PR sets out to fix which is why I'm requesting it be included in the fix.

This change to `focusManager.nav` fixed every case above when I tested it on top of this branch:

```diff
     container = container || (activeElement ? getFocusContainer(activeElement, direction) : getDefaultScope());
 
-    if (!activeElement || activeElement == document.body) {
+    if (!activeElement || activeElement == document.body || !isFocusable(activeElement)) {
         autoFocus(container, true, false);
         return;
     }
```

LLM assistance was used to quickly replicate and confirm existing PR issue and fix.  Additional testing of the confirmed remaining issue and again LLM assistance was used in diagnosing the remaining fix. All write-ups are mine.

PR #8370, opened 2026-08-17
Feature: Preload Next Queued If Audio 

ryco117:
Reduce playback gaps between consecutive audio files by preloading the next queued item if it is also audio. This isn't stricly a gapless audio playback, but it is significantly better than the existing behavior, producing results without a perceivable gap.

Meant to address discussion #7568.

### Changes

Adds a new playback setting to control whether this mechanic should be enabled, defaulting to false. When enabled, playback starts will check if the current and next queued items are audio and if so it will preload the next item so it is ready for playback at the end of the current track. Otherwise, existing behavior is preserved.

Before:
image

After:
image

### Code assistance
Used Copilot CLI and Copilot VS Code plugin for initial prototyping and some bug fixes.

## TODO
* [ ] Remove console debugging prints
* [ ] Refactor so that `invalidatePreloadedTrackIfNextChanged()` isn't using `player._` ("private") properties.
* [ ] Refactor to a cleaner preload state instead of multiple new, and related, preload properties on the player. (Honestly, made the code more complicated, might not prefer this).
* [ ] Avoid race condition issues by using proper async/awaiting on completion of preload step when switching to next track
* [x] Check if `onShuffleQueueModeChange` affects next queued item (and invalidate preload if so).
* [ ] Current implementation discards preloaded track when "Next track" is pressed due to integration into `HtmlAudioPlayer self.stop`. Should be able to leverage preloaded track when skipping to next track directly.

---


src/components/playback/playbackmanager.js
```diff
@@ -3512,7 +3534,61 @@ export class PlaybackManager {
...
+                            setPlaylistState(nextItem.item.PlaylistItemId, nextItem.index);
+
+                            player.activatePreloadedTrack().then(({ item, mediaSource }) => {
+                                // Wire up the new stream info now that the element is playing.
+                                const playerData = getPlayerData(player);
+                                playerData.streamInfo = createStreamInfo(
+                                    ServerConnections.getApiClient(item.ServerId),
+                                    'Audio', item, mediaSource, 0, player
+                                );
+                                playerData.streamInfo.playbackStartTimeTicks = new Date().getTime() * 10000;
```
ryco117, line 3554: `new Date().getTime()` seems to be what the other `playbackStartTimeTicks` setters are using; perhaps a future PR should update them all together for consistent behavior. (In reference to the analysis warning suggesting `Date.now()` be used)

src/plugins/htmlAudioPlayer/plugin.js
```diff
@@ -115,6 +124,8 @@ class HtmlAudioPlayer {
             let val = options.url;
             console.debug('playing url: ' + val);
             import('../../scripts/settings/userSettings').then((userSettings) => {
```
ryco117, line 153: **Observation:** This asynchronous reading of `userSettings` is not tied to the returned `enableHlsPlayer` promise. This is implicitly a race condition. I did not fix because I want this change to be as close to a no-op as I can for users that are not opted in, but let me know what you think,

PR #8390, opened 2026-08-26
Improve Live TV guide cell readability

loister72:
## Changes
- Adds deterministic fallback text for Live TV program cells with missing titles.
- Adds program cell title and aria-label text including program, channel, and time range.
- Tightens program cell text layout so long titles truncate more reliably.

## Validation
- 
- 

Full npm validation was not run locally because this checkout requires Node >=24/npm >=11, and the local shell has Node 20/npm 10. Additionally,  reports the upstream package lock is currently out of sync with package.json.

### DIFF
--- a/src/components/guide/guide.js
+++ b/src/components/guide/guide.js
@@ -374,6 +374,21 @@ function Guide(options) {
         return datetime.getDisplayTime(date).toLowerCase();
     }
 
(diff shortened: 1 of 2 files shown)


thornbill:
- invalid per llm policy

PR #8391, opened 2026-08-26
Declare npm package manager version

loister72:
## Changes
- Adds packageManager metadata for npm 11.17.0, matching the repository's Node 24/npm 11 engine requirement.

## Why
This helps Corepack-aware environments and contributors select the expected package manager version instead of accidentally using npm from an older Node install.

## Validation
- env PATH="/opt/homebrew/opt/node@24/bin:$PATH" npm install --package-lock-only --ignore-scripts
- env PATH="/opt/homebrew/opt/node@24/bin:$PATH" npm run build:check
- git diff --check

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "Web interface for Jellyfin",
   "repository": "https://github.com/jellyfin/jellyfin-web",
   "license": "GPL-2.0-or-later",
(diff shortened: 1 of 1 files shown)


thornbill:
- invalid per llm policy

PR #8392, opened 2026-08-26
Add bundle size reporting script

loister72:
## Changes
- Adds npm run build:report to summarize built asset counts, total size, and largest files in dist.
- Uses a dependency-free Node script so it works with the current Webpack output and can continue to work during a future Vite migration.

## Why
Jellyfin Web already emits Webpack size warnings, but modernization work needs a repeatable summary that can be copied into PRs or CI logs.

## Validation
- env PATH="/opt/homebrew/opt/node@24/bin:$PATH" npm run build:report
- env PATH="/opt/homebrew/opt/node@24/bin:$PATH" node --check scripts/report-bundle-size.mjs
- env PATH="/opt/homebrew/opt/node@24/bin:$PATH" npm run build:check
- git diff --check


thornbill:
- invalid per llm policy
- this type of reporting already exists

PR #8404, opened 2026-08-30
Add global localized alphabet navigation

kostazol:
### Changes

Adds the Jellyfin Web client-side implementation for localized alphabet navigation described in jellyfin/jellyfin#17733.

This depends on the corresponding server-side API changes in:

jellyfin/jellyfin#<SERVER_PR_NUMBER>

The feature is opt-in and disabled by default.

When disabled, the existing alphabet picker and legacy `# / A-Z` filtering behavior remain unchanged.

When enabled, Jellyfin Web:

- selects the primary alphabet from the server UI locale;
- displays the locale's native alphabet instead of forcing navigation through transliterated `A-Z`;
- keeps `#` as the `Other` bucket;
- optionally displays an additional Latin `A-Z` alphabet;
- sends the native alphabet order to the server so groups remain in their actual alphabet order rather than the order produced by transliterated `SortName`;
- applies the same localized navigation to the separate Genres view;
- falls back to the existing legacy picker for unsupported locales and API paths.

For example, with Greek as the UI locale:

```text
# Α Β Γ Δ Ε Ζ Η Θ Ι Κ Λ Μ Ν Ξ Ο Π Ρ Σ Τ Υ Φ Χ Ψ Ω
```

A Greek title such as:

```text
Ωμέγα
```

is navigated through:

```text
Ω
```

instead of only through its transliterated `SortName`:

```text
omega → O
```

If Latin is additionally enabled, a second alphabet is displayed:

```text
Greek:
# Α Β Γ ... Ψ Ω

Latin:
A B C ... Y Z
```

so:

```text
Ωμέγα → Ω
Omega → O
```

The localized `#` button represents items whose native initial does not belong to any enabled alphabet.

#### Configuration

Two global controls are added to Dashboard → General, directly below the existing display-language setting:

- `Use localized alphabet navigation`
- `Include Latin alphabet (A–Z)`

These settings are server-wide rather than stored separately for each media library.

The primary alphabet follows the existing Jellyfin `UICulture`.

Changing the Latin option only adds or removes the `Latn` script and preserves any other configured additional-script values.

#### Supported alphabets

The client currently defines native alphabets for:

- Greek
- Russian
- Ukrainian
- Belarusian
- Bulgarian
- Latin

For locales whose alphabet is not explicitly supported, Jellyfin Web retains the existing legacy alphabet navigation instead of guessing an incorrect alphabet.

#### API behavior

The Web client uses the new server query primitives:

```text
nameInitials
excludeNameInitials
nameInitialSortOrder
```

`nameInitialSortOrder` is sent even when no individual letter is selected so the complete item list is grouped in native alphabet order.

Selecting a native letter sends `nameInitials`.

Selecting `#` sends `excludeNameInitials` for all enabled alphabet initials.

The existing `nameStartsWith` / `nameLessThan` behavior is preserved for the legacy path.

Authors and other endpoints that do not support the new server primitives continue to use the existing legacy behavior.

#### SDK compatibility

All API requests continue to use the Jellyfin TypeScript SDK endpoint clients.

The currently published generated SDK does not yet expose the new server fields and query parameters from the companion server PR, so this change temporarily:

- uses narrow local interfaces for the new configuration/SystemInfo properties;
- passes the new query parameters through the SDK method's `AxiosRequestConfig.params`.

These compatibility bridges can be removed once a generated SDK containing the companion server API changes is available.

#### Localization

Only the source-language `en-us` translation file is changed in this PR.

Translations for other languages should be added through Weblate according to the Jellyfin Web contribution guidelines.

#### Testing

Validated with the same quality checks used by Jellyfin Web CI:

```text
npm run lint
npm run stylelint
npm run build:check
npm run test
npm run build:es-check
```

The tests cover:

- legacy behavior while the feature is disabled;
- Greek alphabet ordering;
- localized `#` / `Other`;
- additional Latin navigation;
- native initial ordering without a selected letter;
- Cyrillic alphabet ordering, including locale-specific letters;
- unsupported-locale fallback;
- SystemInfo configuration mapping;
- unsupported API paths remaining on the legacy behavior.

The implementation was also manually tested against a mixed-script media library together with the companion server changes.

### Issues

Related to jellyfin/jellyfin#17733

Depends on jellyfin/jellyfin#17742

### Code assistance

OpenAI Codex was used to assist with implementation, codebase analysis, debugging, test generation, and review of the changes.

The resulting changes were manually reviewed and validated with the Jellyfin Web lint, type-check, test, production-build, and browser-compatibility checks.

---

### DIFF
--- a/src/apps/dashboard/routes/settings/index.tsx
+++ b/src/apps/dashboard/routes/settings/index.tsx
@@ -26,6 +26,11 @@ import { getSystemApi } from '@jellyfin/sdk/lib/utils/api/system-api';
 import { queryClient } from 'utils/query/queryClient';
 import { ActionData } from 'types/actionData';
 
+interface LocalizedAlphabetConfiguration {
+    EnableLocalizedAlphabetNavigation?: boolean;
+    LocalizedAlphabetAdditionalScripts?: string[] | null;
+}
+
 export const action = async ({ request }: ActionFunctionArgs) => {
     const api = ServerConnections.getApi();
     if (!api) throw new Error('No Api instance available');
@@ -35,6 +40,15 @@ export const action = async ({ request }: ActionFunctionArgs) => {
 
     config.ServerName = formData.get('ServerName')?.toString();
     config.UICulture = formData.get('UICulture')?.toString();
+    const localizedAlphabetConfig = config as typeof config & LocalizedAlphabetConfiguration;
+    localizedAlphabetConfig.EnableLocalizedAlphabetNavigation =
+        formData.get('EnableLocalizedAlphabetNavigation')?.toString() === 'on';
+    const additionalScripts = (localizedAlphabetConfig.LocalizedAlphabetAdditionalScripts ?? [])
+        .filter(script => script !== 'Latn');
+    localizedAlphabetConfig.LocalizedAlphabetAdditionalScripts =
+        formData.get('IncludeLatinAlphabet')?.toString() === 'on' ?
+            [...additionalScripts, 'Latn'] :
+            additionalScripts;
     config.CachePath = formData.get('CachePath')?.toString();
     config.MetadataPath = formData.get('MetadataPath')?.toString();
     config.QuickConnectAvailable = formData.get('QuickConnectAvailable')?.toString() === 'on';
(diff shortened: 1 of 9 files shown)


thornbill:
Invalid per llm policy

PR #8412, opened 2026-09-02
Remove 40 char search restriction

enter-a-random-username:
### Changes
Remove restriction
- Search field is long enough
- Removing it and standard search works same as before (and validate on server, never trust user input)
- Allows searching for a full title of a show
- With search providers more complex queries are possible. Some started to support sematic search

### Issues
-

### Code assistance
NO


josephdburdick (approved):
LGTM
- Nothing downstream depended on the cap. The query flows through the URL parameter into the debounced queries, and every search hook passes the term straight to the SDK with no truncation or fixed-size handling.
- The input element just spreads props onto a native input, so removing the attribute changes nothing else.
- The cap was never a real invariant anyway. The on-screen alphabet picker appends characters through the change handler, and maxLength only constrains direct typing into a controlled input, so that path already bypassed it.
- The empty-results message and the global-search fallback link both handle arbitrary-length text safely.
- The only theoretical downside is a pasted query of several kilobytes producing a long request line and a server-side rejection. That degrades to a failed search, not incorrect behavior. Worth mentioning as a non-blocker if you want to be thorough.
- CI is green and Sonar reports no new issues.

dkanada (approved):

Issue #3575, opened 2022-04-19
 [10.8.0] Books / Comics do not update progress until page reload


MinecraftPlaye:
**Describe The Bug**
When one opens a book or comic, scrolls through it some pages and closes it, the progress get's saved. But when the user immediately opens the book again, the old progress is used. Only after e.g. navigating to the main page of JF and going back in the book library, the new progress is used.
**It's important to note here that "old progress" does not necessarily mean page 0!** 

It looks to me like the reader plugins for comics and books only request the progress of the book on page load and not when the book is opened. I haven't tested it with pdfs, but I can certainly confirm this problem for cbz (Comic Book Archive) and EPUB files.

**Steps To Reproduce**
1. Open a book or comic.
2. Scroll through it and then close it.
3. Open it again.
4. See error :-1: 

---

1. Open a book or comic.
2. Scroll through it and then close it.
3. Go to the main page of Jellyfin.
4. Go back into your book library
5. Open the book
6. See that the book opens at the correct position. :+1: 

**Expected Behavior**
You can close the book and open it immediately and Jellyfin brings you back to the correct position.

**System (please complete the following information):**
- Platform: Linux
- Browser: Firefox 99.0.1
- Jellyfin + Jellyfin Web Version: 10.8.0 Beta 2

MinecraftPlaye:
Can confirm for

- Platform: Fedora 36
- Browser: Firefox 103.0.2
- Jellyfin Server jellyfin/jellyfin@98f6485cdda7c6b0ca07007fe913da2b8bfbc673
- Jellyfin Web Version: 10.8.4 jellyfin/jellyfin-web@bb040b90d12c2e4f4c8d4cfc43e31ccbf1a55cf5

carif:
Can confirm that it's still an issue for:

- Platform: Linux
- Jellyfin Server: 10.8.7
- Jellyfin Web Version: 10.8.7
- Jellyfin Media Player: 1.7.1

carif:
Can confirm that it's still an issue for:

- Platform: Linux
- Jellyfin Server: 10.8.10
- Jellyfin Web Version: 10.8.10
- Jellyfin Media Player: 1.9.1

PR #8421, opened 2026-09-05
Refresh cached resume positions for EPUB books and comics

daniel-stilman:
### Changes

Resuming an EPUB book or comic from a cached details screen can reuse an outdated saved position when a progress notification is missed. Reusing that position can overwrite newer progress already saved on the server. When the requested resume position matches the cached nonzero value, this change fetches the latest user progress before opening the reader. Starts from the beginning and other explicit positions are preserved, and the latest saved position may move backwards. A failed progress lookup stops the resume attempt instead of reusing stale progress.

### Issues

Related to #3575.

### Code assistance

Codex was used to investigate the bug, implement the fix, and write and run regression tests.

---

* [ ] I have provided a *substantive* review of [another web PR](https://github.com/jellyfin/jellyfin-web/pulls).

### DIFF
--- a/src/plugins/bookPlayer/plugin.js
+++ b/src/plugins/bookPlayer/plugin.js
@@ -1,3 +1,4 @@
+import { getBookResumePosition } from '../../utils/bookPlayerResume';
 import { getLibraryApi } from '@jellyfin/sdk/lib/utils/api/library-api';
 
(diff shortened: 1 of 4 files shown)


thornbill:
Invalid per llm guidelines and did not read/follow contributing guidelines.

PR #8425, opened 2026-09-07
Add option to sort unwatched items first

ZaimMarzuki:
### Changes
Add an "Unplayed first" toggle in the sort menu (`libraryBrowser.showSortMenu`).
When active, it prepends `IsPlayed` with `Ascending` sort order to the query, placing unwatched media at the front while keeping watched media visible below instead of hiding them completely.
Also updates `userSettings` to persist the preference and adds the localization string in `en-us.json`.

### Issues
N/A

### Code assistance
Implementation assisted by Claude Code.

---

### DIFF
--- a/src/scripts/libraryBrowser.js
+++ b/src/scripts/libraryBrowser.js
@@ -83,34 +83,16 @@ export function getQueryPagingHtml (options) {
 export function showSortMenu (options) {
     Promise.all([
         import('../components/dialogHelper/dialogHelper'),
-        import('../elements/emby-radio/emby-radio')
+        import('../elements/emby-radio/emby-radio'),
+        import('../elements/emby-checkbox/emby-checkbox')
     ]).then(([{ default: dialogHelper }]) => {
-        function onSortByChange() {
-            const newValue = this.value;
+        const isUnplayedFirst = options.query.UnplayedFirst === true ||
+            (options.query.SortBy || '').startsWith('IsPlayed,');
+        const rawSortBy = (options.query.SortBy || '').replace(/^IsPlayed,/, '');
 
-            if (this.checked) {
-                const changed = options.query.SortBy != newValue;
-                options.query.SortBy = newValue.replace('_', ',');
-                options.query.StartIndex = 0;
-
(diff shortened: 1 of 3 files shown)


josephdburdick (changes requested):
A few issues, the most important beginning here: https://github.com/jellyfin/jellyfin-web/pull/8425/changes#diff-5eaf35ab20e3ba8851a4c3262c2fe2b21092b651f73fb12983a25fe7c03108c6R160

**List-view index headers break.** 
The PR prefixes the sort string with `IsPlayed`, and that same string is handed to the list view as `sortBy`. The index-header logic matches at position zero, using `indexOf('sortname') === 0` and the same for the rating and album-artist variants. I confirmed both halves: the controllers for Movies, Collections, TV Shows, Episodes, and Albums all pass the query sort straight through, and the matching really is anchored at position zero. So turning on the toggle in list view sorted by name makes the A/B/C group headers vanish. 

The fix is either stripping the prefix before it reaches the list view, or teaching the index function to skip a leading `IsPlayed`.

**A non-API field gets sent to the server on every request**
The new `UnplayedFirst` key lives on the same object passed to the items endpoint, and the client serializes every key into the query string. It is also redundant, since the prefix in the sort string already encodes the state. Deriving the checkbox from the sort string alone removes both problems and makes the settings change unnecessary.

**The checkbox can persist as enabled while doing nothing**
When no sort radio is selected, the guard skips adding the prefix but the flag is still stored. Reopening shows a ticked box with no matching sort term. That state is reachable in Songs, where the default sort does not match any menu option on first open.

**The prefixed values get written to server-side display preferences, which are shared across clients.** 
Any older or other `jellyfin-web` build reading them will not recognize the new key but will still read the prefixed sort and order strings, and its sort dialog will open with neither radio selected. The user has to re-pick a sort to escape.

**The new `enableUnplayedSort` parameter is dead configuration.** 
Confirmed it does not exist on master and the PR touches only three files, none of them controllers, so none of the six call sites can pass it. It defaults to on, which means the toggle also appears for Collections, where play state does not apply.

PR #8438, opened 2026-09-08
Reduce search item limit to match other search queries

josephdburdick:
### Code assistance

---

* [ ] I have verified that this is not duplicating changes in an existing PR.

### DIFF
--- a/src/apps/legacy/features/search/api/useSearchItems.ts
+++ b/src/apps/legacy/features/search/api/useSearchItems.ts
@@ -74,8 +74,7 @@ export const useSearchItems = (
                     includeItemTypes: itemTypes,
                     parentId,
                     searchTerm,
-                    isMissing: itemTypes.includes(BaseItemKind.Episode) && !user?.Configuration?.DisplayMissingEpisodes ? false : undefined,
-                    limit: 800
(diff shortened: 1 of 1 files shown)


enter-a-random-username:
Queries should be split, because with new search adapter and fussy or sematic search, the matches can increase. Now you have 1000 high scoring episodes and no more movie or show will be returned.

Personally I split them and every slider get 35 items max, because the mage is already a scroller downwards and adding all sideward scroller and 100 items - nobody will use it. Fells like google 10 pages. Most will refine the search.

Why 35? On 4k screen you will get around 10-12 episodes. That times 3. I would give up after first scroll and add a keyword or so.

This increased the speed a lot, because it won't add so many DOM Nodes nobody will every look at, server needs to load less data and so on

josephdburdick:
I am closing this due to the insights provided by @enter-a-random-username [here](https://github.com/jellyfin/jellyfin-web/pull/8438#issuecomment-5591582534). 

The reasoning is as follows:
- this PR cuts a single shared cap but the cap is not per row. The combined request asks the server for all twelve item types at once and the code buckets the results by type afterward in the browser. Consequently every row draws from 1 pool.
- this means lowing the number does not simply fetch less of each thing, it changes which rows get anything at all

PR #8443, opened 2026-09-09
Add mouse wheel volume control to the audio player

Lastent:
The video player already adjusts the volume when the wheel is scrolled, but the audio player does not. Bind a wheel handler to the volume slider in the now playing bar and in the full screen player so that both behave the same way.

The handler reuses playbackManager.volumeUp/volumeDown, so the direction and the step size always match the video player. It is bound to the slider itself rather than to the document, so it does not interfere with page scrolling or with keyboard navigation elsewhere in the app.

### Changes
* The mouse wheel now changes the volume when it is rolled over the volume control in the now playing bar while playing music.
* This also works in the Now Playing view (/queue).
* Doesn't include any keyboard action.
* This change imitates the video player behavior.

### Issues
This is ralated to the PR #5614 but this one doesn't touch the keyboard actions.

### Code assistance
I used LLMs to help me search for ways to implement this feature using similar behaviors/code in the project and testing it.
---


thedreaddpirate:
Appears to work as described.

https://github.com/user-attachments/assets/418cae74-1167-4f8a-9f3a-ad76792fdbd4

PR #8478, opened 2026-09-12
Show the book OSD when navigating with the keyboard

jesse-anderson:
### Changes
Changes to add an eventlistener on keydown, similar to webvideo, so the OSD stays visible as you page through books.
Paging already works by keyboard, the OSD just can't be recovered at all once it auto hides, so you're moving through a long PDF with no idea where you are.

Progress bar in #8320 works and was what I was after, among other things, for web interface. It's gated on the same visibility flag, so by keyboard it can't be seen either.

Tested: .pdf | .epub | .cbz

### Issues
None

### Code assistance
None

---

### DIFF
--- a/src/plugins/bookPlayer/BookOsd/BookOsd.tsx
+++ b/src/plugins/bookPlayer/BookOsd/BookOsd.tsx
@@ -92,15 +92,22 @@ const BookOsd: FC<BookOsdProps> = ({
             setVisible(state => !state);
         };
 
+        const onKeyDown = () => {
+            scheduleHide();
+            setVisible(true);
+        };
+
         scheduleHide();
         document.addEventListener('pointermove', onPointerMove);
         document.addEventListener('click', onClick);
+        document.addEventListener('keydown', onKeyDown);
 
         return () => {
             clearTimeout(timeout.current);
             updateFullscreen(false);
             document.removeEventListener('pointermove', onPointerMove);
             document.removeEventListener('click', onClick);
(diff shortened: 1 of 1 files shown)


src/plugins/bookPlayer/BookOsd/BookOsd.tsx
```diff
@@ -92,15 +92,22 @@ const BookOsd: FC<BookOsdProps> = ({
...
 
+        const onKeyDown = () => {
+            scheduleHide();
+            setVisible(true);
+        };
+
         scheduleHide();
         document.addEventListener('pointermove', onPointerMove);
         document.addEventListener('click', onClick);
+        document.addEventListener('keydown', onKeyDown);
```
MBR-0001, line 103: this doesn't work on epubs when the focused element is the epub.js frame itself

(bookPlayer/plugin.js -> forwardEvents function only forwards click and pointer events)

MBR-0001 (changes requested):
I have mixed feelings about this but it need fixing first

src/plugins/bookPlayer/BookOsd/BookOsd.tsx
jesse-anderson: Good catch, thanks. I pushed a follow up commit that forwards keydown from the iframe in forwardEvents; dispatched as a custom 'bookplayerkeydown' event. Happy to hear more about the mixed feelings if there's a direction you'd prefer for this.

PR #8491, opened 2026-09-16
feat(trickplay): add option to restrict trickplay to primary media (cleaned up)

TheDMV2:
**Basic intention:** only generate trickplay for the main movie/show, not for every trailer, DVD extra etc, etc. Will vastly speed up trickplay generation (because fewer files need to be created) and will use less disk space.

(This is the web front-end for the server-side change).


thornbill:

TheDMV2:
Valid as per llm policy.

And I'm happy to handle and review feedback.

PR #8501, opened 2026-09-17
Fix Pause key handling on TVs

### LINKED ISSUE #8484: Regression in 12.0: physical Pause key no longer works on Samsung Tizen

### Describe the bug

After updating a Samsung Tizen client from Jellyfin Web 10.11.x to a build using Jellyfin Web 12.0, the physical Pause command no longer pauses playback.

Other playback controls still work:

- Physical Play works.
- The on-screen Pause button works.
- Sending Play to the TV works.
- Sending Pause to the TV (for example Home Assistant "media_player.media_pause", which sends Samsung "KEY_PAUSE") does not work while Jellyfin is playing.

This appears to be a regression in "src/scripts/keyboardNavigation.js".

Regression

In "release-10.11.z", "KeyNames" contains:

13: 'Enter',
19: 'Pause',
27: 'Escape',
32: 'Space',

and the handler contains:

case 'Pause':
    inputManager.handleCommand('pause');
    break;

In "v12.0", the mapping for key code "19" was removed:

const KeyNames = {
    13: 'Enter',
    27: 'Escape',
    37: 'ArrowLeft',
    ...

but the "case 'Pause'" handler is still present.

"MediaPlay" remains explicitly mapped as key code "415", which appears to explain why the physical Play command still works while Pause does not.

### Reproduction Steps

1. Use a Samsung Tizen TV with a Jellyfin Tizen build using Jellyfin Web 12.0.
2. Start video playback.
3. Press the physical Play key: playback responds correctly.
4. Press the physical Pause key: nothing happens.
5. Use the on-screen Pause control: playback pauses correctly.
6. Optionally send Samsung "KEY_PAUSE" externally, for example through Home Assistant: it also has no effect in Jellyfin.
7. Sending "KEY_PLAY" externally still works.

### Expected/Actual behaviour

The physical Pause key should pause playback as it did with Jellyfin Web 10.11.x.

### Logs

_No response_

### Server version

12.0

### Web version

12.0

### Build version

12.0

### Platform

Samsung Tizen TV UE65TU7172UXXH

### Browser

Samsung Tizen embedded WebView (Jellyfin for Tizen)

### Additional information

Suggested fix

Restore the removed mapping:

19: 'Pause',

in "KeyNames".

Related change

PR #7975 restored several "KeyNames" after navigation changes. During review, "Pause" and "Space" were considered unnecessary and were not restored.

For Samsung/Tizen remotes, "Pause" appears to still be required

Environment

- Jellyfin Server: 12.0
- Jellyfin Web in Tizen build: v12.0
- Client: Jellyfin for Tizen
- TV: Samsung UE65TU7172UXXH
- Regression confirmed by comparing "release-10.11.z" and "v12.0" source.

Valley-15:
### Changes

Fixes handling of the physical Pause key on TV platforms.

In Jellyfin Web 12.0, the `Pause` key mapping (`keyCode: 19`) was removed from `src/scripts/keyboardNavigation.js`. The existing keyboard navigation handler still explicitly handles the `Pause` key, but `getKeyName()` could no longer resolve `keyCode: 19` to `Pause` on affected platforms.

This caused physical Pause buttons on some TV remotes to stop working while other playback controls continued to work.

During investigation, the following was found:

- Jellyfin Web 10.11.x mapped `19` to `Pause` in `keyboardNavigation.js`.
- Commit `ab2cf99f2b` removed the `19: 'Pause'` mapping as part of the controller-navigation changes in PR #7975.
- The `Pause` handler itself was not removed and still calls `inputManager.handleCommand('pause')`.
- The reported regression affects Samsung Tizen TVs, where the physical Pause button no longer pauses playback while the on-screen Pause control and physical Play button continue to work.
- The same regression has also been reported for LG webOS TVs after upgrading to Jellyfin Web 12.0.
- The controller-navigation change that introduced the mapping was specifically concerned with gamepad key mappings. `Pause` and `Space` were considered unnecessary for the gamepad mapping, but the `Pause` entry was also being used by the general keyboard/TV input handling.
- `getKeyName()` resolves a key using `event.keyCode` first, followed by `event.code`. Restoring the `19: 'Pause'` entry therefore restores the expected key resolution for environments where the physical remote reports keyCode 19 without a usable `event.code`.
- The legacy KeyboardEvent polyfill also contains the `19: 'Pause'` mapping, which further indicates that keyCode 19 is an intentional/known representation of the Pause key for legacy browser environments supported by Jellyfin Web.

The fix restores:

    19: 'Pause',

to the `KeyNames` mapping in `src/scripts/keyboardNavigation.js`.

No unrelated behavior or refactoring is included.

### Issues

Fixes #8484

### Code assistance

LLM assistance was used during development for codebase exploration, debugging guidance, and understanding the existing keyboard-event and gamepad input handling.

The change itself was reviewed and implemented manually. The investigation included tracing the `getKeyName()` flow, checking the `KeyNames` and `KeyAliases` mappings, examining the existing `Pause` handler, inspecting the legacy keyboard-event polyfill, and reviewing the history of the removed mapping and the PR that introduced its removal.

### Verification

The following checks were run successfully:

- `npm run lint -- src/scripts/keyboardNavigation.js`
- `npm run build:check`
- `git --no-pager diff --check`
- `npm test`

The full test suite passed with 162 tests passing.

---

- [ ] I have provided a *substantive* review of [another web PR](https://github.com/jellyfin/jellyfin-web/pulls?q=is%3Apr+is%3Aopen+-label%3Adependencies+-label%3Ablocked+-label%3Abackend+-label%3Ainvalid+-label%3A%22merge+conflict%22+-label%3Astale+-author%3A%40me).


dmitrylyzo:
`Pause` should be handled by the polyfill. 🤔
https://github.com/jellyfin/jellyfin-web/blob/b54188953dafa4d2f81eeecfa0bb281ac572f777/src/lib/legacy/keyboardEvent.js#L54

Navigation keys were reverted due to the implementation of gamepadtokey.js

dmitrylyzo:
Found the root of the problem: in some versions of Tizen, `KeyboardEvent` has `key` and `code`, and the polyfill doesn't override them, but their values are invalid.

We need to (somehow) apply the polyfill for such cases. Otherwise, some keys will still be unavailable.

It seems that in Tizen 4 and 5, other keys (numeric and alphabetic) have the correct `key` and `code` values. 😵‍💫

src/scripts/keyboardNavigation.js
```diff
@@ -13,6 +13,7 @@ import appSettings from './settings/appSettings';
  */
 const KeyNames = {
     13: 'Enter',
+    19: 'Pause',
```
dmitrylyzo, line 16: It is probably worth adding a comment like `On some platforms (Tizen 4/5), Pause has incorrect key and code values.`

As for webOS, `Pause` works fine on webOS 4/5 (without polyfill). If you know which version of webOS is affected, please add that information as well.

dmitrylyzo (commented):
Since the only key with invalid values is `Pause` (tested on Tizen 4/5), reverting its removal is an acceptable fix of the regression.

Please rebase your branch onto `release-12.z` and change the target branch of PR.

thornbill:

PR #8509, opened 2026-09-17
Add Play Next media segment actions

Car-Role:
### Changes

Adds two new media segment actions, `PlayNext` and `AskToPlayNext`, alongside the existing `Skip` / `AskToSkip`.

Today an outro segment can only be skipped, which seeks to the end of the segment and leaves the remaining seconds of the item to play out. These actions instead advance to the next item in the queue:

- **Play Next** advances immediately when the segment is reached.
- **Ask To Play Next** shows the existing segment button, labelled "Play Next Episode", which calls `nextTrack()` when clicked.

The selected action is now passed alongside the segment on the `PromptSkip` event, so `SkipSegment` knows which label and click behavior to use. The extra argument is optional and the other `PromptSkip` listener ignores it, so existing behavior is unchanged when no action is supplied.

Both new actions are only offered for **Outro** segments in playback settings, since advancing the queue is not meaningful for an intro, recap, preview, or commercial. `PlayNext` reuses the same "seeked back into" guard as `skipSegment`, so scrubbing backwards into an outro will not repeatedly advance the queue.

One behavior worth flagging for review: `SkipSegment.onPromptSkip` already returns early when the segment runs to the end of the item and the Up Next overlay is enabled, to avoid showing two prompts at once. That guard applies to `AskToPlayNext` as well, so for outros that extend to the very end of an episode the Up Next overlay continues to handle the transition and no button is shown. That seemed like the right precedence to preserve rather than change in this PR, but I'm happy to revisit it.

### Issues

None — opened as a feature request rather than a fix for a filed issue.

### Code assistance

This change was written with Claude Code (Claude Opus 5). It was used to trace the media segment code path, write the implementation, and draft this description. The `tsc --noEmit` typecheck, `eslint` on the touched files, and the `vitest` suite (162 tests) were run locally and all pass; the two `sonarjs/fixme-tag` lint warnings on `playbackmanager.js` and `skipsegment.ts` are pre-existing and unrelated.

---

> **Note on the testing checkbox:** this is deliberately left unchecked. The change typechecks, lints, and passes the unit suite, but I have not yet exercised it in a browser against a live server with detected outro segments. I'll report back here once I have, and I'd suggest not merging on my word until then. In particular I have not verified how the "watched" state behaves when advancing this way, given #7039.


thornbill:

Issue #8379, opened 2026-08-19
Multiple gamepads cause erratic UI navigation


MBach:
### This issue respects the following points:

- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-web/issues?q=is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

When two or more gamepads are connected (e.g. two DualShock 4 controllers over Bluetooth), the TV layout navigation becomes completely erratic and unusable. Inputs are repeated, directions get stuck, and the UI becomes unresponsive to intentional input.

The root cause is in `gamepadtokey.js`: the polling function iterates over all connected gamepads via `navigator.getGamepads()`, but writes to a single set of shared state variables for button/axis tracking. When two controllers are present, both overwrite the same pressed/released state on every animation frame, causing conflicting signals.

### Reproduction Steps

1. Connect two game controllers (tested with two DS4 controllers over Bluetooth on Linux)
2. Enable TV display mode and gamepad support
3. Attempt to navigate the UI using the D-pad on either controller
4. Navigation is erratic: selections jump, directions appear stuck, inputs repeat uncontrollably

Disconnecting one controller immediately restores normal behavior.

### Expected/Actual behaviour

With two gamepads connected, only one should control the UI navigation (e.g. the first connected gamepad), or each gamepad should maintain independent state tracking.

In the minified gamepadtokey.js chunk, the polling function loops over all gamepads:

```javascript
function B() {
    for (var e = navigator.getGamepads(), n = 0, a = e.length; n < a; n++) {
        var t = e[n];
        if (t) {
            // processes axes and buttons for EVERY gamepad
            // but writes to shared state variables (d, u, s, f, o, i, c, r...)
        }
    }
    O = requestAnimationFrame(B);
}
```

Each gamepad overwrites the same state variables on every frame. With two controllers, even minor stick drift (which is normal on analog sticks) from both controllers creates rapid state toggling that makes the UI unusable.

### Proposed Fix

The simplest fix is to only process the first connected gamepad for UI navigation:

```javascript
if (t && n === 0) {  // only process gamepad at index 0
```

### A more complete solution could include:

- A user setting to select which gamepad index controls the UI
[...]

PR #8522, opened 2026-09-19
Fix navigation with multiple gamepads

dmitrylyzo:
### Changes
Remember last used gamepad.

### Issues
Fixes #8379 

### Code assistance
None

---


thornbill (commented):
Might be better to identify the gamepad by its `id` property?

dmitrylyzo:
> Might be better to identify the gamepad by its `id` property?

I would say that `index` is better because it is unique. `id` contains information about the model and, iirc, is the same for all controllers of the same model.

I’m not entirely sure, but it seems that the gamepads retain their position in the array, at least within the current browser session. Therefore, I think we can simply use the array index. Besides, we don’t store it for that long (key press/up, stick movement).

> https://developer.mozilla.org/en-US/docs/Web/API/Navigator/getGamepads
> Elements in the array may be `null` if a gamepad disconnects during a session, so that the remaining gamepads retain the same index.

thornbill (approved):

PR #8526, opened 2026-09-20
Fix blurhash to image transition

heyhippari:
### Changes

When transitioning from the BlurHash to the image, the component would remove the BlurHash once the image is loaded, but before it was actually faded in.

Instead, we now wait until the transition is done, then we remove the BlurHash.

### Issues

Did not find any.

### Code assistance

None.

---


thedreaddpirate:
Please confirm the intent/description with what I am observing with this PR vs 12.1.

With 12.1 there is a moment where a poster is black between the removal of the blur hash and the image actually rendering.  This results in a sort of "blinking" affect.

This PR appears to remove that gap, resulting in no "blinking".

If that matches your intent, fantastic.  Definitely a subtle change at first ,but now I can't unsee the blinking as images load.

Side note, when I checked out the PR I tested it as is on master and also when rebased with release-12.z.  Both worked as intended.

heyhippari:
> This PR appears to remove that gap, resulting in no "blinking".

That was indeed the intent: have the transition occur from the blurhash smoothly over to the image, without blinking.

PR #8531, opened 2026-09-21
Improve discoverability of server management and user menu

heyhippari:
### Changes

Rewords the "Dashboard" button to "Manager Server", to more accurately reflect what is behind the action.

Also changes the icon from a dashboard to something that better signals the intent of the menu entry.

Adds a dropdown arrow to clearly indicate to users that the avatar can be interacted with to reveal a menu.

image

### Issues

None, but mentioned in user feedback in https://github.com/orgs/jellyfin/discussions/18019

### Code assistance

None.

---


nielsvanvelzen:
There's more then just settings, you can also see active sessions and logs etc. So perhaps "Manage server" might make more sense

heyhippari:
> There's more then just settings, you can also see active sessions and logs etc. So perhaps "Manage server" might make more sense

I agree, that makes even more sense, I updated it :+1:

nielsvanvelzen (approved):

heyhippari:
Sorry for updating this after approval, but I felt there were still a few improvements to the UX that could be made :p

thornbill:
Not a designer, but my first impressions are:

- I'm not sure the icon is a great choice as it's really similar to the "Select Server" one... not sure what would be better though :sweat_smile: 
- It would probably make sense to update "Metadata Manager" to "Manage Metadata" to be consistent with "Manage Server"
- The down arrow on the user avatar is kind of weird... we don't use it on any other icon that opens a drop down and it makes the hover/active highlight an oval

heyhippari:
> I'm not sure the icon is a great choice as it's really similar to the "Select Server" one... not sure what would be better though 😅

Yeah, I looked at what was available and this seemed "the most okay", but nothing really fits... The previous dashboard didn't really convey that there was more behind that menu entry than, well, a dashboard. This one at least feels somewhat tangentially related to "server" visually...

> It would probably make sense to update "Metadata Manager" to "Manage Metadata" to be consistent with "Manage Server"

I agree, I thought about it when making the other round of changes, but I got lazy lol I'll update that tomorrow.

> The down arrow on the user avatar is kind of weird... we don't use it on any other icon that opens a drop down and it makes the hover/active highlight an oval

Yeah, the idea was essentially that the user menu is just a round image, which doesn't exactly convey "click me, I have stuff you might want". It's a bit of an awkward solution currently.

